PR #5, opened 2022-04-15
Localizations Setup

### LINKED ISSUE #4: Reuse strings from Element-Android
We want to reuse them for several reasons:
- Android is better for supporting plurals
- Better parity between mobile apps
- Less work for translators

We want to rely entirely on them. This means the iOS developer will have to:
- add strings first on the Android project and PR them
- run a script to retrieve from the EA repo

This script could be run like xcodegen when you set up the project.

The final strategy is still to be defined. As part of this research project, we need to define what will be the best (or least bad) in term of devx.

Open questions:
 - [ ] Should those generated strings be in git? 
 - [ ] ...

ismailgulek:
Fixes #4


ElementX/Sources/AppDelegate.swift
```diff
@@ -11,6 +11,15 @@ import UIKit
 class AppDelegate: UIResponder, UIApplicationDelegate {
     
     private var appCoordinator: AppCoordinator!
+
+    func application(_ application: UIApplication, willFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]? = nil) -> Bool {
+
+        //  fixme: Use `Bundle.elementLanguage = ".."` when we have the functionality
+        //  use `en` as fallback language
+        Bundle.elementFallbackLanguage = "en"
```
stefanceriu, line 19: This feels like something that should be in the app coordinator. We should try to keep the AppDelegate as slim as possible.

stefanceriu (changes requested):
The commits don't look quite right. Merges and rebases don't go well together so you'll probably need to cherry pick on top of a clean base. You might also want to squash some of these 22 commits.

I'm out till next week so I'll reassign to @pixlwave.

Screenshot 2022-04-19 at 08 33 53

ismailgulek:
Will create another PR with a clean history.

PR #6, opened 2022-04-19
Localizations Setup

### LINKED ISSUE #4: Reuse strings from Element-Android

ismailgulek:
Fixes #4


Tools/SwiftGen/Templates/Strings/structured-swift5-element.stencil
```diff
@@ -0,0 +1,108 @@
...
+// MARK: - Implementation Details
+
+extension {{enumName}} {
+  static func tr(_ table: String, _ key: String, _ args: CVarArg...) -> String {
+    var set = Set<String>()
+    let languages = ([Bundle.elementLanguage] +
+      Locale.preferredLanguages +
+      [Bundle.elementFallbackLanguage])
+      .compactMap { $0 }
+      .filter { set.insert($0).inserted }
```
pixlwave, line 85: This will iterate through the languages twice. Given this function will be called a lot, it would probably make sense to either add a `.lazy` or combine the `filter` into the `compactMap` so we only need to go through the list once.

Tools/SwiftGen/Templates/Strings/structured-swift5-element.stencil
```diff
@@ -0,0 +1,108 @@
...
+
+    for language in languages {
+      let translation = trIn(language, table, key, args)
+      if translation != key {
```
pixlwave, line 89: I wonder if this check could fail in some instances for a key that is a single word. In EI the `NSLocalizedString` is using `_` for its `value` parameter, which might be more robust in this situation I wonder?

Equally could be over thinking things here, feel free to ignore 🙈

pixlwave (commented):
Hurray, this looks great! I love how simple the fallback language support is compared to the only MXKLanguageBundle implementation 😄 <del>A few</del> Various thoughts I have:

- It would be nice to use `Localizable.strings` instead of `Element.strings` now that we have a single set of strings and so wouldn't need to worry about passing a table name during lookup (also SwiftUI has some lovely magic when using `Localizable.strings` that we wouldn't be using here, but maybe could use in the future).
- `LocaleProvider.swift` looks to be unused to my eyes?
- It seems Android has the concept of regional variations of languages (e.g. the `en-rGB` locale). Does iOS handle this as expected?
- Is there are way we could generate `stringsdict`s for pluralisations instead of the strange `plurals.loc_key_name.item_x` keys that it creates? It would be nice to handle https://github.com/vector-im/element-ios/issues/5991 correctly in EIX if it could.
- Tiny preference here: SwiftPM now uses a folder structure of `Sources/` and `Resources/`. Could we do the same and use `Resources/Assets.xcassets` instead of `Assets/Assets.xcassets`?
- iOS now supports setting custom languages on a per-app basis. It would be nice if we simply supported the current locale and a fallback to `en`, leaving the rest to the Settings app instead. WDYT about us simplifying our implementation like that?
IMG_0375

ElementX/Sources/Other/Bundle.swift
```diff
@@ -0,0 +1,49 @@
...
+    /// - Returns: The lproj bundle if found otherwise the `app` bundle.
+    @objc static func lprojBundle(for language: String) -> Bundle? {
+        guard let lprojURL = Bundle.app.url(forResource: language, withExtension: "lproj") else { return .app }
+        return Bundle(url: lprojURL)
+    }
+
+    /// Whether or not the bundle is the ElementShareExtension.
+    var isShareExtension: Bool {
+        bundleURL.lastPathComponent.contains("ElementShareExtension.appex")
+    }
```
pixlwave, line 38: We should probably remove this for now as 🤞 we won't be using an exception to terminate the extension so wouldn't need to check it.

ismailgulek:
> * It would be nice to use `Localizable.strings` instead of `Element.strings` now that we have a single set of strings and so wouldn't need to worry about passing a table name during lookup (also SwiftUI has some lovely magic when using `Localizable.strings` that we could maybe use in the future).

Done 👍

> * `LocaleProvider.swift` looks to be unused to my eyes?

Yeah, not used anymore. Removed.

> * It seems Android has the concept of regional variations of languages (e.g. the `en-rGB` locale). Does iOS handle this as expected?

Yes, we can check for some specific values though. There might be some differences between iOS and Android, but at least Xcode not complaining about it.

> * Is there are way we could generate `stringsdict`s for pluralisations instead of the strange `plurals.loc_key_name.item_x` keys that it creates? It would be nice to handle [vector-im/element-ios#5991](https://github.com/vector-im/element-ios/issues/5991) correctly in EIX if it could.

We've decided not to use stringsdict for plurals. Actually generated keys make sense to me, but there are some weird keys like `item_x` in Android translations. We should work together with the Android team to improve them i think.

> * Tiny preference here: SwiftPM now uses a folder structure of `Sources/` and `Resources/`. Could we do the same and use `Resources/Assets.xcassets` instead of `Assets/Assets.xcassets`?

Done 👍

> * iOS now supports setting custom languages on a per-app basis. It would be nice if we simply supported the current locale and a fallback to `en`, leaving the rest to the Settings app instead. WDYT about us simplifying our implementation like that?

Makes perfect sense, but i'd like to keep the `Bundle.elementLanguage` at least for testing purposes.

Tools/SwiftGen/Templates/Strings/structured-swift5-element.stencil
ismailgulek: I've added `Bundle.preferredLanguages` to hide this complexity.

ElementX/Sources/Other/Bundle.swift
ismailgulek: 👍

Tools/SwiftGen/Templates/Strings/structured-swift5-element.stencil
ismailgulek: We're not using user readable string keys in Element translations. Shouldn't be an issue. We can check specific keys if we have issues afterwards.

Tools/Scripts/localizer.sh
```diff
@@ -0,0 +1,59 @@
...
+		attranslate "${XML_TO_STRINGS[@]}" --srcFile=$XML_FILE --targetFile=$STRINGS_FILE --srcLng=$LANG_CODE --targetLng=$LANG_CODE
+		sed -i '' 's/""/"/g' $STRINGS_FILE
+		sed -i '' 's/\\";/\\"";/g' $STRINGS_FILE
+		sed -i '' 's/%s/%@/g' $STRINGS_FILE
+		sed -i '' 's/$s/$@/g' $STRINGS_FILE
+		sed -i '' 's/= ";/= "";/g' $STRINGS_FILE
+		sed -i '' 's/"\.";/\.";/g' $STRINGS_FILE
+		sed -i '' 's/"bot"/\\"bot\\""/g' $STRINGS_FILE
+		sed -i '' 's/\${app_name}/%@/g' $STRINGS_FILE
+		sed -i '' 's/\$ {app_name}/%@/g' $STRINGS_FILE
```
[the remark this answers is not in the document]
ismailgulek: Good idea 👍

ElementX/Sources/Other/Bundle.swift
```diff
@@ -0,0 +1,49 @@
...
+
+public extension Bundle {
+
+    /// Returns the real app bundle.
+    /// Can also be used in app extensions.
+    static let app: Bundle = {
+        let bundle = main
+        if bundle.bundleURL.pathExtension == "appex" {
+            // Peel off two directory levels - MY_APP.app/PlugIns/MY_APP_EXTENSION.appex
+            let url = bundle.bundleURL.deletingLastPathComponent().deletingLastPathComponent()
```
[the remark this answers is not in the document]
ismailgulek: It seems similar but this didn't work in a Catalyst app extension. Thanks for the heads up.
The proper way seems to add our `Localizable.strings` files into `Copy Bundle Resources` phase in an extension. Will do it when we have one.

Tools/SwiftGen/Templates/Strings/structured-swift5-element.stencil
```diff
@@ -77,12 +77,7 @@ import Foundation
 
 extension {{enumName}} {
   static func tr(_ table: String, _ key: String, _ args: CVarArg...) -> String {
-    var set = Set<String>()
-    let languages = ([Bundle.elementLanguage] +
-      Locale.preferredLanguages +
-      [Bundle.elementFallbackLanguage])
-      .compactMap { $0 }
-      .filter { set.insert($0).inserted }
+    let languages = Bundle.preferredLanguages
```
pixlwave, line 80: Nice 😎

Tools/Scripts/localizer.sh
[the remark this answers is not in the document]
pixlwave: Thanks 🙏

pixlwave (commented):
> > * It seems Android has the concept of regional variations of languages (e.g. the `en-rGB` locale). Does iOS handle this as expected?
> 
> Yes, we can check for some specific values though. There might be some differences between iOS and Android, but at least Xcode not complaining about it.

Do you think we should at least match the formatting specified in the [docs](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/LanguageandLocaleIDs/LanguageandLocaleIDs.html) for now? E.g. `en-rGB` → `en_GB`? FWIW, Xcode wasn't complaining about `pr.lproj` but it was never available as a language and then ASC generated a warning during upload.

> > * Is there are way we could generate `stringsdict`s for pluralisations instead of the strange `plurals.loc_key_name.item_x` keys that it creates? It would be nice to handle [vector-im/element-ios#5991](https://github.com/vector-im/element-ios/issues/5991) correctly in EIX if it could.
> 
> We've decided not to use stringsdict for plurals. Actually generated keys make sense to me, but there are some weird keys like `item_x` in Android translations. We should work together with the Android team to improve them i think.

Hrm, that sounds like it will be very manual. For example in `sk` we have
```
"plurals.notice_room_aliases_added.item_0"
"plurals.notice_room_aliases_added.item_1"
"plurals.notice_room_aliases_added.item_2"
```
whereas in `en` we only have
```
"plurals.notice_room_aliases_added.item_0"
"plurals.notice_room_aliases_added.item_1"
```
So in code we're going to need check both the count and the language to decide which key we need to use (even if we rename them to `one`, `few`, `many` like they show in the Android keys) 😕

ElementX/Resources/Localizations/sk.lproj/Localizable.strings
```diff
@@ -0,0 +1,2385 @@
...
+"location_share_option_user_live" = "Zdieľať polohu v reálnom čase";
+"a11y_location_share_option_user_current_icon" = "Zdieľať moju aktuálnu polohu";
+"location_share_option_user_current" = "Zdieľať moju aktuálnu polohu";
+"a11y_location_share_locate_button" = "Priblížiť k aktuálnej polohe";
+"a11y_location_share_pin_on_map" = "Označiť vybranú lokalitu na mape";
+"threads_notice_migration_message" = "Blížime sa k vydaniu verejnej beta verzie pre vlákna.
+\n
+\nV rámci príprav na ňu musíme urobiť niekoľko zmien: vlákna vytvorené pred týmto bodom sa budú zobrazovať ako bežné odpovede.
+\n
+\nPôjde o jednorazový prechod, keďže vlákna sú teraz súčasťou špecifikácie Matrix.";
```
pixlwave, line 2243: Sorry, also notice this whilst looking at the plurals. May need a `sed` tweak to fix this.

ismailgulek: Handled in the new script.

ismailgulek:
> > > * It seems Android has the concept of regional variations of languages (e.g. the `en-rGB` locale). Does iOS handle this as expected?
> > 
> > 
> > Yes, we can check for some specific values though. There might be some differences between iOS and Android, but at least Xcode not complaining about it.
> 
> Do you think we should at least match the formatting specified in the [docs](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/LanguageandLocaleIDs/LanguageandLocaleIDs.html) for now? E.g. `en-rGB` → `en_GB`? FWIW, Xcode wasn't complaining about `pr.lproj` but it was never available as a language and then ASC generated a warning during upload.
> 
> > > * Is there are way we could generate `stringsdict`s for pluralisations instead of the strange `plurals.loc_key_name.item_x` keys that it creates? It would be nice to handle [vector-im/element-ios#5991](https://github.com/vector-im/element-ios/issues/5991) correctly in EIX if it could.
> > 
> > 
> > We've decided not to use stringsdict for plurals. Actually generated keys make sense to me, but there are some weird keys like `item_x` in Android translations. We should work together with the Android team to improve them i think.
> 
> Hrm, that sounds like it will be very manual. For example in `sk` we have
> 
> ```
> "plurals.notice_room_aliases_added.item_0"
> "plurals.notice_room_aliases_added.item_1"
> "plurals.notice_room_aliases_added.item_2"
> ```
> 
> whereas in `en` we only have
> 
> ```
> "plurals.notice_room_aliases_added.item_0"
> "plurals.notice_room_aliases_added.item_1"
> ```
> 
> So in code we're going to need check both the count and the language to decide which key we need to use (even if we rename them to `one`, `few`, `many` like they show in the Android keys) 😕

These items are handled in the new script.

pixlwave (approved):
This looks great, thanks for putting in the work to get plurals working properly 🙌

A couple of comments inline. I wonder if it would be worth adding a couple of tests for

- a language such as `sk` that have `one`/`few`/`many` plurals?
- a plural that has a couple of parameters such as `invitations_sent_to_one_and_more_users`?

Side note: There appear to be a couple of commits from Stefan and Manu that have creeped into this PR.

ElementX/Sources/Screens/LoginScreen/View/LoginScreen.swift
```diff
@@ -30,7 +30,7 @@ struct LoginScreen: View {
                 SecureField("Password", text: $context.password)
                     .textFieldStyle(.roundedBorder)
                 
-                Button("Login") {
+                Button(ElementL10n.roomSettingsMentionAndKeywordOnly) {
```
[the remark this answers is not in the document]
ismailgulek: Ah yes, will revert this

ismailgulek:
> * a language such as `sk` that have `one`/`few`/`many` plurals?

added 👍

> * a plural that has a couple of parameters such as `invitations_sent_to_one_and_more_users`?

This is not working 🤦‍♂️ I believe we need to add multiple variables for plurals for this case, for SwiftGen to generate multiple parameters. I'll handle that separately in the future, as i don't know how to do it exactly right now.

> Side note: There appear to be a couple of commits from Stefan and Manu that have creeped into this PR.

I've rebased from develop. Should be normal.

pixlwave:
> I'll handle that separately in the future

Sounds like a good plan to me!

🚢 it!

PR #57, opened 2022-05-13
Add DesignKit package

pixlwave:
Adds a Swift package containing the start of the DesignKit project. This currently contains dynamic colours in an asset catalog that are generated from an Amazon Style dictionary and fonts that are copied from Element iOS.

The PR currently includes a `ElementStyle` directory with the style dictionary project I've been using, however this part will be extracted out into a shared repo for all projects. At that stage DesignKit can become components only, with the fonts/colours living in the style dictionary repo.


DesignKit/Sources/Fonts/ElementFonts.swift
```diff
@@ -0,0 +1,89 @@
...
+
+public extension Font {
+    /// The fonts used by Element as defined in https://www.figma.com/file/X4XTH9iS2KGJ2wFKDqkyed/Compound?node-id=1362%3A0
+    static let element = ElementFonts(values: ElementSharedFonts())
+}
+
+/// Struct for holding fonts for use in SwiftUI.
+@available(iOS 14.0, *)
+public struct ElementFonts: Fonts {
+    public var largeTitle: Font
```
stefanceriu, line 27: These should all be `let` I believe.

DesignKit/Sources/Fonts/ElementUIFonts.swift
```diff
@@ -0,0 +1,88 @@
...
+import UIKit
+
+public extension UIFont {
+    /// The fonts used by Element as defined in https://www.figma.com/file/X4XTH9iS2KGJ2wFKDqkyed/Compound?node-id=1362%3A0
+    @objc static let element = ElementUIFonts(values: ElementSharedFonts())
+}
+
+/// ObjC class for holding fonts for use in UIKit.
+@objcMembers public class ElementUIFonts: NSObject, Fonts {
+    public var largeTitle: UIFont
```
stefanceriu, line 26: These should all be let I believe.

stefanceriu (approved):
Apart from previews not working in the package itself (which we talked about and decided to ignore for now) this looks amazing! I can't wait to start using it. 👏

Package.swift
```diff
@@ -0,0 +1,27 @@
+// swift-tools-version: 5.6
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "DesignKit",
+    platforms: [.iOS(.v14)],
```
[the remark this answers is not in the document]
pixlwave: Scratch that, it only sets the minimum OS version to be used on catalyst. Probably no point enforcing 14 for one and 15 for the other so will remove it.

Issue #53, opened 2022-05-13
Adopt Swift structured concurrency


stefanceriu:
We need to refactor our code to use Swift's new structured concurrency instead of the existing GCD based implementation. Ideally we should let the system to most of the work down to the FFI integration layer.

toshanmugaraj:
This is going to be major rewrite.

PR #63, opened 2022-05-26
#53 - Refactored various components

stefanceriu:
- updated the authentication coordinator to use async await
- replaced the UserSession with a ClientProxy (similar to the RoomProxy)
- renamed `completion` to `callback` on the coordinators and view models
- renamed `viewModelResult` to `viewModelAction`

### DIFF
--- a/ElementX/Sources/AppCoordinator.swift
+++ b/ElementX/Sources/AppCoordinator.swift
@@ -6,6 +6,7 @@
 //
 
 import UIKit
+import Kingfisher
 
 class AppCoordinator: AuthenticationCoordinatorDelegate, Coordinator {
     private let window: UIWindow
@@ -18,8 +19,10 @@ class AppCoordinator: AuthenticationCoordinatorDelegate, Coordinator {
     private let keychainController: KeychainControllerProtocol
     private let authenticationCoordinator: AuthenticationCoordinator!
     
+    private var userSession: UserSession!
+    
     private let memberDetailProviderManager: MemberDetailProviderManager
-        
+    
     private var indicatorPresenter: UserIndicatorTypePresenterProtocol
     private var loadingIndicator: UserIndicator?
     private var errorIndicator: UserIndicator?
@@ -70,15 +73,22 @@ class AppCoordinator: AuthenticationCoordinatorDelegate, Coordinator {
         showLoginErrorToast()
     }
     
-    func authenticationCoordinatorDidSetupUserSession(_ authenticationCoordinator: AuthenticationCoordinator) {
+    func authenticationCoordinatorDidSetupClientProxy(_ authenticationCoordinator: AuthenticationCoordinator) {
+        guard let clientProxy = authenticationCoordinator.clientProxy else {
+            fatalError("User session should be setup at this point")
+        }
+        
+        userSession = .init(clientProxy: clientProxy,
+                            mediaProvider: MediaProvider(clientProxy: clientProxy, imageCache: ImageCache.default))
+        
         presentHomeScreen()
     }
     
-    func authenticationCoordinatorDidTearDownUserSession(_ authenticationCoordinator: AuthenticationCoordinator) {
+    func authenticationCoordinatorDidTearDownClientProxy(_ authenticationCoordinator: AuthenticationCoordinator) {
         if let presentedCoordinator = childCoordinators.first {
             remove(childCoordinator: presentedCoordinator)
         }
-
+        
         mainNavigationController.setViewControllers([splashViewController], animated: false)
         authenticationCoordinator.start()
     }
@@ -89,18 +99,17 @@ class AppCoordinator: AuthenticationCoordinatorDelegate, Coordinator {
         
         hideLoadingIndicator()
         
-        guard let userSession = authenticationCoordinator.userSession else {
+        guard let userSession = userSession else {
             fatalError("User session should be already setup at this point")
         }
         
         let parameters = HomeScreenCoordinatorParameters(userSession: userSession,
-                                                         mediaProvider: userSession.mediaProvider,
                                                          attributedStringBuilder: AttributedStringBuilder(),
                                                          memberDetailProviderManager: memberDetailProviderManager)
         let coordinator = HomeScreenCoordinator(parameters: parameters)
         
-        coordinator.completion = { [weak self] result in
-            switch result {
+        coordinator.callback = { [weak self] action in
+            switch action {
             case .logout:
                 self?.authenticationCoordinator.logout()
             case .selectRoom(let roomIdentifier):
@@ -113,11 +122,11 @@ class AppCoordinator: AuthenticationCoordinatorDelegate, Coordinator {
     }
     
     private func presentRoomWithIdentifier(_ roomIdentifier: String) {
-        guard let userSession = authenticationCoordinator.userSession else {
+        guard let userSession = userSession else {
             fatalError("User session should be already setup at this point")
         }
         
-        guard let roomProxy = userSession.rooms.first(where: { $0.id == roomIdentifier }) else {
+        guard let roomProxy = userSession.clientProxy.rooms.first(where: { $0.id == roomIdentifier }) else {
             MXLog.error("Invalid room identifier: \(roomIdentifier)")
             return
         }
--- a/ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
+++ b/ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
@@ -17,7 +17,7 @@
 import Foundation
 import UIKit
 
-enum HomeScreenViewModelResult {
+enum HomeScreenViewModelAction {
     case logout
     case selectRoom(roomIdentifier: String)
 }
--- a/ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
+++ b/ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
@@ -176,11 +176,11 @@ struct HomeScreen_Previews: PreviewProvider {
                                     htmlBody: nil,
                                     date: .now)
         
-        let rooms = [MockRoomSummary(topic: "Topic", displayName: "Alpha"),
-                     MockRoomSummary(displayName: "Beta"),
-                     MockRoomSummary(displayName: "Omega", lastMessage: eventBrief)]
+        let roomSummaries = [MockRoomSummary(displayName: "Alpha", topic: "Topic"),
+                             MockRoomSummary(displayName: "Beta"),
+                             MockRoomSummary(displayName: "Omega", lastMessage: eventBrief)]
         
-        viewModel.updateWithRoomList(rooms)
+        viewModel.updateWithRoomSummaries(roomSummaries)
         
         if let avatarImage = UIImage(systemName: "person.fill.questionmark") {
             viewModel.updateWithUserAvatar(avatarImage)
--- a/ElementX/Sources/Services/Authentication/AuthenticationCoordinator.swift
+++ b/ElementX/Sources/Services/Authentication/AuthenticationCoordinator.swift
@@ -19,9 +19,9 @@ protocol AuthenticationCoordinatorDelegate: AnyObject {
     
     func authenticationCoordinatorDidStartLoading(_ authenticationCoordinator: AuthenticationCoordinator)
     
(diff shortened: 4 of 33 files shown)


ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
```diff
@@ -17,7 +17,7 @@
 import Foundation
 import UIKit
 
-enum HomeScreenViewModelResult {
+enum HomeScreenViewModelAction {
     case logout
     case selectRoom(roomIdentifier: String)
```
pixlwave, line 22: Seeing those 2 cases, this is definitely the correct name 👌

ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
```diff
@@ -176,11 +176,11 @@ struct HomeScreen_Previews: PreviewProvider {
                                     htmlBody: nil,
                                     date: .now)
         
-        let rooms = [MockRoomSummary(topic: "Topic", displayName: "Alpha"),
-                     MockRoomSummary(displayName: "Beta"),
-                     MockRoomSummary(displayName: "Omega", lastMessage: eventBrief)]
+        let roomSummaries = [MockRoomSummary(topic: "Topic", displayName: "Alpha"),
```
pixlwave, line 179: Dumb question unrelated to this PR, but would it make more sense to have the topic parameter after the displayname parameter?

pixlwave, line 179: Dumb question unrelated to this PR, but would it make more sense to have the topic parameter after the displayname parameter?

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -97,20 +83,26 @@ class UserSession {
         .value
     }
     
-    // MARK: ClientDelegate
+    func mediaSourceForURL(_ url: String) -> MatrixRustSDK.MediaSource {
+        MatrixRustSDK.mediaSourceFromUrl(url: url)
+    }
+    
+    func loadMediaContentForSource(_ source: MatrixRustSDK.MediaSource) throws -> [UInt8] {
```
pixlwave, line 90: Would it be sensible for this function to return `Data` for easier use throughout the app?

ElementX/Sources/Services/Client/ClientProxyProtocol.swift
```diff
@@ -0,0 +1,36 @@
...
+    
+    var rooms: [RoomProxy] { get }
+    
+    func loadUserDisplayName() async -> Result<String, ClientProxyError>
+        
+    func loadUserAvatarURL() async -> Result<String, ClientProxyError>
+    
+    func mediaSourceForURL(_ url: String) -> MatrixRustSDK.MediaSource
+    
+    func loadMediaContentForSource(_ source: MatrixRustSDK.MediaSource) throws -> [UInt8]
```
pixlwave, line 35: Would it be sensible for these to expose `URL`s and `Data` for the types to make them easier to use elsewhere in the app?

pixlwave, line 35: Would it be sensible for these to expose `URL`s and `Data` for the types to make them easier to use elsewhere in the app?

Tools/Scripts/Templates/SimpleScreenExample/ElementX/TemplateSimpleScreenCoordinator.swift
```diff
@@ -56,10 +56,10 @@ final class TemplateSimpleScreenCoordinator: Coordinator, Presentable {
     
     func start() {
         MXLog.debug("[TemplateSimpleScreenCoordinator] did start.")
-        templateSimpleScreenViewModel.completion = { [weak self] result in
+        templateSimpleScreenViewModel.callback = { [weak self] result in
```
pixlwave, line 59:
```suggestion
        templateSimpleScreenViewModel.callback = { [weak self] action in
```

ElementX/Sources/Services/Authentication/AuthenticationCoordinator.swift
```diff
@@ -95,24 +93,23 @@ class AuthenticationCoordinator: Coordinator {
...
+                    switch await self.login(username: result.username, password: result.password) {
                     case .success:
                         completion(.success(()))
                         self.remove(childCoordinator: coordinator)
                         self.navigationRouter.dismissModule()
                     case .failure(let error):
                         completion(.failure(error))
                     }
+                    
                 }
```
pixlwave, line 113: Might be cleaner to extract this into a function and call that from the callback.

pixlwave, line 113: Might be cleaner to extract this into a function and call that from the callback.

ElementX/Sources/Services/Authentication/AuthenticationCoordinator.swift
```diff
@@ -123,58 +120,48 @@ class AuthenticationCoordinator: Coordinator {
...
-                }
-            } catch {
-                MXLog.error("Failed logging in with error: \(error)")
-                
-                DispatchQueue.main.async {
-                    completion(.failure(.failedLoggingIn))
-                }
-            }
+        let basePath = baseDirectoryPathForUsername(username)
+        async let loginTask = Task.detached {
```
pixlwave, line 129: I think this should probably be a plain `let` given that you await the result straight after.

ElementX/Sources/Services/Authentication/AuthenticationCoordinator.swift
```diff
@@ -123,58 +120,48 @@ class AuthenticationCoordinator: Coordinator {
...
+        case .success(let client):
+            return await setupClientProxyForClient(client)
+        case .failure(let error):
+            MXLog.error("Failed logging in with error: \(error)")
+            return .failure(.failedLoggingIn)
         }
     }
     
-    private func restorePreviousLogin(_ usernameTokenTuple: (username: String, token: String), completion: @escaping (Result<Void, AuthenticationCoordinatorError>) -> Void) {
+    private func restorePreviousLogin(_ usernameTokenTuple: (username: String, token: String)) async -> Result<Void, AuthenticationCoordinatorError> {
```
pixlwave, line 144: Nit-pick but could this be `accessToken` just to be clear that its different from a `token` based login.
```suggestion
    private func restorePreviousLogin(_ usernameTokenTuple: (username: String, accessToken: String)) async -> Result<Void, AuthenticationCoordinatorError> {
```
```suggestion
    private func restorePreviousLogin(_ usernameTokenTuple: (username: String, token: String)) async -> Result<Void, AuthenticationCoordinatorError> {
```

pixlwave, line 144: Nit-pick but could this be `accessToken` just to be clear that its different from a `token` based login.
```suggestion
    private func restorePreviousLogin(_ usernameTokenTuple: (username: String, accessToken: String)) async -> Result<Void, AuthenticationCoordinatorError> {
```

ElementX/Sources/Services/Authentication/AuthenticationCoordinator.swift
```diff
@@ -123,58 +120,48 @@ class AuthenticationCoordinator: Coordinator {
...
+        case .success(let client):
+            return await setupClientProxyForClient(client)
+        case .failure(let error):
+            MXLog.error("Failed restoring login with error: \(error)")
+            return .failure(.failedRestoringLogin)
         }
     }
     
-    private func setupUserSessionForClient(_ client: Client) {
+    private func setupClientProxyForClient(_ client: Client) async -> Result<Void, AuthenticationCoordinatorError> {
```
pixlwave, line 164:
```suggestion
    private func setupProxyForClient(_ client: Client) async -> Result<Void, AuthenticationCoordinatorError> {
```

pixlwave, line 164:
```suggestion
    private func setupProxyForClient(_ client: Client) async -> Result<Void, AuthenticationCoordinatorError> {
```

ElementX/Sources/Services/Authentication/AuthenticationCoordinator.swift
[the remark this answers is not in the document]
pixlwave: Same comment on restoring below too.

pixlwave (commented):
GitHub started double posting my comments, so 🤞 I've removed the duplicates.

ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
stefanceriu: Yeah, that does indeed look weird.

ElementX/Sources/Services/Client/ClientProxyProtocol.swift
stefanceriu: Yep yep, good idea 👍

ElementX/Sources/Services/Authentication/AuthenticationCoordinator.swift
stefanceriu: I would need to pass the coordinator around which is now referenced locally, and that makes things annoying. I think it's reasonable enough to leave it like this.

pixlwave: Gotcha 👍

ElementX/Sources/AppCoordinator.swift
```diff
@@ -89,18 +99,17 @@ class AppCoordinator: AuthenticationCoordinatorDelegate, Coordinator {
...
         
         let parameters = HomeScreenCoordinatorParameters(userSession: userSession,
-                                                         mediaProvider: userSession.mediaProvider,
                                                          attributedStringBuilder: AttributedStringBuilder(),
                                                          memberDetailProviderManager: memberDetailProviderManager)
         let coordinator = HomeScreenCoordinator(parameters: parameters)
         
-        coordinator.completion = { [weak self] result in
-            switch result {
+        coordinator.callback = { [weak self] action in
```
pixlwave, line 111: I think this one on the coordinator is still a result fwiw.

pixlwave (approved):
It tells me there are 4 comments when I've only added 2, so lets see what it adds 🙈

Otherwise LGTM

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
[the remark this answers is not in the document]
pixlwave: Yep lets add this a THIRD time 😂

ElementX/Sources/AppCoordinator.swift
```diff
@@ -18,6 +19,9 @@ class AppCoordinator: AuthenticationCoordinatorDelegate, Coordinator {
     private let keychainController: KeychainControllerProtocol
     private let authenticationCoordinator: AuthenticationCoordinator!
     
+    private var clientProxy: ClientProxyProtocol!
+    private var mediaProvider: MediaProviderProtocol!
```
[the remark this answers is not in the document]
pixlwave: Urgh, that was from the first review, but you get the idea. Not necessary right now but would be nice.

ElementX/Sources/Services/Client/ClientProxyProtocol.swift
pixlwave: They still show as `String` for me rather than `URL` but not exactly urgent.

ElementX/Sources/AppCoordinator.swift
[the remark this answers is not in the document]
stefanceriu: Yep yep, I'll try to get that working with the new userSession once I adopt the state machine on the main branch.

ElementX/Sources/AppCoordinator.swift
stefanceriu: Ah, I see. I actually think we have a problem with the templates themselves. They reuse the `ViewModel Result/Action` instead of a separate type.

ElementX/Sources/Services/Client/ClientProxyProtocol.swift
stefanceriu: Oh, my bad, I only read the `Data` part. I thought about using URLs before but the annoying part about that is that they have optional constructors which would need to be handled at the calling place everywhere. Also, none of the stack actually uses URLs, down to the rust side.

PR #64, opened 2022-05-27
Add the DesignKit package to the project.

pixlwave:
Following on from #57, this PR adds the package to the project and uses it in the login screen. I've added it locally using a path and Xcodegen puts it in a `Packages` group which is nice. Also adds:

- An `ElementTextFieldStyle` that uses the `BorderedInputFieldStyle` and adds a header, footer and focus state.
- Static access for all of the SwiftUI styles.

I've applied the styles to the login screen as part of this PR:

Simulator Screen Shot - iPod touch (7th generation) - 2022-06-08 at 18 59 00

### DIFF
--- a/DesignKit/Sources/Buttons/ElementActionButtonStyle.swift
+++ b/DesignKit/Sources/Buttons/ElementActionButtonStyle.swift
@@ -17,38 +17,44 @@
 import SwiftUI
 import DesignTokens
 
-public extension ButtonStyle where Self == PrimaryActionButtonStyle {
-    static func primaryAction(customColor: Color? = nil) -> PrimaryActionButtonStyle {
-        PrimaryActionButtonStyle(customColor: customColor)
+public extension ButtonStyle where Self == ElementActionButtonStyle {
+    /// The CTA button style as defined in Compound.
+    /// - Parameter size: The control size to use. Defaults to regular.
+    /// - Parameter customColor: A custom color for the button's background. Defaults to the accent color.
+    static func elementAction(_ size: ElementControlSize = .regular,
+                              color: Color = .element.accent) -> ElementActionButtonStyle {
+        ElementActionButtonStyle(size: size, color: color)
     }
 }
 
-public struct PrimaryActionButtonStyle: ButtonStyle {
+public struct ElementActionButtonStyle: ButtonStyle {
     @Environment(\.isEnabled) private var isEnabled
     @Environment(\.colorScheme) private var colorScheme
     
-    public var customColor: Color?
+    public var size: ElementControlSize
+    public var color: Color
+    
+    private var verticalPadding: CGFloat { size == .xLarge ? 12 : 4 }
+    private var maxWidth: CGFloat? { size == .xLarge ? .infinity : nil }
     
     private var fontColor: Color {
         // Always white unless disabled with a dark theme.
         .white.opacity(colorScheme == .dark && !isEnabled ? 0.3 : 1.0)
     }
     
-    private var backgroundColor: Color {
-        customColor ?? .element.accent
-    }
-    
-    public init(customColor: Color? = nil) {
-        self.customColor = customColor
+    public init(size: ElementControlSize = .xLarge, color: Color = .element.accent) {
+        self.size = size
+        self.color = color
     }
     
     public func makeBody(configuration: Self.Configuration) -> some View {
         configuration.label
-            .padding(12.0)
-            .frame(maxWidth: .infinity)
+            .padding(.horizontal, 12)
+            .padding(.vertical, verticalPadding)
+            .frame(maxWidth: maxWidth)
             .foregroundColor(fontColor)
             .font(.element.body)
-            .background(backgroundColor.opacity(backgroundOpacity(when: configuration.isPressed)))
+            .background(color.opacity(backgroundOpacity(when: configuration.isPressed)))
             .cornerRadius(8.0)
     }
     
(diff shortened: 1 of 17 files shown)


DesignKit/Sources/TextFields/BorderedInputFieldStyle.swift
```diff
@@ -80,13 +72,16 @@ public struct BorderedInputFieldStyle: TextFieldStyle {
             .font(.element.callout)
             .foregroundColor(textColor)
             .accentColor(accentColor)
-            .frame(height: 48.0)
+            .padding(.vertical, 12.0)
```
pixlwave, line 75: I need to double check this is accurate, but it looks nicer when enabling dynamic type.

DesignKit/Sources/TextFields/BorderedInputFieldStyle.swift
```diff
@@ -80,13 +72,16 @@ public struct BorderedInputFieldStyle: TextFieldStyle {
...
             .padding(.horizontal, 8.0)
             .background(backgroundColor)
             .clipShape(rect)
             .overlay(rect.stroke(borderColor, lineWidth: borderWidth))
             .introspectTextField { textField in
-                textField.returnKeyType = .done
+                if #unavailable(iOS 15.0) {
+                    textField.returnKeyType = .done
+                }
```
pixlwave, line 83: We can use `.submitLabel` on iOS 15 now.

**Note:** Thinking about this, it will affect text fields in EI when using the package 😕

ElementX/Sources/Screens/LoginScreen/View/LoginScreen.swift
```diff
@@ -15,35 +15,46 @@
...
+            
+            SecureField("Password", text: $context.password)
+                .textFieldStyle(.elementInput())
+                .textContentType(.password)
+                .focused($focussedField, equals: .password)
+                .submitLabel(.go)
+                .onSubmit(submit)
+            
+            Button("Login", action: submit)
+                .buttonStyle(.primaryAction())
```
pixlwave, line 46: Looking at the Figma, I think this is an "Extra Large CTA" so this should be `.elementActionXL`. And the secondary action seems to be `.elementGhostXL`.

DesignKit/Sources/Buttons/SecondaryActionButtonStyle.swift
```diff
@@ -30,20 +36,17 @@ public struct SecondaryActionButtonStyle: ButtonStyle {
...
     
-    private func strokeColor(_ isPressed: Bool) -> Color {
-        if let customColor = customColor {
-            return customColor
-        }
-        
-        return isPressed ? .element.accent.opacity(0.6) : .element.accent
+    private func opacity(when isPressed: Bool) -> CGFloat {
+        guard isEnabled else { return 0.6 }
+        return isPressed ? 0.6 : 1.0
```
pixlwave, line 49: https://github.com/vector-im/element-ios/pull/6204#discussion_r884537843

ElementX.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -59,8 +68,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/getsentry/sentry-cocoa.git",
       "state" : {
-        "revision" : "afa4cd596e5cf97a797fb9b5c3afeea4add2c7a3",
-        "version" : "7.15.0"
+        "revision" : "b34ce9048e6b8b082ea6b15e6b50583d46c936ea",
+        "version" : "7.16.0"
```
pixlwave, line 72: Seeing this, I think Xcode might be trying to be "helpful" whenever it resolves packages: makes me wonder if the DesignTokens repo needs tagging (or specifying by commit rather than branch) otherwise navigating back in history could be hard.

ElementX/Sources/Screens/LoginScreen/View/LoginScreen.swift
```diff
@@ -15,35 +15,46 @@
...
+            
+            SecureField("Password", text: $context.password)
+                .textFieldStyle(.elementInput())
+                .textContentType(.password)
+                .focused($focussedField, equals: .password)
+                .submitLabel(.go)
+                .onSubmit(submit)
+            
+            Button("Login", action: submit)
+                .buttonStyle(.elementAction(.xLarge))
```
pixlwave, line 46: The other option here would be `.elementAction.xLarge`, I could give that a go if that sounds any better.

However with a custom colour it would be `.elementAction(color: .pink).xLarge` which probably doesn't read quite as well as `.elementAction(.xLarge, color: .pink)`.

ismailgulek (changes requested):
Looks great. Just a couple of comments.

DesignKit/Sources/Buttons/ElementActionButtonStyle.swift
```diff
@@ -17,32 +17,44 @@
...
         .white.opacity(colorScheme == .dark && !isEnabled ? 0.3 : 1.0)
     }
     
-    private var backgroundColor: Color {
-        customColor ?? .element.accent
-    }
-    
-    public init(customColor: Color? = nil) {
-        self.customColor = customColor
+    public init(size: ElementControlSize = .xLarge, color: Color = .element.accent) {
```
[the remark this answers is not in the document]
pixlwave: Ah yeah good point. I was trying to keep this source compatible with EI but then I decided it was better to rename it so that makes way more sense 👍

ElementX/Sources/Screens/LoginScreen/View/LoginScreen.swift
```diff
@@ -15,35 +15,48 @@
...
-            .onSubmit {
-                context.send(viewAction: .login)
-            }
+        VStack {
+            TextField("Username", text: $context.username)
+                .textFieldStyle(.elementInput())
+                .disableAutocorrection(true)
+                .textContentType(.username)
+                .autocapitalization(.none)
+                .focused($focussedField, equals: .username)
```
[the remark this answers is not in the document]
pixlwave: TBH, its probable a bit overkill to have the `.focused` modifier on it, but what happens is when the text field becomes focused, it updates `focussedField` to `.username` automatically. It means we could query which component has focus and change logic based on that at some point.

DesignKit/Sources/TextFields/ElementTextFieldStyle.swift
```diff
@@ -0,0 +1,71 @@
...
+                .focused($isFocused)
+            
+            if let footerText = footerText {
+                Text(footerText)
+                    .font(.element.footnote)
+                    .foregroundColor(footerColor)
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
pixlwave: Good shout, also made me realise that `isError` should be the last parameter and that the stack wasn't actually aligned properly. Annoyingly previews from a package currently crash Xcode, but its a known bug. I might move the previews into the EX project in a future PR 🤔

DesignKit/Sources/Buttons/SecondaryActionButtonStyle.swift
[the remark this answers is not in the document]
pixlwave: I'm going to leave this for now. Both of the `0.6` values would have to be defined, and it is pretty clear from the code that the first is the disabled opacity and the second is the pressed opacity. Hopefully values like these will come from the DesignTokens package at some point anyway, and then they can be shared between all button styles etc.

ElementX/Sources/Screens/LoginScreen/View/LoginScreen.swift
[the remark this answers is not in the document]
ismailgulek: Ah i didn't get it from the code 🤦‍♂️, ignore me.

Issue #66, opened 2022-05-31
Introduce State Machine component


stefanceriu:
In order to better track app states we would like to introduce a FSM component that will hard assert that states transitions are correct.
We will use [SwiftState](https://github.com/ReactKit/SwiftState) and for starters we will rebuild the AppCoordinator on top of it.

PR #68, opened 2022-05-31
#66 - Rebuild the AppCoordinator on top of a SwiftState FSM

stefanceriu:
This PR introduces a state machine to the AppCoordinator that enforces correct state transitions.

### DIFF
--- a/ElementX/Sources/AppCoordinator.swift
+++ b/ElementX/Sources/AppCoordinator.swift
@@ -11,6 +11,8 @@ import Kingfisher
 class AppCoordinator: AuthenticationCoordinatorDelegate, Coordinator {
     private let window: UIWindow
     
+    private var stateMachine: AppCoordinatorStateMachine
+    
     private let mainNavigationController: UINavigationController
     private let splashViewController: UIViewController
     
@@ -30,6 +32,8 @@ class AppCoordinator: AuthenticationCoordinatorDelegate, Coordinator {
     var childCoordinators: [Coordinator] = []
     
     init() {
+        stateMachine = AppCoordinatorStateMachine()
+        
         splashViewController = SplashViewController()
         mainNavigationController = UINavigationController(rootViewController: splashViewController)
         window = UIWindow(frame: UIScreen.main.bounds)
@@ -50,6 +54,8 @@ class AppCoordinator: AuthenticationCoordinatorDelegate, Coordinator {
                                                               navigationRouter: navigationRouter)
         authenticationCoordinator.delegate = self
         
+        setupStateMachine()
+        
         let loggerConfiguration = MXLogConfiguration()
         loggerConfiguration.logLevel = .verbose
         MXLog.configure(loggerConfiguration)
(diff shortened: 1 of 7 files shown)


ElementX/Sources/AppCoordinator.swift
```diff
@@ -146,14 +191,21 @@ class AppCoordinator: AuthenticationCoordinatorDelegate, Coordinator {
...
+        add(childCoordinator: coordinator)
+        navigationRouter.push(coordinator) { [weak self] in
             guard let self = self else { return }
-            
-            self.remove(childCoordinator: coordinator)
+            self.stateMachine.processEvent(.popRoomScreen)
         }
     }
     
+    private func dismissRoomScreen() {
```
pixlwave, line 201: This naming of this method sounds like you could call it to dismiss the room screen, whereas its use is really to tidy up after the room screen was dismissed.

ElementX/Sources/AppCoordinatorStateMachine.swift
```diff
@@ -0,0 +1,72 @@
...
+//
+
+import Foundation
+import SwiftState
+
+class AppCoordinatorStateMachine {
+    enum State: StateType {
+        case initial
+        case signedOut, signingOut
+        case signedIn, signingIn
```
pixlwave, line 16: It seems that `signingIn` is used for both "I've entered my username and password and am looking at a spinner" as well as "I've opened the app again and am looking at a spinner". I wonder if we this should either be something like `loadingSession`/`restoringSession` or if we should actually have both?

Side note: It would be nice to start adding doc comments to all these cases for clarity :)

pixlwave (commented):
This looks like such a great way to handle app state to me 😍

Do you think it would be a good time to make the `userSession` an associated value that gets passed between the events/states rather than being an implicitly unwrapped property.

Looking at it, I would also imagine (although not necessarily in this PR) this state machine shouldn't go any further than showing either the auth flow or the split view. And then the split view would have its own state machine to handle room navigation as this will become way more complex for Mac/iPad.

ElementX/Sources/AppCoordinator.swift
stefanceriu: Good point. I went with `dismissedRoomScreen` and `tearDownDismissedRoomScreen` for now, we can always change it later if you have better ideas.

ElementX/Sources/AppCoordinatorStateMachine.swift
stefanceriu: I didn't see the need to separate them for now as, intention wise, they're pretty much the same: setup a valid session.
We will of course probably need to change this over time but for now the main goal is to introduce some sort of state machine.

stefanceriu:
Moving the `userSession` to state would require it to conform to Hashable and we would still need to store it locally so that it can be used from other states. I figured it's not really worth it as long as we assert it properly when switching states. I see associated values more useful for passing transient data between states.

Re your second point: absolutely. The AppCoordinator fsm will definitely change as we introduce more components (and more state machines). Ideally it wouldn't do much itself except coordinate other top level components. 
We could have an OnboardingFlowCoordinator, an AuthFlowCoordinator (Create Account + Login + Restore Session + Verify Session), an UserSessionFlowCoordinator (Home, Settings, Room etc) etc. Each would take care of different parts of the application and the AppCoordinator's responsiblity would be to coordinate them. (sounds a bit obvious when written down 😄)

PR #69, opened 2022-05-31
Smooth keyboard presentation animation

stefanceriu:
The main purpose of this PR is to fix janky animations while presenting the keyboard. 
It also cleans up some other animations and fixed the copywright file headers and project template.


ismailgulek (approved):
Didn't test it out but LGTM!

PR #70, opened 2022-05-31
Rageshake Service

### LINKED ISSUE #23: Rageshake Service
We need to have a fresh new and robust service to send rageshakes:

- [ ] Rage shake detection
- [ ] Upload EIX and SDK logs
- [ ] Last crash
- [ ] Text description
- [ ] Screenshot
- [ ] [Anonymised Screenshot](https://github.com/vector-im/element-web/issues/9615#issuecomment-1108992901) would be nice

The UI will be managed in https://github.com/vector-im/element-x-ios/issues/24.

### LINKED ISSUE #24: Rageshake UI
We need a screen and UX to report bug via rage shake:

- [ ] Text description
- [ ] Optional logs
- [ ] Optional screenshot
- [ ] [Anonymised screenshot](https://github.com/vector-im/element-web/issues/9615#issuecomment-1108992901) would be nice

### LINKED ISSUE #61: Integrate with Sentry.io

ismailgulek:
Fixes #23
Fixes #24 
Fixes #61


ElementX/Sources/Screens/BugReport/View/BugReport.swift
```diff
@@ -0,0 +1,160 @@
...
+                .accessibilityLabel("Disable Sending Logs")
+        } else {
+            return Image(uiImage: Asset.Images.selectionUntick.image)
+                .accessibilityLabel("Enable Sending Logs")
+        }
+    }
+
+    var screenshot: some View {
+        if let screenshot = context.screenshot {
+            return AnyView(
```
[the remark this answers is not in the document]
ismailgulek: Great, thanks.

ElementX/Sources/Services/BugReport/BugReportService.swift
```diff
@@ -0,0 +1,207 @@
...
+import Sentry
+
+class BugReportService: BugReportServiceProtocol {
+    private let baseURL: URL
+    private let sentryEndpoint: String
+    private let session: URLSession
+
+    init(withBaseURL baseURL: URL,
+         sentryEndpoint: String,
+         applicationId: String = "riot-ios",
```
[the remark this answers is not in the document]
ismailgulek: It should match with server configuration.

ElementX/Sources/Screens/Settings/View/Settings.swift
```diff
@@ -0,0 +1,75 @@
...
+                buttons
+                    .padding(.horizontal, horizontalPadding)
+            }
+        }
+        .navigationTitle(ElementL10n.settings)
+    }
+
+    /// The action buttons shown at the bottom of the view.
+    var buttons: some View {
+        VStack {
```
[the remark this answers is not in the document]
ismailgulek: thanks for the suggestion, looks much better :)

UnitTests/Sources/ScreenshotDetectorTests.swift
```diff
@@ -0,0 +1,36 @@
...
+import Foundation
+@testable import ElementX
+import XCTest
+
+class ScreenshotDetectorTests: XCTestCase {
+
+    func testDetection() {
+        async { expectation in
+            let detector = ScreenshotDetector()
+            detector.callback = { _, _ in
```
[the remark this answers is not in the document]
ismailgulek: I've tried different versions of this before, need to play more with these actors :/

ElementX/Resources/Assets.xcassets/Images/sample_screenshot.imageset/Contents.json
```diff
@@ -0,0 +1,12 @@
+{
```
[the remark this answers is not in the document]
ismailgulek: Investigated this a bit, didn't seem very possible. I've fixed this by adding sample screenshot to test target directly and removing from the app target.

ElementX/Sources/Services/BugReport/ScreenshotDetector.swift
```diff
@@ -0,0 +1,124 @@
...
+import Foundation
+import UIKit
+import Photos
+
+enum ScreenshotDetectorError: Error {
+    case loadFailed
+    case notAuthorized
+}
+
+class ScreenshotDetector {
```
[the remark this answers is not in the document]
ismailgulek: Yes, i'm aware of that. I'm planning to improve this later, by disabling `autoRequestPHAuthorization` and then implementing file attachment functionality in the bug report screen.

If it's too disturbing i can disable `autoRequestPHAuthorization` now and make this functionality best-effort.

ElementX/Sources/AppCoordinatorStateMachine.swift
```diff
@@ -79,6 +87,8 @@ class AppCoordinatorStateMachine {
                     return .roomScreen(roomId: roomId)
                 case (.dismissedRoomScreen, .roomScreen):
                     return .homeScreen
+                case (.dismissedSettingsScreen, .settingsScreen):
+                    return .homeScreen
```
stefanceriu, line 91: This transition shouldn't be necessary as you're already handling the same one above in `machine.addRoutes(event: .dismissedSettingsScreen, transitions: [ .settingsScreen => .homeScreen ])`

ElementX/Sources/BuildSettings.swift
```diff
@@ -0,0 +1,20 @@
...
+//  BuildSettings.swift
+//  ElementX
+//
+//  Created by Ismail on 2.06.2022.
+//  Copyright © 2022 Element. All rights reserved.
+//
+
+import Foundation
+
+final class BuildSettings {
```
stefanceriu, line 11: 👍

ElementX/Sources/Services/BugReport/BugReportService.swift
```diff
@@ -0,0 +1,213 @@
...
+
+    var applicationWasCrashed: Bool {
+        return SentrySDK.crashedLastRun
+    }
+
+    func crash() {
+        SentrySDK.crash()
+    }
+
+    var applicationId: String
```
stefanceriu, line 66: Shall we move this to the top and make it private (i.e. remove it from the protocol)?

ElementX/Sources/Services/BugReport/BugReportService.swift
```diff
@@ -0,0 +1,213 @@
...
+
+        //  also enable logging crashes, to send them with bug reports
+        MXLogger.logCrashes(true)
+        //  set build version for logger
+        MXLogger.setBuildVersion(ElementInfoPlist.cfBundleShortVersionString)
+    }
+
+    // MARK: - BugReportServiceProtocol
+
+    var applicationWasCrashed: Bool {
```
stefanceriu, line 58: Can I suggest maybe replacing this with `hasApplicationPreviouslyCrashed` or even sentry's own `crashedLastRun`?

ElementX/Sources/AppCoordinator.swift
```diff
@@ -34,7 +37,13 @@ class AppCoordinator: AuthenticationCoordinatorDelegate, Coordinator {
     
     init() {
         stateMachine = AppCoordinatorStateMachine()
-        
+
+        guard let baseURL = URL(string: BuildSettings.bugReportServiceBaseUrlString) else {
+            fatalError("")
```
stefanceriu, line 42: You could remove the quotes here entirely or maybe add a comment à la "Invalid bug report service base URL". Another option would be to move this into a throwing constructor and check the `bugReportSentryEndpoint`for empty too.

ElementX/Sources/AppCoordinator.swift
```diff
@@ -53,12 +62,20 @@ class AppCoordinator: AuthenticationCoordinatorDelegate, Coordinator {
         keychainController = KeychainController(identifier: bundleIdentifier)
         authenticationCoordinator = AuthenticationCoordinator(keychainController: keychainController,
                                                               navigationRouter: navigationRouter)
+
+        screenshotDetector = ScreenshotDetector()
+        screenshotDetector.callback = askAfterScreenshot
```
stefanceriu, line 67: If I might suggest `processScreenshotDetection` instead of `askAfterScreenshot`.

ElementX/Sources/Services/BugReport/ScreenshotDetector.swift
```diff
@@ -0,0 +1,106 @@
...
+
+    init() {
+        startObservingScreenshots()
+    }
+
+    private func startObservingScreenshots() {
+        NotificationCenter.default.addObserver(self, selector: #selector(userDidTakeScreenshot), name: UIApplication.userDidTakeScreenshotNotification, object: nil)
+    }
+
+    @objc @MainActor private func userDidTakeScreenshot() {
```
stefanceriu, line 33: I think we can make the whole class a @MainActor 🤔

ElementX/Sources/Services/BugReport/ScreenshotDetector.swift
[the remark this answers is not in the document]
stefanceriu: I would say let's leave it in for now. It's easy enough to remove or change later.

Otherwsie, Isn't there also a flow in which the user takes the screenshot but doesn't end up saving it to the photo library? I take a lot of screenshots that I just copy to the pasteboard and delete afterwards.
I think the only good option is to let the user choose when/if they want to attach a screenshot.

ElementX/Sources/Services/BugReport/BugReportService.swift
```diff
@@ -0,0 +1,213 @@
...
+            options.debug = true
+            #endif
+
+            // Set tracesSampleRate to 1.0 to capture 100% of transactions for performance monitoring.
+            // We recommend adjusting this value in production.
+            options.tracesSampleRate = 1.0
+
+            options.beforeSend = { event in
+                MXLog.error("Sentry detected crash: \(event)")
+                return nil
```
stefanceriu, line 42: Don't we need to return back the `event` from here instead of `nil`?

ElementX/Sources/Generated/InfoPlist.swift
```diff
@@ -0,0 +1,67 @@
...
+
+import Foundation
+
+// swiftlint:disable superfluous_disable_command
+// swiftlint:disable file_length
+
+// MARK: - Plist Files
+
+// swiftlint:disable identifier_name line_length type_body_length
+internal enum ElementInfoPlist {
```
stefanceriu, line 12: Nice! 👍

ElementX/Sources/Other/UIImage+Element.swift
```diff
@@ -0,0 +1,106 @@
...
+        "&",
+        "+",
+        "-",
+        "_",
+        "\"",
+        "?",
+        "*"
+    ]
+
+    func anonymized(confidenceLevel: Float = 0.5, fillColor: UIColor = .red) async throws -> UIImage {
```
stefanceriu, line 32: 😍 
I wouldn't have made it a public extension though as it feels very specific to the bug report flows. Maybe a static method on a ImageAnonymizer struct is enough.

ElementX/Sources/Screens/BugReport/View/BugReport.swift
```diff
@@ -0,0 +1,148 @@
...
+        VStack {
+            Button { context.send(viewAction: .submit) } label: {
+                Text(ElementL10n.actionSend)
+            }
+            .frame(maxWidth: .infinity)
+            .padding(.vertical, 16)
+            .disabled(context.reportText.count < 5)
+            .accessibilityIdentifier("sendButton")
+
+            Button { context.send(viewAction: .cancel) } label: {
```
stefanceriu, line 108: These buttons when stacked together cover quite a bit part of the screen. I would remove the `cancel` one as you can always navigate backwards and implicitly cancel.

ElementX/Sources/Screens/HomeScreen/HomeScreenCoordinator.swift
```diff
@@ -66,6 +67,8 @@ final class HomeScreenCoordinator: Coordinator, Presentable {
                 self.callback?(.logout)
             case .selectRoom(let roomIdentifier):
                 self.callback?(.selectRoom(roomIdentifier: roomIdentifier))
+            case .tapUserAvatar:
+                self.callback?(.settings)
```
stefanceriu, line 71: `Settings` is a bit generic, maybe `presentSettings` instead. Could do the same for rooms `presentRoom(roomIdentifier: String)? 🤔

UnitTests/Sources/ScreenshotDetectorTests.swift
```diff
@@ -0,0 +1,37 @@
...
+
+                //  just check whether the callback is called
+                XCTAssert(true)
+
+                expectation.fulfill()
+            }
+
+            NotificationCenter.default.post(name: UIApplication.userDidTakeScreenshotNotification, object: nil)
```
stefanceriu, line 26: I think this will invoke the system permissions dialog on the CI which might break the UI tests.

UnitTests/Sources/ImageExtensionTests.swift
```diff
@@ -0,0 +1,53 @@
...
+//  UnitTests
+//
+//  Created by Ismail on 31.05.2022.
+//  Copyright © 2022 element.io. All rights reserved.
+//
+
+import XCTest
+@testable import ElementX
+
+class ImageExtensionTests: XCTestCase {
```
stefanceriu, line 12: You'll need to add `- path: ../Resources` to the UnitTests xcodegen target.yml but the tests would still fail. There's an exception being thrown from the anonymized VNImageRequestHandler -> `Error Domain=com.apple.vis Code=9 "encountered an unexpected condition: Unspecified error" UserInfo={NSLocalizedDescription=encountered an unexpected condition: Unspecified error}`

stefanceriu (changes requested):
Great stuff 👏  Just a few comments inline.

stefanceriu:
> Should we be adding `MXLogger` in its current form as an objc class

I also agree that we should try to get rid of it but at the moment, apart from crash reporting, it's also handling the logging to files and log rotations. Let's tackle this separately when we investigate Logger and rust side logs.

ElementX/Sources/AppCoordinatorStateMachine.swift
ismailgulek: Hmm, forgot this one yeah. First i thought this was needed.

ElementX/Sources/AppCoordinator.swift
ismailgulek: Yeah forgot to fill the fatal error, but a throwing constructor made more sense :)

UnitTests/Sources/ScreenshotDetectorTests.swift
ismailgulek: Right, i've updated `autoRequestPHAuthorization` for tests and checked the current authorization status on the device.

UnitTests/Sources/ImageExtensionTests.swift
ismailgulek: I think updating the yaml file fixed it!

ElementX/Sources/Services/BugReport/ScreenshotDetector.swift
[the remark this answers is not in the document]
ismailgulek: I think it saves the screenshot first and remove if you want afterwards. As we show the popup as soon as the screenshot taken, it seems fine to me. I agree the real solution would be adding functionality to add files into the bug report screen, which i'm planning to provide later.

ElementX/Sources/Screens/BugReport/View/BugReport.swift
ismailgulek: Good idea, i've added a cancel bar button item when we present this screen.

stefanceriu (approved):
👏 👏

ElementX/Sources/Services/BugReport/ScreenshotDetector.swift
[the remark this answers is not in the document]
toshanmugaraj: I would prefer the screenshot option is disabled by default. 
In many cases, it shares the sensitive information to the Rageshake service, and not much helpful. 
I see in one of the demo in Matrix web client, they blur the all dynamic contents while taking the screenshot for Rageshake.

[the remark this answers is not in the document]
pixlwave: > If it's too disturbing i can disable autoRequestPHAuthorization now

No no, like Stefan said, its fine for now :)

ElementX/Resources/Assets.xcassets/Images/sample_screenshot.imageset/Contents.json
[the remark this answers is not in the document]
pixlwave: That sounds even better TBH 👍

pixlwave:
> I also agree that we should try to get rid of it but at the moment, apart from crash reporting, it's also handling the logging to files and log rotations. Let's tackle this separately when we investigate Logger and rust side logs.

Ah yeah that makes more sense :)

PR #73, opened 2022-06-08
Add filtering of rooms by name.

pixlwave:
Part of #26, adds a search bar to the home screen that filters room by name.

Note: We should probably make the user avatar smaller as it imposes on the search bar a little bit on iPhone.


stefanceriu (approved):
Looks good to me but please squash it on merge 👍

PR #75, opened 2022-06-08
Fix Danger's changelog detection.

### LINKED ISSUE #74: Danger doesn't detect new changelog files.
If you create a PR with a new changelog file, Danger comments that you should add one. It looks like it is only checking modified files rather than new files, but switching [this line](https://github.com/vector-im/element-x-ios/blob/54f5493e763be1a7ecccdf545d32d35f7f6944d1/Dangerfile.swift#L20) to use `editedFiles` didn't seem to work when I tested it on a PR.

pixlwave:
🤞 fixes #74


stefanceriu (approved):
Nice! 👏

PR #79, opened 2022-06-10
Use .element.accent instead of elementGreen colour asset.

pixlwave:
I've also removed the local copy of Element Green and replaced the `ASSETCATALOG_COMPILER_GLOBAL_ACCENT_COLOR_NAME` project setting with a `window.tintColor = .element.accent` at app launch. This seems like the right thing to do, but it probably won't take effect if we make another window at some stage so open to other ideas.


ElementX/Sources/AppCoordinator.swift
```diff
@@ -49,6 +49,7 @@ class AppCoordinator: AuthenticationCoordinatorDelegate, Coordinator {
         mainNavigationController = UINavigationController(rootViewController: splashViewController)
         window = UIWindow(frame: UIScreen.main.bounds)
         window.rootViewController = mainNavigationController
+        window.tintColor = .element.accent
```
ismailgulek, line 52: Makes sense to me.

ismailgulek (approved):
LGTM!

PR #82, opened 2022-06-14
Add UserSessionStore and SplashScreen.

pixlwave:
- Adds the FTUE splash screen from EI with a single button for login.
- Adds a UserSessionStore and updates the state machine for this change (in a separate commit so easier to review).
- Replaces the code block background colour with one from DesignKit.

Note: I've left the splash screen as handled by the authentication coordinator so that you can pop from the login screen back to it like in EI.

Simulator Screen Shot - iPhone 13 - 2022-06-14 at 18 31 26


ElementX/Sources/AppCoordinator.swift
```diff
@@ -111,27 +104,36 @@ class AppCoordinator: AuthenticationCoordinatorDelegate, Coordinator {
     private func setupStateMachine() {
```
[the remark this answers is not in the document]
stefanceriu: Let's just disable it for this one please as this is a very special case. Generally I think 40 lines without whitespaces is more than reasonable.

ElementX/Sources/AppCoordinator.swift
```diff
@@ -149,17 +151,28 @@ class AppCoordinator: AuthenticationCoordinatorDelegate, Coordinator {
...
+                self.userSession = userSession
+                stateMachine.processEvent(.succeededRestoringSession)
+            case .failure:
+                MXLog.error("Failed to restore an existing session.")
+                stateMachine.processEvent(.failedRestoringSession)
+            }
         }
+    }
+    
+    private func showAuthentication() {
```
stefanceriu, line 169: Maybe `startAuthentication` would be a better name here considering it includes the onboarding screens.

stefanceriu, line 169: Maybe `startAuthentication` would be a better name here considering it includes the onboarding screens.

ElementX/Sources/Other/String.swift
```diff
@@ -0,0 +1,34 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import SwiftUI
+
+extension String {
```
stefanceriu, line 19: I wonder if we should name these primitive extensions something along the lines of `String+Element.swift` or `String+Tinging.swift` , similar to how we used to do it for categories in ObjC.

stefanceriu, line 19: I wonder if we should name these primitive extensions something along the lines of `String+Element.swift` or `String+Tinging.swift` , similar to how we used to do it for categories in ObjC.

ElementX/Sources/Screens/Authentication/AuthenticationCoordinator.swift
```diff
@@ -0,0 +1,133 @@
...
+    case failedLoggingIn
+}
+
+@MainActor
+protocol AuthenticationCoordinatorDelegate: AnyObject {
+    
+    func authenticationCoordinatorDidStartLoading(_ authenticationCoordinator: AuthenticationCoordinator)
+    
+    func authenticationCoordinator(_ authenticationCoordinator: AuthenticationCoordinator,
+                                   didLoginWithSession userSession: UserSession)
```
stefanceriu, line 22: Same as above, `UserSession` should be`UserSessionProtocol` instead.

stefanceriu, line 22: Same as above, `UserSession` should be`UserSessionProtocol` instead.

ElementX/Sources/Screens/Authentication/AuthenticationMetrics.swift
```diff
@@ -0,0 +1,28 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import SwiftUI
+
+/// Metrics used across the entire onboarding flow.
+struct AuthenticationMetrics {
```
stefanceriu, line 20: Metrics makes me think of analytics and tracking. Would `UIConstants` be a better name maybe?

stefanceriu, line 20: Metrics makes me think of analytics and tracking. Would `Constants` be a better name maybe?

stefanceriu (approved):
Looks great! 👏

ElementX/Sources/Screens/Authentication/AuthenticationMetrics.swift
pixlwave: Yep sure. I changed to Metrics in EI once it grew to include methods that returned non-constant values, but thats not a problem here :)

ElementX/Sources/AppCoordinator.swift
```diff
@@ -93,45 +88,52 @@ class AppCoordinator: AuthenticationCoordinatorDelegate, Coordinator {
...
     }
     
-    func authenticationCoordinator(_ authenticationCoordinator: AuthenticationCoordinator, didFailWithError error: AuthenticationCoordinatorError) {
-        stateMachine.processEvent(.failedSigningIn)
-    }
-    
-    func authenticationCoordinatorDidSetupClientProxy(_ authenticationCoordinator: AuthenticationCoordinator) {
+    func authenticationCoordinator(_ authenticationCoordinator: AuthenticationCoordinator, didLoginWithSession userSession: UserSession) {
+        self.userSession = userSession
+        remove(childCoordinator: authenticationCoordinator)
```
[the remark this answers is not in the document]
pixlwave: 1. That would be nice, its a shame the requirement is `Equatable` and not simple `Identifiable`.
2. Done

ElementX/Sources/Other/String.swift
pixlwave: Moving into an `Extensions` folder for now.

PR #83, opened 2022-06-17
Auto dismiss bug report when completed

### LINKED ISSUE #76: Bug report screen should disappear after sending
It feels weird to stay on this screen after upload. It looks like there is a sending issue.

ismailgulek:
Fixes #76

### DIFF
--- a/ElementX/Sources/Screens/Settings/SettingsCoordinator.swift
+++ b/ElementX/Sources/Screens/Settings/SettingsCoordinator.swift
@@ -33,6 +33,7 @@ final class SettingsCoordinator: Coordinator, Presentable {
     
     private var indicatorPresenter: UserIndicatorTypePresenterProtocol
     private var loadingIndicator: UserIndicator?
+    private var errorIndicator: UserIndicator?
     
     // MARK: Public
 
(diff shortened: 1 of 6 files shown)


ElementX/Sources/Screens/Settings/SettingsCoordinator.swift
```diff
@@ -106,4 +108,9 @@ final class SettingsCoordinator: Coordinator, Presentable {
             self.remove(childCoordinator: coordinator)
         }
     }
+
+    /// Show success indicator
+    private func showSuccess(label: String) {
+        errorIndicator = indicatorPresenter.present(.success(label: label))
```
pixlwave, line 114: I think I commented on one of these before how it seems weird to me to have an `errorIndicator` that shows `success`. Maybe it would make more sense to call this the `statusIndicator`?

pixlwave (approved):
LGTM 👍

ElementX/Sources/Screens/Settings/SettingsCoordinator.swift
ismailgulek: Makes perfect sense, updated throughout the app.

PR #85, opened 2022-06-20
Minimal Settings screen UI

### LINKED ISSUE #37: Minimal Settings Screen UI
We need a very minimal way to

- [x] logout
- [x] Report bug
- [x] App version + Build number
- [ ] Build number navigates to the release / PR

ismailgulek:
Fixes #37 with an exception: Links to the release part will be handled afterwards, when we have a clearer CD pipeline.

### DIFF
--- a/ElementX/Sources/AppCoordinatorStateMachine.swift
+++ b/ElementX/Sources/AppCoordinatorStateMachine.swift
@@ -80,7 +80,7 @@ class AppCoordinatorStateMachine {
             machine.addRoutes(event: .succeededRestoringSession, transitions: [ .restoringSession => .homeScreen ])
             machine.addRoutes(event: .failedRestoringSession, transitions: [ .restoringSession => .signedOut ])
             
-            machine.addRoutes(event: .attemptSignOut, transitions: [ .homeScreen => .signingOut ])
+            machine.addRoutes(event: .attemptSignOut, transitions: [ .settingsScreen => .signingOut ])
             
             machine.addRoutes(event: .succeededSigningOut, transitions: [ .signingOut => .signedOut ])
             machine.addRoutes(event: .failedSigningOut, transitions: [ .signingOut => .homeScreen ])
--- a/ElementX/Sources/Screens/Settings/View/Settings.swift
+++ b/ElementX/Sources/Screens/Settings/View/Settings.swift
@@ -19,6 +19,8 @@ import SwiftUI
 struct Settings: View {
 
     // MARK: Private
+
+    @State private var showingLogoutConfirmation = false
     
     // MARK: Public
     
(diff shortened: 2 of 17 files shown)


ElementX/Sources/AppCoordinatorStateMachine.swift
```diff
@@ -80,7 +80,7 @@ class AppCoordinatorStateMachine {
             machine.addRoutes(event: .succeededRestoringSession, transitions: [ .restoringSession => .homeScreen ])
             machine.addRoutes(event: .failedRestoringSession, transitions: [ .restoringSession => .signedOut ])
             
-            machine.addRoutes(event: .attemptSignOut, transitions: [ .homeScreen => .signingOut ])
+            machine.addRoutes(event: .attemptSignOut, transitions: [ .settingsScreen => .signingOut ])
             
             machine.addRoutes(event: .succeededSigningOut, transitions: [ .signingOut => .signedOut ])
             machine.addRoutes(event: .failedSigningOut, transitions: [ .signingOut => .homeScreen ])
```
pixlwave, line 86: Should this one go back to `.settingsScreen` now?

ElementX/Sources/Screens/Settings/View/Settings.swift
```diff
@@ -28,24 +30,46 @@ struct Settings: View {
...
     var body: some View {
         Form {
-            Button { context.send(viewAction: .reportBug) } label: {
-                Text(ElementL10n.sendBugReport)
+            Section {
+                Button { context.send(viewAction: .reportBug) } label: {
+                    Text(ElementL10n.sendBugReport)
+                }
+                .frame(maxWidth: .infinity)
+                .padding(.vertical, 16)
```
pixlwave, line 38: Could we drop the vertical padding on these buttons so the form matches the standard Settings app layout?
Edit: It might also be nice to remove the infinite frame and left align these first 2 buttons too :)

ElementX/Sources/Screens/Settings/View/Settings.swift
```diff
@@ -28,24 +30,46 @@ struct Settings: View {
...
+                }
             }
-            .frame(maxWidth: .infinity)
-            .padding(.vertical, 16)
-            .accessibilityIdentifier("reportBugButton")
 
-            if context.viewState.crashButtonVisible {
-                Button { context.send(viewAction: .crash) } label: {
-                    Text("Crash the app")
+            Section(footer: versionText) {
```
pixlwave, line 51: This init for adding a footer was soft-deprecated in iOS 15 and replaced by 
```
Section {
    // stuff
} footer: {
    // footer stuff
}
```

ElementX/Sources/Services/BugReport/BugReportService.swift
```diff
@@ -141,21 +141,32 @@ class BugReportService: BugReportServiceProtocol {
...
+            MultipartFormData(key: "utc_time", type: .text(value: utcTime))
         ]
     }
 
-    private var os: String {
-        "\(UIDevice.current.systemName) \(UIDevice.current.systemVersion)"
+    private func localAndUTCTime(for date: Date) -> (String, String) {
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "yyyy-MM-dd HH:mm:ss"
+        let localTime = dateFormatter.string(from: date)
```
pixlwave, line 162: Not sure if you would want to, but we can skip creating date formatters on iOS 15 and instead just do `Date().formatted(.iso8601)` (or whatever formatting options you want)

pixlwave (approved):
Looks good to me 😎

ElementX/Sources/Services/BugReport/BugReportService.swift
ismailgulek: Ah, i've checked the api but it requires lots of configuration to use our familiar date format. I prefer not to use it for bug report service.

ElementX/Sources/AppCoordinatorStateMachine.swift
ismailgulek: It was not used but yes it should, good catch 👍

PR #86, opened 2022-06-21
Room screen header

### LINKED ISSUE #35: Room screen: Header UI
We need an header to use in the room screen.

We could simply reused the one in this [design](https://www.figma.com/file/zVtAKK0hzKu5JkpbzlE1pG/Message-Bubbles?node-id=1%3A29] with no call buttons

ismailgulek:
Fixes #35


pixlwave (approved):
Nice! The room screen is starting looking like Element now 😎

PR #89, opened 2022-06-21
Update SplashScreen to match Element iOS.

pixlwave:
This is the ElementX version of https://github.com/vector-im/element-ios/pull/6320

- The gradient is now separate from the carousel pages
- The carousel pages are now in a VStack with the buttons rather than using a ZStack and ViewFrameReader.
- The performance issue was caused by the `maxWidth` value on the image on each page. Tweaking this seems to have resolved the frame drops on iPhone Mini and the value chosen doesn't cause it to occur on other iPhone models.

### DIFF
--- a/ElementX/Sources/Other/SwiftUI/Layout/ReadableFrameModifier.swift
+++ b/ElementX/Sources/Other/SwiftUI/Layout/ReadableFrameModifier.swift
@@ -1,0 +1,39 @@
+// 
+// Copyright 2022 New Vector Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import SwiftUI
+
+// swiftlint:disable private_over_fileprivate
+
+/// Positions this view within an invisible frame that fills the width of its parent view,
+/// whilst limiting the width of the content to a readable size (which is customizable).
+fileprivate struct ReadableFrameModifier: ViewModifier {
+    var maxWidth: CGFloat
+    
+    func body(content: Content) -> some View {
+        content
+            .frame(maxWidth: maxWidth)
+            .frame(maxWidth: .infinity)
+    }
+}
+
+extension View {
+    /// Positions this view within an invisible frame that fills the width of its parent view,
+    /// whilst limiting the width of the content to a readable size (which is customizable).
+    func readableFrame(maxWidth: CGFloat = 600) -> some View {
+        modifier(ReadableFrameModifier(maxWidth: maxWidth))
+    }
+}
--- a/ElementX/Sources/Screens/SplashScreen/SplashScreenModels.swift
+++ b/ElementX/Sources/Screens/SplashScreen/SplashScreenModels.swift
@@ -28,7 +28,6 @@ struct SplashScreenPageContent {
     let title: AttributedString
     let message: String
     let image: ImageAsset
-    let gradient: Gradient
 }
 
 // MARK: View model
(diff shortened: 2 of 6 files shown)


ElementX/Sources/Other/SwiftUI/Layout/ReadableFrameModifier.swift
```diff
@@ -0,0 +1,39 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import SwiftUI
+
+// swiftlint:disable private_over_fileprivate
```
pixlwave, line 19: The struct is `fileprivate` as it only needs to be used from the extension on `View`. Apparently swiftlint doesn't like this 🤷‍♂️

ElementX/Sources/Screens/SplashScreen/SplashScreenModels.swift
```diff
@@ -41,42 +40,45 @@ enum SplashScreenViewModelAction {
...
+    ]
     
     /// An array containing all content of the carousel pages
     let content: [SplashScreenPageContent]
     var bindings: SplashScreenBindings
     
+    /// The background gradient for all 4 pages and the hidden page at the start of the carousel.
+    var backgroundGradient: Gradient {
+        // Include the extra stop for the hidden page at the start of the carousel.
+        let hiddenPageColor = gradientColors[gradientColors.count - 2]
```
ismailgulek, line 60: why minus 2?

ismailgulek (approved):
Some comments/questions inline, otherwise LGTM!

ElementX/Sources/Other/SwiftUI/Layout/ReadableFrameModifier.swift
```diff
@@ -0,0 +1,39 @@
...
+    var maxWidth: CGFloat
+    
+    func body(content: Content) -> some View {
+        content
+            .frame(maxWidth: maxWidth)
+            .frame(maxWidth: .infinity)
+    }
+}
+
+extension View {
```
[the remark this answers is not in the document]
pixlwave: Personally my preference would be to continue as we have been in EI for modifiers where the file is named `XYZModifier` and includes the view extension without needing to signal this. E.g. [ScreenTrackerViewModifier](https://github.com/vector-im/element-ios/blob/develop/RiotSwiftUI/Modules/Common/Util/ScreenTrackerViewModifier.swift), [ClearViewModifier](https://github.com/vector-im/element-ios/blob/develop/RiotSwiftUI/Modules/Common/Util/ClearViewModifier.swift) and [BorderModifier](https://github.com/vector-im/element-ios/blob/develop/RiotSwiftUI/Modules/Common/Util/BorderModifier.swift). That said, its a good question and probably worth bringing up next week in the Friday discussions.

ElementX/Sources/Other/SwiftUI/Layout/ReadableFrameModifier.swift
```diff
@@ -0,0 +1,39 @@
...
+
+/// Positions this view within an invisible frame that fills the width of its parent view,
+/// whilst limiting the width of the content to a readable size (which is customizable).
+fileprivate struct ReadableFrameModifier: ViewModifier {
+    var maxWidth: CGFloat
+    
+    func body(content: Content) -> some View {
+        content
+            .frame(maxWidth: maxWidth)
+            .frame(maxWidth: .infinity)
```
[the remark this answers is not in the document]
pixlwave: Yeah, it's a pretty clear example of how `.frame` doesn't actually set the frame on the view, but rather embeds it in a view with the frame you defined. It raised comments before too 🙃 https://github.com/vector-im/element-ios/pull/6125#discussion_r866822922

ElementX/Sources/Screens/SplashScreen/SplashScreenModels.swift
pixlwave: Minus 1 would be the right-hand stop of page 4 which we already have as the left-hand stop of page 1. So when adding the hidden page we need the left-hand stop of page 4 which is minus 2. Will add as a comment.

PR #90, opened 2022-06-22
Update DesignTokens to 0.0.2

pixlwave:
The DesignTokens repo now generates the `contentAndAvatars` array along with moving the global colours into a new static `global` struct to separate dynamic and non-dynamic colors.


ismailgulek (approved):
LGTM!

PR #98, opened 2022-06-27
Add the login screen from EI.

pixlwave:
This PR adds the login screen from Element iOS but with the following changes:

- Rename fallback to OIDC (not implemented).
- Remove SSO options in favour of OIDC.
- Select server is disabled and username parsing is faked for now (pending SDK changes).
- Forgot password flow not implemented.

Additional small changes include:
- Splash screen no longer includes the register action as this will all happen via OIDC.
- User indicators use a system material for now so they fit in a bit better.
- Added `MXLog.failure` from EI.
- Run SwiftGen as a pre-build step so strings are generated when there's an error elsewhere in the project.

| Light Mode | Dark Mode | Dark mode loading | Light mode success |
| - | - | - | - |
| Simulator Screen Shot - iPhone 13 mini - 2022-06-27 at 16 18 36 | Simulator Screen Shot - iPhone 13 mini - 2022-06-27 at 16 19 11 | Simulator Screen Shot - iPhone 13 mini - 2022-06-27 at 16 19 15 | Simulator Screen Shot - iPhone 13 mini - 2022-06-27 at 16 20 15 |


ElementX/Sources/Other/SwiftUI/ErrorHandling/AlertInfo.swift
```diff
@@ -0,0 +1,87 @@
...
+    let title: String
+    /// The alert's message (optional).
+    var message: String?
+    /// The alert's primary button title and action. Defaults to an Ok button with no action.
+    var primaryButton: (title: String, action: (() -> Void)?) = (ElementL10n.ok, nil)
+    /// The alert's secondary button title and action.
+    var secondaryButton: (title: String, action: (() -> Void)?)?
+}
+
+#warning("Remove the NSError extensions?")
```
pixlwave, line 39: These initialisers for `NSError` we for the old SDK's `MXError`. Seems sensible to remove these now but thought it was worth checking.

ismailgulek: Removing them makes sense to me 👍
Maybe instead we could introduce an `Error` extension, something like this:
```
extension AlertInfo {
    init(error: Error) where T == String {
        self.init(id: error.localizedDescription,
                  title: error.localizedDescription)
    }
}
```

ElementX/Sources/Screens/Authentication/LoginScreen/LoginViewModel.swift
```diff
@@ -0,0 +1,78 @@
...
+        case .alert(let message):
+            state.bindings.alertInfo = AlertInfo(id: type,
+                                                 title: ElementL10n.dialogTitleError,
+                                                 message: message)
+        case .invalidHomeserver:
+            state.bindings.alertInfo = AlertInfo(id: type,
+                                                 title: ElementL10n.dialogTitleError,
+                                                 message: ElementL10n.loginSigninMatrixIdErrorInvalidMatrixId)
+        case .unknown:
+            state.bindings.alertInfo = AlertInfo(id: type)
```
ismailgulek, line 75: This `AlertInfo` initializer would fail if we remove `NSError` extension.

```suggestion
            state.bindings.alertInfo = AlertInfo(id: type,
                                                 title: ElementL10n.dialogTitleError)
```

ismailgulek (approved):
Some minor comments, otherwise LGTM!

UITests/Sources/LoginScreenUITests.swift
```diff
@@ -15,14 +15,129 @@
 //
 
 import XCTest
+import ElementX
 
-class LoginScreenUITests: XCTestCase {
-    func testInitialStateComponents() {
-        let app = Application.launch()
+@MainActor
+class AuthenticationLoginUITests: XCTestCase {
```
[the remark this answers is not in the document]
pixlwave: Yep my bad, missed that one 👍

UnitTests/Sources/LoginViewModelTests.swift
```diff
@@ -0,0 +1,140 @@
...
+        
+        // When entering a username and an 8-character password.
+        context.username = "bob"
+        context.password = "12345678"
+        
+        // Then the credentials should be considered valid.
+        XCTAssertTrue(context.viewState.hasValidCredentials, "The credentials should be valid when the username and password are valid.")
+    }
+    
+    func testLoadingServer() {
```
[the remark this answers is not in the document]
pixlwave: Good spot I added that one later, definitely should be tested 👍

ElementX/Sources/Screens/Authentication/LoginScreen/LoginMode.swift
```diff
@@ -0,0 +1,56 @@
...
+        default:
+            return false
+        }
+    }
+    
+    var supportsPasswordFlow: Bool {
+        switch self {
+        case .password:
+            return true
+        case .unknown, .unsupported, .oidc:
```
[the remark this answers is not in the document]
pixlwave: I think it made a little bit of sense with the old cases, but even then you're right. Gosh I hope the [`is case`](https://forums.swift.org/t/proposal-draft-for-is-case-pattern-match-boolean-expressions/58260) pitch makes it into the language!

ElementX/Sources/Screens/Authentication/LoginScreen/LoginViewModel.swift
pixlwave: I've added a generic `init(id: T)` for this instance to match the old one and so its easy to switch them all at once.

PR #100, opened 2022-06-28
Background execution

### LINKED ISSUE #99: Background Execution
We'd like to continue execution when the app goes into background.

ismailgulek:
Fixes #99

### DIFF
--- a/ElementX/Sources/AppCoordinator.swift
+++ b/ElementX/Sources/AppCoordinator.swift
@@ -26,6 +26,7 @@ class AppCoordinator: AuthenticationCoordinatorDelegate, Coordinator {
 
     private let bugReportService: BugReportServiceProtocol
     private let screenshotDetector: ScreenshotDetector
+    private let backgroundTaskService: BackgroundTaskServiceProtocol
 
     private var indicatorPresenter: UserIndicatorTypePresenterProtocol
     private var loadingIndicator: UserIndicator?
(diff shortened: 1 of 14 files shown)


ElementX/Sources/AppCoordinator.swift
```diff
@@ -61,6 +62,8 @@ class AppCoordinator: AuthenticationCoordinatorDelegate, Coordinator {
         
         userSessionStore = UserSessionStore(bundleIdentifier: bundleIdentifier)
 
+        backgroundTaskService = UIKitBackgroundTaskService()
```
stefanceriu, line 65: Should we maybe pass the shared UIApplication directly in the constructor instead of letting it rely on that internal `extensionSafeShared` property? Seems weird to have it reach up and fetch the app by itself.

stefanceriu (approved):
This looks great!  The only real question I have is whether we should handle these background tasks directly on the final implementation within the `RoomProxy` and the `MediaProvider`/`ClientProxy`. Those are where the actual work is happening and can be shared between multiple components.

ElementX/Sources/AppCoordinator.swift
ismailgulek: Yes, makes sense for the `AppCoordinator`, updated.

ismailgulek:
> This looks great! The only real question I have is whether we should handle these background tasks directly on the final implementation within the `RoomProxy` and the `MediaProvider`/`ClientProxy`. Those are where the actual work is happening and can be shared between multiple components.

Thanks for the suggestion, moved into them.

PR #121, opened 2022-06-29
Flatten room list

ismailgulek:
Flattens the room list by removing encrypted groups. Also styles the screen a bit.

Simulator Screen Shot - iPhone 13 Pro Max - 2022-06-30 at 02 32 20

### DIFF
--- a/ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
+++ b/ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
@@ -36,20 +36,12 @@ struct HomeScreenViewState: BindableState {
     
     var isLoadingRooms: Bool = false
     
-    var unencryptedDMs: [HomeScreenRoom] {
-        searchFilteredRooms.filter { $0.isDirect && !$0.isEncrypted }
+    var visibleDMs: [HomeScreenRoom] {
+        searchFilteredRooms.filter { $0.isDirect }
     }
-    
-    var encryptedDMs: [HomeScreenRoom] {
-        searchFilteredRooms.filter { $0.isDirect && $0.isEncrypted}
-    }
-    
-    var unencryptedRooms: [HomeScreenRoom] {
-        searchFilteredRooms.filter { !$0.isDirect && !$0.isEncrypted }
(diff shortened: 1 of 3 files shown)


ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
```diff
@@ -77,41 +56,48 @@ struct HomeScreen: View {
...
-                                    .fontWeight(.bold)
-                                    .foregroundColor(.primary)
-                            }
-                        } else {
-                            EmptyView()
-                        }
-                    }
-                    .animation(.default, value: context.viewState.userDisplayName)
-                    .transition(.opacity)
+                HStack(spacing: 0) {
```
pixlwave, line 59: Chiming in un-invited here: Could we make this HStack the label of a single button, rather than having 2 buttons in the HStack? I've been meaning to change it for a while now.

ismailgulek: Makes perfect sense, updated.

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
```diff
@@ -36,20 +36,12 @@ struct HomeScreenViewState: BindableState {
...
-    }
-    
-    var unencryptedRooms: [HomeScreenRoom] {
-        searchFilteredRooms.filter { !$0.isDirect && !$0.isEncrypted }
-    }
-    
-    var encryptedRooms: [HomeScreenRoom] {
-        searchFilteredRooms.filter { !$0.isDirect && $0.isEncrypted }
+
+    var visibleRooms: [HomeScreenRoom] {
```
stefanceriu, line 43: I wouldn't necessarily call these `visible`. Makes me think of subset of rooms that are currently visible on screen.
Perhaps plainly `rooms` and `directMessages` works better.

stefanceriu (approved):
lgtm 👏

pixlwave (approved):
Not sure if you need my full review too, but if you do... ✅

PR #126, opened 2022-06-30
Add basic AuthenticationService and missing UI tests.

pixlwave:
Moves authentication into a basic AuthenticationService, adds a mock of this service and tests for both server selection and the whole auth flow.


ElementX/Sources/Screens/Authentication/LoginScreen/LoginCoordinator.swift
```diff
@@ -136,55 +128,66 @@ final class LoginCoordinator: Coordinator, Presentable {
...
+            }
         }
-        
-        callback?(.login(username: username, password: password))
     }
     
     /// Parses the specified username and looks up the homeserver when a Matrix ID is entered.
     private func parseUsername(_ username: String) {
-        guard isMXID(username: username) else { return }
+        guard authenticationService.usernameIsMatrixID(username) else { return }
```
pixlwave, line 158: I don't like this method name, but failed to come up with something better.

ElementX/Sources/Services/Client/ClientError.swift
```diff
@@ -0,0 +1,30 @@
...
+        
+        for code in MatrixErrorCode.allCases {
+            if message.contains(code.rawValue) {
+                return code
+            }
+        }
+        
+        return .unknown
+    }
+}
```
pixlwave, line 30: These are basic implementations for now. I'd like to learn whether the FFI bridge could return a `.matrix(code: String, description: String` case for the `ClientError` to avoid this wrangling.

UITests/Sources/LoginScreenUITests.swift
```diff
@@ -123,7 +138,7 @@ class LoginScreenUITests: XCTestCase {
     /// Checks that the OIDC button is shown on the screen.
     func validateOIDCButtonIsShown(for state: String) {
         let oidcButton = app.buttons["oidcButton"]
-        XCTAssertTrue(oidcButton.exists, "The OIDC button should be shown for \(state).")
+        XCTAssertTrue(oidcButton.waitForExistence(timeout: 1), "The OIDC button should be shown for \(state).")
```
pixlwave, line 141: Seems needed for CI, ran fine locally without this.

ElementX/Sources/Screens/Authentication/LoginScreen/LoginCoordinator.swift
ismailgulek: How about `isMatrixUserId`? Also does it make sense to move this function to a helper class like we had `MXTools`?

ElementX/Sources/UITestsAppCoordinator.swift
```diff
@@ -11,63 +11,72 @@ import SwiftUI
...
     init() {
-        mainNavigationController = UINavigationController()
+        mainNavigationController = ElementNavigationController()
+        navigationRouter = NavigationRouter(navigationController: mainNavigationController)
+        
         window = UIWindow(frame: UIScreen.main.bounds)
         window.rootViewController = mainNavigationController
         window.tintColor = .element.accent
         
+        UIView.setAnimationsEnabled(false)
```
ismailgulek, line 28: Thanks for adding this. I had to also disable it for screenshot tests

UITests/Sources/RoomScreenUITests.swift
```diff
@@ -20,23 +20,19 @@ import ElementX
 @MainActor
 class RoomScreenUITests: XCTestCase {
 
-    func testPlainNoAvatar() async throws {
+    func testPlainNoAvatar() {
         let app = Application.launch()
         app.goToScreenWithIdentifier(.roomPlainNoAvatar)
 
-        try await Task.sleep(nanoseconds: 400_000_000)
```
ismailgulek, line 27: Thanks for removing these. How do they work now?

ismailgulek (approved):
Looks great to me. It's perfect to see our coverage is climbing 👍

PR #129, opened 2022-07-01
Add SwiftFormat to the project.

pixlwave:
This PR adds SwiftFormat to the project as a build phase and runs it for the first time. I've gone for a regular pre-build phase on the main target which the [docs](https://github.com/nicklockwood/SwiftFormat#xcode-build-phase) don't particularly recommend (instead suggesting to use a step on the tests target). Seems like this should be ok for now, but happy to update now if deemed sensible.

Another option would be to never run manually, instead using [Danger](https://github.com/nicklockwood/SwiftFormat#on-ci-using-danger) to inform you to run it on a PR.

### DIFF
--- a/.swiftformat
+++ b/.swiftformat
@@ -1,0 +1,15 @@
+--swiftversion 5.6
+
+--exclude ElementX/Sources/Generated
+
+--disable all
+
+--enable blankLinesAroundMark
+--enable blankLinesBetweenScopes
+--enable consecutiveBlankLines
+--enable consecutiveSpaces
+--enable linebreakAtEndOfFile
+--enable linebreaks
+--enable trailingSpace
+
+--trimwhitespace nonblank-lines
--- a/.swiftformat
+++ b/.swiftformat
@@ -2,14 +2,16 @@
 
 --exclude ElementX/Sources/Generated
 
---disable all
+--disable wrapMultiLineStatementBraces
+--disable blankLinesAtStartOfScope
+--disable blankLinesAtEndOfScope
+--disable hoistPatternLet
 
---enable blankLinesAroundMark
---enable blankLinesBetweenScopes
---enable consecutiveBlankLines
---enable consecutiveSpaces
---enable linebreakAtEndOfFile
---enable linebreaks
---enable trailingSpace
-
+--commas inline
+--ifdef no-indent
+--nospaceoperators ...,..<
+--stripunusedargs closure-only
 --trimwhitespace nonblank-lines
+--wrapparameters after-first
+--redundanttype inferred
+--emptybraces spaced
--- a/ElementX/Sources/AppDelegate.swift
+++ b/ElementX/Sources/AppDelegate.swift
@@ -15,7 +15,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     func application(_ application: UIApplication, willFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]? = nil) -> Bool {
 
-        //  fixme: Use `Bundle.elementLanguage = ".."` when we have the functionality
+        //  FIXME: Use `Bundle.elementLanguage = ".."` when we have the functionality
         //  use `en` as fallback language
         Bundle.elementFallbackLanguage = "en"
 
@@ -34,17 +34,17 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     
     private var isRunningUnitTests: Bool {
         #if DEBUG
-            ProcessInfo.processInfo.environment["IS_RUNNING_UNIT_TESTS"] == "1"
+        ProcessInfo.processInfo.environment["IS_RUNNING_UNIT_TESTS"] == "1"
         #else
-            false
+        false
         #endif
     }
     
     private var isRunningUITests: Bool {
         #if DEBUG
-            ProcessInfo.processInfo.environment["IS_RUNNING_UI_TESTS"] == "1"
+        ProcessInfo.processInfo.environment["IS_RUNNING_UI_TESTS"] == "1"
         #else
-            false
+        false
         #endif
     }
 }
--- a/ElementX/Sources/AppDelegate.swift
+++ b/ElementX/Sources/AppDelegate.swift
@@ -14,8 +14,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     private lazy var appCoordinator: Coordinator = isRunningUITests ? UITestsAppCoordinator() : AppCoordinator()
 
     func application(_ application: UIApplication, willFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]? = nil) -> Bool {
-
-        //  FIXME: Use `Bundle.elementLanguage = ".."` when we have the functionality
         //  use `en` as fallback language
         Bundle.elementFallbackLanguage = "en"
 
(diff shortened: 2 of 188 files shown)


ElementX/Sources/Other/ImageAnonymizer.swift
```diff
@@ -7,14 +7,14 @@
...
-import Vision
 import UIKit
+import Vision
 
 enum ImageAnonymizerError: Error {
     case noCgImageBased
 }
 
-struct ImageAnonymizer {
+enum ImageAnonymizer {
```
ismailgulek, line 17: Hmm i wonder what is the logic behind this.

ElementX/Sources/Other/Logging/MXLog.swift
```diff
@@ -107,8 +107,8 @@ private var logger: SwiftyBeaver.Type = {
         logger.info(message, file, function, line: line)
     }
     
-    public static func warning(_ message: @autoclosure () -> Any, _
-                                file: String = #file,
+    public static func warning(_ message: @autoclosure () -> Any, _ file: String = #file,
```
ismailgulek, line 110: I'd love to remove this rule, doesn't make perfect sense to me.

ElementX/Sources/Other/Routers/RootRouter.swift
```diff
@@ -76,11 +76,11 @@ final class RootRouter: RootRouterType {
                 UIView.setAnimationsEnabled(false)
                 self.window.rootViewController = rootViewController
                 UIView.setAnimationsEnabled(oldState)
-            }, completion: { (_: Bool) -> Void in
+            }, completion: { (_: Bool) in
```
ismailgulek, line 79: I wonder why it didn't choose `{ _ in`

ElementX/Sources/Services/SessionVerification/MockSessionVerificationControllerProxy.swift
```diff
@@ -56,7 +56,6 @@ struct MockSessionVerificationControllerProxy: SessionVerificationControllerProx
          SessionVerificationEmoji(symbol: "🎂", description: "Cake"),
          SessionVerificationEmoji(symbol: "🎂", description: "Cake"),
          SessionVerificationEmoji(symbol: "🏁", description: "Flag"),
-         SessionVerificationEmoji(symbol: "🌏", description: "Globe")
-        ]
+         SessionVerificationEmoji(symbol: "🌏", description: "Globe")]
```
ismailgulek, line 59: Personally I'd love to revert this, as i find it easier to alter the array.

ElementX/Sources/Services/Room/RoomSummary/MockRoomSummary.swift
```diff
@@ -18,21 +18,19 @@ struct MockRoomSummary: RoomSummaryProtocol {
...
+    var isTombstoned = false
     
     var lastMessage: EventBrief?
     
     var avatar: UIImage?
     
-    func loadDetails() async {
-        
-    }
+    func loadDetails() async { }
```
ismailgulek, line 33: Is Sonarcloud happy with this?

.swiftformat
```diff
@@ -0,0 +1,17 @@
+--swiftversion 5.6
+
+--exclude ElementX/Sources/Generated
+
+--disable wrapMultiLineStatementBraces
+--disable blankLinesAtStartOfScope
+--disable blankLinesAtEndOfScope
```
ismailgulek, line 7: Is there a specific reason to disable them? I find them useful.

ismailgulek (commented):
Looks pretty good to me.

ElementX/Sources/Other/ImageAnonymizer.swift
pixlwave: Oh hmmm I hadn't noticed this one. It's the `enumNamespaces` rule:
https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#enumnamespaces

I can see why it has done it but does feel weird.

ElementX/Sources/Services/SessionVerification/MockSessionVerificationControllerProxy.swift
pixlwave: From what I can tell this rule doesn't enforce one-way or the other, what happened here was the starting `[` mismatched the ending `]` like so:
```
[Emoji
 Emoji,
 Emoji
]
```

Testing with
```
[
  Emoji,
  Emoji,
  Emoji
]
```
And it left it alone.

ElementX/Sources/Other/Routers/RootRouter.swift
pixlwave: Oh yeah, odd. Will update manually 👍

ElementX/Sources/Services/Room/RoomSummary/MockRoomSummary.swift
pixlwave: TBH, I don't understand the logic on Sonarcloud. Its picked out 3 of these and ignored the rest. Either way though, any kind of formatting won't help here as Sonarcloud wants a `/* no-op */` comment in there.

ElementX/Sources/Other/Logging/MXLog.swift
pixlwave: Yeah I didn't understand it either, after experimenting its because the first line ends in `, _` and its moving the `file: String` up alongside it. Putting that whole parameter together on its own line and the formatter leaves it alone, so whichever rule was doing seems fine to leave enabled.

Will fix it manually.

ElementX/Sources/AppDelegate.swift
```diff
@@ -15,8 +15,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     func application(_ application: UIApplication, willFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]? = nil) -> Bool {
 
-        //  fixme: Use `Bundle.elementLanguage = ".."` when we have the functionality
-        //  use `en` as fallback language
+        // FIXME: Use `Bundle.elementLanguage = ".."` when we have the functionality
```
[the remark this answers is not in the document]
pixlwave: Will do

pixlwave (commented):
Thanks for your comments, rule wise they all seem like false positives to me, have fixed them.

ElementX/Sources/Other/ImageAnonymizer.swift
ismailgulek: Ah the explanation makes sense in some manner :)

ElementX/Sources/Services/SessionVerification/MockSessionVerificationControllerProxy.swift
ismailgulek: Ah makes sense 👍 Thanks for checking.

ElementX/Sources/Services/Room/RoomSummary/MockRoomSummary.swift
ismailgulek: Yeah it's a bit sad. Let's keep it as is right now, we can consider adding no-op comments in the future if becomes annoying.

.swiftformat
pixlwave: I'll enable in a separate PR to see what the change looks like in isolation 👍

PR #132, opened 2022-07-06
Enable blank line rules for swiftformat.

pixlwave:
@ismailgulek You were right, these are good too. TBH I thought they did something around `// MARK:` comments which was why I disabled them. Definitely makes sense to enable these 👍


ismailgulek (approved):
LGTM 🤩

PR #133, opened 2022-07-18
Prevent images cached on disk from loading on the main queue

stefanceriu:


ElementX/Sources/Services/Media/MediaProvider.swift
```diff
@@ -54,18 +54,18 @@ struct MediaProvider: MediaProviderProtocol {
...
-            }
-        }
-        
-        if case let .success(cacheResult) = cachedImageLoadResult,
-           let image = cacheResult.image {
-            return .success(image)
-        }
-        
         return await Task.detached { () -> Result<UIImage, MediaProviderError> in
+            let cachedImageLoadResult = await withCheckedContinuation { continuation in
```
pixlwave, line 58: Might be nice to move this continuation into an async function on an extension of the cache object.

pixlwave (approved):
LGTM 👍

ElementX/Sources/Services/Media/MediaProvider.swift
stefanceriu: Nice one 👍

PR #138, opened 2022-07-20
vector-im/element-x-ios/issues/134 - Prevent home screen room last me…

stefanceriu:
…ssage attributed string from being computed on every update

### DIFF
--- a/ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
+++ b/ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
@@ -23,7 +23,11 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol 
     private let attributedStringBuilder: AttributedStringBuilderProtocol
     
     private var roomUpdateListeners = Set<AnyCancellable>()
-    private var roomsUpdateTask: Task<Void, Never>?
+    private var roomsUpdateTask: Task<Void, Never>? {
+        willSet {
+            roomsUpdateTask?.cancel()
+        }
+    }
 
     private var roomSummaries: [RoomSummaryProtocol]? {
         didSet {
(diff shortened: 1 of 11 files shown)


ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -58,12 +59,25 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
     func updateWithRoomSummaries(_ roomSummaries: [RoomSummaryProtocol]) {
         self.roomSummaries = roomSummaries
         
-        state.rooms = roomSummaries.map { roomSummary in
-            buildOrUpdateRoomFromSummary(roomSummary)
+        roomsUpdateTask?.cancel()
```
pixlwave, line 62: Might be worthwhile to move this to the `willSet` of `roomsUpdateTask` in case it is ever updated from another method.

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -58,12 +59,25 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
...
+        roomsUpdateTask = Task {
```
pixlwave, line 63: Maybe this should have a `[weak self]` as its stored, although I didn't look any further into it on the EI side after our discussion the other day.

ElementX/Sources/Services/Timeline/RoomTimelineController.swift
```diff
@@ -96,10 +97,21 @@ class RoomTimelineController: RoomTimelineControllerProtocol {
     }
     
     private func updateTimelineItems() {
+        timelineItemsUpdateTask?.cancel()
+        timelineItemsUpdateTask = Task {
```
pixlwave, line 101: Same thoughts about `willSet` and `[weak self]`.

ElementX/Sources/Services/Timeline/RoomTimelineController.swift
```diff
@@ -96,10 +97,21 @@ class RoomTimelineController: RoomTimelineControllerProtocol {
...
+            await asyncUpdateTimelineItems()
+        }
+    }
+    
+    private func asyncUpdateTimelineItems() async {
```
pixlwave, line 106: Does the method name need `async` in it?

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemFactory.swift
```diff
@@ -47,8 +47,8 @@ struct RoomTimelineItemFactory: RoomTimelineItemFactoryProtocol {
                                                   _ isOutgoing: Bool,
                                                   _ showSenderDetails: Bool,
                                                   _ displayName: String?,
-                                                  _ avatarImage: UIImage?) -> RoomTimelineItemProtocol {
-        let attributedText = (message.htmlBody != nil ? attributedStringBuilder.fromHTML(message.htmlBody) : attributedStringBuilder.fromPlain(message.body))
+                                                  _ avatarImage: UIImage?) async -> RoomTimelineItemProtocol {
+        let attributedText = (message.htmlBody != nil ? await attributedStringBuilder.fromHTML(message.htmlBody) : await attributedStringBuilder.fromPlain(message.body))
```
pixlwave, line 51: I think you could write these with a single await before the ternary expression rather than having 2.

pixlwave (approved):
LGTM, a few minor comments inline.

ElementX/Sources/Services/Timeline/RoomTimelineController.swift
stefanceriu: Yes, unfortunately Xcode complains about `Expression is 'async' but is not marked with 'await'` in the `paginateBackwards` success otherwise ¯\_(ツ)_/¯

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
stefanceriu: Seems that this goes hand in hand with not requiring explicit self as creating retain loops will only happen if the task never finishes. https://forums.swift.org/t/explicit-self-not-required-for-task-closures/54364/8

PR #139, opened 2022-07-22
Switch SwiftFormat to lint mode and have it be lenient on local build…

stefanceriu:
…s but fail on the CI

### DIFF
--- a/ElementX.xcodeproj/project.pbxproj
+++ b/ElementX.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 52;
+	objectVersion = 51;
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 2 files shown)


ElementX.xcodeproj/project.pbxproj
```diff
@@ -1493,7 +1486,6 @@
 				E74CD7681375AD2EAA34D66B /* Authentication */,
 				4009BE2E791C16AC6EE39A7E /* BugReport */,
 				B53CA9BECD3F97805E1432D0 /* HomeScreen */,
-				3180C73BA7B8F5F7447C99B0 /* React */,
```
pixlwave, line 1496: What was this?

pixlwave (approved):
LGTM. Presumably without the warning of the missing tool, the build will hard fail? I wonder if we should do that with the other 2 tools too 🤔

ElementX.xcodeproj/project.pbxproj
stefanceriu: Absolutely no idea, it seems you introduced it in https://github.com/vector-im/element-x-ios/blob/26d8affaa8821971a21e159467de19fca337b635/ElementX.xcodeproj/project.pbxproj#L845

pixlwave: Oh weird, I have no idea what thats about. Good riddance lol!

PR #142, opened 2022-07-22
Improve release process

stefanceriu:
- remove towncrier logs from PR builds and move them to github releases
- set build version based on current time
- bump app version


pixlwave (approved):
LGTM 🚀

PR #143, opened 2022-07-25
Release process improvements

stefanceriu:
Automatically bump the app version and create a commit with all the changes required for the next version


fastlane/Fastfile
```diff
@@ -160,11 +146,31 @@ private_lane :release_to_github do
...
+
+  sh("git add #{target_file_path} #{xcode_project_file_path}")
+
+  sh("git commit -m 'Prepare next release'")
 end
 
 private_lane :export_version_changes do |options|
   Dir.chdir("..") do
-    # Changelog.update_topmost_section(version: options[:version], additional_entries: {})
+    Changelog.update_topmost_section(version: options[:version], additional_entries: {})
```
pixlwave, line 173: Double checking that this won't affect the changes to ElementPR you made last week (IIRC aren't showing towncrier in the PR so I'm sure its fine).

pixlwave (approved):
LGTM 👍

fastlane/Fastfile
stefanceriu: Yep, I comitted this by mistake. We **do** want towncrier to generate notes for each release version.

PR #144, opened 2022-07-26
Use the Rust AuthenticationService.

pixlwave:
PR that uses the Rust authentication service for password logins. Additionally

- Uses the app group container for session storage.
- Renames `accessToken` to `restoreToken` in the keychain controller.
- Updates to the latest version of the Rust SDK.

Closes #40.

**Signing in to localhost**
Simulator Screen Recording - iPhone 13 - 2022-07-26 at 16 05 44

PR #145, opened 2022-07-26
Proof of concept login via OIDC.

### LINKED ISSUE #42: I want to sign in with OIDC
OIDC starts to be ready. We should showcase it using ElementX-iOS Alpha. It will help the OIDC team to get more feedbacks and to improve the UX.

After https://github.com/vector-im/element-x-ios/issues/40, we need to detect from the selected homeserver `.well-known` the `.m.authentication` field ([example](https://synapse-oidc.lab.element.dev/.well-known/matrix/client)) and to display a "Log in via OIDC" that will open the authentication web service.

A sample swift app is available [here](https://gitlab.matrix.org/new-vector/oidc/OidcTest-iOS).
Test homeserver with OIDC: https://synapse-oidc.lab.element.dev/
Hydrogen with OIDC in action: https://hydrogen-oidc.lab.element.dev/

pixlwave:
The implementation uses AppAuth-iOS in the app to negotiate with the OIDC server and ignores the refresh token so the login will only work for a short amount of time.

There's currently an issue where the Rust runtime crashes when starting sync if you have an expired token - the app has no way to recover from this, so it is a good idea to logout within 5-minutes of logging in.

Closes #42. Depends on #144.

### DIFF
--- a/ElementX/Sources/Services/Authentication/AuthenticationServiceProxy.swift
+++ b/ElementX/Sources/Services/Authentication/AuthenticationServiceProxy.swift
@@ -21,7 +21,6 @@ class AuthenticationServiceProxy: AuthenticationServiceProxyProtocol {
     // MARK: Public
     
     private(set) var homeserver = LoginHomeserver(address: BuildSettings.defaultHomeserverAddress, loginMode: .unknown)
-    var oidcUserAgent: OIDExternalUserAgentIOS?
     
     // MARK: - Setup
     
@@ -61,11 +60,8 @@ class AuthenticationServiceProxy: AuthenticationServiceProxyProtocol {
         }
     }
     
-    func loginWithOIDC() async -> Result<UserSessionProtocol, AuthenticationServiceError> {
-        guard
-            let oidcUserAgent = oidcUserAgent,
-            case let .oidc(issuerURL) = homeserver.loginMode
-        else {
+    func loginWithOIDC(userAgent: OIDExternalUserAgentIOS) async -> Result<UserSessionProtocol, AuthenticationServiceError> {
+        guard case let .oidc(issuerURL) = homeserver.loginMode else {
             return .failure(.oidcError(.notSupported))
         }
         
@@ -78,7 +74,7 @@ class AuthenticationServiceProxy: AuthenticationServiceProxyProtocol {
             let authResponse = try await oidcService.presentWebAuthentication(metadata: configuration,
                                                                               clientID: registationResponse.clientID,
                                                                               scope: "urn:matrix:device:\(deviceID)",
-                                                                              userAgent: oidcUserAgent)
+                                                                              userAgent: userAgent)
             let tokenResponse = try await oidcService.redeemCodeForTokens(authResponse: authResponse)
             
             guard let accessToken = tokenResponse.accessToken else { return .failure(.oidcError(.unknown)) }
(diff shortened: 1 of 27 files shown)


ElementX/Sources/Services/Authentication/OIDCService.swift
```diff
@@ -0,0 +1,135 @@
...
+        let scopesArray = scope.components(separatedBy: " ")
+        let request = OIDAuthorizationRequest(configuration: metadata,
+                                              clientId: clientID,
+                                              clientSecret: nil,
+                                              scopes: scopesArray,
+                                              redirectURL: redirectURI,
+                                              responseType: OIDResponseTypeCode,
+                                              additionalParameters: nil)
+        let result: OIDAuthorizationResponse = try await withCheckedThrowingContinuation { continuation in
+            self.session = OIDAuthorizationService.present(request, externalUserAgent: userAgent) { response, error in
```
stefanceriu, line 90: We could do the same thing we did for kingfisher and move this continuation to an extension on OID.

ElementX/Sources/Services/Authentication/OIDCService.swift
```diff
@@ -0,0 +1,135 @@
...
+        #warning("Persist the client ID for reuse.")
+        return registrationResponse
+    }
+    
+    /// Trigger a redirect with standard parameters.
+    /// `acr_values` can be sent as an extra parameter, to control authentication methods.
+    func presentWebAuthentication(metadata: OIDServiceConfiguration,
+                                  clientID: String,
+                                  scope: String,
+                                  userAgent: OIDExternalUserAgent) async throws -> OIDAuthorizationResponse {
```
stefanceriu, line 80: We should probably use Result here instead of throwing exceptions.

stefanceriu (changes requested):
Just a few comments inline 👍

ElementX/Sources/Services/Authentication/AuthenticationServiceProxy.swift
```diff
@@ -59,6 +61,45 @@ class AuthenticationServiceProxy: AuthenticationServiceProxyProtocol {
         }
     }
     
+    func loginWithOIDC() async -> Result<UserSessionProtocol, AuthenticationServiceError> {
+        guard
+            let oidcUserAgent = oidcUserAgent,
```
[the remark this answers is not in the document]
pixlwave: Derp, makes much more sense to me 👍

ElementX/Sources/Services/Authentication/OIDCService.swift
pixlwave: For this temporary implementation, it doesn't seem worth the effort. Especially as it will make the `loginWithOIDC` method grow in the auth service proxy.

ElementX/Sources/Services/Authentication/AuthenticationServiceProxy.swift
```diff
@@ -97,4 +138,13 @@ class AuthenticationServiceProxy: AuthenticationServiceProxyProtocol {
             return .failure(.failedLoggingIn)
         }
     }
+    
+    private func generateDeviceID() -> String {
```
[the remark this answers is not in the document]
pixlwave: This is all temporary, when rust has OIDC we should never need to generate a device ID.

stefanceriu (approved):

PR #146, opened 2022-07-27
Setup tracing through the RustSDK

stefanceriu:


ElementX/Sources/AppCoordinator.swift
```diff
@@ -97,6 +92,25 @@ class AppCoordinator: AuthenticationCoordinatorDelegate, Coordinator {
...
+        
+        #if DEBUG
+        setupTracing(configuration: "info,hyper=warn,sled=warn,matrix_sdk_sled=warn")
+        loggerConfiguration.logLevel = .debug
+        #else
+        setupTracing(configuration: "info,hyper=warn,sled=warn,matrix_sdk_sled=warn")
+        loggerConfiguration.logLevel = .info
+        #endif
+        
+        // Redirect NSLogs to files only if we are not debugging
```
pixlwave, line 106: Seeing as it took both of us a while to understand this:
```suggestion
        // Avoid redirecting NSLogs to files if we are attached to a debugger.
```

ElementX/Sources/AppCoordinator.swift
```diff
@@ -97,6 +92,25 @@ class AppCoordinator: AuthenticationCoordinatorDelegate, Coordinator {
     
     // MARK: - Private
     
+    private func setupLogging() {
+        let loggerConfiguration = MXLogConfiguration()
+        
+        #if DEBUG
+        setupTracing(configuration: "info,hyper=warn,sled=warn,matrix_sdk_sled=warn")
```
pixlwave, line 99: Might be nice to add a comment here saying you can change `warn` to `info` (or whatever it is) for more detailed Rust logging when needed during dev.

pixlwave (approved):
This is going to be amazing 🙌

PR #149, opened 2022-07-28
Add a sync limit of 20 and prefill rooms when loading the last message.

### LINKED ISSUE #93: Have more messages in cache
The room screen is empty when you open it and it stays empty for a long time.

We need to increase the `limit` value of the `/sync` request to have more messages in cache to improve the UX.

Element-iOS has an algo to determine this value according to the screen size. The idea is to have a full page of messages when you open a room. It is roughly 10 for iPhoneSE, 30 for iPads and 20 in between.
It depends a lot on the text density, which is quite light on ElementX right now.

As a quick start, we can use 20 so that we will be able to compare EX and EIX performance on a same amount of data

pixlwave:
This PR also filters rooms in the home screen to only show joined rooms as we were showing invited/left rooms too without differentiating them in anyway (and when trying to back-paginate in a left room the SDK would end up stuck in a loop that blocked other requests).

Updates the SDK to 1.0.13-alpha for the `timeline_limit` and `membership` options.

Fixes #93.

### DIFF
--- a/ElementX/Sources/Screens/HomeScreen/HomeScreenCoordinator.swift
+++ b/ElementX/Sources/Screens/HomeScreen/HomeScreenCoordinator.swift
@@ -119,7 +119,7 @@ final class HomeScreenCoordinator: Coordinator, Presentable {
     
     func updateRoomsList() {
         roomSummaries = parameters.userSession.clientProxy.rooms.compactMap { roomProxy in
-            guard !roomProxy.isSpace, !roomProxy.isTombstoned else {
+            guard roomProxy.isJoined, !roomProxy.isSpace, !roomProxy.isTombstoned else {
                 return nil
             }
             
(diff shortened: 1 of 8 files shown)


ElementX/Sources/Screens/HomeScreen/HomeScreenCoordinator.swift
```diff
@@ -119,7 +119,7 @@ final class HomeScreenCoordinator: Coordinator, Presentable {
     
     func updateRoomsList() {
         roomSummaries = parameters.userSession.clientProxy.rooms.compactMap { roomProxy in
-            guard !roomProxy.isSpace, !roomProxy.isTombstoned else {
+            guard roomProxy.isJoined, !roomProxy.isSpace, !roomProxy.isTombstoned else {
```
pixlwave, line 122: This seemed like the most sensible place to filter the rooms for now so that:
- The `clientProxy.rooms` array contains all rooms if we wanted to show others elsewhere.
- The filtering happens in the coordinator rather than the home screen's view state (which would be computed much more often)

Also I guess sync v3 will change most of this anyway.

ismailgulek (approved):
LGTM!

PR #153, opened 2022-08-01
Fix broken link in CONTRIBUTING.md

vollkorntomate:
Fixes the link to the `createScreen.sh` script.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,7 +32,7 @@ Please see our [pull request guide](https://github.com/vector-im/element-android
 
 ## Implementing a new screen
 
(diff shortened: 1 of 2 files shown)


pixlwave (approved):
LGTM, thanks @vollkorntomate 🙏

pixlwave (changes requested):
Hmmm, I didn't expect Danger to fail. Anyway, it should have said the following:

Could you please add a [changelog file](https://github.com/matrix-org/matrix-ios-sdk/blob/develop/CONTRIBUTING.md#changelog) in ./changelog.d and a [sign off](https://github.com/matrix-org/matrix-ios-sdk/blob/develop/CONTRIBUTING.md#sign-off) to the PR.

pixlwave (approved):
Thanks for the update 🙏

Issue #54, opened 2022-05-13
Research deep linking libraries


stefanceriu:
In order to prepare for opening permalinks and deep linking we should investigate available 3rd party libraries find one that fits our needs.

https://github.com/pointfreeco/swift-url-routing seems like a very good fit

stefanceriu:
Closing in favor of https://github.com/vector-im/element-x-ios/issues/33

PR #155, opened 2022-08-02
Draft: Define URL scheme and support deep links

vollkorntomate:
I looked at #54 and drafted a way to support deep links that I would like to discuss.

In my opinion, there is no real need for a 3rd party library, since iOS handles most of it and the current architecture (with `AppDelegate` and `Coordinator`s) is also beneficial for that.

For the basic URL scheme, I adapted the existing approach from element-ios, so that an example link (e.g. to join a room) could look like that: `elementx://io.element.elementx/room/%23element-ios:matrix.org`.
The `%23` encodes a `#` character, which has a different purpose in URLs (see [RFC 1808](https://datatracker.ietf.org/doc/html/rfc1808#section-2.1)). It will be decoded automatically by Swift.

Note that this is just a draft and I am open to discussing it. There are certainly things that need refinement.


stefanceriu:
Hi Thomas, thanks for raising this. While this is a very good start I do not agree with

> In my opinion, there is no real need for a 3rd party library, since iOS handles most of it and the current architecture (with `AppDelegate` and `Coordinator`s) is also beneficial for that.

We need something to help us parse and decompose custom schemes, universal links and matrix.to links into something, preferably type safe, that the app can converge on and use, something like https://github.com/pointfreeco/swift-url-routing. I would prefer not to write on own. That was the idea behind #54

PR #157, opened 2022-08-03
Small tweaks for CI

pixlwave:
This PR makes 2 small tweaks for CI:
- Danger is now disabled for external repos as it fails due to the missing GH secret.
- SwiftFormat is run as a pre-build script to fail early and free up runners quicker.


ismailgulek (approved):
LGTM!

PR #159, opened 2022-08-09
Use standard file headers everywhere.

### LINKED ISSUE #150: Update file header copyright
We need to make all project file use the right copyright i.e. "New Vector" instead of "Element". We also need to update the project settings so that it's changed for newly created files as well.

More details here https://gitlab.matrix.org/new-vector/internal/-/wikis/Copyright-headers

pixlwave:
This PR contains 2 commits to align the headers in ElementX with those in Element iOS:
- Add a `IDETemplateMacros.plist` file so that new files always have the default file header added.
- A regex replacement of Xcode headers → the default file header (and a find and replace for 2021 → 2022 on existing files).

Fixes #150.

Edit: Last commit fixes an issue where the first line has an extra ` ` in the header.

### DIFF
--- a/DesignKit/Sources/Buttons/ElementActionButtonStyle.swift
+++ b/DesignKit/Sources/Buttons/ElementActionButtonStyle.swift
@@ -1,5 +1,5 @@
 //
-// Copyright 2021 New Vector Ltd
+// Copyright 2022 New Vector Ltd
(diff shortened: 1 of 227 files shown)


ismailgulek (approved):
Looking great!

PR #160, opened 2022-08-12
Use a more robust detection for forks.

pixlwave:
The previous check failed to work as the repo name doesn't change between PRs 🤦‍♂️

This one should check the remote name against the local name.


ismailgulek (approved):
LGTM!

PR #164, opened 2022-08-25
Various logging fixes

stefanceriu:
* imported the ObjCWrapper and updated the MXLogger. Also added back support for app containers
* updated MXLog.swift from El and fixed/improved its logging format
* updated most app log strings

### DIFF
--- a/ElementX/Sources/Other/Logging/MXLog.swift
+++ b/ElementX/Sources/Other/Logging/MXLog.swift
@@ -118,32 +118,57 @@ private var logger: SwiftyBeaver.Type = {
         logger.warning(message, file, function, line: line)
     }
     
-    public static func error(_ message: @autoclosure () -> Any,
+    /// Log error with additional details
+    ///
+    /// - Parameters:
+    ///     - message: Description of the error without any variables (this is to improve error aggregations by type)
+    ///     - context: Additional context-dependent details about the issue
+    public static func error(_ message: String,
                              _ file: String = #file,
                              _ function: String = #function,
                              line: Int = #line,
                              context: Any? = nil) {
-        logger.error(message(), file, function, line: line, context: context)
+        logger.error(message, file, function, line: line, context: context)
     }
     
     @available(swift, obsoleted: 5.4)
-    @objc public static func logError(_ message: String, file: String, function: String, line: Int) {
-        logger.error(message, file, function, line: line)
+    @objc public static func logError(_ message: String,
+                                      file: String,
+                                      function: String,
+                                      line: Int,
+                                      context: Any? = nil) {
+        logger.error(message, file, function, line: line, context: context)
     }
     
-    public static func failure(_ message: @autoclosure () -> Any,
+    /// Log failure with additional details
+    ///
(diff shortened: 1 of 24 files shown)


ElementX/Sources/Other/Logging/MXLog.swift
```diff
@@ -118,32 +118,57 @@ private var logger: SwiftyBeaver.Type = {
...
+    ///
+    /// A failure is any type of programming error which should never occur in production. In `DEBUG` configuration
+    /// any failure will raise `assertionFailure`
+    ///
+    /// - Parameters:
+    ///     - message: Description of the error without any variables (this is to improve error aggregations by type)
+    ///     - context: Additional context-dependent details about the issue
+    public static func failure(_ message: String,
                                _ file: String = #file,
                                _ function: String = #function,
```
pixlwave, line 153: Not related to this PR, but seeing as we basically never enter these, maybe we should add the parameter names at some stage so its clear which default value you are replacing.
```suggestion
                               file: String = #file,
                               function: String = #function,
```

ElementX/Sources/Other/Logging/MXLogger.m
```diff
@@ -142,7 +143,7 @@ + (void)deleteLogFiles
     }
 }
 
-+ (NSArray<NSString*>*)logFiles
++ (NSArray*)logFiles
```
pixlwave, line 146: How come the array type is removed here?

ElementX/Sources/Services/BugReport/BugReportService.swift
```diff
@@ -106,6 +108,11 @@ class BugReportService: BugReportServiceProtocol {
                 params.append(MultipartFormData(key: "compressed-log", type: .file(url: url)))
             }
         }
+        
+        if let crashEventId = lastCrashEventId {
+            params.append(MultipartFormData(key: "crash_report", type: .text(value: "<https://sentry.tools.element.io/organizations/element/issues/?project=44&query=\(crashEventId)>")))
```
pixlwave, line 113: This is clever 🧐

pixlwave (approved):
LGTM, although I'm not sure I understand why we want to add more objc to the project 🫣

ElementX/Sources/Other/Logging/MXLog.swift
```diff
@@ -152,24 +177,27 @@ private var logger: SwiftyBeaver.Type = {
...
             return
         }
         
+        logger.removeAllDestinations()
+        
         let consoleDestination = ConsoleDestination()
         consoleDestination.useNSLog = true
         consoleDestination.asynchronously = false
-        consoleDestination.format = "$M"
+        consoleDestination.format = "$C$N.$F:$l $M $X$c" // Format is `Color Message Context`, see https://docs.swiftybeaver.com/article/20-custom-format
```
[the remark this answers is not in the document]
stefanceriu: Yup, definitely 👍

ElementX/Sources/Other/Logging/MXLogger.m
stefanceriu: Seems that Ismail added that when migrating it over from Element. It got overriden when I copied everything over again

PR #168, opened 2022-08-30
Begin adding the Analytics class from EI.

pixlwave:
This PR focuses on `Analytics.swift` and surrounding infrastructure. There is no UI or SDK support to enable analytics yet (and no tracking is implemented) to help keep PRs at a manageable size.

This forms the first part of #106.

**Note:** There are still some warnings and unanswered questions in this initial commit that need review.

### DIFF
--- a/ElementX/Sources/Services/Analytics/Analytics.swift
+++ b/ElementX/Sources/Services/Analytics/Analytics.swift
@@ -109,7 +109,7 @@ class Analytics {
         
         switch await service.settings() {
         case .success(let settings):
-            self.identify(with: settings)
+            identify(with: settings)
             self.service = nil
         case .failure:
             MXLog.error("Failed to use analytics settings. Will continue to run without analytics ID.")
@@ -165,10 +165,9 @@ class Analytics {
 
 // The following methods are exposed for compatibility with Objective-C as
 // the `capture` method and the generated events cannot be bridged from Swift.
-extension Analytics {
-}
+extension Analytics { }
 
 // MARK: - MXAnalyticsDelegate
 
-//extension Analytics: MXAnalyticsDelegate {
-//}
+// extension Analytics: MXAnalyticsDelegate {
+// }
--- a/ElementX/Sources/Services/Analytics/Analytics.swift
+++ b/ElementX/Sources/Services/Analytics/Analytics.swift
@@ -1,0 +1,174 @@
+//
+// Copyright 2021 New Vector Ltd
+//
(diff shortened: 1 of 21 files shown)


ElementX/Sources/Services/Analytics/Analytics.swift
```diff
@@ -0,0 +1,173 @@
...
+///
+/// Events are managed in a shared repo for all Element clients https://github.com/matrix-org/matrix-analytics-events
+/// To add a new event create a PR to that repo with the new/updated schema. Once merged
+/// into `main`, update the AnalyticsEvents Swift package in `project.yml`.
+///
+class Analytics {
+    // MARK: - Properties
+    
+    /// The singleton instance to be used within the Riot target.
+    static let shared = Analytics()
```
pixlwave, line 37: There's a conversation to have around whether this should be made into a property on `UserSession` instead of being a singleton, but should probably wait until we're closer to implementing tracking so as to understand how it would need to be available from different parts of the app.

ElementX/Sources/Services/Analytics/Analytics.swift
```diff
@@ -0,0 +1,173 @@
...
+/// non-fatal issues and performance. `Analytics` class serves as a façade
+/// to all these use cases.
+///
+/// ## Creating Analytics Events
+///
+/// Events are managed in a shared repo for all Element clients https://github.com/matrix-org/matrix-analytics-events
+/// To add a new event create a PR to that repo with the new/updated schema. Once merged
+/// into `main`, update the AnalyticsEvents Swift package in `project.yml`.
+///
+class Analytics {
```
pixlwave, line 33: Also, some future investigation needs to be done into which parts of this PR should be put on the main actor.

UnitTests/SupportingFiles/target.yml
```diff
@@ -49,5 +49,3 @@ targets:
     - path: ../SupportingFiles
     - path: ../../Tools/Scripts/Templates/SimpleScreenExample/Tests/Unit
     - path: ../Resources
-    - path: ../../ElementX/Sources/BuildSettings.swift
-    - path: ../../ElementX/Sources/Screens/RoomScreen/View/Style/TimelineStyle.swift
```
pixlwave, line 53: Not sure what these were added for but I got compiler errors in BuildSettings and everything seems to be working fine without them added to the target.

stefanceriu (approved):
lgtm! 👍

PR #170, opened 2022-09-05
Update fix snapshots of Settings screen.

pixlwave:
Regenerate snapshots of the Settings screen.


pixlwave:
I think something has gone wrong with this. The snapshots don’t appear to have been picked up by LFS (also looks to be the same for the last commit on `develop` with the analytics ones too 😞

Will have to take another look in the morning.

PR #173, opened 2022-09-06
Add docs around Git LFS and include it in the Brewfile.

pixlwave:
Not sure if CI is going to 💥 trying to install LFS through brew, lets see 🤞

Edit: seems fine says “Using git-lfs” in the logs.


ismailgulek (approved):
LGTM!

Issue #178, opened 2022-09-13
Edit own message


stefanceriu:
Allow the users to redact their own messages
Designs: https://www.figma.com/file/urDzUwqbnLNuXmqg7hYB83/Element-X-%5BEX1%5D?node-id=1596%3A62224

PR #182, opened 2022-09-14
#178: Add redaction menu (disabled).

pixlwave:
Adds a redact item to the context menu.

Also adds icons to the menu as shown in the Figma (although some of the items aren't in there so I've chosen what makes sense to me).

| Light | Dark |
| - | - |
| Simulator Screen Shot - iPhone 13 - 2022-09-14 at 09 49 03 | Simulator Screen Shot - iPhone 13 - 2022-09-14 at 09 49 07 |

### DIFF
--- a/ElementX/Sources/Services/Room/RoomProxy.swift
+++ b/ElementX/Sources/Services/Room/RoomProxy.swift
@@ -194,7 +194,7 @@ class RoomProxy: RoomProxyProtocol {
         .value
     }
     
-    func redactItem(_ eventID: String) async -> Result<Void, RoomProxyError> {
+    func redact(_ eventID: String) async -> Result<Void, RoomProxyError> {
         #warning("Redactions to be enabled on next SDK release.")
         return .failure(.failedRedactingEvent)
 //        let transactionID = genTransactionId()
--- a/ElementX/Sources/Services/Room/RoomProxy.swift
+++ b/ElementX/Sources/Services/Room/RoomProxy.swift
@@ -194,14 +194,14 @@ class RoomProxy: RoomProxyProtocol {
         .value
     }
     
-    func redactItem(_ itemId: String) async -> Result<Void, RoomProxyError> {
+    func redactItem(_ eventID: String) async -> Result<Void, RoomProxyError> {
         #warning("Redactions to be enabled on next SDK release.")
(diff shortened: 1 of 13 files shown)


ElementX/Sources/Services/Room/RoomProxy.swift
```diff
@@ -194,6 +194,22 @@ class RoomProxy: RoomProxyProtocol {
...
     }
     
+    func redactItem(_ itemId: String) async -> Result<Void, RoomProxyError> {
+        return .failure(.failedRedactingEvent)
+        #warning("Redactions to be enabled on next SDK release.")
+//        let transactionID = genTransactionId()
+//
+//        return await Task {
+//            do {
+//                try room.redact(eventId: itemId, reason: nil, txnId: transactionID)
```
pixlwave, line 204: Wasn't sure whether to name these EX methods `redactItem` or `redactEvent` given this weird `eventId: ItemId` but went with item for consistency with others.

Sidenote: I'm dying writing `itemId` instead of `itemID` everywhere for consistency 😭

ElementX/Sources/Services/Room/RoomProxyProtocol.swift
```diff
@@ -24,6 +24,7 @@ enum RoomProxyError: Error {
     case failedRetrievingMemberAvatarURL
     case failedRetrievingMemberDisplayName
     case failedSendingMessage
+    case failedRedactingEvent
```
pixlwave, line 27: This should probably contain the event ID as an associated value, but we don't seem to handle these failures yet, so I didn't add it (and also, if we get the higher-level API from rust it might not even need to fail).

project.yml
```diff
@@ -32,9 +32,9 @@ include:
 
 packages:
   MatrixRustSDK:
-    url: https://github.com/matrix-org/matrix-rust-components-swift
-    exactVersion: 1.0.13-alpha
-    # path: ../matrix-rust-components-swift
+    # url: https://github.com/matrix-org/matrix-rust-components-swift.git
+    # exactVersion: 1.0.11-alpha
+    path: ../matrix-rust-components-swift
```
pixlwave, line 37: Whoops!

stefanceriu (approved):

PR #186, opened 2022-09-14
DesignKit: Update design tokens and create ElementColors locally.

pixlwave:
The latest [version](https://github.com/vector-im/element-design-tokens/commit/63e40f10b336c136d6d05f7967e4565e37d3d760) of element-design-tokens removes the `.element` static property on `Color`/`UIColor` and renames the exported colours as `CompoundColors`. The `.global` property is still available.

This PR updates the version of the tokens consumed, and creates `ElementColors` locally in DesignKit to add the system colours that are in the Figma.

This is likely a temporary set up until colours and Compound in general are better defined.

### DIFF
--- a/DesignKit/Sources/Colors/ElementColors.swift
+++ b/DesignKit/Sources/Colors/ElementColors.swift
@@ -26,20 +26,30 @@ public extension Color {
 public struct ElementColors {
     // MARK: - Compound
     
-    public var accent: Color { .primary }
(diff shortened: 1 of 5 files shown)


stefanceriu (approved):
lgtm 👍

pixlwave:
Surprisingly, with the change of accent colour, the first few UI tests I ran locally all passed. Will merge this and see what happens in the morning 🔍

PR #187, opened 2022-09-14
DesignKit: Update fonts to match Figma.

pixlwave:
Following on from #186, this PR updates the fonts in DesignKit to match the Figma (all semi-bold's are now bolds).

It seemed to make more sense to me to simplify the base fonts and apply the weight modifiers during the SwiftUI/UIKit font structs creation, but happy to revert if that doesn't seem sensible. Also, yes I chickened out of removing the `bold` variants - mainly because having these made the refactor really easy which I think is reason enough to keep them :)


pixlwave:
Whoops, I was wrong about #186 not failing snapshots... I tested them with the wrong device. All updated in this PR.

ismailgulek (approved):
LGTM!

PR #189, opened 2022-09-15
Sliding Sync + New Timeline API

stefanceriu:
This PR is made from a special demo branch in which we merged everything currently ongoing:
* sliding sync
* new timeline
* logout flows and soft logout
* replies, redactions and permalinks
* new home screen avatar menu
* new color scheme

and more..


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -37,46 +41,75 @@ private class WeakClientProxyWrapper: ClientDelegate {
...
+            
+            Benchmark.startTrackingForIdentifier("ClientSync", message: "Started sync.")
+            
+            slidingSync.setObserver(observer: WeakClientProxyWrapper(clientProxy: self))
+            slidingSyncObserverToken = slidingSync.sync()
+        } catch {
+            fatalError("Failed configuring sliding sync")
+        }
+        
+        if !client.isSoftLogout() {
```
ismailgulek, line 108: Probably we should move the sliding sync block inside this if and remove old `startSync` here.

ismailgulek (approved):

ElementX/Sources/Services/Room/RoomProxy.swift
```diff
@@ -210,4 +227,20 @@ class RoomProxy: RoomProxyProtocol {
...
+        memberDisplayNames[userId] = displayName
+    }
+    
+    func update(displayName: String) {
+        self.displayName = displayName
+    }
+    
+    func update(backPaginationOutcome: PaginationOutcome) {
+        self.backPaginationOutcome = backPaginationOutcome
+    }
```
pixlwave, line 245: The 2 are async and the second 2 aren't? Maybe intentional, not sure.

pixlwave: In fact, the first 2 don't need to be async given they never `await` anything anyway.

ElementX/Sources/Services/Room/RoomProxy.swift
```diff
@@ -104,9 +104,13 @@ class RoomProxy: RoomProxyProtocol {
     
     func loadAvatarURLForUserId(_ userId: String) async -> Result<String?, RoomProxyError> {
         await Task.detached { [weak self] () -> Result<String?, RoomProxyError> in
+            guard let self = self else {
+                return .failure(.failedRetrievingMemberAvatarURL)
+            }
```
pixlwave, line 109: If the `@MainActor` fixed the issue, do we actually need the weak self now?

ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu: We actually need both for the moment as some parts of the app still rely on the v2 sync but I'll guard on `client.isSoftLogout` before both

ElementX/Sources/Services/Room/RoomProxy.swift
stefanceriu: Initially they needed to be async so that we can await them from the detached task, but now that the whole class is a MainActor they're awaitable by default. I'll remove `await` from all of them

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -0,0 +1,212 @@
...
+            return buildEmptyRoomSummary()
+        case .filled(let roomId):
+            return buildRoomSummaryForIdentifier(roomId)
+        case .invalidated(let roomId):
+            return buildRoomSummaryForIdentifier(roomId)
+        }
+    }
+    
+    // swiftlint:disable cyclomatic_complexity
+    private func buildDiffFrom(_ diff: SlidingSyncViewRoomsListDiff) -> CollectionDifference<RoomSummary>? {
```
pixlwave, line 163: `buildDiff(from diff: SlidingSyncViewRoomsListDiff)` pls 🥺

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -0,0 +1,212 @@
...
+        case .push(let value), .updateAt(_, let value), .insertAt(_, let value):
+            switch value {
+            case .invalidated:
+                return nil
+            default:
+                break
+            }
+        default:
+            break
+        }
```
pixlwave, line 174: Would be nice to make this a property on `SlidingSyncViewRoomsListDiff` in an extension.

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -0,0 +1,212 @@
...
+            for (index, summary) in roomSummaries.enumerated() {
+                changes.append(.remove(offset: index, element: summary, associatedWith: nil))
+            }
+            
+            values
+                .reversed()
+                .map { buildSummaryForRoomListEntry($0) }
+                .forEach { summary in
+                    changes.append(.insert(offset: 0, element: summary, associatedWith: nil))
+                }
```
pixlwave, line 206: Maybe we should detect the `.replace` case in `updateRoomsWithDiffs` and directly map & assign the room summaries array there, avoiding this extra work?

pixlwave (commented):
Comments inline on the main SS commit :)

ElementX/Sources/Services/Room/RoomProxy.swift
stefanceriu: Okay so after [further reading](https://stackoverflow.com/a/71729233/730924) as long as we don't hold on locally to the task itself retain cycles shouldn't be a problem.

> The rules are exactly the same with Swift concurrency and Task:
> 
> The closure you pass into a Task to initialize it with retains the objects it references by default (unless you use [weak ...])
> Task holds on to the closure for the duration of the task (i.e., while it's executing)
> You will have a retain cycle if self holds on to the Task for the duration of the execution

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
```diff
@@ -44,24 +44,16 @@ struct HomeScreenViewState: BindableState {
...
-    var isLoadingRooms = false
-    
-    var visibleDMs: [HomeScreenRoom] {
-        searchFilteredRooms.filter(\.isDirect)
-    }
-
-    var visibleRooms: [HomeScreenRoom] {
-        searchFilteredRooms.filter { !$0.isDirect }
+    var isLoadingRooms: Bool {
+        rooms.isEmpty
```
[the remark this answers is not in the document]
stefanceriu: Yep, that's true. There's a `SlidingSyncViewRoomsCountObserver` but with how fast things generally are we might be able to get away with not even having a loading state.

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -60,114 +84,36 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
...
-                                  topic: roomSummary.topic,
-                                  lastMessage: await lastMessageFromEventBrief(roomSummary.lastMessage),
-                                  avatar: roomSummary.avatar,
-                                  isDirect: roomSummary.isDirect,
-                                  isEncrypted: roomSummary.isEncrypted,
-                                  isSpace: roomSummary.isSpace,
-                                  isTombstoned: roomSummary.isTombstoned)
+        if let avatarURLString = summary.avatarURLString {
+            Task {
+                let _ = await userSession.mediaProvider.loadImageFromURLString(avatarURLString)
```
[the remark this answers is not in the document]
stefanceriu: 👍

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -64,19 +80,38 @@ class ClientProxy: ClientProxyProtocol {
...
+                .addView(view: slidingSyncView)
+                .build()
+            
+            roomSummaryProvider = RoomSummaryProvider(slidingSyncController: slidingSync,
+                                                      slidingSyncView: slidingSyncView)
+            
+            Benchmark.startTrackingForIdentifier("ClientSync", message: "Started sync.")
+            
+            slidingSync.setObserver(observer: WeakClientProxyWrapper(clientProxy: self))
+            slidingSyncObserverToken = slidingSync.sync()
```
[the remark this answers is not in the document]
stefanceriu: Yep yep, very good idea 👍

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -0,0 +1,212 @@
...
+    
+    func didReceiveUpdate(newState: SlidingSyncState) {
+        stateUpdatePublisher.send(newState)
+    }
+}
+
+class RoomSummaryProvider: RoomSummaryProviderProtocol {
+    private let slidingSyncController: SlidingSyncProtocol
+    private let slidingSyncView: SlidingSyncViewProtocol
+    private let roomMessageFactory = RoomMessageFactory()
```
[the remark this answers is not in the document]
stefanceriu: Yes, my bad 👍

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
stefanceriu: Talked about it offline, will not change it for now.

stefanceriu:
GitHub is currently having issues so we had to build the latest manually. Please use the following QR code or link:

QR code  

If you can't scan the QR code you can install the build via this link: https://i.diawi.com/Pr3Kwc

Hywan:
🎉

PR #196, opened 2022-09-20
Settings Screen New Design

### LINKED ISSUE #180: SettingsScreen - New design
Apply the new designs to the existing settings screen - https://www.figma.com/file/urDzUwqbnLNuXmqg7hYB83/Element-X-%5BEX1%5D?node-id=1543%3A62041

ismailgulek:
Fixes #180 

`Appearance` section not implemented. Also updates placeholder user and room avatars.

|Light|Dark|
|-|-|
|Simulator Screen Shot - iPhone 13 Pro Max - 2022-09-20 at 18 55 56|Simulator Screen Shot - iPhone 13 Pro Max - 2022-09-20 at 18 56 12|
|Simulator Screen Shot - iPhone 13 Pro Max - 2022-09-20 at 19 12 32|Simulator Screen Shot - iPhone 13 Pro Max - 2022-09-20 at 19 12 47|


ElementX/Sources/Screens/Settings/SettingsCoordinator.swift
```diff
@@ -106,12 +111,25 @@ final class SettingsCoordinator: Coordinator, Presentable {
...
         coordinator.start()
-        parameters.navigationRouter.push(coordinator, animated: true) { [weak self] in
+        navigationRouter.push(coordinator, animated: true) { [weak self] in
             guard let self = self else { return }
 
             self.remove(childCoordinator: coordinator)
         }
     }
+
+    private func confirmSignOut() {
```
pixlwave, line 121: Future improvement: We could probably show this with a `.confirmationDialog` in SwiftUI.

ElementX/Sources/Screens/Settings/View/SettingsScreen.swift
```diff
@@ -79,36 +154,66 @@ struct SettingsScreen: View {
...
+            .listRowInsets(listRowInsets)
             .foregroundColor(.element.primaryContent)
             .accessibilityIdentifier("logoutButton")
-            .confirmationDialog(ElementL10n.actionSignOutConfirmationSimple,
-                                isPresented: $showingLogoutConfirmation,
-                                titleVisibility: .visible) {
-                Button(ElementL10n.actionSignOut,
-                       role: .destructive) { context.send(viewAction: .logout)
-                }
-            }
```
pixlwave, line 108: Oh hmmm, talking of the last comment, we did do that. How come you switched to UIKit for this?

ElementX/Sources/Screens/Settings/View/SettingsScreen.swift
```diff
@@ -79,36 +154,66 @@ struct SettingsScreen: View {
...
         } footer: {
             versionText
+                .frame(maxWidth: .infinity)
+        }
+    }
+
+    private var closeButton: some View {
+        Button(action: close) {
+            HStack {
```
pixlwave, line 197: The `HStack` seems a bit redundant here with just an `Image` inside it

ElementX/Sources/AppCoordinatorStateMachine.swift
```diff
@@ -77,11 +74,6 @@ class AppCoordinatorStateMachine {
         case showSessionVerificationScreen
         /// Session verification has finished
         case dismissedSessionVerificationScreen
-        
-        /// Request settings screen presentation
-        case showSettingsScreen
-        /// The settings screen has been dismissed
-        case dismissedSettingsScreen
```
pixlwave, line 84: How come you modified the state machine here? Presumably to inject the user session?

I think we need to separate out into 2 of these now, one top level for the whole app and another for navigation with a single user session navigation.

pixlwave (approved):
Looks good to me, a couple of thoughts inline, but nothing to stop merging this now.

ElementX/Sources/Screens/Settings/View/SettingsScreen.swift
ismailgulek: It was silly, added back with `alert`

ElementX/Sources/Screens/Settings/SettingsViewModel.swift
```diff
@@ -25,21 +25,38 @@ class SettingsViewModel: SettingsViewModelType, SettingsViewModelProtocol {
...
+        Task {
+            if case let .success(userAvatarURLString) = await userSession.clientProxy.loadUserAvatarURLString() {
+                if case let .success(avatar) = await userSession.mediaProvider.loadImageFromURLString(userAvatarURLString, size: MediaProviderDefaultAvatarSize) {
+                    state.userAvatar = avatar
+                }
+            }
+
+            if case let .success(userDisplayName) = await self.userSession.clientProxy.loadUserDisplayName() {
+                state.userDisplayName = userDisplayName
+            }
```
pixlwave, line 50: Might be worth putting this in its own `Task` so that if awaiting the image above is slow, this could still happen in its own time?

ElementX/Sources/Screens/RoomScreen/View/Timeline/PlaceholderAvatarImage.swift
```diff
@@ -18,22 +18,32 @@ import SwiftUI
...
         ZStack {
-            Color.element.accent
+            bgColor
             Text(textForImage)
                 .padding(4)
                 .foregroundColor(.element.background)
-                // Make the text resizable (i.e. Make it large and then allow it to scale down)
-                .font(.system(size: 200).weight(.semibold))
-                .minimumScaleFactor(0.001)
+                .font(.title2.bold())
```
pixlwave, line 29: I'm not 100% sure this is the right choice here. E.g. looking at avatars in the timeline, they're now bottom right of centre. I think its worth deviating from the design for that.

ElementX/Sources/Screens/RoomScreen/View/Timeline/PlaceholderAvatarImage.swift
```diff
@@ -18,22 +18,32 @@ import SwiftUI
...
+    }
+
+    private var bgColor: Color {
+        guard let contentId = contentId else {
+            return .element.accent
+        }
+
+        let colors = Color.element.contentAndAvatars
+        let colorIndex = Int(contentId.hashCode % Int32(colors.count))
+        return Color.element.contentAndAvatars[colorIndex % colors.count]
```
pixlwave, line 46: I reckon this should be an extension on `Element[UI]Colors` so its easy to call something like this anywhere:
```swift
Color.element.avatarBackground(for: contentID)
```
Sidenote: On ThemeV1 in EI we actually have avatar colours separate from display name colours and there are less of them but ThemeV2/SwiftUI actually breaks this. We should try to avoid that again (and fix it at some stage on EI).

pixlwave (commented):
3 small comments in the re-review :)

PR #197, opened 2022-09-21
Made user session store base directory lazy, logging it for easier de…

stefanceriu:
…bugging

### DIFF
--- a/ElementX/Sources/Services/Authentication/AuthenticationServiceProxy.swift
+++ b/ElementX/Sources/Services/Authentication/AuthenticationServiceProxy.swift
@@ -34,7 +34,7 @@ class AuthenticationServiceProxy: AuthenticationServiceProxyProtocol {
     
     init(userSessionStore: UserSessionStoreProtocol) {
         self.userSessionStore = userSessionStore
-        authenticationService = AuthenticationService(basePath: userSessionStore.baseDirectoryPath)
+        authenticationService = AuthenticationService(basePath: userSessionStore.baseDirectory.path)
     }
     
     // MARK: - Public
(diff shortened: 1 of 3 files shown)


ElementX/Sources/Services/UserSessionStore/UserSessionStore.swift
```diff
@@ -136,7 +136,7 @@ class UserSessionStore: UserSessionStoreProtocol {
         }
     }
     
-    func baseDirectory() -> URL {
+    lazy var baseDirectory: URL = {
```
pixlwave, line 139: Probably worth making this a `private(set)`

pixlwave (approved):

ElementX/Sources/Services/UserSessionStore/UserSessionStore.swift
stefanceriu: Upsie 🙃 
I wonder if we should drop the `baseDirectoryPath` and just keep this one?

ElementX/Sources/Services/Authentication/AuthenticationServiceProxy.swift
```diff
@@ -34,7 +34,7 @@ class AuthenticationServiceProxy: AuthenticationServiceProxyProtocol {
     
     init(userSessionStore: UserSessionStoreProtocol) {
         self.userSessionStore = userSessionStore
-        authenticationService = AuthenticationService(basePath: userSessionStore.baseDirectoryPath)
+        authenticationService = AuthenticationService(basePath: userSessionStore.baseDirectory.path)
```
pixlwave, line 37: Lol not sure why I didn't just do this 🙈

pixlwave (approved):
LGTM

PR #200, opened 2022-09-21
Some Post-demo Cleanups

ismailgulek:
Can be reviewed commit by commit. Fixes some remarks from #190 and #196


ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -65,6 +65,10 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
                     state.userAvatar = avatar
                 }
             }
+
+            if case let .success(userDisplayName) = await userSession.clientProxy.loadUserDisplayName() {
+                state.userDisplayName = userDisplayName
+            }
```
pixlwave, line 71: I *think* I made this comment in a different place on one of the other PRs, otherwise, sorry for repeating myself and ignore me!

This could probably go into its own `Task` so that the display name loading isn't blocked the loading above is slow.

UITests/Sources/HomeScreenUITests.swift
```diff
@@ -15,3 +15,14 @@
 //
 
 import XCTest
+
+class HomeScreenUITests: XCTestCase {
+    func testInitialStateComponents() {
+        let app = Application.launch()
+        app.goToScreenWithIdentifier(.home)
+
+        XCTAssert(app.navigationBars[ElementL10n.allChats].exists)
```
pixlwave, line 24: Minor: I'd say the snapshot should catch this so its probably not worth while even checking.

ElementX/Sources/Screens/Settings/View/SettingsScreen.swift
```diff
@@ -186,6 +187,14 @@ struct SettingsScreen: View {
             .listRowInsets(listRowInsets)
             .foregroundColor(.element.primaryContent)
             .accessibilityIdentifier("logoutButton")
+            .alert(ElementL10n.actionSignOut,
```
pixlwave, line 190: 🙌

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -180,6 +180,10 @@ class ClientProxy: ClientProxyProtocol {
             return nil
         }
     }
+
+    func clearRoom(for identifier: String) {
+        roomProxies.removeValue(forKey: identifier)
```
pixlwave, line 185: To go alongside this, I think we could add a `room.removeTimeline()` on the room proxy [deinit](https://github.com/vector-im/element-x-ios/blob/b9f8fb0b6f3fa97ba61f37dcca87d68a68b1b1f7/ElementX/Sources/Services/Room/RoomProxy.swift#L45).

Also, I think we can probably get rid of that `roomProxies` dictionary entirely now tbf. It was a holdover from pre-syncv3

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
```diff
@@ -41,6 +41,9 @@ struct RoomScreen: View {
             }
         }
         .alert(item: $context.alertInfo) { $0.alert }
+        .onDisappear {
+            context.send(viewAction: .viewDisappeared)
+        }
```
pixlwave, line 46: Not an issue now, but when we add another screen that's pushed on top of this one, we'll cancel all the cancellables when the screen is still in the hierarchy. Might need to have the coordinator tell the view model to do this when it knows its finished instead?

ElementX/Sources/Services/Media/MediaProviderProtocol.swift
```diff
@@ -22,17 +22,15 @@ enum MediaProviderError: Error {
     case invalidImageData
 }
 
-let MediaProviderDefaultAvatarSize = CGSize(width: 44.0, height: 44.0)
-
 @MainActor
 protocol MediaProviderProtocol {
-    func imageFromSource(_ source: MediaSource?, size: CGSize?) -> UIImage?
+    func imageFromSource(_ source: MediaSource?, size: AvatarSize?) -> UIImage?
```
pixlwave, line 27: I'm wondering if using `AvatarSize` is one level too far in the media provider? E.g. what if we wanted to get thumbnails for use in the timeline at some point? However, we could also add overloads here that accepted say a `TimelineMediaSize` so I can equally be convinced that this is fine.

ElementX/Sources/Services/Timeline/RoomTimelineController.swift
```diff
@@ -252,7 +252,7 @@ class RoomTimelineController: RoomTimelineControllerProtocol {
                 return
             }
             
-            switch await mediaProvider.loadImageFromURLString(avatarURLString, size: MediaProviderDefaultAvatarSize) {
+            switch await mediaProvider.loadImageFromURLString(avatarURLString, size: .user(on: .timeline)) {
```
pixlwave, line 254: That said, the parameter name should probably be `avatarSize` as when calling like this, the context of what `.user` is, is a little bit lost.

ElementX/Sources/Screens/RoomScreen/View/Timeline/FormattedBodyText.swift
```diff
@@ -26,25 +26,13 @@ struct FormattedBodyText: View {
...
-                            .fixedSize(horizontal: false, vertical: true)
-                            .foregroundColor(.element.primaryContent)
-                            .padding(EdgeInsets(top: 8, leading: 12, bottom: 8, trailing: 12))
-                            .clipped()
-                            .overlay(
-                                RoundedRectangle(cornerRadius: 13)
-                                    .stroke(Color.element.systemGray5)
-                            )
-                    }
+                    Text(component.attributedString.mergingAttributes(blockquoteAttributes))
```
pixlwave, line 29: Woah, that was simple to resize! 👏

ElementX/Sources/Services/Timeline/TimelineItems/EventBasedTimelineItemProtocol.swift
```diff
@@ -59,3 +59,9 @@ protocol EventBasedTimelineItemProtocol: RoomTimelineItemProtocol {
     
     var properties: RoomTimelineItemProperties { get }
 }
+
+extension EventBasedTimelineItemProtocol {
+    var shouldShowSenderDetails: Bool {
```
pixlwave, line 64: Smart 😎

pixlwave (approved):
LGTM, minor comments inline.

Also, the implementation of `AvatarSize` is 🤩

ElementX/Sources/Services/Media/MediaProviderProtocol.swift
ismailgulek: We could also use the `custom` one.

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
ismailgulek: 👍

UITests/Sources/HomeScreenUITests.swift
ismailgulek: Oh yeah, actually i've tried to write more tests about the room list but couldn't find a way yet.

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
ismailgulek: Will refactor in a separate PR.

PR #201, opened 2022-09-21
Replace detached tasks with `Task.dispatch(on:)`

pixlwave:
This should stop us from ever overwhelming the workers with too many detached tasks that don't suspend.

All `Task.detached` that didn't have an `await` and so were presumably blocking, have been replaced with continuations that dispatch onto the specified queue. A future tweak could be to make separate queues or to provide a QoS to the global queue.


ElementX/Sources/Other/Extensions/DispatchQueue.swift
```diff
@@ -0,0 +1,59 @@
...
+    /// a continuation to make it non-blocking and awaitable.
+    ///
+    /// - Parameters:
+    ///   - queue: The queue to run the closure on.
+    ///   - function: A string identifying the declaration that is the notional
+    ///     source for the continuation, used to identify the continuation in
+    ///     runtime diagnostics related to misuse of this continuation.
+    ///   - body: A sendable closure. Use of sendable won't work as it isn't
+    ///     async, but is added to enforce actor semantics.
+    static func awaitable<T>(on queue: DispatchQueue, function: String = #function, _ body: @escaping @Sendable () -> T) async -> T {
```
ismailgulek, line 30: I wonder if we should still keep this in `Task` extension, so we could use it like `Task.awaitable` (which seems a bit better than `DispatchQueue.awaitable` to me) if it makes sense.

ismailgulek (approved):
Just a tiny question inline, otherwise LGTM!

PR #203, opened 2022-09-21
Add validate-lfs.sh check from Android.

pixlwave:
Copied across from the EA [workflow](https://github.com/vector-im/element-android/blob/96901c5afe27fdc972d2733bf4db9f779617ae33/.github/workflows/validate-lfs.yml) and [script](https://github.com/vector-im/element-android/blob/96901c5afe27fdc972d2733bf4db9f779617ae33/tools/validate_lfs.sh).

Will test separately once merged with a fresh clone that didn't `git lfs install`

### DIFF
--- a/.github/workflows/validate-lfs.yml
+++ b/.github/workflows/validate-lfs.yml
@@ -1,0 +1,15 @@
+name: Validate Git LFS
+
+on: [pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: Validate
+    steps:
+      - uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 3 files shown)


stefanceriu (approved):
Good stuff although I was hoping we could have it as a git pre-commit hook as well. Probably not that important, this is not something we'll encounter very often

pixlwave:
> I was hoping we could have it as a git pre-commit hook as well.

Sadly this would work the same way as LFS and only be available for whoever added the hook (at which point they should just `git lfs install` instead). This seems to be a very intentional choice in git 🤷‍♂️

PR #205, opened 2022-09-22
Rename Provider Items to Item Proxies.

pixlwave:
This should be much easier to understand what is what now :)

Hoping it doesn't cause too many conflicts with #200.


ismailgulek (approved):
Definitely much more cleaner 👌

PR #207, opened 2022-09-23
Coordinator stops

ismailgulek:
Adds `stop` method on `Coordinator`s. Calling `stopLoading` by default made sense to me.

### DIFF
--- a/ElementX/Sources/Other/Coordinator.swift
+++ b/ElementX/Sources/Other/Coordinator.swift
@@ -37,7 +37,7 @@ protocol Coordinator: AnyObject {
     /// - Parameter childCoordinator: Child coordinator to remove.
     func remove(childCoordinator: Coordinator)
 
-    /// Stops job of the coordinator. Can be used to clear some resources.
+    /// Stops job of the coordinator. Can be used to clear some resources. Will be automatically called when the coordinator removed.
     func stop()
 }
 
--- a/ElementX/Sources/Other/Coordinator.swift
+++ b/ElementX/Sources/Other/Coordinator.swift
@@ -36,6 +36,9 @@ protocol Coordinator: AnyObject {
     ///
     /// - Parameter childCoordinator: Child coordinator to remove.
     func remove(childCoordinator: Coordinator)
+
+    /// Stops job of the coordinator. Can be used to clear some resources.
+    func stop()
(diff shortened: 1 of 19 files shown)


ElementX/Sources/Screens/RoomScreen/RoomScreenCoordinator.swift
```diff
@@ -58,4 +58,8 @@ final class RoomScreenCoordinator: Coordinator, Presentable {
     func toPresentable() -> UIViewController {
         roomScreenHostingController
     }
+
+    func stop() {
+        roomScreenViewModel.context.send(viewAction: .stop)
```
pixlwave, line 63: I think this is kind of cheating and should really be a new method added to the VM protocol rather than a view action.

ElementX/Sources/Screens/RoomScreen/View/Timeline/PlaceholderAvatarImage.swift
```diff
@@ -26,7 +26,8 @@ struct PlaceholderAvatarImage: View {
             Text(textForImage)
                 .padding(4)
                 .foregroundColor(.white)
-                .font(.title2.bold())
+                .font(.system(size: 200).weight(.semibold))
+                .minimumScaleFactor(0.001)
```
pixlwave, line 30: 🙌

ElementX/Sources/Other/Coordinator.swift
```diff
@@ -36,6 +36,9 @@ protocol Coordinator: AnyObject {
     ///
     /// - Parameter childCoordinator: Child coordinator to remove.
     func remove(childCoordinator: Coordinator)
+
+    /// Stops job of the coordinator. Can be used to clear some resources.
```
pixlwave, line 40: Not sure if anyone will read it, but could add that it is called automatically on remove.

pixlwave (approved):
Great idea to call it when removed 👏

ElementX/Sources/Screens/RoomScreen/RoomScreenCoordinator.swift
ismailgulek: Yeh totally right, updated.

PR #209, opened 2022-09-23
Expose avatar background for a given content id from ElementColors

ismailgulek:
For https://github.com/vector-im/element-x-ios/pull/196#discussion_r977540057

### DIFF
--- a/DesignKit/Sources/Colors/ElementColors.swift
+++ b/DesignKit/Sources/Colors/ElementColors.swift
@@ -57,6 +57,11 @@ public struct ElementColors {
     public var systemGray4: Color { Color(.systemGray4) }
     public var systemGray5: Color { Color(.systemGray5) }
     public var systemGray6: Color { Color(.systemGray6) }
+
+    public func avatarBackground(for contentId: String) -> Color {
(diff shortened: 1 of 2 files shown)


DesignKit/Sources/Colors/ElementColors.swift
```diff
@@ -139,3 +149,17 @@ public extension UIColor {
     public var tempActionBackgroundTint: UIColor { tempActionBackground.withAlphaComponent(0.2) }
     public var tempActionForegroundTint: UIColor { tempActionForeground.withAlphaComponent(0.2) }
 }
+
+private extension String {
+    /// Calculates a numeric hash same as Element Web
+    /// See original function here https://github.com/matrix-org/matrix-react-sdk/blob/321dd49db4fbe360fc2ff109ac117305c955b061/src/utils/FormattingUtils.js#L47
+    var hashCode: Int32 {
```
pixlwave, line 156: Should we remove the implementation in the ElementX target?

pixlwave (approved):
LGTM 😎

DesignKit/Sources/Colors/ElementColors.swift
ismailgulek: We might need it in future somehow, let's keep it for some time.

PR #211, opened 2022-09-23
Run unit test on pull request HEAD

jplatte:
… instead of on a merge commit.

This should make the coverage diffs more reliable since they will report the difference of the base to the actual latest commit of the PR, rather than to what the repo would look like if the PR was merged.


stefanceriu (approved):
Looks good to me although I don't think it's going to change much as linear history is enforced on repository.

PR #215, opened 2022-09-27
Update snapshotting library to 1.10.0

ismailgulek:
Updates snapshotting library to [1.10.0](https://github.com/pointfreeco/swift-snapshot-testing/releases/tag/1.10.0) and enable usage of `perceptualPrecision` as `0.98`.


stefanceriu (approved):
This seems to have done the trick and it only took 37 minutes to run as opposed to the previous 1 hour plus. Really nice! 👏

Issue #216, opened 2022-09-27
Memory leak in RoomProxy


Johennes:
**Describe the bug**

`RoomProxy` and `RoomTimelineProvider` strongly retain each other causing memory to be leaked for every room.

**To Reproduce**

1. Open a room
2. Back out again to the main screen
3. Fire up the memory graph debugger
4. Notice the leak

**Expected behavior**

Should not leak.

**Screenshots**

![Screenshot 2022-09-27 at 20 56 31](https://user-images.githubusercontent.com/1137962/192612785-726386e3-ea53-42e2-9a13-5a91fd189d76.png)

PR #217, opened 2022-09-27
Fix strong reference cycle between RoomProxy and RoomTimelineProvider

Johennes:
Fixes: #216

### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Services/Room/MockRoomProxy.swift
+++ b/ElementX/Sources/Services/Room/MockRoomProxy.swift
@@ -17,10 +17,10 @@
 import Combine
 import MatrixRustSDK
 
-struct MockRoomProxy: RoomProxyProtocol {
+class MockRoomProxy: RoomProxyProtocol {
     let id = UUID().uuidString
     let name: String? = nil
-    let displayName: String?
+    let displayName: String? = nil
     
     let topic: String? = nil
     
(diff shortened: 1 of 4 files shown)


stefanceriu (commented):
I'm not quite sure I follow this one: the roomProxy is a direct dependency of the TimelineProvider and should be strongly retained by it. If there's a cycle it should be handled another way.

LE: I had a closer look at it, we need to take the timelineProvider out of the roomProxy and create it when needed in the AppCoordinator. The timelineProvider can register itself as a listener on its own roomProxy dependency.

Johennes:
> I'm not quite sure I follow this one: the roomProxy is a direct dependency of the TimelineProvider and should be strongly retained by it. If there's a cycle it should be handled another way.

_Johannes tries to get weird retain cycle context from last night back into his head. 😅_

I think they're mutual dependencies, right? The reason I made the reference in `RoomTimelineProvider` the weak one is that `RoomTimelineProvider` seems to only ever be created (and retained) in `RoomProxy`. `RommProxy`, on the other hand, is created and returned in `ClientProxy` so I assumed this was the object the caller would retain.

> LE: I had a closer look at it, we need to take the timelineProvider out of the roomProxy and create it when needed in the ~AppCoordinator~UserSessionFlowCoordinator. The timelineProvider can register itself as a listener on its own roomProxy dependency.

That might be a better solution altogether. I'll have to return to this after hours.

Johennes:
Discarding this over #220.

Issue #218, opened 2022-09-27
Add changelog.d folder in Xcode project


Johennes:
**Is your feature request related to a problem? Please describe.**

Element iOS includes the `changelog.d` folder in the Xcode project which is very handy because this way changelog files – which are required for every pull request – can be edited directly from Xcode.

**Describe the solution you'd like**

Include the folder in `project.yml`.

**Describe alternatives you've considered**

–

**Additional context**

–

PR #219, opened 2022-09-27
Include changelog.d and other common files in Xcode project

Johennes:
Fixes: #218

### Pull Request Checklist

- [ ] Voiceover enabled.


stefanceriu (commented):
While I like the idea I don't love having all these files pile up on top of our project navigator. Is there a way we can drop them inside a miscellaneous folder or something?

Johennes:
Yeah, sure I can do that. I'd also be fine with only including the `changelog.d` folder for now. I only added the others because element-ios also included these. But I don't think they're as widely used as the changelog folder.

stefanceriu:
> sure I can do that

I'm not actually sure you **can** actually do that looking at the specs. If that's the case let's just merge is at as it.

stefanceriu (approved):

Johennes:
> I'm not actually sure you **can** actually do that looking at the specs. If that's the case let's just merge is at as it.

Hm, looks like you're right. The `fileGroups` property doesn't allow any further options. So unless we want to move things around on the file system, all the items will have to be at the top level.

To limit the pain, I removed the markdown files for now and only included the `changelog.d` folder which, I think, is that one that is accessed most often.

PR #220, opened 2022-09-28
Fix strong reference cycle between RoomProxy and RoomTimelineProvider

stefanceriu:
Related to vector-im/element-x-ios/pull/217


stefanceriu:
Perfect, thank you 👍

PR #223, opened 2022-09-29
Add notes for how to debug the network traffic

Johennes:
<del>Hoping this is fine to submit without a changelog entry given that it's just documentation. But let me know otherwise.</del>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,6 +40,10 @@ ElementX uses Fastlane for running actions on the CI and tries to keep the confi
 
 Please run `bundle exec fastlane` to see available options.
 
+### Network debugging proxy
+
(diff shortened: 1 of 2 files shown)


ismailgulek:
> Hoping this is fine to submit without a changelog entry given that it's just documentation. But let me know otherwise.

You can add one with `doc` extension i suppose. Details are at the [changelog section](https://github.com/matrix-org/matrix-ios-sdk/blob/develop/CONTRIBUTING.md#changelog). We could even copy that into here with some modifications.

Johennes:
> > Hoping this is fine to submit without a changelog entry given that it's just documentation. But let me know otherwise.
> 
> You can add one with `doc` extension i suppose. Details are at the [changelog section](https://github.com/matrix-org/matrix-ios-sdk/blob/develop/CONTRIBUTING.md#changelog). We could even copy that into here with some modifications.

Argh, I totally missed the `doc` extension. 🤦‍♂️ Have added a changelog file now.

PR #226, opened 2022-10-06
Set a proper user agent

### LINKED ISSUE #225: Add a proper user agent

The current user agent is "rust-sdk-ios" which isn't very descriptive and limits the usefulness of features such as the upcoming device manager.

Element X should use the standard user agent format similar to what Element iOS does.

–

**Additional context**

With https://github.com/matrix-org/matrix-rust-sdk/pull/1078 it's now possible to specify the user agent when initializing the SDK.

Johennes:
This applies a proper user agent to EX which enables more detailed display of EX sessions in Element's device manager.

This depends on https://github.com/matrix-org/matrix-rust-sdk/pull/1078 so will only be able to move forward once the SDK has been updated to a version that includes this change.

Fixes #225 

### Pull Request Checklist

- [ ] Voiceover enabled.


Johennes:
As the SDK was updated this is now ready for review. I believe the LFS quota errors are unrelated to my changes but please let me know otherwise.

ElementX/Sources/Other/Extensions/Bundle.swift
```diff
@@ -17,6 +17,20 @@
 import Foundation
 
 public extension Bundle {
+    /// Returns the real app bundle.
+    /// Can also be used in app extensions.
+    static let app: Bundle = {
```
stefanceriu, line 22: This doesn't seem to be used. What was the initial reason for adding it?

ElementX/Sources/Other/Extensions/String.swift
```diff
@@ -51,4 +51,15 @@ extension String {
         }
         return abs(hash)
     }
+    
+    func asciified() -> String {
+        guard !canBeConverted(to: .ascii) else {
+            return self
```
stefanceriu, line 57: This is potentially unsafe if it silently fails converiting to ascii. It would be safe to return a nullable and let the user handle it.

ElementX/Sources/Other/Extensions/String.swift
```diff
@@ -51,4 +51,15 @@ extension String {
...
+        }
+        let mutableString = NSMutableString(string: self)
+        guard CFStringTransform(mutableString, nil, "Any-Latin; Latin-ASCII; [:^ASCII:] Remove" as CFString, false) else {
```
stefanceriu, line 60: I wonder if we can use something like `folding(options: .diacriticInsensitive, locale: NSLocale.current)`, seems to be passing all the tests..

ElementX/Sources/Other/UserAgentFactory.swift
```diff
@@ -0,0 +1,71 @@
...
+import Foundation
+import UIKit
+
+#if !os(OSX)
+import DeviceKit
+#endif
+
+final class UserAgentFactory {
+    class func makeASCIIUserAgent() -> String? {
+        makeUserAgent()?.asciified()
```
stefanceriu, line 26: If `asciified` is only used here perhaps we should make the String extension private to this file.

ElementX/Sources/Other/UserAgentFactory.swift
```diff
@@ -0,0 +1,71 @@
...
+    }
+    
+    private class func makeUserAgent() -> String? {
+        let clientName = ElementInfoPlist.cfBundleDisplayName
+        let clientVersion = ElementInfoPlist.cfBundleShortVersionString
```
stefanceriu, line 31: These should probably be passed as constructor dependencies but I can see why they're not with all the Device/UIDevice/UIScreen etc. usages. Doing that would've helped with mocking the factory for the unit tests which are not entirely bulletproof atm

ElementX/Sources/Other/UserAgentFactory.swift
```diff
@@ -0,0 +1,71 @@
...
+//
+
+import Foundation
+import UIKit
+
+#if !os(OSX)
+import DeviceKit
+#endif
+
+final class UserAgentFactory {
```
stefanceriu, line 24: I think tehnically this is a builder not a factory as it only produces strings

stefanceriu (changes requested):

ElementX/Sources/Other/UserAgentFactory.swift
```diff
@@ -0,0 +1,71 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+import UIKit
+
+#if !os(OSX)
```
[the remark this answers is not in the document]
Johennes: Hm, I thought we needed this because DeviceKit itself didn't support macOS: https://github.com/devicekit/DeviceKit/blob/master/Package.swift#L17

ElementX/Sources/Other/Extensions/Bundle.swift
Johennes: Oh, I think I was using this before I noticed the `ElementInfoPlist` enum and then forgot cleaning it up. 🤦‍♂️

Thanks for catching. Have removed it now.

ElementX/Sources/Other/Extensions/String.swift
Johennes: Oh, excellent point! 👍

ElementX/Sources/Other/Extensions/String.swift
Johennes: It's a bit esoteric but this doesn't seem to work for some emojis (e.g. 🏩). I've added another test case for that. It's possible that there's a way to change that but I couldn't find one easily. The string transform code is ported from AlamoFire's Objective-C code.

ElementX/Sources/Other/Extensions/String.swift
Johennes: Interestingly `"someascii".canBeConverted(to: .ascii)` is `false` so I had to add another check to see if the string is already in ASCII.

ElementX/Sources/Other/UserAgentFactory.swift
Johennes: I think it'd still have to be `internal` due to the unit tests but I can still move it here if that's ok?

ElementX/Sources/Other/UserAgentFactory.swift
Johennes: Hm, I thought builder was more of a

```
let userAgent = UserAgentBuilder().setName("foo").setVersion("1.0").build()
```

pattern? Or maybe that's only valid in Java? 😅

It felt a little odd to me from the beginning to have this class with a single method but I couldn't think of a better way.

ElementX/Sources/Other/UserAgentFactory.swift
Johennes: Yeah, that seems tricky especially since the arguments would also change based on the OS. However, if you want to mock the factory itself, maybe we could just extract a protocol containing only `func makeASCIIUserAgent()` and a mock class returning a predefined value?

ElementX/Sources/Other/UserAgentFactory.swift
Johennes: Have renamed it now.

ElementX/Sources/Other/UserAgentFactory.swift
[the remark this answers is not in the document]
stefanceriu: Okay, let's leave it alone for now 👍

ElementX/Sources/Other/UserAgentFactory.swift
stefanceriu: I'm having second thoughts now, probably not worth the trouble

ElementX/Sources/Other/UserAgentFactory.swift
stefanceriu: The way I think about is based on what it returns, if from a heterogenous collection of types or just always the same type.

ElementX/Sources/Other/Extensions/String.swift
```diff
@@ -60,6 +60,6 @@ extension String {
         guard CFStringTransform(mutableString, nil, "Any-Latin; Latin-ASCII; [:^ASCII:] Remove" as CFString, false) else {
             return nil
         }
-        return mutableString as String
+        return mutableString.trimmingCharacters(in: .whitespaces)
```
stefanceriu, line 70: Should we add a test for this too?

stefanceriu (approved):
Great stuff 👍

ElementX/Sources/Other/UserAgentFactory.swift
stefanceriu: I don't think that's worth it to be honest..

ElementX/Sources/Other/UserAgentFactory.swift
Johennes: Oh, I see. That makes sense. It's all about the definition, I guess. 🙂

ElementX/Sources/Other/Extensions/String.swift
Johennes: It is actually covered in the `testStripsTheHeartbreakHotel` test. Sorry for only being able to reply after you already merged this. Thanks for the review! Looking forward to see my Element X sessions display nicely in device manager. 🙂

PR #227, opened 2022-10-07
Include app name in default session display name

Johennes:
This ports https://github.com/vector-im/element-ios/pull/6828 over to ElementX. Please see the Element iOS PR for the reasoning.

### DIFF
--- a/ElementX/Resources/Localizations/en.lproj/Untranslated.strings
+++ b/ElementX/Resources/Localizations/en.lproj/Untranslated.strings
@@ -27,3 +27,6 @@
 "login_tablet_device" = "Tablet";
 
 "a11y_all_chats_user_avatar_menu" = "User menu";
+
+// Parameter is the application display name (e.g. "ElementX")
+"default_session_display_name" = "%@ iOS";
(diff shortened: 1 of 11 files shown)


Johennes:
Hm, `testVerificationCancellation` keeps failing but it also fails on `develop` so I don't think this is related to my changes.

pixlwave (approved):
LGTM, although we're currently using `cfBundleName` everywhere else and the `cfBundleDisplayName` has the same value, so I'm wondering if it adds anything using that?

Johennes:
Ah, reading up in the docs it seems there is a slight difference:

> CFBundleName: A user-visible short name for the bundle.
> 
> This name can contain up to 15 characters. The system may display it to users if [CFBundleDisplayName](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundledisplayname) isn't set.

> CFBundleDisplayName: The user-visible name for the bundle, used by Siri and visible on the iOS Home screen.
>
> Use this key if you want a product name that's longer than [CFBundleName](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundlename).

So I think `CFBundleDisplayName` might be better to use actually because it allows for longer app names.

pixlwave:
Yep you're right, that makes sense to me. Would you mind updating any other uses of `ElementInfoPlist.cfBundleName` to be `ElementInfoPlist.cfBundleDisplayName` for consistency?

Johennes:
Ah, good idea, yes. Have replaced all the usages now.

pixlwave:
Thanks 🙏

PR #229, opened 2022-10-14
Switch to Xcode 14 and handle the UICollectionView-backed List.

pixlwave:
This PR does the following:
- Bump required OS versions to iOS 16/macOS 13
- Use Xcode 14.0 when building with Fastlane
- Add an `introspectCollectionView` method ala https://github.com/siteline/SwiftUI-Introspect/pull/169
- Introspect the List as a collection view.
- Only perform pagination handling when the timeline item count changes (otherwise its likely images/avatars/names having loaded in).
- Fix warnings on run scripts having no file outputs
- Adopt `if`/`guard let optional {` syntax everywhere.

The collection view based timeline isn't super stable in terms of back pagination scrolling, but seems better to merge these changes now and work on that afterwards. It's probably easiest reviewed commit by commit.


ismailgulek (approved):
LGTM!

PR #230, opened 2022-10-14
Attempt fixing flakey unit test

stefanceriu:
* switches from DispatchQueue.asyncAfter to Task.sleep
* pins the Xcode version on all fastlane actions
* makes Alpha builds only be triggered on PRs labelled as `Trigger-PR-Build`


pixlwave (approved):
LGTM 😎

PR #232, opened 2022-10-17
Stop syncing on resign active and start again on resume

ismailgulek:


stefanceriu (approved):
Nice use of the state machine! 👍

PR #233, opened 2022-10-19
Prevent `Unable to activate constraint with anchors .. because they h…

stefanceriu:
…ave no common ancestor.` crashes. Only link toasts to the top safe area instead of the navigation controller


ismailgulek (approved):
LGTM!

PR #236, opened 2022-10-21
Add URLInteractionModifier

alfogrillo:
This PR adds a new view modifier to be able to customize the tint color on Views when the user taps on a link.
⚠ The implementation override the `EnvironmentValues.openURL` variable to make the job.

https://user-images.githubusercontent.com/19324622/197225039-ba604a2d-2251-4af4-aaa9-c5f212663aa2.mp4

### DIFF
--- a/ElementX/Sources/Other/SwiftUI/Modifiers/URLInteractionModifier.swift
+++ b/ElementX/Sources/Other/SwiftUI/Modifiers/URLInteractionModifier.swift
@@ -1,0 +1,54 @@
+//
+// Copyright 2022 New Vector Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


stefanceriu (commented):
I might be missing something but it seems that this isn't really visible in ElementX. If this isn't something that the design team requested I would suggest we only keep this Element side until we know better how to handle links w/ contextual menus in ElX.

Otherwise I would rename the View extensions to `urlInteractionTintColor` instead of `tintColorURLInteraction` as I think it flows better. I would also maybe just keep the one that takes 2 colors as the opacity can very well be one of those. 

As swiftlint suggest we also need to remove the unused modified `url` parameter.

alfogrillo:
> I might be missing something but it seems that this isn't really visible in ElementX. If this isn't something that the design team requested I would suggest we only keep this Element side until we know better how to handle links w/ contextual menus in ElX.
> 
> Otherwise I would rename the View extensions to `urlInteractionTintColor` instead of `tintColorURLInteraction` as I think it flows better. I would also maybe just keep the one that takes 2 colors as the opacity can very well be one of those.
> 
> As swiftlint suggest we also need to remove the unused modified `url` parameter.

It's not visible since the tint color now is the same of the text. I didn't change it here. 
Btw I agree, let's keep this PR aside, since we don't really know (at least I don't) how the UX is gonna be in the future.
Also noticing that this PR isn't really solid since the tint color is gonna change for the whole cell (if you have two links in the message it will look like you tapped on both).
Refining a bit the code like you suggested btw. ;-)

PR #268, opened 2022-10-26
Fixed XcodeCloud release workflows

stefanceriu:
This PR will make Xcode Cloud:
* correctly access the path where towncrier (python dependency) is installed
* configure git with `Element CI` / `ci@element.io`
* generate version bump commit and push it using the Github Access Token to (Xcode Cloud doesn't have write access)

There's a discussion to be had on whether we should create release branches, automatically merge to master etc. but for now this will do


ismailgulek (approved):
LGTM!

PR #274, opened 2022-10-28
Enable e2ee support

stefanceriu:
This PR enable the sliding sync extensions and adds a missing `m.room.encryption` SS required state.
It also moves some extra screens to the `UserSessionFlowCoordinatorStateMachine` (bugReport still left to be done)


ismailgulek (approved):
LGTM!

PR #281, opened 2022-11-02
User session state machine crashes

stefanceriu:
- Fix UserSession state machine crashes on foregrounding and add the bug report screen to it
- Rename home screen actions


ismailgulek (approved):
LGTM!

PR #283, opened 2022-11-02
Concurrency Fixes

ismailgulek:
- Reviewed the places where we used `@MainActor`.
- Moved Rust client operations into a dedicated concurrent queue, so made sure we don't use them on main thread.
- Removed a couple of `Task.detached` operations


stefanceriu (approved):
Looks great 👍

PR #284, opened 2022-11-02
Fixes #279 - Fix plain timeline style layout, fix various TimelineVie…

### LINKED ISSUE #279: Fix plain room timeline layout
The plain timeline layout is rather broken atm as we only foused on the bubbled one. We should fix it

stefanceriu:
…ws' previews


ismailgulek (approved):
LGTM!

PR #288, opened 2022-11-04
Offline support

stefanceriu:
The application was previously cleaning any existing sessions if session restoration would've failed for any reason, including offline networking errors but the root problem is that we had to do network requests to being with.
This PR adopts a new RestorationToken format (while keeping compatibility with the old one) that contains the resolved homeserver URL which we can them pass back to the client builder and have it not need to do a request anymore


ismailgulek (approved):
LGTM 👏

PR #289, opened 2022-11-04
Fixes #273 - Remove home screen animations

### LINKED ISSUE #273: fancy roomlist sliding animation gets in the way of loading & filtering
**Describe the bug**
We currently use a fancy swiftui animation to slide rows of the roomlist into view as they appear. i think this might be overall detrimental when filtering and loading the roomlist via SS, as it makes it feel slower and less responsive than if it immediately updated. A fast fade might be okay, but the sliding makes you feel like you're having to wait for the app to stop moving around before you can use it, and so slows you down.

stefanceriu:
This PR removes home screen list change animations as they're not quite pretty or useful


ismailgulek (approved):
LGTM!

PR #290, opened 2022-11-04
Fixes #271 - Allow bubbles to horizontally fill the available space

### LINKED ISSUE #271: Bubbles should not be fixed width
The original EX design had fixed width bubbles. @nadonomy confirmed this was a mistake (as the massive bubbles with lots of internal padding look weird). It also wastes screen real estate if bubbles can’t expand to fill the screen (as telegram allows).

There should be new designs somewhere with more accurate bubbles.

stefanceriu:
This PR will drop the previous percent based bubble width layout and allow them to fill all availalbe horizontal space. It also remoes the `timelineWidth` env key as it's no longer needed.
Screenshot 2022-11-04 at 16 36 16


ismailgulek (approved):
Great!

PR #299, opened 2022-11-08
Reset keyboard after message sent

### LINKED ISSUE #269: Keyboard doesn’t change back to alphabetic mode after you send a msg
**Describe the bug**
Keyboard layout should reset to alphabetic mode every time you send a message. Instead, it stays in whatever mode you were previously in.

**To Reproduce**
1. Send a message that ends with a punctuation symbol
2. Hit send
3. Note the keyboard is still in punctuation mode

Nightly, iOS 16.1, iPhone 12 Pro Max

ismailgulek:
Fixes #269


stefanceriu (approved):
Nice one 👍

PR #308, opened 2022-11-11
Video playback

### LINKED ISSUE #238: Video message playback support
Allow video message playback from the timeline using Apple's default video player

ismailgulek:
Fixes #238

`FileCache` is definitely not a perfect implementation, but it should be good enough for now.
Pushing the media player is not perfect neither, we might implement some transitions in future.


stefanceriu (approved):
Nice!
The only comment I have is about the the caching: is it maybe time to completely drop the file caches from our side (only keeping the in memory image cache) and completely rely on the RustSDK's ones?

PR #318, opened 2022-11-15
Fixes #296 - Improved bubble timeline item corner radii

### LINKED ISSUE #296: Fix timeline bubble corner radii
In [290](https://github.com/vector-im/element-x-ios/pull/290) We have decided make the width of timeline bubbles dynamic, based on their content, but that breaks the corner radii treatment for grouped bubbles. We need to sync up with the design team and decide on what the correct fix is.

Image

stefanceriu:
image


ismailgulek (commented):
👌

PR #319, opened 2022-11-16
Introduce sending and sent state for timeline messages

### LINKED ISSUE #27: I want to see the sending state in the timeline
As an user I would like to see the sending state of recently sent messages:
- [ ] show an icon indicating that the message is currently in local echo state
- [ ] show another icon when the message has been sent
- [ ] remove all icons after a small delay

We don't currently have designs for this but it's not a big concern as tweaking things SwiftUI side is easy. Please use current Element for inspiration e.g. empty circle while in local echo, circle with checkmark after sent.

Rust SDK issue: https://github.com/matrix-org/matrix-rust-sdk/issues/639

paleksandrs:
closes #27 

Simulator Screen Recording - iPhone 14 Pro - 2022-11-16 at 16 28 59


ElementX/Sources/Services/Timeline/TimeLineItemContent/MessageTimelineItem.swift
```diff
@@ -23,6 +23,14 @@ protocol MessageContentProtocol: RoomMessageEventContentProtocol {
     var body: String { get }
 }
 
+/// The delivery status for the item.
+enum MessageTimelineItemDeliveryStatus: Hashable, Identifiable {
```
stefanceriu, line 27: Why does this need to be `Hashable` and `Identifiable` when `Equatable` is enough?

ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineDeliveryStatusView.swift
```diff
@@ -0,0 +1,49 @@
...
+            return "checkmark.circle"
+        }
+    }
+    
+    var body: some View {
+        Image(systemName: systemImageName)
+            .task {
+                if case .sent = deliveryStatus {
+                    try? await Task.sleep(nanoseconds: 1_000_000_000)
+                    showMe = false
```
stefanceriu, line 36: If this view already knows how to dismiss itself then it should probably handle all the visibility logic and stop relying on the bubble styler. The styler should only add it as a subview in the right place.

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemFactory.swift
```diff
@@ -166,7 +166,8 @@ struct RoomTimelineItemFactory: RoomTimelineItemFactoryProtocol {
                                     senderDisplayName: displayName,
                                     senderAvatar: avatarImage,
                                     properties: RoomTimelineItemProperties(isEdited: message.isEdited,
-                                                                           reactions: aggregateReactions(message.reactions)))
+                                                                           reactions: aggregateReactions(message.reactions),
+                                                                           deliveryStatus: message.deliveryStatus))
```
stefanceriu, line 170: Why is this implemented only for text messages? What about the rest?

Screenshot 2022-11-16 at 17 32 48

ElementX/Sources/Services/Timeline/TimeLineItemContent/MessageTimelineItem.swift
```diff
@@ -23,6 +23,14 @@ protocol MessageContentProtocol: RoomMessageEventContentProtocol {
...
+    var id: Self { self }
+    case unknown
+    case sending
+    case sent(secondsAgo: TimeInterval)
```
stefanceriu, line 31: Let's rename this associated value to `elapsedTime`

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
```diff
@@ -25,7 +25,24 @@ struct TimelineItemBubbledStylerView<Content: View>: View {
...
+            switch timelineItem.properties.deliveryStatus {
+            case .sending, .unknown:
+                _showDeliveryStatus = State(initialValue: true)
+            case let .sent(secondsAgo: secondsAgo):
+                _showDeliveryStatus = State(initialValue: secondsAgo < 3)
+            }
+        } else {
+            _showDeliveryStatus = State(initialValue: false)
+        }
+    }
```
stefanceriu, line 44: Please remove all of this and let the `TimelineDeliveryStatusView` handle the view state itself based on the `deliveryStatus` its constructed with.

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
```diff
@@ -37,6 +54,11 @@ struct TimelineItemBubbledStylerView<Content: View>: View {
                     HStack {
                         Spacer()
                         styledContentWithReactions
+                        if showDeliveryStatus {
+                            TimelineDeliveryStatusView(deliveryStatus: timelineItem.properties.deliveryStatus,
```
stefanceriu, line 58: This needs to happen for the `TimelineItemPlainStylerView` too, with a tweaked design preferably. You can use the `timelineStyle` Environment value

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
```diff
@@ -37,6 +54,11 @@ struct TimelineItemBubbledStylerView<Content: View>: View {
...
+                                                       showMe: $showDeliveryStatus.animation())
```
stefanceriu, line 59: Please handle animations and transitions internally in the `TimelineDeliveryStatusView`

ElementX/Sources/Services/Timeline/TimeLineItemContent/MessageTimelineItem.swift
```diff
@@ -23,6 +23,14 @@ protocol MessageContentProtocol: RoomMessageEventContentProtocol {
     var body: String { get }
 }
 
+/// The delivery status for the item.
+enum MessageTimelineItemDeliveryStatus: Hashable, Identifiable {
+    var id: Self { self }
+    case unknown
```
stefanceriu, line 29: Seeing as we don't really know how to construct or use `unknown` perhaps we should default to `sent/delivered` for everything unless we know otherwise

ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineDeliveryStatusView.swift
paleksandrs: > The styler should only add it as a subview in the right place.
if it is needed, right? If the view is not needed in the view hierarchy, it shouldn't be added and that should be controlled by the parent view. Can you please explain how do you see this working?

stefanceriu: Similar to `styledContentWithReactions` which returns an EmptyView if there's no content for it. That way the logic is nicely handled in one single place.

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemFactory.swift
paleksandrs: how did you manage to send other than text message? I thought this is not supported yet

stefanceriu: I sent them from a different client but the app doesn't handle them properly and shows a circle next to them.

ElementX/Sources/Services/Timeline/TimeLineItemContent/MessageTimelineItem.swift
paleksandrs: To be honest I would prefer to keep `unknow` status, to make sure that the status reflects the actual state of the message.  Also, I added[ new case](https://github.com/vector-im/element-x-ios/blob/6fbb453c941b1d630cd60d266aaeca5815fd1d08/ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineDeliveryStatusView.swift#L40) to hide the delivery status view if state is `unknow`

paleksandrs:
@stefanceriu I have finished with all fixes, please let me know if you want to change anything else. Thanks for the comments.

ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineDeliveryStatusView.swift
```diff
@@ -0,0 +1,67 @@
...
+        
+        switch deliveryStatus {
+        case .sending:
+            _showDeliveryStatus = State(initialValue: true)
+        case let .sent(elapsedTime: elapsedTime):
+            _showDeliveryStatus = State(initialValue: elapsedTime < 3)
+        case .unknown:
+            _showDeliveryStatus = State(initialValue: false)
+        }
+    }
```
stefanceriu, line 43: Why do you need to all of this when you can simply assign a value to `showDeliveryStatus`?

```
init(deliveryStatus: MessageTimelineItemDeliveryStatus) {
        self.deliveryStatus = deliveryStatus
        
        switch deliveryStatus {
        case .sending:
            showDeliveryStatus = true
        case let .sent(elapsedTime: elapsedTime):
            showDeliveryStatus = elapsedTime < 3.0
        case .unknown:
            showDeliveryStatus = false
        }
    }
```

stefanceriu (approved):
Thanks, this looks good to me 👍  (apart from that small @State var comment)

ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineDeliveryStatusView.swift
paleksandrs: Oh, I didn't know that. Earlier when swiftui came out, this was the way to set it. It seems that it changed. Thanks 👍

PR #320, opened 2022-11-17
Expose sliding sync proxy URL on the server selection screen

stefanceriu:
This PR will:
* move the sliding sync server url to the user defaults
* expose it on the server selection screen
* make the app more resilient to slidinc sync configuration errors. Instead of throwing fatal errors it will allow the app to start and show the home screen but will always stay in the skeletons view. The user can choose to log out and change the server settings.

Screenshot 2022-11-17 at 09 27 03

### DIFF
--- a/ElementX/Sources/Application/BuildSettings.swift
+++ b/ElementX/Sources/Application/BuildSettings.swift
@@ -20,8 +20,9 @@ final class BuildSettings {
     // MARK: - Servers
 
     static let defaultHomeserverAddress = "matrix.org"
-    static let slidingSyncProxyBaseURL = URL(staticString: "https://slidingsync.lab.element.dev")
-
+    
+    static let defaultSlidingSyncProxyBaseURLString = "https://slidingsync.lab.element.dev"
+    
     // MARK: - Bug report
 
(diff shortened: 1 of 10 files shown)


ElementX/Sources/Application/BuildSettings.swift
```diff
@@ -17,11 +17,20 @@
...
+        return userDefaults
+    }()
+    
     // MARK: - Servers
 
     static let defaultHomeserverAddress = "matrix.org"
-    static let slidingSyncProxyBaseURL = URL(staticString: "https://slidingsync.lab.element.dev")
-
+    
+    @UserDefault(key: "slidingSyncProxyBaseURLString", defaultValue: "https://slidingsync.lab.element.dev")
```
ismailgulek, line 31: I think `ElementSettings` would be more appropriate to store this. Here we can just store just the default value. We're also using `AppStorage` there so the new `UserDefault` wrapper seems redundant to me.

stefanceriu: 😮 didn't even think about AppStorage and completely forgot about ElementSettings 😅

ismailgulek (approved):

Issue #314, opened 2022-11-15
Send messages on return


stefanceriu:
Sending messages using the hardware keyboard is currently configured on the `cmd+return` shortcut. 
We should find a way to  do that on just `return` and ideally have it insert newlines with `shift`+`return`

PR #325, opened 2022-11-18
Send messages on return

paleksandrs:
Closes #https://github.com/vector-im/element-x-ios/issues/314

- Send a message by selecting Enter key from an external keyboard
- Enter a new line by selecting Shift+Enter key combination on an external keyboard


ElementX/Sources/Screens/RoomScreen/View/MessageComposer.swift
```diff
@@ -34,7 +34,10 @@ struct MessageComposer: View {
                 MessageComposerTextField(placeholder: ElementL10n.roomMessagePlaceholder,
                                          text: $text,
                                          focused: $focused,
-                                         maxHeight: 300)
+                                         maxHeight: 300,
+                                         onEnterKeyHandler: {
```
stefanceriu, line 38: Maybe `returnKeyActionHandler` would work better here .. and in the ComposerTextField (as in `return` instead of `enter`). 
And maybe it should be optional? 🤔

stefanceriu (approved):
Looks great 👏 
Just a small comment re naming inline

ElementX/Sources/Screens/RoomScreen/View/MessageComposer.swift
paleksandrs: Initially, I was using "return" as well, but then I checked :

> As a general rule, when they differ, Return is simply the key for typing a newline character, whereas Enter enters what you've already typed without adding a new line.

I think we are tracking enter key, no? 

> And maybe it should be optional? 

It is not used anywhere else and we can always change it to optional if needed, let's keep it as not optional for now. What do you think?

stefanceriu: Alrighty, I agree with you on both then 👍

PR #326, opened 2022-11-18
Re-write MXLogger in Swift

### LINKED ISSUE #166: Convert MXLogger to Swift
MXLogger has been brought across wholesale from EI as an Objective-C class. Given that #102 is a no-go, we should re-write this class in Swift which would remove the need for the objc wrapper part of MXLog too.

pixlwave:
This PR re-writes MXLogger as a Swift file. It is best reviewed as separate commits. The first adds some extra tests on the objc implementation and the second replaces this with the Swift implementation.

This is basically a straight re-write of the objc version, the only change is that it uses URLs instead of path strings where possible.

Fixes #166.


stefanceriu (approved):
Looks great! 👏

PR #327, opened 2022-11-18
Various fixes

stefanceriu:
* Decrease timeline back pagination size from 50 to 20
* Switch the roomSummaryProvider and roomTimelineProvider to collect on the default global queue instead of background
* Force the timeline to stay at the bottom when first loading
* Reverted *reverted* commit that was handling offsets on item changes, like edits

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/View/TimelineItemList.swift
+++ b/ElementX/Sources/Screens/RoomScreen/View/TimelineItemList.swift
@@ -82,20 +82,12 @@ struct TimelineItemList: View {
                     }
                 }
             }
-            .onChange(of: pinnedItem) { item in
-                guard let item else {
-                    return
-                }
-                
-                if item.animated {
-                    withAnimation(Animation.elementDefault) {
-                        proxy.scrollTo(item.id, anchor: item.anchor)
-                    }
-                } else {
-                    proxy.scrollTo(item.id, anchor: item.anchor)
-                }
-                
-                pinnedItem = nil
+            .onChange(of: timelineItems) { _ in
+                // Run this multiple times on first appearance to fix offset issues
+                updatePinnedOffset(scrollViewProxy: proxy)
+            }
+            .onChange(of: pinnedItem) { _ in
+                updatePinnedOffset(scrollViewProxy: proxy)
             }
         }
         .scrollDismissesKeyboard(.immediately)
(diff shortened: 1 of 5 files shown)


ElementX/Sources/Screens/RoomScreen/View/TimelineItemList.swift
```diff
@@ -198,6 +180,25 @@ struct TimelineItemList: View {
...
+            return
+        }
+        
+        if item.animated {
+            withAnimation(Animation.elementDefault) {
+                proxy.scrollTo(item.id, anchor: item.anchor)
+            }
+        } else {
+            proxy.scrollTo(item.id, anchor: item.anchor)
+        }
```
pixlwave, line 194: Could this be simplified to? Genuinely don't know if it is 100% equivalent or not.
```suggestion
        withAnimation(item.animated ? .elementDefault : .noAnimation) {
            proxy.scrollTo(item.id, anchor: item.anchor)
        }
```

ElementX/Sources/Screens/RoomScreen/View/TimelineItemList.swift
```diff
@@ -114,17 +106,17 @@ struct TimelineItemList: View {
...
+                timelineItems = items
             }
             
             // Pin to the bottom if empty
             if timelineItems.isEmpty {
-                if let lastItem = context.viewState.items.last {
+                if let lastItem = items.last {
                     let pinnedItem = PinnedItem(id: lastItem.id, anchor: .bottom, animated: false)
-                    timelineItems = context.viewState.items
+                    timelineItems = items
```
pixlwave, line 119: There's a few of these, but aren't they redundant with that `defer` above?

pixlwave (approved):
LGTM 👍

ElementX/Sources/Screens/RoomScreen/View/TimelineItemList.swift
stefanceriu: Unfortunately they are not, it behaves like it's fully animated and doesn't even scroll all the way down.

ElementX/Sources/Screens/RoomScreen/View/TimelineItemList.swift
stefanceriu: Kind of, the idea was to set the `pinnedItem` only after the timelineItems have been updated. But it doesn't really matter anymore as I ended up reverting this whole commit as it causing problems, presumably because of updating the pinnedItem too many times

PR #332, opened 2022-11-22
Use a VStack on the timeline

pixlwave:
This PR switches from a `LazyVStack` to a regular `VStack` for the timeline. This fixes various issues seen with the lazy stack not being able to properly estimate its content size (more specifically that is doesn't seem to cache previous item heights as part of this estimate).

`ReversedScrollView` has been customised to be specific for the timeline and renamed accordingly.

One tradeoff of this change is that memory usage will be higher than a LazyVStack when back paginating a lot in a room. In my testing an heavy amount of back pagination (mainly text items) resulted in around 500MB of memory usage which doesn't seem unworkable, but in the future we could put a limit on the number of items shown and start removing items as you scroll back.


ElementX/Sources/Screens/RoomScreen/View/TimelineItemList.swift
```diff
@@ -23,76 +23,74 @@ struct TimelineItemList: View {
...
+                                .id(item.id)
+                        }
+                        .opacity(opacityForItem(item))
+                        .padding(settings.timelineStyle.rowInsets)
+                        .onAppear {
+                            context.send(viewAction: .itemAppeared(id: item.id))
+                        }
+                        .onDisappear {
+                            context.send(viewAction: .itemDisappeared(id: item.id))
+                        }
```
pixlwave, line 60: I'm not sure if this is definitely needed any more but have left it in for now. Can remove if you think it makes sense.

stefanceriu: Let's leave them in for now.

stefanceriu (approved):
This looks great! 🚀

PR #333, opened 2022-11-22
Timeline scrolling performance issues

### LINKED ISSUE #330: Timeline scrolling performance issues
It seems that the context menus we create for timeline items are causing the scrolling to drop in performance. It gets worse the more you back paginate.

The only reference I found to it was from a WWDC 22 digital lounge here https://blog.timing.is/wwdc-swift-slack/

```
- I'm noticing a huge hit to scrolling performance when I add a contextMenu to cells in my lazyvstack. When I profiled the app, it showed that buttons in the contextMenu was what was taking so long to render. Does a context menu have to render on creation of a view? Is there any way to create a lazy context menu?
- We believe that this issue was fixed in a recent software update. If you are still seeing this, could you please file a feedback?
```

paleksandrs:
closes #330 

The issue was due to the repeated loading of avatars and usernames for list items. Updating list items with the loaded avatar or username, triggered body of `TimelineItemList` to get re-evaluated and that caused all of the list items to be recreated, causing a performance hit. 

There are other things that could potentially be improved. For example, when a user tries to download more items, all items in the view model are recreated, causing again all of the list items to be recreated. Maybe it could be possible just to update state objects by adding new items only.

### DIFF
--- a/ElementX/Sources/Services/Timeline/RoomTimelineController.swift
+++ b/ElementX/Sources/Services/Timeline/RoomTimelineController.swift
@@ -378,7 +378,7 @@ class RoomTimelineController: RoomTimelineControllerProtocol {
     }
     
     private func loadUserAvatarForTimelineItem(_ timelineItem: EventBasedTimelineItemProtocol) async {
-        if timelineItem.shouldShowSenderDetails == false {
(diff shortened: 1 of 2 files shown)


stefanceriu (approved):
Looking good! 🚀

PR #335, opened 2022-11-23
UI test fixes

stefanceriu:
Various UI tests fixes
* remove status bar when running snapshotting
* remove stored OS versions, allow running on same device on any OS
* fix ServiceLocator and fallbackLanguage when running UITests
* switch to iPhone 14
* update all screenshots


ElementX/Sources/Application/ServiceLocator.swift
```diff
@@ -0,0 +1,39 @@
...
+    static var shared: ServiceLocator {
+        if let serviceLocator {
+            return serviceLocator
+        }
+        
+        let serviceLocator = ServiceLocator()
+        self.serviceLocator = serviceLocator
+        
+        return serviceLocator
+    }
```
pixlwave, line 30: Would this not be equivalent to
```
private(set) static var shared = ServiceLocator()
```
which is lazy by the fact it is `static`

UITests/SupportingFiles/target.yml
```diff
@@ -52,7 +52,7 @@ targets:
     - package: Sentry
       linkType: static
     - package: SnapshotTesting
-      linkType: static
+      linkType: staticfasd
```
pixlwave, line 55: Is this a real `linkType` or a typo?? 😅

.gitignore
```diff
@@ -44,9 +44,9 @@ Tools/Scripts/element-android
...
-!/UITests/Sources/__Snapshots__/Application/16-0-en-GB-iPad-9th-generation.*.png
-!/UITests/Sources/__Snapshots__/Application/16-0-de-DE-iPhone-13-Pro-Max.*.png
-!/UITests/Sources/__Snapshots__/Application/16-0-fr-FR-iPhone-13-Pro-Max.*.png
-!/UITests/Sources/__Snapshots__/Application/16-0-en-GB-iPhone-13-Pro-Max.*.png
+!/UITests/Sources/__Snapshots__/Application/de-DE-iPad-9th-generation.*.png
+!/UITests/Sources/__Snapshots__/Application/fr-FR-iPad-9th-generation.*.png
+!/UITests/Sources/__Snapshots__/Application/en-GB-iPad-9th-generation.*.png
+!/UITests/Sources/__Snapshots__/Application/de-DE-iPhone-13-Pro-Max.*.png
+!/UITests/Sources/__Snapshots__/Application/fr-FR-iPhone-13-Pro-Max.*.png
+!/UITests/Sources/__Snapshots__/Application/en-GB-iPhone-13-Pro-Max.*.png
```
pixlwave, line 52: The phone screenshots are missing from this PR, presumably because they're being ignored. Untested, but presumably we should be unignoring like so:
```suggestion
!/UITests/Sources/__Snapshots__/Application/de-DE-iPhone-14.*.png
!/UITests/Sources/__Snapshots__/Application/fr-FR-iPhone-14.*.png
!/UITests/Sources/__Snapshots__/Application/en-GB-iPhone-14.*.png
```

ElementX/Sources/Application/Application.swift
```diff
@@ -39,7 +39,12 @@ struct Application: App {
                     .task {
                         applicationCoordinator.start()
                     }
+                    .statusBarHidden(shouldHideStatusBar)
```
pixlwave, line 42: Haha, that was easier than I expected 👏

pixlwave (commented):
Looks good to me, but we'll need to include the phone screenshots too :)

.gitignore
stefanceriu: This is why I don't like the .gitignore thing, sure to forget about it 😁

ElementX/Sources/Application/ServiceLocator.swift
stefanceriu: It sure is, don't know what i was thinking ¯\_(ツ)_/¯

UITests/SupportingFiles/target.yml
stefanceriu: I have this bad habbit of chaning a file to have it pop up in SourceTree so I can easily check it's history. Strange that it built with this here though. Seems it doesn't matter if they're static or the default dynamic, so I'll just remove linkType from all of them.

stefanceriu: Actually doing that doesn't even change the project file, I think it was unnecessary to being with

PR #336, opened 2022-11-24
Setup tracing with a typed configuration.

pixlwave:
This change allows us to create tracing presents so that its easier to increase the logging as needed.

I've also added the `RUST_BACKTRACE` env var to the run scheme (but disabled by default) as that seems like it would be useful to easily toggle on and off.


stefanceriu (approved):
This looks ah-mazing, nicely done 👍

PR #337, opened 2022-11-24
Crash fixes

stefanceriu:
* fix diffing crashes because of collection on a concurrent dispatch queue
* fix rust side crashes when restoring SS data from invitations
* add diffing logs
* adopt new addTimelineListener API and log errors accordingly

### DIFF
--- a/ElementX/Sources/Services/Timeline/RoomTimelineProvider.swift
+++ b/ElementX/Sources/Services/Timeline/RoomTimelineProvider.swift
@@ -28,6 +28,7 @@ private class RoomTimelineListener: TimelineListener {
 
 class RoomTimelineProvider: RoomTimelineProviderProtocol {
     private let roomProxy: RoomProxyProtocol
+    private let serialDispatchQueue: DispatchQueue
     private var cancellables = Set<AnyCancellable>()
     
     let itemsPublisher = CurrentValueSubject<[TimelineItemProxy], Never>([])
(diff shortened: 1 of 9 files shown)


ElementX/Sources/Services/Timeline/RoomTimelineProvider.swift
```diff
@@ -63,10 +70,13 @@ class RoomTimelineProvider: RoomTimelineProviderProtocol {
         // Set this back to false after actually updating the items or if failed
         backPaginationPublisher.send(true)
         
+        MXLog.info("Started back pagination request")
```
pixlwave, line 73: This file has a mixture of `info` and `verbose` added. I'm sure the distinction is fine, but just pointing it out incase you changed your mind part way though.

ElementX/Sources/Services/Timeline/RoomTimelineProvider.swift
```diff
@@ -78,95 +88,143 @@ class RoomTimelineProvider: RoomTimelineProviderProtocol {
...
-                    return partialResult
+            .reduce(itemProxies) { currentItems, diff in
+                guard let collectionDiff = buildDiff(from: diff, on: currentItems) else {
+                    MXLog.error("Failed building CollectionDifference from \(diff)")
+                    return currentItems
                 }
                 
-                return partialResult.applying(collectionDiff) ?? partialResult
+                guard let updatedItems = currentItems.applying(collectionDiff) else {
+                    MXLog.error("Failed applying diff: \(collectionDiff)")
```
pixlwave, line 144: Will logging these diffs potentially log the contents of messages? Same comment in the room summary provider if this is true.

pixlwave (approved):
LGTM 👍

PR #347, opened 2022-11-28
Various fixes

stefanceriu:
This PR will:
* re-enable sliding sync based timelines
* cleaning up of room components on screen pop
* moving the user display name and avatar fetching to a serial queue
* other small tweaks

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/RoomScreenCoordinator.swift
+++ b/ElementX/Sources/Screens/RoomScreen/RoomScreenCoordinator.swift
@@ -25,10 +25,16 @@ struct RoomScreenCoordinatorParameters {
 }
 
 final class RoomScreenCoordinator: CoordinatorProtocol {
-    private let parameters: RoomScreenCoordinatorParameters
+    private var parameters: RoomScreenCoordinatorParameters?
 
-    private var viewModel: RoomScreenViewModelProtocol
-    private var navigationController: NavigationController { parameters.navigationController }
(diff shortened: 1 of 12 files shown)


ElementX/Sources/Screens/RoomScreen/RoomScreenCoordinator.swift
```diff
@@ -25,10 +25,16 @@ struct RoomScreenCoordinatorParameters {
...
 final class RoomScreenCoordinator: CoordinatorProtocol {
-    private let parameters: RoomScreenCoordinatorParameters
+    private var parameters: RoomScreenCoordinatorParameters?
 
-    private var viewModel: RoomScreenViewModelProtocol
-    private var navigationController: NavigationController { parameters.navigationController }
+    private var viewModel: RoomScreenViewModelProtocol?
+    private var navigationController: NavigationController {
+        guard let parameters else {
+            fatalError()
```
ismailgulek, line 33: I'm not very sure this is a good pattern. We may need to tweak it again in future.

ismailgulek (approved):

ElementX/Sources/Screens/RoomScreen/RoomScreenCoordinator.swift
stefanceriu: Yeah, I know what you mean, I'm not particularly happy with it either. I'm hoping we won't have to do it in many places or that we can figure out the NavigationStack retaining in the meantime

PR #349, opened 2022-12-02
Use a UITableView component for the timeline.

pixlwave:
This PR replaces the SwiftUI timeline with one based upon `UITableView`. The cells are configured with a `UIHostingConfiguration` so are still SwiftUI based. This gives us the ability to update the scroll position after changing items but before the changes are rendered meaning the scroll jumps should be fixed.

- The back pagination threshold is now double the height of the table view
- The table will keep back paginating to fill the threshold.
- Items are only added/updated when the table isn't scrolling as updating the scroll position kills any inertia. (Further pagination is blocked until items are added from a pending pagination).
- Conforms timeline items to `Hashable` for use in the diffable data source.
- `TimelineView` has been merged into `RoomScreen` (happy to revert this if it doesn't make sense).

### Todo
- The syncing indicator still needs a `ProgressView` to match the figma (future PR).
- Add UI tests so we notice if any desired behaviours are broken (future PR).
- [ ] Find out why timeline cells are leaked.

https://user-images.githubusercontent.com/6060466/205295673-c0e1041f-68bc-46f5-9846-554d538bd2f7.mp4

| Small Timeline |
| - |
| Simulator Screen Shot - iPhone 14 - 2022-12-02 at 12 41 42 |


ElementX/Sources/Screens/RoomScreen/View/TimelineTableView.swift
```diff
@@ -0,0 +1,400 @@
...
+        viewModelContext.viewState.paginateBackwardsPublisher.send(())
+        return tableView
+    }
+    
+    func updateUIView(_ uiView: UITableView, context: Context) {
+        context.coordinator.update()
+        
+        if context.coordinator.timelineStyle != timelineStyle {
+            context.coordinator.timelineStyle = timelineStyle
+        }
```
pixlwave, line 50: Thinking about it, this probably highlights that we should observe the settings inside of the view model and only inject it into the environment on the actually cell.

ElementX/Sources/Other/ScrollViewAdapter.swift
```diff
@@ -25,30 +25,29 @@ class ScrollViewAdapter: NSObject, UIScrollViewDelegate {
...
         
-    private func update() {
-        guard let scrollView else { return }
+    private func update(_ scrollView: UIScrollView) {
         isScrolling.send(scrollView.isDragging || scrollView.isDecelerating)
     }
     
     func scrollViewWillBeginDragging(_ scrollView: UIScrollView) {
-        update()
+        update(scrollView)
```
stefanceriu, line 35: Do we really need all of these? 🤔

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -60,14 +64,26 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
                     self.state.items[viewIndex] = timelineViewFactory.buildTimelineViewFor(timelineItem: timelineItem)
                 case .startedBackPaginating:
                     self.state.isBackPaginating = true
+                    ServiceLocator.shared.userNotificationController.submitNotification(UserNotification(id: "pagination",
+                                                                                                         type: .toast,
+                                                                                                         title: ElementL10n.roomTimelineSyncing,
+                                                                                                         persistent: true))
                 case .finishedBackPaginating:
                     self.state.isBackPaginating = false
+                    ServiceLocator.shared.userNotificationController.retractNotificationWithId("pagination")
```
stefanceriu, line 73: In other places we went with a static var for notification identifiers on Ismail's requests, you  might want to do the same here.

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -26,6 +27,9 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
     private let timelineController: RoomTimelineControllerProtocol
     private let timelineViewFactory: RoomTimelineViewFactoryProtocol
     private let mediaProvider: MediaProviderProtocol
+    
+    /// A publisher used to throttle back pagination requests.
+    private let paginateBackwardsPublisher = PassthroughSubject<Void, Never>()
```
stefanceriu, line 32: Why do we need to throttle them to begin with?

ElementX/Sources/Screens/RoomScreen/View/TimelineTableView.swift
```diff
@@ -0,0 +1,424 @@
...
+        /// Configures a diffable data source for the timeline's table view.
+        private func configureDataSource() {
+            guard let tableView else { return }
+            
+            dataSource = .init(tableView: tableView) { tableView, indexPath, timelineItem in
+                let cell = tableView.dequeueReusableCell(withIdentifier: TimelineItemCell.reuseIdentifier, for: indexPath)
+                guard let cell = cell as? TimelineItemCell else { return cell }
+                
+                cell.item = timelineItem
+                #warning("Do we need a weak self here???")
```
stefanceriu, line 146: When in doubt, always weak 😅

stefanceriu (approved):
This looks so promising, hope we can squash all the little edge cases and get something really solid out of it 👏

ElementX/Sources/Other/ScrollViewAdapter.swift
pixlwave: So in this instance I couldn't set `adapter.scrollView = timelineTableView` because then the `Coordinator` is no longer the delegate. So I'm calling all these methods externally and passing the scrollView to update. Its definitely cheating though but wasn't sure how else to reuse this class...

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
pixlwave: My bad, I missed this 👍

stefanceriu (approved):
lgtm 👍 👏

PR #354, opened 2022-12-06
Fix keychain access on macOS.

pixlwave:
Running ElementX on macOS (Designed for iPad) didn't have the correct access to the App Group for sharing the keychain. This PR adds the Keychain Access Group back to properly support macOS builds.

Note: This PR will require everyone to sign in once again on iOS.


stefanceriu (approved):
Excellent! 👏  Thank you for taking care of it 👍

PR #356, opened 2022-12-07
Update codecov  

paleksandrs:
- excluded some folders that shouldn't be added to test coverage 
- added integration tests flag

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -9,7 +9,10 @@ coverage:
     patch: false
 
 ignore:
-  - "ElementX/Source/Generated"  # ignore the folder and all its contents
+  - "ElementX/Sources/Generated"
+  - "ElementX/Sources/Vendor"
+  - "ElementX/Sources/UITests"
+  - "Tools"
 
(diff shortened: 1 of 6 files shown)


codecov.yml
```diff
@@ -9,7 +9,10 @@ coverage:
     patch: false
 
 ignore:
-  - "ElementX/Source/Generated"  # ignore the folder and all its contents
```
stefanceriu, line 12: 🤦‍♂️

stefanceriu (changes requested):
This looks good but I don't think coverage is enabled for the integration tests https://github.com/vector-im/element-x-ios/blob/develop/ElementX/SupportingFiles/target.yml#L31

paleksandrs:
> This looks good but I don't think coverage is enabled for the integration tests https://github.com/vector-im/element-x-ios/blob/develop/ElementX/SupportingFiles/target.yml#L31

Good catch! Thank you 👍

stefanceriu (approved):

PR #359, opened 2022-12-09
Use a view controller for the timeline.

pixlwave:
This PR changes from wrapping a `UITableView` to wrapping a `UIViewController` with its own table view.
- This exposes lifecycle methods which allows a tidier fix for when the timeline is pre-filled with items from the start.
- The table view's frame changes can now be managed in `viewWillLayoutSubviews` instead of using a frame observer.
- It also tidies up the responsibility of the Coordinator to simply synchronise view state and pass view actions back up.
- `TimelineTableView` has been renamed to `TimelineView`.


stefanceriu (approved):
Looks great 👍

PR #363, opened 2022-12-12
Add basic UI tests for the timeline.

pixlwave:
This initial PR includes 2 tests: a room with 2 items, and the same room with both a single incoming message and a single message in a pagination resposne.

To do this, the MockTimelineController is now configurable with incoming items and pagination responses.


stefanceriu (approved):
lgtm 👏

PR #372, opened 2022-12-15
Resetting the app after deletion and cleanup settings

stefanceriu:
This PR will make it so the user state is completely reset if the app has been deleted since the last run
It will also merge together the BuildSetting and ElementSettings into ApplicationSettings and move it to the ServiceLocator


ElementX/Sources/Application/ApplicationSettings.swift
```diff
@@ -0,0 +1,134 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+import SwiftUI
+
+/// Store Element specific app settings.
+final class ApplicationSettings: ObservableObject {
```
pixlwave, line 21: Me again, your annoying "can we rename x to y" reviewer...

Could we just go with `AppSettings` to match `AppCoordinator`, `AppDelegate` etc?

ElementX/Sources/Application/ApplicationSettings.swift
```diff
@@ -0,0 +1,134 @@
...
+            fatalError("Fail to load shared UserDefaults")
+        }
+        return userDefaults
+    }
+    
+    static func reset() {
+        let dictionary = Self.store.dictionaryRepresentation()
+        dictionary.keys.forEach { key in
+            Self.store.removeObject(forKey: key)
+        }
```
pixlwave, line 44: I believe this can be simplified to
```
Self.store.removePersistentDomain(forName: InfoPlistReader.target.appGroupIdentifier)
```

ElementX/Sources/Application/ApplicationSettings.swift
```diff
@@ -0,0 +1,134 @@
...
+    @AppStorage(UserDefaultsKeys.enableInAppNotifications.rawValue, store: store)
+    var enableInAppNotifications = true
+
+    /// Tag describing which set of device specific rules a pusher executes.
+    @AppStorage(UserDefaultsKeys.pusherProfileTag.rawValue, store: store)
+    var pusherProfileTag: String?
+        
+    // MARK: - Other
+    
+    var permalinkBaseURL = URL(staticString: "https://matrix.to")
```
pixlwave, line 133: I realise this was a `var` in `BuildSettings`, but it should probably be a `let` no?

ElementX/Sources/Application/ApplicationSettings.swift
```diff
@@ -0,0 +1,134 @@
...
+            Self.store.removeObject(forKey: key)
+        }
+    }
+    
+    // MARK: - Application
+    
+    /// Simple flag to check if app has been deleted between runs.
+    /// Used to clear data stored in the shared container and keychain
+    @AppStorage(UserDefaultsKeys.wasAppPreviouslyRan.rawValue, store: store)
+    var wasAppPreviouslyRan = false
```
pixlwave, line 52: A slightly cleaner name might be `hasAppLaunched`. I'll stop with the renaming suggestions now 🙈

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineStyle.swift
```diff
@@ -35,7 +35,7 @@ enum TimelineStyle: String, CaseIterable {
 // MARK: - Environment
 
 private struct TimelineStyleKey: EnvironmentKey {
-    static let defaultValue = BuildSettings.defaultRoomTimelineStyle
+    static let defaultValue = TimelineStyle.bubbles
```
pixlwave, line 38: This should probably read the default value from the app settings tbh. Its only for situations when the value is accessed from the environment but we haven't injected one higher up.

ElementX/Sources/Screens/Settings/View/SettingsScreen.swift
```diff
@@ -110,30 +110,26 @@ struct SettingsScreen: View {
...
-                       role: .destructive) { context.send(viewAction: .crash)
-                }
-                .listRowInsets(listRowInsets)
-                .accessibilityIdentifier("crashButton")
+            
+            Button("Crash app",
+                   role: .destructive) { context.send(viewAction: .crash)
             }
+            .listRowInsets(listRowInsets)
+            .accessibilityIdentifier("crashButton")
```
pixlwave, line 118: Probably worth wrapping in a `#if DEBUG`?

ElementX/Sources/Services/UserSession/UserSessionStore.swift
```diff
@@ -36,6 +36,17 @@ class UserSessionStore: UserSessionStoreProtocol {
         MXLog.debug("Setup base directory at: \(baseDirectory)")
     }
     
+    /// Deletes all data stored in the shared container and keychain
+    func reset() {
+        do {
+            try FileManager.default.removeItem(at: baseDirectory)
+        } catch {
+            MXLog.failure("Failed resetting user session store: \(error)")
```
pixlwave, line 44: I think this will send a failure to Sentry if the user installed the app, deleted the app without signing in and then installed it again. Probably checking that the directory exists before deleting it.

UnitTests/Sources/AnalyticsTests.swift
```diff
@@ -19,10 +19,15 @@ import AnalyticsEvents
 import XCTest
 
 class AnalyticsTests: XCTestCase {
+    private var applicationSettings: ApplicationSettings!
+    
+    override func setUp() {
+        ApplicationSettings.reset()
```
pixlwave, line 25: Just to double check... With this in place, running the unit tests forces your device/simulator to sign out on next launch.

pixlwave (approved):
Much cleaner 👏 Some thoughts inline.

ElementX/Sources/Application/ApplicationSettings.swift
stefanceriu: Should we do `App.swift`? 😁

pixlwave:
Ok... One more thought on naming. On merging build settings with user settings, do you think it might be helpful to have a convention for constants e.g. suffixing them with `featureFlag`/`flag` or something so you can tell at the point of use that its read only.

ElementX/Sources/Application/ApplicationSettings.swift
pixlwave: Sadly I think thats the one place where we have to use `Application` due to the `App` protocol existing.

stefanceriu:
I don't think that's necessary, it's pretty easy to check.

ElementX/Sources/Application/ApplicationSettings.swift
stefanceriu: I don't think it's cleaner.. makes me think about the current instance "having been launched"

stefanceriu: Agreed on `hasAppLaunchedOnce`

PR #375, opened 2022-12-19
Fixes #329 - Simplify video player UI until we agree on final designs.

### LINKED ISSUE #329: Video playback doesn't hide the iOS topbar menu
* Record a screencast vid on an iphone
* Send it in a room
* Play it back on EX
* Observe the top is a mangled mess due to the recorded UI getting overlaid on the real UI
* Declare the app flakey

stefanceriu:


pixlwave (approved):
Makes sense for now 👍

PR #376, opened 2022-12-19
Added unit tests for NotificationManager

paleksandrs:
- added unit test for `NotificationManager` (increasing code coverage for the file from 0 to 93%)


ElementX.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -23,8 +23,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/Cocoanetics/DTCoreText",
       "state" : {
-        "revision" : "b664664825da565b4c2b7a17dbe2369f68ae43d9",
-        "version" : "1.6.26"
+        "revision" : "9d2d4d2296e5d2d852a7d3c592b817d913a5d020",
+        "version" : "1.6.27"
```
pixlwave, line 27: These updates to `Package.resolved` occur as an annoying side effect of running `xcodegen` whilst the project is opened in Xcode. I don't think we should be checking them in with unrelated PRs though 🙂

ElementX/Sources/Services/Notification/Manager/NotificationManager.swift
```diff
@@ -45,22 +46,25 @@ class NotificationManager: NSObject, NotificationManagerProtocol {
...
             }
         }
     }
 
-    func register(with deviceToken: Data) {
-        setPusher(with: deviceToken, clientProxy: clientProxy)
+    func register(with deviceToken: Data, completion: ((Bool) -> Void)? = nil) {
+        setPusher(with: deviceToken,
+                  clientProxy: clientProxy,
+                  completion: completion)
```
pixlwave, line 67: Could we mark `register` and `setPusher` as `async` functions rather than adding a completion handler?

ElementX/Sources/Services/Notification/Manager/UserNotificationCenterProtocol.swift
```diff
@@ -0,0 +1,25 @@
...
+// limitations under the License.
+//
+
+import Foundation
+import UserNotifications
+
+protocol UserNotificationCenterProtocol {
+    var delegate: UNUserNotificationCenterDelegate? { get set }
+    func add(_ request: UNNotificationRequest, withCompletionHandler completionHandler: ((Error?) -> Void)?)
+    func requestAuthorization(options: UNAuthorizationOptions, completionHandler: @escaping (Bool, Error?) -> Void)
```
pixlwave, line 23: Again, could we use the `async` variants of these methods rather than completion handlers?

ElementX/Sources/Services/Notification/Manager/NotificationManager.swift
```diff
@@ -45,22 +46,25 @@ class NotificationManager: NSObject, NotificationManagerProtocol {
                                                    options: [])
         notificationCenter.setNotificationCategories([replyCategory])
         notificationCenter.delegate = self
-        Task {
-            do {
-                let granted = try await notificationCenter.requestAuthorization(options: [.alert, .sound, .badge])
+        notificationCenter.requestAuthorization(options: [.alert, .sound, .badge]) { [weak self] granted, error in
```
pixlwave, line 49: How come this was switched to use the non-async variant?

ElementX/Sources/Services/Notification/Manager/NotificationManager.swift
```diff
@@ -19,11 +19,12 @@ import UIKit
 import UserNotifications
 
 class NotificationManager: NSObject, NotificationManagerProtocol {
-    private let notificationCenter = UNUserNotificationCenter.current()
+    private var notificationCenter: UserNotificationCenterProtocol
```
pixlwave, line 22: We could make `UserNotificationCenterProtocol` only apply to `AnyObject` and then the `var` could remain a `let` here :)

pixlwave (commented):
Nice job, I've added some comments about using async/await more which will tidy up some of the tests by removing the need for expectations :)

ElementX/Sources/Services/Notification/Manager/NotificationManager.swift
paleksandrs: `start` method is not async method, so in my oppinon there is no benefit of using `requestAuthorization` as async. If async is used you need to wrap it in `Task` and because `start` is not async, testing this method is not possible.

ElementX/Sources/Services/Notification/Manager/NotificationManager.swift
paleksandrs: I tried to keep the interface of the class as it was. There is no need for `register` to be async because the caller doesn't need to wait for it. I added the completion because otherwise, it is impossible to test the method (because of the `Task` used).

ElementX/Sources/Services/Notification/Manager/UserNotificationCenterProtocol.swift
paleksandrs: Explained my reasoning here https://github.com/vector-im/element-x-ios/pull/376/files#r1053120656

paleksandrs:
@pixlwave thank you for your comments. I fixed some of them, but some I left as it was. If you feel strongly about changing methods to async, let's discuss it first.

paleksandrs:
@stefanceriu because @pixlwave is away, could you please finish to review?

stefanceriu (approved):
This looks very reasonable to me, nicely done 👍

PR #379, opened 2022-12-19
Fixes #258 - Display an indicator if the network is currently unreachable

### LINKED ISSUE #258: Connectivity status
Handle and inform the user about connectivity issues. Display an indicator if the network is currently unreachable.

stefanceriu:
Screenshot 2022-12-19 at 17 50 41


ElementX/Sources/Other/NetworkMonitor.swift
```diff
@@ -0,0 +1,50 @@
...
+    private let queue: DispatchQueue
+    let reachabilityPublisher: CurrentValueSubject<Bool, Never>
+    
+    var isCurrentConnectionExpensive: Bool {
+        pathMonitor.currentPath.isExpensive
+    }
+    
+    var isCurrentConnectionConstrained: Bool {
+        pathMonitor.currentPath.isConstrained
+    }
```
pixlwave, line 32: Very nice, love that this is so easily available 👏

ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -329,12 +330,12 @@ class AppCoordinator: AppCoordinatorProtocol {
...
             }
-            .store(in: &cancellables)
+            .store(in: &userSessionCancellables)
     }
 
     private func deobserveUserSessionChanges() {
-        cancellables.forEach { $0.cancel() }
-        cancellables.removeAll()
+        userSessionCancellables.forEach { $0.cancel() }
+        userSessionCancellables.removeAll()
```
pixlwave, line 338: I probably made this comment before, but calling `removeAll()` is enough here, `cancel` will get called automatically on `deinit`.

pixlwave (approved):
Looks great 📶

ElementX/Sources/Application/AppCoordinator.swift
stefanceriu: Yep, good point, I went through the whole app and removed these extra cancellations

PR #384, opened 2022-12-21
Add retry decryption encrypted timeline item debug menu option

stefanceriu:


ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -85,8 +85,8 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
                     } else {
                         self?.state.roomListMode = .rooms
                     }
-                } else if totalCount == 0 {
-                    #warning("Empty state but it never happens because SS never goes into live for empty accounts")
+                } else if totalCount == 0 { // Empty state
```
paleksandrs, line 88: As a suggestion. Following clean code principles, I always try to explain myself in code, instead of writing comments, that you need to maintain. You could define a variable like `let isEmptyState = totalCount == 0 ` and use it.

paleksandrs (approved):
looks good to me 👍🏼 just one small suggestion, up to you if you want to use it

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
stefanceriu: Fair point, this was bound for a cleanup 👍  (which I've now done)

PR #395, opened 2023-01-03
Fix stuck timeline pagination because of too many membership events

### LINKED ISSUE #394: Pagination stuck because of too many membership events
If a room contains too many events that we do not display like membership events, the pagination stops to work. `Matrix HQ` is a good example around Jan, 2nd.

The reason is that the pagination listener receives no diff updates during a pagination cycle in this case. The app stays in pagination in progress mode.

At the app level , we did not have a reliable way to know when a pagination cycle is complete. The FFI layer actually breaks the async workflow defined by the SDK where the `Room.paginateBackwards()` should return only after all the calls of `TimelineListener onUpdate()`. It is actually called before.

This SDK [PR](https://github.com/matrix-org/matrix-rust-sdk/pull/1314) adds more information to `PaginationOutcome` to handle and fix this case.

manuroe:
Fixes #394.
Requires https://github.com/matrix-org/matrix-rust-sdk/pull/1314


stefanceriu (approved):

pixlwave:
Nice! 👏

PR #404, opened 2023-01-05
Storage location change

stefanceriu:
- remove the old restoration token handling
- move session data to the shared container's `Library/Application Support` folder and handle migrations. Also excludes it from iCloud backups. Left a warning in place so we can remove the migration in a couple of releases


pixlwave (approved):
LGTM.

PR #406, opened 2023-01-05
Minor fixes.

pixlwave:
A couple of minor tweaks:
- Use the `links` colour from Compound for links.
- Avoid constantly recomputing the RoomScreen view hierarchy while scrolling.


stefanceriu (approved):
Nice! 👍

PR #407, opened 2023-01-05
Replace completion handlers with async/await.

pixlwave:
Small follow up to #376 which keeps everything using async/await instead of completion handlers.

### DIFF
--- a/UnitTests/Sources/NotificationManager/NotificationManagerTests.swift
+++ b/UnitTests/Sources/NotificationManager/NotificationManagerTests.swift
@@ -33,48 +33,29 @@ final class NotificationManagerTests: XCTestCase {
         sut = NotificationManager(clientProxy: clientProxySpy, notificationCenter: notificationCenter)
     }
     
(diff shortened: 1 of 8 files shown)


UnitTests/Sources/NotificationManager/NotificationManagerTests.swift
```diff
@@ -143,21 +116,18 @@ final class NotificationManagerTests: XCTestCase {
...
         authorizationStatusWasGranted = false
         notificationCenter.requestAuthorizationGrantedReturnValue = true
         sut.delegate = self
         sut.start()
-        let expectation = expectation(description: "Wait for main thread")
-        DispatchQueue.main.async {
-            expectation.fulfill()
-        }
-        waitForExpectations(timeout: 1, handler: nil)
+        try await Task.sleep(for: .milliseconds(100))
```
pixlwave, line 130: A simple `yield` didn't work here but a short sleep should be fine instead.

stefanceriu (approved):
Oh niiice one! 🙌

PR #420, opened 2023-01-06
Remove French Locale from UI tests.

pixlwave:
We haven't really got any value from this locale as the strings aren't particularly long, so it doesn't identify layout issues and just means creating and running tests is a slower process with not much benefit. Also fixes some a build error on the tests and removes some unnecessary sources from the target.

It would be kind of nice if we could eventually run 3 simulators simultaneously each with a different locale but that doesn't appear to be supported by test plans.


stefanceriu (approved):
I agree, lgtm! 👍

PR #431, opened 2023-01-10
Room list fixes

stefanceriu:
* start using custom identifiers for invalided list room entries
* cleanup logic and avatar loading
* re-enable the allRoomsView with a hard 500 room limit and bigger batch size

### DIFF
--- a/ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomCell.swift
+++ b/ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomCell.swift
@@ -24,7 +24,9 @@ struct HomeScreenRoomCell: View {
     
     var body: some View {
         Button {
-            context.send(viewAction: .selectRoom(roomIdentifier: room.id))
+            if let roomId = room.roomId {
+                context.send(viewAction: .selectRoom(roomIdentifier: roomId))
+            }
         } label: {
             HStack(spacing: 16.0) {
                 if let avatar = room.avatar {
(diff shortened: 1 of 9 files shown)


ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomCell.swift
```diff
@@ -55,7 +57,7 @@ struct HomeScreenRoomCell: View {
                                 .lineLimit(2)
                                 .multilineTextAlignment(.leading)
                                 .padding(.top, 2)
-                                .id(lastMessage)
+                                .id(UUID())
```
pixlwave, line 60: What are these `lastMessage` and `timestamp` IDs actually about anyway?? Can they go?

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -86,7 +86,7 @@ class ClientProxy: ClientProxyProtocol {
                 let slidingSyncBuilder = try client.slidingSync().homeserver(url: ServiceLocator.shared.settings.slidingSyncProxyBaseURLString)
 
                 let visibleRoomsView = try SlidingSyncViewBuilder()
-                    .timelineLimit(limit: 30)
+                    .timelineLimit(limit: 20)
```
pixlwave, line 104: I wonder if this should be shared with the RoomProxy constant somehow? Even if we wanted unique values later, it would be nice to define these constants in the same place.

pixlwave (approved):
LGTM 👏

ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomCell.swift
stefanceriu: I use them so that transition animations between new values work properly

stefanceriu: Decided to use the lastMessage and timestampt texts as identifiers

PR #432, opened 2023-01-10
Fix back pagination

pixlwave:
The pagination logic was broken following a recent change for the SDK, this PR uses the new timeline items for the pagination indicator and the room start to update the view state.

Additionally updates some snapshots that were broken following the removal of the `top` section in the timeline, causing minor layout changes.


stefanceriu (approved):
This looks great, so much cleaner 👏

PR #437, opened 2023-01-11
Various fixes

stefanceriu:
* Fix information leaking in RoomSummaryDetail logs
* Prevent crash popups when force quitting the application
* Cleanup crash detection popup display
* Wait for logout confirmation before changing the app state

### DIFF
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -260,6 +260,11 @@ class AppCoordinator: AppCoordinatorProtocol {
         
         deobserveUserSessionChanges()
         
+        guard !isSoftLogout else {
+            stateMachine.processEvent(.completedSigningOut)
+            return
+        }
+        
         Task {
             showLoadingIndicator()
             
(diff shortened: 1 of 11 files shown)


ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -255,21 +260,22 @@ class AppCoordinator: AppCoordinatorProtocol {
...
+            
+            //  regardless of the result, clear user data
+            userSessionStore.logout(userSession: userSession)
+            userSession = nil
+            notificationManager?.delegate = nil
+            notificationManager = nil
+            
+            stateMachine.processEvent(.completedSigningOut)
+            
+            hideLoadingIndicator()
```
pixlwave, line 277: In theory we should still handle `isSoftLogout` separately to this path as you don't want to clear the user data. We should remember their user ID so all they need to enter is their password to continue from where they left off.

pixlwave (commented):
One comment on soft logout, otherwise looks good to me.

ElementX/Sources/Application/AppCoordinator.swift
stefanceriu: Oh I'm so sorry, I had a guard in there that I managed to delete somehow

pixlwave (approved):
Ahhh, that makes sense. LGTM 👍

PR #440, opened 2023-01-11
Move the RoomSummaryProvider id based room updated to the same serial…

stefanceriu:
… dispatch queue as the normal diffing. Cleanup logging and RoomTimelineProvider diffing


pixlwave (approved):
LGTM

PR #446, opened 2023-01-12
Various sliding sync tweaks

stefanceriu:
- [x] refactor the sliding sync setup and make so it's possibly to add views to it later
- [x] implement pop and clear diff support on the room summary provider
- [x] use allRoomsView room summaries for invalidated visibleRoomsView entries
- [x] switch visible range setting from the ScrollViewAdapter to a publisher + debounce
- [ ] ~fix room subscriptions on entering~ currently broken rust side
- [ ] ~fix visibleRoomsView going into live before any networking is done (rust side)~ switched to waiting for the first diff update

### DIFF
--- a/ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
+++ b/ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
@@ -185,7 +185,7 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol 
         
         for (index, summary) in visibleRoomsSummaryProvider.roomListPublisher.value.enumerated() {
             switch summary {
-            case .empty:
+            case .empty, .invalidated:
                 guard let allRoomsRoomSummary = allRoomsSummaryProvider?.roomListPublisher.value[safe: index] else {
                     rooms.append(HomeScreenRoom.placeholder())
                     continue
@@ -200,9 +200,6 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol 
                 }
             case .filled(let details):
                 let room = buildRoom(with: details, invalidated: false)
-                rooms.append(room)
-            case .invalidated(let details):
-                let room = buildRoom(with: details, invalidated: true)
                 rooms.append(room)
             }
         }
--- a/ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
+++ b/ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
@@ -20,10 +20,16 @@ import SwiftUI
 typealias HomeScreenViewModelType = StateStoreViewModel<HomeScreenViewState, HomeScreenViewAction>
 
 class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol {
+    enum Constants {
+        static let slidingWindowBoundsPadding = 5
+    }
+    
     private let userSession: UserSessionProtocol
     private let visibleRoomsSummaryProvider: RoomSummaryProviderProtocol?
     private let allRoomsSummaryProvider: RoomSummaryProviderProtocol?
     private let attributedStringBuilder: AttributedStringBuilderProtocol
+    
+    private let visibleItemRangePublisher = CurrentValueSubject<Range<Int>, Never>(0..<0)
     
     var callback: ((HomeScreenViewModelAction) -> Void)?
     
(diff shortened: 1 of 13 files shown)


ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
```diff
@@ -191,6 +184,22 @@ struct HomeScreen: View {
...
+        
+        guard !result.isEmpty else {
+            return
+        }
+        
+        guard let firstIndex = result.first, let lastIndex = result.last else {
+            return
+        }
+        
+        context.send(viewAction: .updatedVisibleItemRange(firstIndex..<lastIndex))
```
pixlwave, line 201: Maybe its me, but I'd be tempted to put the throttle on the view side to avoid sending lots of actions that will get filtered. Although that might not be clean separation, not sure?

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -297,6 +233,116 @@ class ClientProxy: ClientProxyProtocol {
...
+            registerSlidingSyncView(visibleRoomsSlidingSyncView)
+            
+        } catch {
+            MXLog.error("Failed building sliding sync with error: \(error)")
+        }
+    }
+    
+    // swiftlint:disable:next function_body_length
+    private func configureSlidingSyncViews(slidingSync: SlidingSyncProtocol) {
+        let requiredState = [RequiredState(key: "m.room.avatar", value: ""),
```
pixlwave, line 269: Maybe this should have a `guard visible/allRoomsProvider == nil` at the start for safety? Equally might not matter, I'm just wondering about if we ended up with multiple views `sink`'d things could go crazy.

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -216,6 +217,18 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
             for (index, value) in values.enumerated() {
                 changes.append(.insert(offset: index, element: buildSummaryForRoomListEntry(value), associatedWith: nil))
             }
+        case .clear:
+            MXLog.verbose("Clear all items, current total count: \(rooms.count)")
+            for (index, value) in rooms.enumerated() {
+                changes.append(.remove(offset: index, element: value, associatedWith: nil))
+            }
```
pixlwave, line 224: Stupid question perhaps, but should this be on `rooms.reversed().enumerated()` otherwise you're removing the index `count - 1` by the time there's only 1 item left in the array?

pixlwave (approved):
LGTM

ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
stefanceriu: I though about that but I don't have access to the view from the view model ¯\_(ツ)_/¯

ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu: Yeah, good idea. Rebuilding these will basically break everything else

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
stefanceriu: I think the CollectionDifference takes care of making this happen the right way, it does for the rest as well

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -53,6 +59,14 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
             }
             .store(in: &cancellables)
         
+        visibleItemRangePublisher
+            .debounce(for: 0.1, scheduler: RunLoop.main)
```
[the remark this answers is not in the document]
stefanceriu: Talked about it offline, we'll go with this for now until we know for sure we want to make more requests

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
pixlwave: Nice, TIL 😎

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -236,8 +232,7 @@ class ClientProxy: ClientProxyProtocol {
     // MARK: Private
     
     private func restartSync() {
-        slidingSyncObserverToken?.cancel()
-        slidingSync?.setObserver(observer: WeakClientProxyWrapper(clientProxy: self))
+        stopSync()
         slidingSyncObserverToken = slidingSync?.sync()
```
pixlwave, line 236: At this point I think we could simply do the following right?
```
stopSync()
startSync()
```

Issue #418, opened 2023-01-06
Cannot swipe to leave the media fullscreen view


manuroe:
Instead, we have to tap the `Back` button in the top navigation bar. This is kind of old school.

pixlwave:
I looked into this with #447, however using `QLPreviewController` (which gives us a head-start with a *lot* of functionality for free) wrapped in an `NSViewControllerRepresentable` struct doesn't provide us with the "automatic" swipe gestures  that it provides within a UIKit hierarchy. I tested this with the controller wrapped in both `UINavigationController` (it updates the toolbars automatically in this one) and within `NavigationStack` and didn't get anything out of it.

The decision for now was to deprioritise the gestures so this issue should probably have the `TF` label removed?

PR #447, opened 2023-01-12
QuickLook media.

pixlwave:
Initial result as a of #418, but doesn't actually close it. For now we're just switching to use QuickLook for videos too (as the player is much nicer, and you get a share sheet for free) and using a full screen presentation of this rather than pushing it onto the details view stack. PR also includes a couple of minor tweaks to DesignKit.

Videos start like this, and a tap will make it go full screen.

Simulator Screen Shot - iPhone 14 - 2023-01-12 at 15 42 49


ElementX/Sources/Application/Navigation/NavigationCoordinators.swift
```diff
@@ -301,6 +336,11 @@ private struct NavigationSplitCoordinatorView: View {
         // through the NavigationSplitCoordinator as well.
         .sheet(item: $navigationSplitCoordinator.sheetModule) { module in
             module.coordinator.toPresentable()
+                .tint(.element.accent)
```
pixlwave, line 339: Added a tint here as it isn't inherited by sheets, so this avoids us needed to fix per screen.

PR #452, opened 2023-01-15
Room member details cleanup

stefanceriu:
* made it so members aren't loaded twice
* switch from a form to a lazyvstack for performance reasons
* rename everything from RoomMembers* to RoomMemberDetails*


pixlwave (approved):
I reviewed this one but forgot to press the button 🤦‍♂️

LGTM 😎

PR #454, opened 2023-01-16
chore: Setup logging for the crypto crate

poljar:
Not sure if this is fully correct and compiles, but it seemed sensible looking at the rest of the code.


stefanceriu (approved):
lgtm 👍

PR #459, opened 2023-01-17
Break out DesignKit into its own sub-package

pixlwave:
We'll be using https://github.com/vector-im/compound-ios eventually so this tidies up the Xcode project dependencies in the mean time.


stefanceriu (approved):
🚢 IT! 🚀

PR #460, opened 2023-01-17
Fix unit and UI tests.

pixlwave:
This PR does the following:
- Fix race condition on session verification tests.
- Fix clear button on text field being blocked by a tap gesture.
- Add a timeout to UI test signalling to stop tests from hanging.
- Use snapshot testing for login screen tests.


UnitTests/Sources/SessionVerificationViewModelTests.swift
```diff
@@ -74,7 +74,7 @@ class SessionVerificationViewModelTests: XCTestCase {
         let waitForAcceptance = XCTestExpectation(description: "Wait for acceptance")
         
         let cancellable = sessionVerificationController.callbacks
-            .debounce(for: .seconds(2.0), scheduler: DispatchQueue.main)
+            .delay(for: .seconds(0.1), scheduler: DispatchQueue.main)
```
stefanceriu, line 77: I would still add a comment here, I'm sure we'll forget again why we need to delay (combine subscription callbacks race conditions between the view model and the test)

stefanceriu (approved):
Fabulous 🤩

Issue #414, opened 2023-01-05
Rooms stay unread after visiting them


manuroe:
The unread state in the room list \(bolded last message and circle\) should be reset when coming to the room list.

manuroe:
- [ ] We will need to create the rust API (should be not hard)
- [ ] Rust SDK does not provide updates when reading a room from another client updating the marker.

pixlwave:
I'm going to mark this as complete. We send read markers on entering/exiting the room, and there's a Rust issue for the fully read markers not tracking your other clients.

PR #462, opened 2023-01-17
Mark rooms as read on push and pop.

pixlwave:
We don't have visibility into timeline item read status from Rust yet, so can't simply do this as items appear on screen.

This will look a bit strange on iPad/Mac when messages come in to a room that is open as the room will remain unread in the sidebar until you tap a different room.

Additionally this PR simplifies `RoomTimelineProvider` by removing it as a middleman for calls to `RoomProxy from `RoomTimelineController`.

Note: Requires a new release of the Rust SDK

Part of #414 - Maybe closes it depending on how we feel about the split view behaviour (and pending on https://github.com/matrix-org/matrix-rust-sdk/issues/1363 for fully read marker updates.)


ElementX/Sources/Services/Room/RoomProxy.swift
```diff
@@ -156,7 +176,7 @@ class RoomProxy: RoomProxyProtocol {
         
         let transactionId = genTransactionId()
         
-        return await Task.dispatch(on: .global()) {
+        return await Task.dispatch(on: serialDispatchQueue) {
```
pixlwave, line 179: This one was wrong, as we want the order of messages to be preserved. Should maybe do the same with edits, but as we don't have local echoes of those yet (or the ability to edit a local echo) that seems less important for now.

ElementX/Sources/Services/Timeline/RoomTimelineProvider.swift
```diff
@@ -61,79 +61,6 @@ class RoomTimelineProvider: RoomTimelineProviderProtocol {
         }
     }
     
-    func paginateBackwards(requestSize: UInt, untilNumberOfItems: UInt) async -> Result<Void, RoomTimelineProviderError> {
```
stefanceriu, line 64: 👏

ElementX/Sources/Services/Room/RoomProxyProtocol.swift
```diff
@@ -58,6 +60,8 @@ protocol RoomProxyProtocol {
     
     func paginateBackwards(requestSize: UInt, untilNumberOfItems: UInt) async -> Result<Void, RoomProxyError>
     
+    func sendReadReceipt(for eventID: String) async -> Result<Void, RoomProxyError>
+    
     func sendMessage(_ message: String, inReplyToEventId: String?) async -> Result<Void, RoomProxyError>
     
     func sendReaction(_ reaction: String, for eventId: String) async -> Result<Void, RoomProxyError>
```
stefanceriu, line 67: We should probably agree on consistent naming for all of these:
`sendMessage inReplyTo`, `editMessage of`, `sendReaction for`, `debugDescriptionFor` but this probably needs us to sync offline

ElementX/Sources/Screens/RoomScreen/RoomScreenCoordinator.swift
```diff
@@ -64,6 +64,7 @@ final class RoomScreenCoordinator: CoordinatorProtocol {
     }
     
     func stop() {
+        viewModel?.context.send(viewAction: .markRoomAsRead)
```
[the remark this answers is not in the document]
pixlwave: If more messages came in whilst the room was open, then you'll return to the home screen and the room will still be unread. Caveat: there's no guarantee you saw those messages though so its not perfect. Ultimately when we can detect if visible items are read this will make the implementation a lot better.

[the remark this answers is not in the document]
stefanceriu: Oh okay, yes, that makes perfect sense to me 👍

ElementX/Sources/Services/Room/RoomProxy.swift
```diff
@@ -148,6 +152,22 @@ class RoomProxy: RoomProxyProtocol {
         }
     }
     
+    func sendReadReceipt(for eventID: String) async -> Result<Void, RoomProxyError> {
+        sendMessageBackgroundTask = backgroundTaskService.startBackgroundTask(withName: "SendMessage", isReusable: true)
```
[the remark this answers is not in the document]
pixlwave: Discussed this one and we're going to put all the send events on the same serial queue and task.

ElementX/Sources/Services/Room/RoomProxyProtocol.swift
pixlwave: Updated 👍

PR #463, opened 2023-01-18
Timeline item sender profiles

stefanceriu:
This PR adopts an initial version of the timeline items sender profiles i.e. resolved display names and avatar URLs coming directly in the event timeline items. Unfortunately this isn't quite reliable yet but in an attempt to unblock sdk development I decided to go ahead with it.
This PR also rewrites our URL handling and removes all stringly typed URLs from throughout the app


ElementX/Sources/Services/Timeline/TimelineItemProxy.swift
```diff
@@ -98,6 +98,18 @@ struct EventTimelineItemProxy: CustomDebugStringConvertible {
...
+        item.senderProfile().displayName
+    }
+    
+    var senderAvatarURL: URL? {
+        guard let urlString = item.senderProfile().avatarUrl else {
+            return nil
+        }
+        
+        return URL(string: urlString)
+    }
```
pixlwave, line 112: Could we collect these together into a
```swift
struct Sender: Identifiable {
    let id: String
    let displayName: String?
    let avatarURL: URL?
}
```
and just have a single method on the proxy?

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemFactory.swift
```diff
@@ -19,66 +19,69 @@ import UIKit
...
     // swiftlint:disable:next cyclomatic_complexity
     func buildTimelineItemFor(eventItemProxy: EventTimelineItemProxy,
                               groupState: TimelineItemGroupState) -> RoomTimelineItemProtocol {
-        let displayName = roomProxy.displayNameForUserId(eventItemProxy.sender)
-        let avatarURL = roomProxy.avatarURLStringForUserId(eventItemProxy.sender)
-        let avatarImage = mediaProvider.imageFromURLString(avatarURL, avatarSize: .user(on: .timeline))
+        var avatarImage: UIImage?
+        if let senderAvatarURL = eventItemProxy.senderAvatarURL {
+            avatarImage = mediaProvider.imageFromURL(senderAvatarURL, avatarSize: .user(on: .timeline))
+        }
```
pixlwave, line 41: Another `flatMap` - please do just ignore these if you don't want to do it.

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemFactory.swift
```diff
@@ -110,17 +113,21 @@ struct RoomTimelineItemFactory: RoomTimelineItemFactoryProtocol {
...
         let width = imageInfo.width.map(CGFloat.init)
         let height = imageInfo.height.map(CGFloat.init)
         if let width, let height {
             aspectRatio = width / height
         }
+        
+        var image: UIImage?
+        if let url = URL(string: imageURLString) {
+            image = mediaProvider.imageFromURL(url)
+        }
```
pixlwave, line 130: Do we want to load in the factory rather than on appear?

pixlwave (approved):
Looks great thanks for the tidy-up, although I've read the word URL enough to last me for a lifetime now!
Sorry about lots of comments, they're all minor.

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemFactory.swift
pixlwave: My bad this is loading from the cache 🤦‍♂️

stefanceriu (commented):
This looks soo much cleaner, thank you for taking care of it 👏

PR #470, opened 2023-01-18
Sliding sync tweaks

stefanceriu:
* fix the splash screen not being shown after a session restoration error
* fix the cold cache for the visible rooms view
* remove custom room identifiers for invalid rooms (we still have duplicates on filled rooms)
* fixed the visible range debouncer, using the main run loop was wrong
* add more tracing

### DIFF
--- a/ElementX/Sources/Services/Client/ClientProxy.swift
+++ b/ElementX/Sources/Services/Client/ClientProxy.swift
@@ -250,7 +250,19 @@ class ClientProxy: ClientProxyProtocol {
         do {
             let slidingSyncBuilder = try client.slidingSync().homeserver(url: ServiceLocator.shared.settings.slidingSyncProxyBaseURLString)
             
+            // Build the visibleRoomsSlidingSyncView here so that it can take advantage of the SS builder cold cache
+            // We will still register the allRoomsSlidingSyncView later, and than will have no cache
+            let visibleRoomsView = try SlidingSyncViewBuilder()
+                .timelineLimit(limit: 20)
+                .requiredState(requiredState: slidingSyncRequiredState)
+                .filters(filters: slidingSyncFilters)
+                .name(name: "CurrentlyVisibleRooms")
+                .syncMode(mode: .selective)
+                .addRange(from: 0, to: 20)
+                .build()
+            
             let slidingSync = try slidingSyncBuilder
+                .addView(v: visibleRoomsView)
                 .withCommonExtensions()
                 .coldCache(name: "ElementX")
                 .build()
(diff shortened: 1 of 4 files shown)


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -253,103 +275,91 @@ class ClientProxy: ClientProxyProtocol {
...
         } catch {
-            MXLog.error("Failed building sliding sync views with error: \(error)")
+            MXLog.error("Failed building the all rooms sliding sync view with error: \(error)")
         }
     }
     
+    private lazy var slidingSyncRequiredState: [RequiredState] = {
+        [RequiredState(key: "m.room.avatar", value: ""),
+         RequiredState(key: "m.room.encryption", value: "")]
+    }()
```
pixlwave, line 338: This should be fine
```suggestion
    private lazy var slidingSyncRequiredState = [
        RequiredState(key: "m.room.avatar", value: ""),
        RequiredState(key: "m.room.encryption", value: "")
    ]
```

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -253,103 +275,91 @@ class ClientProxy: ClientProxyProtocol {
...
+                                      spaces: [],
+                                      isEncrypted: nil,
+                                      isInvite: false,
+                                      isTombstoned: false,
+                                      roomTypes: [],
+                                      notRoomTypes: ["m.space"],
+                                      roomNameLike: nil,
+                                      tags: [],
+                                      notTags: [])
+    }()
```
pixlwave, line 351: Same here (indentation would need fixing)
```suggestion
    private lazy var slidingSyncFilters = SlidingSyncRequestListFilters(isDm: nil,
                                                                        spaces: [],
                                                                        isEncrypted: nil,
                                                                        isInvite: false,
                                                                        isTombstoned: false,
                                                                        roomTypes: [],
                                                                        notRoomTypes: ["m.space"],
                                                                        roomNameLike: nil,
                                                                        tags: [],
                                                                        notTags: [])
```

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -253,103 +275,91 @@ class ClientProxy: ClientProxyProtocol {
...
-            configureSlidingSyncViews(slidingSync: slidingSync)
-            
-            guard let visibleRoomsSlidingSyncView else {
-                MXLog.error("Visible rooms sliding sync view unavailable")
-                return
-            }
-            
-            registerSlidingSyncView(visibleRoomsSlidingSyncView)
-            
+            buildAndConfigureVisibleRoomsSlidingSyncView(slidingSync: slidingSync, visibleRoomsView: visibleRoomsView)
```
pixlwave, line 278: Probably should be called `configureVisibleRoomsSlidingSyncView` given this one doesn't build the view?

pixlwave (approved):
LGTM 😎

ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu: I bounced between the 2 as it still builds the `SlidingSyncViewProxy` and `RoomSummaryProvider` so I figured it's best to user more words than less 😅

PR #473, opened 2023-01-19
Render state items in the timeline and as the last message on the home screen.

pixlwave:
- Adds `RoomStateStringBuilder` which will build the right localised strings for state changes.
- Uses this state string builder inside of the timeline item factory.
- Adds `RoomEventStringBuilder` which will build a localised string for showing events as attributed strings in the room list. This allows us to handle all event types and uses `RoomStateStringBuilder` for state events.
- Adds a timeline item and view to show state events in the timeline.

| Room list | Timeline |
| - | - |
| RoomSummaryState | TimelineState |

### DIFF
--- a/ElementX/Sources/Services/Room/RoomSummary/RoomEventStringBuilder.swift
+++ b/ElementX/Sources/Services/Room/RoomSummary/RoomEventStringBuilder.swift
@@ -16,15 +16,15 @@
 
 import Foundation
 
-struct EventTimelineItemSummaryFactory {
-    private let roomStateTimelineItemFactory: RoomStateTimelineItemFactory
+struct RoomEventStringBuilder {
+    private let roomStateStringBuilder: RoomStateStringBuilder
     
     /// The Matrix ID of the current user.
     private let userID: String
     
-    init(userID: String, roomStateTimelineItemFactory: RoomStateTimelineItemFactory) {
+    init(userID: String, roomStateStringBuilder: RoomStateStringBuilder) {
         self.userID = userID
-        self.roomStateTimelineItemFactory = roomStateTimelineItemFactory
+        self.roomStateStringBuilder = roomStateStringBuilder
     }
     
     // swiftlint:disable:next cyclomatic_complexity
@@ -61,13 +61,13 @@ struct EventTimelineItemSummaryFactory {
                 message = messageContent.body()
             }
             return prefix(message, with: sender)
-        case .state(let stateKey, let content):
-            return roomStateTimelineItemFactory
-                .textForOtherState(content, stateKey: stateKey, sender: sender, isOutgoing: isOutgoing)
+        case .state(let stateKey, let state):
+            return roomStateStringBuilder
+                .buildString(for: state, stateKey: stateKey, sender: sender, isOutgoing: isOutgoing)
                 .map(AttributedString.init)
         case .roomMembership(userId: let userID, change: let change):
-            return roomStateTimelineItemFactory
-                .textForMembershipChange(change, member: userID, sender: sender, isOutgoing: isOutgoing)
+            return roomStateStringBuilder
+                .buildString(for: change, member: userID, sender: sender, isOutgoing: isOutgoing)
                 .map(AttributedString.init)
         }
     }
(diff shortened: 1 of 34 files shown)


ElementX/Sources/Services/Timeline/TimelineItems/RoomStateStringBuilder.swift
```diff
@@ -0,0 +1,200 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import MatrixRustSDK
+import UIKit
+
+struct RoomStateStringBuilder {
```
stefanceriu, line 20: Should we call this `RoomStateEventStringBuilder` to match the `RoomEventStringBuilder` one?

ElementX/Sources/Services/Room/RoomSummary/RoomEventStringBuilder.swift
```diff
@@ -0,0 +1,84 @@
...
+// limitations under the License.
+//
+
+import Foundation
+
+struct RoomEventStringBuilder {
+    private let roomStateStringBuilder: RoomStateStringBuilder
+    
+    /// The Matrix ID of the current user.
+    private let userID: String
```
stefanceriu, line 23: I think this is unused

ElementX/Sources/Services/Timeline/TimelineItems/RoomStateStringBuilder.swift
```diff
@@ -0,0 +1,200 @@
...
+            break
+        case .roomTombstone: // Handle as a virtual timeline item with a link to the upgraded room.
+            break
+        case .spaceChild, .spaceParent: // Users shouldn't see the timeline of a Space.
+            break
+        case .custom: // Won't provide actionable information to the user.
+            break
+        }
+        
+        MXLog.verbose("Filtering timeline item for state: \(state)")
```
stefanceriu, line 197: I would promote this to info but I guess we need to look out for leaking PII

ElementX/Sources/Services/Timeline/TimelineItems/RoomStateStringBuilder.swift
```diff
@@ -0,0 +1,200 @@
...
+//
+
+import MatrixRustSDK
+import UIKit
+
+struct RoomStateStringBuilder {
+    let userID: String
+    
+    // swiftlint:disable:next cyclomatic_complexity function_body_length
+    func buildString(for change: MembershipChange, member: String, sender: TimelineItemSender, isOutgoing: Bool) -> String? {
```
stefanceriu, line 24: I wonder if these methods should return attributed strings and bold the member display name like we do for normal messages

stefanceriu (approved):
This is great, very nicely done! 👏

ElementX/Sources/Services/Room/RoomSummary/RoomEventStringBuilder.swift
pixlwave: Ah yes, I forgot this when I allowed the state builder to be injected.

ElementX/Sources/Services/Timeline/TimelineItems/RoomStateStringBuilder.swift
pixlwave: Not sure, I think it would be hard to reason about when to do it e.g.

> @alice:matrix.org changed their name from Alice to Bob

or 

> Alice invited Bob to join the room

I think unless Design highlights it in the future, we should simply treat these as low priority notices that create lots of noise.

ElementX/Sources/Services/Timeline/TimelineItems/RoomStateStringBuilder.swift
pixlwave: Yeah, plus we can clearly see above which are the state events we filter on, I'm not sure what it would add to the logs, and I was thinking it could become pretty noisy e.g. during spam attacks with ACL events. Will leave for now, can update later.

PR #474, opened 2023-01-19
Various performance tweaks

stefanceriu:
* Store and reuse room list placeholder avatars and last messages
* Cache and reuse HomeScreenRooms
* Reduce RoomSummaryProvider diff collection time
* Promote more logs to info
* Slighty tweak RustTracing to make it easier to configure
* Move TimelineProvider and RoomTimelineController item processing to background queues
* Prevent the timeline from stopping an ogoing decelerating scroll when starting backpagination

### DIFF
--- a/ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
+++ b/ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
@@ -31,6 +31,8 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol 
     
     private let visibleItemRangePublisher = CurrentValueSubject<Range<Int>, Never>(0..<0)
     
+    private var roomsForIdentifiers = [String: HomeScreenRoom]()
+    
     var callback: ((HomeScreenViewModelAction) -> Void)?
     
     // MARK: - Setup
@@ -178,7 +180,7 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol 
     // MARK: - Private
     
     private func loadDataForRoomIdentifier(_ identifier: String) {
-        guard let room = state.rooms.first(where: { $0.roomId == identifier }),
+        guard let room = roomsForIdentifiers[identifier],
               room.avatar == nil,
               let avatarURL = room.avatarURL else {
             return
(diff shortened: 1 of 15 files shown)


ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -105,7 +105,7 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
         }
 
         guard let newSummaries = rooms.applying(diff) else {
-            MXLog.error("Failed applying diff: \(diff)")
+            MXLog.info("Failed applying diff: \(diff)")
```
pixlwave, line 108: Just double checking you intended to drop this down to `info`

ElementX/Sources/Services/Timeline/RoomTimelineProvider.swift
```diff
@@ -47,8 +47,8 @@ class RoomTimelineProvider: RoomTimelineProviderProtocol {
             
             roomTimelineListener
                 .itemsUpdatePublisher
-                .receive(on: DispatchQueue.main)
-                .sink { [weak self] in self?.updateItemsWithDiff($0) }
+                .collect(.byTime(DispatchQueue.global(), 0.025))
```
pixlwave, line 50: Are you sure you want to put this onto a concurrent queue? Couldn't we end up applying the diffs in the wrong order?

UnitTests/Sources/TracingConfigurationTests.swift
```diff
@@ -19,32 +19,18 @@ import XCTest
...
-        XCTAssertTrue(filterComponents.contains("matrix_sdk_base::sliding_sync=warn"))
+    func testConfiguration() {
+        let configuration = TracingConfiguration(overrides: [.common: .trace,
+                                                             .matrix_sdk_base_sliding_sync: .error,
+                                                             .matrix_sdk_http_client: .warn,
+                                                             .matrix_sdk_crypto: .info,
+                                                             .hyper: .debug])
+        
+        let filterComponents = configuration.filter.components(separatedBy: ",")
+        XCTAssertTrue(filterComponents.contains("trace"))
```
pixlwave, line 30: Untested and possibly a reason why I didn't do this, but seems like it should always be at the start of the string right?
```suggestion
        XCTAssertTrue(filterComponents.hasPrefix("trace"))
```

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -235,24 +239,34 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
...
     
     private func buildRoom(with details: RoomSummaryDetails) -> HomeScreenRoom {
-        let avatarImage = details.avatarURL.flatMap { userSession.mediaProvider.imageFromURL($0, avatarSize: .room(on: .home)) }
+        var room: HomeScreenRoom! = roomsForIdentifiers[details.id]
+        
+        if room == nil {
+            room = HomeScreenRoom(id: details.id,
+                                  roomId: details.id,
+                                  avatarURL: details.avatarURL)
+        }
```
pixlwave, line 253: Totally unimportant, but this is simpler (although it looks a mess with GH formatting)
```suggestion
        var room: HomeScreenRoom! = roomsForIdentifiers[details.id] ?? HomeScreenRoom(id: details.id,
                                                                                      roomId: details.id,
                                                                                      avatarURL: details.avatarURL)
```

pixlwave (approved):
LGTM.

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
stefanceriu: I think I prefer mine better as it's more in your face

ElementX/Sources/Services/Timeline/RoomTimelineProvider.swift
stefanceriu: Absolutely, good catch 👍

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
stefanceriu: Nop, total error on my part

PR #478, opened 2023-01-20
Improvements to state items

pixlwave:
Minor changes to #473:
- Remove context menu (and fix long press when there are no items).
- Add membership change tests.
- Fix new/old name incorrect ordering.
- Fix message grouping when it comes before/after a state item from the same user.
- Don't show the reactions picker when double tapping virtual or state items.

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
+++ b/ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
@@ -32,11 +32,11 @@ enum RoomScreenComposerMode: Equatable {
 
 enum RoomScreenViewAction {
     case displayRoomDetails
-    case displayEmojiPicker(itemId: String)
     case paginateBackwards
     case itemAppeared(id: String)
     case itemDisappeared(id: String)
     case itemTapped(id: String)
+    case itemDoubleTapped(id: String)
     case linkClicked(url: URL)
     case sendMessage
     case sendReaction(key: String, eventID: String)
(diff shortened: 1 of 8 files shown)


ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
```diff
@@ -32,11 +32,11 @@ enum RoomScreenComposerMode: Equatable {
 
 enum RoomScreenViewAction {
     case displayRoomDetails
-    case displayEmojiPicker(itemId: String)
     case paginateBackwards
     case itemAppeared(id: String)
     case itemDisappeared(id: String)
     case itemTapped(id: String)
+    case itemDoubleTapped(id: String)
```
stefanceriu, line 39: 👍

ElementX/Sources/Services/Timeline/TimelineItemProxy.swift
```diff
@@ -83,6 +89,10 @@ struct EventTimelineItemProxy: CustomDebugStringConvertible {
         content.asMessage() != nil
     }
     
+    var isState: Bool {
```
stefanceriu, line 92: `isState` sounds a bit too generic, can we call it `isRoomState` instead?

stefanceriu (approved):
👍

ElementX/Sources/Services/Timeline/TimelineItemProxy.swift
pixlwave: Oh yeah good point 👍

PR #483, opened 2023-01-23
Media loading flow changes

stefanceriu:
This PR introduces a new way to load media
* media requests are coalesced for the same item and only one actual request fires
* media can be loaded directly from views through the new imageProvider view context parameter
* there's a new LoadableImage SwiftUI component that handles loading transparently

Todo:
- [x] Reading images from the in-memory cache on init doesn't always seem to work
- [x] Fetching images after setting a view's URL property doesn't work (current user avatar on the home screen or settings)
- [x] We need to merge all avatars into one single component


ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemFactory.swift
```diff
@@ -38,64 +38,61 @@ struct RoomTimelineItemFactory: RoomTimelineItemFactoryProtocol {
...
+            return buildEncryptedTimelineItem(eventItemProxy, encryptedMessage, isOutgoing, groupState)
         case .redactedMessage:
-            return buildRedactedTimelineItem(eventItemProxy, sender, isOutgoing, groupState)
-        case .sticker(let body, let imageInfo, let url):
-            return buildStickerTimelineItem(eventItemProxy, body, imageInfo, url, sender, isOutgoing, groupState)
+            return buildRedactedTimelineItem(eventItemProxy, isOutgoing, groupState)
+        case .sticker(let body, let imageInfo, let urlString):
+            guard let url = URL(string: urlString) else {
+                MXLog.error("Invalid sticker url string: \(urlString)")
+                return nil
```
pixlwave, line 51: Not for now, but might be nice to add a todo here to use `buildUnsupportedTimelineItem` so that we have the same behaviour here as we do with decoding invalid events in the SDK. Could maybe do the same for media too but I think there's more nuance in those events.

ElementX/Sources/Services/Media/MediaLoader.swift
```diff
@@ -0,0 +1,84 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Combine
+import Foundation
+import MatrixRustSDK
+import UIKit
+
+private final class MediaRequest: @unchecked Sendable {
```
pixlwave, line 22: Everything compiles just fine without the `Sendable` conformance anyway?
```suggestion
private final class MediaRequest {
```

ElementX/Sources/Services/Media/MediaLoader.swift
```diff
@@ -0,0 +1,84 @@
...
+    private let clientQueue: DispatchQueue
+    private var ongoingRequests = [MediaSourceProxy: MediaRequest]()
+
+    init(client: ClientProtocol,
+         clientQueue: DispatchQueue = .global()) {
+        self.client = client
+        self.clientQueue = clientQueue
+    }
+
+    func mediaSourceForURL(_ url: URL) -> MediaSourceProxy {
```
pixlwave, line 37: Seeing this, I'm wondering if we could keep `MediaSourceProxy`'s use internal to `MediaLoader` and then use `URL` everywhere in our types? Feels like theres a load of redundancy given you construct them with purely a URL.

Probably a bigger question for another PR though.

ElementX/Sources/Other/SwiftUI/Views/LoadableImage.swift
```diff
@@ -0,0 +1,104 @@
...
+            }
+        }
+        
+        ZStack {
+            if let image = image {
+                transformer(
+                    Image(uiImage: image)
+                        .resizable()
+                )
+            } else if let blurhash = blurhash,
```
pixlwave, line 87:
```suggestion
            if let image {
                transformer(
                    Image(uiImage: image)
                        .resizable()
                )
            } else if let blurhash,
```

ElementX/Sources/Other/SwiftUI/Views/LoadableImage.swift
```diff
@@ -0,0 +1,104 @@
...
+                  mediaSource: mediaSource,
+                  blurhash: blurhash,
+                  avatarSize: avatarSize,
+                  transformer: transformer,
+                  placeholder: placeholder)
+    }
+    
+    var body: some View {
+        let _ = Task {
+            guard image == nil, let mediaSource else { return }
```
pixlwave, line 74: At some stage we should probably not `guard image == nil` instead checking the URL of the cached image so that we support URL changes live on screen, but thats for another day.

pixlwave (approved):
This is such a satisfying tidy up to the codebase, stoked to have this available. `LoadableAvatarImage` is such a nice reusable abstraction everywhere 🙌

Sorry loads of comments, but none of them are major.

ElementX/Sources/Screens/RoomDetails/RoomDetailsViewModel.swift
```diff
@@ -19,29 +19,25 @@ import SwiftUI
...
-                                           roomTitle: roomProxy.displayName ?? roomProxy.name ?? "Unknown Room",
-                                           roomTopic: roomProxy.topic,
+                                           title: roomProxy.displayName ?? roomProxy.name ?? "Unknown Room",
+                                           topic: roomProxy.topic,
+                                           avatarURL: roomProxy.avatarURL,
                                            members: [],
-                                           bindings: .init()))
-
+                                           bindings: .init()),
+                   imageProvider: mediaProvider)
```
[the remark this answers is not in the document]
pixlwave: Or not, its the parameter of the outer method, not the inner one 🤦‍♂️

ElementX/Sources/Other/SwiftUI/Views/LoadableAvatarImage.swift
```diff
@@ -0,0 +1,52 @@
...
+    
+    init(imageProvider: ImageProviderProtocol?, url: URL?, avatarSize: AvatarSize, text: String, contentID: String?) {
+        self.imageProvider = imageProvider
+        self.url = url
+        self.avatarSize = avatarSize
+        self.text = text
+        self.contentID = contentID
+        
+        _frameSize = ScaledMetric(wrappedValue: avatarSize.value)
+    }
```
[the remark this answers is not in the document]
pixlwave: Hmmm ok, this wasn't as simple as I thought due to the scaled metric, leaving the init in place 👍

ElementX/Sources/Other/SwiftUI/Views/LoadableImage.swift
```diff
@@ -0,0 +1,104 @@
...
+         transformer: @escaping (Image) -> TransformerView = { $0 },
+         placeholder: @escaping () -> PlaceholderView) {
+        self.imageProvider = imageProvider
+        self.mediaSource = mediaSource
+        self.blurhash = blurhash
+        self.avatarSize = avatarSize
+        
+        self.transformer = transformer
+        self.placeholder = placeholder
+    }
```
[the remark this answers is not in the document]
pixlwave: Again not as simple as I thought due to the default value of `transformer` and its generic type. Leaving the init here too.

PR #493, opened 2023-01-26
Sliding sync tweaks

stefanceriu:
* Stop updating room list items based on sliding sync identifier changes
* Remove 500 room hard limit from the all rooms view
* Hardcode server versions
* Drop the timeline limit to 1 on the visible rooms view

### DIFF
--- a/ElementX.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/ElementX.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -86,8 +86,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/matrix-org/matrix-rust-components-swift",
       "state" : {
-        "revision" : "a119a8f16bbe3adc34cfdf2e092886d44e01705a",
-        "version" : "1.0.32-alpha"
+        "revision" : "2d702d0d52805e4f81924507b39ad81c8a74a63f",
+        "version" : "1.0.33-alpha"
       }
     },
     {
--- a/ElementX/Sources/Services/Client/ClientProxy.swift
+++ b/ElementX/Sources/Services/Client/ClientProxy.swift
@@ -19,7 +19,7 @@ import Foundation
 import MatrixRustSDK
 import UIKit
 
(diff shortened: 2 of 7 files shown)


ElementX.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -86,8 +86,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/matrix-org/matrix-rust-components-swift",
       "state" : {
-        "revision" : "a119a8f16bbe3adc34cfdf2e092886d44e01705a",
-        "version" : "1.0.32-alpha"
+        "revision" : "2d702d0d52805e4f81924507b39ad81c8a74a63f",
+        "version" : "1.0.33-alpha"
```
pixlwave, line 90: Was starting to wonder if I'd forgot to check this in 🙏

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -263,12 +252,13 @@ class ClientProxy: ClientProxyProtocol {
...
             // We will still register the allRoomsSlidingSyncView later, and than will have no cache
             let visibleRoomsView = try SlidingSyncViewBuilder()
-                .timelineLimit(limit: 20)
+                .timelineLimit(limit: 1)
                 .requiredState(requiredState: slidingSyncRequiredState)
                 .filters(filters: slidingSyncFilters)
                 .name(name: "CurrentlyVisibleRooms")
                 .syncMode(mode: .selective)
                 .addRange(from: 0, to: 20)
+                .sendUpdatesForItems(enable: true)
```
pixlwave, line 261: Seems like this should be the default to me, but I guess that's a question for Rust.

pixlwave (approved):
Looks much cleaner 🙌

ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu: For me too, not entirely sure what @gnunicorn decided to go down this path.

PR #499, opened 2023-01-27
Workaround #418 - Allow swiping back from file previews

stefanceriu:
After trying to hack my way through the QLPreviewController I basically gave up on the idea
What I did instead was to change the file preview presentation to a stack push and add an extra (invisible) view on top of the hierarchy so that swiping backwards works again. While it's not perfect it's probably better than what we had.

### DIFF
--- a/ElementX/Sources/Screens/FilePreview/View/FilePreviewScreen.swift
+++ b/ElementX/Sources/Screens/FilePreview/View/FilePreviewScreen.swift
@@ -22,10 +22,20 @@ struct FilePreviewScreen: View {
     @ObservedObject var context: FilePreviewViewModel.Context
     
     var body: some View {
-        PreviewView(context: context,
-                    fileURL: context.viewState.fileURL,
-                    title: context.viewState.title)
-            .ignoresSafeArea()
+        ZStack(alignment: .leading) {
+            PreviewView(context: context,
+                        fileURL: context.viewState.fileURL,
+                        title: context.viewState.title)
+                .ignoresSafeArea()
+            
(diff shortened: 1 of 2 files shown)


pixlwave:
Simulator Screen Shot - iPhone 14 Pro - 2023-01-30 at 15 36 13

amshakal:
We seem to be hitting a place where the back button from the full screen image overlaps with the All chats back button, like in the picture above that Doug shared. For that reason, it seems like for FOSDEM it might be best to stick to the full screen view we have. It will make the app look more polished, removing any chance of the user seeing that bug. 

In the future we might want to add genstures to swipe left/right to browse through different images and swipe down to dismiss along with a back button on top to get out of image viewing mode. (Like whatsapp)

ElementX/Sources/Screens/FilePreview/View/FilePreviewScreen.swift
```diff
@@ -22,10 +22,20 @@ struct FilePreviewScreen: View {
...
     var body: some View {
-        PreviewView(context: context,
-                    fileURL: context.viewState.fileURL,
-                    title: context.viewState.title)
-            .ignoresSafeArea()
+        ZStack(alignment: .leading) {
+            PreviewView(context: context,
+                        fileURL: context.viewState.fileURL,
+                        title: context.viewState.title)
+                .ignoresSafeArea()
```
pixlwave, line 29:
```suggestion
                .ignoresSafeArea(.bottom)
```

pixlwave (approved):
Ok, consensus was that ignoring the top safe area is a good compromise for now.

PR #502, opened 2023-01-30
Hardcode the sliding sync proxy.

pixlwave:
This PR hard codes the sliding sync proxy to the new matrix.org one for the demo. It removes the proxy input field on server selection (as this won't be needed in the future anyway) and disables the server selection button on the login screen (as no other servers are going to work for now anyway).

Merging this PR is blocked until the Rust SDK has been updated to handle the new JSON response format for sliding sync.

| Login Screen |
| - |
| Simulator Screen Shot - iPhone 14 - 2023-01-30 at 12 26 09 |

### DIFF
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -17,6 +17,7 @@
 import Combine
 import MatrixRustSDK
 import SwiftUI
+import Version
 
 class AppCoordinator: AppCoordinatorProtocol {
     private let stateMachine: AppCoordinatorStateMachine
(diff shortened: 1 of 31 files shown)


stefanceriu (approved):
👌 👌 👌

pixlwave:
@stefanceriu https://github.com/vector-im/element-x-ios/pull/502/commits/be61fc681ffcda21e3f485b7066de09cb5c6c9f2 adds migrations and bumps the version patch with a migration that signs out the user.

ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -125,6 +129,18 @@ class AppCoordinator: AppCoordinatorProtocol {
         MXLog.configure(loggerConfiguration)
     }
     
+    /// Perform any required migrations for the app to function correctly.
+    private func migrateIfNecessary(from oldVersion: Version, to newVersion: Version) {
```
stefanceriu, line 133: I have no good explanation for this but I find `performMigrationsIfNecessary` a better name

ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -125,6 +129,18 @@ class AppCoordinator: AppCoordinatorProtocol {
...
+        guard oldVersion != newVersion else { return }
+        
+        if oldVersion < Version(1, 0, 17) {
```
stefanceriu, line 136: I wonder if we should check these versions better. As per their documentation `We use a tolerant parser, so strings like "10.1" or even "3" will parse.` and it might even return `Version.null`

ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -125,6 +129,18 @@ class AppCoordinator: AppCoordinatorProtocol {
...
+            // Version 1.0.17 hardcoded a new sliding sync proxy for matrix.org
+            // Force a sign out for the user to log in with the new proxy.
+            MXLog.warning("Clearing user data for hardcoded proxy.")
+            userSessionStore.reset()
```
stefanceriu, line 140: Should we maybe extract "resetting" into a shared method? I have a feeling we will add things to it with time

stefanceriu (approved):

PR #503, opened 2023-01-30
Memory management

stefanceriu:
Starting from the user session not getting dealocated properly I ended up fixing retain cycles caused by SwiftUI's NavigationStack, which seems to hold on to it's path based object for longer than expected.

The solution here is to manually nil a NavigationModule's coordinator and let it deallocate properly. We don't care much about navigationModules being retained as they're of no consequence

This PR will also fix a long standing retaing cycle in the timeline item contex menu builder and clean up various other parts of the code.

NB I haven't gone through the whole app, there might still very well be other memory mangement issues

### DIFF
--- a/ElementX/Sources/Application/Navigation/NavigationModule.swift
+++ b/ElementX/Sources/Application/Navigation/NavigationModule.swift
@@ -34,7 +34,7 @@ class NavigationModule: Identifiable, Hashable {
         self.dismissalCallback = dismissalCallback
     }
     
-    func dismiss() {
+    func tearDown() {
         coordinator?.stop()
         dismissalCallback?()
         coordinator = nil
--- a/ElementX/Sources/Application/Navigation/NavigationModule.swift
+++ b/ElementX/Sources/Application/Navigation/NavigationModule.swift
@@ -18,9 +18,14 @@ import Foundation
 
 /// A CoordinatorProtocol wrapper and type erasing component that allows
 /// dynamically presenting arbitrary screens
-struct NavigationModule: Identifiable, Hashable {
+class NavigationModule: Identifiable, Hashable {
     let id = UUID()
-    let coordinator: any CoordinatorProtocol
+    
+    /// The NavigationStack has a tendency to hold on to path items for longer than needed. We work around that by manually nilling the coordinator
+    /// when a NavigationModule is dismissed, reason why this is an optional property
+    /// As the NavigationModule is just a wrapper multiple instances of it continuing living is of no consequence
+    /// https://stackoverflow.com/questions/73885353/found-a-strange-behaviour-of-state-when-combined-to-the-new-navigation-stack/
+    var coordinator: (any CoordinatorProtocol)?
     let dismissalCallback: (() -> Void)?
     
     init(_ coordinator: any CoordinatorProtocol, dismissalCallback: (() -> Void)? = nil) {
--- a/ElementX/Sources/Application/Navigation/NavigationModule.swift
+++ b/ElementX/Sources/Application/Navigation/NavigationModule.swift
@@ -18,6 +18,7 @@ import Foundation
 
 /// A CoordinatorProtocol wrapper and type erasing component that allows
 /// dynamically presenting arbitrary screens
+@MainActor
 class NavigationModule: Identifiable, Hashable {
     let id = UUID()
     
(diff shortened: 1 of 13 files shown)


ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -72,7 +70,13 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
             }
             .store(in: &cancellables)
         
-        state.contextMenuBuilder = buildContextMenuForItemId(_:)
+        state.contextMenuActionProvider = { [weak self] itemId -> TimelineItemContextMenuActions? in
+            guard let self = self else {
```
pixlwave, line 74: 😉
```suggestion
            guard let self else {
```

ElementX/Sources/Screens/RoomScreen/View/TimelineItemContextMenu.swift
```diff
@@ -17,11 +17,8 @@
 import SwiftUI
 
 struct TimelineItemContextMenuActions {
-    static var empty: TimelineItemContextMenuActions { .init(actions: [], debugActions: []) }
-    
     let actions: [TimelineItemContextMenuAction]
     let debugActions: [TimelineItemContextMenuAction]
-    var isEmpty: Bool { actions.isEmpty && debugActions.isEmpty }
```
pixlwave, line 24: Small tweak that might be useful for the future - we could leave this property and check it at the end of `contextMenuActionsForItemId` to return `nil` if it turns out that we built no actions for that item (for whatever reason).

ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
```diff
@@ -44,6 +44,7 @@ enum RoomScreenViewAction {
     case cancelEdit
     /// Mark the entire room as read - this is heavy handed as a starting point for now.
     case markRoomAsRead
+    case handleContextMenuAction(itemID: String, action: TimelineItemContextMenuAction)
```
pixlwave, line 47: Could probably just call this `contextMenuAction` to be fair.

ElementX/Sources/Application/Navigation/NavigationModule.swift
```diff
@@ -18,21 +18,33 @@ import Foundation
...
+    var coordinator: (any CoordinatorProtocol)?
     let dismissalCallback: (() -> Void)?
     
     init(_ coordinator: any CoordinatorProtocol, dismissalCallback: (() -> Void)? = nil) {
         self.coordinator = coordinator
         self.dismissalCallback = dismissalCallback
     }
     
-    static func == (lhs: NavigationModule, rhs: NavigationModule) -> Bool {
+    func dismiss() {
```
pixlwave, line 37: Can we call this something different? At the call site it reads like it this is the method that dismisses the module. Something like `onDismiss`/`didDismiss`/`stop` perhaps?

ElementX/Sources/Application/Navigation/NavigationModule.swift
```diff
@@ -18,21 +18,33 @@ import Foundation
 
 /// A CoordinatorProtocol wrapper and type erasing component that allows
 /// dynamically presenting arbitrary screens
-struct NavigationModule: Identifiable, Hashable {
+@MainActor
+class NavigationModule: Identifiable, Hashable {
```
pixlwave, line 22: How come this went from a `struct` to an `@MainActor class`?

pixlwave (approved):
LGTM, nice job 🙌

ElementX/Sources/Application/Navigation/NavigationModule.swift
stefanceriu: How about tearDown?

ElementX/Sources/Application/Navigation/NavigationModule.swift
stefanceriu: It's a main actor because it now calls a method on another main actor and it's a class because it's easier mutating its coordinator this way

ElementX/Sources/Screens/RoomScreen/View/TimelineItemContextMenu.swift
stefanceriu: I wonder if it's simpler to just have a optional constructor that just checks this internally..

ElementX/Sources/Application/Navigation/NavigationModule.swift
pixlwave: Yep yep, thanks that feels much clearer 🙏

PR #505, opened 2023-01-31
Initial timeline design tweaks

pixlwave:
This PR contains the first set of design tweaks to the timeline screen. The changes are mainly fonts, colours and spacing.

- Images now have a maximum height in the timeline.
- Fixes timeline view previews that needed a context in the environment.
- Message bubble logic has been simplified with much more code reuse between ingoing/outgoing messages.
- Context menu previews are fixed and only show the message bubble.
- The message composer height calculation is entirely handled in the representable view struct now.
- Encrypted items are no longer buttons - If its useful, maybe we should put the ID into the Developer menu as a disabled action?
- When a room has no last message we show "New message" to plug the gap.

### DIFF
--- a/ElementX/Resources/Localizations/en.lproj/Untranslated.strings
+++ b/ElementX/Resources/Localizations/en.lproj/Untranslated.strings
@@ -4,6 +4,8 @@
 "action_confirm" = "Confirm";
 "action_match" = "Match";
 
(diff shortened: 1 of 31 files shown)


stefanceriu (approved):
nothing jumps out, lgtm 👍

PR #506, opened 2023-01-31
Handle API changes from Rust.

pixlwave:
There are some bad assumptions about profile changes in here as we don't have the member/sender key from Rust.


stefanceriu (approved):
Looks great 👏

PR #507, opened 2023-01-31
Dynamically adjusting the visibleRoomsView's timeline limit based on …

stefanceriu:
…the app state

Requires https://github.com/matrix-org/matrix-rust-sdk/pull/1399 and https://github.com/vector-im/element-x-ios/pull/502

This will start the app with `visibleRooms[timelineLimit=0]`, then, as soon as that receives updates, switch to `visibleRooms[timeline=1]` + `allRooms` and later, once the initial data has been displayed, switch to `visibleRooms[timeline=20]`+`allRooms`

After that, while scrolling and chaning the visible range the timeline limit with be set to 1 and then set again to 20 when the scrolling has stopped to precache room history

### DIFF
--- a/ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
+++ b/ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
@@ -19,14 +19,10 @@ import SwiftUI
 struct HomeScreen: View {
     @ObservedObject var context: HomeScreenViewModel.Context
     
+    @State private var scrollViewAdapter = ScrollViewAdapter()
     @State private var showingLogoutConfirmation = false
-    @State private var visibleItemIdentifiers = Set<String>() {
-        didSet {
-            if visibleItemIdentifiers != oldValue {
-                updateVisibleRange()
-            }
-        }
-    }
+    @State private var visibleItemIdentifiers = Set<String>()
+    @State private var hasTriggeredInitialVisibleItemUpdate = false
     
     var body: some View {
         ScrollView {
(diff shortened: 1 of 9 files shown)


ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
```diff
@@ -76,6 +76,17 @@ struct HomeScreen: View {
                 userMenuButton
             }
         }
+        .onChange(of: visibleItemIdentifiers, perform: { _ in
+            if !hasTriggeredInitialVisibleItemUpdate {
+                updateVisibleRange()
+                hasTriggeredInitialVisibleItemUpdate = true
+            }
+        })
```
pixlwave, line 84: I'm sure something will be grumbling about this somewhere, but:
```suggestion
        .onChange(of: visibleItemIdentifiers) { _ in
            if !hasTriggeredInitialVisibleItemUpdate {
                updateVisibleRange()
                hasTriggeredInitialVisibleItemUpdate = true
            }
        }
```

manuroe:
> This will start the app with `visibleRooms[timelineLimit=0]`, then, as soon as that receives updates, switch to `visibleRooms[timeline=1]` + `allRooms` and later, once the initial data has been displayed, switch to `visibleRooms[timeline=20]`+`allRooms`
> 
> After that, while scrolling and chaning the visible range the timeline limit with be set to 1 and then set again to 20 when the scrolling has stopped to precache room history

If it works, we need to add those sentences somewhere in the documentation. There are super clear.

pixlwave (approved):
LGTM

Issue #500, opened 2023-01-28
Money is inexplicably coloured blue


ara4n:
![59FCFCF4-EB68-47B8-A77D-6D2DF2088B18](https://user-images.githubusercontent.com/1294269/215231567-9fecafba-1b9a-4862-9229-33a8d6a2415d.jpeg)

PR #509, opened 2023-01-31
iOS: Removing event id links

Velin92:
Event Identifiers were actually recognised as links, however this behaviour does not actually exist on any other version of the app, (like El-Web) and even if they were recognised as link, they were not handled (the link just contained the event identifier, which is actually not a valid URL resulting in an untappable link).

This also caused a bug where any $ amount was always being seen as an untappable link:
https://github.com/vector-im/element-x-ios/issues/500


stefanceriu (changes requested):
Good catch but I don't think we should remove the mxid regex parser. It feels like something that we will definitely need in the future and it's already well tested.

What might be interesting to look into is whether it detecting money is valid or if there's a bug somewhere.

Velin92:
> Good catch but I don't think we should remove the mxid regex parser. It feels like something that we will definitely need in the future and it's already well tested.
> 
> What might be interesting to look into is whether it detecting money is valid or if there's a bug somewhere.

I see, however even if we have the link detector for event id, the link actually does nothing even on valid event identifiers. Maybe we should keep the code for the future, but remove the check of converting these identifiers to links? Since even with valid ones it's going to create untappable links

stefanceriu:
> but remove the check of converting these identifiers to links

I'm good with removing the line from the AttributedStringBuilder, I just want to keep what's in `MatrixEntityRegex` and friends.

And after another look, we should definitely keep the check from `permalinkTo(eventIdentifier`, I missed that in my initial review

Velin92:
> > but remove the check of converting these identifiers to links
> 
> I'm good with removing the line from the AttributedStringBuilder, I just want to keep what's in `MatrixEntityRegex` and friends.
> 
> And after another look, we should definitely keep the check from `permalinkTo(eventIdentifier`, I missed that in my initial review

done! commented the code and added a comment on top of it, so we could reuse it later when we actually know what to do with such links

stefanceriu (approved):
great 👏

PR #511, opened 2023-02-01
Introduce back custom room list item identifiers for invalidated rooms

stefanceriu:
With duplicates for filled rooms now being fixed (no more duplicated sliding sync pos requests) we can confirm that indeed sometimes there are duplicated identifiers between an invalid and a filled room. This was expected but we were expecting the rest of the bugs to be solved before acting on it


pixlwave (approved):
LGTM

PR #513, opened 2023-02-01
Added a progress tracker to the bug report view

Velin92:
https://user-images.githubusercontent.com/34335419/216094889-6c693abd-5fed-4726-90cd-1e01a12bb0d7.MP4

Added a custom progress tracker class that can be used to observe progress in a data task.
Such object can be attached to a user notification and user to observe the progress of any loading to create progress views that track such progress.

Other improvements I want to add:

- [x] The signature for `submitBugReport` is too big, maybe is worth create a dedicated struct
- [x]  Move the ProgressTracker class in a specific file
- [x] I guess I could add a mock version of the UserNotificationView that also displays a progress bar


ElementX/Sources/Screens/BugReport/BugReportViewModel.swift
```diff
@@ -49,13 +49,16 @@ class BugReportViewModel: BugReportViewModelType, BugReportViewModelProtocol {
...
-                                                                    includeCrashLog: true,
-                                                                    githubLabels: [],
-                                                                    files: [])
+            let bugReport = BugReport(text: context.reportText,
+                                      includeLogs: context.sendingLogsEnabled,
+                                      includeCrashLog: true,
+                                      githubLabels: [],
+                                      files: [])
+            let result = try await bugReportService.submitBugReport(bugReport,
+                                                                    progressTracker: progressTracker)
```
stefanceriu, line 61: Should the BugReportService be the one that provides the progress through a simple value combine publisher? The fact that the `ProgressTracker` is now a `URLSessionTaskDelegate` leaks internal details for no good reason.
The `ProgressTracker` should probably be a simple protocol and have different services provide progress in whatever way they choose fit.

ElementX/Sources/Other/ProgressTracker.swift
```diff
@@ -0,0 +1,36 @@
...
+        $progressFraction
+            .receive(on: RunLoop.main)
+            .eraseToAnyPublisher()
+    }
+
+    init(initialValue: Double = 0.0) {
+        progressFraction = initialValue
+    }
+
+    func urlSession(_ session: URLSession, didCreateTask task: URLSessionTask) {
```
stefanceriu, line 33: See my comment below but I don't think this belongs here. The fact that the BugService is using an URLSession should remain an implementation detail.

ElementX/Sources/Other/ProgressTracker.swift
```diff
@@ -0,0 +1,36 @@
...
+
+import Combine
+import Foundation
+
+final class ProgressTracker: NSObject, URLSessionTaskDelegate {
+    @Published private var progressFraction: Double
+
+    var progressFractionPublisher: AnyPublisher<Double, Never> {
+        $progressFraction
+            .receive(on: RunLoop.main)
```
stefanceriu, line 25: Why is this RunLoop.main instead of DispatchQueue.main?

ElementX/Sources/Other/ProgressTracker.swift
```diff
@@ -0,0 +1,36 @@
...
+// limitations under the License.
+//
+
+import Combine
+import Foundation
+
+final class ProgressTracker: NSObject, URLSessionTaskDelegate {
+    @Published private var progressFraction: Double
+
+    var progressFractionPublisher: AnyPublisher<Double, Never> {
```
stefanceriu, line 23: Let's simply name this `progressPublisher`

ElementX/Sources/Other/ProgressTracker.swift
```diff
@@ -0,0 +1,36 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Combine
+import Foundation
+
+final class ProgressTracker: NSObject, URLSessionTaskDelegate {
+    @Published private var progressFraction: Double
```
stefanceriu, line 21: Why do we even need to store this if we're just passing values through?  Might be a good idea to expose a CurrentValueSubject instead

ElementX/Sources/Other/UserNotifications/UserNotificationModalView.swift
```diff
@@ -14,16 +14,22 @@
...
-    
+    @State private var progressFraction: Double?
+
     var body: some View {
         ZStack {
             VStack(spacing: 12.0) {
-                ProgressView()
-                
+                if let progressFraction = progressFraction {
+                    ProgressView(value: progressFraction)
```
stefanceriu, line 28: We might want to set a max frame width now that this has been introduced

stefanceriu (changes requested):
Left a couple of comments inline, hope they make sense.

Also, unrelated to your PR (I noticed in the video) but I think the transition animation for the modal is wrong. It should probably be a simple `.opacity` instead of the `asymmetric .top .bottom` it's doing now. Would appreciate it if you could have a look at it.

ElementX/Sources/Other/ProgressTracker.swift
stefanceriu, line 21: Also, does this `@Published` even work without it being an `ObservableObject`?

Velin92: Yup, @Published is just essentially creating a wrapped publisher of type `CurentValueSubject<T,Never>` where `T` is the type of the variable, if the variable is also contained in an `ObservavbleObject`, it will fire the `objectWillChange.send()` otherwise the `objectWillChange.send()`  will not be called.

However you can still listen to the changes of the wrapped publisher. I usually just find more readable, easier and faster to write a simple a `Published var` instead of a `CurentValueSubject` if I know that there is no error case, and that if the `Failure`  type is `Never`

Velin92: `CurrentValueSubject` would still store a value, since it has an `init` with value and a `value` field, also explained the reason for my preference in the other comment. However since they are essentially almost the same thing I can use a `CurrentValueSubject` instead.

ElementX/Sources/Other/ProgressTracker.swift
Velin92: Uhm `DispatchQueue.main` is actually better now that I think of it... `RunLoop.main` can get busy by user interactions making it less for general use.

Velin92:
> Left a couple of comments inline, hope they make sense.
> 
> Also, unrelated to your PR (I noticed in the video) but I think the transition animation for the modal is wrong. It should probably be a simple `.opacity` instead of the `asymmetric .top .bottom` it's doing now. Would appreciate it if you could have a look at it.

Will give a look at the presentation issue.
I see is actually also happening on TF, so maybe is some kind of regression, so you want me to take a look at it in a different PR/branch tomorrow (I'll be back from my holidays/leave), unrelated to the progress tracker?

ElementX/Sources/Other/ProgressTracker.swift
stefanceriu: That's pretty smart, didn't tink about it like that before.

I'm generaly weary of using SwiftUI specific wrappers in other places and in this particular case it seems there's a few gotchas with it: it fires on `willChange` instead of `didChange` and is restricted to structs https://www.avanderlee.com/swiftui/published-property-wrapper/

As for protocols, presumably you would stilexpose a `CurentValueSubject` if you need access to the current value without exposing a separate property, no?

stefanceriu:
> Will give a look at the presentation issue.
> I see is actually also happening on TF, so maybe is some kind of regression, so you want me to take a look at it in a different PR/branch tomorrow (I'll be back from my holidays/leave), unrelated to the progress tracker?

I don't think it's a regression, I think I just copy pasted it from the `UserNotificationToastView` and forgot to change the animation style. Up to you if you want to handle it in this PR or separately. The change should be pretty simple:
`.transition(.opacity)` instead of that custom `toastTransition: AnyTransition`

ElementX/Sources/Other/ProgressTracker.swift
Velin92: The thing is that the `@Published` technically speaking is part of Combine (even if you don't even need to import it, to use it, Foundation is sufficient), but yeah it is constricted to structs.

For protocols yeah if you really need to use one you could either do that with a `CurrentValueSubject` which could use the `$` publisher of the struct since it's literally a `CurrentValueSubject` (if you use `po` you will see it), or you can have the protocol expose the value directly as T, and have an AnyPublisher<T, Never>, and in the implementation use the `@Published`, depends if you really want to have the interface clearly say that the value will be published or you want to hide it.

Regarding the willSet, that could be risky if you have a case where you are listening to the change provided by the publisher... but at the same time you also checking for the value wrapped in the publisher... which would not really make a lot of sense to use it in that way.

stefanceriu (approved):
Looking good, nicely done! 👏

PR #514, opened 2023-02-01
Allow different states for the last message.

pixlwave:
This PR makes a small change to the last message in the room list. It now has 3 states defined in an enum and we choose whether to display the message, a loading placeholder, or a placeholder string when the message is unknown.

https://user-images.githubusercontent.com/6060466/216105909-0a475d52-055f-49fc-a740-bbb06ec1ba63.mp4


stefanceriu (approved):
lgtm 👍

PR #519, opened 2023-02-02
Fix remaining retain cycles, allow coordinators to be weakly referenc…

stefanceriu:
…ed and make sure the NavigationStackCoordinator cleans up after itself


pixlwave (approved):
LGTM

PR #520, opened 2023-02-02
Refactor the dynamic timeline modifiers and make sure they fire in th…

stefanceriu:
…e right order


ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -136,9 +119,14 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
             .store(in: &cancellables)
         
         allRoomsSummaryProvider.roomListPublisher
+            .dropFirst(1) // We don't care about its initial value
```
pixlwave, line 122: I'm sure its fine, but will this do anything wacky to UI tests with mocks that don't update?

pixlwave (approved):
Wooop, this looks good and testing it on my device fixes the problem 👍

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
stefanceriu: I don't think so but that's definitely something to look out for when we fix the tests

PR #521, opened 2023-02-02
Prevent duplicated timeline items from crashing the diffable datasource

stefanceriu:
The Rust side timeline doesn't correctly handle timeline event trickling at the moment and generates duplicated items. This in turn crashes the Swift side timeline's diffable datasource.

This PR will remove those duplicates just before they hit the UI level


pixlwave (approved):
🤞

PR #522, opened 2023-02-03
Load member profiles through the timeline

stefanceriu:
This PR simplifies the way we load data for timeline item sender profiles. It replaces the need to call `loadDisplayNameForUserId` with a `fetchMembers` internal to the room proxy any time we add a timeline listener.
It also fixes flashing display names in the timeline and avatars that don't load

### DIFF
--- a/ElementX.xcodeproj/project.pbxproj
+++ b/ElementX.xcodeproj/project.pbxproj
@@ -4012,7 +4012,7 @@
 			repositoryURL = "https://github.com/matrix-org/matrix-rust-components-swift";
 			requirement = {
 				kind = exactVersion;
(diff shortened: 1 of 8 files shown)


pixlwave (approved):
Nice, super excited to have this now 👏👏👏

Can we remove the following from `RoomProxy` too now?

```
func avatarURLForUserId()
var memberAvatars
func displayNameForUserId()
var memberDisplayNames
```

stefanceriu:
> Can we remove the following from `RoomProxy` too now?

Seeing as there's no other direct entry point for fetching them I figured I'm just going to leave them in for now ¯\_(ツ)_/¯

PR #525, opened 2023-02-03
Revert decryption banner & try fix context menu crash.

pixlwave:
Reverts the decryption banner for a button in the context menu and potentially fix a crash when building context menus.


stefanceriu (approved):
🚢  it!

PR #538, opened 2023-02-06
Various tweaks

stefanceriu:
* remove now unneeded migration from caches to application support
* add `function_body_length` swiftlint rule with default values in an attempt to fix false positives on the CI (DangerSwift)


pixlwave (approved):
LGTM 👍

Issue #532, opened 2023-02-06
SwiftUI environment object crash


stefanceriu:
The app keeps crashing while trying to access a SwiftUI environment object

https://sentry.tools.element.io/organizations/element/issues/61228/?project=44&query=is%3Aunresolved+release.version%3A1.0.18&statsPeriod=14d

We believe this is to do with UIKit layer we use for the timeline and it not respecting the environment hierarchy when reusing cells.

We will try manually injecting the environment object at that point but if that fails we should switch to manual constructor injection

PR #539, opened 2023-02-06
Explicitly pass the context into the cell's environment.

pixlwave:
Potential fix for #532 on the assumption that at times the UIKit components of the timeline may not be part of the view hierarchy/have access to the environment 🤞


stefanceriu (approved):
👍

PR #544, opened 2023-02-07
Add shimmer effect to room skeletons.

pixlwave:
Adds a shimmer effect to the skeletons when waiting for the server.

https://user-images.githubusercontent.com/6060466/217235495-d94a4434-7be7-4853-8bb5-432a7409a4e9.mov


stefanceriu (approved):
🤩

PR #552, opened 2023-02-08
Fixes #367 - Refactor AppSettings usage patterns

### LINKED ISSUE #367: Element settings usage patterns
We are not happy with the current way of changing application settings in which the view reaches up to the ServiceLocator and it's settings object to directly set values.

After much though we decided to expose publishers for each setting by writing a custom property wrapper. The viewModel should sink the publishers its interested in and pass those values through the view state to the view. The view on the other hand should send back a view action that the view model can act on and update the settings \(or not, depending on what the logic is\)

stefanceriu:
We are not happy with the current way of changing application settings in which the view reaches up to the ServiceLocator and it's settings object to directly set values.

After much thought we decided to expose publishers for each setting by writing a custom property wrapper. The viewModel should sink the publishers it's interested in and pass those values through the view state to the view. The view on the other hand should send back a view action that the view model can act on and update the settings (or not, depending on what the logic is)

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
+++ b/ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
@@ -40,6 +40,7 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol 
         super.init(initialViewState: RoomScreenViewState(roomId: timelineController.roomID,
                                                          roomTitle: roomName ?? "Unknown room 💥",
                                                          roomAvatarURL: roomAvatarUrl,
+                                                         timelineStyle: ServiceLocator.shared.settings.timelineStyle,
                                                          bindings: .init(composerText: "", composerFocused: false)),
                    imageProvider: mediaProvider)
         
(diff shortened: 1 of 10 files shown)


ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -78,6 +79,11 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
             return self.contextMenuActionsForItemId(itemId)
         }
         
+        ServiceLocator.shared.settings.$timelineStyle.sink { [weak self] timelineStyle in
+            self?.state.timelineStyle = timelineStyle
+        }
+        .store(in: &cancellables)
+
```
Velin92, line 86: I hate that we need to use `sink` for a single value in a single field to avoid retain cycles, instead of using the easier `assign(to:, on:)`.
If you want we could add this extension:
```
extension Publisher where Self.Failure == Never {
    public func assignNoRetain<Root>(to keyPath: ReferenceWritableKeyPath<Root, Self.Output>, on object: Root) -> AnyCancellable where Root: AnyObject {
        sink { [weak object] (value) in
        object?[keyPath: keyPath] = value
    }
  }
}
```

To have an assign that will always prevent the strong reference usage.

We can also tweak this a step further by overloading the original assign and always use a weak behind the shadows when dealing with Objects:
```
extension Publisher where Failure == Never {
    func assign<Root: AnyObject>(to keyPath: ReferenceWritableKeyPath<Root, Output>, on root: Root) -> AnyCancellable {
       sink { [weak root] in
            root?[keyPath: keyPath] = $0
        }
    }
}
```

pixlwave (approved):
Looks good from my side 👍

PR #554, opened 2023-02-08
Updated Screenshot Tests and fixed an issue with the MockMediaProvider

Velin92:
Updated all the screenshots to adapt to the new designs.

Also fixed an issue that did not fetch an image for the MockMediaProvider to render mocked avatar rooms, and an issue where the roomID was incorrectly using the id of the proxy and not the roomId coming from the timeline


ElementX/Sources/UITests/UITestsAppCoordinator.swift
```diff
@@ -275,10 +275,29 @@ class MockScreen: Identifiable {
         case .roomDetailsScreen:
             let navigationStackCoordinator = NavigationStackCoordinator()
             let userNotificationController = UserNotificationController(rootCoordinator: navigationStackCoordinator)
+            var roomProxy = MockRoomProxy(
+                displayName: "Room",
+                isEncrypted: true,
+                members: [.mockAlice, .mockBob, .mockCharlie]
+            )
+            roomProxy.id = "MockRoomIdentifier"
```
pixlwave, line 283: Style and you can use the synthesised init for the `id`.
```suggestion
            var roomProxy = MockRoomProxy(id: "MockRoomIdentifier",
                                          displayName: "Room",
                                          isEncrypted: true,
                                          members: [.mockAlice, .mockBob, .mockCharlie])
```

UITests/Sources/RoomDetailsScreenUITests.swift
```diff
@@ -22,6 +22,15 @@ class RoomDetailsScreenUITests: XCTestCase {
         let app = Application.launch()
         app.goToScreenWithIdentifier(.roomDetailsScreen)
         
+        XCTAssert(app.staticTexts["roomAvatarImage"].exists)
         app.assertScreenshot(.roomDetailsScreen)
     }
+
+    func testInitialStateComponentsWith() {
```
pixlwave, line 29:
```suggestion
    func testInitialStateComponentsWithAvatar() {
```

pixlwave (approved):
Thanks for all these updates 👏

UITests/Sources/RoomDetailsScreenUITests.swift
Velin92: done

ElementX/Sources/UITests/UITestsAppCoordinator.swift
Velin92: done

PR #561, opened 2023-02-09
BrowserStack support

stefanceriu:
We want to be able to test the app through BrowserStack and they do not currently support Application Groups. In order to get that working I slightly changed the way we create the path where we store user data. It shouldn't affect normal users in any way
I also added support for automatic ipa file uplods to BrowserStack for PR builds

### DIFF
--- a/ElementX/Sources/Other/Extensions/URL.swift
+++ b/ElementX/Sources/Other/Extensions/URL.swift
@@ -28,8 +28,16 @@ extension URL {
     /// The URL of the primary app group container.
     static var appGroupContainerDirectory: URL {
         guard let url = FileManager.default.containerURL(forSecurityApplicationGroupIdentifier: InfoPlistReader.main.appGroupIdentifier) else {
-            fatalError("Should always be able to retrieve the container directory")
+            // Browserstack doesn't properly handle AppGroup entitlements so this fails, presumably because of the resigning happening on their side
+            // Try using the normal app folder instead of the app group
+            // https://www.browserstack.com/docs/app-automate/appium/troubleshooting/entitlements-error
+            guard let url = FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask).first?.deletingLastPathComponent().deletingLastPathComponent() else {
+                fatalError("Should always be able to retrieve the app directory")
+            }
+            
+            return url
         }
+        
         return url
     }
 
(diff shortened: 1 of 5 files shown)


ElementX/Sources/Other/Extensions/URL.swift
```diff
@@ -28,8 +28,17 @@ extension URL {
...
-            fatalError("Should always be able to retrieve the container directory")
+            MXLog.error("Application Group unavailable, falling back to the application folder")
+            // Browserstack doesn't properly handle AppGroup entitlements so this fails, presumably because of the resigning happening on their side
+            // Try using the normal app folder instead of the app group
+            // https://www.browserstack.com/docs/app-automate/appium/troubleshooting/entitlements-error
+            guard let url = FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask).first?.deletingLastPathComponent().deletingLastPathComponent() else {
+                fatalError("Should always be able to retrieve the app directory")
+            }
+            
+            return url
```
pixlwave, line 39: We discussed that this fallback isn't going to cause any harm elsewhere.

Small suggestion:
```suggestion
            return URL.applicationSupportDirectory.deletingLastPathComponent().deletingLastPathComponent()
```
Also, how come we don't use the directory provided and instead go up 2 levels?

pixlwave (approved):
LGTM

ElementX/Sources/Other/Extensions/URL.swift
stefanceriu: Because I didn't want to add 1 `Library/Application Support` layer to the existing one

stefanceriu: I'm not sure I understand the suggestion: `applicationSupportBaseDirectory` is using the `appGroupContainerDirectory` internally 🤔

pixlwave: > Because I didn't want to add 1 `Library/Application Support` layer to the existing one

Oh sorry yes this overall property is the bast directory I see. The suggestion is purely to use the non-failable `URL.applicationSupportDirectory` instead of `urls(for: .applicationSupportDirectory, in:) which returns an optional.

stefanceriu: Ah sorry, I mistook `applicationSupportDirectory` for our own `applicationSupportBaseDirectory`

PR #562, opened 2023-02-09
Message composer icon and scroll to bottom button dark mode

pixlwave:
This PR
- Updates the composer send icon (not enough of a change to fail UI Tests) to use our custom paper plane.
- Remove animations on the composer when entering/exiting edit/reply mode as the animations were conflicting with the keyboard dismissal/appearance.
- Updates the scroll to bottom button in the room screen to support dark mode.

| Light Mode, send enabled | Dark Mode, send disabled |
| - | - |
| Simulator Screen Shot - iPhone 14 - 2023-02-09 at 15 57 42 | Simulator Screen Shot - iPhone 14 - 2023-02-09 at 15 57 22 |


ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
```diff
@@ -17,6 +17,8 @@
 import SwiftUI
 
 struct RoomScreen: View {
+    @Environment(\.colorScheme) private var colorScheme
```
stefanceriu, line 20: Is this unused? 🤔

stefanceriu (approved):
👍 👍

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
pixlwave: Whooops yes I didn't need it in the end, thanks.

PR #563, opened 2023-02-09
Swiftformat on githooks + New `swift run tools setup` command

Velin92:
Moved githooks to a new a folder called .githooks, so that they can be committed and tweaked for all the team.

Added a pre-commit file that will run swiftformat --lint . and if finds any linting issue, will prevent the commit to occur and then swiftformat the project.
Once done it will ask the contributor to try committing again, since their commit needed some formatting.

To allow every contributor to use xcodegen, swiftformat and align the usage of githooks in our custom folder, I created a new `swift run tool setup` that could be used to setup the project for a first run (I am wondering if we want to remove the xcodeprj, to enforce the setup as a first run).

Also added a githook to run xcodegen after every checkout.
Improved docs.
Added Brewfile.lock.json to the .gitignore


Tools/Sources/Setup.swift
```diff
@@ -0,0 +1,24 @@
+import ArgumentParser
+import Foundation
+
+struct Setup: ParsableCommand {
```
stefanceriu, line 4: Let's maybe rename this to something less generic like something along the lines of `SetupProject` or `SetupDevelopmentMode`

Tools/Sources/Tools.swift
```diff
@@ -4,5 +4,6 @@ import Foundation
 @main
 struct Tools: ParsableCommand {
     static var configuration = CommandConfiguration(abstract: "A collection of command line tools for ElementX",
-                                                    subcommands: [BuildSDK.self])
```
stefanceriu, line 7: Actually while we're on renaming things perhaps `BuildSDK` should be renamed too as it does more than just building it. Maybe `SetupSDKForLocalDevelopment`

.githooks/pre-commit
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+export PATH="$PATH:/opt/homebrew/bin"
+
+if ! swiftformat --lint . > /dev/null 2>&1
+then
+  echo "pre-commit: Commit aborted to run SwiftFormat, please try again"
```
stefanceriu, line 7:
```suggestion
  echo "pre-commit: Commit aborted due to SwiftFormat warnings. Please check the automatically generated fixes and try again"
```

stefanceriu (changes requested):
This looks (and works) great! 👏  
Please update the documentation to make it absolutely clear to everybody that this is a mandatory step.
We can handle removing the project file entirely in a separate PR

Tools/Sources/Tools.swift
pixlwave: I don't want to `swift run tool setup-sdk-for-local-development` though. You can specify a custom command name, but that seems unhelpful to me once you have more commands as you have to search for the right one.

From the user's perspective, the `abstract` clearly explains what the command does.

stefanceriu: Fair enough 👍

Tools/Sources/Utils.swift
```diff
@@ -0,0 +1,37 @@
+import ArgumentParser
+import Foundation
+
+enum Utils {
```
pixlwave, line 4: Pet peeve, could we call this `Utilities` rather shortening it? Or if you want a shorter name then `Tools` perhaps?

pixlwave (approved):
Looks great to me 👏

stefanceriu (approved):

.githooks/post-checkout
```diff
@@ -1,3 +1,8 @@
 #!/bin/sh
 command -v git-lfs >/dev/null 2>&1 || { echo >&2 "\nThis repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, remove this hook by deleting '.git/hooks/post-checkout'.\n"; exit 2; }
 git lfs post-checkout "$@"
+
+#!/bin/bash
```
pixlwave, line 5: Does this actually do anything when its not the first line of the file (genuine question, I have no idea).

PR #564, opened 2023-02-10
Update PR Build workflow.

pixlwave:
- Allows the workflow to run when the PR is labelled after being created.
- Removes manual runs as there isn't anywhere to post the link.
- Simplifies the run conditions seeing as this is a manual action now.


stefanceriu (approved):
lgtm 👍

PR #567, opened 2023-02-10
Align accessibility IDs with elementX android

michaelkaye:
The reason to change these is to bring them in line with android; such that we can (ideally!) use only one appium flow to automate either client. I doubt it will last for too long; but it will reduce duplication if we don't need to maintain two separate drivers.

The logic behind the naming scheme is a fairly simple:

`<screen>-<option>`

So `change_server`-`continue` is the continue button on change server. This means we can tell it apart from say `onboarding`-`continue` which is the continue button on the onboarding page - both could have the same text on them but we don't want to click an onboarding screen button if the test is out of sync with the app, and were trying to click the change_server button, for instance. 

The screen names don't align perfectly with either client's Views etc (eg the button to change server is part of a sub-view of the login page), because of slight differences in how the clients are coded, so the naming system for screens considers the point of view of an user looking at the app without knowledge of the underlying code.

I note that there's the SoftLogout screen that hasn't been changed as part of this, while still retaining some of the same IDs. I've not changed that because we've not automated that section yet.

There will be more PRs but this is an initial section that works.

I'm not an iOS developer and don't have a proper development environment, so please directly correct any oddities in my coding style - the tests are still passing so I view it as "good enough", but there may be more that needs to be done here.

### DIFF
--- a/DesignKit/Sources/TextFields/ElementTextFieldStyle.swift
+++ b/DesignKit/Sources/TextFields/ElementTextFieldStyle.swift
@@ -21,8 +21,9 @@ import SwiftUI
 public extension TextFieldStyle where Self == ElementTextFieldStyle {
     static func elementInput(labelText: String? = nil,
                              footerText: String? = nil,
-                             isError: Bool = false) -> ElementTextFieldStyle {
-        ElementTextFieldStyle(labelText: labelText, footerText: footerText, isError: isError)
+                             isError: Bool = false,
+                             accessibilityIdentifier: String? = nil) -> ElementTextFieldStyle {
+        ElementTextFieldStyle(labelText: labelText, footerText: footerText, isError: isError, accessibilityIdentifier: accessibilityIdentifier)
     }
 }
 
@@ -38,6 +39,7 @@ public struct ElementTextFieldStyle: TextFieldStyle {
     public let labelText: String?
     public let footerText: String?
     public let isError: Bool
+    public let accessibilityIdentifier: String?
     
     /// The color of the text field's border.
     private var borderColor: Color {
@@ -91,10 +93,11 @@ public struct ElementTextFieldStyle: TextFieldStyle {
     ///   - labelText: The text shown in the label above the field.
     ///   - footerText: The text shown in the footer label below the field.
     ///   - isError: Whether or not the text field is currently in the error state.
-    public init(labelText: String? = nil, footerText: String? = nil, isError: Bool = false) {
+    public init(labelText: String? = nil, footerText: String? = nil, isError: Bool = false, accessibilityIdentifier: String? = nil) {
         self.labelText = labelText
         self.footerText = footerText
         self.isError = isError
+        self.accessibilityIdentifier = accessibilityIdentifier
     }
     
     public func _body(configuration: TextField<_Label>) -> some View {
@@ -128,8 +131,9 @@ public struct ElementTextFieldStyle: TextFieldStyle {
                     textField.clearButtonMode = .whileEditing
                     textField.attributedPlaceholder = NSAttributedString(string: textField.placeholder ?? "",
                                                                          attributes: [NSAttributedString.Key.foregroundColor: UIColor(placeholderColor)])
+                    textField.accessibilityIdentifier = accessibilityIdentifier
                 }
-            
+ 
             if let footerText {
                 Text(footerText)
                     .font(.element.caption1)
--- a/ElementX/Sources/Other/AccessibilityIdentifiers.swift
+++ b/ElementX/Sources/Other/AccessibilityIdentifiers.swift
@@ -1,0 +1,89 @@
+//
+// Copyright 2023 New Vector Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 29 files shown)


michaelkaye:
(Ideally we'd refactor these into a file of static constants or similar, to make it easier to use the same IDs across tests etc)

michaelkaye:
```xml
<XCUIElementTypeScrollView type="XCUIElementTypeScrollView" enabled="true" visible="true" accessible="false" x="0" y="0" width="390" height="844" index="0">
  <XCUIElementTypeOther type="XCUIElementTypeOther" enabled="true" visible="true" accessible="false" x="0" y="91" width="390" height="378" index="0">
    <XCUIElementTypeStaticText type="XCUIElementTypeStaticText" value="Select your server" name="Select your server" label="Select your server" enabled="true" visible="true" accessible="true" x="102" y="195" width="186" height="27" index="0"/>
    <XCUIElementTypeStaticText type="XCUIElementTypeStaticText" value="What is the address of your server? This is like a home for all your data" name="What is the address of your server? This is like a home for all your data" label="What is the address of your server? This is like a home for all your data" enabled="true" visible="true" accessible="true" x="32" y="229" width="326" height="43" index="1"/>
    <XCUIElementTypeStaticText type="XCUIElementTypeStaticText" value="Homeserver URL" name="change_server-server" label="Homeserver URL" enabled="true" visible="true" accessible="true" x="32" y="307" width="103" height="17" index="2"/>
    <XCUIElementTypeTextField type="XCUIElementTypeTextField" value="matrix.org" name="change_server-server" label="" enabled="true" visible="true" accessible="true" x="32" y="342" width="331" height="22" index="3"/>
    <XCUIElementTypeStaticText type="XCUIElementTypeStaticText" value="You can only connect to an existing server" name="change_server-server" label="You can only connect to an existing server" enabled="true" visible="true" accessible="true" x="32" y="382" width="239" height="15" index="4"/>
    <XCUIElementTypeButton type="XCUIElementTypeButton" name="change_server-continue" label="Next" enabled="true" visible="true" accessible="true" x="16" y="420" width="358" height="49" index="5"/>
    <XCUIElementTypeOther type="XCUIElementTypeOther" name="change_server-server" label="IntrospectionUIView&lt;UITextField>" enabled="true" visible="false" accessible="false" x="195" y="353" width="0" height="0" index="6"/>
  </XCUIElementTypeOther>
  <XCUIElementTypeOther type="XCUIElementTypeOther" value="0%" name="Vertical scroll bar, 1 page" label="Vertical scroll bar, 1 page" enabled="true" visible="true" accessible="false" x="357" y="91" width="30" height="719" index="1"/>
</XCUIElementTypeScrollView>
```

I'm not sure if this is something weird going on, but the accessibilityId is being spread to all the decorations of the text field, not just the text field that's being decorated.

It's hitting the staticText top and tail as well as the box itself - which means it's not identifying a single item so we can't use this to click on it.

This might be something odd in the way the decoration is happening - it's reannotating everything with the same accesssability ID. I'm not going to get dug into this today so adding this comment for future-me.

michaelkaye:
This is fundimentally working; i've limited where the ID is set; however i'm not understanding the swift-way to optionally call a function to set the accessibility ID - i think i'm running it twice so there's two textfields being created.

```
<XCUIElementTypeScrollView type="XCUIElementTypeScrollView" enabled="true" visible="true" accessible="false" x="0" y="0" width="390" height="844" index="0">
  <XCUIElementTypeOther type="XCUIElementTypeOther" enabled="true" visible="true" accessible="false" x="0" y="91" width="390" height="408" index="0">
    <XCUIElementTypeStaticText type="XCUIElementTypeStaticText" value="Select your server" name="Select your server" label="Select your server" enabled="true" visible="true" accessible="true" x="102" y="195" width="186" height="27" index="0"/>
    <XCUIElementTypeStaticText type="XCUIElementTypeStaticText" value="What is the address of your server? This is like a home for all your data" name="What is the address of your server? This is like a home for all your data" label="What is the address of your server? This is like a home for all your data" enabled="true" visible="true" accessible="true" x="32" y="229" width="326" height="43" index="1"/>
    <XCUIElementTypeStaticText type="XCUIElementTypeStaticText" value="Homeserver URL" name="Homeserver URL" label="Homeserver URL" enabled="true" visible="true" accessible="true" x="32" y="307" width="103" height="17" index="2"/>
    <XCUIElementTypeOther type="XCUIElementTypeOther" name="IntrospectionUIView&lt;UITextField>" label="IntrospectionUIView&lt;UITextField>" enabled="true" visible="false" accessible="false" x="195" y="353" width="0" height="0" index="3"/>
    <XCUIElementTypeTextField type="XCUIElementTypeTextField" value="matrix.org" label="" enabled="true" visible="true" accessible="true" x="32" y="342" width="331" height="22" index="4"/> <!-- this is the extra one -->
    <XCUIElementTypeTextField type="XCUIElementTypeTextField" value="matrix.org" name="change_server-server" label="" enabled="true" visible="true" accessible="true" x="16" y="382" width="358" height="23" index="5"/> <!-- this one is right -->
    <XCUIElementTypeStaticText type="XCUIElementTypeStaticText" value="You can only connect to an existing server" name="You can only connect to an existing server" label="You can only connect to an existing server" enabled="true" visible="true" accessible="true" x="32" y="412" width="239" height="15" index="6"/>
    <XCUIElementTypeButton type="XCUIElementTypeButton" name="change_server-continue" label="Next" enabled="true" visible="true" accessible="true" x="16" y="450" width="358" height="49" index="7"/>
  </XCUIElementTypeOther>
  <XCUIElementTypeOther type="XCUIElementTypeOther" value="0%" name="Vertical scroll bar, 1 page" label="Vertical scroll bar, 1 page" enabled="true" visible="true" accessible="false" x="357" y="91" width="30" height="719" index="1"/>
</XCUIElementTypeScrollView>
```

Issue #396, opened 2023-01-03
Holes in the room list


manuroe:
Some holes can appear when scrolling in the room list. There are of 2 types. We can have blank cells or default placeholders:

![IMG\_5164](https://user-images.githubusercontent.com/8418515/210395029-bb5a0534-68dc-4e2d-889b-70f19969a1b5.jpg)

Note there is a 3rd type of hole. When the [allRoomsView](https://github.com/vector-im/element-x-ios/blob/ab10278ff54de5fa9b3b9a582b78579b7451790f/ElementX/Sources/Services/Client/ClientProxy.swift#L108) is enabled, some rooms can hide and show when scrolling slowly. I can share a video in a private matrix room.

stefanceriu:
This is still happening, on "filled" rooms as it seems. Needs more debuggin, reopening.

stefanceriu:
Tracked this down to a RustSDK problem in which the pos query param isn't incremented correct and list diff operations are applied multiple times, which can create duplicates in both filled and invalidated rooms.

stefanceriu:
We have worked around this client side [here](https://github.com/vector-im/element-x-ios/commit/ee4e4a212b6d6848c079d177a6f65eac14e6264c) but the underlying issue still remains. I've added a lot more logging around the issue to at least give us an idea what to look for. The problem here is that it's reproducible but very hard to track down exactly why it happens and on what level

ara4n:
This came back with a vengance in the week before the public launch of EX, so reopening it. It looks like this sort of thing (from June 30th Nightly):

![image](https://github.com/vector-im/element-x-ios/assets/1294269/4cc55285-2d9e-4847-bd39-b755d927a867)

ara4n:
actually, i guess this is now https://github.com/vector-im/element-x-ios/issues/1157

PR #571, opened 2023-02-13
#396 - Work around duplicated room list items and add error logs

stefanceriu:


ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -205,16 +205,32 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
         MXLog.info("Updating rooms")
         
         var rooms = [HomeScreenRoom]()
+        var createdRoomIdentifiers = [String: Bool]()
+        
+        func appendRoom(_ room: HomeScreenRoom, allRoomsProvider: Bool, invalidated: Bool) {
```
pixlwave, line 210: Could the `invalidated` parameter here be read from `room` directly?

pixlwave: Also, should we add a warning here similar to the timeline:
https://github.com/vector-im/element-x-ios/blob/b0a7d05a8cb3d8a7ddad9200c2f301849ecddca3/ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift#L242

pixlwave (approved):
Looks reasonable to me :)

PR #573, opened 2023-02-13
Tweak the design of various Forms to match Figma

### LINKED ISSUE #565: FOSDEM implementation inconsistencies
1. Background colour - for full screen pages the background colour should be the lightest colour \(Compound/Light/Background\) and the opposite in dark mode. For modals, the background colour should be the next lightest colour \(Compound/Light/System Light\) and the opposite in dark mode. Currently there are inconsistencies on the settings and room settings pages across both themes - I’ve updated our new file to match this so you can use it as the source of truth.  This background colour should be the same across all pages and modals of the app - is it possible to set this as a rule instead of individually adjusting them?

Settings: https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X-\(new\)?node-id=5%3A7387
 Room settings: https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X-\(new\)?node-id=5%3A7388

Room list header - this should have the same opacity as the All Chats header if possible?

[Room settings page](https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X-(new)?node-id=5%3A7388\) - we are missing the Matrix ID and Topic from the implementation. I think Gil worked to add these, so I’m not sure why they’re not in there. Could you add them? There are also some changes needed for text weight and size of the ‘Topic’, ‘About’ and ‘Security’ headers. Again, the file is accurate and the source of truth for these.

pixlwave:
Mainly small tweaks, notable changes:
- `formBackground` and `formRowBackground` colours added as these are special cases where we use different colours in light/dark modes - (`SettingsScreen` was doing this internally but not the others).
- Replaced a couple of broken instances of introspection with `.scrollContentBackground` modifiers.
- Modified one room details UI test to include buttons and topic.

Fixes #565.

| Settings | Bug Report | Room Details |
| - | - | - |
| Screenshot 2023-02-13 at 2 25 01 pm | Screenshot 2023-02-13 at 2 24 53 pm | Screenshot 2023-02-13 at 2 24 45 pm |


Velin92 (approved):
LGTM!

I am wondering if we should add UI Tests also for dark mode... we run them 4 times already which is quite a lot, but we might need a dark mode on too?

stefanceriu (approved):

pixlwave:
> I am wondering if we should add UI Tests also for dark mode... we run them 4 times already which is quite a lot, but we might need a dark mode on too?

That's definitely something to think about.

PR #576, opened 2023-02-14
Refactored UserNotification into UserIndicator

Velin92:
replaced UserNotification and all its mentions/references as UserIndicator


pixlwave (approved):
LGTM 😎

PR #577, opened 2023-02-14
Disable broken UI Tests and remove navigation.

pixlwave:
- Disables broken tests for now
- Removes navigation instead opening the app directly.
- Adds a 60-second timeout on UI tests.
- Enabled parallel execution of English and German when running locally.


UITests/Sources/Application.swift
```diff
@@ -72,7 +64,11 @@ extension XCUIApplication {
     }
 
     private var deviceName: String {
-        UIDevice.current.name
+        var name = UIDevice.current.name
+        if name.starts(with: "Clone "), let range = name.range(of: " of ") {
```
stefanceriu, line 68: Should probably leave a comment here to explain what this is about

ElementX/Sources/Other/Tests.swift
```diff
@@ -34,4 +34,13 @@ public enum Tests {
         false
         #endif
     }
+    
+    /// The identifier of the screen to be loaded when running UI tests.
+    static var screenID: UITestsScreenIdentifier? {
+        #if DEBUG
+        ProcessInfo.processInfo.environment["UI_TESTS_SCREEN"].flatMap(UITestsScreenIdentifier.init)
```
stefanceriu, line 41: Nice! 👍

stefanceriu (approved):
Looks good appart from those XCUIElement changes

PR #578, opened 2023-02-14
Attach Screenshot in Bug Report Screen

Velin92:
Added an Attach Screenshot button that will add a screenshot to the bug report:

https://user-images.githubusercontent.com/34335419/219050822-84f27815-5dae-4c15-9d73-ed6a68e0bdda.MP4


ElementX/Sources/Screens/BugReport/BugReportViewModel.swift
```diff
@@ -52,11 +54,18 @@ class BugReportViewModel: BugReportViewModelType, BugReportViewModelProtocol {
         let progressTracker = ProgressTracker()
         callback?(.submitStarted(progressTracker: progressTracker))
         do {
+            var files: [URL] = []
+            if let screenshot = context.viewState.screenshot {
+                let imageURL = URL(fileURLWithPath: NSTemporaryDirectory()).appendingPathComponent("Screenshot.png")
```
pixlwave, line 59: Small nicety:
```suggestion
                let imageURL = URL.temporaryDirectory.appendingPathComponent("Screenshot.png")
```

ElementX/Sources/Screens/BugReport/View/BugReportScreen.swift
```diff
@@ -113,6 +126,21 @@ struct BugReportScreen: View {
             }
             .padding(.vertical, 16)
             .padding(.horizontal, 16)
+        } else {
+            PhotosPicker(
+                selection: $selectedScreenshot,
+                matching: .screenshots,
+                photoLibrary: .shared()
+            ) {
```
pixlwave, line 134: Sorry, style:
```suggestion
            PhotosPicker(selection: $selectedScreenshot,
                         matching: .screenshots,
                         photoLibrary: .shared()) {
```

pixlwave (approved):
Looks good to me. One small comment from the screen recording, could the progress bar say "Sending" instead of loading?

Also should probably ask for a design review on this :)

ElementX/Sources/Screens/BugReport/View/BugReportScreen.swift
pixlwave: I've created https://github.com/vector-im/element-x-ios/pull/579

stefanceriu (approved):
lgtm too 👍

amshakal (changes requested):
Hey!! Thanks for working on this. I have a few suggestions/tweaks:
1. Is it possible be reduce the size of the text box so that the info/buttons below it are more visible?
2. Can we add the 'attach screenshot' button above 'send logs'? It fit's better with the overall heiarchy of this screen.
3.  Can we add an icon to the left of attach screenshot to make it super obvious what the user is supposed to do here, it also matches with the treatment we are giving buttons within settings. 

Screenshot for ref:
Screenshot 2023-02-15 at 12 11 30 pm

Figma file for ref:
https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X-(new)?node-id=123%3A21855&t=cBA1DmfD4awGuMCW-1

ElementX/Sources/Other/SwiftUI/Views/SettingsDefaultRow.swift
```diff
@@ -26,24 +26,21 @@ struct SettingsDefaultRow: View {
...
-                    .font(.element.body)
-                    .foregroundColor(.element.primaryContent)
+                Label(title: {
+                    Text(title)
+                }, icon: {
+                    image
+                })
+                .labelStyle(RowLabelStyle(titleColor: .element.primaryContent,
+                                          iconColor: .element.systemGray,
+                                          backgroundColor: .element.systemGray6))
```
pixlwave, line 43: Seems these 2 colours were wrong. They should be `secondaryContent` and `formBackground` just like you did in the bug report screen. Can we update this and hard code them into the style?

ElementX/Sources/Other/SwiftUI/Views/RowLabelStyle.swift
```diff
@@ -0,0 +1,39 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import SwiftUI
+
+struct RowLabelStyle: LabelStyle {
```
pixlwave, line 19: Should probably call this `SettingsRowLabelStyle` to match the others. I think at some point it might make sense to rename all of the `Settings...` items to `Form...` to be less specific.

pixlwave (approved):
Changes look good to me. There's a couple more places where we could use the label style: `RoomDetailsScreen` in the `aboutSection` and the `securitySection`.

amshakal (approved):
💯

ElementX/Sources/Other/SwiftUI/Views/SettingsRowLabelStyle.swift
```diff
@@ -16,24 +16,25 @@
...
-    let backgroundColor: Color
-    
-    @ScaledMetric private var menuiconSize = 30.0
+struct SettingsRowLabelStyle: LabelStyle {
+    @ScaledMetric private var menuIconSize = 30.0
+    private let alignment: VerticalAlignment
+
+    init(alignment: VerticalAlignment = .center) {
+        self.alignment = alignment
+    }
```
pixlwave, line 25: You could just go for
```
var alignment: VerticalAlignment = .center
```
and use the synthesised init here.

Personally I'd be tempted to use `.firstTextBaseline` as the default too.

PR #579, opened 2023-02-14
Add SwiftFormat wraparguments rule.

pixlwave:
Matches the wrapparameters rule, seems it was missed and left an inconsistency between `func` and `init` styles.

Maybe controversial, I've disabled formatting on the Package manifests.


stefanceriu (approved):
👍

Issue #534, opened 2023-02-06
Fix UI tests


stefanceriu:
Following recents changes the UI tests stopped working. We need to investigate and fix them

PR #585, opened 2023-02-15
Use files instead of UDP for signalling.

pixlwave:
Turns out a big part of the issue was the use of the continuation to `await receive()`. This implementation switches to use a publisher instead. I stayed with the file-based approach after discovering this as it simplifies the code a fair amount.

Keeping an eye on this run for the UI Tests: https://github.com/vector-im/element-x-ios/actions/runs/4183990641


stefanceriu (approved):
Looks great, nicely done! 👏

pixlwave:
UI Tests failed (but on a different one) so I'll ignore that and merge this. Will leave #534 open I guess.

PR #586, opened 2023-02-15
Fixed UI test problems introduced after the accessibility identifier …

stefanceriu:
…changes

### DIFF
--- a/ElementX/Sources/UITests/UITestsAppCoordinator.swift
+++ b/ElementX/Sources/UITests/UITestsAppCoordinator.swift
@@ -34,13 +34,13 @@ class UITestsAppCoordinator: AppCoordinatorProtocol {
     }
     
     func start() {
+        Bundle.elementFallbackLanguage = "en"
+        
         guard let screenID = Tests.screenID else { fatalError("Unable to launch with unknown screen.") }
         
         let mockScreen = MockScreen(id: screenID)
         navigationRootCoordinator.setRootCoordinator(mockScreen.coordinator)
         self.mockScreen = mockScreen
-        
-        Bundle.elementFallbackLanguage = "en"
     }
     
     func toPresentable() -> AnyView {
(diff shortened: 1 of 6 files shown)


Velin92 (changes requested):
Maybe would be better to remove the line `"parallelizable" : true` from the UI tests .xctestplan to avoid multiple UI tests to run in parallel, which seems to break the testing in Xcode. (At least for me locally).

Also the following screenshot test seems to be failing:
```
class OnboardingUITests: XCTestCase {
    func testInitialStateComponents() {
        let app = Application.launch(.onboarding)
        app.assertScreenshot(.onboarding)
    }
```

stefanceriu:
> Maybe would be better to remove the line `"parallelizable" : true` from the UI tests .xctestplan to avoid multiple UI tests to run in parallel, which seems to break the testing in Xcode. (At least for me locally).

I think we should, doesn't work properly for me either

stefanceriu:
> Also the following screenshot test seems to be failing:

Seems we have broken our `elementFallbackLanguage` in the UITests refactoring

Screenshot 2023-02-16 at 08 56 47

ElementX/Sources/UITests/UITestsAppCoordinator.swift
```diff
@@ -34,13 +34,13 @@ class UITestsAppCoordinator: AppCoordinatorProtocol {
     }
     
     func start() {
+        Bundle.elementFallbackLanguage = "en"
```
pixlwave, line 37: Whoops sorry, didn't think about that!

UITests/Sources/Application.swift
```diff
@@ -24,7 +24,6 @@ struct Application {
             "IS_RUNNING_UI_TESTS": "1",
             "UI_TESTS_SCREEN": identifier.rawValue
         ]
-        Bundle.elementFallbackLanguage = "en"
```
pixlwave, line 27: I think this is about setting the fallback language in the tests runner? Not sure it really matters with our strings though.

pixlwave (approved):
LGTM

UITests/Sources/Application.swift
pixlwave: Changed my mind, we probably do need to keep this to match a label value in the tests to what is shown in the app when running in German.

pixlwave (approved):

PR #587, opened 2023-02-15
Report Content

Velin92:
- [x] Implemented the report content UI
- [x] implemented the report content with reason UI
- [x] Implemented the logic to send the report (NOTE: Only works on main branch of the Matrix Rust SDK As of right now)

https://user-images.githubusercontent.com/34335419/219120343-c8a1c403-c7af-4d00-a288-d0f7e0d9720a.MP4


Velin92:
@amshakal added a screen recording of the current UI for reporting content that I have implemented so far.
I am using the existing Element-X context menu to add such option, while for the Report Content reason I am using the old Element Alert style UI.
Ofc these are just temporary solutions to start testing and working on the implementation, but these can be improved/changed if necessary.

Velin92:
@amshakal confirmed that so far the UI is okay for her!
The idea is that we can split the report content UI and implementation from the report content informative UI (success in sending the report/informing what should/could happen after the report).

So I'll work on those on another PR once such designs are ready

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -120,6 +120,10 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
             await markRoomAsRead()
         case .contextMenuAction(let itemID, let action):
             processContentMenuAction(action, itemID: itemID)
+        case let .report(itemID, reason):
+            Task {
```
stefanceriu, line 124: You might not need a task here as you're already in an async context

ElementX/Sources/Other/Extensions/Alert.swift
```diff
@@ -0,0 +1,45 @@
...
+// limitations under the License.
+//
+
+import SwiftUI
+
+protocol AlertItem {
+    var title: String { get }
+}
+
+extension View {
```
stefanceriu, line 23: Nice 👍

ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift
```diff
@@ -187,6 +187,16 @@ class RoomTimelineController: RoomTimelineControllerProtocol {
...
         }
     }
+
+    func reportContent(_ itemID: String, reason: String?) async {
+        MXLog.info("Send report content in \(roomID)")
+        switch await roomProxy.reportContent(itemID, reason: reason) {
+        case .success:
+            MXLog.info("Finished reporting content")
+        case .failure(let error):
+            MXLog.error("Failed reporing content with error: \(error)")
```
stefanceriu, line 197:
```suggestion
            MXLog.error("Failed reporting content with error: \(error)")
```

stefanceriu (approved):
lgtm, nicely done 👏 
Let's wait until we release a new SDK version before merging it

PR #588, opened 2023-02-16
Fix integration tests

stefanceriu:


Velin92 (approved):
aside from the weird Danger error, LGTM!

PR #589, opened 2023-02-16
Add required state on subscribe.

pixlwave:
Small PR to request the room's topic and canonical alias for display in the room details screen. Its a little bit slow initially, but remembers it when you've opened the room once.


ElementX/Sources/Other/Logging/RustTracing.swift
```diff
@@ -48,7 +49,8 @@ struct TracingConfiguration {
         .matrix_sdk_crypto: .debug,
         .matrix_sdk_http_client: .debug,
         .matrix_sdk_sliding_sync: .trace,
-        .matrix_sdk_base_sliding_sync: .trace
+        .matrix_sdk_base_sliding_sync: .trace,
+        .matrix_sdk_room_timeline: .info
```
pixlwave, line 53: Added this while I was here. Presumably `.info` is the common level so this is how it would have been before being added to the configuration?

stefanceriu (approved):
Noice! 👍

PR #590, opened 2023-02-16
Add identifiers for verification cross signing

michaelkaye:


michaelkaye:
Running into a few problems automating this:

#591  - worked around by stripping "https://" and "http://" from the URLs in the adapter.

#594  - kinda blocked; i can probably create a room in the automation if that's what's needed to complete the process, but this feels like some sort of bug. ele-web <-> ele-web seems fine though.

pixlwave (approved):

PR #597, opened 2023-02-17
Auto-Mocking with Sourcery

Velin92:
Added the brew package Sourcery.

This allows to essentially use a stencil file (I am using the one provided by Sourcery itself with the added swiftlint disable directive) to generate auto-mocks for any protocol that is marked with a mark protocol called AutoMockable (the protocol itself is just an empty protocol is only used as a marker).

Generated mocks for the BugReport service and showed a simple implementation for how to use such mock.
Also created a special mock for the SessionVerificationControllerProxy where we are able to inject more custom behaviour in the mocks.

Replaced these two former mocks in the code with the newer ones


ElementX/Sources/Mocks/Generated/GeneratedMocks.swift
```diff
@@ -0,0 +1,195 @@
...
+
+
+
+
+
+
+
+
+
+
```
pixlwave, line 36: What's this about, can we remove it?

ElementX/Sources/Mocks/AutoMockable.swift
```diff
@@ -14,18 +14,6 @@
...
-        SubmitBugReportResponse(reportUrl: "https://www.example/com/123")
-    }
-
-    var crashedLastRun = false
-
-    func crash() {
-        // no-op
-    }
-}
+// This protoclo is used only as a marker
```
pixlwave, line 17:
```suggestion
// This protocol is used only as a marker
```

ElementX/Sources/Mocks/MockSessionVerificationControllerProxy.swift
```diff
@@ -0,0 +1,88 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Combine
+
+final class MockSessionVerificationControllerProxy: SessionVerificationControllerProxyProtocolMock {
```
pixlwave, line 19: Rather than introducing another type here, is it possible to make a static property/method on `SessionVerificationControllerProxyProtocolMock` that returns a pre-configured `SessionVerificationControllerProxyProtocolMock`? It feels very confusing to have `MockSomething` and `SomethingMock`.

Sourcery/AutoMockable.stencil
```diff
@@ -0,0 +1,228 @@
+// swiftlint:disable all
```
pixlwave, line 1: Is it possible to move these into `Tools/Sourcery` to keep the project root tidy (well tidier anyway, its not exactly clean)?

Sourcery/AutoMockable.stencil
```diff
@@ -0,0 +1,228 @@
...
+
+import Foundation
+#if os(iOS) || os(tvOS) || os(watchOS)
+import UIKit
+#elseif os(OSX)
+import AppKit
+#endif
```
pixlwave, line 8: Can probably simplify this somewhat...
```suggestion
import SwiftUI
```

UnitTests/Sources/BugReportServiceTests.swift
```diff
@@ -19,7 +19,13 @@ import Foundation
 import XCTest
 
 class BugReportServiceTests: XCTestCase {
-    let bugReportService = MockBugReportService()
+    var bugReportService: BugReportServiceProtocolMock!
+
+    override func setUpWithError() throws {
+        bugReportService = BugReportServiceProtocolMock()
+        bugReportService.underlyingCrashedLastRun = false
+        bugReportService.submitBugReportProgressListenerReturnValue = SubmitBugReportResponse(reportUrl: "https://www.example/com/123")
```
pixlwave, line 27: I'm guessing this return value isn't actually checked?
```suggestion
        bugReportService.submitBugReportProgressListenerReturnValue = SubmitBugReportResponse(reportUrl: "https://www.example.com/123")
```

UnitTests/Sources/SessionVerificationViewModelTests.swift
```diff
@@ -32,13 +32,13 @@ class SessionVerificationViewModelTests: XCTestCase {
...
 
-    func testRequestVerification() async {
+    func testRequestVerification() async throws {
         XCTAssertEqual(context.viewState.verificationState, .initial)
         
         context.send(viewAction: .requestVerification)
         
-        await Task.yield()
-        
+        try await Task.sleep(for: .seconds(1))
```
pixlwave, line 40: Won't 100 milliseconds do here like the rest?

pixlwave (commented):
Ok, this looks pretty cool 😎
I've added some thoughts inline.

ElementX/Sources/Mocks/Generated/GeneratedMocks.swift
Velin92: I think I can, I just need to play around with the stencil, I actually already tweaked the stencil we are using (I am using the official one from the Sourcery repo here: https://github.com/krzysztofzablocki/Sourcery/blob/master/Templates/Templates/AutoMockable.stencil ) to include stuff like `// swiftlint: disable all`.

These stencil files need to be periodically checked on the original repo in case they make new improvements for them.
Of course we make our own tweaks but then we might need to diff them with the newer ones the avoid conflicts.
Actually there is ALREADY a an update for this stencil file, so I might update it before the PR is approved.

ElementX/Sources/Mocks/Generated/GeneratedMocks.swift
```diff
@@ -0,0 +1,195 @@
...
+class BugReportServiceProtocolMock: BugReportServiceProtocol {
```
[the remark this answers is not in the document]
Velin92: Uhm okay this might be a bit complicated, need to check how the .stencil actually do work, I imagine that we want to check if the word Protocol is at the end of word and remove it if present.

Sourcery/AutoMockable.stencil
Velin92: Yeah I guess it's not a problem at all, is just a matter of moving the file and changing the directory of the config for the run script.

Sourcery/AutoMockable.stencil
Velin92: Honestly I think we could even remove it.
In the end I am already able to decide what needs to be imported from the configuration file, that should be enough.

UnitTests/Sources/SessionVerificationViewModelTests.swift
Velin92: It's because by default the mock uses a wait delay of `.seconds(1)` (it's the same we were using before transitioning to the auto generated ones)

pixlwave: Oh, it runs fine on my side when set to `.milliseconds(100)`. Does that call count get incremented before the wait?

Velin92:
```
func requestVerification() async -> Result<Void, SessionVerificationControllerProxyError> {
        requestVerificationCallsCount += 1
        if let requestVerificationClosure = requestVerificationClosure {
            return await requestVerificationClosure()
        } else {
            return requestVerificationReturnValue
        }
    }
```

Actually yeah, by default call increments are always called before their customising closure. Probably was just a matter of the yield being a bit too fast

ElementX/Sources/Mocks/Generated/GeneratedMocks.swift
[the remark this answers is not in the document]
Velin92: done!

ElementX/Sources/Mocks/MockSessionVerificationControllerProxy.swift
Velin92: done

pixlwave (approved):
Nice, LGTM 😎

PR #598, opened 2023-02-18
Fixes #594 - Do not disable the session verification banner if the ro…

### LINKED ISSUE #594: Unable to click "Continue" to get access to encrypted messages for entirely clean newly registered user.
### Steps to reproduce

1. Register a user against a server. Perform no actions on that new user except to create it. (Specifically, do not create a room). Leave the element-web stable client running so we'll be able to share keys later.
2. Login on element X iOS with the credentials above
3. Attempt to click the "Continue" button

(Nothing happens)

Looking at the appium source, the button is tagged `enabled=false`, so I imagine clicking won't help.

waiting (up to 90s) does not enable the button.

https://user-images.githubusercontent.com/1917473/219683111-2d96b06d-0bf6-4542-a8c3-f7ae1fa0de3b.mp4

#### What did you expect?
To be able to share keys from the first login.

#### What happened instead?

Never able to click the button.

I have been able to login to an existing user with a room already created and share credentials there.

Logs are available on request via browserstack (however they have my test username/password embedded in them so will not supply public link here). Am also able to share the username/password if developers wish to try locally.

Practically this isn't a big issue; most users will create a room and use the device a little on element-web before they try to sign in on their phone; however, if you don't create that room then the entire app appears frozen/broken with no way forward.

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

_No response_

### Homeserver

_No response_

### Will you send logs?

Yes

stefanceriu:
…om list is the empty or loading state

### DIFF
--- a/ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
+++ b/ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
@@ -38,10 +38,10 @@ struct HomeScreen: View {
                     ForEach(context.viewState.visibleRooms) { room in
                         HomeScreenRoomCell(room: room, context: context)
                             .redacted(reason: .placeholder)
-                            .disabled(true)
                     }
                 }
                 .shimmer()
+                .disabled(context.viewState.roomListMode == .skeletons)
             } else {
                 LazyVStack(spacing: 0) {
                     ForEach(context.viewState.visibleRooms) { room in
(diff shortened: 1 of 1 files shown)


Velin92 (approved):
LGTM, but next time avoid such big PRs pls xD

ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
```diff
@@ -75,7 +75,6 @@ struct HomeScreen: View {
             }
         }
         .scrollDismissesKeyboard(.immediately)
-        .disabled(context.viewState.roomListMode == .skeletons)
```
pixlwave, line 78: Sorry I'm too slow on this comment, but this change makes the skeletons scrollable. Don't think we want that.

PR #601, opened 2023-02-20
State event collapsing

stefanceriu:
This PR mainly focuses on state even collapsing:
* it will group more than 1 consecutive room state events together (and only room state events together)

It also cleans up a couple of things:
* take out bubble grouping from the timeline items
* removes the RoomTimelineViewFactory
* moves bubble computations in the RoomScreenViewModel and passes those down to the timelineStyles through an environment variable
* makes the state grouping configurable through a developer option menu
* allows only in-memory storage on the UserSetting property wrappers

https://user-images.githubusercontent.com/637564/220632741-0127f99d-1830-42e0-9e79-ad7f9c04e0bd.mp4


ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineViewProvider.swift
```diff
@@ -78,43 +121,74 @@ enum RoomTimelineViewProvider: Identifiable, Hashable {
...
+             .notice(_, let grouping),
+             .redacted(_, let grouping),
+             .encrypted(_, let grouping),
+             .readMarker(_, let grouping),
+             .paginationIndicator(_, let grouping),
+             .sticker(_, let grouping),
+             .unsupported(_, let grouping),
+             .timelineStart(_, let grouping),
+             .state(_, let grouping),
+             .group(_, let grouping):
```
pixlwave, line 190: Ohhh this is a neater form of the switch. Could also do the same on the `id` switch with a cast:
```
.emote(let item as RoomTimelineItemProtocol)
```

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineViewProvider.swift
```diff
@@ -18,53 +18,96 @@ import Foundation
...
+    case readMarker(ReadMarkerRoomTimelineItem, TimelineStyleGrouping)
+    case paginationIndicator(PaginationIndicatorRoomTimelineItem, TimelineStyleGrouping)
+    case sticker(StickerRoomTimelineItem, TimelineStyleGrouping)
+    case unsupported(UnsupportedRoomTimelineItem, TimelineStyleGrouping)
+    case timelineStart(TimelineStartRoomTimelineItem, TimelineStyleGrouping)
+    case state(StateRoomTimelineItem, TimelineStyleGrouping)
+    case group(GroupTimelineItem, TimelineStyleGrouping)
+    
+    // swiftlint:disable:next cyclomatic_complexity
+    init(timelineItem: RoomTimelineItemProtocol, grouping: TimelineStyleGrouping = .single) {
```
pixlwave, line 39: The default value leaves me a bit wary about it being ignored in places. Am I right in thinking we only plan to create these items in a single place?

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineStyle.swift
```diff
@@ -32,21 +32,50 @@ enum TimelineStyle: String, CaseIterable {
     }
 }
 
+enum TimelineStyleGrouping: Hashable {
```
pixlwave, line 35: I'm not convinced on the name of this, mainly due to the clash with groupable items. Should we remove grouping from the name entirely and call it something like `TimelineMessageStyle`, `TimelineDisplayState`, `TimelineCellStyle`, `TimelineConversationState`, `TimelineContinuityState` or something? Lets bikeshed offline.

ElementX/Sources/Screens/RoomScreen/View/Timeline/GroupRoomTimelineView.swift
```diff
@@ -0,0 +1,72 @@
...
+
+struct GroupRoomTimelineView_Previews: PreviewProvider {
+    static var previews: some View {
+        let item = GroupTimelineItem(items: [SeparatorRoomTimelineItem(text: "This is a separator"),
+                                             SeparatorRoomTimelineItem(text: "This is another separator")])
+        GroupRoomTimelineView(timelineItem: item)
+    }
+}
+
+struct CustomDisclosureGroupStyle<Label: View>: DisclosureGroupStyle {
```
pixlwave, line 50: I'm guessing this needed to be generic in a previous implementation?
```suggestion
struct CustomDisclosureGroupStyle: DisclosureGroupStyle {
```

pixlwave: Also, needs a more descriptive name :)

ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift
```diff
@@ -218,62 +218,31 @@ class RoomTimelineController: RoomTimelineControllerProtocol {
...
-                        createdIdentifiers[timelineItem.id] = true
-                    } else {
-                        MXLog.error("Found duplicated timeline item, ignoring")
-                    }
+        
+        let itemGroups = timelineProvider.itemsPublisher.value.groupBy { isItemGroupable($0) }
+        
+        for (index, itemGroup) in itemGroups.enumerated() {
+            if itemGroup.count == 1, let itemProxy = itemGroup.first {
+                let isLastItem = index == itemGroups.indices.last
```
pixlwave, line 232: <del>This value is always true when `itemGroup.count == 1`.</del>
Ok looking elsewhere I get it now, its for the very last item on the timeline to hide the read marker. I don't think we should do this on our side tbh. Would be worth making an issue in the Rust SDK instead to make sure it doesn't happen.

pixlwave (commented):
This looks like a great start 👏 I've added some thoughts inline

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineViewProvider.swift
stefanceriu: Good idea 👍

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineViewProvider.swift
stefanceriu: Unfortunately it needs to be in 2 places: the view model and the `GroupRoomTimelineView`. So I think you're right, better to make it explicit

ElementX/Sources/Screens/RoomScreen/View/Timeline/GroupRoomTimelineView.swift
stefanceriu: I went with `CollapsibleRoomTimelineItemDisclosureGroupStyle` ¯\_(ツ)_/¯

ElementX/Sources/Other/Extensions/Array.swift
```diff
@@ -0,0 +1,67 @@
...
+    func groupBy(_ isGroupable: (Element) -> Bool) -> [[Element]] {
+        var newItems = [[Element]]()
+        
+        var itemsToBeGrouped = [Element]()
+        
+        for (index, currentItem) in enumerated() {
+            let previousItem = self[safe: index - 1]
+            let nextItem = self[safe: index + 1]
+            
+            if isGroupable(currentItem) {
```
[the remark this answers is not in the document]
stefanceriu: > maybe it would make sense to extract

This is actually trickier than expected as Element would need to conform to Equatable for `itemsGroupingAllowed.contains` to work, and our TimelineItems do not



> might be worth trying out Apple's [Algorithms](https://github.com/apple/swift-algorithms) and using `chunked(by:)`
Tried it out but doesn't produce the results we're expecting. On the other hand it works beautifuly on computing the bubble grouping

ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift
stefanceriu: Rust ticket here https://github.com/matrix-org/matrix-rust-sdk/issues/1546

stefanceriu:
@amshakal this PR is now finally ready for design review. Please see the attached video but do let me know if you need a build you can play around with or more screenshots.

amshakal (approved):
It's great! Thank you.

PR #604, opened 2023-02-20
Associated Domains 

Velin92:
For now I just added all the domains of the current El-iOS app, and opened up an issue [here](https://github.com/vector-im/element.io-website/issues/137) to ask the website to update the apple-app-associated-domains file on the website and on all related website that we still host and have one.

In case some of this website do not have such file anymore, we can remove them later from the entitlments, now just in case, for backwards compatibility let's keep all of them.

### DIFF
--- a/ElementX/SupportingFiles/ElementX.entitlements
+++ b/ElementX/SupportingFiles/ElementX.entitlements
@@ -4,6 +4,20 @@
 <dict>
 	<key>aps-environment</key>
 	<string>development</string>
+	<key>com.apple.developer.associated-domains</key>
+	<array>
+		<string>applinks:www.vector.im</string>
+		<string>applinks:vector.im</string>
+		<string>applinks:riot.im</string>
+		<string>applinks:www.riot.im</string>
+		<string>webcredentials:riot.im</string>
+		<string>applinks:element.io</string>
+		<string>applinks:app.element.io</string>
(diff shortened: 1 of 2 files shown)


ElementX/SupportingFiles/ElementX.entitlements
```diff
@@ -4,6 +4,20 @@
 <dict>
 	<key>aps-environment</key>
 	<string>development</string>
+	<key>com.apple.developer.associated-domains</key>
+	<array>
+		<string>applinks:www.vector.im</string>
+		<string>applinks:vector.im</string>
+		<string>applinks:riot.im</string>
+		<string>applinks:www.riot.im</string>
+		<string>webcredentials:riot.im</string>
```
pixlwave, line 13: Presumably these ones can go given they're redirects, or did I miss something here?

Velin92: Yeah I guess they can go, however just to be extra sure in this [issue](https://github.com/vector-im/element.io-website/issues/137) I asked if we still keep an apple-app-associated domain for these websites.

Velin92: Double checked with @manuroe they should be safe to remove, also the old websites seem to be just redirects as of right now.

pixlwave (approved):
LGTM 🔗

PR #606, opened 2023-02-20
Removed Edit/Reply Dimming

Velin92:
Figma is not requiring such dimming, so we are removing it.

Waiting confirmation from design, before making the PR ready.

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
+++ b/ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
@@ -196,14 +196,12 @@ class TimelineTableViewController: UIViewController {
             
             // A local reference to avoid capturing self in the cell configuration.
             let coordinator = self.coordinator
-            let opacity = self.opacity(for: timelineItem)
-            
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
```diff
@@ -351,12 +349,6 @@ class TimelineTableViewController: UIViewController {
         
         coordinator.send(viewAction: .paginateBackwards)
     }
-    
-    /// Returns the opacity that the supplied timeline item's cell should be.
-    private func opacity(for item: RoomTimelineViewProvider) -> CGFloat {
-        guard case let .reply(selectedItemID, _) = composerMode else { return 1.0 }
```
pixlwave, line 357: You can remove the `didSet` on `composerMode too now :)

Velin92: done!

pixlwave (approved):
LGTM 👍

PR #611, opened 2023-02-22
Fix visible room sliding sync view ranges changing while inside rooms

stefanceriu:
**Some room** list cells were getting appearance calls while the user was in the room screen. This would incorrectly trigger sliding sync range updates and making unnecessary requests.
This PR tracks the view's visibility and ignores said changes


pixlwave (approved):
LGTM!

PR #613, opened 2023-02-23
Fix navigationStackCoordinator getting torn down when being reset on …

stefanceriu:
…the navigationSplitCoordinator, added unit tests for it

Navigating to rooms in split mode is currently broken. This happens because we set the `detailNavigationStackCoordinator` multiple times as the `splitCoordinator`'s detail. This in turn make the `splitCoordinator` tear down its previous detail which just happens to be the same as the current one. As the detail is a `navigationStack` that makes it reset itself and release all coordinators

As Coordinators are not currently equatable there's no good way to prevent this on the navigation coordinators level so instead we check against resetting that detail multiple times to the same instance

### DIFF
--- a/ElementX/Sources/Application/Navigation/NavigationCoordinators.swift
+++ b/ElementX/Sources/Application/Navigation/NavigationCoordinators.swift
@@ -138,6 +138,10 @@ class NavigationSplitCoordinator: CoordinatorProtocol, ObservableObject, CustomS
             return
         }
         
+        if sidebarModule?.coordinator === coordinator {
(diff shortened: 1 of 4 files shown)


Velin92:
More of a question: why we can't just compare the references of the two coordinators by using the === operator? There is a case where even if they are two different coordinators they could still be considered the same coordinator according to some specific internal state?

pixlwave (approved):
LGTM 🚀

stefanceriu:
> why we can't just compare the references of the two coordinators by using the `=== `operator

That's a very very good idea 👏  I'm going to give it a shot

Issue #615, opened 2023-02-23
timeline gets stuck blank, then wedges, then unwedges when you send a message after ss expiration


ara4n:
### Steps to reproduce

1. Use a big account (maybe? it's 100% repro for Matthew, but may not be related to large accounts)
2. View a given room
3. Bg the app and wait for SS to expire
4. Fg the app, observe the room goes blank and doesn't refresh its timeline
5. Switch to roomlist and back again, observe that the old content of the room from step 2 is shown.
6. Send a message in the room, observe that the room contents suddenly updates.

#### What did you expect?

Consistent room history

### Application version

10.21 (135) nightly

### Will you send logs?

Yes

ara4n:
current working theory is that we're dropping the subscription for the room somehow and not reestablishing it.

this is 100% repro for me.

stefanceriu:
Fixed in #618 and #628

PR #618, opened 2023-02-23
Manually cancel timeline listeners when a RoomScreen is torn down ins…

stefanceriu:
…tead of waiting for a delayed deinit

Potential (partial?) fix for https://github.com/vector-im/element-x-ios/issues/615


pixlwave (approved):
LGTM

PR #619, opened 2023-02-23
Add UI_TESTS_SCREEN as a disabled env var for quicker debugging.

pixlwave:
Removes the old `IS_RUNNING_UI_TESTS` and instead checks if the screen was set.


stefanceriu (approved):
👌

PR #628, opened 2023-02-24
Sliding sync session reset manual backpagination

stefanceriu:
Request backpagination whenever the timeline item list is empty. This can happen when first loading loading the timeline or after a sliding sync session reset


pixlwave (approved):
🥳

PR #633, opened 2023-02-24
Cached avatar support (from Rust)

Velin92:
For now this PR will not work on develop so opening it as a draft.

Added a new API to expose a cached version of the url, and use it by default always if available, and then do the actual real avatar url request, and override it if it is successful.

Also changed the exposed avatar_url function to return an optional string to not treat the case where the avatar_url is missing as an error case.


ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -91,9 +98,7 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
                 // Delay user profile detail loading until after the initial room list loads
                 if roomListMode == .rooms {
                     Task {
-                        if case let .success(url) = await userSession.clientProxy.loadUserAvatarURL() {
-                            self.state.userAvatarURL = url
-                        }
+                        await userSession.clientProxy.loadUserAvatar()
```
Velin92, line 101: @pixlwave we always delay the load user avatar function until the initial room list loads, however this means that the avatar will not appear if the user launches the app offline, even if now we cache it.
Could I move this outside the first load?

Velin92: The other option would be to have two different functions, one for the cached avatar and one for the normal avatar API call, and I could use the cached one outside the sink, and the API call one inside the sync

Velin92: what concerns me is that we never do a retry for the avatar fetching... so let's say there is no cache and no connections the user launches the app, they are offline, they then have finally some connection and get the first rooms, the avatar would not get updted if I place this outside the sink. On the other end placing it inside the sync means loading this thing only when we have connection.

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -58,6 +58,13 @@ class ClientProxy: ClientProxyProtocol {
     
     var allRoomsSlidingSyncView: SlidingSyncList?
     var allRoomsSummaryProvider: RoomSummaryProviderProtocol?
+
+    private let avatarUrlSubject = CurrentValueSubject<URL?, Never>(nil)
+    var avatarUrlPublisher: AnyPublisher<URL?, Never> {
```
pixlwave, line 63:
```suggestion
    private let avatarURLSubject = CurrentValueSubject<URL?, Never>(nil)
    var avatarURLPublisher: AnyPublisher<URL?, Never> {
```

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -154,23 +161,34 @@ class ClientProxy: ClientProxyProtocol {
...
-                return .success(url)
-            } catch {
-                return .failure(.failedRetrievingAvatarURL)
             }
         }
     }
+
+    func loadUserAvatar() async {
+        await loadUserAvatar(policy: .default)
+    }
```
pixlwave, line 191: Could we move this in an extension of `ClientProxyProtocol` so that it doesn't need to be implemented?

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -54,6 +54,17 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
                 }
             }
             .store(in: &cancellables)
+
+        userSession.clientProxy.avatarUrlPublisher
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] url in
+                self?.state.userAvatarURL = url
+            }
+            .store(in: &cancellables)
```
pixlwave, line 63: Nice 😎

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -154,23 +161,34 @@ class ClientProxy: ClientProxyProtocol {
...
+        if policy != .cacheOnly {
+            await Task.dispatch(on: clientQueue) {
+                do {
+                    if let urlString = try self.client.avatarUrl() {
+                        self.avatarUrlSubject.value = URL(string: urlString)
+                    } else {
+                        self.avatarUrlSubject.value = nil
+                    }
+                } catch {
+                    return
```
pixlwave, line 183: We should probably log something here.

ElementX/Sources/Screens/Settings/SettingsScreenViewModel.swift
```diff
@@ -43,11 +43,16 @@ class SettingsScreenViewModel: SettingsScreenViewModelType, SettingsScreenViewMo
         ServiceLocator.shared.settings.$timelineStyle
             .weakAssign(to: \.state.bindings.timelineStyle, on: self)
             .store(in: &cancellables)
+
+        userSession.clientProxy.avatarUrlPublisher
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] url in
+                self?.state.userAvatarURL = url
+            }
```
pixlwave, line 51: Can this use `weakAssign`? Same comment for the other place you sink it?
Also wdyt about moving `receive(on:)` to the publisher itself?

pixlwave (approved):

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -162,34 +169,20 @@ class ClientProxy: ClientProxyProtocol {
...
-                } catch {
-                    return
+    func loadUserAvatar() async {
+        await Task.dispatch(on: clientQueue) {
+            do {
+                if let urlString = try self.client.avatarUrl() {
+                    self.avatarURLSubject.value = URL(string: urlString)
+                } else {
+                    self.avatarURLSubject.value = nil
                 }
```
pixlwave, line 179: A nice little improvement here would be to
```suggestion
                let urlString = try self.client.avatarUrl()
                self.avatarURLSubject.value = urlString.flatMap(URL.init)
```

pixlwave (approved):
Looks great! 👏 Sorry I found a typo though.

PR #646, opened 2023-02-27
Collapsible timeline item tweaks

stefanceriu:
* Fix collapsible timeline items grouping just one item, remove duplicate timeline item checking, clean up the logic
* Cleanup collapsible, separator and state room timeline view paddings
* Increase collapsible button tappable area

Screenshot 2023-02-27 at 16 33 53


callumu:
This looks good!

callumu:
I noticed an error with how the expand / collapse behaves when it's the bottom item. Initially it adjusts the content so it's all visible (preferable imo), and then after it collapses again it doesn't do this. Is this intended?

https://user-images.githubusercontent.com/110109933/221596621-d4527c2c-3ac1-4592-8a3a-50f338a704d7.MP4

stefanceriu:
> Is this intended?

None of it is really. Getting the timeline to automatically scroll to the bottom when it's expanded would be a significant chunk of work and we didn't tackle it here. In your case the expanding probably coincided with a timeline update and kicked in the scroll to bottom mecanism but it's by no means intended.

callumu:
Okay cool, thanks for the clarification. Perhaps we can revisit this later

pixlwave (approved):
LGTM

PR #650, opened 2023-02-28
Various fixes

stefanceriu:
This PR fixes 2 bugs:
1) Incorrect stack and split coordinator teardowns leading to possible multiple stops for the same coordinator
2) Crashes on the settings screen which was showing the "complete verification" button before the session verification controller proxy was ready


pixlwave (approved):
LGTM 👍

PR #651, opened 2023-02-28
Update login and server selection screens

pixlwave:
Fixes #632.

Additionally tidies up some uses of the old `NavigationView`, opting for `NavigationStack` instead.

| Login Screen | Server Selection Screen |
| - | - |
| Simulator Screen Shot - iPhone 14 - 2023-03-02 at 12 49 16 | Simulator Screen Shot - iPhone 14 - 2023-03-02 at 12 49 23 |

PR #659, opened 2023-03-01
Report Content v2

Velin92:
New improved Report Content UI

I only need to add a bunch of tests.

https://user-images.githubusercontent.com/34335419/222234844-d93db558-36e4-4012-b1b0-d2bf3196a112.mp4


amshakal (approved):
TY!

ElementX/Sources/Screens/BugReport/BugReportCoordinator.swift
```diff
@@ -57,7 +57,7 @@ final class BugReportCoordinator: CoordinatorProtocol {
             case .cancel:
                 self.completion?(.cancel)
             case let .submitStarted(progressTracker):
-                self.startLoading(label: ElementL10n.bugReportScreenSending, progressPublisher: progressTracker)
+                self.startLoading(label: ElementL10n.sending, progressPublisher: progressTracker)
```
pixlwave, line 60: Thanks it feels much nicer to have generic strings for stuff like this 👏

ElementX/Sources/Screens/ReportContent/ReportContentCoordinator.swift
```diff
@@ -0,0 +1,90 @@
...
+        viewModel.callback = { [weak self] action in
+            guard let self else { return }
+            switch action {
+            case .submitStarted:
+                self.startLoading()
+            case let .submitFailed(error):
+                self.stopLoading()
+                self.showError(description: error.localizedDescription)
+            case .submitFinished:
+                self.stopLoading()
```
pixlwave, line 54: It might be nice to pass the indicator presenter (as weak) to the view model and then not need to worry about all these extra callback actions.

ElementX/Sources/Screens/RoomScreen/RoomScreenCoordinator.swift
```diff
@@ -115,4 +117,27 @@ final class RoomScreenCoordinator: CoordinatorProtocol {
 
         navigationStackCoordinator.push(coordinator)
     }
+
+    private func displayReportContent(for itemId: String) {
+        let navCoordinator = NavigationStackCoordinator()
```
pixlwave, line 122: Lets not abbreviate please :)
```suggestion
        let navigationCoordinator = NavigationStackCoordinator()
```

ElementX/Sources/Screens/ReportContent/View/ReportContentScreen.swift
```diff
@@ -0,0 +1,109 @@
...
+                    .allowsHitTesting(false)
+            }
+
+            RoundedRectangle(cornerRadius: 14, style: .continuous)
+                .stroke(Color.element.quaternaryContent)
+        }
+        .frame(maxWidth: .infinity)
+        .frame(height: 220)
+        .font(.body)
+    }
```
pixlwave, line 81: Can we make some kind of a `formTextEditorStyle()` modifier here that we could reuse between here and the bug report screen?

ElementX/Sources/Screens/ReportContent/View/ReportContentScreen.swift
```diff
@@ -0,0 +1,109 @@
...
+
+    /// The main content of the view to be shown in a scroll view.
+    var mainContent: some View {
+        VStack(alignment: .leading, spacing: 24) {
+            Text(ElementL10n.reportContentInfo)
+                .font(.element.body)
+                .foregroundColor(Color.element.primaryContent)
+            descriptionTextEditor
+        }
```
pixlwave, line 49: It would be nice to use a `Form` here instead of a `VStack` this time around as the bug report screen feels like a bunch of hacks to emulate one. Then we can remove the `ScrollView` in `body` too and it simplifies things a lot.

ElementX/Sources/Screens/ReportContent/View/ReportContentScreen.swift
```diff
@@ -0,0 +1,109 @@
...
+            Button(ElementL10n.actionCancel) {
+                context.send(viewAction: .cancel)
+            }
+        }
+
+        ToolbarItem(placement: .confirmationAction) {
+            Button(ElementL10n.actionSend) {
+                context.send(viewAction: .submit)
+            }
+            .accessibilityIdentifier(A11yIdentifiers.bugReportScreen.send)
```
pixlwave, line 95: Can probably remove this as its using the bug report screen identifier and seems unused.

ElementX/Sources/Screens/ReportContent/View/ReportContentScreen.swift
```diff
@@ -0,0 +1,109 @@
...
+
+// MARK: - Previews
+
+struct ReportContent_Previews: PreviewProvider {
+    static var previews: some View {
+        Group {
+            let viewModel = ReportContentViewModel(itemID: "", roomProxy: MockRoomProxy(displayName: nil))
+            ReportContentScreen(context: viewModel.context)
+        }
+    }
```
pixlwave, line 108: Sorry, I've tidied most of these up in #651, so this is just to keep consistency.
```suggestion
    static let viewModel = ReportContentViewModel(itemID: "", roomProxy: MockRoomProxy(displayName: nil))
    
    static var previews: some View {
        ReportContentScreen(context: viewModel.context)
    }
```

pixlwave (approved):
LGTM, various thoughts inline but nothing major 😎

ElementX/Sources/Screens/ReportContent/View/ReportContentScreen.swift
Velin92: Sadly it actually makes thing a bit more complicated, because forms by default always wrap the content in at least one Section, so the UI ends up looking like this:
Screenshot 2023-03-02 at 16 56 45

I think this is why the bug report screen was not made in this way

ElementX/Sources/Screens/ReportContent/ReportContentCoordinator.swift
Velin92: I would still need to handle a finish callback so in the end I would just remove 1 or 2 actions, not that big of a deal.

ElementX/Sources/Screens/ReportContent/View/ReportContentScreen.swift
Velin92: Maybe a specific view to wrap this entire piece here?

Velin92: I guess I could just make a reusable view for it and call it FormTextEditor

PR #663, opened 2023-03-02
Remove unused media loading from URL.

pixlwave:
There are all unused as we are now using `MediaSource` everywhere (and should continue to do so).


stefanceriu (approved):
🚢 it! 😄

PR #667, opened 2023-03-02
Bump the Rust SDK

stefanceriu:
* bump the rust sdk and fix breaking changes
* cleanup where and how we present loading indicators on the app coordinator

### DIFF
--- a/ElementX/Sources/Services/Client/ClientProxy.swift
+++ b/ElementX/Sources/Services/Client/ClientProxy.swift
@@ -58,10 +58,10 @@ class ClientProxy: ClientProxyProtocol {
     private var slidingSyncObserverToken: TaskHandle?
     private var slidingSync: SlidingSync?
     
-    var visibleRoomsSlidingSyncView: SlidingSyncView?
+    var visibleRoomsSlidingSyncView: SlidingSyncList?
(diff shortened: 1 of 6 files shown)


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -58,10 +58,10 @@ class ClientProxy: ClientProxyProtocol {
     private var slidingSyncObserverToken: TaskHandle?
     private var slidingSync: SlidingSync?
     
-    var visibleRoomsSlidingSyncView: SlidingSyncView?
+    var visibleRoomsSlidingSyncView: SlidingSyncList?
```
pixlwave, line 61: Probably not the comment you want if you're busy right now, but should we be updating all the class, variable and parameter names to match?

pixlwave (approved):
LGTM 👍

ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu: I thought about it but I honestly don't think it's worth it. If we're ever going to move this logic to the Rust side then maybe at that point.

pixlwave: Very good point 👍

PR #683, opened 2023-03-07
Update danger message.

pixlwave:
Seeing as we tend to use Squash merges 99% of the time we decided to update this warning message.

This PR also runs `brew update` on CI before installing packages so that SwiftFormat is on the latest version. Includes some changes for the a couple of new `hoist` rules.


Velin92 (approved):
LGTM

PR #684, opened 2023-03-07
Updated `setPusher` function 

Velin92:
Updated the set pusher function to work again (will only work once my [Rust PR](https://github.com/matrix-org/matrix-rust-sdk/pull/1608) is merged and released)


ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -473,11 +473,11 @@ extension AppCoordinator: NotificationManagerDelegate {
     func notificationTapped(_ service: NotificationManagerProtocol, content: UNNotificationContent) async {
         MXLog.info("[AppCoordinator] tappedNotification")
 
-        guard let roomId = content.userInfo[NotificationConstants.UserInfoKey.roomIdentifier] as? String else {
+        guard !content.threadIdentifier.isEmpty else {
```
pixlwave, line 476: Nice! Could we add a small doc comment to clarify that this is the notification's `thread-id` and nothing todo with matrix threads? I was a bit confused by this change at first 🙈

ElementX/Sources/Mocks/Generated/GeneratedMocks.swift
```diff
@@ -11,18 +10,20 @@ class BugReportServiceMock: BugReportServiceProtocol {
...
         return crashCallsCount > 0
     }
     var crashClosure: (() -> Void)?
+
     func crash() {
         crashCallsCount += 1
         crashClosure?()
     }
-    // MARK: - submitBugReport
+    //MARK: - submitBugReport
```
pixlwave, line 26: Why the loss of the space here?

ElementX/Sources/Services/Client/ClientProxyProtocol.swift
```diff
@@ -31,37 +31,21 @@ enum ClientProxyError: Error {
...
-//    }
-}
-
 enum SlidingSyncConstants {
     static let initialTimelineLimit: UInt = 0
     static let lastMessageTimelineLimit: UInt = 1
     static let timelinePrecachingTimelineLimit: UInt = 20
 }
 
+struct Pusher {
```
pixlwave, line 40: Could we add some docs to the one please? I don't think the name `Pusher` would be immediately obvious to someone who is new to the project, so would be nice to add a description.

pixlwave (commented):
Looks great, nice simplification with the Rust types 👍

ElementX/Sources/Services/Client/ClientProxyProtocol.swift
Velin92: renaming it also to PusherConfiguration

pixlwave: Nice, much cleaner 👏

pixlwave (approved):
🚢 it!

PR #690, opened 2023-03-08
Hide the avatar when the Dynamic Text is large

sempruijs:
Currently, when the user has larger font on (settings > accessibility > display & text size  > larger text) and they try to read the content in the Home Screen, the profile pictures are so big that they cannot read the groep name anymore.

I think the group name and the preview message is more important than someones profile picture. (And  whatsapp agrees with me)

old, a screenshot of the home view where the profile pictures are too big

This pr fixes that by hiding the avatar when the user has a larger text size.

normal, a screenshot with the regular text size
fixed, a screenshot of the home view with larger text, where the profile pictures are hidden

### Pull Request Checklist

- [x] Pull request includes a [sign off](https://github.com/matrix-org/matrix-ios-sdk/blob/develop/CONTRIBUTING.md#sign-off)

- [x] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomCell.swift
+++ b/ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomCell.swift
@@ -20,6 +20,8 @@ struct HomeScreenRoomCell: View {
     let room: HomeScreenRoom
     let context: HomeScreenViewModel.Context
     
+    @Environment(\.dynamicTypeSize) var size
+    
     var body: some View {
         Button {
             if let roomId = room.roomId {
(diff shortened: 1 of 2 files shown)


sempruijs:
Signed-off-by: Sem Pruijs <sem@pruijs.net>

cor:
LGTM

pannie:
LGTM

pixlwave (changes requested):
Hey @sempruijs thanks for this PR. I've run it past our design team and there are a couple of tweaks we would like to make:
- Hide the avatar at AX3.
- Don't allow the avatar to grow any bigger than AX1.

I've included suggestions in your PR for this :)

We've realised we will need to create designs for a different cell layout when using the larger AX sizes at some stage in the future to hopefully improve things further for visually impaired users.

sempruijs:
Ahhh thanks for the suggestions! I can implement them probably somewhere next week :) 

I am very happy that you take accessibility serious. My goal is to make element x the most accessible private messenger and I am glad that you give feedback and help me along the way to make the app and me as a developer better ❤️

ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomCell.swift
```diff
@@ -20,6 +20,8 @@ struct HomeScreenRoomCell: View {
     let room: HomeScreenRoom
     let context: HomeScreenViewModel.Context
     
+    @Environment(\.dynamicTypeSize) var size
```
pixlwave, line 23: One last suggestion looking through this again. Its nice to keep the name of environment values consistent with their keys:
```suggestion
    @Environment(\.dynamicTypeSize) private var dynamicTypeSize
```

PR #695, opened 2023-03-10
Update the layout of HomeScreenRoomCell for Dynamic Type

pixlwave:
Following on from #690, this PR makes a couple of changes to the room cells:
- Make sure the separator always aligns to the text when dynamic type changes the size of the avatar.
- Use vertical padding for the minimum height to allow more cells to be shown at smaller dynamic type sizes.

It also fixes a bug where the colour of the separator wasn't actually being applied and tidies up the button style to easily set a selected state when we have that.

| Extra Small | Large (Default) | AX1 |
| - | - | - |
| Screenshot 2023-03-10 at 1 57 31 pm | Screenshot 2023-03-10 at 1 57 14 pm | Screenshot 2023-03-10 at 1 57 25 pm |


ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomCell.swift
```diff
@@ -80,43 +100,50 @@ struct HomeScreenRoomCell: View {
             ZStack(alignment: .topLeading) {
                 // Hidden text with 2 lines to maintain consistent height, scaling with dynamic text.
                 Text(" \n ").lastMessageFormatting().hidden()
```
Velin92, line 102:
```suggestion
                Text(" \n ")
                     .lastMessageFormatting()
                     .hidden()
```

Velin92 (approved):
LGTM, 1 suuggestion

PR #700, opened 2023-03-14
Allow theming of the search bar.

pixlwave:
This changes adds a `searchableStyle` modifier so that search bars can be themed to match the Figma. This is then applied to the HomeScreen, RoomMemberDetailsScreen and EmojiPicker.

Before and after.
Screenshot 2023-03-14 at 12 21 20 pm


Velin92 (commented):
LGTM

PR #701, opened 2023-03-14
Fix accent colour not being applied to Alerts etc.

pixlwave:
The use of SwiftUI's `.tint` modifier was really manual and couldn't be applied to Alerts. This PR sets the tint on `UIView` which means it is carried all over the UI (including to Xcode previews) and behaves as thought the accent colour was defined within the asset catalog.

| Alert | Modals (unchanged but showing it is automatic) |
| - | - |
| Simulator Screen Shot - iPhone 14 - 2023-03-14 at 12 28 23 | Simulator Screen Shot - iPhone 14 - 2023-03-14 at 12 30 04 |

### DIFF
--- a/ElementX/Sources/Screens/AnalyticsPrompt/View/AnalyticsPrompt.swift
+++ b/ElementX/Sources/Screens/AnalyticsPrompt/View/AnalyticsPrompt.swift
@@ -69,6 +69,7 @@ struct AnalyticsPrompt: View {
                 .font(.element.body)
                 .multilineTextAlignment(.center)
                 .foregroundColor(.element.secondaryContent)
+                .tint(.element.links)
             
             Divider()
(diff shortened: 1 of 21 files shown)


ElementX/Sources/Screens/AnalyticsPrompt/View/AnalyticsPrompt.swift
```diff
@@ -69,6 +69,7 @@ struct AnalyticsPrompt: View {
                 .font(.element.body)
                 .multilineTextAlignment(.center)
                 .foregroundColor(.element.secondaryContent)
+                .tint(.element.links)
```
pixlwave, line 72: Tiny fix that makes sense on a screen that isn't live so very low risk.

Velin92 (approved):
LGTM

PR #702, opened 2023-03-14
Move media file loading logic to the SDK.

### LINKED ISSUE #316: Drop Swift side disk caches
We are currently caching media files on disk on the ElX side but the RustSDK also has a cache. We should expose file paths on the MediaRequest \(or similar\) and drop client side caching

### LINKED ISSUE #444: Improvements to FileCache logic.
Whenever we download media to show via AVPlayer/QuickLook the file is cached using naive guesses of the file extension. As we're writing `Data` to disk, the file doesn't have any useful type metadata, so if the file extension is wrong \(e.g. #443\) then playback/previewing will fail. We should either:
- Add more robust file type detection by reading the mimetype from the event, and then checking the data received.
- Pass this over to Rust itself and instead have it return us a valid URL to open the file from rather than data which we write to disk.

pixlwave:
This PR uses the Rust [`getMediaFile`](https://github.com/matrix-org/matrix-rust-sdk/pull/1649) method to allow us to drop our implementation of file caching and instead let the SDK handle it. We use a `MediaFileProxy` which maintains ownership of a temporary file on disk. When the proxy is released, Rust will delete the file from disk. This stops the app from caching decrypted files on disk.

Closes #316
Closes #444


ElementX/Sources/Services/Media/MediaLoader.swift
```diff
@@ -46,6 +47,14 @@ actor MediaLoader: MediaLoaderProtocol {
         }
     }
     
+    func loadMediaFileForSource(_ source: MediaSourceProxy, contentType: UTType) async throws -> MediaFileProxy {
+        let result = try await Task.dispatch(on: clientQueue) {
+            try self.client.getMediaFile(source: source.underlyingSource, mimeType: contentType.preferredMIMEType ?? "application/octet-stream")
```
pixlwave, line 52: Using `application/octet-stream` as a basic "some kind of unknown data" fallback here.

stefanceriu: Does this fallback maybe belong sdk side?

stefanceriu (approved):
This is such a great addition 👏

ElementX/Sources/Services/Media/MediaLoader.swift
pixlwave: Discussed and leaving it here for now as it seems nicer to not have an optional in the SDK's API. Maybe in the future we'll decide to show an error and avoid loading the file when the mimetype is missing..

ElementX/Sources/Services/Media/MediaFileProxy.swift
```diff
@@ -0,0 +1,68 @@
...
+    init(handle: MediaFileHandle) {
+        self.handle = handle
+    }
+    
+    /// Creates a new instance as a mock, using a raw `URL`
+    ///
+    /// A media file created from a URL won't have the automatic clean-up mechanism
+    /// that is provided by the SDK's `MediaFileHandle`.
+    init(url: URL) {
+        handle = MockMediaFileHandle(url: url)
```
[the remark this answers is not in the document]
pixlwave: Tried using a protocol for the whole proxy but ran into issues equating `any MediaFileHandleProxyProtocol`. Have renamed this to `UnmanagedMediaFileHandle` and moved the init to a `static func unmanaged`

PR #706, opened 2023-03-17
Add Localazy for Strings.

pixlwave:
- Namespaced under `L10n` so we can easily watch our migration from the old strings to the new ones.
- Old strings moved to `Legacy.strings[dict]`.
- Removes old scripts for updating EA strings.


alfogrillo (approved):
LGTM!

PR #708, opened 2023-03-20
Adopt Compound for the settings and bug report screens.

pixlwave:
This PR is the first part of adopting the tokens from https://github.com/vector-im/compound-design-tokens, using https://github.com/vector-im/compound-ios for pre-defined component styles that match the Compound Figma.

Dark mode will be updated at some stage with further semantic tokens to replace the current `formBackground` and `formRowBackground`, but this isn't a blocker for now.

| Settings | Bug Report |
| - | - |
| Simulator Screen Shot - iPhone 14 - 2023-03-20 at 15 27 24 | Simulator Screen Shot - iPhone 14 - 2023-03-20 at 15 27 27 |


stefanceriu (approved):
Very nice 👌

PR #711, opened 2023-03-20
Room Member is now an interface SDK update

Velin92:


ElementX/Sources/Services/Room/RoomMember/RoomMemberProxyProtocol.swift
```diff
@@ -0,0 +1,33 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+import MatrixRustSDK
+
+// sourcery: AutoMockable
+protocol RoomMemberProxyProtocol: Identifiable {
```
alfogrillo, line 21: Extending `Identifiable` has a price: 
now `RoomMemberProxyProtocol` is generic meaning that we need this `any` keyword around.
I was wondering: wouldn't be easier just using `userId` whenever is needed and remove the `Identifiable` extension? What do we risk if we do it?

Velin92: It was already existing for the original RoomMemberProxy (before it was also hidden behind a protocol).
The conformance is used here:
```
return VStack {
            ForEach(members) { member in
                RoomMemberDetailsMemberCell(member: .init(withProxy: member), context: viewModel.context)
            }
        }
```

But yeah honestly I think I can just remove it, is not really that useful, I can just rewrite that piece as this:
```
return VStack {
            ForEach(members, id: \.userId) { member in
                RoomMemberDetailsMemberCell(member: .init(withProxy: member), context: viewModel.context)
            }
        }
```

Velin92: done!

ElementX/Sources/Services/Room/RoomMember/RoomMemberProxy.swift
```diff
@@ -0,0 +1,54 @@
...
+// limitations under the License.
+//
+
+import Foundation
+import MatrixRustSDK
+
+final class RoomMemberProxy: RoomMemberProxyProtocol {
+    private let member: RoomMemberProtocol
+
+    init(with member: RoomMemberProtocol) {
```
pixlwave, line 23: Again, only if you want to handle the changes elsewhere, we could rename this one too. I think `init`s should be clear about what is being passed in.
```suggestion
    init(member: RoomMemberProtocol) {
```

ElementX/Sources/Services/Room/RoomMember/RoomMemberProxyProtocol.swift
```diff
@@ -0,0 +1,29 @@
...
+
+// sourcery: AutoMockable
+protocol RoomMemberProxyProtocol {
+    var userId: String { get }
+    var displayName: String? { get }
+    var avatarURL: URL? { get }
+    var membership: MembershipState { get }
+    var isNameAmbiguous: Bool { get }
+    var powerLevel: Int64 { get }
+    var normalizedPowerLevel: Int64 { get }
```
pixlwave, line 28: Do you think it would make sense to convert these to a regular `Int`?

pixlwave (approved):
LGTM, just a few minor comments.

ElementX/Sources/Services/Room/RoomMember/RoomMemberProxyProtocol.swift
Velin92: Uhm I guess so in the end I think power levels can't exceed 100 so... should be safe to do so

PR #712, opened 2023-03-21
Various tweaks

stefanceriu:
This PR will:
* expose build profiles on Tools/BuildSDK and move targets to an enum
* add a Github CODEOWNERS file so PRs are automatically assigned to @vector-im/element-x-ios-reviewers


pixlwave (approved):
LGTM, couple of thoughts on naming.

PR #716, opened 2023-03-21
Show or create direct message room

flescio:
Resolve: https://github.com/vector-im/element-x-ios/issues/595

New flow for creating a direct room or presenting an already existing one.

dm_rooms


ElementX/Sources/Screens/StartChat/StartChatCoordinator.swift
```diff
@@ -45,6 +46,27 @@ final class StartChatCoordinator: CoordinatorProtocol {
             case .createRoom:
                 // TODO: start create room flow
```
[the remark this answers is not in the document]
stefanceriu: Please refrain from adding todo's unless it's something that you're actually blocked on.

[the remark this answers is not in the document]
stefanceriu: Same here

ElementX/Sources/Screens/StartChat/StartChatCoordinator.swift
```diff
@@ -45,6 +46,27 @@ final class StartChatCoordinator: CoordinatorProtocol {
...
                 break
+            case .userSelected(let user):
```
stefanceriu, line 49: `close` and `createRoom` are imperatives so this one should be too: `selectUser` to keep them consistent. Also worth mentioning that we select an `User` but the associated value is a `Member`. We should probably just stick with one of them

stefanceriu, line 49: On another note, why don't we handle this logic in the viewModel directly instead of forwarding it up to the coordinator?

ElementX/Sources/Screens/StartChat/StartChatModels.swift
```diff
@@ -19,21 +19,33 @@ import Foundation
...
 }
 
 struct StartChatViewState: BindableState {
     var bindings = StartChatScreenViewStateBindings()
     
+    var isSearching: Bool {
+        !bindings.searchQuery.isEmpty
+    }
+    
+    // user resulting from search
```
stefanceriu, line 32: Remove this comment, doesn't help in any way

ElementX/Sources/Screens/StartChat/StartChatModels.swift
```diff
@@ -19,21 +19,33 @@ import Foundation
...
+    var searchedUsers: [RoomMemberProxy] = []
+    
     // TODO: bind with real service, and mock data only in preview
```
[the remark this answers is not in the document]
stefanceriu: Same here

ElementX/Sources/Screens/StartChat/StartChatModels.swift
```diff
@@ -19,21 +19,33 @@ import Foundation
...
+    
+    /// Information describing the currently displayed alert.
+    var alertInfo: AlertInfo<ClientProxyError>?
 }
 
 enum StartChatViewAction {
     case close
     case createRoom
     case inviteFriends
+    case userSelected(RoomMemberProxy)
```
stefanceriu, line 50: Same comment as above `selectUser`

ElementX/Sources/Screens/StartChat/StartChatModels.swift
```diff
@@ -19,21 +19,33 @@ import Foundation
...
 
 struct StartChatViewState: BindableState {
     var bindings = StartChatScreenViewStateBindings()
     
+    var isSearching: Bool {
+        !bindings.searchQuery.isEmpty
+    }
+    
+    // user resulting from search
+    var searchedUsers: [RoomMemberProxy] = []
```
stefanceriu, line 33: Let's maybe rename this to something more appropiate like `searchResults` or `suggestedUsers`

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
```diff
@@ -26,6 +27,8 @@ class StartChatViewModel: StartChatViewModelType, StartChatViewModelProtocol {
     init(userSession: UserSessionProtocol) {
         self.userSession = userSession
         super.init(initialViewState: StartChatViewState(), imageProvider: userSession.mediaProvider)
+        
+        start()
```
stefanceriu, line 31: What's this about, why do need a start method? If you really want to extract that code call is something appropiate e.g. `setupBindings`

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
```diff
@@ -39,6 +42,39 @@ class StartChatViewModel: StartChatViewModelType, StartChatViewModelProtocol {
...
         }
     }
+    
+    func displayError(_ type: ClientProxyError) {
+        switch type {
+        case .failedRetrievingDirectRoom:
+            state.bindings.alertInfo = AlertInfo(id: type,
+                                                 title: ElementL10n.dialogTitleError,
+                                                 message: ElementL10n.retrievingDirectRoomError)
+        case .failedCreatingRoom: // this will likely be in the Room's screen while sending the first message
```
stefanceriu, line 56: Same ask with this comment as well. Don't leave comments for things that aren't 100% clear. `Will likely` doesn't bring any value in the long run

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
```diff
@@ -39,6 +42,39 @@ class StartChatViewModel: StartChatViewModelType, StartChatViewModelProtocol {
         case .inviteFriends:
             // TODO: start invite people flow
```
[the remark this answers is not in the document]
stefanceriu: Same as the others

ElementX/Sources/Screens/StartChat/View/StartChatScreen.swift
```diff
@@ -21,9 +21,13 @@ struct StartChatScreen: View {
     
     var body: some View {
         Form {
-            createRoomSection
-            inviteFriendsSection
-            suggestionsSection
+            if context.viewState.isSearching {
+                searchedUsersSection
```
stefanceriu, line 25: This links to the comment above, let's go with `userSuggestionSection` or whatever

ElementX/Sources/Screens/StartChat/View/StartChatScreen.swift
```diff
@@ -69,6 +74,18 @@ struct StartChatScreen: View {
         .formSectionStyle()
     }
     
+    private var searchedUsersSection: some View {
+        Section {
+            ForEach(context.viewState.searchedUsers, id: \.userId) { user in
+                StartChatSuggestedUserCell(user: user, imageProvider: context.imageProvider)
+                    .onTapGesture {
```
stefanceriu, line 81: Shouldn't this be a Button instead so it gets correct highlight states?

ElementX/Sources/Screens/StartChat/View/StartChatSuggestedUserCell.swift
```diff
@@ -32,14 +32,14 @@ struct StartChatSuggestedUserCell: View {
             VStack(alignment: .leading, spacing: 4) {
                 // covers both nil and empty state
                 let displayName = user.displayName ?? ""
-                Text(displayName.isEmpty ? user.userId : displayName)
+                let firstRowText = displayName.isEmpty ? user.userId : displayName
```
stefanceriu, line 35: This logic feels a bit convoluted, I would just simply drop everything in the Text() i.e. `Text(user.displayName.isEmpty ? user.userId : user.displayName)`

ElementX/Sources/Screens/StartChat/View/StartChatSuggestedUserCell.swift
```diff
@@ -32,14 +32,14 @@ struct StartChatSuggestedUserCell: View {
...
+                Text(firstRowText)
                     .font(.element.title3)
                     .foregroundColor(.element.primaryContent)
-                if !displayName.isEmpty {
-                    Text(user.userId)
-                        .font(.element.subheadline)
-                        .foregroundColor(.element.tertiaryContent)
-                }
+                let secondRowText = displayName.isEmpty ? "" : user.userId
+                Text(secondRowText)
```
stefanceriu, line 40: Same comment as above and you should probably center things vertically when the second line is missing. Looks a bit weird otherwise

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -141,6 +141,31 @@ class ClientProxy: ClientProxyProtocol {
         slidingSyncObserverToken = nil
     }
     
+    func currentDirectRoomWithUser(_ userId: String) async -> Result<String?, ClientProxyError> {
```
stefanceriu, line 144: What does "current" mean in this context? In keeping with the name of `roomForIdentifier` can we maybe call this `directRoomForUserIdentifier`?

ElementX/Sources/Services/Room/RoomMemberProxy.swift
```diff
@@ -23,6 +23,17 @@ struct RoomMemberProxy {
     init(with member: RoomMember) {
         self.member = member
     }
+    
+    // convenience init for start chat flow while a proper User Struct is missing from search result
```
stefanceriu, line 27: A `convenience` constructor without a `convenience` specifier? 😅 I don't think we need this comment, just having the constructor in here is clear enough, no matter where it's used from.

ElementX/Sources/Screens/StartChat/StartChatCoordinator.swift
```diff
@@ -45,6 +46,27 @@ final class StartChatCoordinator: CoordinatorProtocol {
...
+                Task {
+                    let currentDirectRoom = await self.parameters.userSession.clientProxy.currentDirectRoomWithUser(user.userId)
```
stefanceriu, line 51: Don't you need some sort of loading indicator to block interaction while these potentially long taks running? same for creating a room

ElementX/Sources/Services/UserSession/UserSessionFlowCoordinator.swift
```diff
@@ -267,6 +267,9 @@ class UserSessionFlowCoordinator: CoordinatorProtocol {
             switch action {
             case .close:
                 self.navigationSplitCoordinator.setSheetCoordinator(nil)
+            case .openRoom(let identifier):
+                self.navigationSplitCoordinator.setSheetCoordinator(nil)
+                self.presentRoomWithIdentifier(identifier)
```
stefanceriu, line 272: This is wrong, room presentation should go through state machine changes not be invoked directly i.e. `stateMachine.processEvent(.selectRoom(roomId: roomId))` the same as we do for `case .presentRoom`. Otherwise you will leave the state machine in an inconsistent state and it will crash

stefanceriu (changes requested):

ElementX/Sources/Screens/StartChat/StartChatCoordinator.swift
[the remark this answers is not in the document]
alfogrillo: What are people using to mark spots in the code where future developments are required in the near future?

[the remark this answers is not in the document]
stefanceriu: Nothing, chances are 3 years down the line you're going to end up with a ton of TODO's that don't make sense to anybody. If you, for example, have an enum case that's not handled just break on it. It would be very clear to everybody that nothing is happening there and if you do need to come back to it I'm sure it's going to be easy enough to find out without a TODO. If it's not then we're probably doing something wrong

ElementX/Sources/Screens/StartChat/StartChatCoordinator.swift
alfogrillo: I think we won't use real "members" in the near future, since here is when we are creating rooms so there isn't any membership concept involved yet.

ElementX/Sources/Screens/StartChat/StartChatModels.swift
alfogrillo, line 33: To clarify: this is a mock. I think we are going to use [this type](https://github.com/matrix-org/matrix-rust-sdk/pull/1689/files#diff-f4219660dd4687809f8c9f0a5f4f17b5e782ea7e28f5b6e40cccbb67c38e1308R552) (or a proxy of it) instead.
Does it make sense to updated here since the above has been released already?

alfogrillo:
@flescio you can use the same pattern for naming the branches we had on Element.
eg. `flescio/show_or_create_dm` instead of `epic/create_and_join_rooms`.
No need to change it here imho.

ElementX/Sources/Screens/StartChat/StartChatCoordinator.swift
flescio: @stefanceriu  Since I moved this logic to the viewModel, in order to use the indicator, should I expose the userIndicatorController to the viewModel, or should I create 2 actions inside StartChatViewModelAction like showIndicator and hideIndicator, so the coordinator will be in charge of that?

ElementX/Sources/Other/SwiftUI/List Styles/ListSection.swift
```diff
@@ -0,0 +1,32 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import SwiftUI
+
+/// Standard style for list sections
+struct ListSectionStyle: ViewModifier {
```
stefanceriu, line 20: Same comment as above

stefanceriu (approved):
Looks good to me, nicely done!

ElementX/Sources/Screens/StartChat/StartChatModels.swift
```diff
@@ -19,21 +19,31 @@ import Foundation
...
     
-    // TODO: bind with real service, and mock data only in preview
-    var suggestedUsers: [RoomMemberProxyMock] = [.mockAlice, .mockBob, .mockCharlie]
+    var isSearching: Bool {
+        !bindings.searchQuery.isEmpty
+    }
+    
+    var searchResults: [UserProfileProxyProtocol] = []
+    
+    var suggestedUsers: [UserProfileProxyProtocol] = [UserProfileProxyMock.mockAlice, UserProfileProxyMock.mockBob, UserProfileProxyMock.mockCharlie]
```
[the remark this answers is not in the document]
stefanceriu: Then maybe you should create a Section struct that has the title and an `UserProfileProxyProtocol` array?

ElementX/Sources/Other/SwiftUI/List Styles/ListRow.swift
```diff
@@ -0,0 +1,22 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import SwiftUI
+
+enum ListRow {
+    /// Element specific insets that are used for all our Form rows.
+    static let insets = EdgeInsets(top: 10, leading: 16, bottom: 10, trailing: 16)
```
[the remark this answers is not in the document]
pixlwave: Eventually yes, but we'll be doing that screen by screen and moved stuff as we need it so for now, this is fine.

The comment should say all our `List` rows though 😜

ElementX/Sources/Services/Users/UserProfile.swift
```diff
@@ -0,0 +1,37 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+import MatrixRustSDK
+
+// sourcery: AutoMockable
+protocol UserProfileProxyProtocol {
+    var userId: String { get }
```
pixlwave, line 22:
```suggestion
    var userID: String { get }
```

ElementX/Sources/Services/Client/ClientProxyProtocol.swift
```diff
@@ -70,6 +72,10 @@ protocol ClientProxyProtocol: AnyObject, MediaLoaderProtocol {
     
     func stopSync()
     
+    func directRoomForUserIdentifier(_ userId: String) async -> Result<String?, ClientProxyError>
+    
+    func createDirectRoom(with userId: String) async -> Result<String, ClientProxyError>
```
pixlwave, line 77:
```suggestion
    func directRoomForUserID(_ userID: String) async -> Result<String?, ClientProxyError>
    
    func createDirectRoom(with userID: String) async -> Result<String, ClientProxyError>
```

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
```diff
@@ -37,8 +42,80 @@ class StartChatViewModel: StartChatViewModelType, StartChatViewModelProtocol {
...
+            state.bindings.alertInfo = AlertInfo(id: type)
         }
     }
+    
+    // MARK: - Private
+    
+    private func setupBindings() {
+        context.$viewState
+            .map(\.bindings.searchQuery)
+            .debounce(for: .milliseconds(300), scheduler: DispatchQueue.main)
```
pixlwave, line 84: Should there be a `removeDuplicates` in here too? That would make sure not to call the closure when
a) The user types and then removes a character in the search field within the debounce time, leaving the query unchanged.
b) Any other part of the view state is updated that isn't the search query.

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
```diff
@@ -37,8 +42,80 @@ class StartChatViewModel: StartChatViewModelType, StartChatViewModelProtocol {
...
+            .sink(receiveValue: { [weak self] searchQuery in
+                if MatrixEntityRegex.isMatrixUserIdentifier(searchQuery) {
+                    self?.state.searchResults = [UserProfileProxy(userId: searchQuery, displayName: nil, avatarURL: nil)]
+                } else {
+                    self?.state.searchResults = []
+                }
+            })
```
pixlwave, line 91: Can we prefer trailing closures please?
```suggestion
            .sink { [weak self] searchQuery in
                if MatrixEntityRegex.isMatrixUserIdentifier(searchQuery) {
                    self?.state.searchResults = [UserProfileProxy(userId: searchQuery, displayName: nil, avatarURL: nil)]
                } else {
                    self?.state.searchResults = []
                }
            }
```

pixlwave (commented):
Had a quick skim through whilst I was here and added a couple of tiny comments :)

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
```diff
@@ -81,20 +82,30 @@ class StartChatViewModel: StartChatViewModelType, StartChatViewModelProtocol {
     private func setupBindings() {
         context.$viewState
             .map(\.bindings.searchQuery)
+            .removeDuplicates()
```
pixlwave, line 85: Sorry my bad, my comment was very specific. You want this *after* the debounce I believe.

pixlwave (approved):
Nice!

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
flescio: You're right, I also got it wrong

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -141,6 +141,31 @@ class ClientProxy: ClientProxyProtocol {
...
+            do {
+                if let room = try self.client.getDmRoom(userId: userID) {
+                    return .success(room.id())
+                }
+            } catch {
+                return .failure(.failedRetrievingDirectRoom)
+            }
+            return .success(nil)
+        }
+    }
```
alfogrillo, line 155: We can make it a bit more readable
```suggestion
    func directRoomForUserID(_ userID: String) async -> Result<String?, ClientProxyError> {
        await Task.dispatch(on: clientQueue) {
            do {
                let room = try self.client.getDmRoom(userId: userID)?.id()
                return room.map { .success($0) } ?? .success(nil)
            } catch {
                return .failure(.failedRetrievingDirectRoom)
            }
        }
    }
```

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -141,6 +141,31 @@ class ClientProxy: ClientProxyProtocol {
...
+    
+    func createDirectRoom(with userID: String) async -> Result<String, ClientProxyError> {
+        await Task.dispatch(on: clientQueue) {
+            do {
+                let parameters = CreateRoomParameters(name: "", topic: nil, isEncrypted: true, isDirect: true, visibility: .private, preset: .privateChat, invite: [userID], avatar: nil)
```
alfogrillo, line 160:
```suggestion
                let parameters = CreateRoomParameters(name: "", topic: nil, isEncrypted: true, isDirect: true, visibility: .private, preset: .trustedPrivateChat, invite: [userID], avatar: nil)
```

I think trusted private chat is the default on Element legacy. Please have a look there. 
I think it makes more sense.

ElementX/Sources/Services/Users/UserProfile.swift
```diff
@@ -0,0 +1,41 @@
...
+import MatrixRustSDK
+
+// sourcery: AutoMockable
+protocol UserProfileProxyProtocol {
+    var userID: String { get }
+    var displayName: String? { get }
+    var avatarURL: URL? { get }
+}
+
+struct UserProfileProxy: UserProfileProxyProtocol {
```
alfogrillo, line 27: I don't think we need `UserProfileProxyProtocol`.  `struct UserProfileProxy` should be enough.

alfogrillo (approved):
LGTM! 🎉

PR #718, opened 2023-03-22
Use strings from Localazy

pixlwave:
The PR does the following
- Includes latest strings from Localazy.
- Updates all screens to use these strings.
- Removes old Android strings and any untranslated strings included in Localazy.
- Moves untranslated strings into `UntranslatedL10n`.
- Disables the German UI tests as for now, none of the strings are translated.
    - We should likely pick the double length pseudo-language instead so that translation updates don't break the tests.

UI tests are a mess outside of this PR anyway so I'll address them collectively in a followup PR.


Velin92 (approved):
LGTM

stefanceriu (approved):
Reaaally nice 👏

PR #719, opened 2023-03-22
Fixes #676 - Fixes regression on updating room list states when using…

### LINKED ISSUE #676: App browsing whilst offline

I want to be able to cold start ElementX when i'm on airplane mode

To be able to browse the current known rooms and view known history in those rooms 

N/A

**Additional context**

Nothing is currently shown, only the room list loading placeholder

stefanceriu:
… data from the cold cache

### DIFF
--- a/ElementX/Sources/Services/Client/ClientProxy.swift
+++ b/ElementX/Sources/Services/Client/ClientProxy.swift
@@ -51,9 +51,10 @@ class ClientProxy: ClientProxyProtocol {
     private let clientQueue: DispatchQueue
     
     private var slidingSyncObserverToken: TaskHandle?
-    private var slidingSync: SlidingSync?
+    private var slidingSync: SlidingSyncProtocol?
     
     var visibleRoomsSlidingSyncView: SlidingSyncList?
+    var visibleRoomsViewProxy: SlidingSyncViewProxy?
     var visibleRoomsSummaryProvider: RoomSummaryProviderProtocol?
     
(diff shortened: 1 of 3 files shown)


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -51,9 +51,10 @@ class ClientProxy: ClientProxyProtocol {
     private let clientQueue: DispatchQueue
     
     private var slidingSyncObserverToken: TaskHandle?
-    private var slidingSync: SlidingSync?
+    private var slidingSync: SlidingSyncProtocol?
```
pixlwave, line 54: I wonder if this should be implicitly unwrapped? It will always exist after init, and if building fails, the app isn't going to be at all usable. I'm coming at this from seeing `if let slidingSync` which made me wonder what state sliding sync shouldn't be available. Might tidy up a few other places too.

pixlwave (approved):
One general question otherwise lgtm 👍

ElementX/Sources/Services/Client/ClientProxy.swift
pixlwave: Ah ha makes sense! Should probably have a comment in there to that effect at some point :)

stefanceriu: Yep, but I'll wait until we move parts of this to the Rust side to see how I can refactor things

PR #721, opened 2023-03-22
Dpendencies update + improved dependencies version management

Velin92:
- Updated the dependencies
- Forced minor versions in the project.yml for dependencies
- Added a tool that allows to check for outdated dependencies in the project

NOTE:
I had to set an exactVersion as 1.6.26 for DTCoreText, because a tag for a version 1.6.27 exists but is not in the official release list (there are no release notes for it in fact), and creates some flaky behaviours with block quotes.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -13,7 +13,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMinor(from: "1.2.0")),
-        .package(url: "https://github.com/jpsim/Yams", .upToNextMinor(from: "5.0.0")),
+        .package(url: "https://github.com/jpsim/Yams", .upToNextMinor(from: "5.0.0"))
     ],
     targets: [
         .executableTarget(name: "Tools",


stefanceriu (commented):
I'm getting an error when running `swift run tools setup-project` on this branch. Apparently 
```
Computed https://github.com/apple/swift-argument-parser at 1.2.2 (0.00s)
error: Couldn’t check out revision ‘f47ba4838c30dbd59998a4e4c87ab620ff959e8a’:
    fatal: reference is not a tree: f47ba4838c30dbd59998a4e4c87ab620ff959e8a
```

Not sure how to fix it, any idea what this is about?

Tools/Sources/Tools.swift
```diff
@@ -5,5 +5,6 @@ import Foundation
 struct Tools: ParsableCommand {
     static var configuration = CommandConfiguration(abstract: "A collection of command line tools for ElementX",
                                                     subcommands: [BuildSDK.self,
-                                                                  SetupProject.self])
+                                                                  SetupProject.self,
+                                                                  OutdatedPackages.self])
```
stefanceriu, line 9: How is this OutdatedPackages supposed to be used without any source files? 🤔

Package.resolved
```diff
@@ -5,17 +5,17 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/apple/swift-argument-parser",
       "state" : {
-        "revision" : "fddd1c00396eed152c45a46bea9f47b98e59301d",
-        "version" : "1.2.0"
+        "revision" : "fee6933f37fde9a5e12a1e4aeaa93fe60116ff2a",
+        "version" : "1.2.2"
```
stefanceriu, line 9: Should we specify the Package dependencies using `.upToNextMinor(from:)` as well? e.g.

```
dependencies: [
        .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMinor(from: "1.2.0")),
        .package(url: "https://github.com/jpsim/Yams", .upToNextMinor(from: "5.0.1")),
    ],
```

stefanceriu (changes requested):

stefanceriu (commented):
Should we also take this opportunity to fix the typo in `echo "warning: Sourcert not installed`?

I also want to start a discussion around whether we should actually run this step on every build or make it part of the tools.

Tools/Sources/Tools.swift
Velin92: whoops, here it is, forgot to push it

Package.resolved
Velin92: yeah! that would be better

Velin92:
> I'm getting an error when running `swift run tools setup-project` on this branch. Apparently
> 
> ```
> Computed https://github.com/apple/swift-argument-parser at 1.2.2 (0.00s)
> error: Couldn’t check out revision ‘f47ba4838c30dbd59998a4e4c87ab620ff959e8a’:
>     fatal: reference is not a tree: f47ba4838c30dbd59998a4e4c87ab620ff959e8a
> ```
> 
> Not sure how to fix it, any idea what this is about?

i solved this by using in this order
swift package resolve
swift package update
swift package resolve

I had the same issue, is likely there an SPM bug that happens in repo Package.swift managed manually
https://github.com/apple/swift-argument-parser/issues/517

Velin92:
> Should we also take this opportunity to fix the typo in `echo "warning: Sourcert not installed`?
> 
> I also want to start a discussion around whether we should actually run this step on every build or make it part of the tools.

We actually already had a discussion with @pixlwave and @alfogrillo  and Sourcery takes essentially only 0.5 seconds to run as part of the run phase script, so it's not big deal to have a separate tool for now. Maybe if we are going to have TONS of protocols mocked could be worth it, but for now is already blazingly fast

stefanceriu:
> but for now is already blazingly fast

Alrighty, fair enough 👍

stefanceriu:
> swift package resolve

Doesn't work for me unfortunately. Does it work on your side @pixlwave ?

Velin92:
> > swift package resolve
> 
> Doesn't work for me unfortunately. Does it work on your side @pixlwave ?

This is a bit drastic but you could try to delete the ~/Library/Caches/org.swift.swiftpm/repositories and search for the folder of the argument parser and delete it, and try running the resolve command again

Package.swift
```diff
@@ -12,14 +12,14 @@ let package = Package(
         .executable(name: "tools", targets: ["Tools"])
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-argument-parser", from: "1.2.0"),
-        .package(url: "https://github.com/jpsim/Yams", from: "5.0.1")
+        .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMinor(from: "1.2.0")),
+        .package(url: "https://github.com/jpsim/Yams", .upToNextMinor(from: "5.0.0")),
```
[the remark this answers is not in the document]
stefanceriu: Do you need to actually address this? 🤔

stefanceriu (approved):
Very nice, 🚢 it! 👏

Package.swift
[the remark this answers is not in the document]
Velin92: uhm even in the swift pm documentation package.swift files should always use trailing commas... so not really? But ofc Danger is mixing them up with project code
https://www.swift.org/package-manager/

Tools/Sources/SetupProject.swift
```diff
@@ -15,7 +15,10 @@ struct SetupProject: ParsableCommand {
     }
 
     func brewBundleInstall() throws {
-        try Utilities.zsh("brew install xcodegen swiftgen swiftlint swiftformat git-lfs sourcery")
+        // required for swift-outdated
+        try Utilities.zsh("brew tap kiliankoe/formulae")
+        
+        try Utilities.zsh("brew install xcodegen swiftgen swiftlint swiftformat git-lfs sourcery swift-outdated")
```
pixlwave, line 21: You could simplify this like so:
```suggestion
        try Utilities.zsh("brew install xcodegen swiftgen swiftlint swiftformat git-lfs sourcery kiliankoe/formulae/swift-outdated")
```

Package.swift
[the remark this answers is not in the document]
stefanceriu: I would say the documentation is wrong in this case 😅

Tools/Sources/OutdatedPackages.swift
```diff
@@ -0,0 +1,23 @@
+import ArgumentParser
+import Foundation
+
+struct OutdatedPackages: ParsableCommand {
+    static var configuration = CommandConfiguration(abstract: "A tool to check outdated package dependencies, requires to have setup the project first")
```
pixlwave, line 5: Maybe a little bit clearer:
```suggestion
    static var configuration = CommandConfiguration(abstract: "A tool to check outdated package dependencies. Please make sure you have already run setup-project before using this tool.")
```

Package.swift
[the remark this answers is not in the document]
pixlwave: Yeah ditto on that, I wouldn't leave the trailing comma tbh. But we could equally ignore both of our Package.swift files from SwiftLint like we currently do from SwiftFormat right?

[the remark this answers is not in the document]
Velin92: okay removing the trailing comma then

PR #730, opened 2023-03-23
Update UI tests

pixlwave:
Follow up to #718 which makes the following changes:
- Update all failing snapshot tests in the codebase (some of these have been failing for quite a while it seems).
- Replace German snapshots with double-length pseudolanguage so that downloading new translations won't suddenly break our tests (also they seem much more extreme in a lot of cases too).
- Fix the logic for grouping messages with reactions.

Quite possibly easier to take this one commit-by-commit.

Run of the UI tests is going on [here](https://github.com/vector-im/element-x-ios/actions/runs/4542578034).


manuroe:
>  double-length pseudolanguage 

TIL `NSDoubleLocalizedStrings`. Super idea 👍

ElementX/Sources/Services/Media/MockMediaProvider.swift
```diff
@@ -23,14 +23,10 @@ struct MockMediaProvider: MediaProviderProtocol {
             return nil
         }
         
-        #warning("Fix me. this is stupid!")
```
stefanceriu, line 26: Yey!

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -196,7 +196,7 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
         }
         
         //  can be improved by adding a date threshold
-        return otherEventTimelineItem.properties.reactions.isEmpty && eventTimelineItem.sender == otherEventTimelineItem.sender
+        return eventTimelineItem.properties.reactions.isEmpty && eventTimelineItem.sender == otherEventTimelineItem.sender
```
stefanceriu, line 199: Even more yey, good catch!

UITests/Sources/Application.swift
```diff
@@ -76,6 +76,11 @@ extension XCUIApplication {
         
         return name
     }
+    
+    private var localeCode: String {
+        guard !UserDefaults.standard.bool(forKey: "NSDoubleLocalizedStrings") else { return "pseudo" }
```
stefanceriu, line 81: This feels a bit convoluted, I would go with

```
if UserDefaults.standard.bool(forKey: "NSDoubleLocalizedStrings") == true {
   return "pseudo"
}
```

UnitTests/Sources/RoomScreenViewModelTests.swift
```diff
@@ -14,6 +14,154 @@
 // limitations under the License.
 //
 
+@testable import ElementX
 import XCTest
 
-class RoomScreenViewModelTests: XCTestCase { }
+@MainActor
+class RoomScreenViewModelTests: XCTestCase {
```
stefanceriu, line 21: Awesome 👏

stefanceriu (approved):
Fabulous! 👏

PR #737, opened 2023-03-27
Ignore User UI

Velin92:
Requires an SDK update to work properly


Velin92:
@stefanceriu so I either can release a new version of the SDK and add it inside this PR, everything will work only for the online cases, or we can wait for Ivan to finish also their piece of work and merge this when everything is ready

amshakal:
Looks good! TY. :D 
As we discussed the two missing things are both 1:1 DMs related:
1. Adding block/unblock functionality to 1:1 DMs
2. Removing the composer when a user is blocked in 1:1 DMs

stefanceriu:
> everything will work only for the online cases

Let's roll with it as it is and add Ivan's resetting later

ElementX/Sources/Other/SwiftUI/Form Styles/FormButtonStyles.swift
```diff
@@ -19,13 +19,16 @@ import SwiftUI
 /// A view to be added on the trailing edge of a form row.
 enum FormRowAccessory: View {
     case navigationLink
+    case spinner
```
stefanceriu, line 22: Should we maybe call this `progressView`?

stefanceriu: Otherwise, presumably this should also be added to Compound? 🤔

ElementX/Sources/Screens/RoomMemberDetails/RoomMemberDetailsModels.swift
```diff
@@ -25,6 +25,7 @@ struct RoomMemberDetailsViewState: BindableState {
     let isAccountOwner: Bool
     let permalink: URL?
     var isIgnored: Bool
+    var isIgnoreLoading = false
```
stefanceriu, line 28: Not sure I like this name, does `isProcessingIgnoreRequest` work better?

stefanceriu (approved):
Great stuff 👏

Velin92:
> > everything will work only for the online cases
> 
> Let's roll with it as it is and add Ivan's resetting later

ok then let me do a quick release of the sdk and add it here

PR #740, opened 2023-03-28
Add test plans for all test targets

alfogrillo:
This PR adds test plan for the test targets:
- **UnitTests**: with `"defaultTestExecutionTimeAllowance" : 60`
- **IntegrationTests**: with the default configuration

### DIFF
--- a/ElementX/SupportingFiles/target.yml
+++ b/ElementX/SupportingFiles/target.yml
@@ -33,8 +33,9 @@ schemes:
       gatherCoverageData: true
       coverageTargets:
         - ElementX
-      targets:
-        - UnitTests
+      testPlans:
+        - path: ../../UnitTests/SupportingFiles/UnitTests.xctestplan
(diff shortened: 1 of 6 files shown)


pixlwave (approved):
LGTM, great idea 👍

Maybe worth noting that running the Unit tests when the ElementX scheme is selected doesn't appear to use the test plan, but I don't think that actually matters from the perspective of making sure the right thing happens on CI.

alfogrillo:
> LGTM, great idea 👍
> 
> Maybe worth noting that running the Unit tests when the ElementX scheme is selected doesn't appear to use the test plan, but I don't think that actually matters from the perspective of making sure the right thing happens on CI.

Yeah It doesn't matter, but I think it's still better to be consistent. :)
Applied [here](https://github.com/vector-im/element-x-ios/commit/cb2452decc44465d9db7e0c7e201e76485c66be4).

PR #742, opened 2023-03-29
Add Block user to Report Content screen.

pixlwave:
This PR adds the Block user toggle to the Report Content screen. It additionally adopts compound-ios for the screen as it is the same as the Bug Report screen so can use all the existing tokens/styles.

Draft for now until we have an SDK release containing https://github.com/matrix-org/matrix-rust-sdk/pull/1713

| Initial State | With Report |
| - | - |
| Simulator Screen Shot - iPhone 14 - 2023-03-29 at 13 08 19 | Simulator Screen Shot - iPhone 14 - 2023-03-29 at 14 14 56 |


pixlwave:
Still needs an update to the SDK, but marking as ready for review given the `room.ignore` method is now merged.

UITests/Sources/__Snapshots__/Application/pseudo-iPhone-14.reportContent-1.png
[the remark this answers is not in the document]
pixlwave: It is for now (its worse on dark mode where its a white toggle on an almost white background) - the Toggle style hasn't been confirmed yet on compound, but we can update it as soon as it has.

ElementX/Sources/Services/Room/RoomProxy.swift
```diff
@@ -255,7 +255,7 @@ class RoomProxy: RoomProxyProtocol {
         }
     }
 
-    func reportContent(_ eventID: String, reason: String?) async -> Result<Void, RoomProxyError> {
+    func reportContent(_ eventID: String, reason: String?, ignoring senderID: String? = nil) async -> Result<Void, RoomProxyError> {
```
Velin92, line 258: you know what... actually instead of having `reportContent(eventID, reason, senderID)` at this point maybe I would prefer to mirror the SDK and have two functions `reportContent` and `ignoreUser`.
For two reasons:
- this ignoreUser function is quite useful, and may be useful also for my current DM Work, and for the future when we implement the unblock function on the blocked composer for blocked users
- In this way in case one errors out and the other one does not, we can display an error message saying x was sucessfull but y was not.

Ofc for now we only have the ignore exposed on the room, and not the unignore but it could a start. WDYT?

Velin92: Also is worth testing what could happen if you ignore an already ignored user... prolly nothing but who knows

pixlwave: Sure thing, makes sense to me, I'll update 👍

Velin92 (changes requested):
I would maybe prefer to have ignore and reportContent as separate functions at this point

UnitTests/Sources/ReportContentViewModelTests.swift
```diff
@@ -20,10 +20,54 @@ import XCTest
...
-
-        XCTAssertEqual(context.reasonText, "")
+    let itemID = "test-id"
+    let senderID = "@meany:server.com"
+    let reportReason = "I don't like it."
+    
+    func testReportContent() async {
+        // Given the report content view for some content.
+        let roomProxy = RoomProxyMock(with: .init(displayName: "test"))
+        roomProxy.reportContentReasonIgnoringReturnValue = .success(())
```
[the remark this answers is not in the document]
pixlwave: We discussed this and decided it wasn't useful in this instance.

UnitTests/Sources/MediaProvider/MediaLoaderTests.swift
```diff
@@ -97,6 +97,8 @@ private class MockMediaLoadingClient: ClientProtocol {
     
     func getMediaFile(source: MatrixRustSDK.MediaSource, mimeType: String) throws -> MatrixRustSDK.MediaFileHandle { fatalError() }
     
+    func getProfile(userId: String) throws -> MatrixRustSDK.UserProfile { fatalError() }
```
pixlwave, line 100: Seems this was missed with the latest SDK update.

Velin92 (approved):

Velin92:
maybe we want to change the loading screen to be less of a blocker for the user UI? not required to do in this PR

Issue #747, opened 2023-03-30
[MediaUpload] File preview screen


stefanceriu:
After the user chooses a file to upload we need to show a preview screen with a send option

Designs TBD

Issue #746, opened 2023-03-30
[MediaUpload] Support various media pickers


stefanceriu:
We need to be able to allow the user to pick media from their photo library, files app or directly from the camera.
This needs to be done through 3 different system pickers and grouped together in a nice, reusable, interface

Issue #745, opened 2023-03-30
[MediaUpload] Room composer attachments options


stefanceriu:
The media upload flow starts with a new menu available close to the room composer from which the user can chose the source

Designs TBD

PR #749, opened 2023-03-30
Implements #745, #746, #747 - Initial media uploading flows

stefanceriu:
- room composer contextual menu for selecting source
- coordinator for presenting the different sources
- system picker handling and URL passback
- file preview after selection
- feature flag

### DIFF
--- a/ElementX/Resources/Localizations/en.lproj/Untranslated.strings
+++ b/ElementX/Resources/Localizations/en.lproj/Untranslated.strings
@@ -24,3 +24,11 @@
 "soft_logout_clear_data_submit" = "Clear all data";
 "soft_logout_clear_data_dialog_title" = "Clear data";
 "soft_logout_clear_data_dialog_content" = "Clear all data currently stored on this device?\nSign in again to access your account data and messages.";
+
+// MARK: - Media upload
+
+"media_upload_camera_picker" = "Camera";
+
+"media_upload_photo_and_video_picker" = "Photo & Video Library";
+
+"media_upload_document_picker" = "Document";
--- a/ElementX/Sources/Application/AppSettings.swift
+++ b/ElementX/Sources/Application/AppSettings.swift
@@ -27,7 +27,7 @@ final class AppSettings: ObservableObject {
         case enableInAppNotifications
         case pusherProfileTag
         case shouldCollapseRoomStateEvents
-        case showStartChatFlow
+        case startChatFlowEnabled = "showStartChatFlow"
         case mediaUploadingFlowEnabled
     }
     
@@ -160,8 +160,8 @@ final class AppSettings: ObservableObject {
     
     // MARK: Start Chat
     
-    @UserSetting(key: UserDefaultsKeys.showStartChatFlow.rawValue, defaultValue: false, persistIn: store)
-    var startChatFlowFeatureFlag
+    @UserSetting(key: UserDefaultsKeys.startChatFlowEnabled.rawValue, defaultValue: false, persistIn: store)
+    var startChatFlowEnabled
     
     // MARK: Media Uploading
     
--- a/ElementX/Sources/Application/AppSettings.swift
+++ b/ElementX/Sources/Application/AppSettings.swift
@@ -28,6 +28,7 @@ final class AppSettings: ObservableObject {
         case pusherProfileTag
         case shouldCollapseRoomStateEvents
         case showStartChatFlow
+        case mediaUploadingFlowEnabled
     }
     
     private static var suiteName: String = InfoPlistReader.main.appGroupIdentifier
@@ -161,4 +162,9 @@ final class AppSettings: ObservableObject {
     
     @UserSetting(key: UserDefaultsKeys.showStartChatFlow.rawValue, defaultValue: false, persistIn: store)
     var startChatFlowFeatureFlag
+    
+    // MARK: Media Uploading
+    
+    @UserSetting(key: UserDefaultsKeys.mediaUploadingFlowEnabled.rawValue, defaultValue: false, persistIn: nil)
+    var mediaUploadingFlowEnabled
 }
--- a/ElementX/Sources/Other/Extensions/FileManager.swift
+++ b/ElementX/Sources/Other/Extensions/FileManager.swift
@@ -31,4 +31,21 @@ extension FileManager {
         }
         try createDirectory(at: url, withIntermediateDirectories: withIntermediateDirectories)
     }
+    
+    func copyFileToTemporaryLocation(url: URL) throws -> URL {
+        let newURL = URL.temporaryDirectory.appendingPathComponent(url.lastPathComponent)
+        
+        try? removeItem(at: newURL)
+        try copyItem(at: url, to: newURL)
+        
+        return newURL
+    }
+    
+    func writeDataToTemporaryLocation(data: Data, fileName: String) throws -> URL {
+        let newURL = URL.temporaryDirectory.appendingPathComponent(fileName)
+        
+        try data.write(to: newURL)
+        
+        return newURL
+    }
 }
--- a/ElementX/Sources/Other/Extensions/FileManager.swift
+++ b/ElementX/Sources/Other/Extensions/FileManager.swift
@@ -32,7 +32,7 @@ extension FileManager {
         try createDirectory(at: url, withIntermediateDirectories: withIntermediateDirectories)
     }
     
-    func copyFileToTemporaryLocation(url: URL) throws -> URL {
+    func copyFileToTemporaryDirectory(url: URL) throws -> URL {
         let newURL = URL.temporaryDirectory.appendingPathComponent(url.lastPathComponent)
         
         try? removeItem(at: newURL)
@@ -41,7 +41,7 @@ extension FileManager {
         return newURL
     }
     
-    func writeDataToTemporaryLocation(data: Data, fileName: String) throws -> URL {
+    func writeDataToTemporaryDirectory(data: Data, fileName: String) throws -> URL {
         let newURL = URL.temporaryDirectory.appendingPathComponent(fileName)
         
         try data.write(to: newURL)
--- a/ElementX/Sources/Screens/MediaPicker/MediaPickerCoordinator.swift
+++ b/ElementX/Sources/Screens/MediaPicker/MediaPickerCoordinator.swift
@@ -1,0 +1,89 @@
+//
+// Copyright 2023 New Vector Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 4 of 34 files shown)


ElementX/Sources/Screens/MediaPicker/MediaPickerCoordinator.swift
```diff
@@ -0,0 +1,89 @@
...
+        AnyView(mediaPicker)
+    }
+    
+    @ViewBuilder
+    private var mediaPicker: some View {
+        switch source {
+        case .camera:
+            cameraPicker
+        case .photoLibrary:
+            PhotoLibraryPicker { [weak self] action in
```
pixlwave, line 50: Would it work to use the same Action type for all 3 pickers and then we could have a `let pickerCallback = …` in the coordinator and pass the same one in all 3 cases of the switch? Feels like it might be a bit neater. Not sure how the `[weak self]` would play into that though 🤔

ElementX/Sources/Screens/MediaPicker/PhotoLibraryPicker.swift
```diff
@@ -0,0 +1,84 @@
...
+        
+        // MARK: PHPickerViewControllerDelegate
+        
+        func picker(_ picker: PHPickerViewController, didFinishPicking results: [PHPickerResult]) {
+            guard let provider = results.first?.itemProvider else {
+                parent.callback(.cancel)
+                return
+            }
+            
+            provider.loadFileRepresentation(forTypeIdentifier: "public.item") { @MainActor [weak self] url, error in
```
pixlwave, line 64: Given that this is nothing more than a false impression
```suggestion
            provider.loadFileRepresentation(forTypeIdentifier: "public.item") { [weak self] url, error in
```

ElementX/Sources/Screens/MediaPickerPreviewScreen/MediaPickerPreviewScreenCoordinator.swift
```diff
@@ -0,0 +1,56 @@
...
+    let url: URL
+    let title: String?
+}
+
+enum MediaPickerPreviewScreenCoordinatorAction {
+    case send
+    case cancel
+}
+
+final class MediaPickerPreviewScreenCoordinator: CoordinatorProtocol {
```
pixlwave, line 29: General question, shouldn't this be generated as `MediaPickerPreviewCoordinator` along with the same for actions, parameters, models, viewmodels etc etc?

ElementX/Sources/Screens/MediaPickerPreviewScreen/View/MediaPickerPreviewScreenScreen.swift
```diff
@@ -0,0 +1,110 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import QuickLook
+import SwiftUI
+
+struct MediaPickerPreviewScreenScreen: View {
```
pixlwave, line 20: Ah yeah, `ScreenScreen` I think the script should probably drop a trailing `Screen` if its supplied to the command.

ElementX/Sources/Screens/RoomScreen/RoomScreenCoordinator.swift
```diff
@@ -74,8 +80,32 @@ final class RoomScreenCoordinator: CoordinatorProtocol {
...
+    
     // MARK: - Private
+    
+    private func displayMediaPickerWithSource(_ source: MediaPickerSource) {
+        let mediaPickerCoordinator = MediaPickerCoordinator(source: source) { [weak self] action in
+            switch action {
+            case .cancel:
+                self?.navigationStackCoordinator.setSheetCoordinator(nil)
+            case .error:
+                break
```
pixlwave, line 92: Just noting that I haven't seen any logging of the errors yet. Not saying this it the right place, simply noting it here.

ElementX/Sources/Services/Room/RoomProxy.swift
```diff
@@ -218,6 +218,10 @@ class RoomProxy: RoomProxyProtocol {
             }
         }
     }
+    
+    func sendImage(body: String, url: URL) async -> Result<Void, RoomProxyError> {
```
pixlwave, line 222: Should probably handle this Rust side using the last filename from the URL for the body as in the spec.
```suggestion
    func sendImage(url: URL) async -> Result<Void, RoomProxyError> {
```

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
```diff
@@ -23,7 +23,16 @@ struct RoomScreen: View {
...
-            .safeAreaInset(edge: .bottom, spacing: 0) { messageComposer }
+            .safeAreaInset(edge: .bottom, spacing: 0) {
+                HStack(spacing: 4.0) {
+                    if context.viewState.mediaUploadingFlowEnabled {
+                        sendAttachmentButton
+                    }
+                    
+                    messageComposer
+                }
+                .padding()
```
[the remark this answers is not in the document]
pixlwave: Also, this PR might need to update some snapshots before merging seeing as the UI tests are happily working again.

pixlwave (commented):
Nice, this will be great to have! 👏

ElementX/Sources/Application/AppSettings.swift
```diff
@@ -28,6 +28,7 @@ final class AppSettings: ObservableObject {
         case pusherProfileTag
         case shouldCollapseRoomStateEvents
         case showStartChatFlow
+        case mediaUploadingFlowEnabled
```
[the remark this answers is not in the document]
stefanceriu: No, I intentionally didn't as `showing` is specifically linked to the UI and I don't think that's what we want on this level. If anything I would change the `showStartChatFlow` to `startChatFlowEnabled`

ElementX/Sources/Other/Extensions/FileManager.swift
```diff
@@ -31,4 +31,21 @@ extension FileManager {
         }
         try createDirectory(at: url, withIntermediateDirectories: withIntermediateDirectories)
     }
+    
+    func copyFileToTemporaryLocation(url: URL) throws -> URL {
```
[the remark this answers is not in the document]
stefanceriu: Why is `Directory` better? I intentionally wanted to make it abstract so you don't care where it's copied to.

ElementX/Sources/Screens/RoomScreen/RoomScreenCoordinator.swift
```diff
@@ -74,8 +80,32 @@ final class RoomScreenCoordinator: CoordinatorProtocol {
...
+            case .selectMediaAtURL(let url):
+                let mediaPickerPreviewScreenCoordinator = MediaPickerPreviewScreenCoordinator(parameters: .init(url: url, title: url.lastPathComponent)) { action in
```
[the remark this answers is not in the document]
stefanceriu: I figured the `MediaPickerCoordinator` should only handle chosing a file through the system components. This particular preview is linked to the room media uploading and something like the avatar changes should probably use a different one. This way we can reuse the `MediaPickerCoordinator` in other places

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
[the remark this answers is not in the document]
stefanceriu: As far as I could tell this is what the original composer was doing too. With the flow disabled there shouldn't be any UI changes

ElementX/Sources/Services/Room/RoomProxy.swift
stefanceriu: Fair point, I wasn't sure what we were going to do with this from the UI side so I just left it in

ElementX/Sources/Screens/RoomScreen/RoomScreenCoordinator.swift
stefanceriu: I'm thinking I'll need to pass those into the view model and have it show an alert or something but I'm not sure yet

ElementX/Sources/Screens/MediaPickerPreviewScreen/View/MediaPickerPreviewScreenScreen.swift
stefanceriu: Oh 🤦

ElementX/Sources/Screens/MediaPickerPreviewScreen/MediaPickerPreviewScreenCoordinator.swift
stefanceriu: Something like `HomeCoordinator` would feel to generic. I would be inclined to actually add `Screen` to all of them

ElementX/Sources/Screens/MediaPicker/PhotoLibraryPicker.swift
stefanceriu: Oh yes, I need to check if it needs dispatching 👍

ElementX/Resources/Localizations/en.lproj/Untranslated.strings
```diff
@@ -24,3 +24,11 @@
...
 "soft_logout_clear_data_dialog_title" = "Clear data";
 "soft_logout_clear_data_dialog_content" = "Clear all data currently stored on this device?\nSign in again to access your account data and messages.";
+
+// MARK: - Media upload
+
+"media_upload_camera_picker" = "Camera";
+
+"media_upload_photo_and_video_picker" = "Photo & Video Library";
+
+"media_upload_document_picker" = "Document";
```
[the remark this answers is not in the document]
stefanceriu: Blatantly copied this from Whatsapp but it will most likely change when design comes back

ElementX/Sources/Other/Extensions/FileManager.swift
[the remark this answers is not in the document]
pixlwave: I would say that as its an extension on `FileManager` and `FileManager` works with directories it is clearer and more consistent with the existing API, but I see you point. (Although I do think there's a benefit for a developer to know that it is putting the data into the temporary directory).

ElementX/Sources/Application/AppSettings.swift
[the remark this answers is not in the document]
pixlwave: Yep I wouldn't be against aligning the other way, would just be nice to have a convention here :)

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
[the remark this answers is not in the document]
pixlwave: Derp yes you're right its disabled isn't it, my bad!

ElementX/Sources/Screens/MediaPicker/MediaPickerCoordinator.swift
stefanceriu: Talked about it offline, decide to keep them separate of ease of reuse, at least for now.

ElementX/Sources/Screens/MediaPicker/PhotoLibraryPicker.swift
stefanceriu: Nevermind, it's already dispatched

pixlwave (approved):
LGTM 👍

PR #751, opened 2023-03-30
Fix use of zsh in our tools.

pixlwave:
Just a tiny fix to unblock us for now, this PR makes sure `zsh` doesn't inherit its environment from `swift run`. We can look at launching the process directly instead of through a shell at a later date.


stefanceriu (commented):
Looks good to me too and considering this is a developer tool I think we can live with zsh just fine

PR #752, opened 2023-03-31
Bump the RustSDK, bump the app minor version and force user logout

stefanceriu:


ElementX/Sources/Services/UserSession/UserSessionStore.swift
```diff
@@ -38,6 +38,7 @@ class UserSessionStore: UserSessionStoreProtocol {
     
     /// Deletes all data stored in the shared container and keychain
     func reset() {
+        MXLog.warning("Resetting the UserSessionStore.")
```
pixlwave, line 41:
```suggestion
        MXLog.warning("Resetting the UserSessionStore. All accounts will be affected.")
```

pixlwave (approved):
LGTM, could we also add one of the resetting logs to `KeychainController.removeAllRestorationTokens`?

PR #754, opened 2023-03-31
Improve user search using the GET profile

alfogrillo:
Fixes https://github.com/vector-im/element-x-ios/issues/729

**Result**
r

### DIFF
--- a/ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
+++ b/ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
@@ -59,9 +59,6 @@ class StartChatViewModel: StartChatViewModelType, StartChatViewModelProtocol {
                     self.displayError(failure)
                 }
             }
-        case .showProfilePopover:
-            #warning("do proper action")
-            displayError(.failedGettingUserProfile)
         }
     }
     
--- a/ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
+++ b/ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
@@ -109,7 +109,7 @@ class StartChatViewModel: StartChatViewModelType, StartChatViewModelProtocol {
         // copies the current query to check later if fetched data must be shown or not
         let committedQuery = searchQuery
         
-        async let queriedProfile = getProfileIfNeeded()
+        async let queriedProfile = getProfileIfPossible()
         async let searchedUsers = clientProxy.searchUsers(searchTerm: committedQuery, limit: 5)
         
         await updateState(committedQuery: committedQuery,
@@ -156,7 +156,7 @@ class StartChatViewModel: StartChatViewModelType, StartChatViewModelProtocol {
         }
     }
     
-    private func getProfileIfNeeded() async -> UserProfile? {
+    private func getProfileIfPossible() async -> UserProfile? {
         guard searchQuery.isMatrixIdentifier else {
             return nil
         }
--- a/ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
+++ b/ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
@@ -96,12 +96,13 @@ class StartChatViewModel: StartChatViewModelType, StartChatViewModelProtocol {
     }
     
     private func fetchData() {
-        if searchQuery.isValidSearchQuery {
-            Task {
-                await searchProfiles()
-            }
-        } else {
+        guard searchQuery.count >= 3 else {
             fetchSuggestions()
+            return
+        }
+        
+        Task {
+            await searchProfiles()
         }
     }
     
@@ -189,10 +190,6 @@ class StartChatViewModel: StartChatViewModelType, StartChatViewModelProtocol {
 }
 
 private extension String {
-    var isValidSearchQuery: Bool {
-        count >= 3
-    }
-    
     var isMatrixIdentifier: Bool {
         MatrixEntityRegex.isMatrixUserIdentifier(self)
     }
--- a/ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
+++ b/ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
@@ -24,10 +24,6 @@ class StartChatViewModel: StartChatViewModelType, StartChatViewModelProtocol {
     
     var callback: ((StartChatViewModelAction) -> Void)?
     weak var userIndicatorController: UserIndicatorControllerProtocol?
-    private var fetchedProfiles: [String: UserProfile] = .init()
-    private var searchTask: Task<Void, Error>? {
-        didSet { oldValue?.cancel() }
-    }
     
     init(userSession: UserSessionProtocol, userIndicatorController: UserIndicatorControllerProtocol?) {
         self.userSession = userSession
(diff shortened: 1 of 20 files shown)


callumu:
Designs:

DM - https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X-(new)?node-id=1735%3A91051&t=ImpiMq9VI2MqCeg8-1

No profile information can be fetched DM


Room - https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X-(new)?node-id=1735%3A91238&t=ImpiMq9VI2MqCeg8-1

No profile information can be fetched Room

UnitTests/Sources/StartChatViewModelTests.swift
```diff
@@ -33,26 +33,84 @@ class StartChatScreenViewModelTests: XCTestCase {
...
-        XCTAssertEqual(context.viewState.usersSection.users.count, 1)
-        XCTAssertFalse(context.viewState.hasEmptySearchResults)
+        XCTAssertEqual(context.viewState.usersSection.users.count, count)
+        XCTAssertEqual(context.viewState.hasEmptySearchResults, count == 0)
+    }
+    
+    @discardableResult
+    private func search(query: String) async -> StartChatViewState? {
+        viewModel.context.searchQuery = query
+        return await context.$viewState.nextValue
```
stefanceriu, line 106: Nice 👏

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
```diff
@@ -92,31 +89,64 @@ class StartChatViewModel: StartChatViewModelType, StartChatViewModelProtocol {
...
     
-    private func updateState(searchQuery: String) {
-        searchTask = nil
-        
-        if searchQuery.count < 3 {
-            fetchSuggestions()
-        } else if MatrixEntityRegex.isMatrixUserIdentifier(searchQuery) {
-            state.usersSection = .init(type: .searchResult, users: [UserProfile(userID: searchQuery)])
+    private func fetchData() {
+        if searchQuery.isValidSearchQuery {
```
stefanceriu, line 99: I would've used a guard here, bailing out early is a very nice habit to get into

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
```diff
@@ -92,31 +89,64 @@ class StartChatViewModel: StartChatViewModelType, StartChatViewModelProtocol {
...
+            fetchSuggestions()
         }
     }
     
+    private func searchProfiles() async {
+        // copies the current query to check later if fetched data must be shown or not
+        let committedQuery = searchQuery
+        
+        async let queriedProfile = getProfileIfNeeded()
+        async let searchedUsers = clientProxy.searchUsers(searchTerm: committedQuery, limit: 5)
```
stefanceriu, line 113: 👌 on the concurrency

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
```diff
@@ -126,17 +156,20 @@ class StartChatViewModel: StartChatViewModelType, StartChatViewModelProtocol {
...
-        searchTask = Task { @MainActor in
-            guard
-                case let .success(result) = await userSession.clientProxy.searchUsers(searchTerm: searchTerm, limit: 5),
-                !Task.isCancelled
-            else {
-                return
-            }
-            
-            state.usersSection = .init(type: .searchResult, users: result.results)
+    private func getProfileIfNeeded() async -> UserProfile? {
```
stefanceriu, line 159: I wouldn't call this `needed`, more like a `possible` to me

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
```diff
@@ -126,17 +156,20 @@ class StartChatViewModel: StartChatViewModelType, StartChatViewModelProtocol {
...
+        guard searchQuery.isMatrixIdentifier else {
+            return nil
         }
+        
+        return try? await clientProxy.getProfile(for: searchQuery).get()
```
stefanceriu, line 164: Can't this result in false positives if `getProfile` fails with an actual error and not because the user doesn't exist?

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
```diff
@@ -154,3 +187,22 @@ class StartChatViewModel: StartChatViewModelType, StartChatViewModelProtocol {
         userIndicatorController?.retractIndicatorWithId(Self.loadingIndicatorIdentifier)
     }
 }
+
+private extension String {
```
stefanceriu, line 191: Not sure how I feel about these extensions considering they're not used in a lot of places, especially `isValidSearchQuery`. I would find the logic easier to understand if they would be closer to the place where they're actually used.

ElementX/Sources/Screens/StartChat/View/StartChatSuggestedUserCell.swift
```diff
@@ -29,17 +29,39 @@ struct StartChatSuggestedUserCell: View {
...
+                    .font(.compound.bodyXS)
+                    .padding(.top, 4)
+                }
             }
             .accessibilityElement(children: .combine)
         }
     }
 }
+
+private extension UserProfile {
```
stefanceriu, line 61: This doesn't belong on the view layer.

ElementX/Sources/UITests/UITestsScreenIdentifier.swift
```diff
@@ -50,6 +50,7 @@ enum UITestsScreenIdentifier: String {
     case reportContent
     case startChat
     case startChatWithSearchResults
+    case startChatSearchingInexistingID
```
stefanceriu, line 53: `nonexistent` please 🙃

stefanceriu (approved):
Left a few small comments inline, looks good to me otherwise 👏

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
alfogrillo: To me when I have two small arms the `if/else` looks easier to be understood.
If you have many things to do after a condition, a `guard` is better instead.
eg:

```swift
if cond {
    a
} else {
    b
}
```

Reads better than
```swift
guard cond else {
    b
}
a
```
I don't mind to make the change if people think this is less readable.

stefanceriu: In this particular way I don't mind either way honestly just pointing out the idea.

Another way to look at it would be from a semantical pov: `isValidSearchQuery` is a top level check without which `fetchingData` wouldn't work. It would make sense to be an early bailout.

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
alfogrillo: We don't need to distinguish client (the query not being a MXID) vs server errors in this case.
`nil` is ok for both.

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
alfogrillo: Maybe I can get rid of `isValidSearchQuery` since I use it just once?

ElementX/Sources/UITests/UITestsScreenIdentifier.swift
alfogrillo: Applied

ElementX/Sources/Screens/StartChat/View/StartChatSuggestedUserCell.swift
alfogrillo: Applied

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
alfogrillo: Applied

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
alfogrillo: Applied the suggestion.
Since we have a `Task` I think guard is looking good as well. ;)

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
alfogrillo: Applied

PR #758, opened 2023-04-03
change name to nil in direct room parameters

flescio:
The name of a direct room should be nil in order to reflect the other participant.


pixlwave (approved):
LGTM 👍

PR #760, opened 2023-04-03
Notifications Enabled

Velin92:
Enabled notifications with default static text.

Also implemented a way to get to the original user that sent the notification (which will be useful for multi account support)

WHAT IS MISSING:

- Permalink support
- A way to handle navigations where we set coordinators without animations


ElementX/Sources/Other/Extensions/Bundle.swift
```diff
@@ -54,6 +54,18 @@ public extension Bundle {
         }
     }
 
+    static var appBundle: Bundle {
```
stefanceriu, line 57: `Bundle.appBundle` sounds a bit weird, let's call it `Bundle.app`

ElementX/Resources/Localizations/de.lproj/Localizable.strings
stefanceriu, line 1: Why are these translation changes part of this PR? 🤔

ElementX/Sources/Services/Notification/Manager/NotificationManager.swift
```diff
@@ -84,15 +84,19 @@ class NotificationManager: NSObject, NotificationManagerProtocol {
...
-                        "loc-args": []
-                    ]
-                ]
-            ]
+                        "loc-args": [] as [Any]
+                    ] as [String: Any]
+                ] as [String: Any]
+            ] as [String: Any]
+            if let notificationID = clientProxy.restorationToken?.notificationID {
+                defaultPayload["notificationID"] = notificationID
```
stefanceriu, line 97: Please extract this `notificationID` into a constant somewhere. We should also rename it, check the other comment.

ElementX/Sources/Services/Notification/Proxy/NotificationItemProxy.swift
```diff
@@ -52,15 +52,15 @@ struct NotificationItemProxy {
 //    }
 
     var title: String {
-        "Title"
+        InfoPlistReader(bundle: .appBundle).bundleDisplayName
     }
 
     var subtitle: String? {
-        nil
+        L10n.notification
```
stefanceriu, line 59: How do translations even work within the NSE context? We need to make sure we're reading them properly from the main app bundle.

ElementX/Sources/Services/UserSession/RestorationToken.swift
```diff
@@ -19,6 +19,14 @@ import MatrixRustSDK
 
 struct RestorationToken: Codable, Equatable {
     let session: MatrixRustSDK.Session
+    let notificationID: Int
+
+    init(session: MatrixRustSDK.Session) {
+        self.session = session
+        var hasher = Hasher()
+        hasher.combine(session.userId)
```
stefanceriu, line 27: I would use CryptoKit and a standard hash function implementation instead of whatever haher uses. That way it would be easier to track this identifier when debugging

ElementX/Sources/Services/UserSession/RestorationToken.swift
```diff
@@ -19,6 +19,14 @@ import MatrixRustSDK
 
 struct RestorationToken: Codable, Equatable {
     let session: MatrixRustSDK.Session
+    let notificationID: Int
```
stefanceriu, line 22: Let's please name this something along the lines of `UniqueNotificationPusherIdentifer` or similar.

NSE/Sources/NotificationServiceExtension.swift
```diff
@@ -36,11 +36,13 @@ class NotificationServiceExtension: UNNotificationServiceExtension {
         guard !DataProtectionManager.isDeviceLockedAfterReboot(containerURL: URL.appGroupContainerDirectory),
               let roomId = request.roomId,
               let eventId = request.eventId,
-              let credentials = keychainController.restorationTokens().first else {
+              let notificationID = request.content.userInfo["notificationID"] as? Int,
```
stefanceriu, line 39: Same comment as above about extracting it into a constant.

Velin92:
> Looks good to me but I left some changes inline.
> 
> Also, today's merge commit should've been a rebase instead.

actually regarding this... I used the default github update branch option, but this always defaults to a merge, at this point could we just disable it, or there is a way to force it to be always a rebase?

ElementX/Resources/Localizations/de.lproj/Localizable.strings
Velin92: localazy download, just downloads everything, I just wanted to add a small change in a copy (use the action_share key instead of the action_share_link key),

ElementX/Sources/Services/Notification/Proxy/NotificationItemProxy.swift
stefanceriu: I don't think that's how it works: the translations will use the current bundle and search for the lproj in it. Seeing as you're in the NSE that will fail

Velin92: I see here that the used Bundle is initialised through a url of the lproj:
```
let bundle = Bundle(url: lprojURL)
Self.cachedLocalizationBundles[language] = bundle
```
[which points to the file specified by the url](https://developer.apple.com/documentation/foundation/bundle/1409352-init)

You know what? Let me try adding the italian translation for the `Notification` key on localazy and try it, so we can test if the translation gets the appropriate value

ElementX/Sources/Services/Notification/Manager/NotificationManager.swift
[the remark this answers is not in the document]
Velin92: yes! time to make this a struct

ElementX/Sources/Services/Notification/Proxy/NotificationItemProxy.swift
stefanceriu: Yes, but that `url(forResource: language, withExtension: "lproj")` isn't available in the NSE bundle, just the main app one.

> You know what? Let me try adding the italian translation for the Notification key on localazy and try it
You could just modify the localization files locally, no need to bother with Localazy

Velin92: Yeah, I see what you mean know, because the url function defaults to the path of NSE. Okay I'll look into it

ElementX/Sources/Services/UserSession/RestorationToken.swift
Velin92: I choose SHA256 I hope is okay

ElementX/Sources/Services/Notification/Proxy/NotificationItemProxy.swift
Velin92: Okay surprisingly enough now that the translation bug is fixed, we don't need to do actually anything!
the url(forResource) function is actually smart enough to get the resource directly from the main app, without the need to go through its bundle, or copying the bundle resources into the NSE:
`_url	NSURL	"file:///var/containers/Bundle/Application/43EFEBC2-A8BE-4CB2-A657-C7CE186607AB/ElementX.app/it.lproj/"	0x000000010610a280`

stefanceriu: Yey 🎉

stefanceriu (approved):
🚢 it!

PR #762, opened 2023-04-04
m.room.join_rules added as a required state

Velin92:
This allow the `isPublic` value to be correct when used

### DIFF
--- a/ElementX/Sources/Services/Client/ClientProxy.swift
+++ b/ElementX/Sources/Services/Client/ClientProxy.swift
@@ -379,7 +379,8 @@ class ClientProxy: ClientProxyProtocol {
     }
     
     private lazy var slidingSyncRequiredState = [RequiredState(key: "m.room.avatar", value: ""),
-                                                 RequiredState(key: "m.room.encryption", value: "")]
+                                                 RequiredState(key: "m.room.encryption", value: ""),
+                                                 RequiredState(key: "m.room.join_rules", value: "")]
     
     private lazy var slidingSyncFilters = SlidingSyncRequestListFilters(isDm: nil,
                                                                         spaces: [],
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -379,7 +379,8 @@ class ClientProxy: ClientProxyProtocol {
     }
     
     private lazy var slidingSyncRequiredState = [RequiredState(key: "m.room.avatar", value: ""),
-                                                 RequiredState(key: "m.room.encryption", value: "")]
+                                                 RequiredState(key: "m.room.encryption", value: ""),
+                                                 RequiredState(key: "m.room.join_rules", value: "")]
```
pixlwave, line 383: Considering we don't have a context menu on the room list yet, we don't need this state here - it will slow down the sync responses. Lets add required state only when we need it.

ElementX/Sources/Services/Room/RoomProxy.swift
```diff
@@ -139,7 +139,8 @@ class RoomProxy: RoomProxyProtocol {
         
     func addTimelineListener(listener: TimelineListener) -> Result<[TimelineItem], RoomProxyError> {
         let settings = RoomSubscription(requiredState: [RequiredState(key: "m.room.topic", value: ""),
-                                                        RequiredState(key: "m.room.canonical_alias", value: "")],
+                                                        RequiredState(key: "m.room.canonical_alias", value: ""),
+                                                        RequiredState(key: "m.room.join_rules", value: "")],
```
pixlwave, line 143: I'm surprised the trailing `,` didn't trigger a warning 🤔
```suggestion
                                                        RequiredState(key: "m.room.join_rules", value: "")]
```

Velin92: it's a function argument

pixlwave (approved):

PR #763, opened 2023-04-04
Improve tests' reliability

alfogrillo:
This PR:
- Refactors unit tests removing `Task.sleep` or `Task.yield` whenever is possible with the goal to make tests more reliable.
- Makes `process(viewAction: action)` a sync method
- Fixes integration tests env variables


UnitTests/Sources/Extensions/ViewModelContext.swift
```diff
@@ -0,0 +1,24 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+@testable import ElementX
+
+extension ViewModelContext {
+    @discardableResult
+    func nextViewState() async -> ViewState? {
```
stefanceriu, line 21: Can't we put this directly where the `ViewModelContext` is defined? 🤔

ElementX/Sources/Screens/RoomMemberDetails/RoomMemberDetailsViewModel.swift
```diff
@@ -32,21 +32,22 @@ class RoomMemberDetailsViewModel: RoomMemberDetailsViewModelType, RoomMemberDeta
...
+            Task { await ignoreUser() }
         case .unignoreConfirmed:
-            await unignoreUser()
+            Task { await unignoreUser() }
         }
     }
 
     // MARK: - Private
 
+    @MainActor
```
stefanceriu, line 50: Do these really need to be dispatched on the main actor, what happens without it? 🤔

UITests/Sources/StartChatScreenUITests.swift
```diff
@@ -50,7 +50,7 @@ class StartChatScreenUITests: XCTestCase {
     }
     
     func testSearchExactNotExistingMatrixID() {
-        let app = Application.launch(.startChatSearchingInexistingID)
+        let app = Application.launch(.startChatSearchingNonExistingID)
```
stefanceriu, line 53:
```suggestion
        let app = Application.launch(.startChatSearchingNonexistentID)
```

stefanceriu (approved):
Such a nice addition, very well done 👏 

Let's also please add publishers to the template vm and coordinator so new screens are done the right way from the start

UnitTests/Sources/Extensions/ViewModelContext.swift
alfogrillo: Wanted to put this into an extension just for test targets.
We don't want the EX app to use this stuff right?

ElementX/Sources/Screens/RoomMemberDetails/RoomMemberDetailsViewModel.swift
alfogrillo: These methods make side effect on the view state which is supposed to be updated just from the main thread right?

stefanceriu: Ah right, good point! 👍

UnitTests/Sources/Extensions/ViewModelContext.swift
stefanceriu: Ah right, good point! 👍

ElementX/Sources/Screens/RoomMemberDetails/RoomMemberDetailsViewModel.swift
stefanceriu: Right right, but I guess what I'm asking is why they go off the main thread to being with

UITests/Sources/StartChatScreenUITests.swift
alfogrillo: Applied

ElementX/Sources/Screens/RoomMemberDetails/RoomMemberDetailsViewModel.swift
alfogrillo: They are called like this:
```swift
Task { await ignore() }
```

However I noticed now the whole view model protocol is a main actor:
```swift
@MainActor
protocol RoomDetailsViewModelProtocol
```
So I can delete them. 👍🏻

stefanceriu: Awesome, thanks 👍

pixlwave (approved):

UnitTests/Sources/BugReportViewModelTests.swift
```diff
@@ -70,19 +68,19 @@ class BugReportViewModelTests: XCTestCase {
...
+        
+        _ = await viewModel
+            .callbackPublisher
+            .filter {
+                guard case .submitFinished = $0 else {
+                    return false
+                }
+                return true
+            }
+            .firstValue
```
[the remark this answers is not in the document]
alfogrillo: Like this slightly more.
Added also a convenience method in `AsyncSequence` to get just the first value when needed.
[Here](https://github.com/vector-im/element-x-ios/pull/763/commits/4ec7597b31b2756a456552f1b2189662477758f4) 
Wdyt?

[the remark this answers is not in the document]
pixlwave: Oh yeah, that's nice 👍

PR #765, opened 2023-04-04
Fixed translation bug when the locale contained the region

Velin92:
Also added some tests to prevent the issue from happening again... and the italian translation


UnitTests/Sources/LocalizationTests.swift
```diff
@@ -19,87 +19,77 @@ import XCTest
...
-        XCTAssertEqual(L10n.inviteFriendsText("Element", "element.io"), "Yalnızca Söz ve Anahtar Kelimeler")
+        XCTAssertEqual(L10n.testLanguageIdentifier, "it")
     }
 
     /// Test fallback language for a language not supported at all
-    func disabled_testFallbackOnNotSupportedLanguage() {
-        //  set app language to something Element don't support at all (chose 'Malay' language)
-        Bundle.elementLanguage = "ms"
+    func testStripRegionIfRegionalTranslationIsNotAvailable() {
+        //  set app language to something that includes also a region (it-IT)
```
stefanceriu, line 36:
```suggestion
        // set app language to something that includes also a region (it-IT)
```

UnitTests/Sources/LocalizationTests.swift
```diff
@@ -19,87 +19,77 @@ import XCTest
 
 class LocalizationTests: XCTestCase {
```
stefanceriu, line 20: Very nice 👌

ElementX/Sources/Generated/Strings+Untranslated.swift
```diff
@@ -71,24 +71,12 @@ public enum UntranslatedL10n {
 
 extension UntranslatedL10n {
   static func tr(_ table: String, _ key: String, _ args: CVarArg...) -> String {
-    let languages = Bundle.preferredLanguages
```
stefanceriu, line 74: What's the new logic here, don't we need preferred languages anymore? If so shouldn't delete the implementation from the Bundle extension?

stefanceriu (approved):
Looks good to me but honestly I'm still a bit confused on how this actually works within the NSE context.

ElementX/Sources/Generated/Strings+Untranslated.swift
Velin92: essentially Untranslated is just a collection of strings that are set in the "en" language, in fact if you check the project navigator other languages are not selected (if you select one Xcode will automatically create the collection of string, but we dont' need more than one for Untranslated ofc)
Before we were cycling through AAAAAALL the languages, but for Untranslated is not really needed to do all this shenanigans, so we just go straight to the point and always get the "en" strings inside Untranslated.

Also yeah we need preferred language we just use it for L10n instead for UntranslatedL10n.

stefanceriu: Oooh I just realised these are the untranslated ones, sorry, my bad 🤦

Velin92:
> Looks good to me but honestly I'm still a bit confused on how this actually works within the NSE context.

@stefanceriu 
Yeah I'll explain it better in the other PR when is merged but essentially the `url(resource, extension)` function just gets any resource that is assigned to the target, so we can either assign the strings directly to the NSE target (by just adding their import in the target.yml), or we can get them from the Bundle.app. Better to discuss this in the other PR

stefanceriu:
> Better to discuss this in the other PR

Fair but I think this is the way to go:
>or we can get them from the Bundle.app.

PR #768, opened 2023-04-05
Translations workflow and tools

Velin92:
- No need to use localazy download (actually that will fail now since I have changed the name of the jsons so that we can only add translation through our controlled tool)
- You can run `swift run tools download-translations` which will just download the english keys from localazy, so that we don't get too overwhelmed or annoyed by huge string changes, but we only fetch the basic keys we need for implementation purposes
- If you instead want ALL the translations just run `swift run tools download-translations --all`
- Also created a workflow that should automatically create a PR every week with all translation updates from all the languages


Tools/Sources/DownloadTranslations.swift
```diff
@@ -0,0 +1,23 @@
...
+        try localazyDownload()
+    }
+
+    private func localazyDownload() throws {
+        let json: String
+        if all {
+            json = "localazy-all.json"
+        } else {
+            json = "localazy-en.json"
+        }
```
pixlwave, line 20: Totally nit picky and unimportant, but seems simpler to use a ternary in this instance.
```suggestion
        let json = all ? "localazy-all.json" : "localazy-en.json"
```

.github/workflows/translations-pr.yml
```diff
@@ -0,0 +1,32 @@
...
+
+      - name: Setup project
+        run: swift run tools setup-project
+
+      - name: Download All Translations
+        run: swift run tools download-translations --all
+
+      - name: Create PR for Translations
+        uses: peter-evans/create-pull-request@v5
+        with:
```
pixlwave, line 26: Is there anything we could add in here to make sure it only includes `ElementX/Resources/Localizations`? I'm thinking if for some random reason the Package.resolved was updated, we probably wouldn't want that included in this PR.

ElementX/Resources/Localizations/es.lproj/Localizable.stringsdict
```diff
@@ -0,0 +1,78 @@
...
+			<key>NSStringFormatSpecTypeKey</key>
+			<string>NSStringPluralRuleType</string>
+			<key>NSStringFormatValueTypeKey</key>
+			<string>d</string>
+			<key>zero</key>
+			<string>No hay miembros</string>
+			<key>one</key>
+			<string>%1$d miembro</string>
+			<key>two</key>
+			<string></string>
```
pixlwave, line 20: Hmmm, unrelated but these empty plurals aren't ok.

Maybe we should also add a locheck run to the PR. It would (hopefully) identify the cases that don't have a placeholder to alert us that it shouldn't be merged.

.github/workflows/translations-pr.yml
Velin92: we could even force exactVersion on the package.swift to avoid this issue entirely

Velin92: otherwise I guess I could use this:


add-paths | A comma or newline-separated list of file paths to commit. Paths should follow git's pathspec syntax. If no paths are specified, all new and modified files are added. See Add specific paths.
-- | --


add-paths	A comma or newline-separated list of file paths to commit. Paths should follow git's [pathspec](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefpathspecapathspec) syntax. If no paths are specified, all new and modified files are added. See [Add specific paths](https://github.com/peter-evans/create-pull-request#add-specific-paths).

pixlwave: `add-paths` sounds good to me 👍

ElementX/Resources/Localizations/es.lproj/Localizable.stringsdict
Velin92: locheck lives, in our hearts, and in this branch now.

Tools/Sources/Locheck.swift
```diff
@@ -0,0 +1,44 @@
...
+            switch self {
+            case .missingMint:
+                return "💥 Unable to find mint. Fix by running:\nbrew install mint\n"
+            case .outputError:
+                return "💥 Failed to read the output from locheck."
+            }
+        }
+    }
+
+    static var configuration = CommandConfiguration(abstract: "A tool that verifies bad strings containes in localization files")
```
pixlwave, line 19:
```suggestion
    static var configuration = CommandConfiguration(abstract: "A tool that verifies bad strings contained in localization files")
```

pixlwave (approved):
Nice, above and beyond as always 👏

Tools/Sources/DownloadTranslations.swift
```diff
@@ -0,0 +1,18 @@
+import ArgumentParser
+import Foundation
+
+struct DownloadTranslations: ParsableCommand {
```
bmarty, line 4: My 2 cents: since it will not always download translations, maybe rename to `DownloadStrings`?

Velin92: uhm yeah maybe should be DownloadString and the command --allTranslations

PR #769, opened 2023-04-05
Improve view model/context architecture

alfogrillo:
This PR:
- Embeds a `Context` into the associated view model reducing boilerplate on generic types
- Deletes the publisher `viewActions` since now a context can talk directly with the view model
- Makes the `Context.init` private since it doesn't make sense creating `Context` outside the view model.


stefanceriu (approved):
As discussed offline, great change. Let's try it out 👍

pixlwave (approved):

PR #770, opened 2023-04-05
Guard user suggestions behind feature flag so that they don't impact releasability of other room creation features

Johennes:
Suggestions have been descoped from https://github.com/vector-im/element-meta/issues/988. This wraps them into a separate feature switch so that we can still safely release other pieces of the epic.


ElementX/Sources/Screens/StartChat/View/StartChatScreen.swift
```diff
@@ -89,7 +89,7 @@ struct StartChatScreen: View {
                 }
             }
         } header: {
-            if let title = context.viewState.usersSection.type.title {
+            if !context.viewState.usersSection.users.isEmpty, let title = context.viewState.usersSection.type.title {
```
alfogrillo, line 92: I feel like we should make `context.viewState.usersSection` optional. 
But it looks like a good workaround for now

alfogrillo (approved):
Just a minor comment.
LGTM otherwise.

ElementX/Sources/Application/AppSettings.swift
```diff
@@ -164,6 +165,9 @@ final class AppSettings: ObservableObject {
     @UserSetting(key: UserDefaultsKeys.startChatFlowEnabled.rawValue, defaultValue: false, persistIn: store)
     var startChatFlowEnabled
     
+    @UserSetting(key: UserDefaultsKeys.startChatUserSuggestionsEnabled.rawValue, defaultValue: false, persistIn: store)
```
pixlwave, line 168: I would probably suggest not persisting this one (at least until the feature is brought back into scope for someone to work on).
```suggestion
    @UserSetting(key: UserDefaultsKeys.startChatUserSuggestionsEnabled.rawValue, defaultValue: false, persistIn: nil)
```

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
```diff
@@ -142,7 +142,7 @@ class StartChatViewModel: StartChatViewModelType, StartChatViewModelProtocol {
     }
     
     private func fetchSuggestions() {
-        state.usersSection = .init(type: .suggestions, users: [.mockAlice, .mockBob, .mockCharlie])
+        state.usersSection = .init(type: .suggestions, users: ServiceLocator.shared.settings.startChatUserSuggestionsEnabled ? [.mockAlice, .mockBob, .mockCharlie] : [])
```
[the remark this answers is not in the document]
Johennes: Hm, I think this may not work the way you had intended it. Without setting `state.usersSection` it won't reset the search results when you erase the input. Say you enter "ddd", wait for the results to load, then delete the input. You're then left with this state:

simulator_screenshot_43A11A4D-3D34-47CB-9902-6D200FE86045

But I think we'd want it to instead show nothing in this case?

ElementX/Sources/Application/AppSettings.swift
Johennes: Oh interesting, I hadn't thought of this. What's your thinking with not persisting it?

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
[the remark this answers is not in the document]
alfogrillo: I see the problem.
While the section stays non optional, I think we can fix it in the `fetchData` like this:

```swift
    private func fetchData() {
       state.usersSection = .init(type: .suggestions, users: [])
        
       guard searchQuery.count >= 3 else {
            fetchSuggestions()
            return
        }
        
        Task {
            await searchProfiles()
        }
    }
```

wdyt?

ElementX/Sources/Application/AppSettings.swift
pixlwave: Ideally we don't want to persist feature flags unless we really need to. The thought being that someone could toggle a load of random flags (that might not play well with each other), get their app in a mess of some unhandled state and end up reporting bugs that don't really exist. I suggested to do the same for `startChatFlowEnabled` as I personally don't see the value in persisting this one either.

ElementX/Sources/Screens/StartChat/View/StartChatScreen.swift
Johennes: Have made it so

ElementX/Sources/Application/AppSettings.swift
Johennes: Ok, got it. Have made it so.

For `startChatFlowEnabled` I think it's ok to persist because we're actively working on this and without persisting you'd have to turn it on again after every local build.

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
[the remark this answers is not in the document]
Johennes: Ok, have made it so. One small remaining problem I'm seeing is that the form background ends up white when there is no content. So as you type and results are added and removed, it flickers back and forth between white and grey background a lot.

Simulator Screenshot - iPhone 14 Pro - 2023-04-06 at 11 40 24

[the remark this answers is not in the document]
alfogrillo: Ok, let's stick with your proposal for now.
We will refactor the section as optional later.
Thanks for trying it 👍🏻

ElementX/Sources/Application/AppSettings.swift
pixlwave: Thank you!

You could set the `default` value to `true` using it like a build flag for local builds.

Johennes: Ah, that's true, yes. Well, I'll leave this one alone for now since it's Alfonso's. 😅

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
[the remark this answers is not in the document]
Johennes: Hm, I had to move the state setting back into `fetchSuggestions` because it broke the tests ... which would have had to wait for two state updates now. Hope that's ok.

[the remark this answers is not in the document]
alfogrillo: Yeah this is what I meant! ;) 👍🏻

PR #771, opened 2023-04-05
Remove styling for developer toggles

Johennes:
I know this may not yet be decided but there's simply no way I'm gonna be able to look at two different shades of green on this screen. 😅


stefanceriu (approved):
No no, this is a mistake, we don't want any kind of styling on this screen 👍

Issue #748, opened 2023-03-30
[MediaUpload] Media pre-processing


stefanceriu:
Before we can upload media we need to take it through several pre-processing steps:
* resize large images and videos (videos are hard to convert rust side)
* generate thumbnails
* strip exif information
* generate media information
* potentially generate a blurhash

PR #772, opened 2023-04-06
#748 - Implement methods for processing various media types and prepa…

stefanceriu:
…ring them for upload

This PR introduced a `MediaUploadingPreprocessor` that is responsible for processing any kind of media file and returning the final file and thumbnail URLs and their associated metadata. These should be then passed over to the rust side for the actual upload

What's missing (both will come as separate PRs)
- [ ] Unit tests
- [ ] RustSDK support

Will do both as separate PRs


ElementX/Sources/Services/Media/MediaUploadingPreprocessor.swift
```diff
@@ -0,0 +1,350 @@
...
+    /// - Returns: the URL for the resulting thumbnail and its sizing info as an `ImageProcessingResult`
+    private func generateThumbnailForImageAt(_ url: URL) async -> Result<ImageProcessingResult, MediaUploadingPreprocessorError> {
+        guard let data = try? Data(contentsOf: url),
+              let image = UIImage(data: data) else {
+            return .failure(.failedGeneratingImageThumbnail(nil))
+        }
+        
+        switch await resizeImage(image, targetSize: Constants.maximumThumbnailSize) {
+        case .success(let thumbnail):
+            guard let data = thumbnail.pngData() else {
```
pixlwave, line 205: I would think we should send thumbnails as `.jpeg` rather than `.png` no? For photos this could make a big difference to the file size.

ElementX/Sources/Services/Media/MediaUploadingPreprocessor.swift
```diff
@@ -0,0 +1,350 @@
...
+    private func generateThumbnailForVideoAt(_ url: URL) async -> Result<ImageProcessingResult, MediaUploadingPreprocessorError> {
+        let assetImageGenerator = AVAssetImageGenerator(asset: AVAsset(url: url))
+        assetImageGenerator.appliesPreferredTrackTransform = true
+        assetImageGenerator.maximumSize = Constants.maximumThumbnailSize
+        
+        do {
+            // Avoid the first frames as on a lot of videos they're black.
+            // If the specified seconds are longer than the actual video a frame close to the end of the video will be used, at AVFoundation's discretion
+            let location = CMTime(seconds: Constants.videoThumbnailTime, preferredTimescale: 1)
+            let cgImage = try assetImageGenerator.copyCGImage(at: location, actualTime: nil)
```
pixlwave, line 265: Note: This method is deprecated as it is synchronous. There's a new awaitable `assetImageGenerator.image(at: time)` we could use instead.

ElementX/Sources/Services/Media/MediaUploadingPreprocessor.swift
```diff
@@ -0,0 +1,350 @@
...
+            return .failure(.failedGeneratingImageThumbnail(error))
+        }
+    }
+    
+    /// Aspect ratio resizes an image so it fits in the given size
+    /// - Parameters:
+    ///   - image: the original image
+    ///   - targetSize: maximum resulting size
+    /// - Returns: the resized image
+    private func resizeImage(_ image: UIImage, targetSize: CGSize) async -> Result<UIImage, MediaUploadingPreprocessorError> {
```
pixlwave, line 226: Doing a bit of [reading](https://stackoverflow.com/questions/5860215/resizing-a-uiimage-without-loading-it-entirely-into-memory), I think we should try and use `ImageIO` here instead of `UIGraphicsImageRenderer` as this should save on memory usage for the share extension and allow larger images to be sent.

I think we can also keep the CGImage and write PNG/JPEG data to disk from there too for the same benefits. Sorry for suggesting more CG apis 🙈

ElementX/Sources/Services/Media/MediaUploadingPreprocessor.swift
```diff
@@ -0,0 +1,350 @@
...
+        var fileName = url.deletingPathExtension().appendingPathExtension("mp4").lastPathComponent
+        let isOriginalFileInTemporaryDirectory = FileManager.default.temporaryDirectoryURLFor(fileName: fileName) == url
+        
+        // AVAssetExportSession will fail if the output URL already exists
+        if isOriginalFileInTemporaryDirectory {
+            fileName = "converted-\(fileName)"
+        }
+        let outputURL = FileManager.default.temporaryDirectoryURLFor(fileName: fileName)
+        
+        try? FileManager.default.removeItem(at: outputURL)
```
pixlwave, line 306: Ok so seeing this makes me worry what happens if I go and send the same file to a couple of friends separately. We could conceivably have multiple uses of the same file URL given slow encoding/uploading processes? I think we should discuss the file naming a bit more.

pixlwave (commented):
This is such a neat way of implementing it 👏

ElementX/Sources/Other/BlurHashDecode.swift
```diff
@@ -142,5 +142,3 @@ private extension String {
         return self[start..<end]
     }
 }
-
-// swiftlint:enable all
```
pixlwave, line 146: There's a new SwiftLint rule that means we need to put this one back. Alternatively I wonder if we should have a directory for 3rd party code which we can then exclude from linting, formatting & sonarcloud?

ElementX/Sources/Services/Media/MediaUploadingPreprocessor.swift
```diff
@@ -57,60 +62,71 @@ private struct VideoProcessingResult {
...
-    func processMedia(url: URL) async -> Result<MediaInfo, MediaUploadingPreprocessorError> {
+    /// - Returns: a specific type of `MediaInfo` depending on the file type and its associated details
+    func processMedia(at url: URL) async -> Result<MediaInfo, MediaUploadingPreprocessorError> {
+        // Start by moving the file to a unique temporary location in order to avoid conflicts if processing it multiple times
+        // All the other operations will be made relative to it
+        let uniqueFolder = FileManager.default.temporaryDirectory.appendingPathComponent(UUID().uuidString)
+        let newURL = uniqueFolder.appendingPathComponent(url.lastPathComponent)
+        do {
+            try FileManager.default.createDirectory(at: uniqueFolder, withIntermediateDirectories: true)
+            try FileManager.default.moveItem(at: url, to: newURL)
```
pixlwave, line 80: How come this is a move instead of a copy? Could we risk moving something out of the user's documents directory?

ElementX/Sources/Services/Media/MediaUploadingPreprocessor.swift
```diff
@@ -0,0 +1,416 @@
...
+    
+    /// Removes the GPS dictionary from an image's metadata
+    /// - Parameters:
+    ///   - url: the URL for the original image
+    ///   - type: its UTType
+    /// - Returns: the URL for the modified image and its size as an `ImageProcessingResult`
+    private func stripLocationFromImage(at url: URL, type: UTType, mimeType: String) async -> Result<ImageProcessingInfo, MediaUploadingPreprocessorError> {
+        guard let originalData = NSData(contentsOf: url),
+              let originalCGImage = UIImage(data: originalData as Data)?.cgImage,
+              let imageSource = CGImageSourceCreateWithData(originalData, nil) else {
```
pixlwave, line 221: I think these 3 lines could be simplified down to
```suggestion
        guard let imageSource = CGImageSourceCreateWithURL(url as CFURL, nil) else {
```

ElementX/Sources/Services/Media/MediaUploadingPreprocessor.swift
```diff
@@ -0,0 +1,416 @@
...
+        assetImageGenerator.appliesPreferredTrackTransform = true
+        assetImageGenerator.maximumSize = Constants.maximumThumbnailSize
+        
+        do {
+            // Avoid the first frames as on a lot of videos they're black.
+            // If the specified seconds are longer than the actual video a frame close to the end of the video will be used, at AVFoundation's discretion
+            let location = CMTime(seconds: Constants.videoThumbnailTime, preferredTimescale: 1)
+            let cgImage = try await assetImageGenerator.image(at: location).image
+            
+            switch await resizeImage(UIImage(cgImage: cgImage), targetSize: Constants.maximumThumbnailSize) {
```
pixlwave, line 334: Give that you've passed a `maximumSize` to the image generator, this resize step is unnecessary right? And then…

ElementX/Sources/Services/Media/MediaUploadingPreprocessor.swift
```diff
@@ -0,0 +1,416 @@
...
+    }
+    
+    private func resizeImage(_ image: UIImage, targetSize: CGSize) async -> Result<UIImage, MediaUploadingPreprocessorError> {
+        guard let data = image.pngData(),
+              let imageSource = CGImageSourceCreateWithData(data as NSData, nil) else {
+            return .failure(.failedResizingImage)
+        }
+        
+        return await resizeImage(withSource: imageSource, targetSize: targetSize)
+    }
```
pixlwave, line 284: … this method isn't needed anymore.

pixlwave (approved):
Nice 👏

ElementX/Sources/Other/BlurHashDecode.swift
stefanceriu: We would probably need to move the generated code and mocks in there too.. let's look into it later

ElementX/Sources/Services/Media/MediaUploadingPreprocessor.swift
stefanceriu: Well it can but I need the `originalCGImage` to read its size and `CGImageDestinationAddImage` it back to the destination

ElementX/Sources/Services/Media/MediaUploadingPreprocessor.swift
stefanceriu: Oh yes, of course, nice catch!

ElementX/Sources/Services/Media/MediaUploadingPreprocessor.swift
stefanceriu: I honestly don't know if that would even possible, but you're right, we should definitely not rely on it. Not sure what I was thinking, even the comment says `Start by moving..` 🙈

PR #773, opened 2023-04-06
Removed string references from UI Tests

Velin92:
We don't really need these anymore, we can use identifiers or just asserting screenshots.


pixlwave (approved):
LGTM 🙏

PR #778, opened 2023-04-06
Removing the about section title + Fix notification tap crash

Velin92:
A small to update the designs.

And a fix to the issue that notifications crash


UITests/Sources/__Snapshots__/Application/en-GB-iPad-9th-generation.roomDetailsScreen.png
pixlwave, line 1: I think all of these deleted iPad screenshots still need to be generated?

pixlwave (approved):
Code LGTM for both the room details and handling the notification navigation via an `AppRoute`. Only thing is the comment about iPad screenshots.

UITests/Sources/__Snapshots__/Application/en-GB-iPad-9th-generation.roomDetailsScreen.png
Velin92: oh weird maybe I forgot to push them

PR #781, opened 2023-04-07
Adopt new timeline item debug info and overhaul the timeline item deb…

stefanceriu:
…ug menu

This will show 3 collapsible sections (if available) on the debug menu like so:

Screenshot 2023-04-07 at 12 40 49

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/View/TimelineItemDebugView.swift
+++ b/ElementX/Sources/Screens/RoomScreen/View/TimelineItemDebugView.swift
@@ -26,16 +26,21 @@ struct TimelineItemDebugView: View {
     var body: some View {
         NavigationStack {
             ScrollView {
-                Divider()
-                
                 VStack {
-                    disclosureGroup(title: "Model", text: info.model)
-                    disclosureGroup(title: "Original JSON", text: info.originalJson)
-                    disclosureGroup(title: "Latest edit JSON", text: info.latestEditJson)
+                    TimelineItemInfoDisclosureGroup(title: "Model", text: info.model, isExpanded: $isModelExpanded)
+                    
+                    if let originalJSONInfo = info.originalJSON {
+                        TimelineItemInfoDisclosureGroup(title: "Original JSON", text: originalJSONInfo)
+                    }
+                    
+                    if let latestEditJSONInfo = info.latestEditJSON {
+                        TimelineItemInfoDisclosureGroup(title: "Latest edit JSON", text: latestEditJSONInfo)
+                    }
                 }
             }
             .navigationBarTitleDisplayMode(.inline)
             .navigationTitle("Timeline item")
+            .toolbarBackground(.visible, for: .navigationBar)
             .toolbar {
                 ToolbarItem(placement: .cancellationAction) {
                     Button(L10n.actionCancel) {
@@ -54,41 +59,44 @@ struct TimelineItemDebugView: View {
     
     // MARK: - Private
     
-    @ViewBuilder
-    private func disclosureGroup(title: String, text: String?, isExpanded: Binding<Bool>? = nil) -> some View {
-        if let text {
+    private struct TimelineItemInfoDisclosureGroup: View {
+        let title: String
+        let text: String
+        var isExpanded: Binding<Bool>?
+        
+        var body: some View {
             VStack(spacing: 0.0) {
                 Group {
                     if let isExpanded {
                         DisclosureGroup(title, isExpanded: isExpanded) {
-                            groupContent(for: text)
+                            disclosureGroupContent
                         }
                     } else {
                         DisclosureGroup(title) {
-                            groupContent(for: text)
+                            disclosureGroupContent
                         }
                     }
                 }
-                .listRowInsets(.none)
                 .font(.element.subheadline)
                 .padding()
                 
                 Divider()
             }
         }
-    }
-    
-    private func groupContent(for text: String) -> some View {
-        VStack(alignment: .leading) {
-            Spacer()
-            
-            Divider()
-            
-            Text(text)
-                .font(.element.caption1.monospaced())
-                .foregroundColor(.element.primaryContent)
+        
+        @ViewBuilder
+        var disclosureGroupContent: some View {
+            VStack(alignment: .leading) {
+                Spacer()
+                
+                Divider()
+                
+                Text(text)
+                    .font(.element.caption1.monospaced())
+                    .foregroundColor(.element.primaryContent)
+            }
+            .frame(maxWidth: .infinity)
         }
-        .frame(maxWidth: .infinity)
     }
 }
 
@@ -110,7 +118,7 @@ struct TimelineItemDebugView_Previews: PreviewProvider {
     
     static var previews: some View {
         TimelineItemDebugView(info: .init(model: smallContent,
-                                          originalJson: smallContent,
-                                          latestEditJson: smallContent))
+                                          originalJSON: "{\"Hi\": \"Alice\"}",
+                                          latestEditJSON: "{\"Hi\": \"Bob\"}"))
     }
 }
--- a/ElementX/Sources/Screens/RoomScreen/View/TimelineItemDebugView.swift
+++ b/ElementX/Sources/Screens/RoomScreen/View/TimelineItemDebugView.swift
@@ -17,40 +17,78 @@
 import SwiftUI
 
 struct TimelineItemDebugView: View {
-    struct DebugInfo: Identifiable {
-        let id = UUID()
-        let title: String
-        var content: String
-    }
-    
     @Environment(\.dismiss) private var dismiss
     
-    let info: DebugInfo
+    @State private var isModelExpanded = true
+    
+    let info: TimelineItemDebugInfo
     
     var body: some View {
         NavigationStack {
             ScrollView {
-                Text(info.content)
-                    .padding()
-                    .font(.element.footnote)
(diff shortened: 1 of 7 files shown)


ElementX/Sources/Services/Timeline/TimelineItemProxy.swift
```diff
@@ -142,3 +129,44 @@ extension TimelineItemContentKind {
...
+    
+    private static func prettyJsonFormattedString(from string: String?) -> String? {
+        guard let string,
+              let data = string.data(using: .utf8),
+              let jsonDictionary = try? JSONSerialization.jsonObject(with: data, options: .mutableContainers),
+              let jsonData = try? JSONSerialization.data(withJSONObject: jsonDictionary, options: [.prettyPrinted]) else {
+            return nil
+        }
+        
+        return String(data: jsonData, encoding: .ascii)
```
pixlwave, line 170: How come this uses `ascii` and not `utf8`?

ElementX/Sources/Screens/RoomScreen/View/TimelineItemDebugView.swift
```diff
@@ -17,41 +17,79 @@
...
     
     var body: some View {
         NavigationStack {
             ScrollView {
-                Text(info.content)
-                    .padding()
-                    .font(.element.footnote)
-                    .foregroundColor(.element.primaryContent)
-                    .frame(maxWidth: .infinity, alignment: .leading)
+                Divider()
```
pixlwave, line 29: From what I can tell, this is to separate the content from the navigation bar? In which case I would suggest replacing it with an additional modifier on the scroll view:
```
.toolbarBackground(.visible, for: .navigationBar)
```

ElementX/Sources/Screens/RoomScreen/View/TimelineItemDebugView.swift
```diff
@@ -17,41 +17,79 @@
...
-    struct DebugInfo: Identifiable {
-        let id = UUID()
-        let title: String
-        var content: String
-    }
-    
     @Environment(\.dismiss) private var dismiss
     
-    let info: DebugInfo
+    @State private var isModelExpanded = true
```
pixlwave, line 22: This is unused.

ElementX/Sources/Screens/RoomScreen/View/TimelineItemDebugView.swift
```diff
@@ -17,41 +17,79 @@
...
+                    }
+                }
+            }
+        }
+    }
+    
+    // MARK: - Private
+    
+    @ViewBuilder
+    private func disclosureGroup(title: String, text: String?, isExpanded: Binding<Bool>? = nil) -> some View {
```
pixlwave, line 58: Could we extract this into its own view struct?

ElementX/Sources/Screens/RoomScreen/View/TimelineItemDebugView.swift
```diff
@@ -17,41 +17,79 @@
...
+        if let text {
+            VStack(spacing: 0.0) {
+                Group {
+                    if let isExpanded {
```
pixlwave, line 62: `isExpanded` is never passed in to this view. Could be much simpler without it as the group wouldn't be necessary.

ElementX/Sources/Screens/RoomScreen/View/TimelineItemDebugView.swift
```diff
@@ -17,41 +17,79 @@
...
+                        DisclosureGroup(title, isExpanded: isExpanded) {
+                            groupContent(for: text)
+                        }
+                    } else {
+                        DisclosureGroup(title) {
+                            groupContent(for: text)
+                        }
                     }
                 }
+                .listRowInsets(.none)
```
pixlwave, line 72: I might be wrong, but I don't think this does anything outside of a `List`/`Form`

ElementX/Sources/Screens/RoomScreen/View/TimelineItemDebugView.swift
```diff
@@ -17,41 +17,79 @@
...
+                .padding()
+                
+                Divider()
             }
         }
     }
+    
+    private func groupContent(for text: String) -> some View {
+        VStack(alignment: .leading) {
+            Spacer()
```
pixlwave, line 83: Inside of a `ScrollView` I have a feeling that `Spacer()` basically does nothing.

ElementX/Sources/Screens/RoomScreen/View/TimelineItemDebugView.swift
```diff
@@ -71,6 +109,8 @@ struct TimelineItemDebugView_Previews: PreviewProvider {
     """
     
     static var previews: some View {
-        TimelineItemDebugView(info: .init(title: "Timeline item", content: smallContent))
+        TimelineItemDebugView(info: .init(model: smallContent,
+                                          originalJson: smallContent,
+                                          latestEditJson: smallContent))
```
pixlwave, line 114: I can't figure out why, but in the Xcode preview, I see neither of the json sections.

ElementX/Sources/Services/Timeline/TimelineItemProxy.swift
```diff
@@ -142,3 +129,44 @@ extension TimelineItemContentKind {
...
     }
 }
+
+struct TimelineItemDebugInfo: Identifiable, CustomStringConvertible {
+    let id = UUID()
+    let model: String
+    let originalJson: String?
+    let latestEditJson: String?
+    
+    init(model: String, originalJson: String?, latestEditJson: String?) {
```
pixlwave, line 139: Just doing a few of these, but I would say that in Swift, we should capitalise this ala `JSONEncoder` `JSONSerialization` etc etc.
```suggestion
    let originalJSON: String?
    let latestEditJSON: String?
    
    init(model: String, originalJSON: String?, latestEditJSON: String?) {
```

pixlwave (approved):
Looks great, will be nice to have this update 🙌

ElementX/Sources/Services/Timeline/TimelineItemProxy.swift
stefanceriu: Upsie 😬

ElementX/Sources/Screens/RoomScreen/View/TimelineItemDebugView.swift
stefanceriu: Well.. it should'be been, for always expading the model info from the start

ElementX/Sources/Screens/RoomScreen/View/TimelineItemDebugView.swift
stefanceriu: Sorry, this was me trying to remove some random padding fromt he disclosure group, forgot to remove it

ElementX/Sources/Screens/RoomScreen/View/TimelineItemDebugView.swift
stefanceriu: It does work, adds a bit of space between the disclosure group header and the divider

ElementX/Sources/Screens/RoomScreen/View/TimelineItemDebugView.swift
stefanceriu: I mean we could but does it really matter? 🤔

ElementX/Sources/Screens/RoomScreen/View/TimelineItemDebugView.swift
stefanceriu: I couldn't figure it out initially either but it just ocurred to me that it needs to be valid json

ElementX/Sources/Screens/RoomScreen/View/TimelineItemDebugView.swift
```diff
@@ -54,41 +59,44 @@ struct TimelineItemDebugView: View {
     
     // MARK: - Private
     
-    @ViewBuilder
-    private func disclosureGroup(title: String, text: String?, isExpanded: Binding<Bool>? = nil) -> some View {
-        if let text {
+    private struct TimelineItemInfoDisclosureGroup: View {
+        let title: String
+        let text: String
+        var isExpanded: Binding<Bool>?
```
pixlwave, line 65: Alrighty, so now with this tidied up, instead of passing in a binding to state in the parent view. We could have a `@State private var isExpanded: Bool` and an init that takes a `isExpandedInitially: Bool` parameter to set the state's initial value. That way we can get rid of the yucky looking `if let isExpanded`. Sound good?

ElementX/Sources/Screens/RoomScreen/View/TimelineItemDebugView.swift
pixlwave: Ohhhh, I stared at it for ages trying to understand why it wasn't working 😅

ElementX/Sources/Screens/RoomScreen/View/TimelineItemDebugView.swift
pixlwave: Looks much tidier now thanks 🙏

pixlwave (approved):
Nice! One last comment :)

ElementX/Sources/Screens/RoomScreen/View/TimelineItemDebugView.swift
stefanceriu: Haha, okay, can do 👍

PR #789, opened 2023-04-12
Move search users into a dedicated service

flescio:
Fix https://github.com/vector-im/element-x-ios/issues/627
Create a UsersProvider service which will provide searched users and suggested users.
Move the current search logic into this service.


ElementX/Sources/Other/Extensions/Publisher.swift
```diff
@@ -23,3 +24,16 @@ extension Publisher where Self.Failure == Never {
         }
     }
 }
+
+extension Publisher where Output == String, Failure == Never {
+    func delayTextOrImmediateClear() -> AnyPublisher<String, Never> {
```
alfogrillo, line 29:
```suggestion
    What about calling it like `searchQuery`?
```

ElementX/Sources/Services/Users/UsersProvider.swift
```diff
@@ -0,0 +1,75 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+
+class UsersProvider: UsersProviderProtocol {
+    var clientProxy: ClientProxyProtocol
```
alfogrillo, line 20:
```suggestion
    private let clientProxy: ClientProxyProtocol
```

ElementX/Sources/Services/Users/UsersProviderProtocol.swift
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+//
+
+import Foundation
+
+// sourcery: AutoMockable
+protocol UsersProviderProtocol {
+    func searchProfiles(with searchQuery: String) async -> [UserProfile]
+    func fetchSuggestions() -> [UserProfile]
+}
```
alfogrillo, line 23:
```suggestion
protocol UsersProviderProtocol {
    func searchProfiles(with searchQuery: String) async -> Result<[UserProfile], UsersProviderError>
    func fetchSuggestions() async -> Result<[UserProfile], UsersProviderError>
}
```

Let's try to have a more generic signature to fulfill possible future requirements:
eg:
- Suggestions not sync
- Error handling

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
```diff
@@ -102,51 +97,18 @@ class StartChatViewModel: StartChatViewModelType, StartChatViewModelProtocol {
...
-        
-        let filteredSearchResult = searchResults?.filter {
-            $0.userID != localProfile.userID
-        } ?? []
-
-        return [localProfile] + filteredSearchResult
     }
     
     private func fetchSuggestions() {
         guard ServiceLocator.shared.settings.startChatUserSuggestionsEnabled else {
```
alfogrillo, line 106: Possibly we can move it in the `UsersProvider`?

ElementX/Sources/UITests/UITestsAppCoordinator.swift
```diff
@@ -312,22 +312,28 @@ class MockScreen: Identifiable {
...
             return navigationStackCoordinator
         case .startChatSearchingNonexistentID:
             let navigationStackCoordinator = NavigationStackCoordinator()
             let clientProxy = MockClientProxy(userID: "@mock:client.com")
-            clientProxy.searchUsersResult = .success(.init(results: [.mockAlice], limited: true))
-            clientProxy.getProfileResult = .failure(.failedGettingUserProfile)
-            let coordinator = StartChatCoordinator(parameters: .init(userSession: MockUserSession(clientProxy: clientProxy, mediaProvider: MockMediaProvider()), navigationStackCoordinator: navigationStackCoordinator))
+            let usersProviderMock = UsersProviderMock()
+            usersProviderMock.fetchSuggestionsReturnValue = .success([.mockAlice, .mockBob, .mockCharlie])
+            usersProviderMock.searchProfilesWithReturnValue = .success([.init(userID: "@a:b.com"), .mockAlice, .mockBob])
```
alfogrillo, line 335:
```suggestion
            usersProviderMock.searchProfilesWithReturnValue = .success([.mockAlice, .mockBob])
```
Isn't this enough for this test?

pixlwave (commented):
Nicely done, its good to see a separate service being added.

ElementX/Sources/Services/Users/UsersProviderProtocol.swift
```diff
@@ -0,0 +1,23 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+
+// sourcery: AutoMockable
+protocol UsersProviderProtocol {
```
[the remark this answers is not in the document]
pixlwave: Ok, after a bit of a conversation with Stefan we're wondering if this should actually be something like `UserSearchResultsService`. Saying this is providing users doesn't really give you much context as to how/when it is providing those users. This may also give a nice small module to put other common things like for example `UserSearchResultsSection`

UnitTests/Sources/UsersProvider/UsersProviderTests.swift
```diff
@@ -0,0 +1,97 @@
...
+    
+    override func setUpWithError() throws {
+        clientProxy = .init(userID: "")
+        provider = UsersProvider(clientProxy: clientProxy)
+    }
+    
+    func testQueryShowingResults() async throws {
+        clientProxy.searchUsersResult = .success(.init(results: [UserProfile.mockAlice], limited: true))
+        
+        let results = await (try? search(query: "AAA").get()) ?? []
```
pixlwave, line 34: So I thought I hadn't seen this before. `.get()` was [deprecated](https://developer.apple.com/documentation/swift/task/get()-4i2gt) and replaced by `.value`. Would you mind having a search for all of these and replacing them with `.value` please?
```suggestion
        let results = await (try? search(query: "AAA").value) ?? []
```

ElementX/Sources/UITests/UITestsAppCoordinator.swift
```diff
@@ -312,22 +312,28 @@ class MockScreen: Identifiable {
...
+            usersProviderMock.searchProfilesWithReturnValue = .success([])
+            let coordinator = StartChatCoordinator(parameters: .init(userSession: MockUserSession(clientProxy: MockClientProxy(userID: "@mock:client.com"), mediaProvider: MockMediaProvider()), navigationStackCoordinator: navigationStackCoordinator, usersProvider: usersProviderMock))
             navigationStackCoordinator.setRootCoordinator(coordinator)
             return navigationStackCoordinator
         case .startChatWithSearchResults:
             let navigationStackCoordinator = NavigationStackCoordinator()
             let clientProxy = MockClientProxy(userID: "@mock:client.com")
-            clientProxy.searchUsersResult = .success(.init(results: [.mockAlice], limited: true))
-            let coordinator = StartChatCoordinator(parameters: .init(userSession: MockUserSession(clientProxy: clientProxy, mediaProvider: MockMediaProvider()), navigationStackCoordinator: navigationStackCoordinator))
+            let usersProviderMock = UsersProviderMock()
```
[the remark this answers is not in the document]
flescio: I created a more appealing test, showing results accordingly with the input, but this stuff is mocked, I don't get point on filtering an array just to show the desired results, because the mock already expose the final result and has no logic.

UnitTests/Sources/UsersProvider/UsersProviderTests.swift
flescio: I'm calling .get() on Result<> and should be the right one

pixlwave: Derp, sorry, my bad.

ElementX/Sources/Services/Users/UsersProvider.swift
```diff
@@ -16,46 +16,48 @@
...
+final class UsersProvider: UsersProviderProtocol {
     private let clientProxy: ClientProxyProtocol
     
     init(clientProxy: ClientProxyProtocol) {
         self.clientProxy = clientProxy
     }
     
-    func fetchSuggestions() -> [UserProfile] {
-        [.mockAlice, .mockBob, .mockCharlie]
+    func fetchSuggestions() async -> Result<[UserProfile], ClientProxyError> {
```
pixlwave: It would be cleaner for the service to have its own Error type and map the ClientProxyError to this as needed. That way anything calling methods on this service have a much more limited number of error cases to handle.

pixlwave (approved):
Great, looks good to me 😎

ElementX/Sources/Other/CancellableTask.swift
```diff
@@ -0,0 +1,35 @@
...
+            storedValue = newValue
+        }
+    }
+    
+    private var storedValue: Task<S, F>?
+    
+    init(_ value: Task<S, F>? = nil) {
+        storedValue = value
+    }
+}
```
alfogrillo, line 35:
```suggestion
@propertyWrapper
struct CancellableTask<S: Sendable, F: Error> {
    private var storedValue: Task<S, F>?
    
    init(_ value: Task<S, F>? = nil) {
        storedValue = value
    }
    
    var wrappedValue: Task<S, F>? {
        get {
            storedValue
        } set {
            storedValue?.cancel()
            storedValue = newValue
        }
    }
}
```
**Optional**
Sorted things like we usually do

ElementX/Sources/Services/Users/UserDiscoveryService.swift
```diff
@@ -0,0 +1,77 @@
...
+        return [localProfile] + filteredSearchResult
+    }
+    
+    private func profileIfPossible(with searchQuery: String) async -> Result<UserProfile?, ClientProxyError> {
+        guard searchQuery.isMatrixIdentifier else {
+            return .success(nil)
+        }
+        
+        return await clientProxy.profile(for: searchQuery).map { $0 }
+    }
```
alfogrillo, line 61: To me it looks better like this
```suggestion
    private func profileIfPossible(with searchQuery: String) async -> Result<UserProfile, ClientProxyError> {
        guard searchQuery.isMatrixIdentifier else {
            return .failure(.failedGettingUserProfile)
        }
        
        return await clientProxy.profile(for: searchQuery)
    }
```
wdyt?

alfogrillo (approved):
LGTM.
Left just a couple of minor comments

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
```diff
@@ -72,15 +72,17 @@ class StartChatViewModel: StartChatViewModelType, StartChatViewModelProtocol {
...
         switch type {
         case .failedRetrievingDirectRoom:
-            state.bindings.alertInfo = AlertInfo(id: type,
+            state.bindings.alertInfo = AlertInfo(id: .failedRetrievingDirectRoom,
                                                  title: L10n.commonError,
                                                  message: L10n.screenStartChatErrorStartingChat)
         case .failedCreatingRoom:
-            state.bindings.alertInfo = AlertInfo(id: type,
+            state.bindings.alertInfo = AlertInfo(id: .failedCreatingRoom,
                                                  title: L10n.commonError,
```
pixlwave, line 80: As these 2 are the same you could

```
case .failedRetrievingDirectRoom, .failedCreatingRoom:
    state.bindings.alertInfo = AlertInfo(id: .failedCreatingRoom,
                                         title: L10n.commonError,
                                         message: L10n.screenStartChatErrorStartingChat)
```
And then remove `StartChatErrorType.failedRetrievingDirectRoom` entirely as it isn't a user facing error

ElementX/Sources/Screens/StartChat/StartChatViewModel.swift
```diff
@@ -72,15 +72,17 @@ class StartChatViewModel: StartChatViewModelType, StartChatViewModelProtocol {
...
                                                  message: L10n.screenStartChatErrorStartingChat)
+        case .failedSearchingUsers:
+            state.bindings.alertInfo = AlertInfo(id: .failedSearchingUsers)
```
pixlwave, line 83: And then as this has no error message, this `StartChatErrorType.failedSearchingUsers` can become `StartChatErrorType.unknown`.

ElementX/Sources/Screens/InviteUsers/InviteUsersModels.swift
```diff
@@ -45,7 +45,7 @@ struct InviteUsersViewStateBindings {
     var searchQuery = ""
     
     /// Information describing the currently displayed alert.
-    var alertInfo: AlertInfo<ClientProxyError>?
+    var alertInfo: AlertInfo<UserDiscoveryErrorType>?
```
pixlwave, line 48: And then this would be `InviteUsersErrorType`.

pixlwave (approved):

ElementX/Sources/Screens/InviteUsers/InviteUsersViewModel.swift
```diff
@@ -65,11 +67,58 @@ class InviteUsersViewModel: InviteUsersViewModelType, InviteUsersViewModelProtoc
...
+    
+    @CancellableTask
+    private var searchTask: Task<Void, Never>?
+    
+    private func fetchUsers() {
+        guard searchQuery.count >= 3 else {
+            fetchSuggestions()
+            return
+        }
+        suggestionTask = nil
```
alfogrillo, line 88: It looks like `suggestionTask` and `searchTask` are mutually exclusive.
It would be better at this point to have just one `@CancellableTask` called like `fetchUsersTask` and use it for both right?

Issue #796, opened 2023-04-12
Room list can become unusable from nightly build 172


manuroe:
From this morning (23/04/12), the app (1.1.10 Build 175) started to be unusable for Matthew and I.

The room list does not load or refresh as crazy when scrolling. We needed to come back to build 171 to have it usable again.
On my side, the app stopped working properly for unknown reason. It was working well for 2 hours until I tried to reproduce the bug where messages miss in case of a gappy sync.

It might be related to a SS proxy update that happened the evening before including bug fixes like https://github.com/matrix-org/sliding-sync/issues/52.

manuroe:
I wonder if it could be due to a cache issue. Running the develop code with the same account on a different app id (io.element.elementx) does not have the issue.

manuroe:
New logs show that we have inconsistency on the room count. It is fine if the view is static. The state stays in `fullyLoaded` but when we scroll, the state turns into `notLoaded`. The room count inconsistency becomes an issue.

The issue seems to be around the diff mechanism and its output. My actual room count is 849. I do not know why there is an additional room.


```
2023-04-13 18:04:34.152736+0200 ElementX[45402:3284985] HomeScreenViewModel.init():96 roomSummaryProviderState: notLoaded. rooms.count: 850 / totalCount: 849
2023-04-13 18:04:34.661697+0200 ElementX[45402:3284985] HomeScreenViewModel.init():96 roomSummaryProviderState: notLoaded. rooms.count: 850 / totalCount: 849
2023-04-13 18:04:36.626117+0200 ElementX[45402:3284985] HomeScreenViewModel.init():96 roomSummaryProviderState: notLoaded. rooms.count: 850 / totalCount: 849
2023-04-13 18:04:37.144815+0200 ElementX[45402:3284985] HomeScreenViewModel.init():96 roomSummaryProviderState: notLoaded. rooms.count: 850 / totalCount: 849
2023-04-13 18:04:40.595978+0200 ElementX[45402:3284985] HomeScreenViewModel.init():96 roomSummaryProviderState: notLoaded. rooms.count: 850 / totalCount: 849

manuroe:
> The issue seems to be around the diff mechanism and its output

I was wrong. I got disturbed by the fact we have 2 instances of `RoomSummaryProvider`,`VisibleRooms` and `AllRooms`, getting updates at the same time.

[...]

PR #800, opened 2023-04-13
Display the room list even if the room count is not exact

manuroe:
This is a workaround for #796 to cover the case where the `rooms.count` at the output of the diffing mechanism is not correct as described [here](https://github.com/vector-im/element-x-ios/issues/796#issuecomment-1507288114).

The pre-commit hook forced me to commit a change on another file ¯\_(ツ)_/¯


pixlwave (approved):
LGTM 👍

The SwiftFormat step is nothing to worry about - the tools are updated naturally by brew so we will see constant little style improvements like this :)

PR #801, opened 2023-04-14
Notifications are now handled when the app is in a killed state

Velin92:
The issue was that we were not registering the NotificationManager as the UNUserNotificationCenter soon enough, so we couldn't manage notifications that woke the app from a killed state.

This small refactor alongside some improvements to store the route states should allow this work now.

Also added the SwiftURLRouting package, but is a WIP we still do nothing with it.


ElementX/Sources/Other/Extensions/UNNotificationContent.swift
```diff
@@ -15,23 +15,21 @@
...
 import Foundation
+import UserNotifications
 
-class MockNotificationManager: NotificationManagerProtocol {
-    // MARK: NotificationManagerProtocol
-
-    var isAvailable: Bool {
-        false
+extension UNNotificationContent {
+    @objc var receiverId: String? {
```
pixlwave, line 21:
```suggestion
    @objc var receiverID: String? {
```

ElementX/Sources/Other/Consumable.swift
```diff
@@ -0,0 +1,37 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+
+@propertyWrapper struct Consumable<Value> {
```
pixlwave, line 19: Woah, move-only via a property wrapper, I love it 🤩

ElementX/Sources/Services/Notification/Manager/NotificationManager.swift
```diff
@@ -46,6 +41,9 @@ class NotificationManager: NSObject, NotificationManagerProtocol {
                                                    options: [])
         notificationCenter.setNotificationCategories([replyCategory])
         notificationCenter.delegate = self
+    }
+
+    func requestAuth() {
         Task {
```
pixlwave, line 47: Not particularly related to this PR, but should this method have a guard on `notificationCentre.notificationSettings().authorizationStatus == .notDetermined` before calling request authorisation? Not sure if that would break anything if `authorizationStatusUpdated` wasn't called though, so better to double check.

ElementX/Sources/Services/Notification/Manager/NotificationManagerProtocol.swift
```diff
@@ -31,12 +31,14 @@ protocol NotificationManagerDelegate: AnyObject {
...
-    var isAvailable: Bool { get }
+// sourcery: AutoMockable
+protocol NotificationManagerProtocol: AnyObject {
     var delegate: NotificationManagerDelegate? { get set }
 
     func start()
     func register(with deviceToken: Data) async -> Bool
     func registrationFailed(with error: Error)
     func showLocalNotification(with title: String, subtitle: String?) async
+    func setClientProxy(_ clientProxy: ClientProxyProtocol?)
```
pixlwave, line 42: I'm wondering if it would be sensible to update this to be multi-account aware now or if that would be a distraction. Feels like ultimately we should give it a publisher of user sessions but this might be quite a lot of work right now.

pixlwave (approved):

ElementX.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -160,15 +169,33 @@
...
+      "state" : {
+        "revision" : "c6e2241daa46e5c6e5027a93b161bca6ba692bcc",
+        "version" : "0.12.0"
+      }
+    },
     {
       "identity" : "swift-snapshot-testing",
       "kind" : "remoteSourceControl",
-      "location" : "https://github.com/pointfreeco/swift-snapshot-testing",
+      "location" : "https://github.com/pointfreeco/swift-snapshot-testing.git",
```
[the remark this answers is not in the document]
Velin92: no idea but I hate it

ElementX/Sources/Services/Notification/Manager/NotificationManagerProtocol.swift
```diff
@@ -31,12 +31,14 @@ protocol NotificationManagerDelegate: AnyObject {
...
+    func requestAuth()
```
pixlwave, line 43:
```suggestion
    func requestAuthorization()
```

ElementX/Sources/Other/Consumable.swift
Velin92: use it once, and is destroyed yeah.

ElementX/Sources/Services/Notification/Manager/NotificationManagerProtocol.swift
Velin92: like every time the user session changes we should call this? For now this is automatically happening through the didSet observer in the AppCoordinator that is on the user session. But maybe worth checking this again in a Multi Account PR in the future

ElementX/Sources/Services/Notification/Manager/NotificationManager.swift
Velin92: Uhm I am not sure but since the requestAuthorization state could have been changed externally before launching the app or the user session, I feel that we should run this code even if the state != .notDetermined.

PR #803, opened 2023-04-14
Use iOS localization handling for strings.

pixlwave:
Previously we had our own `calculatePreferredLanguages` method based upon `Locale.preferredLanguages` to decide suitable languages for strings and then before translating we were doing additional work to manage locale regions if necessary.

Turns out that `Bundle` has a method called `preferredLocalizations` which does this for us and makes sure to only provide a single language (plus any regional variations) so you don't end up with the app showing strings in 3 or more languages.

I've removed `Bundle.elementFallbackLanguage` in favour of the `developmentLocalization` property, but left in an override so that we can still run unit tests.

There's a run of UI tests going on [here](https://github.com/vector-im/element-x-ios/actions/runs/4702454042), to make sure this works there too


ElementX/Sources/Generated/Strings.swift
```diff
@@ -800,31 +800,23 @@ public enum L10n {
...
   private static func trIn(_ language: String, _ table: String, _ key: String, _ args: CVarArg...) -> String? {
-    guard let bundle = Bundle(for: BundleToken.self).lprojBundle(for: language) else {
-      // no translations for the desired language
-      return nil
-    }
+    guard let bundle = Bundle.lprojBundle(for: language) else { return nil }
     let format = NSLocalizedString(key, tableName: table, bundle: bundle, comment: "")
-    return String(format: format, locale: Locale(identifier: language), arguments: args)
+    let translation = String(format: format, locale: Locale(identifier: language), arguments: args)
+    guard translation != key else { return nil }
```
pixlwave, line 817: This check was missing before which left a possible path where the key would be shown if it was missing from the translations instead of falling back to the next potential language or the development language.

Velin92: what if the key and the translations match? We would discard that translation string

ElementX/Sources/Services/BugReport/BugReportService.swift
```diff
@@ -160,9 +160,9 @@ class BugReportService: NSObject, BugReportServiceProtocol {
             MultipartFormData(key: "version", type: .text(value: InfoPlistReader.main.bundleShortVersionString)),
             MultipartFormData(key: "build", type: .text(value: InfoPlistReader.main.bundleVersion)),
             MultipartFormData(key: "os", type: .text(value: os)),
-            MultipartFormData(key: "resolved_language", type: .text(value: Bundle.preferredLanguages[0])),
-            MultipartFormData(key: "user_language", type: .text(value: Bundle.elementLanguage ?? "null")),
-            MultipartFormData(key: "fallback_language", type: .text(value: Bundle.elementFallbackLanguage ?? "null")),
+            MultipartFormData(key: "resolved_language", type: .text(value: Bundle.app.preferredLocalizations[0])),
+            MultipartFormData(key: "user_language", type: .text(value: Locale.preferredLanguages.first ?? "null")),
```
Velin92, line 164: what about logging the whole array?

Velin92 (approved):
LGTM just one question and one suggestion

ElementX/Sources/Generated/Strings.swift
pixlwave: Given they're all prefixed by `action_`, `common_`, `screen_` etc, the only potential place this would fail is `notification = Notification` and that wouldn't happen due to the casing. We shouldn't have a key that matches the string.

Nicolas just hit an issue where he downloaded the `en` strings with the tool and in his simulator in French all the strings he just added show up as the keys.

ElementX/Sources/Services/BugReport/BugReportService.swift
pixlwave: Honestly I wasn't sure how this would come through, so didn't but that's a good point. Let me try it.

ElementX/Sources/Generated/Strings.swift
Velin92: I think it's Notification with the capital letter right now for that specific one... maybe we should then change it or give it a more appropriate name, like... default_push_notification_title

pixlwave: Lets discuss outside of this PR.

PR #805, opened 2023-04-17
Navigations can now have their animations controlled with a parameter

Velin92:
In this PR, a refactor has been done on the navigation coordinators, so that they can now have all their navigation functions be fine tuned to show an animation or not.

This required also to turn the compact layout into a computed binding.

### DIFF
--- a/ElementX/Sources/Application/Navigation/NavigationCoordinators.swift
+++ b/ElementX/Sources/Application/Navigation/NavigationCoordinators.swift
@@ -37,8 +37,8 @@ class NavigationSplitCoordinator: CoordinatorProtocol, ObservableObject, CustomS
                 logPresentationChange("Set sidebar", sidebarModule)
                 sidebarModule.coordinator?.start()
                 if let observableCoordinator = sidebarModule.coordinator as? NavigationStackCoordinator {
-                    sidebarWillChangeCancellable = observableCoordinator.$stackModules.sink { _ in
-                        self.objectWillChange.send()
+                    sidebarWillChangeCancellable = observableCoordinator.$stackModules.sink { [weak self] _ in
+                        self?.objectWillChange.send()
                     }
                 }
             }
@@ -63,8 +63,8 @@ class NavigationSplitCoordinator: CoordinatorProtocol, ObservableObject, CustomS
                 logPresentationChange("Set detail", detailModule)
                 detailModule.coordinator?.start()
                 if let observableCoordinator = detailModule.coordinator as? NavigationStackCoordinator {
-                    detailWillChangeCancellable = observableCoordinator.$stackModules.sink { _ in
-                        self.objectWillChange.send()
+                    detailWillChangeCancellable = observableCoordinator.$stackModules.sink { [weak self] _ in
+                        self?.objectWillChange.send()
                     }
                 }
             }
--- a/ElementX/Sources/Application/Navigation/NavigationCoordinators.swift
+++ b/ElementX/Sources/Application/Navigation/NavigationCoordinators.swift
@@ -24,20 +24,21 @@ class NavigationSplitCoordinator: CoordinatorProtocol, ObservableObject, CustomS
     
     private var cancellables = Set<AnyCancellable>()
 
-    var sidebarWillChangeCancellable: AnyCancellable?
+    var sidebarStackModuleCancellable: AnyCancellable?
 
     @Published fileprivate var sidebarModule: NavigationModule? {
         didSet {
             if let oldValue {
                 logPresentationChange("Remove sidebar", oldValue)
                 oldValue.tearDown()
+                sidebarStackModuleCancellable = nil
             }
             
             if let sidebarModule {
                 logPresentationChange("Set sidebar", sidebarModule)
                 sidebarModule.coordinator?.start()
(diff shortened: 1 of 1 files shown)


ElementX/Sources/Application/Navigation/NavigationCoordinators.swift
```diff
@@ -23,7 +23,9 @@ class NavigationSplitCoordinator: CoordinatorProtocol, ObservableObject, CustomS
     fileprivate let placeholderModule: NavigationModule
     
     private var cancellables = Set<AnyCancellable>()
-    
+
+    var sidebarWillChangeCancellable: AnyCancellable?
```
pixlwave, line 27: Optional: maybe slightly clearer to call this
```suggestion
    var sidebarStackModulesCancellable: AnyCancellable?
```

ElementX/Sources/Application/Navigation/NavigationCoordinators.swift
```diff
@@ -34,16 +36,21 @@ class NavigationSplitCoordinator: CoordinatorProtocol, ObservableObject, CustomS
             if let sidebarModule {
                 logPresentationChange("Set sidebar", sidebarModule)
                 sidebarModule.coordinator?.start()
+                if let observableCoordinator = sidebarModule.coordinator as? NavigationStackCoordinator {
+                    sidebarWillChangeCancellable = observableCoordinator.$stackModules.sink { [weak self] _ in
```
pixlwave, line 40: Looking at this with fresh eyes, I wonder if would be sensible to set these new cancellables to `nil`  in the `oldValue` block? 

E.g. if we replace the sidebar or detail module with something that isn't a navigation stack, then it won't get cancelled (probably wouldn't cause any harm leaving it, but seems worthwhile nonetheless).

ElementX/Sources/Application/Navigation/NavigationCoordinators.swift
```diff
@@ -34,16 +36,21 @@ class NavigationSplitCoordinator: CoordinatorProtocol, ObservableObject, CustomS
...
-            updateCompactLayoutComponents()
         }
     }
     
     /// The currently displayed sidebar coordinator
     var sidebarCoordinator: (any CoordinatorProtocol)? {
         sidebarModule?.coordinator
     }
+
+    var detailWillChangeCancellable: AnyCancellable?
```
pixlwave, line 53:
```suggestion
    var detailStackModulesCancellable: AnyCancellable?
```

ElementX/Sources/Application/Navigation/NavigationCoordinators.swift
```diff
@@ -141,15 +204,20 @@ class NavigationSplitCoordinator: CoordinatorProtocol, ObservableObject, CustomS
...
+            sidebarModule = NavigationModule(coordinator, dismissalCallback: dismissalCallback)
+        }
     }
     
     /// Set the coordinator to be used on the split's right pannel
     /// - Parameters:
     ///   - coordinator: the detail coordinator
     ///   - dismissalCallback: called when this particular detail coordinator has removed/replaced
-    func setDetailCoordinator(_ coordinator: (any CoordinatorProtocol)?, dismissalCallback: (() -> Void)? = nil) {
+    func setDetailCoordinator(_ coordinator: (any CoordinatorProtocol)?, animated: Bool = true, dismissalCallback: (() -> Void)? = nil) {
```
pixlwave, line 220: This one is missing the doc comment.

pixlwave, line 220: This one is missing the doc comment.

ElementX/Sources/Application/Navigation/NavigationCoordinators.swift
```diff
@@ -328,28 +352,6 @@ class NavigationSplitCoordinator: CoordinatorProtocol, ObservableObject, CustomS
...
-            guard navigationStackCoordinator.stackModules != stackModules else { return }
-            DispatchQueue.main.async { self?.updateCompactLayoutComponents() }
-        }
-        .store(in: &cancellables)
-        
-        navigationStackCoordinator.$sheetModule.sink { [weak self] sheetModule in
-            guard navigationStackCoordinator.sheetModule != sheetModule else { return }
-            DispatchQueue.main.async { self?.updateCompactLayoutComponents() }
-        }
-        .store(in: &cancellables)
```
pixlwave, line 351: Ok looking at this it also observed the `root` and `sheet` modules. From the comments it seems sheets was pointless as we have to present them on the split coordinator (although in the future perhaps we could fix that with another `Binding`) but I think we probably also need to include the root in our observation:
```
Publishers.CombineLatest2(coordinator.$rootModule, 
                          coordinator.$stackModules)
    .sink {
    …
```
Note: I don't think the unit tests will capture this as they use the `get` which will recompute and be correct whereas SwiftUI will only recompute after receiving a `willChange`.

pixlwave (approved):
LGTM, but I think we need to handle the `rootModule` too.

ElementX/Sources/Application/Navigation/NavigationCoordinators.swift
Velin92: Why using combine latest? We could just set the observation as we did for the stack and the sidebar modules?

PR #806, opened 2023-04-17
Accept and decline invites

alfogrillo:
Fixes https://github.com/vector-im/element-x-ios/issues/621

**Result**
demo


UITests/Sources/__Snapshots__/Application/pseudo-iPhone-14.invites-1.png
alfogrillo, line 1: Idk what's happening here (didn't investigate yet):

- Some keys aren't resolved
- Some text is in Italian

Any idea?

pixlwave: - First one should be fixed on `develop` now 🤞
- Second one I think we need to make sure the device's language list is English and English alone so it matches the default config:
Simulator Screenshot - iPhone 14 Pro - 2023-04-18 at 11 34 19

alfogrillo: > First one should be fixed on develop now 

It worked! 🎉

> Second one I think we need to make sure the device's language list is English and English alone so it matches the default config:

I expected the language in the test plan to override the simulator one. But it looks not working for the pseudo language I guess. 🤔
The workaround you suggested worked anyway. 👍🏻

ElementX/Sources/Services/UserSession/UserSessionFlowCoordinatorStateMachine.swift
```diff
@@ -39,7 +39,7 @@ class UserSessionFlowCoordinatorStateMachine {
         case startChatScreen(selectedRoomId: String?)
         
         /// Showing invites list screen
-        case invitesScreen(selectedRoomId: String?)
+        case invitesScreen(selectedRoomId: String?, shownRoomId: String?)
```
stefanceriu, line 42: What's the difference between the `selectedRoomId` and the `shownRoomId`? 🤔

stefanceriu (changes requested):

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -129,7 +129,7 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
         
         let details = RoomSummaryDetails(id: room.roomId(),
                                          name: room.name() ?? room.roomId(),
-                                         isDirect: room.isDm() ?? false,
+                                         isDirect: fullRoom?.isDirect() ?? room.isDm() ?? false,
```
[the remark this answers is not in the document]
alfogrillo: We have now a custom logic in the sdk for the `isDirect` flag for rooms in the invite state.
This logic is available on a `Room`. More context [here](https://github.com/matrix-org/matrix-rust-sdk/pull/1775).

ElementX/Sources/Services/UserSession/UserSessionFlowCoordinator.swift
```diff
@@ -200,12 +205,18 @@ class UserSessionFlowCoordinator: CoordinatorProtocol {
...
                 
-                // Move the state machine to no room selected if the room currently being dimissed
+                // Move the state machine to no room selected if the room currently being dismissed
                 // is the same as the one selected in the state machine.
                 // This generally happens when popping the room screen while in a compact layout
-                if case let .roomList(selectedRoomId) = self.stateMachine.state, selectedRoomId == roomIdentifier {
+                switch self.stateMachine.state {
+                case
+                    let .roomList(selectedRoomId) where selectedRoomId == roomIdentifier,
+                    let .invitesScreen(_, selectedRoomId) where selectedRoomId == roomIdentifier:
```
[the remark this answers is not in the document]
alfogrillo: Not really: as described [here](https://github.com/vector-im/element-x-ios/pull/806#discussion_r1171151338) the actual timeline in case of invites is the second component, the first one is there just to fix the above mentioned technical problem.
If this is confusing I can add two separate cases (doing the same thing though).

ElementX/Sources/Services/UserSession/UserSessionFlowCoordinatorStateMachine.swift
stefanceriu: Talked about this offline: we will try to simplify the logic by replacing the currently selected room with the one that needs to be opened after accepting an invititation.

alfogrillo: Done here -> https://github.com/vector-im/element-x-ios/pull/806/commits/a166bc67fd7713f344540595d8fa138ed468a046

stefanceriu (approved):
Great stuff 👌

PR #807, opened 2023-04-17
Remove CurrentValueSubject as protocol requirement

alfogrillo:
This pr introduces the `CurrentValuePublisher`.  
It's purpose is to hide the `send` method for clients that need just to subscribe and/or the current value of a `CurrentValueSubject`.

### DIFF
--- a/ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
+++ b/ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
@@ -25,13 +25,25 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
     
     private var cancellables = Set<AnyCancellable>()
     
-    let roomListPublisher = CurrentValueSubject<[RoomSummary], Never>([])
-    let statePublisher = CurrentValueSubject<RoomSummaryProviderState, Never>(.notLoaded)
-    let countPublisher = CurrentValueSubject<UInt, Never>(0)
+    private let roomListSubject = CurrentValueSubject<[RoomSummary], Never>([])
+    private let stateSubject = CurrentValueSubject<RoomSummaryProviderState, Never>(.notLoaded)
+    private let countSubject = CurrentValueSubject<UInt, Never>(0)
+    
+    var roomListPublisher: CurrentValuePublisher<[RoomSummary], Never> {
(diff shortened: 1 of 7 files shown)


ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -25,13 +25,25 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
...
-    let roomListPublisher = CurrentValueSubject<[RoomSummary], Never>([])
-    let statePublisher = CurrentValueSubject<RoomSummaryProviderState, Never>(.notLoaded)
-    let countPublisher = CurrentValueSubject<UInt, Never>(0)
+    private let roomListSubject = CurrentValueSubject<[RoomSummary], Never>([])
+    private let stateSubject = CurrentValueSubject<RoomSummaryProviderState, Never>(.notLoaded)
+    private let countSubject = CurrentValueSubject<UInt, Never>(0)
+    
+    var roomListPublisher: CurrentValuePublisher<[RoomSummary], Never> {
+        roomListSubject.asCurrentValuePublisher()
+    }
```
pixlwave, line 34: Ok so dumb question here… If `provider.roomListPublisher.value` is called often, do you think its bad that we create a new publisher each time it gets called?

I guess that's the long way of saying "Should all these new publishers be stored properties?"

pixlwave (commented):
Ok, in theory I like this change a lot. One thought about the use of the publisher inline though.

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
alfogrillo: I don't see this as a real problem. 
`CurrentValuePublisher` is a lightweight value type that wraps around a ref type (`CurrentValueSubject`).
I guess we can think of it like a [AnyPublisher](https://developer.apple.com/documentation/combine/anypublisher/) which wraps a publisher with a value type as well.

pixlwave: Ok, fair point.

pixlwave (approved):

PR #810, opened 2023-04-18
Update rust sdk to 1.0.54-alpha

alfogrillo:


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -319,7 +319,6 @@ class ClientProxy: ClientProxyProtocol {
             let slidingSync = try slidingSyncBuilder
                 .addList(v: visibleRoomsSlidingSyncView)
                 .withCommonExtensions()
-                .coldCache(name: "ElementX")
```
stefanceriu, line 322: Why are we removing the cold cache?

stefanceriu: Same comment as in the other pr `I think it was just renamed to storage_key instead, please don't remove it entirely. The "cold cache" is responsible for providing a room list while in offline mode or before the initial sync comes down the wire`

stefanceriu: Nice, thank you 👍

PR #811, opened 2023-04-19
Fixes #785 - Speed up restoring the room list from the cold cache

### LINKED ISSUE #785: Start from cold is slower
The display of the room list in the nightlies is slower since #775 landed.

stefanceriu:
Following changes on the rust side we were forced to move the sliding sync view building to before invoking the sliding sync builder so that observers would be in place. This had the side effect of returning an empy current room list when creating the room summary providers (as part of the `buildAndConfigure..SlidingSyncView`)

This PR extracts the creation of the roomSummaryProviders into a separate method that we manually invoke after the sliding sync setup process is complete


Velin92 (approved):
LGTM

PR #812, opened 2023-04-19
Room subscription limits

stefanceriu:
Given the following scenario:
* a room moves outside of the visible rooms list
* it continues to receive messages
* the room is then opened through the search (which doesn't change the SS range)

then that room will be missing messages that were sent after it exited the visible list.

The fix for that is to actually use a timeline limit when subscribing to it, same limit we use for the normal room list when not scrolling.


manuroe (approved):
👏

pixlwave (approved):
Nice!

PR #814, opened 2023-04-20
Refactor UserSetting property wrapper to support Codable

alfogrillo:
This PR replaces the property wrapper `UserSetting` and `UserSettingRawRepresentable` with a single property wrapper called `UserPreference`.
The new property wrapper also allows to store `Codable` values in any provided storage.


Velin92:
Could we add some tests for it? both for the volatile case, the storage case, for each one of this a test with PlistValues that need encoding to data and those whic don't.
We could use just a user defaults test storage

alfogrillo:
> Amazing!
> Could we add some tests for it? both for the volatile case, the storage case, for each one of this a test with PlistValues that need encoding to data and those whic don't.
> We could use just a user defaults test storage

On it already ;)

Velin92 (approved):
LGTM

pixlwave (approved):
LGTM 👍

PR #816, opened 2023-04-20
Fix: forget the user’s consents for analytics on logout

nimau:
This PR reset the user's consents for analytics on logout (as requested in #106)

### DIFF
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -295,7 +295,8 @@ class AppCoordinator: AppCoordinatorProtocol {
             tearDownUserSession()
             
             // reset analytics
-            ServiceLocator.shared.analytics.reset()
+            ServiceLocator.shared.analytics.optOut()
+            AppSettings.forgetAnalyticsConsents()
             
             stateMachine.processEvent(.completedSigningOut(isSoft: isSoft))
         }
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -296,7 +296,7 @@ class AppCoordinator: AppCoordinatorProtocol {
             
             // reset analytics
             ServiceLocator.shared.analytics.optOut()
-            AppSettings.forgetAnalyticsConsents()
+            ServiceLocator.shared.analytics.resetConsentState()
             
             stateMachine.processEvent(.completedSigningOut(isSoft: isSoft))
         }
--- a/ElementX/Sources/Application/AppSettings.swift
+++ b/ElementX/Sources/Application/AppSettings.swift
@@ -134,6 +134,12 @@ final class AppSettings: ObservableObject {
     /// `true` when the user has opted in to send analytics.
     @UserPreference(key: UserDefaultsKeys.enableAnalytics, defaultValue: false, storageType: .userDefaults(store))
     var enableAnalytics
+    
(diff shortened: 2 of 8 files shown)


ElementX/Sources/Application/AppSettings.swift
```diff
@@ -128,6 +128,12 @@ final class AppSettings: ObservableObject {
     /// `true` when the user has opted in to send analytics.
     @UserSetting(key: UserDefaultsKeys.enableAnalytics.rawValue, defaultValue: false, persistIn: store)
     var enableAnalytics
+    
+    /// Reset user consents for analytics
+    static func forgetAnalyticsConsents() {
```
stefanceriu, line 133: @pixlwave and I were chatting about this and we believe we have a better solution:
- change `enableAnalytics` to be an enum with 3 values stored as a RawRepresentable with a default value of `.unknown` or similar
- remove the `forgetAnalyticsConsents` and just use `ServiceLocator.shared.settings.enableAnalytics  = .unknown` when you want to reset it

stefanceriu (changes requested):

ElementX/Sources/Services/Analytics/AnalyticsConsentState.swift
```diff
@@ -0,0 +1,23 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+
+enum AnalyticsConsentState: String {
+    case optOut
+    case optIn
```
stefanceriu, line 21: Should these be called `optedIn` and `optedOut` maybe?

ElementX/Sources/Application/AppSettings.swift
nimau: Hi @stefanceriu, I've updated the PR to reflect this change.
Also I took the liberty of renaming the UserDefaultKey from `enableAnalytics` to `analyticsConsentState` to better reflect the purpose of this key.

ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -295,7 +295,8 @@ class AppCoordinator: AppCoordinatorProtocol {
             tearDownUserSession()
             
             // reset analytics
-            ServiceLocator.shared.analytics.reset()
+            ServiceLocator.shared.analytics.optOut()
+            ServiceLocator.shared.analytics.resetConsentState()
```
stefanceriu, line 299: At this point wouldn't it be better to keep the `reset` method that would `optOut` + `resetConsent` internally?

stefanceriu (approved):
Looks great 👍

ElementX/Sources/Services/Analytics/AnalyticsConsentState.swift
nimau: Done.

ElementX/Sources/Application/AppCoordinator.swift
nimau: There is a `reset` method on `Analytics` but it is for resetting the analytics client and the bug report service. Do you mean to move these two calls inside a new private `reset` method inside the `AppCoordinator`?

stefanceriu: No no, sorry, I didn't realise there's another reset method in there

PR #817, opened 2023-04-20
Fixes #641 - Room previews failing to load because of incorrect slidi…

### LINKED ISSUE #641: Scrolling around the room list can end up with room previews failing to load
### Steps to reproduce

1. Scroll back in the roomlist
2. Notice that all the room previews are blank
3. Wait for them to trickle in
4. they never trickle in

#### What did you expect?

reliable room list

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

1.0.21 (137)

### Homeserver

_No response_

### Will you send logs?

Yes

stefanceriu:
…ng sync view ranges

Using the search functionality in conjuncture with the existing appearance call based visible range computations proved to be unreliable and resulted in not only incorrect ranges but clunky code.

This PR rewrites all of that and makes the range computation now rely on the underlying scroll view, its content size and offset.

It relies heavily on the deduplication and throttling happening view model side and while the computations might not be perfect they're good enough considering the paddings we're using

### DIFF
--- a/ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
+++ b/ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
@@ -23,17 +23,8 @@ struct HomeScreen: View {
     
     @ObservedObject var context: HomeScreenViewModel.Context
     
-    @State private var isViewVisible = false
-    
     @State private var scrollViewAdapter = ScrollViewAdapter()
     @State private var showingLogoutConfirmation = false
-    @State private var visibleItemIdentifiers = Set<String>() {
-        didSet {
-            if isViewVisible {
-                updateVisibleRange()
-            }
-        }
-    }
     
     var body: some View {
         ScrollView {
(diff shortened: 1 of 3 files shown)


ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
```diff
@@ -69,38 +60,37 @@ struct HomeScreen: View {
...
+        })
+        .onChange(of: context.searchQuery, perform: { searchQuery in
+            if searchQuery.isEmpty {
+                // Allow the view to update after changing the query
+                DispatchQueue.main.async {
+                    updateVisibleRange()
+                }
+            }
+        })
+        .onChange(of: context.viewState.visibleRooms, perform: { _ in
```
pixlwave, line 88: These 3 could use trailing closures.

pixlwave (approved):
Looks much more sane to me 👍

ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
stefanceriu: Ups, I keep missing these 🙈

PR #820, opened 2023-04-21
Media uploading preprocessor tests

stefanceriu:
This PR add some simple tests for checking the results of media uploading preprocessing for various media types. 
It also prevents the app coordinator from even starting in unit tests


UnitTests/Resources/test_image.png
pixlwave, line 1: What's going on with this file? Looks like it was edited to an empty image?

UnitTests/Sources/MediaUploadingPreprocessorTests.swift
```diff
@@ -0,0 +1,186 @@
...
+            return
+        }
+        
+        guard case let .success(result) = await mediaUploadingPreprocessor.processMedia(at: url),
+              case let .video(videoURL, thumbnailURL, videoInfo) = result else {
+            XCTFail("Failed processing asset")
+            return
+        }
+        
+        XCTAssertEqual(videoURL.pathExtension, "mp4")
```
pixlwave, line 55: Would it be nice to check that it is still called `test_video.mp4` too? Possibly a good check for all of these tests that the name remains unchanged - could also do a file with a space in the name for example.

UnitTests/Sources/MediaUploadingPreprocessorTests.swift
```diff
@@ -0,0 +1,186 @@
...
+        
+        // Check that the thumbnail is generated correctly
+        guard let thumbnailData = try? Data(contentsOf: thumbnailURL),
+              let thumbnail = UIImage(data: thumbnailData) else {
```
pixlwave, line 59: Totally unimportant: Isn't there a `UIImage(contentsOf:)`?

.gitattributes
```diff
@@ -1 +1,2 @@
 UITests/Sources/__Snapshots__/** filter=lfs diff=lfs merge=lfs -text
+UnitTests/Resources/** filter=lfs diff=lfs merge=lfs -text
```
pixlwave, line 2: I wonder if these should go inside another subdirectory e.g. `Media` so that we could add more basic resources such as JSON files for test fixtures in the future.

pixlwave (approved):
Looks great 👌

UnitTests/Resources/test_image.png
stefanceriu: The image is a weird transparent version of our logo 😅 Github is also acting up, it looks fine if you `View file` and drag it around

PR #824, opened 2023-04-24
Update template screen naming.

pixlwave:
This PR updates the template and the script so that

`./createScreen Home` creates

```
Screens/HomeScreen/View/HomeScreen.swift
Screens/HomeScreen/HomeScreenCoordinator.swift
Screens/HomeScreen/HomeScreenModels.swift
Screens/HomeScreen/HomeScreenViewModel.swift
```

PR #826, opened 2023-04-24
Rename screens

stefanceriu:
Rename a lot of the screens following https://github.com/vector-im/element-x-ios/pull/824


ElementX/Sources/Screens/RoomMemberListScreen/RoomMembersListScreenCoordinator.swift
```diff
@@ -16,26 +16,26 @@
...
     let members: [RoomMemberProxyProtocol]
 }
 
-enum RoomMembersListCoordinatorAction { }
+enum RoomMembersListScreenCoordinatorAction { }
 
-final class RoomMembersListCoordinator: CoordinatorProtocol {
-    private let parameters: RoomMembersListCoordinatorParameters
-    private var viewModel: RoomMembersListViewModelProtocol
+final class RoomMembersListScreenCoordinator: CoordinatorProtocol {
```
pixlwave, line 27: I'm not a huge fan of this one given `List` is a view component. `ListScreen` sounds like writing `ButtonScreen` would be a thing too. That said I can't think of anything better than simply `RoomMembersScreenCoordinator` which might not be great either.

pixlwave (approved):
Nice work! I bet that was tedious.

ElementX/Sources/Screens/RoomMemberListScreen/RoomMembersListScreenCoordinator.swift
stefanceriu: That's the exact same thought process I went through. MemberListScreen is fine, really, it is 😄

PR #827, opened 2023-04-24
Fixes #641 - Prevent large ranges before the view renders the room list

### LINKED ISSUE #641: Scrolling around the room list can end up with room previews failing to load

stefanceriu:
On first load, when the room list is not rendered yet, the range computation mechanism generates ranges equal to the total number of rooms. This prevents that from happening


pixlwave (approved):

pixlwave:
I can't wait to get the rooms/visible rooms lists merged in the SDK 😄

Issue #821, opened 2023-04-22
Room not coming up when filtering roomlist despite being visible when roomlist unfiltered


ara4n:
### Steps to reproduce

https://github.com/vector-im/element-x-ios-rageshakes/issues/278

#### What did you expect?

ability to filter roomlist, especially for a room i’m looking at right now.

#### What happened instead?

broken filter.

### Will you send logs?

Yes

ara4n:
haven't seen this in forever

PR #828, opened 2023-04-24
#821 - Remove lazy room list filter (long shot)

stefanceriu:


pixlwave (approved):
Looking at the code now, I'm remembering we possibly had to filter out spaces before, so that's why we had `lazy` as we could chain that property onto this one.

PR #829, opened 2023-04-25
Renaming screens returns

stefanceriu:
Rename remaining screens following https://github.com/vector-im/element-x-ios/pull/824 and https://github.com/vector-im/element-x-ios/pull/826


pixlwave (approved):
LGTM 👍

PR #832, opened 2023-04-25
Don't trigger checks for actions

kittykat:
Make it so changes to files only in the .github directory don't trigger other automation

### Pull Request Checklist

- [ ] I read the [contributing guide](https://github.com/vector-im/element-ios/blob/develop/CONTRIBUTING.md)

- [ ] Voiceover enabled.

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -3,7 +3,7 @@ name: Danger CI
 on:
   pull_request:
     paths-ignore:
-    - '.github/**'
+      - '.github/**'
 
 jobs:
   build:
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -1,6 +1,9 @@
 name: Danger CI
 
-on: [pull_request]
+on:
+  pull_request:
+    paths-ignore:
+    - '.github/**'
 
 jobs:
   build:
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -3,6 +3,8 @@ name: PR Build
 on:
   pull_request:
     types: [ labeled, synchronize, opened, reopened ]
+    paths-ignore:
+    - '.github/**'
 
 jobs:
   build:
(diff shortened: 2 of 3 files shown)


kittykat:
Need to squash commits when merging

.github/workflows/danger.yml
```diff
@@ -1,6 +1,9 @@
 name: Danger CI
 
-on: [pull_request]
+on:
+  pull_request:
+    paths-ignore:
+      - '.github/**'
```
pixlwave, line 6: Could we maybe be less general here? E.g. something like
```
- '.github/CODEOWNERS'
- '.github/PULL_REQUEST_TEMPLATE.md'
- '.github/ISSUE_TEMPLATE'
- '.github/workflows/triage_*'
```
Otherwise this change will prevent us from testing out changes to a PR workflow by using a PR.

kittykat: I'm not really sure what that specific workflow does, but also I don't think that the new change is applied to the PR that introduces it so you'd need to test on the next PR anyway. I could be wrong…

pixlwave: We can definitely iterate on a workflow within a PR, e.g. see this [run](https://github.com/vector-im/element-x-ios/actions/runs/4800489532/jobs/8541480244?pr=836#step:2:5) from this [test PR](https://github.com/vector-im/element-x-ios/pull/836).

pixlwave: Also, sorry my comment probably wasn't very clear. I was suggesting to make the same change to all the ignores on all the workflows. So that we can always test out updates on all of them easily.

.github/workflows/pr-build.yml
pixlwave, line 1: FWIW, this workflow will only run when a PR is manually tagged with the `Trigger-PR-Build` label, so its probably not worth adding the exclusion on this one at all.

stefanceriu:
@kittykat What are we doing with this PR?

stefanceriu:
Talked about this offline with both kat and doug and we agreed that it's not worth the effort, at least not at the moment.

PR #833, opened 2023-04-25
Adopt the bump event types sliding sync proxy API. Allow only `m.room…

stefanceriu:
….message and `m.room.encrypted` to reorder the room list


pixlwave (approved):
Looks good, should we add stickers too as we support those?

PR #834, opened 2023-04-25
Compound Fonts

pixlwave:
This PR switches all the fonts over to use `CompoundFonts` instead of `ElementFonts`. `Font.element` and `UIFont.element` are marked as deprecated. The following fonts were changed:
- `(edited)` suffix on messages is slightly bigger (using `bodyXS`).
- The `NEW` indicator on the unread marker is slightly bigger (using `bodyXS`).
- The size of text in `ElementTextFieldStyle` has gone from `callout` to `bodyLG`.
- The size of the timestamp in the Modern layout is slightly bigger.

Note, the fonts that went from `Bold` to `Semibold` actually seem to render the same as SwiftUI's `.bold()` modifier appears to use semibold when the font size is `body` or smaller 🤷‍♂️


stefanceriu (approved):
👏

PR #835, opened 2023-04-25
Fix flakey unit tests.

pixlwave:
Await specific view states rather than the next one.

Additionally adds a `UnitTestsAppCoordinator` that configures the ServiceLocator with default mock services. Some tests update these with pre-configured mocks, but that seems fine to me.


stefanceriu (approved):
Yey 👏

stefanceriu (approved):

PR #839, opened 2023-04-27
Use search field style from Compound.

pixlwave:
This PR updates `compound-ios` and uses the `compoundSearchField()` modifier to style all uses of `searchable` in the app.


ElementX/Sources/Other/SwiftUI/Views/UserProfileCell.swift
```diff
@@ -44,7 +44,7 @@ struct UserProfileCell: View {
                 if !user.isVerified {
                     HStack(alignment: .firstTextBaseline, spacing: 4) {
                         Text(Image(systemName: "exclamationmark.circle.fill"))
-                            .foregroundColor(.compound.textActionCritical)
+                            .foregroundColor(.compound.textCriticalPrimary)
```
pixlwave, line 47: This token was renamed.

Velin92 (approved):
-114 and only +16? I like this

PR #842, opened 2023-04-28
Contextual menu for rooms

alfogrillo:
Fixes https://github.com/vector-im/element-x-ios/issues/736

**Result**
iPhone


ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenModels.swift
```diff
@@ -106,6 +106,7 @@ struct LeaveRoomAlertItem: AlertItem {
         case `private`
     }
 
+    let roomId: String
```
Velin92, line 109: I don't think we need to expose this here, I guess the `leaveRoomAlertActions(_ item: LeaveRoomAlertItem)` can just use `room.id` directly

ElementX/Sources/Services/UserSession/UserSessionFlowCoordinator.swift
```diff
@@ -126,6 +126,11 @@ class UserSessionFlowCoordinator: CoordinatorProtocol {
                 self.presentRoomWithIdentifier(roomId)
             case (.invitesScreen, .deselectRoom, .invitesScreen):
                 break
+                
+            case (.roomList(let currentRoomId), .selectRoomDetails(let roomId), .roomList) where currentRoomId == roomId:
+                break
+            case (.roomList, .selectRoomDetails(let roomId), .roomList(let selectedRoomId)) where roomId == selectedRoomId:
+                self.presentRoomDetails(roomIdentifier: roomId)
```
Velin92, line 133: I think this should also include the animated parameter (actually also the presentRoomWithIdentifier above too)

Velin92 (approved):
LGTM, just a few comments

ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenModels.swift
alfogrillo: I needed the id just in the `HomeScreen` [here](https://github.com/vector-im/element-x-ios/pull/842/files#diff-339f2c71d6740123007026ab3b894f3053921783e0f1665181db568e79a3847aR262) since there is no concept of "room" there.
In the `RoomDetailsViewModel` it still works with `room.id` like you say.

ElementX/Sources/Services/UserSession/UserSessionFlowCoordinator.swift
alfogrillo: Applied: https://github.com/vector-im/element-x-ios/pull/842/commits/7e60f3b41774e964d6e8621e05270b26e7d1f38a

PR #848, opened 2023-05-02
Avoid race and duplication conditions between local and remote notifications

Velin92:
Implemented a system to avoid race condition and duplication between local notifications and other local notifications, and most importantly between local notifications and remote notifications:

- LocalW/Local: Using a timestamp value exposed on the event wrapped in the notification item, this is compared with the date of the last app launch, if the timestamp is above this value then we can show the notification otherwise we discard it
- LocalW/Remote: Saving a notification identifier value in a shared store between NSE and the main app, this notification identifier is just a sha256 of the client's user id, the room id and the event id. When a remote notification/local notification is about to be served we check first if the identifier is already present, if it is we discard the notification, if not we save the identifier and then we display the notification.

The PR also features some various code improvements to reduce the size of the process functions.

The only thing left to do is decide if we want to enable local notifications immediately right now, or we want to wait for remote notifications to also decrypt their content.

As of right now I tested the code, and local notifications are always faster, so they always take precedence over NSE remote notifications, so when the app is not in a killed state, we are able to serve rich push notifications that show content and media! However not sure if given bad network connections or old devices remotes could take precedence hindering a bit the experience, for now it has never happened once to me testing it on my physical device.

### DIFF
--- a/ElementX/Sources/Application/AppSettings.swift
+++ b/ElementX/Sources/Application/AppSettings.swift
@@ -69,6 +69,8 @@ final class AppSettings {
     /// deleted between runs so should clear data in the shared container and keychain.
     @UserPreference(key: UserDefaultsKeys.lastVersionLaunched, storageType: .userDefaults(store))
     var lastVersionLaunched: String?
+
+    let lastLaunchDate = Date()
     
     /// The Set of room identifiers of invites that the user already saw in the invites list.
     /// This Set is being used to implement badges for unread invites.
@@ -155,8 +157,6 @@ final class AppSettings {
     
     let permalinkBaseURL = URL(staticString: "https://matrix.to")
 
-    let lastAppLaunchDate = Date()
-    
     // MARK: - Feature Flags
     
     // MARK: Start Chat
--- a/ElementX/Sources/Application/AppSettings.swift
+++ b/ElementX/Sources/Application/AppSettings.swift
@@ -150,6 +150,8 @@ final class AppSettings: ObservableObject {
     // MARK: - Other
     
     let permalinkBaseURL = URL(staticString: "https://matrix.to")
+
+    let lastAppLaunchDate = Date()
     
     // MARK: - Feature Flags
     
--- a/ElementX/Sources/Application/AppSettings.swift
+++ b/ElementX/Sources/Application/AppSettings.swift
@@ -148,7 +148,7 @@ final class AppSettings {
     var pusherProfileTag: String?
 
     /// A set of all the notification identifiers that have been served so far, it's reset every time the app is launched
-    @UserPreference(key: SharedUserDefaultsKeys.servedNotificationIdentifiers, defaultValue: [], storageType: .userDefaults(store))
+    @UserPreference(key: SharedUserDefaultsKeys.servedNotificationIdentifiers, initialValue: [], storageType: .userDefaults(store))
     var servedNotificationIdentifiers: Set<String>
         
     // MARK: - Other
--- a/ElementX/Sources/Application/AppSettings.swift
+++ b/ElementX/Sources/Application/AppSettings.swift
@@ -18,7 +18,7 @@ import Foundation
 import SwiftUI
 
 /// Store Element specific app settings.
-final class AppSettings: ObservableObject {
+final class AppSettings {
     private enum UserDefaultsKeys: String {
         case lastVersionLaunched
         case seenInvites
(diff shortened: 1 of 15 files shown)


ElementX/Sources/Application/AppSettings.swift
```diff
@@ -146,10 +146,16 @@ final class AppSettings: ObservableObject {
...
+
+    /// A set of all the notification identifiers that have been served so far, it's reset every time the app is launched
+    @UserPreference(key: SharedUserDefaultsKeys.servedNotificationIdentifiers, initialValue: [], storageType: .userDefaults(store))
+    var servedNotificationIdentifiers: Set<String>
         
     // MARK: - Other
     
     let permalinkBaseURL = URL(staticString: "https://matrix.to")
+
+    let lastAppLaunchDate = Date()
```
pixlwave, line 158: Probably should go alongside `lastVersionLaunched` in the Application section. Could also call it `lastLaunchDate` for consistency.

ElementX/Sources/Other/Extensions/UNNotificationContent.swift
```diff
@@ -22,6 +22,18 @@ extension UNNotificationContent {
     @objc var receiverID: String? {
         userInfo[NotificationConstants.UserInfoKey.receiverIdentifier] as? String
     }
+
+    @objc var notificationIdentifier: String? {
```
pixlwave, line 26: Perhaps for consistency with the others?
```suggestion
    @objc var notificationID: String? {
```

ElementX/Sources/Services/Notification/Proxy/NotificationItemProxy.swift
```diff
@@ -39,8 +42,20 @@ protocol NotificationItemProxyProtocol {
     var isEncrypted: Bool { get }
 }
 
+extension NotificationItemProxyProtocol {
+    var identifier: String? {
```
pixlwave, line 46: I wonder if both this and `notificationIdentifier` should simply become `id` and then we could use it for `Identifiable` in the future if we ever needed that.
```suggestion
    var id: String? {
```

NSE/Sources/NotificationServiceExtension.swift
```diff
@@ -29,8 +30,8 @@ class NotificationServiceExtension: UNNotificationServiceExtension {
         guard !DataProtectionManager.isDeviceLockedAfterReboot(containerURL: URL.appGroupContainerDirectory),
               let roomId = request.roomId,
               let eventId = request.eventId,
-              let notificationID = request.pusherNotificationClientIdentifier,
-              let credentials = keychainController.restorationTokens().first(where: { $0.restorationToken.pusherNotificationClientIdentifier == notificationID }) else {
+              let clientID = request.pusherNotificationClientIdentifier,
```
pixlwave, line 33: I almost suggested this before when it was called `notificationID`, how about we call this something like `accountID`. To make it very much separate from notifications, clients etc?

ElementX/Sources/Other/UserPreference.swift
```diff
@@ -75,6 +75,11 @@ extension UserPreference {
     convenience init<R: RawRepresentable>(key: R, defaultValue: T, storageType: StorageType) where R.RawValue == String {
         self.init(key: key.rawValue, defaultValue: defaultValue, storageType: storageType)
     }
+
+    convenience init<R: RawRepresentable>(key: R, initialValue: T, storageType: StorageType) where R.RawValue == String {
+        self.init(key: key, defaultValue: initialValue, storageType: storageType)
+        wrappedValue = initialValue
+    }
```
pixlwave, line 82: This is confusing me a bit. How come it doesn't use `key.rawValue` like above? And then are we saying that it will wipe out the stored value every time it is created?

If so that should probably be loudly documented, and maybe the parameter renamed to something like `resetsValueTo` because it is very close and I worry people would use this by mistake not realising what it is doing.

pixlwave (approved):
That's much neater not having to pass the IDs around everywhere 👍

NSE/Sources/Other/NSESettings.swift
```diff
@@ -0,0 +1,28 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+
+final class NSESettings {
```
pixlwave, line 19: Oh, looking at this again though, I don't think this is what Stefan was suggesting.

Instead we should make a `protocol NSESettings` with whatever you want access to and conform `AppSettings` to that. Then store an instance of `AppSettings` as an `NSESettings`.

Velin92: The problem that I had, that made me discard the solution is that this requires AppSettings to be imported in the NSE, whjich creates a ton of issues, we need to import so many other not needed classes (like Analytics, Timeline types) and so on, and just doesn't feel it's worth it.

ElementX/Sources/Other/UserPreference.swift
Velin92: It's not using the `rawValue`, because it's actively using the convenience constructor defined above.

What it does is that it does not only create the UserPreference, but also initialises the user preference to have its default value already stored immediately in the store.

Instead the default value is just a default value, but is not really stored.

I can add a piece of documentation to better explain it

Velin92: I added documentation for it

pixlwave (approved):

PR #850, opened 2023-05-04
Pass attachment bodies when loading media

stefanceriu:
We now use the attachment body string to generate the correct temporary filename so that it's available to system dialogs later.


pixlwave (approved):
LGTM 👍

PR #852, opened 2023-05-04
Add developer option for clearing the application cache

stefanceriu:
This will delete a particular folder from within the user session and then reload the whole application without logging out the user.

### DIFF
--- a/ElementX/Sources/Screens/DeveloperOptionsScreen/View/DeveloperOptionsScreen.swift
+++ b/ElementX/Sources/Screens/DeveloperOptionsScreen/View/DeveloperOptionsScreen.swift
@@ -69,12 +69,11 @@ struct DeveloperOptionsScreen: View {
             }
             
             Section {
-                Button {
+                Button(role: .destructive) {
                     context.send(viewAction: .clearCache)
                 } label: {
                     Text("Clear cache")
                         .frame(maxWidth: .infinity)
-                        .foregroundColor(.red)
                 }
             }
         }
--- a/ElementX/Sources/Screens/DeveloperOptionsScreen/View/DeveloperOptionsScreen.swift
+++ b/ElementX/Sources/Screens/DeveloperOptionsScreen/View/DeveloperOptionsScreen.swift
@@ -66,9 +66,17 @@ struct DeveloperOptionsScreen: View {
                     Text("🥳")
                         .frame(maxWidth: .infinity)
                 }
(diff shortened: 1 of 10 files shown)


ElementX/Sources/Screens/DeveloperOptionsScreen/View/DeveloperOptionsScreen.swift
```diff
@@ -66,9 +66,17 @@ struct DeveloperOptionsScreen: View {
...
-            .formSectionStyle()
+            
+            Section {
+                Button {
+                    context.send(viewAction: .clearCache)
+                } label: {
+                    Text("Clear cache")
+                        .frame(maxWidth: .infinity)
+                        .foregroundColor(.red)
+                }
```
pixlwave, line 78: Not that it matters in here, but lets stick to good habbits
```suggestion
                Button(role: .destructive) {
                    context.send(viewAction: .clearCache)
                } label: {
                    Text("Clear cache")
                        .frame(maxWidth: .infinity)
                }
```

ElementX/Sources/Services/UserSession/UserSessionStore.swift
```diff
@@ -150,4 +155,16 @@ class UserSessionStore: UserSessionStoreProtocol {
             MXLog.failure("Failed deleting the session data: \(error)")
         }
     }
+    
+    private func deleteCachesFolder(for userID: String) {
+        // Rust sanitises the user ID replacing invalid characters with an _
+        let sanitisedUserID = userID.replacingOccurrences(of: ":", with: "_")
+        let url = baseDirectory.appendingPathComponent(sanitisedUserID).appendingPathComponent("matrix-sdk-state")
```
pixlwave, line 162: At some stage we should probably expose a method from Rust that gives use the path for a `userID` with the shared logic so we don't need to keep this in sync. But for the moment, it might be nice to extract this (excluding the `state` dir) into its own `func path(for userID: String)` with a warning/todo on it so we actually do it?

pixlwave (approved):
Nice, the state machine definitely helps here 👏

ElementX/Sources/Services/UserSession/UserSessionStore.swift
stefanceriu: Done 👍

pixlwave (approved):
LGTM 😎

PR #854, opened 2023-05-05
Remote Push Notifications can now be decrypted

Velin92:
Remote pushes can now be decrypted.

NOTE: This only works if the current keys are valid, working on the Rust side to find a way to do one single sliding sync to refresh keys


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -105,7 +105,8 @@ class ClientProxy: ClientProxyProtocol {
 
         let delegate = WeakClientProxyWrapper(clientProxy: self)
         client.setDelegate(delegate: delegate)
-        // Uncomment to test local notifications
+
+        // Comment to remove local notifications
```
pixlwave, line 109: Possibly more relevant?
```suggestion
        // Set up sync listener for generating local notifications.
```

NSE/Sources/NotificationServiceExtension.swift
```diff
@@ -165,3 +149,18 @@ class NotificationServiceExtension: UNNotificationServiceExtension {
         MXLog.info("\(tag) deinit")
     }
 }
+
+private extension Client {
+    func getNotificationItemProxy(roomID: String, eventID: String) async throws -> NotificationItemProxyProtocol {
```
pixlwave, line 154: As this is async, no need for the `get`.
```suggestion
    func notificationItemProxy(roomID: String, eventID: String) async throws -> NotificationItemProxyProtocol {
```

NSE/Sources/NotificationServiceExtension.swift
```diff
@@ -66,80 +66,64 @@ class NotificationServiceExtension: UNNotificationServiceExtension {
...
-        } else {
-            // This is not very likely, as it should discard the notification sooner
+        } catch {
+            MXLog.error("NSE run error: \(error)")
             return discard()
         }
     }
 
-    private func createMediaProvider(with credentials: KeychainCredentials) throws -> MediaProviderProtocol {
+    private func getClient(from credentials: KeychainCredentials) throws -> Client {
```
pixlwave, line 105: Maybe `buildClient(for:)` or `makeClient(for:)` to indicate its creating and restoring a new client?

ElementX/Sources/Services/Notification/Proxy/NotificationItemProxy.swift
```diff
@@ -102,10 +102,7 @@ struct NotificationItemProxy: NotificationItemProxyProtocol {
     }
 }
 
-// The mock and the protocol are just temporary until we can handle
-// and decrypt notifications both in background and in foreground
-// but they should not be necessary in the future
-struct MockNotificationItemProxy: NotificationItemProxyProtocol {
+struct GenericNotificationItemProxy: NotificationItemProxyProtocol {
```
pixlwave, line 105: I had to read through the code to understand what a generic notification is. Might be better to call it something like an empty, placeholder or unknown notification?

pixlwave (approved):

PR #857, opened 2023-05-06
Improve media uploading content type selection. Fix pasting in screenshots not working

stefanceriu:
Sometimes selected media reports multiple registered content types e.g. a live photo will report

```
▿ 3 elements
  ▿ 0 : <UTType 0x282692150> com.apple.live-photo-bundle (not dynamic, declared)
    - _type : <UTType 0x282692150> com.apple.live-photo-bundle (not dynamic, declared)
  ▿ 1 : <_UTCoreType 0x1e8395670> public.jpeg (not dynamic, declared)
    - _type : <_UTCoreType 0x1e8395670> public.jpeg (not dynamic, declared)
  ▿ 2 : <_UTCoreType 0x1e83957d0> public.heic (not dynamic, declared)
    - _type : <_UTCoreType 0x1e83957d0> public.heic (not dynamic, declared)
```

and in those cases we need to make sure we don't just select the first one but filter down to the types we actually support. More than that, in this case we also receive 2 possible image types and we should try to use jpeg if available.

Fixes https://github.com/vector-im/element-x-ios/issues/858

### DIFF
--- a/ElementX/Sources/Other/Extensions/NSItemProvider.swift
+++ b/ElementX/Sources/Other/Extensions/NSItemProvider.swift
@@ -15,12 +15,30 @@
 //
 
 import Foundation
+import UniformTypeIdentifiers
 
 extension NSItemProvider {
     var isSupportedForPasteOrDrop: Bool {
-        !registeredContentTypes
-            .compactMap(\.preferredMIMEType)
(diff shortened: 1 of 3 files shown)


pixlwave (approved):
LGTM, a little bit of me wonders if iOS could ever provide a jpeg as supporting content to something else, but if that's ever possible I'm sure we'll find out in time 😃

PR #860, opened 2023-05-08
Fix missing UI tests simulator

stefanceriu:
This change will atempt to create the desired simulator if not available by default. This happens because Xcode 14.3 no longer includes the `iPad (9th generation)` and `iPhone 13 Pro` simulators by default


Velin92 (approved):
LGTM

PR #862, opened 2023-05-08
Fix for the sender icon not being rendered in notifications sometimes

Velin92:
This was happening because we were using both the flow for groups and normal senders at the same time, which made the notification confused on which image the render (the one set on the sender, or the one set on the intent).

Now we only set the image on the intent when we are sure we want to render a group.


Tools/Sources/Utilities.swift
```diff
@@ -19,6 +19,7 @@ enum Utilities {
     @discardableResult
     static func zsh(_ command: String, workingDirectoryURL: URL = projectDirectoryURL) throws -> String? {
         let process = Process()
+        process.environment = [:]
```
stefanceriu, line 22: You should probably put this in a separate commit so there's some history around it

PR #864, opened 2023-05-08
Low effort fix for timeline date separator glitches

stefanceriu:
While this is by no means perfect it does seem to mitigate the glitches. This would be great to come directly from the Rust side that's apparently quite hard to do.

Problem is seeing this for a split second, SwiftUI trying to render the same separator with different identifiers:
Screenshot 2023-05-08 at 18 27 38

### DIFF
--- a/ElementX/Sources/Services/Timeline/RoomTimelineProvider.swift
+++ b/ElementX/Sources/Services/Timeline/RoomTimelineProvider.swift
@@ -53,7 +53,7 @@ class RoomTimelineProvider: RoomTimelineProviderProtocol {
             
             roomTimelineListener
                 .itemsUpdatePublisher
-                .collect(.byTime(serialDispatchQueue, 0.025))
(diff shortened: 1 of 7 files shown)


ElementX/Sources/Services/Timeline/RoomTimelineProvider.swift
```diff
@@ -53,7 +53,7 @@ class RoomTimelineProvider: RoomTimelineProviderProtocol {
             
             roomTimelineListener
                 .itemsUpdatePublisher
-                .collect(.byTime(serialDispatchQueue, 0.025))
+                .collect(.byTime(serialDispatchQueue, 0.1))
```
pixlwave, line 56: Presumably changing this also helps with the glitches by slowing the updates?

pixlwave (approved):
I haven't tried it, but seems like a good start until we get stable identifiers out of the SDK.

PR #865, opened 2023-05-08
Removed clean process.environment

Velin92:
Sadly didn't really fix the issue, but caused only more trouble with brew....


stefanceriu (approved):
👍

pixlwave (approved):

PR #871, opened 2023-05-11
Tag nightlies after a successful build

stefanceriu:


fastlane/Fastfile
```diff
@@ -238,17 +235,46 @@ lane :prepare_next_release do
 
   xcodegen(spec: "project.yml")
 
-  sh("git config --global user.name 'Element CI'")
-  sh("git config --global user.email 'ci@element.io'")
```
pixlwave, line 242: Wouldn't we want these set before committing so that the commit is shown as made by CI?

fastlane/Fastfile
```diff
@@ -238,17 +235,46 @@ lane :prepare_next_release do
...
+
+lane :tag_nightly do |options|
+  build_number = options[:build_number]
+  UI.user_error!("Invalid build number.") unless !build_number.to_s.empty?
+
+  xcodegen_project_file_path = "../project.yml"
+  data = YAML.load_file xcodegen_project_file_path
+  current_version = data["settings"]["MARKETING_VERSION"]
+
+  tag_name = "nightly-#{current_version}.#{build_number}"
```
pixlwave, line 253: Maybe this should be `nightly/versoionnumber` so they're more easily hidden in a git client?

pixlwave (approved):

fastlane/Fastfile
stefanceriu: Yeah, good catch 👍

fastlane/Fastfile
stefanceriu: Yep, makes perfect sense

PR #882, opened 2023-05-11
Disable Local Push Notifications + Start Handling invite notifications

Velin92:
Local Push notifications are quite a niche case, so we don't really need that much, also sometimes can be a bit flaky, so we decided to disable them on iOS.

Also
- did some refactor in the code for push notifications
- started implementing the `processInvite` function for invite notifications (however they can't be properly handled yet since the NotificationItem that generated them throws an error)
- Finally found the fix for the missing avatars
- Removed the reply action  in the notification since we don't support it yet


ElementX/Sources/Other/Extensions/UNNotificationContent.swift
```diff
@@ -127,7 +127,7 @@ extension UNMutableNotificationContent {
         if let mediaSource = iconType.mediaSource {
             switch await mediaProvider?.loadFileFromSource(mediaSource) {
             case .success(let mediaFile):
-                image = try INImage(imageData: Data(contentsOf: mediaFile.url))
+                image = try await INImage(imageData: Data(from: mediaFile.url))
```
stefanceriu, line 130: Shouldn't you use the `MediaLoaders` `loadMediaContentForSource` here instead?

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -540,7 +544,7 @@ class ClientProxy: ClientProxyProtocol {
         if let notificationsListBuilder {
             MXLog.info("Registering notifications view")
             _ = slidingSync?.addList(listBuilder: notificationsListBuilder)
-        } else {
+        } else if ServiceLocator.shared.settings.enableLocalPushNotifications {
```
stefanceriu, line 547: Should you maybe wrap the whole `if let notificationsListBuilder...` block in this check instead?

stefanceriu (approved):
Left some comments inline, otherwise lgtm 👍

ElementX/Sources/Services/Client/ClientProxy.swift
Velin92: oh yeah that's better

ElementX/Sources/Other/Extensions/UNNotificationContent.swift
Velin92: I have the MediaProviderProtocol here

Velin92: oh wait you suggest actually instantiating a MediaLoader and using it in place of creating the data? Ok let's give it a try honestly I am not a fan of this AI generated solution of using a URLSessionProtocol

PR #861, opened 2023-05-08
Reply rendering

stefanceriu:
The RustSDK has stopped seding us reply fallbacks as Matrix as a whole is moving away from them so we now need to start using the already available `InReplyToDetails`. 

This PR refactors the existing code to extract out the types common to `MessageTimelineItems` and `RepliedToEventDetails` so that we can now bundle `MessageContent` together with both. All message type timeline items now have a `replyDetails` property that they can use in rendering.
It also exposes a new Room method for fetching `.unavailable` event details when a certain reply appears on screen (`RoomTimelineController.processItemAppearance`)

Screenshot 2023-05-08 at 12 14 34
Screenshot 2023-05-08 at 12 14 59

Known issues:
- [ ] Reply details sometimes don't load
- [ ] They are only rendered in textual format for all message content types (e.g. we won't show a thumbnail in the reply bubble)
- [ ] The reply bubble doesn't expand to fill all the available space. Couldn't figure out how to do that, not even with a custom Layout

PR #883, opened 2023-05-11
Use the existing quote bubble layout with TimelineReplyView.

pixlwave:
Follow-up to #861, this PR uses the existing layout and formatting from `FormattedBodyText` for the new `TimelineReplyView` when contained within a message bubble. The result should be the same layout as we had before, but now with the sender's name formatted as in the Figma:

Screenshot 2023-05-11 at 6 54 47 pm


stefanceriu (approved):
Whoa this is awesome, thank you! 👏

PR #888, opened 2023-05-12
Fixing Tests for Xcode 14.3

Velin92:


.github/workflows/integration-tests.yml
```diff
@@ -36,6 +36,15 @@ jobs:
           INTEGRATION_TESTS_HOST: ${{ secrets.INTEGRATION_TESTS_HOST }}
           INTEGRATION_TESTS_USERNAME: ${{ secrets.INTEGRATION_TESTS_USERNAME }}
           INTEGRATION_TESTS_PASSWORD: ${{ secrets.INTEGRATION_TESTS_PASSWORD }}
+      
+      - name: Archive artifacts
```
stefanceriu, line 40: Niice 👏

UITests/Sources/Application.swift
```diff
@@ -30,24 +30,31 @@ struct Application {
...
+    func assertScreenshot(_ identifier: UITestsScreenIdentifier, step: Int? = nil, insets: UIEdgeInsets? = nil, delayInMilliseconds: UInt = 2000, precision: Float = 0.99) async throws {
         var snapshotName = identifier.rawValue
         if let step {
             snapshotName += "-\(step)"
         }
-        
+
+        // Sometimes the CI might be too slow to load the content so let's wait the delay time
+        if delayInMilliseconds > 0 {
+            try await Task.sleep(for: .milliseconds(delayInMilliseconds))
```
stefanceriu, line 46: Should we maybe just `try?` this delay instead of exposing it through the whole API and forcing the test to throw?

stefanceriu (approved):
lgtm, nicely done 👏

UITests/Sources/Application.swift
Velin92: what could be the issue if the tests throw?

stefanceriu: not much, just having to write that everywhere is annoying 😄

Velin92: no idea, but I guess if it would fail, maybe getting that exception could be insightful, so I would prefer to keep the throw.

stefanceriu: okay, fair enough 👍

PR #889, opened 2023-05-12
Fixes cache cleaning behavior after sdk change

stefanceriu:
The new Sqlite state store is no longer in the matrix-sdk-state subfolder. Also, the client wasn't properly dealocated at the time this was run so I've added some artifical delays


Velin92 (approved):
LGTM however couldn't the sdk implement a callback to tell us that the Sync has stopped so we don't need to wait for an arbitrary amount of time?

stefanceriu:
> that the Sync has stopped

This is more about the userSession+client getting deallocated on our side. Because the Settings screen is dismissed with an animation and because that holds the user session it takes a bit longer to clean up things.

PR #890, opened 2023-05-12
Disable unhandled_throwing_task rule.

pixlwave:
Disable the new [`unhandled_throwing_task`](https://realm.github.io/SwiftLint/unhandled_throwing_task.html) rule in SwiftLint. This is going to get annoying if we have to `do { } catch { }` around every `Task.sleep`.


stefanceriu (approved):
🚀

PR #892, opened 2023-05-15
Add background app refresh support

stefanceriu:
This PR add support for background app refreshes. 

The app currently stops syncing after a background task that starts when the app resigns active expires. We will now also schedule a background refresh task to run 30 seconds after this happens (just a suggestion for the system). When that happens the sync is restarted. It will attempt to run for a maximum or 10 sync responses or 10 seconds and then shut down again.

How to test this amazing piece of kit:

1) Run on a real device, simulators are not supported
2) Put a breakpoint right after scheduling the task -> `Successfully scheduled background app refresh task`
3) Execute this debugger command `e -l objc -- (void)[[BGTaskScheduler sharedScheduler] _simulateLaunchForTaskWithIdentifier:@"io.element.elementx.background.refresh"]`
4) Continue the execution
5) See the background task getting invoked and profit

(https://developer.apple.com/documentation/backgroundtasks/starting_and_terminating_tasks_during_development)

### DIFF
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -20,7 +20,7 @@ import MatrixRustSDK
 import SwiftUI
 import Version
 
-class AppCoordinator: AppCoordinatorProtocol {
+class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate, NotificationManagerDelegate {
     private let stateMachine: AppCoordinatorStateMachine
     private let navigationRootCoordinator: NavigationRootCoordinator
     private let userSessionStore: UserSessionStoreProtocol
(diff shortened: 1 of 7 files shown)


ElementX/Sources/Application/AppSettings.swift
```diff
@@ -84,6 +84,9 @@ final class AppSettings {
     /// that don't yet have an officially trusted proxy configured in their well-known.
     let slidingSyncProxyURL: URL? = nil
     
+    /// The task identifier used for background app refresh. Also used in main target's the Info.plist
+    let backgroundAppRefreshTaskIdentifier = "io.element.elementx.background.refresh"
```
Velin92, line 88: shouldn't we use the InfoPlistReader for this value here? So the string is always tied (don't think we are changing bundle name, but for for forks could be useful)

ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -342,7 +347,7 @@ class AppCoordinator: AppCoordinatorProtocol {
                 case .didReceiveAuthError(let isSoftLogout):
                     stateMachine.processEvent(.signOut(isSoft: isSoftLogout))
                 case .updateRestorationToken:
-                    userSessionStore.refreshRestorationToken(for: userSession)
+                    _ = userSessionStore.refreshRestorationToken(for: userSession)
```
Velin92, line 350: @discardableResult on this func

ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -412,11 +416,17 @@ class AppCoordinator: AppCoordinatorProtocol {
...
         backgroundTask = backgroundTaskService.startBackgroundTask(withName: "SuspendApp: \(UUID().uuidString)") { [weak self] in
-            self?.stopSync()
+            guard let self else { return }
             
-            self?.backgroundTask = nil
-            self?.isSuspended = true
+            stopSync()
+        
+            backgroundTask = nil
+            isSuspended = true
```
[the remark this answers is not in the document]
stefanceriu: Ha, same here, but it's just soo nice not to use

ElementX/Sources/Application/AppSettings.swift
Velin92: Wha if we make sure that the plist arrat is always ordered in a certain way?
I guess a comment in the .yml on the expected order could help

stefanceriu: I think that's finicky and can lead to pretty bad bugs. Better to be safe than sorry.

PR #895, opened 2023-05-15
Resurect the Xcode project file

stefanceriu:
Not checking in the project file is causing problems while switching branches and rebasing because of Xcode automatically modifying the resolved packages file.
This PR reintroduces a checked in project file, removes if from gitignore (together with that ugly hack) and stops re-generating it on CI runs


Velin92 (approved):

Velin92 (approved):

pixlwave:
Wooo 🥳

PR #896, opened 2023-05-15
Run single Test on CI for faster debugging

Velin92:
Added the possibility of providing the name of the test, so that we can run only ONE single test to debug what is wrong with it on the CI without the need of waiting 1 hour of tests to complete.

if no test name is provided, all tests will run.

Test names need to be provided in the following way:
ClassName/testName

### DIFF
--- a/.github/workflows/ui_tests.yml
+++ b/.github/workflows/ui_tests.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       test_name:
-        description: 'Name of the specific test to run'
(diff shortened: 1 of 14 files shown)


.github/workflows/ui_tests.yml
```diff
@@ -30,12 +34,16 @@ jobs:
...
 
       - name: Setup environment
-        run:
-          source ci_scripts/ci_common.sh && setup_github_actions_environment
+        run: source ci_scripts/ci_common.sh && setup_github_actions_environment
 
       - name: Run tests
-        run: bundle exec fastlane ui_tests
-      
+        run: |
```
stefanceriu, line 40: Mega useful 👏

stefanceriu (approved):
🙌

PR #903, opened 2023-05-16
Cleanup the background task that pauses the SDK when resigning active

Velin92:
From the video you can see that only the third time I background the app (where I waited about 30 seconds) when I foregrounded it, I got back the loading indicator showing the the sync has restarted.
 
 So we are extending the duration of the sync for 30 seconds after suspension, and then stopping it in the expiration handler
 
 We restart the sync only if the app is coming from a suspended state and the sync was stopped.

https://github.com/vector-im/element-x-ios/assets/34335419/1ec1f508-f4f2-43b7-95ac-5c736c4c643f

### DIFF
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -348,7 +348,7 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
             hideLoadingIndicator()
         }
         
-        userSession.clientProxy.stopSync()
+        stopSync()
         userSessionFlowCoordinator?.stop()
         
         guard !isSoft else {
@@ -462,7 +462,7 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
         
         navigationRootCoordinator.setRootCoordinator(SplashScreenCoordinator())
         
-        userSession.clientProxy.stopSync()
+        stopSync()
         userSessionFlowCoordinator?.stop()
         
         let userID = userSession.userID
@@ -498,8 +498,8 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
 
     // MARK: - Application State
 
-    private func stopSync() {
-        userSession?.clientProxy.stopSync()
+    private func stopSync(completionHandler: (() -> Void)? = nil) {
+        userSession?.clientProxy.stopSync(completionHandler: completionHandler)
     }
 
     private func startSync() {
@@ -544,9 +544,12 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
         backgroundTask = backgroundTaskService.startBackgroundTask(withName: "SuspendApp: \(UUID().uuidString)") { [weak self] in
             guard let self else { return }
             
-            stopSync()
-        
-            backgroundTask = nil
+            stopSync { [weak self] in
+                guard let self else { return }
+                backgroundTask?.stop()
+                backgroundTask = nil
+            }
+
             isSuspended = true
         }
         
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -542,10 +542,13 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
         }
 
         backgroundTask = backgroundTaskService.startBackgroundTask(withName: "SuspendApp: \(UUID().uuidString)") { [weak self] in
-            self?.userSession?.clientProxy.stopSync { [weak self] in
-                guard let self else { return }
-                backgroundTask?.stop()
-                backgroundTask = nil
+            guard let self else {
+                return
+            }
+            userSession?.clientProxy.stopSync {
+                // No need to weakify self, this is a non escaping closure
+                self.backgroundTask?.stop()
+                self.backgroundTask = nil
             }
         }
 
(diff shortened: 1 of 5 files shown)


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -161,10 +161,20 @@ class ClientProxy: ClientProxyProtocol {
...
+            return
+        }
+
         MXLog.info("Stopping sync")
-        slidingSyncObserverToken?.cancel()
-        slidingSyncObserverToken = nil
+        slidingSyncObserverToken.cancel()
+
+        Task {
+            while !slidingSyncObserverToken.isFinished() { }
```
Velin92, line 174: not super elegant I know, but as of right now we don't have callbacks that tell us when a TaskHandle has finished

stefanceriu: I wonder if we could benefit from exposing another rust Client delegate method to let us know when this is done 🤔

ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -517,9 +517,12 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
         backgroundTask = backgroundTaskService.startBackgroundTask(withName: "SuspendApp: \(UUID().uuidString)") { [weak self] in
             guard let self else { return }
             
-            stopSync()
-        
-            backgroundTask = nil
+            stopSync { [weak self] in
+                guard let self else { return }
```
stefanceriu, line 521: I wonder if running this later might cause problems if quickly backgrounding and foregrounding the app, problems like `startSync` failing

Velin92:
I am actually wondering, do we really need this background task at all? Don't we just need to suspend the sync when the AppCoordinator sees that the app is suspending, and nothing more?

ElementX/Sources/Application/AppCoordinator.swift
Velin92: should have solved this in the last commit

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -161,7 +165,17 @@ class ClientProxy: ClientProxyProtocol {
...
         slidingSyncObserverToken = slidingSync?.sync()
     }
-    
+
+    func stopSync(completionHandler: () -> Void) {
+        guard let slidingSyncObserverToken else {
+            MXLog.info("No sync is present")
+            return
+        }
+        stopSync()
```
stefanceriu, line 174: Can we not merge these 2 stopSyncs into one? 🤔 If not we should at least make one of them private

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -152,6 +152,10 @@ class ClientProxy: ClientProxyProtocol {
             return nil
         }
     }
+
+    var isSyncing: Bool {
```
stefanceriu, line 156: Wouldn't it be cleaner if this check was done internal to the ClientProxy and it would result in a no-op if already syncing or already not syncing?

startSync already has a check for this

```
MXLog.info("Starting sync")
        guard slidingSyncObserverToken == nil else {
            return
        }
```

stefanceriu (changes requested):
Looks good to me but I would like some api changes that I commented on inline.

Separately from this, the usage of the background tasks in the roomProxy feels fishy to me. For example `sendMessage` starts a background task and then stops it on exiting the method. In reality the async call to the SDK doesn't return when the message was actually sent but when it was enqueued. I think that basically means that the background task here is useless. What do you think?

ElementX/Sources/Services/Client/ClientProxy.swift
Velin92: the problem is that the startSync() implementation inside the AppCoordinator, which has code that runs regardless of such state

ElementX/Sources/Services/Client/ClientProxy.swift
Velin92: the problem is that optional closures are automatically marked as @escaping, and since escaping closure are slightly slower than non escaping closure (and the expiration handler needs to complete ASAP), I don't want to create a function that uses an optional closure. It's a very minor optimisation honestly, but I would like to keep it just fort his specific bg case (I wrote a doc about this verison of the function) since we are still running a while loop which could take 1 second at most.

Velin92:
> Looks good to me but I would like some api changes that I commented on inline.
> 
> Separately from this, the usage of the background tasks in the roomProxy feels fishy to me. For example `sendMessage` starts a background task and then stops it on exiting the method. In reality the async call to the SDK doesn't return when the message was actually sent but when it was enqueued. I think that basically means that the background task here is useless. What do you think?

I actually tested it and the background task is actually needed,  I added an artificial  delay before the enqueueing of let's say 3 seconds before the sending of the message, without that bg task it will only get sent when I relaunch the app.

ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu: Okay, fair enough

ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu: Oh okay, got it 👍

stefanceriu:
> instead with it it gets sent in bg

Huh, interesting, okay then 👍

stefanceriu (approved):

ElementX/Sources/Services/Client/ClientProxy.swift
Velin92: but you know what, let me replace these slidingSyncObserverToken == nil internally with the isSyncing check, so it's a tad bit cleaner

PR #925, opened 2023-05-19
Flescio/create room action

flescio:
Fix a UI bug in the invites screen
Resolve https://github.com/vector-im/element-x-ios/issues/630
Add the creation action of a room. 
Add the task runner with timeout.
After a room is created it's not available right away, so it waits for the room id and room name to be present in the room summary list.

room-creation

### DIFF
--- a/ElementX/Sources/Other/Runner.swift
+++ b/ElementX/Sources/Other/Runner.swift
@@ -1,0 +1,53 @@
+//
+// Copyright 2023 New Vector Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+
+enum RunnerError: Error {
+    case timeout
+}
+
+actor Runner<T> {
+    private var continuation: CheckedContinuation<T, Error>?
+    
+    private var task: () async throws -> T
+    
+    init(_ task: @escaping () async throws -> T) {
+        self.task = task
+    }
+    
+    func run(timeout: TimeInterval) async throws -> T {
+        try await withCheckedThrowingContinuation {
+            continuation = $0
+            
+            Task {
+                try await Task.sleep(for: .seconds(timeout))
+                continuation?.resume(with: .failure(RunnerError.timeout))
+                continuation = nil
+            }
+            
+            Task {
+                do {
+                    let result = try await task()
+                    continuation?.resume(with: .success(result))
+                } catch {
+                    continuation?.resume(with: .failure(error))
+                }
+                continuation = nil
+            }
+        }
+    }
+}
--- a/ElementX/Sources/Screens/CreateRoom/CreateRoomViewModel.swift
+++ b/ElementX/Sources/Screens/CreateRoom/CreateRoomViewModel.swift
@@ -97,10 +97,11 @@ class CreateRoomViewModel: CreateRoomViewModelType, CreateRoomViewModelProtocol 
     }
     
     private func createRoom() async {
+        defer {
+            hideLoadingIndicator()
+        }
         showLoadingIndicator()
-        let result = await clientProxy.createRoom(with: createRoomParameters, users: state.selectedUsers)
-        hideLoadingIndicator()
-        switch result {
+        switch await clientProxy.createRoom(with: createRoomParameters, userIDs: state.selectedUsers.map(\.userID)) {
         case .success(let roomId):
             actionsSubject.send(.openRoom(withIdentifier: roomId))
         case .failure(let failure):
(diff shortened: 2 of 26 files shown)


ElementX/Sources/Screens/CreateRoom/CreateRoomViewModel.swift
```diff
@@ -71,4 +78,48 @@ class CreateRoomViewModel: CreateRoomViewModelType, CreateRoomViewModelProtocol
...
+    }
+    
+    private var clientProxy: ClientProxyProtocol {
+        userSession.clientProxy
+    }
+    
+    private func createRoom() async {
+        showLoadingIndicator()
+        let result = await clientProxy.createRoom(with: createRoomParameters, users: state.selectedUsers)
+        hideLoadingIndicator()
```
stefanceriu, line 102: This is minor but you could use a defer {} to hide the loading indicator and then switch directly on the `await clientProxy.createRoom`

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -192,17 +192,63 @@ class ClientProxy: ClientProxyProtocol {
             }
         }
     }
-
-    func createDirectRoom(with userID: String) async -> Result<String, ClientProxyError> {
-        await Task.dispatch(on: clientQueue) {
+    
+    func createDirectRoom(with userProfile: UserProfile) async -> Result<String, ClientProxyError> {
```
stefanceriu, line 196: I wonder if it would be better to just pass the `userID` and a `expectedRoomName` here so that it's clear that we don't need the full UserProfile
That way `createRoom(with parameters` won't need an array of UserProfiles either.

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -192,17 +192,63 @@ class ClientProxy: ClientProxyProtocol {
...
+    private func waitForRoomSummary(with result: Result<String, ClientProxyError>, name: String?) async -> Result<String, ClientProxyError> {
+        guard case .success(let roomId) = result else { return result }
+        let runner = Runner { [weak self] in
+            guard let roomLists = self?.visibleRoomsSummaryProvider?.roomListPublisher.values else {
+                return
+            }
+            // for every list of summaries, we check if we have a room summary with matching ID and name (if present)
+            for await roomList in roomLists {
+                guard let summary = roomList.first(where: { $0.id == roomId }) else { continue }
+                guard let name else { break }
```
stefanceriu, line 242: You should make this check even before running the "Runner",  no? Actually, this name shouldn't even be an optional at this point, the check should be done in the caller e.g. `createRoom(with parameters` and not even invoke waitForRoomSummary

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -192,17 +192,63 @@ class ClientProxy: ClientProxyProtocol {
...
+        
+        return await waitForRoomSummary(with: result, name: parameters.name)
+    }
+    
+    /// Await the room to be available in the room summary list
+    /// - Parameter result: the result of a room creation Task with the roomId
+    private func waitForRoomSummary(with result: Result<String, ClientProxyError>, name: String?) async -> Result<String, ClientProxyError> {
+        guard case .success(let roomId) = result else { return result }
+        let runner = Runner { [weak self] in
+            guard let roomLists = self?.visibleRoomsSummaryProvider?.roomListPublisher.values else {
```
stefanceriu, line 236: Wouldn't it be enough to just check the latest `value` instead of all `values?

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -192,17 +192,63 @@ class ClientProxy: ClientProxyProtocol {
...
+                return
+            }
+            // for every list of summaries, we check if we have a room summary with matching ID and name (if present)
+            for await roomList in roomLists {
+                guard let summary = roomList.first(where: { $0.id == roomId }) else { continue }
```
stefanceriu, line 241: Can we simplify everything in here to :?
```
if roomList.first(where: { $0.name == name }) != nil {
    break
}
```

ElementX/Sources/Other/Runner.swift
```diff
@@ -0,0 +1,53 @@
...
+// limitations under the License.
+//
+
+import Foundation
+
+enum RunnerError: Error {
+    case timeout
+}
+
+actor Runner<T> {
```
stefanceriu, line 23: Nifty solution 👍
Can we maybe rename it to something less generic though? I'm thinking `ExpiringTaskRunner`.

We also really **need** unit tests for this.

ElementX/Sources/Other/Runner.swift
```diff
@@ -0,0 +1,53 @@
...
+    private var continuation: CheckedContinuation<T, Error>?
+    
+    private var task: () async throws -> T
+    
+    init(_ task: @escaping () async throws -> T) {
+        self.task = task
+    }
+    
+    func run(timeout: TimeInterval) async throws -> T {
```
stefanceriu, line 32: Should we maybe directly use a `Duration` here instead of a `TimeInterval`?

stefanceriu (changes requested):
Looks good to me but I left some comments inline 👍

ElementX/Sources/Services/Client/ClientProxy.swift
flescio: We need to check the name on here, because if the name is nil, for any reason, we still need to wait the room to be present in the summary (after the creation is ok), because if not we can end up showing the home room list, without the freshly created room.

ElementX/Sources/Services/Client/ClientProxy.swift
flescio: sometimes we receive an array without the room or with the room but not fully loaded and a "Empty Room" name. We need to cycle the values of this publisher, Because it can be the first event, the second, or another one. As soon as we get it we break the loop.

ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu: Right, got it 👍

stefanceriu (approved):
🚀it!

PR #926, opened 2023-05-19
Add share link in start chat flow

alfogrillo:
This PR adds the share of the profile logic in the "start a chat" flow.

Fix https://github.com/vector-im/element-x-ios/issues/924

**Result**
share


Velin92 (approved):
LGTM

Issue #918, opened 2023-05-18
Introduce a FlowCoordinator protocol


stefanceriu:
We want FlowCoordinators to be special objects that will coordinate together multiple "screen" coordinators and the flows between them.
They will also usually have a state machine and be able to `handleAppRoute`s

Issue #919, opened 2023-05-18
Introduce more flow coordinators


stefanceriu:
With the application flows becoming more complex we should stop relying on screen coordinators from pushing other screens and instead extract that logic into dedicated flow coordinators.
Candidates are: the room coordinator, the create room coordinator and the room details coordinator

stefanceriu:
This is now done for the important flows.

PR #927, opened 2023-05-19
Flow coordinators

stefanceriu:
This PR is the first step towards introducing more flow coordinators, solidifying our navigation and fixing existing bugs, while making a whole class of potential problems easy to handle.

Fixes https://github.com/vector-im/element-x-ios/issues/918
Relates to https://github.com/vector-im/element-x-ios/issues/919

We've added a new RoomFlowCoordinator that is responsible for everything happening within a Room's context: 
* takes over some of the work that used to be done in the UserSessionFlowCoordinator, simplifying both
* works in tandem with the above to enable more complex flows like invitations
* has a state machine that correctly asserts the state between all the involved screens at all times (and clearly prints out what flows don't work)
* also simplifies the old RoomScreenCoordinator removing things like the media uploading and room details from it
* converges on a single place from where room details are presented, both from within the room header and from the room list contextual menu
* can handle app routes and can easily be used to implement deep linking into member details for example
* fixes bugs like 
    * the userSessionStateMachine current state often being wrong
    * not being able to select a room after opening its details from the room list contextual menu
    * not being able to a room's details from the room list contextual menu if that room is already opened
    * not rewidinging the navigation stack when re-selecting the room
    * and most likely others we haven't found yet

### DIFF
--- a/ElementX/Sources/Application/FlowCoordinatorProtocol.swift
+++ b/ElementX/Sources/Application/FlowCoordinatorProtocol.swift
@@ -1,0 +1,22 @@
+//
+// Copyright 2023 New Vector Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+
+@MainActor
+protocol FlowCoordinatorProtocol {
+    func handleAppRoute(_ appRoute: AppRoute, animated: Bool)
+}
--- a/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
@@ -1,0 +1,480 @@
+//
+// Copyright 2023 New Vector Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Combine
+import Foundation
+import SwiftState
+
+enum RoomFlowCoordinatorAction: Equatable {
+    case presentedRoom(String)
+    case dismissedRoom
+}
+
+class RoomFlowCoordinator: FlowCoordinatorProtocol {
+    private let userSession: UserSessionProtocol
+    private let roomTimelineControllerFactory: RoomTimelineControllerFactoryProtocol
+    private let navigationStackCoordinator: NavigationStackCoordinator
+    private let navigationSplitCoordinator: NavigationSplitCoordinator
+    private let emojiProvider: EmojiProviderProtocol
+    
+    private let stateMachine: StateMachine<State, Event> = .init(state: .initial)
+    
+    private var cancellables: Set<AnyCancellable> = .init()
+    
+    private let actionsSubject: PassthroughSubject<RoomFlowCoordinatorAction, Never> = .init()
+    var actions: AnyPublisher<RoomFlowCoordinatorAction, Never> {
+        actionsSubject.eraseToAnyPublisher()
+    }
+    
+    private var roomProxy: RoomProxyProtocol?
+    private var timelineController: RoomTimelineControllerProtocol?
+    
+    init(userSession: UserSessionProtocol,
+         roomTimelineControllerFactory: RoomTimelineControllerFactoryProtocol,
+         navigationStackCoordinator: NavigationStackCoordinator,
+         navigationSplitCoordinator: NavigationSplitCoordinator,
+         emojiProvider: EmojiProviderProtocol) {
+        self.userSession = userSession
+        self.roomTimelineControllerFactory = roomTimelineControllerFactory
+        self.navigationStackCoordinator = navigationStackCoordinator
+        self.navigationSplitCoordinator = navigationSplitCoordinator
+        self.emojiProvider = emojiProvider
+        
+        setupStateMachine()
+    }
+        
+    // MARK: - FlowCoordinatorProtocol
+    
+    func handleAppRoute(_ appRoute: AppRoute, animated: Bool) {
+        switch appRoute {
+        case .room(let roomID):
+            if case .room(let identifier) = stateMachine.state,
+               roomID == identifier {
+                return
+            }
+            
+            stateMachine.tryEvent(.presentRoom(roomID: roomID), userInfo: EventUserInfo(animated: animated))
+        case .roomDetails(let roomID):
+            stateMachine.tryEvent(.presentRoomDetails(roomID: roomID), userInfo: EventUserInfo(animated: animated))
+        case .roomList:
+            stateMachine.tryEvent(.dismissRoom, userInfo: EventUserInfo(animated: animated))
+        }
+    }
+    
+    // MARK: - Private
+    
+    // swiftlint:disable:next cyclomatic_complexity function_body_length
+    private func setupStateMachine() {
+        stateMachine.addRouteMapping { event, fromState, _ in
+            switch (event, fromState) {
+            case (.presentRoom(let roomID), _):
+                return .room(roomID: roomID)
+            case (.dismissRoom, .room):
+                return .initial
+                
+            case (.presentRoomDetails(let roomID), .initial):
+                return .roomDetails(roomID: roomID)
+            case (.presentRoomDetails(let roomID), .room):
+                return .roomDetails(roomID: roomID)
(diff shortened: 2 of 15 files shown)


ElementX/Sources/Application/FlowCoordinatorProtocol.swift
```diff
@@ -0,0 +1,22 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+
+@MainActor
+protocol FlowCoordinatorProtocol {
```
alfogrillo, line 20: It's a bit strange seeing this protocol not extending `CoordinatorProtocol`. I know these are different things, but conceptually it would be nice to have a common root protocol.

ElementX/Sources/Application/FlowCoordinatorProtocol.swift
```diff
@@ -0,0 +1,22 @@
...
+    func handleAppRoute(_ appRoute: AppRoute, animated: Bool)
```
alfogrillo, line 21: I don't expect in general any `FlowCoordinator` to be able to handle any `AppRoute`.

To me a better solution would be something like this:

```swift
@MainActor
protocol FlowCoordinatorProtocol {
    associatedtype Route
    
    func handleAppRoute(_ appRoute: Route, animated: Bool)
}
```

So that any flow coordinator can handle just the subset of routes he knows.

alfogrillo (commented):
Nice refactor! 🎉
Left some comments/ideas inline.

Also noticed the navigation to room's settings using the contextual menu changed.
- Before: the room's settings were pushed on the Home Screen
- Now: the is the room's timeline pushed in between

To me it looks still good, but let's check everyone is happy.

ElementX/Sources/Application/FlowCoordinatorProtocol.swift
stefanceriu: Well flow coordinators themselves can't (and shouldn't) be presented as they the ones that should be presenting Coordinators, on top of one of the root navigation components. I'm not sure if mixing them will bring any benefits.

ElementX/Sources/Application/FlowCoordinatorProtocol.swift
stefanceriu: I wasn't sure if we want to bother splitting up the Routes before passing them into each FlowCoordinator or whether each FlowCoordinator should just use whatever it can handle from one big route. I have no strong feelings either way, curious what the others think.

stefanceriu:
> let's check everyone is happy

Definitely, thanks for the review!

ElementX/Sources/Application/FlowCoordinatorProtocol.swift
Velin92: I feel that splitting the routes into specific subtypes may make the implementation way more difficult (and strain us away from using a simple enum type), not saying we shouldn't but maybe let's start simple, and see how it evolves

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -0,0 +1,332 @@
...
+            }
+        }
+        
+        stateMachine.addErrorHandler { context in
+            fatalError("Failed transition with context: \(context)")
+        }
+    }
+    
+    private func selectRoom(_ roomID: String, animated: Bool) {
+        Task { @MainActor in
```
pixlwave, line 127: Do you need the `@MainActor in` given the protocol is `@MainActor` and so the method will be too?

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -74,7 +93,7 @@ class UserSessionFlowCoordinator: CoordinatorProtocol {
         }
         switch appRoute {
         case .room(let roomID):
-            stateMachine.processEvent(.selectRoom(roomId: roomID), userInfo: .init(animated: animated))
+            roomFlowCoordinator.presentRoom(roomID: roomID, animated: animated)
```
pixlwave, line 96: Being particularly picky here, but these changes are making the assumption that the detail is always going to contain the `RoomFlowCoordinator`. Would it be cleaner for `roomFlowCoordinator` to be an optional that is nil when a room isn't selected, and then this creates one if needed before presenting it?

Appreciate that adds extra complexity though, just depends how flexible we want the navigation to be.

Also, I think this should probably pass the route down to the next flow coordinator for it to decide the next course of action, instead of telling it what to do.

ElementX/Sources/Screens/RoomScreen/RoomScreenCoordinator.swift
```diff
@@ -153,51 +111,10 @@ final class RoomScreenCoordinator: CoordinatorProtocol {
...
-                self?.showSuccess(label: L10n.commonReportSubmitted)
-            }
-        }
-        navigationCoordinator.setRootCoordinator(coordinator)
-        navigationStackCoordinator.setSheetCoordinator(userIndicatorController)
-    }
-
-    private func showSuccess(label: String) {
-        ServiceLocator.shared.userIndicatorController.submitIndicator(UserIndicator(title: label, iconName: "checkmark"))
+        navigationStackCoordinator.setSheetCoordinator(emojiPickerNavigationStackCoordinator)
```
pixlwave, line 118: How come the emoji picker is left here with the screen coordinator still having access to the stack? I would have expected this coordinator to lose all access to that stack so that e.g. navigating to a different room could dismiss the emoji picker as part of the room flow coordinator's responsibilities?

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -0,0 +1,332 @@
...
+            await selectRoom(roomID, animated: animated)
+        }
+    }
+    
+    // swiftlint:disable:next cyclomatic_complexity function_body_length
+    private func selectRoom(_ roomID: String, animated: Bool) async {
```
pixlwave, line 133: I'm curious. Is overloading a method as async (with the sync one calling the async one) a sensible idea? What happens if you call this inside an existing Task without the await, will it spawn a new task needlessly?

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -0,0 +1,332 @@
...
+        stateMachine.processEvent(.deselectRoom)
+    }
+    
+    func presentRoomDetails(roomID: String) {
+        stateMachine.processEvent(.presentRoomDetails(roomID: roomID))
+    }
+    
+    // MARK: - FlowCoordinatorProtocol
+    
+    func handleAppRoute(_ appRoute: AppRoute, animated: Bool) { }
```
pixlwave, line 76: This should have an implementation before we merge the PR really.

pixlwave (approved):
Overall seems like a decent change to me. We discussed offline about a possibility of only having `handleAppRoute` as the public method to encourage the use of `AppRoute`s as a single system for navigation throughout the app.

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
stefanceriu: Probably not, i think I just coped it from the original implementation 👍

ElementX/Sources/Screens/RoomScreen/RoomScreenCoordinator.swift
stefanceriu: Honestly, I forgot to move it 😅

ElementX/Sources/FlowCoordinators/RoomFlowCoordinatorStateMachine.swift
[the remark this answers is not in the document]
stefanceriu: The original idea was to keep them separate so we can unit test the state machine so let's see how unit testing the floow coordinators themselves go and if that's okay we can merge them

ElementX/Sources/FlowCoordinators/RoomFlowCoordinatorStateMachine.swift
```diff
@@ -0,0 +1,122 @@
...
+    // swiftlint:disable:next cyclomatic_complexity
+    private func configure() {
+        stateMachine.addRouteMapping { event, fromState, _ in
+            switch (event, fromState) {
+            case (.selectRoom(let roomID), .initial):
+                return .roomSelected(roomID: roomID)
+            case (.selectRoom(let roomID), .roomSelected):
+                return .roomSelected(roomID: roomID)
+            case (.selectRoom(let roomID), .roomDetails):
+                return .roomSelected(roomID: roomID)
```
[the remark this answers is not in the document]
stefanceriu: > via a notification when the app is currently presenting media, reporting content etc

Very good point 👍 

> this flow coordinator should be allowed to go from anything to anything else

I agree, I think this is what needs to happen.. eventually, might be a bit of overkill for now

ElementX/Sources/FlowCoordinators/RoomFlowCoordinatorStateMachine.swift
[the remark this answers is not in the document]
pixlwave: Sounds like a plan 👍

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
stefanceriu: Pretty much yeah 😅 I'm gonna keep it simple and call it `asyncSelectRoom`

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
stefanceriu: I understand where you're coming from but imho having it as an optional complicates things without any obvious advantage. It's a lot easier to reason about things if it's always there

PR #942, opened 2023-05-22
Trigger PR Build labelled PR to trigger a PR build.

michaelkaye:
XXX do not merge.


michaelkaye:
> Oh, unfortunately PR builds are currently broken because of a missing fastlane plugin. How soon do you need this to work again?

Oh, no worries. I was just coming back to this area after spending time with EC - getting a grip with where things are once more.

I'm going to be on vacation until the end of the month; it would be ace if i could pick this up after then?

PR #950, opened 2023-05-24
Fix UI tests

stefanceriu:
The UI tests were behaving weirdly when the timeline was involved. It would jump up every time we would want to find any element through the XCUIApp.
Turns out this was happening because the diffable datasource's cellProvider was invoked for every single possible accessibility identifier. 
To avoid that we're not completely disabling accessibility on the timeline's table view based on an envar set through UITests. We can do this on a per test basis as we're relaunching the app


ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
```diff
@@ -170,18 +167,14 @@ class TimelineTableViewController: UIViewController {
     override func viewWillLayoutSubviews() {
         super.viewWillLayoutSubviews()
         
-        guard tableView.frame.size != view.frame.size else { return }
-        tableView.frame = CGRect(origin: .zero, size: view.frame.size)
-        
-        // Update the table's layout if necessary after the frame changed.
-        updateTopPadding()
-        
-        guard composerMode == .default else { return }
```
stefanceriu, line 179: I wasn't sure about this one, why wouldn't we want to keep the timeline at the bottom if the composer mode is different? 🤔

ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
```diff
@@ -170,18 +167,14 @@ class TimelineTableViewController: UIViewController {
...
+        if tableView.frame.size != view.frame.size {
```
stefanceriu, line 170: I saw cases in which the subviews layout but the frame doesn't change. I think it's safer to just do the scrollToBottom even if the frame doesn't change

ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
```diff
@@ -87,8 +87,8 @@ class TimelineTableViewController: UIViewController {
     private let paginateBackwardsPublisher = PassthroughSubject<Void, Never>()
     /// Whether or not the ``timelineItems`` value should be applied when scrolling stops.
     private var hasPendingUpdates = false
-    /// Yucky hack to fix some layouts where the scroll view doesn't make it to the bottom on keyboard appearance.
-    private var keyboardWillShowLayout: LayoutDescriptor?
+    /// We need to store the previous layout as computing it on the fly leads to problems.
+    private var previousLayout: LayoutDescriptor?
```
stefanceriu, line 91: I saw the layout computed on the fly in willLayout report the wrong visible item index paths. It's safer to just store it, and we can reuse it for the keyboard too

ElementX/Sources/Other/Tests.swift
```diff
@@ -43,4 +43,16 @@ public enum Tests {
...
         #endif
     }
+    
+    static var shouldDisableTimelineAccessibility: Bool {
+        guard isRunningUITests else {
+            return false
+        }
+        
+        #if DEBUG
+        return ProcessInfo.processInfo.environment["UI_TESTS_DISABLE_TIMELINE_ACCESSIBILITY"] != nil
```
Velin92, line 53: this was breaking the scroll?!
gotta love apple

Velin92 (approved):
Approved however I also changed some UI tests of the room in my PR for the timestamp which is ready now

ElementX/Sources/Other/Tests.swift
stefanceriu: Yeah, what happens is that XCUIAccessibility queries all the possible cells the TimelineTableViewController can create, which in turn invokes the diffable dataSource's cellProvider block and change the layout of the timeline.
The fix for that is just to disable accessibility on the whole table view. This envar just allows us to do that on a per test basis

stefanceriu:
> I also changed some UI tests of the room in my PR

Alrighty, let's then wait for that to get merged first

pixlwave (approved):
LGTM, I'll see in a separate PR if we can now reuse some timeline snapshots to assert that it didn't move.

PR #952, opened 2023-05-24
Add room's update publisher

alfogrillo:
This PR adds a new publisher in the room's proxy to listen room's updates.
This publisher will be useful everywhere a scene needs fresh information from the room.
It also refactors `Room` and `RoomDetails` to handle live state events.
The PR also fixes a caching problem in the `LoadableAvatarImage` that prevented updates to be visible after a change in the URL.

### DIFF
--- a/ElementX/Sources/Other/SwiftUI/Views/LoadableAvatarImage.swift
+++ b/ElementX/Sources/Other/SwiftUI/Views/LoadableAvatarImage.swift
@@ -53,6 +53,9 @@ struct LoadableAvatarImage: View {
             } placeholder: {
                 PlaceholderAvatarImage(name: name, contentID: contentID)
             }
+            // Binds the lifecycle of the LoadableImage to the associated URL.
+            // This fixes the problem of the cache returning old values after a change in the URL.
+            .id(url)
         } else {
             PlaceholderAvatarImage(name: name, contentID: contentID)
         }
--- a/ElementX/Sources/Services/Room/RoomProxy.swift
+++ b/ElementX/Sources/Services/Room/RoomProxy.swift
@@ -41,6 +41,12 @@ class RoomProxy: RoomProxyProtocol {
     var membersPublisher: AnyPublisher<[RoomMemberProxyProtocol], Never> {
         membersSubject.eraseToAnyPublisher()
     }
+    
+    private var timelineListener: RoomTimelineListener?
+    private let updatesSubject = PassthroughSubject<TimelineDiff, Never>()
+    var updatesPublisher: AnyPublisher<MatrixRustSDK.TimelineDiff, Never> {
+        updatesSubject.eraseToAnyPublisher()
+    }
         
     init(slidingSyncRoom: SlidingSyncRoomProtocol,
          room: RoomProtocol,
@@ -139,12 +145,20 @@ class RoomProxy: RoomProxyProtocol {
         }
     }
         
-    func addTimelineListener(listener: TimelineListener) -> Result<[TimelineItem], RoomProxyError> {
+    func setupTimelineListenerIfNeeded() -> Result<[TimelineItem], RoomProxyError> {
+        guard timelineListener == nil else {
+            return .success([])
+        }
+        
         let settings = RoomSubscription(requiredState: [RequiredState(key: "m.room.topic", value: ""),
                                                         RequiredState(key: "m.room.canonical_alias", value: ""),
                                                         RequiredState(key: "m.room.join_rules", value: "")],
                                         timelineLimit: UInt32(SlidingSyncConstants.timelinePrecachingTimelineLimit))
         roomSubscriptionObservationToken = slidingSyncRoom.subscribeToRoom(settings: settings)
+        
+        let listener = RoomTimelineListener { [weak self] timelineDiff in
+            self?.updatesSubject.send(timelineDiff)
+        }
         
         if let result = try? slidingSyncRoom.addTimelineListener(listener: listener) {
             roomTimelineObservationToken = result.taskHandle
@@ -473,6 +487,13 @@ class RoomProxy: RoomProxyProtocol {
         UInt(room.activeMembersCount())
     }
     
+    deinit {
+        Task { @MainActor [roomTimelineObservationToken, slidingSyncRoom] in
+            roomTimelineObservationToken?.cancel()
+            _ = slidingSyncRoom.unsubscribeFromRoom()
+        }
+    }
+
     // MARK: - Private
     
     /// Force the timeline to load member details so it can populate sender profiles whenever we add a timeline listener
@@ -485,5 +506,17 @@ class RoomProxy: RoomProxyProtocol {
         
     private func update(displayName: String) {
         self.displayName = displayName
+    }
+}
+
+private class RoomTimelineListener: TimelineListener {
+    private let onUpdateClosure: (TimelineDiff) -> Void
+   
+    init(_ onUpdateClosure: @escaping (TimelineDiff) -> Void) {
+        self.onUpdateClosure = onUpdateClosure
+    }
+    
+    func onUpdate(update: TimelineDiff) {
+        onUpdateClosure(update)
     }
 }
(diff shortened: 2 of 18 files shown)


ElementX/Sources/Other/SwiftUI/Views/LoadableAvatarImage.swift
```diff
@@ -53,6 +53,9 @@ struct LoadableAvatarImage: View {
             } placeholder: {
                 PlaceholderAvatarImage(name: name, contentID: contentID)
             }
+            // Binds the lifecycle of the LoadableImage to the associated URL.
+            // This fixes the problem of the cache returning old values after a change in the URL.
+            .id(url)
```
stefanceriu, line 58: Nice but should we maybe do this directly in the LoadableImage? 🤔

ElementX/Sources/Services/Room/RoomProxy.swift
```diff
@@ -139,15 +145,25 @@ class RoomProxy: RoomProxyProtocol {
         }
     }
         
-    func addTimelineListener(listener: TimelineListener) -> Result<[TimelineItem], RoomProxyError> {
+    func registerTimelineListenerIfNeeded() -> Result<[TimelineItem]?, RoomProxyError> {
+        guard timelineListener == nil else {
+            return .success(nil)
```
stefanceriu, line 150: I would expect this to give you back the current timeline items, not nil. If we can't get our hands on hands on that perhaps we should throw an "already listening" error

ElementX/Sources/Services/Room/RoomProxy.swift
```diff
@@ -473,6 +480,13 @@ class RoomProxy: RoomProxyProtocol {
         UInt(room.activeMembersCount())
     }
     
+    deinit {
```
stefanceriu, line 483: Please move this above the constructor

ElementX/Sources/Services/Room/RoomProxy.swift
```diff
@@ -473,6 +480,13 @@ class RoomProxy: RoomProxyProtocol {
...
+        Task { @MainActor [roomTimelineObservationToken, slidingSyncRoom] in
+            roomTimelineObservationToken?.cancel()
+            _ = slidingSyncRoom.unsubscribeFromRoom()
```
stefanceriu, line 486: In practice this never works, subscriptions are never removed from the outgoing calls.

Can confirm that it works propery on develop

ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift
```diff
@@ -237,6 +238,8 @@ class RoomTimelineController: RoomTimelineControllerProtocol {
                     isBackPaginating = true
                 } else if timelineItem is TimelineStartRoomTimelineItem {
                     canBackPaginate = false
+                } else if timelineItem is StateRoomTimelineItem {
```
stefanceriu, line 241: Keep in mind that the timeline item factory can chose not to render some of the state events, resulting in the wrong value here. If product decides to not render room name changes in the UI this whole logic will break.

ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift
```diff
@@ -237,6 +238,8 @@ class RoomTimelineController: RoomTimelineControllerProtocol {
...
+                    hasNewStateEvents = true
                 }
```
stefanceriu, line 243: I'm not a particular big fan of this, the timeline controller is too close to the UI level. Ideally the fact that we have room name/topic/etc change should come directly from the RoomProxy through its own publisher. Not everybody has a TimelineController they can listen to for changes i.e. room details screen.

So how about this plan: 
- move the RoomTimelineProvider within the RoomProxy and expose it as a plain property so that the rest can still use it
- have the RoomProxy listen to RoomTimelineProvider updates as well and check if any of the `TimelineItemProxy` are `event(EventTimelineItemProxy)`s and their `isRoomState` is true 
- use that in the `updatesPublisher` to inform about state changes
- bonus points: have different types of state change updates based on TimelineItemProxy's  TimelineItemContentKind.state stateKey property

stefanceriu (changes requested):
Nice start but we need to iron out some details 👍

ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift
stefanceriu: Otherwise it would be great if we could get these things direcly from rust through a dedicated callback interface ¯\_(ツ)_/¯

ElementX/Sources/Other/SwiftUI/Views/LoadableAvatarImage.swift
alfogrillo: The issue is that parent views define identifiers for children.
A possible approach can be refactoring the `LoadableImage` using a private nested view:

- LoadableImage (which defines the id on `LoadableImageInner`)
    - LoadableImageInner (which owns the `ContentLoader`)
    
Since this is a bigger refactor it should be better considering a separate PR.
wdyt?

ElementX/Sources/Services/Room/RoomProxy.swift
alfogrillo: I'm not sure how to fix this. 
Should I store the `[TimelineItem]` array we get back when we add the listener and return always that?
I though that info may be out of sync.

Maybe throwing an error looks better?

ElementX/Sources/Services/Room/RoomProxy.swift
alfogrillo: How do I know if this works?
The problem is having the `unsubscription` being deallocated before the rust task ends?

stefanceriu: > How do I know if this works?

What I generally do is run Proxy man and see what the payload for SS is, it should have an "unsubscribe_rooms" in it when you unsubscribe.



> The problem is having the `unsubscription` being deallocated before the rust task ends?

I think so, yes. One solution would be to do a `while task.notFinished` to make sure it's done but I don't know how that plays with Task ran from deinit

ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift
alfogrillo: >  Ideally the fact that we have room name/topic/etc change should come directly from the RoomProxy through its own publisher.

This is exactly the purpose of `var updatesPublisher: AnyPublisher<TimelineDiff, Never>` in `RoomProxy`.
I had workaround this in the `RoomScreenViewModel` since it hasn't a reference to a `RoonProxy`.
How bad is having one here?

ElementX/Sources/Services/Room/RoomProxy.swift
alfogrillo: > I think so, yes. One solution would be to do a while task.notFinished to make sure it's done but I don't know how that plays with Task ran from deinit

It should be fine, trying it...

ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift
stefanceriu: Not bad at all 👍

ElementX/Sources/Services/Room/RoomProxy.swift
stefanceriu: I think so too, yeah

ElementX/Sources/Other/SwiftUI/Views/LoadableAvatarImage.swift
alfogrillo: Discussed off line

ElementX/Sources/Services/Room/RoomProxy.swift
alfogrillo: introducing a new error

ElementX/Sources/Services/Room/RoomProxy.swift
alfogrillo: Added logic to wait in the deinit

ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift
alfogrillo: discussed offline

stefanceriu (approved):
Looks great 👏

Issue #720, opened 2023-03-22
Update room properties from room's detail


alfogrillo:

stefanceriu:
Please use the `MediaPickerCoordinator` introduced in https://github.com/vector-im/element-x-ios/pull/749

PR #956, opened 2023-05-25
Room's details edit screen

alfogrillo:
This PR adds a new screen to edit room's details:
- Name 
- Topic
- Avatar

Fixes https://github.com/vector-im/element-x-ios/issues/720

**Result**
edit

### DIFF
--- a/ElementX/Sources/Other/SwiftUI/Views/LoadableEditableAvatarImage.swift
+++ b/ElementX/Sources/Other/SwiftUI/Views/LoadableEditableAvatarImage.swift
@@ -1,0 +1,59 @@
+//
+// Copyright 2023 New Vector Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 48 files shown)


ElementX/Sources/Other/SwiftUI/Views/LoadableEditableAvatarImage.swift
```diff
@@ -0,0 +1,59 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import SwiftUI
+
+struct LoadableEditableAvatarImage: View {
```
stefanceriu, line 19: This name makes me think that this view actually has something to do with the editing process. Perhaps we can use `LocallyOverloadableAvatarImage`? Might a be bit cheeky.

stefanceriu (approved):
Looks good to me! 👍

ElementX/Sources/Other/SwiftUI/Views/LoadableEditableAvatarImage.swift
alfogrillo: Changed to ` OverridableAvatarImage`. The alternative URL actually may also be another remote url. 👍🏻

PR #959, opened 2023-05-25
Server Confirmation Screen

pixlwave:
Implements the server confirmation screen, removing the server selection field on the login screen in favour of setting the server up front.

Not handled in this PR:
- Showing the OIDC webview directly from the server confirmation screen.
- Adding a state machine to the `AuthenticationCoordinator`.
- Matching the layout of the bottom buttons to the OnboardingScreen (I'd like to tidy that up first and then make a shared layout component).
- Register flow (might not make sense until OIDC is more commonplace).

Integration tests run is here: https://github.com/vector-im/element-x-ios/actions/runs/5093122043

| Default Server | Changing Server |
| - | - |
| Simulator Screen Recording - iPhone 14 Pro - 2023-05-26 at 17 39 44 | Simulator Screen Recording - iPhone 14 Pro - 2023-05-26 at 17 40 40 |


ElementX/Sources/Other/DidSetPublisher.swift
```diff
@@ -0,0 +1,30 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Combine
+
+/// Creates a regular property that provides a publisher via it's projected value.
+/// Similar to using `@Published` but works outside of an `ObservableObject` and the
+/// publisher is fired on `didSet` instead of `willSet`.
+@propertyWrapper struct DidSetPublisher<Value> {
```
stefanceriu, line 22: How we use this seems very similar to to how we use a currentValueSubject + currentValuePublisher but I believe the latter has an advantage as we don't need to expose 2 public properties when we need the current value to be read from the outside e.g.

```
private(set) var homeserver
var homeserverPublisher
```

stefanceriu (approved):
Looks good to me apart from the `DidSetPublisher` which we should talk about.

ElementX/Sources/Other/DidSetPublisher.swift
pixlwave: Discussed offline, removed for now and will make a separate PR with just this change.

PR #961, opened 2023-05-26
Flescio/add room image

flescio:
Resolve https://github.com/vector-im/element-x-ios/issues/664 and https://github.com/vector-im/element-x-ios/issues/665

Add the image picker flow for the creation of a room, the image is transformed into a MediaInfo for having the thumbnail and infos, the thumbnail is presented in the UI, the user has the liberty of removing it or change it.

When the user press the create button, the image will be uploaded and then the room will be created.

create room

### DIFF
--- a/ElementX/Sources/Screens/CreateRoom/View/CreateRoomScreen.swift
+++ b/ElementX/Sources/Screens/CreateRoom/View/CreateRoomScreen.swift
@@ -18,6 +18,8 @@ import SwiftUI
 
 struct CreateRoomScreen: View {
     @ObservedObject var context: CreateRoomViewModel.Context
+    @State private var showAttachmentPopover = false
+    @State private var sheetContentHeight = CGFloat(0)
     
     var body: some View {
         mainContent
@@ -32,6 +34,25 @@ struct CreateRoomScreen: View {
                 }
             }
             .background(ViewFrameReader(frame: $frame))
+            .confirmationDialog("", isPresented: $showAttachmentPopover, actions: {
+                Button {
+                    context.send(viewAction: .displayCameraPicker)
+                } label: {
+                    Text(L10n.actionTakePhoto)
+                }
+                Button {
+                    context.send(viewAction: .displayMediaPicker)
+                } label: {
+                    Text(L10n.actionChoosePhoto)
+                }
+                if context.viewState.roomImage != nil {
+                    Button(role: .destructive) {
+                        context.send(viewAction: .removeImage)
+                    } label: {
+                        Text(L10n.actionRemove)
+                    }
+                }
+            })
     }
     
     /// The main content of the view to be shown in a scroll view.
@@ -49,11 +70,22 @@ struct CreateRoomScreen: View {
     private var roomSection: some View {
         Section {
             HStack(alignment: .center, spacing: 16) {
-                Image(systemName: "camera")
-                    .foregroundColor(.element.secondaryContent)
-                    .frame(width: roomIconSize, height: roomIconSize)
-                    .background(Color.element.quinaryContent)
-                    .clipShape(Circle())
+                Button {
+                    showAttachmentPopover = true
+                } label: {
+                    if let data = context.viewState.roomImage, let image = UIImage(data: data) {
+                        Image(uiImage: image)
+                            .resizable()
+                            .frame(width: roomIconSize, height: roomIconSize)
+                            .clipShape(Circle())
+                    } else {
+                        Image(systemName: "camera")
+                            .foregroundColor(.element.secondaryContent)
+                            .frame(width: roomIconSize, height: roomIconSize)
+                            .background(Color.element.quinaryContent)
+                            .clipShape(Circle())
+                    }
+                }
                 VStack(alignment: .leading, spacing: 8) {
                     Text(L10n.screenCreateRoomRoomNameLabel.uppercased())
                         .font(.compound.bodyXS)
--- a/ElementX/Sources/Screens/CreateRoom/View/CreateRoomScreen.swift
+++ b/ElementX/Sources/Screens/CreateRoom/View/CreateRoomScreen.swift
@@ -93,6 +93,7 @@ struct CreateRoomScreen: View {
                         cameraImage
                     }
                 }
+                .buttonStyle(.plain)
                 VStack(alignment: .leading, spacing: 8) {
                     Text(L10n.screenCreateRoomRoomNameLabel.uppercased())
                         .font(.compound.bodyXS)
--- a/ElementX/Sources/Screens/CreateRoom/View/CreateRoomScreen.swift
+++ b/ElementX/Sources/Screens/CreateRoom/View/CreateRoomScreen.swift
@@ -18,11 +18,9 @@ import SwiftUI
 
 struct CreateRoomScreen: View {
     @ObservedObject var context: CreateRoomViewModel.Context
-    @State private var showAttachmentPopover = false
-    @State private var sheetContentHeight = CGFloat(0)
     @FocusState private var focus: Focus?
 
-    enum Focus {
+    private enum Focus {
         case name
         case topic
     }
@@ -40,6 +38,7 @@ struct CreateRoomScreen: View {
                 }
             }
             .background(ViewFrameReader(frame: $frame))
+            .alert(item: $context.alertInfo) { $0.alert }
     }
     
     /// The main content of the view to be shown in a scroll view.
@@ -58,7 +57,7 @@ struct CreateRoomScreen: View {
         Section {
             HStack(alignment: .center, spacing: 16) {
                 Button {
-                    showAttachmentPopover = true
+                    context.showAttachmentConfirmationDialog = true
                 } label: {
                     if let url = context.viewState.roomImage {
                         AsyncImage(url: url) { image in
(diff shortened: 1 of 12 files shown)


ElementX/Sources/Screens/CreateRoom/View/CreateRoomScreen.swift
```diff
@@ -49,11 +57,43 @@ struct CreateRoomScreen: View {
...
+                            cameraImage
+                        }
+                        .frame(width: roomIconSize, height: roomIconSize)
+                        .clipShape(Circle())
+                    } else {
+                        cameraImage
+                    }
+                }
+                .buttonStyle(.plain)
+                .confirmationDialog("", isPresented: $showAttachmentPopover, actions: {
```
pixlwave, line 78: Same comment on preferring a trailing closure.

ElementX/Sources/Screens/CreateRoom/View/CreateRoomScreen.swift
```diff
@@ -49,11 +57,43 @@ struct CreateRoomScreen: View {
...
+                    } label: {
+                        Text(L10n.actionChoosePhoto)
+                    }
+                    if context.viewState.roomImage != nil {
+                        Button(role: .destructive) {
+                            context.send(viewAction: .removeImage)
+                        } label: {
+                            Text(L10n.actionRemove)
+                        }
+                    }
```
pixlwave, line 95: As the label is simply a text that is styled by the system it would be a bit cleaner to simplify these to
```
Button(L10n.someString) {
    /// do the action
}
```

ElementX/Sources/Screens/CreateRoom/View/CreateRoomScreen.swift
```diff
@@ -49,11 +57,43 @@ struct CreateRoomScreen: View {
     private var roomSection: some View {
         Section {
             HStack(alignment: .center, spacing: 16) {
-                Image(systemName: "camera")
-                    .foregroundColor(.element.secondaryContent)
-                    .frame(width: roomIconSize, height: roomIconSize)
-                    .background(Color.element.quinaryContent)
-                    .clipShape(Circle())
+                Button {
+                    showAttachmentPopover = true
```
pixlwave, line 61: Picky naming here but as this is only a popover on iPad it would be more accurate to call this `showAttachmentConfirmationDialog`.

ElementX/Sources/Services/Client/ClientError.swift
```diff
@@ -21,6 +21,19 @@ enum MatrixErrorCode: String, CaseIterable {
     case unknown = "M_UNKNOWN"
     case userDeactivated = "M_USER_DEACTIVATED"
     case forbidden = "M_FORBIDDEN"
+    case fileTooLarge = "M_TOO_LARGE"
+}
+
+extension ClientError {
+    var code: MatrixErrorCode {
```
pixlwave, line 28: I'm really not a fan of expanding this silly hack any further. Would it be within the scope of your work to add a `FileTooLarge` field on `ClientError` and throw this in the `uploadMedia` method instead of more "look in the error string and hope for the best" solutions?

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -197,25 +197,40 @@ class ClientProxy: ClientProxyProtocol {
...
-        return await waitForRoomSummary(with: result, name: parameters.name)
+        return await waitForRoomSummary(with: result, name: name)
+    }
+    
+    func uploadMedia(_ media: MediaInfo) async -> Result<String, ClientProxyError> {
+        guard let mimeType = media.mimeType else { return .failure(ClientProxyError.mediaFileError) }
+        return await Task.dispatch(on: clientQueue) {
+            do {
+                let data = try Data(contentsOf: media.mainURL)
+                let matrixUrl = try self.client.uploadMedia(mimeType: mimeType, data: [UInt8](data))
```
pixlwave, line 226:
```suggestion
                let matrixURL = try self.client.uploadMedia(mimeType: mimeType, data: [UInt8](data))
```
Sidenote, shouldn't this accept the original file's URL and let Rust do whatever it wants with the file rather than having us load the whole thing into memory on the app side? Not sure how related this binding is to your PR.

pixlwave (commented):
Nice addition 📸

ElementX/Sources/Screens/StartChatScreen/StartChatScreenCoordinator.swift
```diff
@@ -99,35 +107,75 @@ final class StartChatScreenCoordinator: CoordinatorProtocol {
...
+                self.processUrl(url)
+            }
+        }
+        
+        stackCoordinator.setRootCoordinator(mediaPickerCoordinator)
+        
+        navigationStackCoordinator?.setSheetCoordinator(userIndicatorController)
+    }
+    
+    private func processUrl(_ url: URL) {
```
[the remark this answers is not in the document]
flescio: it won't upload anything yet, it's only pre-processing the image into a MediaInfo object with thumbnail

ElementX/Sources/Services/Client/ClientProxyProtocol.swift
```diff
@@ -92,7 +94,9 @@ protocol ClientProxyProtocol: AnyObject, MediaLoaderProtocol {
     
     func createDirectRoom(with userID: String, expectedRoomName: String?) async -> Result<String, ClientProxyError>
     
-    func createRoom(with parameters: CreateRoomFlowParameters, userIDs: [String]) async -> Result<String, ClientProxyError>
+    func createRoom(name: String, topic: String?, isRoomPrivate: Bool, userIDs: [String], roomImageMatrixUrl: String?) async -> Result<String, ClientProxyError>
```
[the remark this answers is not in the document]
flescio: the uploadMedia function is returning a String, should I cast it into an URL, and then in here cast it back into a String, or should I change the name like "avatarPath" instead?

ElementX/Sources/Screens/StartChatScreen/StartChatScreenCoordinator.swift
```diff
@@ -45,6 +45,14 @@ final class StartChatScreenCoordinator: CoordinatorProtocol {
         selectedUsers.asCurrentValuePublisher()
     }
     
+    private var navigationStackCoordinator: NavigationStackCoordinator? {
+        parameters.navigationStackCoordinator
+    }
+    
+    private var userIndicatorController: UserIndicatorControllerProtocol? {
+        parameters.userIndicatorController
+    }
```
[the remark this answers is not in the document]
flescio: well the navigationStackCoordinator it's ok, but the userIndicatorController is a weak var reference in the parameters, so should be optional right? Or it's better to have a failable init, with a guard let for that?

ElementX/Sources/Services/Client/ClientProxy.swift
flescio: I think you are right but It seems out of the scope of this PR, If you prefer I can create an issue about it so we won't lose track of it. Also probably the media upload Rust API will change anyways with the upload progress right?

ElementX/Sources/Services/Client/ClientError.swift
flescio: I wanted to avoid having too many errors in ClientProxyError enum: let's say we'll have file too large, or format not supported or anything else. 
If this is growing, those cases shall all fall onto the failedUploadingMedia(MatrixErrorCode)

ElementX/Sources/Screens/StartChatScreen/StartChatScreenCoordinator.swift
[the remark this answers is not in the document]
pixlwave: Ah no, it's weak that explains it, I didn't look close enough 👍

ElementX/Sources/Services/Client/ClientError.swift
pixlwave: Sorry I specifically meant the `ClientError` coming from Rust. We should really start to type these errors rather than looking for things in the `Generic` field's message.

ElementX/Sources/Services/Client/ClientProxy.swift
pixlwave: Yep yep, that would be great if you would 👍

ElementX/Sources/Screens/StartChatScreen/StartChatScreenCoordinator.swift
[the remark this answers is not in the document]
pixlwave: Ahhh right. Then maybe `processAvatar(from url: URL)`?

ElementX/Sources/Services/Client/ClientError.swift
alfogrillo: @flescio if we aren't showing any custom error for the `FileTooLarge` thing, we may revert this logic for now.

ElementX/Sources/Services/Client/ClientProxyProtocol.swift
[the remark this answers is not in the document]
pixlwave: Sorry I missed this question earlier. Yep, create a URL from it to pass around in Swift, and then when you eventually pass it back use the string value again.

ElementX/Sources/Services/Client/ClientError.swift
alfogrillo: > @flescio if we aren't showing any custom error for the FileTooLarge thing, we may revert this logic for now.

Discussed offline. We still need to distinguish the error to show a custom copy. 
To me it would be good to raise a separate PR later on Rust to fix how errors are exposed in UniFFI. 
This is a more general problem not directly related to this PR imho.

PR #965, opened 2023-05-26
Showing the iOS default silhoutte when the user/room avatar is missing

Velin92:
IMG_0175
IMG_66D166C0BFB0-1


ElementX/Sources/Other/Extensions/UNNotificationContent.swift
```diff
@@ -123,7 +123,7 @@ extension UNMutableNotificationContent {
                        senderID: String,
                        senderName: String,
                        iconType: NotificationIconType) async throws -> UNMutableNotificationContent {
-        var image: INImage?
+        var image = INImage(named: "")
```
Velin92, line 126: there is also `INImage()` but instead of generating the default image for a missing image, instead it's like setting the image to nil, and not having any image at all.

PR #969, opened 2023-05-26
Logging the failed device token registration

Velin92:


ElementX/Sources/Services/Notification/Manager/NotificationManager.swift
```diff
@@ -86,7 +86,9 @@ class NotificationManager: NSObject, NotificationManagerProtocol {
         }
     }
 
-    func registrationFailed(with error: Error) { }
+    func registrationFailed(with error: Error) {
+        MXLog.info("[NotificationManager] device token registration failed with error: \(error)")
```
pixlwave, line 90:
```suggestion
        MXLog.error("[NotificationManager] device token registration failed with error: \(error)")
```

pixlwave (approved):
Do we know if this error will contain any PII?

Velin92:
> Do we know if this error will contain any PII?

https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622962-application

Is not specified but I imagine that this means that probably it doesn't

PR #981, opened 2023-05-29
Add indicator after leaving a room

alfogrillo:
This PR adds a confirmation after leaving a room from the room list.

**Result**
result

### DIFF
--- a/ElementX/Sources/Other/UserIndicator/UserIndicator.swift
+++ b/ElementX/Sources/Other/UserIndicator/UserIndicator.swift
@@ -14,6 +14,7 @@
 // limitations under the License.
 //
 
+import Combine
 import Foundation
 
 enum UserIndicatorType {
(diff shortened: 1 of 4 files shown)


alfogrillo:
@callumu the UI looks different compared with [this](https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X?type=design&node-id=1386%3A82060&t=7P2Fl48uuexeQFn0-1) since I reused the indicator we already had in the app.
I guess is this worth to take an app wide decision in a next PR?

ElementX/Sources/Other/UserIndicator/UserIndicator.swift
```diff
@@ -30,10 +31,24 @@ struct UserIndicator: Equatable, Identifiable {
...
     var persistent = false
-    var progressPublisher: ProgressPublisher?
+    var loaderType: LoaderType? = .unknownProgress
+    
+    var progressPublisher: AnyPublisher<Double, Never> {
+        switch loaderType {
+        case .none, .unknownProgress:
+            return Empty().eraseToAnyPublisher()
+        case .some(.progress(let publisher)):
+            return publisher.publisher.eraseToAnyPublisher()
```
Velin92, line 51: publisher.publisher feels a bit weird can we have the variable be called differently?

Velin92 (approved):

ElementX/Sources/Other/UserIndicator/UserIndicator.swift
alfogrillo: Applied!

callumu:
This looks good 👍

PR #982, opened 2023-05-29
Generating SDK Mocks through the usage of `swift run tools generate-sdk-mocks`

Velin92:
You need to specify the version for the mocks you want to generate.
For example the current mocks are generated using 1.0.69-alpha.
`swift run tools generate-sdk-mocks 1.0.69-alpha`

However if you like to generate mocks for your local implementation, you just ned to use:
`swift run tools generate-sdk-mocks local`

For now the only mock that we are generating is the one for the `ClientProtocol` however we can add them for as many protocols as we want, remember that they need to be placesd inside the Sources/Mocks/SDK folder to be parsed by the tool.

These mock generation tool is separated from the internal project default one used each time the app is run.


stefanceriu (approved):
Veeery very nice! 👏

PR #984, opened 2023-05-30
Composer mode UI updates

stefanceriu:
- Slightly tweak the editing composer mode UI
- Update reply composer mode UI to include message being replied to

Screenshot 2023-05-30 at 10 43 11

https://github.com/vector-im/element-x-ios/assets/637564/c770d11a-3b66-4522-a482-ad84742f0f5a

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/View/MessageComposer.swift
+++ b/ElementX/Sources/Screens/RoomScreen/View/MessageComposer.swift
@@ -20,7 +20,7 @@ struct MessageComposer: View {
     @Binding var text: String
     @Binding var focused: Bool
     let sendingDisabled: Bool
-    let type: RoomScreenComposerMode
+    let mode: RoomScreenComposerMode
     
     let sendAction: EnterKeyHandler
     let pasteAction: PasteHandler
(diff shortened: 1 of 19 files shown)


callumu:
Everything in the video looks great imo

stefanceriu:
> Can the size of the icons within the message replies be reduced to 16x16?

@callumu Sure, but these views are shared between the composer reply mode and the timeline reply bubbles. Should I change both?

stefanceriu:
> can it be 20x20 instead of what I previously said?

Of course, thanks 👍

stefanceriu:
> One more small thing

Is this what you had in mind? The composers uses 9pts and the timeline 4.

Screenshot 2023-05-30 at 16 20 27

ElementX/Sources/Screens/RoomScreen/View/MessageComposer.swift
```diff
@@ -209,22 +215,47 @@ struct MessageComposer_Previews: PreviewProvider {
...
+                            mode: .edit(originalItemId: UUID().uuidString),
                             sendAction: { },
                             pasteAction: { _ in },
                             replyCancellationAction: { },
                             editCancellationAction: { })
         }
         .padding(.horizontal)
+        
+        ScrollView {
+            VStack {
```
Velin92, line 227: since this contains a for each loop wouldn't be better to use a LazyVStack ?

Velin92 (approved):

ElementX/Sources/Screens/RoomScreen/View/MessageComposer.swift
stefanceriu: I guess, but it's just a preview so who cares? 😁

Velin92: oh yeah mb, that's preview so it's statically defined.

PR #988, opened 2023-05-30
Fix navigation back from room's details

alfogrillo:
This PR fixes the navigation back to the room list when the room's details have been pushed using the room's **context menu**.

**Currently**: the back from room's details goes into the room's timeline
**Expected**: the back from room's details goes to the room's list 

**iPhone**
Simulator Screen Recording - iPhone 14 Pro - 2023-05-31 at 09 56 35

**iPad**
ipad

### DIFF
--- a/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
@@ -311,8 +311,6 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
         let coordinator = RoomDetailsScreenCoordinator(parameters: params)
         coordinator.callback = { [weak self] action in
             switch action {
-            case .cancel:
-                self?.navigationStackCoordinator.pop()
             case .leftRoom:
                 self?.dismissRoom(animated: animated)
             }
--- a/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
@@ -317,7 +317,6 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
         }
         
         if isRoot {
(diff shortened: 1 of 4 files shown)


stefanceriu:
@amshakal @VolkerJunginger can we get some product input on this please?

alfogrillo:
> @amshakal @VolkerJunginger can we get some product input on this please?

@callumu already confirmed we want this behavior. 
But other opinions are welcome. :-)

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -275,20 +278,30 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
...
         actionsSubject.send(.dismissedRoom)
     }
     
-    private func presentRoomDetails(roomID: String, animated: Bool) async {
-        if roomProxy?.id != roomID {
-            await asyncPresentRoom(roomID, animated: true)
+    private func presentRoomDetails(roomID: String, isRoot: Bool, animated: Bool) {
+        Task {
+            await asyncPresentRoomDetails(roomID: roomID, isRoot: isRoot, animated: animated)
+        }
```
Velin92, line 291: is it possible to trigger another navigation during the await? Like if I tap very fast on another room while loading another one?

Velin92 (approved):

stefanceriu (approved):
Nice! 👌

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
alfogrillo: I think we are preventing to show a UI feedback since the UX would be ugly (imagine a loader each time you tap on a room/open the room details).
In practice this doesn't seem to be a real problem since getting back the room proxy from the rust sdk is very very fast (measured around  `0.002` seconds).

PR #994, opened 2023-05-31
Authentication Layouts

pixlwave:
- Remove a lot of old unused code from the onboarding screen as we don't have a carousel anymore.
- Adds a `FullscreenDialog` container used for layouts of screens that have content with buttons at the bottom.
    - This layout no longer scrolls when the content fits on the screen.
    - When the content doesn't fit on the screen, the buttons now move out the way instead of being an overlay so the entire screen scrolls.
- Uses `FullscreenDialog` for the analytics prompt and server confirmation screen.
- Fix a bug in the analytics prompt where the 2 points with HTML weren't scaling due to having fixed fonts set by DTCoreText.
- Bumps the minimum OS to 16.4 for the scroll fix.

The original plan was to also use the new layout in the onboarding screen however, it turns out it didn't really make sense as that one doesn't need to scroll with accessibility text sizes, instead the logo can simply get smaller.

https://github.com/vector-im/element-x-ios/assets/6060466/e7955e86-b556-4098-bd51-759478f66497


UITests/Sources/__Snapshots__/Application/pseudo-iPhone-14.analyticsPrompt.png
Velin92, line 1: uhm... maybe this should be scrollable to not overlap with the bottom of the screen if the text is too long?

Velin92 (approved):
LGTM, just one small thing, and maybe worth asking to the design team for the scrolling behaviour of the view, because I feel the overlapping with the bottom indicator is a bit too evident for buttons maybe?

PR #998, opened 2023-05-31
Sliding sync hang fix

stefanceriu:
This PR fixes sliding sync not restarting properly when coming back from an app suspended mode. When that happens scrolling the list results in more range requests that the internal channel can support and blocks the main thread indefinitely

### DIFF
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -436,19 +436,22 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
     
     private func observeNetworkState() {
         let reachabilityNotificationIdentifier = "io.element.elementx.reachability.notification"
-        networkMonitorObserver = ServiceLocator.shared.networkMonitor.reachabilityPublisher.sink { reachable in
-            MXLog.info("Reachability changed to \(reachable)")
-            
-            if reachable {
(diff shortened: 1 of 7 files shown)


ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -436,19 +436,22 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
...
-            
-            if reachable {
-                ServiceLocator.shared.userIndicatorController.retractIndicatorWithId(reachabilityNotificationIdentifier)
-            } else {
-                ServiceLocator.shared.userIndicatorController.submitIndicator(.init(id: reachabilityNotificationIdentifier,
-                                                                                    title: L10n.commonOffline,
-                                                                                    persistent: true))
+        networkMonitorObserver = ServiceLocator.shared.networkMonitor
+            .reachabilityPublisher
+            .removeDuplicates()
```
pixlwave, line 441: Good spot

pixlwave (approved):
LGTM

ElementX/Sources/Application/AppCoordinator.swift
stefanceriu: Didn't really affect anything, just a bunch of useless logs

ara4n:
would this explain https://github.com/vector-im/element-x-ios-rageshakes/issues/389?

ara4n:
answer: nope, as in that rageshake the app UI hasn't locked up.

PR #1005, opened 2023-06-01
Create & Join rooms UI tweaks

alfogrillo:
This PR adjusts a lot of small UI things related with creation of rooms and invites flows.

Changes of behavior are:
- Show/Hide bottom toolbar in the `HomeScreen`
- The invites list now show the mxid of the inviters

**TODO**: update snapshot tests ref. screenshots


ElementX/Sources/Other/SwiftUI/Form Styles/FormButtonStyles.swift
```diff
@@ -20,7 +20,7 @@ import SwiftUI
 enum FormRowAccessory: View {
     case navigationLink
     case progressView
-    case selection(isSelected: Bool)
+    case selection(isSelected: Bool, isDisabled: Bool = false)
```
pixlwave, line 23: Couldn't we just use `isEnabled` from the environment instead of adding an additional parameter? I would imagine that would work just as well here.

alfogrillo: Great suggestion!
But I guess I need to refactor `FormRowAccessory` to a struct to use the `@Environment` property wrapper.
Are we still happy with it?
(current `case`s will be refactored as static func var/func I guess)

alfogrillo: Refactored [here](https://github.com/vector-im/element-x-ios/pull/1005/commits/71720fd31c01fe91131c35e564b981f66df9ed8a).
LGTM! 
wdyt?

pixlwave: Yep yep, that works for me, not sure I would have thought about doing that 👏
Could we go one step further and remove `isDisabled` as a parameter on the style, instead preferring to call `.disabled` on the button itself. That feels much more SwiftUIy to me.

Also, aren't you meant to be on a national holiday?! 🙈

pixlwave: Haha, thanks for keeping me company 😁

I like it, that's nice and clean now 👌

stefanceriu (approved):
Looks good to me 👍

PR #1006, opened 2023-06-02
Update PostHog to fix Accent Colour bug.

pixlwave:
Since enabling PostHog, there was an issue for users who had opted into Analytics where their accent colour was broken. This was [addressed](https://github.com/PostHog/posthog-ios/pull/55) by PostHog so this PR updated the package to fix the colour.

| Before | After |
| - | - |
| IMG_4D84BB75FF73-1 | Simulator Screenshot - iPhone 14 Pro - 2023-06-02 at 09 11 11 |


aringenbach (approved):
LGTM

PR #1009, opened 2023-06-02
Backport base Form style from Compound.

pixlwave:
I had thought we'd be using the `.compoundForm()` style by now, but as we aren't this PR makes it easier than having to manually apply a colour.


aringenbach (approved):
LGTM

PR #1010, opened 2023-06-02
Fix authentication UI tests.

pixlwave:
The flow tests weren't updated after making OIDC present from the confirmation screen.
The analytics prompt snapshot wasn't updated after reverting a change to the layout for regular text sizes.


aringenbach (approved):
LGTM !

PR #1018, opened 2023-06-05
Improved bug report error handling

pixlwave:
This PR improves bug report uploading by
- Checking for a 200 response before decoding the response body.
- Showing the error from the server if the upload fails so its clearer what went wrong (as we can't get logs to find out)
- Chunking the log files every 10MB (before compression).
- Adding a max upload size and only adding files to the upload that don't exceed the limit.


stefanceriu (approved):
lgtm 👍

PR #1019, opened 2023-06-05
Update compound-ios

pixlwave:
Includes new `CompoundIcon` component.


Velin92 (approved):
LGTM, maybe worth running UI tests and see if there are any changes that require an update?

PR #1038, opened 2023-06-06
Read Receipts display with the user avatar

Velin92:
The Read Receipts coming from the SDK do not account for all possible hidden events, also a bug seems to duplicate the read receipts for some users, however generally the UI to show the avatar is completed, and hidden behind a feature flag so that we can release it only when we will be read, here are some screens:

Simulator Screenshot - iPhone 14 - 2023-06-07 at 17 26 34
Simulator Screenshot - iPhone 14 - 2023-06-07 at 17 28 42
Simulator Screenshot - iPhone 14 - 2023-06-07 at 18 16 20

Video here:
https://github.com/vector-im/element-x-ios/assets/34335419/af9406cd-7785-4024-8ca2-b24749804d7a

### DIFF
--- a/ElementX/Sources/Services/Timeline/Fixtures/RoomTimelineItemFixtures.swift
+++ b/ElementX/Sources/Services/Timeline/Fixtures/RoomTimelineItemFixtures.swift
@@ -90,6 +90,28 @@ enum RoomTimelineItemFixtures {
          TextRoomTimelineItem(text: "How are you?",
                               senderDisplayName: "Alice")]
     }
+
+    /// A small chunk of events, containing 2 text items.
+    static var smallChunkWithReadReceipts: [RoomTimelineItemProtocol] {
+        [TextRoomTimelineItem(text: "Hey there 👋",
+                              senderDisplayName: "Alice")
+                .withReadReceipts([ReadReceipt(userID: "a1", formattedTimestamp: nil)]),
+            TextRoomTimelineItem(text: "How are you?",
+                                 senderDisplayName: "Alice")
+                .withReadReceipts([ReadReceipt(userID: "a2", formattedTimestamp: nil),
+                                   ReadReceipt(userID: "b2", formattedTimestamp: nil)]),
+            TextRoomTimelineItem(text: "Fine, Thanks!",
+                                 senderDisplayName: "Bob")
+                .withReadReceipts([ReadReceipt(userID: "a3", formattedTimestamp: nil),
+                                   ReadReceipt(userID: "b3", formattedTimestamp: nil),
+                                   ReadReceipt(userID: "c3", formattedTimestamp: nil)]),
+            TextRoomTimelineItem(text: "What about you?",
+                                 senderDisplayName: "Bob")
+                .withReadReceipts([ReadReceipt(userID: "a4", formattedTimestamp: nil),
+                                   ReadReceipt(userID: "b4", formattedTimestamp: nil),
+                                   ReadReceipt(userID: "c4", formattedTimestamp: nil),
+                                   ReadReceipt(userID: "d4", formattedTimestamp: nil)])]
+    }
     
     /// A chunk of events that contains a single text item.
     static var singleMessageChunk: [RoomTimelineItemProtocol] {
@@ -197,5 +219,13 @@ private extension TextRoomTimelineItem {
                   isEditable: false,
                   sender: .init(id: "", displayName: senderDisplayName),
                   content: .init(body: text))
+    }
+}
+
+private extension TextRoomTimelineItem {
+    func withReadReceipts(_ receipts: [ReadReceipt]) -> TextRoomTimelineItem {
+        var newSelf = self
+        newSelf.properties.orderedReadReceipts = receipts
+        return newSelf
     }
 }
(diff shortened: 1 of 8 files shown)


ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -88,6 +88,15 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
...
             .store(in: &cancellables)
+
+        roomProxy.membersPublisher
+            .map { members in
+                members.reduce(into: [String: RoomMemberState]()) { dictionary, member in
+                    dictionary[member.userID] = RoomMemberState(displayName: member.displayName, avatarURL: member.avatarURL)
+                }
+            }
+            .weakAssign(to: \.state.members, on: self)
+            .store(in: &cancellables)
```
Velin92, line 99: this is how I update the current state of the members in the viewState.
This membersPublisher essentially tells the room when the Proxy has done the fetching (which is automatically started always when entering the room)

callumu:
This looks good 👍

Velin92:
@stefanceriu question! Do we want to include separate UI tests with read receipts, (maybe even just a simple one) where we turn on the feature flag just for the test? (is it even possible?)

stefanceriu:
> Do we want to include separate UI tests with read receipts

Oh, that's a great idea! Should be pretty easy to set up the right settings as we create a new instance every time we start a test in the `UITestsAppCoordinator`. That being said we shouldn't probably access `ServiceLocator.shared.settings.readReceiptsEnabled` directly from the TimelineStatusView

The only thing I'm really worried about the stability of the timeline UI tests but for a simple example it should work well

ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineStatusView.swift
```diff
@@ -16,10 +16,10 @@
 
 import SwiftUI
 
-// This is also going to be used for read receipts in the future
-struct TimelineReceiptView: View {
+struct TimelineStatusView: View {
```
stefanceriu, line 19: Should this maybe be called `Timeline**Item**StatusView` as it relates to an item and not the entire timeline? 🤔

ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineStatusView.swift
```diff
@@ -34,7 +34,9 @@ struct TimelineReceiptView: View {
     }
 
     var body: some View {
-        if shouldShowDeliveryStatus {
+        if !timelineItem.properties.orderedReadReceipts.isEmpty, readReceiptsEnabled {
```
stefanceriu, line 37: Does this mean we never show the delivery status if we have a read receipt? Does this work properly with my own read receipt? 🤔 
Wouldn't it be simpler keep them separate and let them update individually?

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
```diff
@@ -69,6 +69,7 @@ struct RoomScreen: View {
             .id(context.viewState.roomId)
             .environmentObject(context)
             .timelineStyle(context.viewState.timelineStyle)
+            .environment(\.readReceiptsEnabled, context.viewState.readReceiptsEnabled)
```
stefanceriu, line 72: Perhaps we should get rid of that timelineStyle modifier, what do you think?

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
stefanceriu: Yup yup, this looks absolutely fine to me 👍

stefanceriu (approved):
Nice one! 👏

ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineStatusView.swift
Velin92: Exactly that's also what design wants.
Also the delivery status becomes nil once a message has being echoed, and before being echoed there are no read receipts. 
Sadly we have a bug in the SDK where even if I read the message, we are not updating the read receipt in the timeline currently, but this is because the SDK itself is not returning any read receipt for hidden events, like web does.

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
Velin92: yeah it's a bit uselss would prefer to just read .environment(\.timelineStyle)

Velin92:
> > Do we want to include separate UI tests with read receipts
> 
> Oh, that's a great idea! Should be pretty easy to set up the right settings as we create a new instance every time we start a test in the `UITestsAppCoordinator`. That being said we shouldn't probably access `ServiceLocator.shared.settings.readReceiptsEnabled` directly from the TimelineStatusView
> 
> The only thing I'm really worried about the stability of the timeline UI tests but for a simple example it should work well

Yeah moved the settings usage to the view model, and used the environment instead in the view so the intents are now separated!
Then I'll create at least one test so we can keep track of how the read receipts evolve over time before we enable them fully in the app.

PR #1043, opened 2023-06-07
SDK Bump to 1.0.73-alpha

Velin92:


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -437,7 +437,7 @@ class ClientProxy: ClientProxyProtocol {
             let slidingSync = try slidingSyncBuilder
                 .addList(listBuilder: visibleRoomsListBuilder)
                 .withCommonExtensions()
-                .bumpEventTypes(bumpEventTypes: roomListRecencyOrderingAllowedEventTypes)
+//                .bumpEventTypes(bumpEventTypes: roomListRecencyOrderingAllowedEventTypes)
```
Velin92, line 440: not sure about this @stefanceriu  this function seems to have been removed

stefanceriu: Yes, as https://github.com/matrix-org/sliding-sync/pull/142 the bunp event types are per list now

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -466,12 +462,15 @@ class ClientProxy: ClientProxyProtocol {
...
         let visibleRoomsListProxy = SlidingSyncListProxy(name: listName)
+
+        let roomListRecencyOrderingAllowedEventTypes = ["m.room.message", "m.room.encrypted", "m.sticker"]
         
         let visibleRoomsListBuilder = SlidingSyncListBuilder(name: listName)
             .timelineLimit(limit: UInt32(SlidingSyncConstants.initialTimelineLimit)) // Starts off with zero to quickly load rooms, then goes to 1 while scrolling to quickly load last messages and 20 when the scrolling stops to load room history
             .requiredState(requiredState: slidingSyncRequiredState)
             .filters(filters: slidingSyncFilters)
             .syncModeSelective(selectiveModeBuilder: .init().addRange(start: 0, endInclusive: 20))
+            .bumpEventTypes(bumpEventTypes: roomListRecencyOrderingAllowedEventTypes)
```
stefanceriu, line 473: Oh.. I think we need this in the allRoomsList as well otherwise their ordering will be different and merging the 2 lists will be a mess.

stefanceriu (changes requested):
Looks fine except the bump event types that need to be shared

ElementX/Sources/Services/Client/ClientProxy.swift
Velin92: should it be only for this list right?

ElementX/Sources/Services/Client/ClientProxy.swift
Velin92: done!

ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu: No no, as per the other comment here it needs to be on both visible and all

Velin92:
@stefanceriu  I have also removed all the code related to the local notifications since it's kind of a thing we will probably never need again or way later in the future, so to avoid confusion I removed the FF and the code related to it.

Issue #1016, opened 2023-06-05
Room details: improve members count UX


alfogrillo:
Currently users are waiting for all the members to be fetched to see the members count in the room's details.

The same applies for seeing additional features like:
- Edit the room
- Invite more people

Rust gives us the number of joined/invited people without doing any additional call.
We may show it while fetching more members.

Ideally Rust should also provide a quick way to get `RoomMember` for the current user.
This is required to check power levels in the room.

**ACs**
- The members count is instantly visible in the room's details
- **TBC**: power levels can be checked quickly without fetching all the room members

related to: https://github.com/vector-im/element-x-android/issues/505

**Impacted screen**
![poc](https://github.com/vector-im/element-x-ios/assets/19324622/bfc31afe-d2d0-4f30-8e70-9faa21f31e90)

PR #1050, opened 2023-06-08
Room details: improve members count UX 

alfogrillo:
Fix https://github.com/vector-im/element-x-ios/issues/1016

| Before | After |
| --- | --- |
| b|   a|


ElementX/Sources/Screens/InviteUsersScreen/InviteUsersScreenViewModel.swift
```diff
@@ -60,13 +60,15 @@ class InviteUsersScreenViewModel: InviteUsersScreenViewModelType, InviteUsersScr
             actionsSubject.send(.toggleUser(user))
         }
     }
+
+    // MARK: - Private
     
-    private func buildMembershipStateIfNeeded() {
-        guard case let .room(members, userIndicatorController) = roomType else {
+    private func buildMembershipStateIfNeeded(members: [RoomMemberProxyProtocol]) {
+        guard case let .room(_, userIndicatorController) = roomType else {
```
Velin92, line 67: since this function and the fetchMembershipIfNeeded function share the same check at the start, maybe we can remove this check entirely and just call this buildMembershipState ? 
Or is this function used somewhere else?

ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenViewModel.swift
```diff
@@ -88,74 +88,69 @@ class RoomDetailsScreenViewModel: RoomDetailsScreenViewModelType, RoomDetailsScr
...
+    
+    private func updatePowerLevelPermissions() {
+        state.canInviteUsers = accountOwner?.canInviteUsers ?? false
+        state.canEditRoomName = accountOwner?.canSendStateEvent(type: .roomName) ?? false
+        state.canEditRoomTopic = accountOwner?.canSendStateEvent(type: .roomTopic) ?? false
+        state.canEditRoomAvatar = accountOwner?.canSendStateEvent(type: .roomAvatar) ?? false
+    }
+    
+    private var isEncryptedDirectRoom: Bool {
+        roomProxy.isDirect && roomProxy.isEncrypted && roomProxy.joinedMembersCount == 2
```
Velin92, line 153: actually this could be useful to have on the roomProxy itself as an extension, so we may reuse it in the future for other views, WDYT? (for example in the future we would want the room screen's message composer to be blocked if this condition is met and the other user is blocked)

ElementX/Sources/UITests/UITestsAppCoordinator.swift
```diff
@@ -306,8 +306,11 @@ class MockScreen: Identifiable {
             let roomProxy = RoomProxyMock(with: .init(id: "MockRoomIdentifier",
                                                       displayName: "Room",
                                                       isEncrypted: true,
-                                                      members: members))
-            let coordinator = RoomDetailsScreenCoordinator(parameters: .init(navigationStackCoordinator: navigationStackCoordinator,
+                                                      members: members,
+                                                      memberForID: .mockOwner(allowedStateEvents: [], canInviteUsers: false),
+                                                      joinedMembersCount: UInt(members.count)))
```
Velin92, line 311: I see that we are doing a lot of back and forth between UInt and Int, at this point maybe we could expose the joined members and the invited members count as Int in the proxy, and always use it as Int, since Swift is a bit lacking in terms of handling UInt in the most general context.
In this way the count values can be used on a swifty way and the conversion only happens at a Proxy level.

ElementX/Sources/Screens/InviteUsersScreen/InviteUsersScreenViewModel.swift
```diff
@@ -103,6 +103,26 @@ class InviteUsersScreenViewModel: InviteUsersScreenViewModelType, InviteUsersScr
             .store(in: &cancellables)
     }
     
+    private func fetchMembersIfNeeded() {
+        guard case let .room(roomProxy, userIndicatorController) = roomType else {
+            return
+        }
+        
+        Task {
+            showLoader(using: userIndicatorController)
```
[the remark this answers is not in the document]
alfogrillo: Cool,having a look to see if it's ok here 👍🏻

ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenViewModel.swift
alfogrillo: Yep, now it can be an extension like you say.
Making a private extension for now.
People can export it later if they need it.
Wdyt?

ElementX/Sources/UITests/UITestsAppCoordinator.swift
alfogrillo: Although a counter conceptually is a UInt, you are right: very often we deal with just Int.
The only problem I see refactoring the proxy to expose Int, is the potential crashes we may get when very big numbers cannot be represented like Int. 
I think in our case it's safe enough.
Wdyt?

ElementX/Sources/Screens/InviteUsersScreen/InviteUsersScreenViewModel.swift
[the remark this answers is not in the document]
alfogrillo: [Applied](https://github.com/vector-im/element-x-ios/pull/1050/commits/2c11486d0d3f6347a3b9c956455fb003257d4c44)

ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenViewModel.swift
alfogrillo: Made private extension [here](https://github.com/vector-im/element-x-ios/pull/1050/commits/2c11486d0d3f6347a3b9c956455fb003257d4c44)

ElementX/Sources/UITests/UITestsAppCoordinator.swift
Velin92: yeah I mean, not even Matrix itself has more than 18446744073709551615 users (should be Int overflow value on 64 bit), so probably not an issue we will need to deal ever. I guess is safe to convert to Int

ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenViewModel.swift
Velin92: maybe I would already make it public in the RoomProxyProtocol file itself, I feel is just easier to find when the time comes that will need it for other models

ElementX/Sources/UITests/UITestsAppCoordinator.swift
alfogrillo: Yeah, it should be safe enough!
[Applied](https://github.com/vector-im/element-x-ios/pull/1050/commits/4d8bf7dba697072e97dc47a65732b2f5c720a2c4)

ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenViewModel.swift
alfogrillo: Fair enough. [Applied](https://github.com/vector-im/element-x-ios/pull/1050/commits/4d8bf7dba697072e97dc47a65732b2f5c720a2c4)

PR #1055, opened 2023-06-08
RTL and LTR mixed languages fix for optimal timestamp support

Velin92:
RTL layout:
Screenshot 2023-06-08 at 14 52 22
LTR layout:
Screenshot 2023-06-08 at 14 36 54


changelog.d/1052.feature
Velin92, line 1: forgot to add this important changelog in my previous PR, adding it here.

ElementX/Sources/Screens/RoomScreen/View/Timeline/FormattedBodyText.swift
```diff
@@ -36,7 +36,19 @@ struct FormattedBodyText: View {
 
     // These is needed to create the slightly off inlined timestamp effect
     private var additionalWhitespacesSuffix: String {
-        .generateBreakableWhitespaceEnd(whitespaceCount: additionalWhitespacesCount, isRTL: layoutDirection == .rightToLeft)
+        .generateBreakableWhitespaceEnd(whitespaceCount: additionalWhitespacesCount, layoutDirection: layoutDirection)
+    }
+
+    // This alllows to render the start and only the start of the string with the layout of the UI for the current user
```
alfogrillo, line 42:
```suggestion
    // This allows to render the start and only the start of the string with the layout of the UI for the current user
```

ElementX/Sources/Screens/RoomScreen/View/Timeline/FormattedBodyText.swift
```diff
@@ -36,7 +36,19 @@ struct FormattedBodyText: View {
...
+    private var layoutPrefix: AttributedString {
+        switch layoutDirection {
+        case .leftToRight:
+            return "\u{2066}"
+        case .rightToLeft:
+            return "\u{2067}"
+        default:
+            return ""
```
alfogrillo, line 50: Saw this switch already.
Is it worth to extend `LayoutDirection` and put it there?

alfogrillo (approved):
LGTM.
Left minor comments

ElementX/Sources/Screens/RoomScreen/View/Timeline/FormattedBodyText.swift
Velin92: oh yeah that's a good idea

Velin92:
i had to do a a bit of change, I removed the prefix thing because it could mess up with messages that contained only blockquotes, the timestamp still works fine for RTL strings, but however it could be displayed slightly closer (so the spacing is less consistent), probably there is a fix for that too, but it's going to be a bit too time consuming, and worth revisting later.

PR #1059, opened 2023-06-09
SDK Bump to 1.0.74

Velin92:
Includes also some quality of life adjustments like auto generating mocks from local when you switch to local usage of the sdk, some swiftlint corrections, some flaky tests fixes.

This SDK update should get rid of some SDK bugs like the the calculated name being used instead of the name for the rooms sometimes.


UnitTests/Sources/RoomScreenViewModelTests.swift
```diff
@@ -19,6 +19,12 @@ import XCTest
 
 @MainActor
 class RoomScreenViewModelTests: XCTestCase {
+    var userIndicatorControllerMock: UserIndicatorControllerMock!
+
+    override func setUp() async throws {
+        userIndicatorControllerMock = nil
```
alfogrillo, line 25: Why assigning nil in the setup?
I expected more:
```swift
userIndicatorControllerMock = UserIndicatorControllerMock.default
```

alfogrillo (approved):
LGTM, left just a comment

UnitTests/Sources/RoomScreenViewModelTests.swift
Velin92: oh yeah actually this is even better I'll do that

PR #1063, opened 2023-06-09
buildSDK.sh script, build the sdk without encountering target problems

Velin92:
After struggling for half a day I think that is the only reliable way to run the swift run tools build-sdk without encountering any issues. Not saying you should use this script, it only makes sense to use if you are having trouble and you need to build the sdk also for the iPhone.

At least this should allow the sdk to be built locally for who has encountered issues without too many headaches.

### DIFF
--- a/Tools/Sources/BuildSDK.swift
+++ b/Tools/Sources/BuildSDK.swift
@@ -91,7 +91,9 @@ struct BuildSDK: ParsableCommand {
     
     /// Build the Rust SDK as an XCFramework with the debug profile.
     func buildFramework() throws {
-        var buildCommand = "cargo xtask swift build-framework"
+        // unset fixed an issue where swift compilation prevents building from other targets
+        // other than macOS X
+        var buildCommand = "unset SDKROOT && cargo xtask swift build-framework"
         buildCommand.append(" --profile \(profile.rawValue)")
         if let target {
             buildCommand.append(" --only-target \(target.rawValue)")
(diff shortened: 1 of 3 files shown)


stefanceriu (changes requested):
I decided to have a go at this and try to understand exactly what's happening. I went and compared the environment between a normal shell and whatever swift is doing and found that the SDKROOT is automatically set before each process invocation to something along the lines of `SDKROOT=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.3.sdk`

With that in hand I went and changed the `buildFramework` step to `unset SDKROOT && cargo xtask swift build-framework` which seems to fix the problem without any other changes.

Velin92:
Seems to be working perfectly!
Thanks you are my saviour @stefanceriu 

I really needed this for the future notifications api tests on device.

Tools/Sources/BuildSDK.swift
```diff
@@ -91,7 +91,9 @@ struct BuildSDK: ParsableCommand {
     
     /// Build the Rust SDK as an XCFramework with the debug profile.
     func buildFramework() throws {
-        var buildCommand = "cargo xtask swift build-framework"
+        // unset fixed an issue where swift compilation prevents building from other targets
+        // other than macOS X
```
stefanceriu, line 95:
```suggestion
        // unset fixes an issue where swift compilation prevents building for targets other than macOS X
```

stefanceriu (approved):
🚀 !

PR #1068, opened 2023-06-12
Disable RR extension when feature flag is off

Velin92:
since we need the extension to be disable to fully support the new Room List API


stefanceriu (approved):
👌

PR #1070, opened 2023-06-12
Workaround: fix search related memory leaks

alfogrillo:
### Problem
the app is leaking coordinators/view models when:

- There is a wrapping `UserIndicatorController` for the scene
- The scene is presented as a modal sheet
- The user starts the search and then dismisses the modal before aborting the search. The leak doesn't happen if the search is dismissed first.

**Affected scenes**
- StartChatScreen
- InviteUsersScreen

The leaks seem somehow related with the `searchable` modifier, and isn't hard reproducing the same situation in a new project.

#### Tried also
- Save the `searchQuery` into a `@State` instead of inside the context
- Just dismiss the search when views' `onDisappear` is called


stefanceriu (approved):
Mega weird but ¯\_(ツ)_/¯ Nicely done on the workaround 👍

PR #1074, opened 2023-06-13
Fix modal sheet presentations when invoked from the timeline item men…

stefanceriu:
…u sheet

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/View/TimelineItemMenu.swift
+++ b/ElementX/Sources/Screens/RoomScreen/View/TimelineItemMenu.swift
@@ -215,7 +215,10 @@ public struct TimelineItemMenu: View {
     
     private func send(_ action: TimelineItemMenuAction) {
         presentationMode.wrappedValue.dismiss()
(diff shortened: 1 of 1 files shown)


Velin92:
Could we try instead of using `PresentationMode` which is also another thing that apple is deprecating the environment variable dismiss and isPresented?

https://developer.apple.com/documentation/swiftui/environmentvalues/dismiss
https://developer.apple.com/documentation/swiftui/environmentvalues/ispresented

I think what you can do is to call the dismiss, then listen to the change of isPresented to become false to then call the other present without the need of dispatching after some time.

Velin92 (approved):

stefanceriu:
> Could we try instead of using `PresentationMode`

Talked about this offline, will merge this fix as it is and try the other solution in a subsequent one asap

stefanceriu:
> try the other solution in a subsequent PR asap

Had a look at adopting this but unfortunately `isPresented` never changes. We believe it's because we're using a UIViewRepresentable in the timeline but we have no real proof at the moment

PR #1076, opened 2023-06-13
Update camera and microphone usage descriptions

stefanceriu:


Velin92 (approved):

ElementX/SupportingFiles/target.yml
```diff
@@ -62,8 +62,8 @@ targets:
         NSUserActivityTypes: [
           INSendMessageIntent
         ]
-        NSCameraUsageDescription: The camera is used to take and upload photos and videos.
-        NSMicrophoneUsageDescription: The microphone is used to take videos.
+        NSCameraUsageDescription: To take pictures or videos and send them as a message Element X needs access to the camera.
```
alfogrillo, line 65: We could replace Element X with `$(APP_DISPLAY_NAME)`
@flescio already tested it in the location permissions case and it looks working.

alfogrillo (approved):
LGTM, just left a minor comment

ElementX/SupportingFiles/target.yml
stefanceriu: Oh nice, thanks 👍

PR #1082, opened 2023-06-14
Two sync loop experimental solution to decrypt push notifications

Velin92:
This is using an external matrix rust sdk experimental branch, not merged yet in main, so it will likely fail to compile on CI for now.

The idea is to use the notification sync (which I feel it should be renamed to encryption or decryption sync) to actually update the keys.

The sync runs both in the main app and in the NSE, we can now wait for the notification to be fully decrypted (we wait for about 10 seconds at most for now but it takes way less) before showing it to the user.

~~This requires to not use all common extensions but remove e2ee and toDeviceEvents from the normal sync, since the notification sync does that job instead.~~

Updated to work with RoomListAPI

Adding also an ff, to enable or not this, since could require some testing for quite some in the nightly time before we know if is stable or not.


ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -290,8 +299,8 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
     }
     
     private func dismissRoom(animated: Bool) {
-        navigationStackCoordinator.popToRoot(animated: animated)
         navigationSplitCoordinator.setDetailCoordinator(nil, animated: animated)
+        navigationStackCoordinator.popToRoot(animated: animated)
```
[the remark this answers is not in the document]
alfogrillo: Had the same problem with animations and fixed it differently [here](https://github.com/vector-im/element-x-ios/commit/003085aa57799afdc5095b5be94bdff446f485d2). 
Not sure what's best. 
If we go with your fix, we may want to add a comment to highlight the fact that lines shouldn't be inverted again. 
Also worth checking what's happening on iPad after this change.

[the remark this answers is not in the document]
Velin92: Actually now that I think about this, since the invite screen on the iPad lives in the master and not in the details of the iPad, maybe we should not dismiss the room at all? Maybe we should not touch the state of the room flow coordinator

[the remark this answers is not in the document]
Velin92: I think we should only clear the route if we are on iPhone and keep it as it is on iPad, WDYT?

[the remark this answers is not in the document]
alfogrillo: (discussed privately)

NSE/Sources/Other/NSEUserSession.swift
```diff
@@ -19,17 +19,23 @@ import MatrixRustSDK
 
 final class NSEUserSession {
     private let client: ClientProtocol
+    private var encryptionSync: EncryptionSync!
```
stefanceriu, line 22: Can we call this `encryptionSyncService` or something else along those lines?

NSE/Sources/NotificationServiceExtension.swift
```diff
@@ -70,9 +71,21 @@ class NotificationServiceExtension: UNNotificationServiceExtension {
...
+                Self.userSession[credentials.userID] = userSession
+            }
+
+            // TODO: The following wait with a timeout should be handled by the SDK
+            // We try to decrypt the notification for 10 seconds at most
+            var itemProxy: NotificationItemProxyProtocol
+            let date = Date()
+            repeat {
+                itemProxy = try await userSession.notificationItemProxy(roomID: roomId, eventID: eventId)
+            } while itemProxy.isEncrypted && date.timeIntervalSinceNow < 10
```
stefanceriu, line 88: Probably worth leaving a warning here

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -400,7 +433,7 @@ class ClientProxy: ClientProxyProtocol {
                 if state == .allRooms {
                     Task {
                         // Subscribe to invites later as the underlying SlidingSync list is only added when entering AllRooms
-                        await self.inviteSummaryProvider?.subscribeIfNecessary(entriesFunction: roomListService.invites(listener:),
+                        await self.inviteSummaryProvider?.subscribeIfNecessary(entriesFunction: roomListService.entries(listener:),
                                                                                // This state function is wrong but it's not used
```
stefanceriu, line 437: What's this about? 🤔

stefanceriu, line 437: Also, I fixed this here https://github.com/vector-im/element-x-ios/pull/1109

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -19,6 +19,18 @@ import Foundation
 import MatrixRustSDK
 import UIKit
 
+private class ConcreteEncryptionSyncListener: EncryptionSyncListener {
```
stefanceriu, line 22: I tried going for block based listeners for these (see the end of the file) maybe we should do the same here

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -376,7 +397,19 @@ class ClientProxy: ClientProxyProtocol {
             self.avatarURLSubject.value = urlString.flatMap(URL.init)
         }
     }
-        
+
+    private func configureEncryptionSync(listener: EncryptionSyncListener) {
+        guard encryptionSync == nil else {
+            MXLog.info("Encryption sync is already configured")
```
stefanceriu, line 403: I would fatalError here, no reason why it should be configured twice

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -376,7 +397,19 @@ class ClientProxy: ClientProxyProtocol {
...
+            return
+        }
+        do {
+            try encryptionSync = client.mainEncryptionSync(id: "Main App", listener: listener, withLock: true)
```
stefanceriu, line 407: I was expecting these services to be similar, why doesn't encryptionSync return a taskHandle? 🤔

stefanceriu (commented):
Left some comments inline but it looks good to me 👍

NSE/Sources/NotificationServiceExtension.swift
stefanceriu: Ah right, cool cool 👍

ElementX/Sources/Services/Client/ClientProxy.swift
Velin92: Oh yeah I like this, and I can reuse it for NSE too.

ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu: I personally believe we should crash as soon as we encounter unexpected states. Crashes are a lot easier to track down and people generally report them more often.

ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu: Alrighty ¯\_(ツ)_/¯

ElementX/Sources/Services/Client/ClientProxy.swift
Velin92: Okay actually I refactored the code, now the room list api and the encryption sync will always work together, so I can use the fatal error now

ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -553,6 +553,16 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
...
+
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(applicationWillTerminate),
+                                               name: UIApplication.willTerminateNotification,
+                                               object: nil)
+    }
+
+    @objc
+    private func applicationWillTerminate() {
+        userSession?.clientProxy.stopSync()
```
stefanceriu, line 565: 👍

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -376,14 +391,40 @@ class ClientProxy: ClientProxyProtocol {
...
+        guard ServiceLocator.shared.settings.isEncryptionSyncEnabled else {
+            return
+        }
+        configureEncryptionSyncService()
+    }
+    
+    private func configureEncryptionSyncService() {
+        do {
+            let listener = EncryptionSyncListenerProxy { [weak self] in
+                MXLog.info("Encryption Sync did terminate for user: \(self?.userID ?? "unknown")")
```
stefanceriu, line 405: This is a tad weird, is the listener only good for knowing when the encryptionSync stopped? 🤔

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -123,23 +127,34 @@ class ClientProxy: ClientProxyProtocol {
...
+        roomListSyncTaskHandle != nil && isEncryptionSyncing
     }
     
     func startSync() {
         MXLog.info("Starting sync")
-        guard !isSyncing else {
-            return
-        }
-        
+        guard !isSyncing else { return }
```
stefanceriu, line 135: I believe this will be a problem if the room list loop fails (for whatever reason) and we try to restart it. The encryption sync will still be running and it will return early without actually restarting it.

stefanceriu (changes requested):
I'm worried about isSyncing and merging together the 2 loops, otherwise it looks good to me 👍

ElementX/Sources/Services/Client/ClientProxy.swift
Velin92: yeah, and it can either stop because we are stopping it, or maybe because there is no connection or other issues.
Benjamin wants to introduce a TerminationReason enum in the future.

ElementX/Sources/Services/Client/ClientProxy.swift
Velin92: however now isSyncing is based on:
`roomListSyncTaskHandle != nil && isEncryptionSyncing`

So the sync to be seen as really running, both of them need to be running at the same time.
So if the room list has failed the start function is going to recreate another encryption sync which is okay because it will have the same identifier, so it will still keep the previous lock.

Velin92: otherwise we could create `startRoomListSync()` and a `stopRoomListSync()` and a isRoomListSyncing flag and use those instead to restart the room list sync in the termination cases (essentially use the same pattern we are using now for the encryption sync but for the room list sync, let me know if you prefer this solution)

stefanceriu: > the start function is going to recreate another encryption sync

Well that won't happen because there's an check in there for `guard !isSyncing else { return }`

I'm happy to ignore `&& isEncryptionSyncing` in `isSyncing` if you thing that's going to work properly.

stefanceriu (approved):

PR #1084, opened 2023-06-14
Refactored alert info to not use the soon to be deprecated API

Velin92:
Refactored the `AlertInfo` so that it will now use the new Apple's API to display alerts instead of the soon to be deprecated one.


ElementX/Sources/Screens/RoomMemberDetailsScreen/View/RoomMemberDetailsScreen.swift
```diff
@@ -29,7 +29,7 @@ struct RoomMemberDetailsScreen: View {
         }
         .elementFormStyle()
         .alert(item: $context.ignoreUserAlert, actions: blockUserAlertActions, message: blockUserAlertMessage)
-        .errorAlert(item: $context.errorAlert)
+        alert(item: $context.alertInfo)
```
pixlwave, line 32: Not 100% sure how the unit tests pass in this case unless I'm being very blind and missing something?
```suggestion
        .alert(item: $context.alertInfo)
```

pixlwave (approved):
Love it! 👏

PR #1085, opened 2023-06-15
Add reverse geocoding service

flescio:
fixes https://github.com/vector-im/element-x-ios/issues/1015

Add reverse geocoding request, that for a given coordinate will return the place name.


ElementX/Sources/Application/AppSettings.swift
```diff
@@ -184,6 +184,9 @@ final class AppSettings {
     let lightTileStaticMapStyleURL = "https://api.maptiler.com/maps/9bc819c8-e627-474a-a348-ec144fe3d810/{z}/{x}/{y}.png"
     let darkTileStaticMapStyleURL = "https://api.maptiler.com/maps/dea61faf-292b-4774-9660-58fcef89a7f3/{z}/{x}/{y}.png"
     
+    // maptiler geocoding url
+    let geocodingURL = "https://api.maptiler.com/geocoding/%f,%f.json"
```
stefanceriu, line 188: We should probably call this `geocodingURLFormatString` or something as it's not a valid URL

ElementX/Sources/Other/MapLibre/MapTilerGeocoding.swift
```diff
@@ -0,0 +1,83 @@
...
+        
+        let features: [FeatureCollection]
+    }
+    
+    func reverseGeoCoding(for coordinate: CLLocationCoordinate2D) async -> Result<String, MapTilerGeocodingError> {
+        let latitude = coordinate.latitude
+        let longitude = coordinate.longitude
+        
+        let path = String(format: geocodingURL, longitude, latitude)
+        guard var url = URL(string: path) else { return .failure(.geocodingFailed) }
```
stefanceriu, line 48: I wonder if we should fatalError here, this should always be a valid URL, no?

ElementX/Sources/Other/MapLibre/MapTilerGeocoding.swift
```diff
@@ -0,0 +1,83 @@
...
+        let authorization = MapTilerAuthorization(key: key)
+        url = authorization.authorizateURL(url)
```
stefanceriu, line 50: `authorizate`.. that's a new one 😅

ElementX/Sources/Other/MapLibre/MapTilerGeocoding.swift
```diff
@@ -0,0 +1,83 @@
...
+            let (data, response) = try await session.dataWithRetry(for: request)
+            
+            guard let httpResponse = response as? HTTPURLResponse else {
+                let errorDescription = String(data: data, encoding: .utf8)?.trimmingCharacters(in: .whitespacesAndNewlines) ?? "Unknown"
+                MXLog.error("Failed to get reverse geocoding: \(errorDescription)")
+                MXLog.error("Response: \(response)")
+                return .failure(.geocodingFailed)
+            }
+            
+            guard httpResponse.statusCode == 200 else {
```
stefanceriu, line 66: Can't you merge this with the one above? Seems very similar

ElementX/SupportingFiles/Info.plist
```diff
@@ -28,6 +28,8 @@
 	<false/>
 	<key>NSCameraUsageDescription</key>
 	<string>To take pictures or videos and send them as a message $(APP_DISPLAY_NAME) needs access to the camera.</string>
+	<key>NSLocationWhenInUseUsageDescription</key>
+	<string>When you share your location to people, $(APP_DISPLAY_NAME) needs access to show them a map.</string>
```
stefanceriu, line 32: Did you validate this with product people? I imagine it's going to change when we start sending locations ourselves, no?

ElementX/Sources/Other/MapLibre/MapTilerGeocoding.swift
```diff
@@ -0,0 +1,83 @@
...
+                MXLog.error("Failed to get reverse geocoding: \(errorDescription) (\(httpResponse.statusCode))")
+                MXLog.error("Response: \(httpResponse)")
+                return .failure(.geocodingFailed)
+            }
+            
+            // Parse the JSON data
+            let decoder = JSONDecoder()
+            decoder.keyDecodingStrategy = .convertFromSnakeCase
+            let decodedResponse = try decoder.decode(GeocodedPlace.self, from: data)
+            guard let placeName = decodedResponse.features.first?.placeName else { return .failure(.geocodingFailed) }
```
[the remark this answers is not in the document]
flescio: the team agreed on that, also we only receive 1 result since url.append(queryItems: [.init(name: "limit", value: "1")])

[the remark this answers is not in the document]
stefanceriu: Fair enough 👍

ElementX/Sources/Other/MapLibre/MapTilerGeocoding.swift
flescio: I'm not quite sure about that, since the main point of this and a future task is to make it also easily forkable, I prefer to shout out an error than terminating, even if in our case, most likely should never occur

stefanceriu: Okay well in that case you should create a different error case just for this

ElementX/Sources/Other/Extensions/MapTiler.swift
```diff
@@ -21,3 +21,9 @@ extension MapTilerStyleBuilder {
         self.init(lightURL: appSettings.lightTileMapStyleURL, darkURL: appSettings.darkTileMapStyleURL, key: appSettings.mapTilerApiKey)
     }
 }
+
+extension MapTilerGeoCodingService {
+    init(session: URLSession = .shared, appSettings: AppSettings) {
```
stefanceriu, line 26: Out of curiosity, why are these 2 constructors better than the normal way?

ElementX/Sources/Other/MapLibre/MapTilerGeoCodingServiceProtocol.swift
```diff
@@ -0,0 +1,22 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import CoreLocation
+import Foundation
+
+protocol MapTilerGeoCodingServiceProtocol {
+    func reverseGeoCoding(for coordinate: CLLocationCoordinate2D) async -> Result<String, MapTilerGeocodingError>
```
stefanceriu, line 21: The `MapTilerGeocodingError` errors should be part of this protocol (same file) as they pertain to it

ElementX/Sources/Other/MapLibre/MapTilerGeocoding.swift
```diff
@@ -0,0 +1,72 @@
...
+    }
+    
+    func reverseGeoCoding(for coordinate: CLLocationCoordinate2D) async -> Result<String, MapTilerGeocodingError> {
+        let latitude = coordinate.latitude
+        let longitude = coordinate.longitude
+        
+        let path = String(format: geocodingURL, longitude, latitude)
+        guard var url = URL(string: path) else { return .failure(.wrongGeocodingURL) }
+        let authorization = MapTilerAuthorization(key: key)
+        url = authorization.authorizeURL(url)
```
stefanceriu, line 46: I just noticed that thi authorizeURL thing is implemented in a different struct. Do we really need that, isn't building that url the right part of this service's responsibility? Do you plan on reusing it anywhere? 🤔

stefanceriu (approved):
lgtm 👍

ElementX/Sources/Other/MapLibre/MapTilerGeocoding.swift
flescio: it's already being used in 2 separate services and another on it's way, basically the map, the reverse geocoding and the static map image (in timeline) will use this

ElementX/Sources/Other/Extensions/MapTiler.swift
flescio: The idea is that, the MapLibre implementation should live as a separate component and shouldn't know about AppSettings. While not within the scope of this prep activity, I prepared the work that way as much as possible

PR #1086, opened 2023-06-15
Adopt Compound Colours

pixlwave:
- Deprecate all of the old SwiftUI colour tokens in DesignKit.
- Delete all of the old custom SwiftUI colours from DesignKit.
- Delete all of the UIKit colours from DesignKit using `UIColor(.compound.someColor)` instead.
- Migrates all of the tokens used from DesignKit to Compound.
- A few other minor tweaks discussed with Design such as small font changes.

Snapshot updates to follow when all app polish is finished.


DesignKit/Sources/Colors/ElementColors.swift
```diff
@@ -27,22 +27,33 @@ public extension Color {
 public struct ElementColors {
     // MARK: - Legacy Compound
     
-    private let compound = DesignTokens.CompoundColors()
```
alfogrillo, line 30: Ideally we want to remove `DesignKit` from the project in the long run right?

I see two stronger alternatives to desecration's:
- Delete `ElementColors` at all if possible
- Mark properties as `obsoleted` to generate build errors instead of warnings

wdyt?

DesignKit/Sources/TextFields/ElementTextFieldStyle.swift
```diff
@@ -57,8 +57,7 @@ public struct ElementTextFieldStyle: TextFieldStyle {
     
     /// The color of the text field's border.
     private var borderColor: Color {
-        guard !isError else { return .element.alert }
-        return isFocused ? .element.tertiaryContent : .element.quinaryContent
+        isError ? .compound.textCriticalPrimary : .compound._borderTextFieldFocused
```
alfogrillo, line 60: `isFocused` disappeared. Is it correct?

DesignKit/Sources/TextFields/ElementTextFieldStyle.swift
```diff
@@ -67,39 +66,32 @@ public struct ElementTextFieldStyle: TextFieldStyle {
...
+        isEnabled ? .compound.textPrimary : .compound.textDisabled
     }
     
     /// The color of the text field's background.
     private var backgroundColor: Color {
-        if !isEnabled, colorScheme == .dark {
-            return .element.quinaryContent
-        }
-        return .element.system
+        .compound.bgSubtleSecondary.opacity(isEnabled ? 1 : 0.5)
```
alfogrillo, line 79: I guess the `colorScheme` is managed inside the color asset itself right?

ElementX/Sources/Other/HTMLParsing/AttributedStringBuilder.swift
```diff
@@ -112,7 +112,7 @@ struct AttributedStringBuilder: AttributedStringBuilderProtocol {
         attributedString.enumerateAttribute(.backgroundColor, in: .init(location: 0, length: attributedString.length), options: []) { value, range, _ in
             if let value = value as? UIColor,
                value == temporaryCodeBlockMarkingColor {
-                attributedString.addAttribute(.backgroundColor, value: UIColor.element.quinaryContent as Any, range: range)
+                attributedString.addAttribute(.backgroundColor, value: UIColor(.compound._bgCodeBlock) as Any, range: range)
```
alfogrillo, line 115: Why we have colors prefixed with `_`?

alfogrillo (commented):
Great job!
I trust you on color mapping ofc. 🙂

Left some clarification questions before we merge. ;)

DesignKit/Sources/Colors/ElementColors.swift
pixlwave: Correct yes, its just a few button/text field styles left in here now, they can become part of the app in the near future tbf.

I deprecated just to avoid breaking anyones PRs that are in-flight right now, my plan was to come back in a couple of weeks and delete that. If we want to be more aggressive and obsolete I can do that in this PR (so it still shows the suggested replacements for existing PRs)?

DesignKit/Sources/TextFields/ElementTextFieldStyle.swift
pixlwave: Correct, because the border width is set to 0 when `!isFocused` so it seems to me to be a redundant check added.

DesignKit/Sources/TextFields/ElementTextFieldStyle.swift
pixlwave: Correct, everything from compound is built as [colorsets](https://github.com/vector-im/compound-design-tokens/blob/develop/assets/ios/swift/Compound.xcassets/colorGray100.colorset/Contents.json) for light/dark/light-hc/dark-hc.

ElementX/Sources/Other/HTMLParsing/AttributedStringBuilder.swift
pixlwave: Sorry I should have linked the [PR](https://github.com/vector-im/compound-ios/pull/9) for this in the description. These are temporary colours that Callum requested were manually added until we have time to define them for real in the design tokens.

So I followed the Swift [convention](https://github.com/apple/swift/blob/main/docs/ReferenceGuides/UnderscoredAttributes.md) here to indicate that they aren't stable and could change.

DesignKit/Sources/Colors/ElementColors.swift
alfogrillo: > If we want to be more aggressive and obsolete I can do that in this PR

Looks good to me, I think we don't have big UI PR in-flight at the moment.
Fixing error should be easy anyway since you are suggesting the fix 👍🏻

pixlwave: Updated with https://github.com/vector-im/element-x-ios/pull/1086/commits/396653da2c3c8559709f2e7c64293f4c7a02b1e8

alfogrillo (approved):
LGTM! 🎉

PR #1091, opened 2023-06-16
Invitation Notification design improvements and navigation

Velin92:
This is only going to work properly when the two sync loop is released, but since it's its own separate piece of work, and is not going to impact the current invite notifications (since they appear as generic ones most of the time), it's nice to already have it ready for when the two sync loop is ready

NOTE:
This branch is mostly cherry picks of the Two Syncs branch

### DIFF
--- a/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
@@ -299,9 +299,8 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
     }
     
     private func dismissRoom(animated: Bool) {
-        // The room isn't in the same navigation stack of the home screen.
-        // Animating the popToRoot causes weird animations on when the room is left from room's details
-        navigationStackCoordinator.popToRoot(animated: false)
+        // Setting the detail coordinator to nil afirst allows the dismiss to work properly
+        // if followed immediately by another navigation on iPhone
         navigationSplitCoordinator.setDetailCoordinator(nil, animated: animated)
         navigationStackCoordinator.popToRoot(animated: animated)
         roomProxy = nil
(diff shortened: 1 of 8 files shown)


ElementX/Sources/Services/Notification/Proxy/NotificationItemProxy.swift
```diff
@@ -41,7 +41,9 @@ protocol NotificationItemProxyProtocol {
 
     var isDirect: Bool { get }
 
-    var isEncrypted: Bool? { get }
+    var isRoomEncrypted: Bool? { get }
+
+    var isEncrypted: Bool { get }
```
pixlwave, line 46: I presume there's a time when these would be different? Might be worth documenting it briefly to help people coming to the code with fresh eyes.

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -290,10 +299,10 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
     }
     
     private func dismissRoom(animated: Bool) {
-        // The room isn't in the same navigation stack of the home screen.
-        // Animating the popToRoot causes weird animations on when the room is left from room's details
-        navigationStackCoordinator.popToRoot(animated: false)
+        // Setting the detail coordinator to nil afirst allows the dismiss to work properly
+        // if followed immediately by another navigation on iPhone
         navigationSplitCoordinator.setDetailCoordinator(nil, animated: animated)
+        navigationStackCoordinator.popToRoot(animated: false)
```
pixlwave, line 305: Does this still need to be `animated: false` with your tweak to use `if isPhone` below?

pixlwave (approved):
LGTM 👍

ElementX/Sources/Services/Notification/Proxy/NotificationItemProxy.swift
Velin92: yeah essentially the first just tells you if the room is an encrypted room, the second if the notification is still encrypted (so was not possible to encrypt it) They will be needed for the two sync loop, going to add some comments then

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
Velin92: i mean the animated value doesn't matter anymore , because the detailCoordinator is getting set to nil, before the pop to root, so in reality doesn't really matter the animation value, I guess we can just keep it as false.

PR #1098, opened 2023-06-16
App Polish

pixlwave:
First phase of app polish work. This PR makes the following tweaks:

- Skeleton avatar colour on app switching.
- Bug report icon and string.
- Semi-bold font on room list timestamps.
- Room attachments button spacing and alignment.
- Replace link tint with semi-bold font in AnalyticsSettingsScreen.
- Semi-bold font on Invites button.
- Add padding to media attachment sheet.
- Message composer header layout tweaks.
- Remove indicator after changing homeserver with an MXID.
- Change copy permalink string.
- Formatting of link in Analytics Prompt.
- Only semi-bold the room timestamp when unread.
- Tweak the colour of the date separator.
- Add activity indicator to toast style user indicator.

### DIFF
--- a/ElementX/Sources/Other/UserIndicator/UserIndicator.swift
+++ b/ElementX/Sources/Other/UserIndicator/UserIndicator.swift
@@ -18,12 +18,11 @@ import Combine
 import Foundation
 
 enum UserIndicatorType: Equatable {
-    case toast
-    case modal(interactiveDismissDisabled: Bool)
+    case toast(progress: UserIndicator.Progress?)
+    case modal(progress: UserIndicator.Progress?, interactiveDismissDisabled: Bool)
     
-    static var modal: Self {
-        .modal(interactiveDismissDisabled: false)
-    }
+    static var toast: Self { .toast(progress: .none) }
+    static var modal: Self { .modal(progress: .indeterminate, interactiveDismissDisabled: false) }
 }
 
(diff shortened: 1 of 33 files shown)


ElementX/Sources/Other/UserIndicator/UserIndicator.swift
```diff
@@ -35,32 +34,51 @@ struct UserIndicator: Equatable, Identifiable {
...
     }
 
-    enum LoaderType {
-        case unknownProgress
-        case progress(ProgressPublisher)
+    enum Progress: Equatable {
+        static func == (lhs: UserIndicator.Progress, rhs: UserIndicator.Progress) -> Bool {
+            switch (lhs, rhs) {
+            case (.indeterminate, .indeterminate): return true
+            case (.published, .published): return true // Not technically correct but close enough for now.
```
Velin92, line 41: maybe you could get the two ProgressPublisher from both values, and compare them by reference?
```
 case (.published(let lhsPublisher), .published(let rhsPublisher)): return lhsPublisher === rhsPublisher
```

Velin92 (approved):
LGTM

ElementX/Sources/Other/UserIndicator/UserIndicator.swift
pixlwave: Nice idea 😎

PR #1100, opened 2023-06-17
Add analytics tracking for room creation

Johennes:
This adds screen and event tracking for room and DM creation.

### DIFF
--- a/ElementX/Sources/Services/Analytics/Analytics.swift
+++ b/ElementX/Sources/Services/Analytics/Analytics.swift
@@ -125,4 +125,10 @@ extension Analytics {
         let event = AnalyticsEvent.MobileScreen(durationMs: milliseconds, screenName: screen.screenName)
         client.screen(event)
     }
+
+    /// Track the creation of a room
+    /// - Parameter isDM: true if the created room is a direct message, false otherwise
+    func trackCreatedRoom(isDM: Bool) {
+        capture(event: AnalyticsEvent.CreatedRoom(isDM: isDM))
+    }
 }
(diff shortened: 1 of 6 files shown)


Johennes:
CC @giomfo because I think you might have had this on your list, too.

ElementX/Sources/Services/Analytics/Analytics.swift
```diff
@@ -125,4 +116,11 @@ extension Analytics {
         let event = AnalyticsEvent.MobileScreen(durationMs: milliseconds, screenName: screen.screenName)
         client.screen(event)
     }
+
+    /// Capture an event in the `client`.
+    /// - Parameter event: The event to capture.
+    func capture(event: AnalyticsEventProtocol) {
```
pixlwave, line 122: Please can we keep this private, instead adding public methods for tracking events like we do on EI:
[trackCreatedRoom](https://github.com/vector-im/element-ios/blob/2831265badf17c77f64e228b5372beb150d8050a/Riot/Modules/Analytics/Analytics.swift#L363)

This has the advantage of not needing to import AnalyticsEvents everywhere we're going to track things so it should be easier for forks to customise analytics with different schemas.

pixlwave (changes requested):
Nice, its good to see PostHog starting to be used!

ElementX/Sources/Services/Analytics/Analytics.swift
Johennes: Ah ok. Was already wondering what the additional abstraction layer in the old app was good for. Makes sense and have changed.

pixlwave (approved):
LGTM 😎

ElementX/Sources/Services/Analytics/Analytics.swift
pixlwave: Super thanks

PR #1109, opened 2023-06-19
Loading state functions

stefanceriu:
* Prevent invites from stealing away the roomList's state listener
* Retract all user indicators when tearing down the user session


alfogrillo (approved):
LGTM!

PR #1112, opened 2023-06-19
Stefan/various UI tweaks

stefanceriu:
* bigger invites button
* persistent highlighting of selected room list items
* stop showing room when focusing the search field with an empty query

### DIFF
--- a/ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
+++ b/ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
@@ -14,6 +14,7 @@
 // limitations under the License.
 //
 
+import Combine
 import Compound
 import SwiftUI
 
@@ -49,7 +50,7 @@ struct HomeScreen: View {
             if context.viewState.roomListMode == .skeletons {
                 LazyVStack(spacing: 0) {
                     ForEach(context.viewState.visibleRooms) { room in
-                        HomeScreenRoomCell(room: room, context: context)
+                        HomeScreenRoomCell(room: room, context: context, isSelected: false)
                             .redacted(reason: .placeholder)
                     }
                 }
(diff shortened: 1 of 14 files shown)


ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomList.swift
```diff
@@ -0,0 +1,54 @@
...
+    @Environment(\.isSearching) var isSearching
+    
+    @ObservedObject var context: HomeScreenViewModel.Context
+    
+    var body: some View {
+        if isSearching, context.searchQuery.count == 0 {
+            EmptyView()
+        } else {
+            ForEach(context.viewState.visibleRooms) { room in
+                Group {
```
pixlwave, line 29: Also here, I don' think this `Group` is doing anything?

pixlwave (approved):
LGTM ✨

ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
```diff
@@ -25,57 +26,38 @@ struct HomeScreen: View {
...
         ScrollView {
             if context.viewState.showSessionVerificationBanner {
                 sessionVerificationBanner
             }
-
+            
             if context.viewState.hasPendingInvitations {
-                HomeScreenInvitesButton(title: L10n.actionInvitesList, hasBadge: context.viewState.hasUnreadPendingInvitations) {
-                    context.send(viewAction: .selectInvites)
+                ZStack {
```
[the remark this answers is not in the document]
stefanceriu: The ZsStack has a max width of infinity with trailing alignment and the button only has some padding on it. This is so that the button doesn't go full width and sticks to the trailing anchor. I couldn't figure out how to do that without introducing an extra layer

ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomList.swift
stefanceriu: That was already there, but I can't say I remember why 🤔

pixlwave (approved):

Issue #1036, opened 2023-06-06
Room list and timeline polish


manuroe:
We want to polish those 2 core screens before going to the app store.

PR #1118, opened 2023-06-20
App polish part 2

pixlwave:
This PR makes the following changes as part of #1036.

- Fix sheet backgrounds in dark mode.
- Fix background colours for reaction buttons.
- Reply view layout.
- Update separator colours.
- Don't show sender name/avatar in DMs.
- Fix inconsistent line heights with formatted body text.
- Use plain text for reply
- Increase tappable size of collapsed state changes button.
- Blockquote layout and text colour.
- Tap to expand the topic in room details.
- Change the topic and security font size in room details.
- Add cancel button when inviting someone to an existing room.
- Reword Add Friends to Add People in start chat screen.

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
+++ b/ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
@@ -20,6 +20,7 @@ import SwiftUI
 struct TimelineItemBubbledStylerView<Content: View>: View {
     @EnvironmentObject private var context: RoomScreenViewModel.Context
     @Environment(\.timelineGroupStyle) private var timelineGroupStyle
+    @Environment(\.isEncryptedOneToOneRoom) private var isEncryptedOneToOneRoom
     
     private let feedbackGenerator = UIImpactFeedbackGenerator(style: .heavy)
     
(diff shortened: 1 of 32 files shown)


ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
```diff
@@ -20,6 +20,7 @@ import SwiftUI
 struct TimelineItemBubbledStylerView<Content: View>: View {
     @EnvironmentObject private var context: RoomScreenViewModel.Context
     @Environment(\.timelineGroupStyle) private var timelineGroupStyle
+    @Environment(\.isEncryptedOneToOneRoom) private var isEncryptedOneToOneRoom
```
Velin92, line 23: why using the environment? We also have the context available of the view model of the room, isn't that already holding the state of the encrypted one to one room?

Velin92 (approved):
LGTM, one comment

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
pixlwave: Oh yeah, I didn't think about that and just followed the others. Was helpful for previewing but I'm done with that now. Will update.

PR #1120, opened 2023-06-20
Fixes #1102 - Crashes when opening the invites screen

### LINKED ISSUE #1102: app crashes on opening invites
### Steps to reproduce

1. select invites in room list
2. app crashes

see sentry

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

242

### Homeserver

_No response_

### Will you send logs?

Yes

stefanceriu:
For some reason that still eludes me opening invites crashed on opening the screen while accessing the view state, **only** on release builds.
The fix is to make them non-optional and, while I have no idea why that works, it's simple enough to not warrant any further investigation


alfogrillo (approved):
Strange issue. 🤔
LGTM anyway

Issue #1097, opened 2023-06-16
Add missing posthog events


giomfo:
### Steps to reproduce

The following posthog events are needed for the EX with at least the listed sub-fields:

- [x] [Composer](https://github.com/matrix-org/matrix-analytics-events/blob/main/schemas/Composer.json)
   - isEditing
   - isReply

- [x] [ViewRoom](https://github.com/matrix-org/matrix-analytics-events/blob/main/schemas/ViewRoom.json)
   - isDM

- [x] [CreatedRoom](https://github.com/matrix-org/matrix-analytics-events/blob/main/schemas/CreatedRoom.json)
   - isDM

- [x] [JoinedRoom](https://github.com/matrix-org/matrix-analytics-events/blob/main/schemas/JoinedRoom.json)
  - isDM

- [x] screen (I would consider [MobileScreen](https://github.com/matrix-org/matrix-analytics-events/blob/main/schemas/MobileScreen.json))
   - screenName
   - durationMs

### Will you send logs?

Yes

giomfo:
linked to [android-x issue](https://github.com/vector-im/element-x-android/issues/624)

giomfo:
@nimau FYI a PR has been already prepared for RoomCreated event: https://github.com/vector-im/element-x-ios/pull/1100

PR #1126, opened 2023-06-21
Add posthog events (#1097)

nimau:
This PR add some missing posthog events as requested by #1097 

Added events:
- `Composer`
- `ViewRoom`
- `JoinedRoom`

Added screens:
- `invites`
- `roomDetails`
- `roomMembers`
- `roomSettings`
- `user`

### DIFF
--- a/ElementX/Sources/Services/Analytics/Analytics.swift
+++ b/ElementX/Sources/Services/Analytics/Analytics.swift
@@ -148,4 +148,15 @@ extension Analytics {
             await capture(event: AnalyticsEvent.ViewRoom(activeSpace: nil, isDM: roomProxy.isDirect, isSpace: roomProxy.isSpace, trigger: nil, viaKeyboard: nil))
         }
     }
+    
+    /// Track the action of joining a room
+    func trackJoinedRoom(_ roomProxy: RoomProxyProtocol) {
+        Task {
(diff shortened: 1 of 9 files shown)


ElementX/Sources/Services/Analytics/Analytics.swift
```diff
@@ -131,4 +131,32 @@ extension Analytics {
...
+    ///   - inThread: whether the composer is used in a Thread
+    ///   - isEditing: whether the composer is used to edit a message
+    ///   - isReply: whether the composer is used to reply a message
+    ///   - startsThread: whether the composer is used to start a new thread
+    func trackComposer(inThread: Bool, isEditing: Bool, isReply: Bool, startsThread: Bool?) {
+        capture(event: AnalyticsEvent.Composer(inThread: inThread, isEditing: isEditing, isReply: isReply, startsThread: startsThread))
+    }
+    
+    /// Track the presentation of a room
+    func trackViewRoom(_ roomProxy: RoomProxyProtocol) {
```
stefanceriu, line 146: While I can understand the appeal for this I don't think Analytics has any business knowing about a `RoomProxyProtocol`. Can we pass individual properties instead?

stefanceriu (changes requested):

ElementX/Sources/Services/Analytics/Analytics.swift
nimau: Ok done :)

stefanceriu (approved):
Looking good 👏

PR #1130, opened 2023-06-22
Media upload progress

stefanceriu:
This PR adds support for reporting media upload progress. It also refactors the UserIndicators to use a CurrentValuePublisher instead of the ProgressTracker

https://github.com/vector-im/element-x-ios/assets/637564/be52ac4b-a3d3-488c-88fe-2cf6941204ae


ElementX/Sources/Other/ProgressTracker.swift
pixlwave, line 1: Buh bye 👋

ElementX/Sources/Other/UserIndicator/UserIndicator.swift
```diff
@@ -30,13 +30,13 @@ struct UserIndicator: Equatable, Identifiable {
...
             case (.indeterminate, .indeterminate): return true
-            case (.published(let lhsPublisher), .published(let rhsPublisher)): return lhsPublisher === rhsPublisher
+            case (.published(let lhsPublisher), .published(let rhsPublisher)): return lhsPublisher.value == rhsPublisher.value
             default: return false
             }
         }
         
         case indeterminate
-        case published(ProgressPublisher)
+        case published(CurrentValuePublisher<Double, Never>)
```
pixlwave, line 39: Should we be correct here and use `Progress` instead of `Double`?

ElementX/Sources/Other/UserIndicator/UserIndicator.swift
```diff
@@ -30,13 +30,13 @@ struct UserIndicator: Equatable, Identifiable {
         static func == (lhs: UserIndicator.Progress, rhs: UserIndicator.Progress) -> Bool {
             switch (lhs, rhs) {
             case (.indeterminate, .indeterminate): return true
-            case (.published(let lhsPublisher), .published(let rhsPublisher)): return lhsPublisher === rhsPublisher
+            case (.published(let lhsPublisher), .published(let rhsPublisher)): return lhsPublisher.value == rhsPublisher.value
```
pixlwave, line 33: Oh yeah, that's actually more correct isn't it 🙌

ElementX/Sources/Services/Room/RoomProxy.swift
```diff
@@ -611,3 +619,17 @@ private class RoomTimelineListener: TimelineListener {
         onUpdateClosure(diff)
     }
 }
+
+private class UploadProgressListener: ProgressWatcher {
+    private let onUpdateClosure: (Double) -> Void
```
pixlwave, line 624: Minor, and just thinking out loud here: maybe this would work instead of the dispatch queue?
```suggestion
    private let onUpdateClosure: @MainActor (Double) -> Void
```

pixlwave (approved):
Very nice 😎

ElementX/Sources/Services/Room/RoomProxy.swift
stefanceriu: Maybe, definitely worth a shot 👍

callumu:
Nice! Is there a reason for the pause after the progress bar completes?

callumu:
It would be ideal if as soon as the progress bar completes, then the next screen gets pushed

callumu:
Does that mean that the message isn't actually sending while the progress bar is animating?

stefanceriu:
You first upload the thumbnail, then upload the image and then send the room message with the links to those. It's a 3 step process.

callumu:
Okay thanks for clarifying

PR #1131, opened 2023-06-22
SDK Bump to 1.0.80 and push notifications filtering

Velin92:
Now notifications that are able to be decrypted will also be filtered out based on the push context.


alfogrillo (approved):
LGTM 👍🏻

PR #1132, opened 2023-06-22
Stop asserting integration test run time & move to performance testing runner.

michaelkaye:
Performance tests like these are not best run on a 100% success rate. We should not assert a failure due to a timeout and instead capture the range of times that the test runs in.

A single failure means nothing; we need to look at patterns instead, and allow ourselves to work around networking glitches or cpu spikes on the test server.


michaelkaye:
We also aim to persist the device logs as well as the test logs here.

stefanceriu:
In my defense the values we assert on are averages over 3 runs (+ a warmup one)

michaelkaye:
This now additionally moves the integration tests over to the performance testing GHA runner in the office.

.github/workflows/integration-tests.yml
```diff
@@ -2,14 +2,14 @@ name: Integration tests
 
 on:
   schedule:
-    - cron:  '0 2 * * 1-5'
```
michaelkaye, line 5: We own the hardware so can run this ... as often as we like ^^

michaelkaye:
There were some unit test failures but i believe they're unrelated; have updated to latest develop and hopefully that should clear it up.

michaelkaye:
https://github.com/vector-im/element-x-ios/actions/runs/5345012631 contains a demonstration of the integration tests running (against the branch).

https://vector-im.github.io/element-x-ios/performance/ shows the improvement in performance.

https://michaelkaye.github.io/element-x-ios/performance/ shows historical data (hourly for ~24h) is also more consistent over time.

Velin92 (approved):

PR #1138, opened 2023-06-22
Room list service

stefanceriu:
Adopt latest RoomListService changes
* discrete lists
* new list states (abstracted away from sliding sync) that include the total number of rooms

### DIFF
--- a/ElementX/Sources/Services/Client/ClientProxy.swift
+++ b/ElementX/Sources/Services/Client/ClientProxy.swift
@@ -26,7 +26,7 @@ class ClientProxy: ClientProxyProtocol {
     private let mediaLoader: MediaLoaderProtocol
     private let clientQueue: DispatchQueue
         
-    private var roomListService: RoomList?
+    private var roomListService: RoomListService?
     private var roomListStateUpdateTaskHandle: TaskHandle?
     
     var roomSummaryProvider: RoomSummaryProviderProtocol?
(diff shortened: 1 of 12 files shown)


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -400,8 +400,7 @@ class ClientProxy: ClientProxyProtocol {
                     
                     Task {
                         // Subscribe to invites later as the underlying SlidingSync list is only added when entering AllRooms
-                        await self.inviteSummaryProvider?.subscribeIfNecessary(entriesFunction: roomListService.invites(listener:),
-                                                                               entriesLoadingStateFunction: nil)
+                        try await self.inviteSummaryProvider?.setRoomList(roomListService.invites())
```
pixlwave, line 403: Is it intentional to let the Task swallow up any errors? Might be better to do/catch/log?

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProviderProtocol.swift
```diff
@@ -20,9 +20,25 @@ import MatrixRustSDK
...
+    var isLoaded: Bool {
+        switch self {
+        case .loaded:
+            return true
+        default:
+            return false
+        }
+    }
+    
+    var totalNumberOfRooms: UInt {
```
pixlwave, line 34: Given this is only used to check for `state.isLoaded && state.totalNumberOfRooms == 0`, I think it would be safer to make this property `hasNoRooms` and do the check in here. Otherwise its possible for someone to check `totalNumberOfRooms` without checking `isLoaded` and introduce bugs.

pixlwave (approved):
Oh this is nicer 👍

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProviderProtocol.swift
stefanceriu: Very good point!

ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu: Nop, absolutely totally missed it. Talked about it offline, will re-enable the swiftlint rule to prevent problems like this in the future

PR #1141, opened 2023-06-23
Keyboard is kept dismissed when dismissing the actions or the attachments menu

Velin92:
https://github.com/vector-im/element-x-ios/assets/34335419/4fe02197-5029-4d0b-836e-59a8fb0f5e4c


pixlwave (approved):
Nice, thanks for figuring this one out 👏

PR #1144, opened 2023-06-23
ServiceLocator cleanup

### LINKED ISSUE #920: Stop directly accessing the ServiceLocator
The service locator should always be passed in the constructor from at least one layer above to allow for cleaner code and easier mocking

stefanceriu:
Fixes #920

While using the ServiceLocator directly sure is cozy the rule of thumb is that any dependency should be passed at least from one level above. This PR takes care of that.


pixlwave (commented):
I love the potential of this change, but I'm not sure it really fixes https://github.com/vector-im/element-x-ios/issues/920 in its current form. I stopped adding comments about "should we pass this in" because I was becoming boring 🙈, but at the very least I see issues with things doing `ServiceLocator.shared.analytics` when we get into a multi-user world.

As I see it, anything downstream from the user flow coordinator shouldn't simply access the services via the singleton (whether view, coordinator or view model). 2 ideas of how to continue:
1. Rename `ServiceLocator.shared` into `ServiceLocator.mock` so its clearly only for tests and previews. Pass the services (or even an instance of the locator) through all the coordinators and down into the view models, and any necessary models - feels like a *lot* of work.
2. Make functions like `ServiceLocator.analytics(for: userSession)` to get the service (or if we prefer more generally a service locator) for a specific user session/ID and continue to use static access everywhere. Would involve all the things knowing about the current userID though so I guess still plenty of work.

Does that actually make any sense?

stefanceriu:
So we talked about all of this offline. The key takeaways are that the ServiceLocator isn't really working out for us (especially in a multi-account env) and in order to not complicate things we've agreeed that for now we should just go with plain old constructor dependency injection and re-assess later. For now we extracted everything one level above and we can continue doing this until the service locator is no more, as time allows it.

As an example for why this will be better let's take the global user indicator controller. We know that atm it won't work on screens presented as sheets, and if by any reason one of the screens turns into that there's no indicator that the breaking will occur, as the interface doesn not say anything about the userNotificationController.

PR #1145, opened 2023-06-23
Reaction highlights

langleyd:
Fixes https://github.com/vector-im/element-x-ios/issues/1146

## Whats in this PR?
- This makes use of the newly exposed reaction sender ids in the sdk
- Updates the timeline to use the senders to highlight the current users reactions
- Fixes the style of the reactions when highlighted
- Updates the message bottom sheet to only show the reactions as highlighted if it's our own id(previously it would show if anybody had added that reaction.)


ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineItemStatusView.swift
```diff
@@ -58,7 +58,7 @@ struct TimelineItemStatusView: View {
             TimelineDeliveryStatusView(deliveryStatus: .sending)
         case .sent:
             TimelineDeliveryStatusView(deliveryStatus: .sent)
-        case .none:
+        case .none, .cancelled:
```
alfogrillo, line 61: Are we sure about a `.cancelled` being treated as sent if is the last item in the timeline?
Wdyt @stefanceriu?

ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineReactionsView.swift
```diff
@@ -46,21 +46,38 @@ struct TimelineReactionButton: View {
...
     var backgroundShape: some InsettableShape {
         RoundedRectangle(cornerRadius: 12, style: .continuous)
     }
+    
+    var textColor: Color {
+        reaction.isHighlighted ? Color.compound.textPrimary : .compound.textSecondary
+    }
+    
+    var overlayBackgroundColor: Color {
+        reaction.isHighlighted ? Color.compound._bgBubbleOutgoing : .compound._bgBubbleIncoming
```
alfogrillo, line 75: Selecting `_bgBubbleOutgoing`/`_bgBubbleIncoming` depending on the `isHighlighted` flag looks weird.
I guess the outgoing/incoming bubble state doesn't play a role here right?

ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineReactionsView.swift
```diff
@@ -46,21 +46,38 @@ struct TimelineReactionButton: View {
...
         .background(
             backgroundShape
-                .strokeBorder(reaction.isHighlighted ? Color.compound.textSecondary : .compound.bgCanvasDefault, lineWidth: 2)
-                .background(reaction.isHighlighted ? Color.compound.textPrimary.opacity(0.1) : .compound._bgReactionButton, in: backgroundShape)
+                .fill(Color.compound.bgCanvasDefault)
+                .overlay(
+                    backgroundShape
+                        .strokeBorder(overlayBorderColor)
+                        .background(overlayBackgroundColor, in: backgroundShape)
+                        .padding(1)
```
alfogrillo, line 60: I'm not sure why the previous `strokeBorder` wasn't enough. Is it because of the padding?

alfogrillo (commented):
I don't see any update on snapshot references.
If we have ui tests rendering this piece of UI they are probably gonna fail.

ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineReactionsView.swift
langleyd: They are just the closest colours I could see in compound that match the designs. If you look at them the figma the highlighted color looks the same as the bubble outgoing and the non-highlighted looks the same as the incoming. but yea, not dependant on whether it's actually and incoming/outgoing message.

ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineReactionsView.swift
langleyd: There is a kind of double border in the designs, with the white wrapping the dark in the highlighted state:


Before introducing my code the white border was replaced by the dark, and wasn't wrapped with the white.

langleyd: I think the padding should be 2 actually, looking at it now.

ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineReactionsView.swift
langleyd: Happy to change if you see a better way of doing things though? :)

ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineItemStatusView.swift
alfogrillo: fyi: I applied the change [here](https://github.com/vector-im/element-x-ios/pull/1147)

ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineReactionsView.swift
alfogrillo: > but yea, not dependant on whether it's actually and incoming/outgoing message.

Ok thanks for the clarification. I don't have better suggestions.
Maybe @pixlwave has since he has all the colors in his mind. :D

ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineReactionsView.swift
alfogrillo: I think you can do the same easier like this:

```swift
        .background(
            backgroundShape
                .fill(overlayBackgroundColor)
                .overlay(
                    backgroundShape
                        .strokeBorder(overlayBorderColor)
                )
        )
```
and
```swift
    var overlayBorderColor: Color {
        reaction.isHighlighted ? Color.compound.textSecondary : Color.clear
    }
```

**Result**
demo

Is it what you want?

langleyd: But over the bubble that gives you this:
image

with no white boarder on the outside of the dark border. What I have uses padding to inset the dark border inside the white one:
image

ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineReactionsView.swift
pixlwave: (For context, we're awaiting more semantic tokens to cover these before they're finalised hence the use of temporary ones)

ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineReactionsView.swift
pixlwave: You can also do `backgroundShape.inset(by: 2).fill()`/`.strokeBorder()` if that's any help :)

ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineReactionsView.swift
pixlwave: @langleyd These are the correct colours from @callumu

Added by someone else: `.bgSubtleSecondary`

 Added by you: `.bgSubtlePrimary` with a border colour of `.borderInteractivePrimary`

ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineReactionsView.swift
alfogrillo: > But over the bubble that gives you this:

I edited the comment once, not sure you picked up the final suggestion.

This is what I see on a real timeline:

Screenshot 2023-06-23 alle 18 00 30

Is it good?

langleyd: Works for me, thanks!

langleyd:
> I don't see any update on snapshot references. If we have ui tests rendering this piece of UI they are probably gonna fail.

Thanks @alfogrillo, I haven't had to do this before. How does on do this? any docs?

alfogrillo:
> > I don't see any update on snapshot references. If we have ui tests rendering this piece of UI they are probably gonna fail.
> 
> Thanks @alfogrillo, I haven't had to do this before. How does on do this? any docs?

Mmmh I don't think we have any docs.
Impacted tests should be just `RoomScreenUITests`.

We need to check they pass:

- iPhone 14
- iPad (9th generation)

If they pass, we are good.
Otherwise we you need to delete failing reference screenshots at `element-x-ios/UITests/Sources/__Snapshots__/Application` and record new ones.
To record new ones just run the tests again after the deletion of old screenshots.
After that you should see the new screenshot references in the git working area.

alfogrillo (approved):
LGTM.

langleyd:
I added a snapshot test for reactions as we didn't have one that included them before.
I also used the info you provided as to documentation for snapshot tests in the contributing guide.
Thanks for your help today. @alfogrillo

alfogrillo:
> I added a snapshot test for reactions as we didn't have one that included them before.
> I also used the info you provided as to documentation for snapshot tests in the contributing guide.
> Thanks for your help today. @alfogrillo

Cool, thanks for documenting it! :)

Issue #1117, opened 2023-06-20
Emotes aren't rendered properly in push


ara4n:
### Steps to reproduce

1. receive a push for an emote
2. wince at the emoji in the notif rather than a profile

![image](https://github.com/vector-im/element-x-ios/assets/1294269/d37d9a16-28ad-46aa-85d8-4d364d9ffd2d)

#### What did you expect?

emotes to be rendered correctly

#### What happened instead?

placeholder 🫥

### Application version

244

### Will you send logs?

No

PR #1152, opened 2023-06-25
Render emote notifications like in the timeline

Johennes:
This uses the `* DISPLAYNAME ...` format that's also used in the timeline to render emotes. It's likely not exactly perfect, but emotes are used fairly widely and this at least makes their notifications look somewhat less WIP than the 🫥 emoji prefix.

Fixes: #1117

### DIFF
--- a/ElementX/Sources/Services/Notification/Proxy/NotificationItemProxy.swift
+++ b/ElementX/Sources/Services/Notification/Proxy/NotificationItemProxy.swift
@@ -360,7 +360,7 @@ extension NotificationItemProxyProtocol {
     private func processEmote(content: EmoteMessageContent,
                               mediaProvider: MediaProviderProtocol?) async throws -> UNMutableNotificationContent {
         let notification = try await processCommon(mediaProvider: mediaProvider)
-        notification.body = "🫥 " + content.body
+        notification.body = L10n.commonEmote(senderDisplayName ?? roomDisplayName, content.body)
 
         return notification
     }
--- a/changelog.d/1117.feature
+++ b/changelog.d/1117.feature
@@ -1,0 +1,1 @@
+Render emote notifications like in the timeline


ElementX/Sources/Services/Notification/Proxy/NotificationItemProxy.swift
```diff
@@ -360,7 +360,7 @@ extension NotificationItemProxyProtocol {
     private func processEmote(content: EmoteMessageContent,
                               mediaProvider: MediaProviderProtocol?) async throws -> UNMutableNotificationContent {
         let notification = try await processCommon(mediaProvider: mediaProvider)
-        notification.body = "🫥 " + content.body
+        notification.body = L10n.commonEmote(senderDisplayName ?? roomDisplayName, content.body)
```
Velin92, line 363: the senderDisplay name is already in the title of the notification, are we sure we want to include it also in the body?

Johennes: This is what the old iOS app does at least and emotes in the timeline on Element X look similar. I haven't consulted with design on this. My main goal was to make it look less WIP.

Velin92: Okay I guess we can use this solution for now then, let's just check if this temporary solution is okay for @callumu 

This will display for an emote notification the following message layout:
{Sender Name}
* {Sender Name} {Emote}

callumu: Would this vary from DM to room?

Velin92: in room would be the same thing, but we would have the room as subtitle:
{Sender Name}
{Room Name}
* {Sender Name} {Emote}

Velin92: i checked with @callumu the preferred solution would be:
Reacted with {emote} 

for the body.

I have made a new string notification_reaction_body string on localazy, @Johennes could you. use that one instead?

Velin92 (approved):
for now I guess we can keep the old behaviour will check with @callumu  what would be the preferred emote to have in the future

ElementX/Sources/Services/Notification/Proxy/NotificationItemProxy.swift
Velin92: disregard my previous message, confused emote with reaction

Velin92 (approved):

ElementX/Sources/Services/Notification/Proxy/NotificationItemProxy.swift
Johennes: You don't want to know how often this has already happened to me. 😅

Thanks for doing the design coordination that I should have done in the first place. 🙏

PR #1158, opened 2023-06-26
Reaction summary and flow layout for reactions

langleyd:
## What's in this PR
- It removes the old single line of reactions that would scroll off the screen
- It introduces a flow layout where reactions wrap on to the next line.
  - They also align left-to-right or right-to-left when in bubble mode on the timeline (and just left-to-right in plain/modern mode).
- Reactions are now sorted by count
- Adds the reactions summary view via long-press on a reaction
  - This shows the name and id and avatar of who left the reactions.
 
 The reason I added the flow layout to this PR was that the long press on the reaction button was difficult to achieve when the reaction button was contained within a scrollview and also needs to be able to be tapped. Removing the scroll view simplified things.

## Whats left to do in this PR.
- [x] Safe inset at the bottom of the summary view
- [x] Scroll to the initially selected emoji when summary view first displayed

## Whats left to do in a future PR.
- We need to expose the timestamp of reactions from the rust sdk
- This should be displayed in the summary view
- I think this can be used as a secondary sort after count
- We need to add the more/less button that collapses the reactions down to less lines.
- Add the smiley "Add More" button that allows the user quickly add another emoji.


ElementX/Sources/Mocks/AggregatedReactionMock.swift
```diff
@@ -0,0 +1,56 @@
...
+import Foundation
+
+extension AggregatedReaction {
+    fileprivate static func mockIds(_ count: Int) -> [String] {
+        Array(1...count).map {
+            "@user\($0):matrix.org"
+        }
+    }
+    
+    fileprivate static var alice: String {
```
alfogrillo, line 26:
```suggestion
    private static var alice: String {
```

ElementX/Sources/Screens/RoomScreen/View/Supplementary/ReactionsSummaryView.swift
```diff
@@ -0,0 +1,136 @@
...
+            }
+            .tabViewStyle(.page(indexDisplayMode: .never))
+        }
+        .presentationDetents([.medium])
+        .presentationBackground(Color.compound.bgCanvasDefault)
+        .presentationDragIndicator(.visible)
+    }
+}
+
+struct ReactionSummaryButton: View {
```
alfogrillo, line 74:
```suggestion
private struct ReactionSummaryButton: View {
```

ElementX/Sources/Screens/RoomScreen/View/Supplementary/ReactionsSummaryView.swift
```diff
@@ -0,0 +1,136 @@
...
+                                avatarSize: .user(on: .timeline),
+                                imageProvider: imageProvider)
+            VStack(alignment: .leading) {
+                Text(displayName)
+                    .font(.compound.bodyMDSemibold)
+                Text(sender)
+                    .font(.compound.bodySM)
+                    .foregroundColor(.compound.textSecondary)
+            }
+            Spacer()
```
alfogrillo, line 123: We try to don't use `Spacer`.
I think this layout should do the same thing:

```swift
        HStack {
            LoadableAvatarImage(url: member?.avatarURL,
                                name: displayName,
                                contentID: sender,
                                avatarSize: .user(on: .timeline),
                                imageProvider: imageProvider)
            VStack(alignment: .leading) {
                Text(displayName)
                    .font(.compound.bodyMDSemibold)
                Text(sender)
                    .font(.compound.bodySM)
            }
        }
        .frame(maxWidth: .infinity, alignment: .leading)
        .padding(.vertical, 8)
  ```

File: ElementX/Sources/Services/Timeline/TimelineItemContent/AggregratedReaction.swift [hunk 4]
```diff
@@ -18,10 +18,22 @@ import Foundation
 
 /// Represents all reactions of the same type for a single event.
 struct AggregatedReaction: Hashable {
+    /// The user id of the current user.
+    let currentUserId: String
```
alfogrillo, line 22: We use the `owner` wording for the ID of the account.
maybe `let accountOwnerID: String`?
The `currentUserId` is always there even if the user didn't react for this cluster right?

alfogrillo (commented):
Great addition! 👏🏻
Left several small comments

ElementX/Sources/Screens/RoomScreen/View/Supplementary/ReactionsSummaryView.swift
```diff
@@ -0,0 +1,136 @@
...
+            ScrollView(.horizontal, showsIndicators: false) {
+                ScrollViewReader { scrollView in
+                    HStack {
+                        ForEach(reactions, id: \.self) { reaction in
+                            ReactionSummaryButton(reaction: reaction, highlighted: selectedReactionKey == reaction.key) { key in
+                                withAnimation(.easeInOut) {
+                                    selectedReactionKey = key
+                                }
+                            }
+                            .id(reaction.key)
```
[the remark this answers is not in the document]
langleyd: Looks like it's needed in my tests.
Steps without this line:
1. Add about 10 reactions
image
2. Long press on the the last reaction to bring up the summary
3. Observed: When the summary shows the selected item is off the screen.
image

ElementX/Sources/Services/Timeline/TimelineItemContent/AggregratedReaction.swift
langleyd: Correct yea

alfogrillo (commented):
LGTM! 👍🏻

langleyd:
I’ve noticed a bug on this where the long press from displaying the summary can cause the tap to toggle a reaction to fire

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
```diff
@@ -92,10 +92,12 @@ struct TimelineItemBubbledStylerView<Content: View>: View {
...
             
             if !timelineItem.properties.reactions.isEmpty {
-                TimelineReactionsView(reactions: timelineItem.properties.reactions,
-                                      alignment: alignment) { key in
+                TimelineReactionsView(reactions: timelineItem.properties.reactions) { key in
                     context.send(viewAction: .toggleReaction(key: key, eventID: timelineItem.id))
+                } showReactionSummary: { key in
+                    context.send(viewAction: .reactionSummary(itemID: timelineItem.id, key: key))
                 }
+                .environment(\.layoutDirection, timelineItem.isOutgoing ? .rightToLeft : .leftToRight)
```
pixlwave, line 100: Can't wait to have this land on EX 🤩
One small thought on this change: `LayoutDirection` is intended for RTL language handling, so whilst this change use is overall fine, it should also handle the RTL case too? E.g. something like

```
@Environment(\.layoutDirection) var layoutDirection: LayoutDirection
var reactionsLayoutDirection: LayoutDirection {
    guard isOutgoing else { return layoutDirection }
    return layoutDirection == .leftToRight ? .rightToLeft : .leftToRight
}
```

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemPlainStylerView.swift
```diff
@@ -115,10 +115,12 @@ struct TimelineItemPlainStylerView<Content: View>: View {
...
             
             if !timelineItem.properties.reactions.isEmpty {
-                TimelineReactionsView(reactions: timelineItem.properties.reactions,
-                                      alignment: .leading) { key in
+                TimelineReactionsView(reactions: timelineItem.properties.reactions) { key in
                     context.send(viewAction: .toggleReaction(key: key, eventID: timelineItem.id))
+                } showReactionSummary: { key in
+                    context.send(viewAction: .reactionSummary(itemID: timelineItem.id, key: key))
                 }
+                .environment(\.layoutDirection, .leftToRight)
```
pixlwave, line 123: And this one should probably just be left as the default :)

alfogrillo (approved):

ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineReactionsView.swift
```diff
@@ -40,17 +40,15 @@ struct TimelineReactionButton: View {
     let toggleReaction: (String) -> Void
     let showReactionSummary: (String) -> Void
     
+    @State private var didLongPress = false
+    
     var body: some View {
-        Button {
+        label.onTapGesture {
```
pixlwave, line 46: We're going to need to do something for accessibility here as it won't be exposed to VoiceOver as a button now, but that can a follow-up and tbh we need to do a sweep over the timeline anyway, because there are various long press and swipe gestures that probably need putting into a rotor.

pixlwave (approved):
Thanks for the RTL tweaks 😎

PR #1161, opened 2023-06-26
Handling cancelled as a sendingFailed case

Velin92:
There is no reason on the app side for which this state should be handled differently from the sendingFailed case.


stefanceriu (approved):
Ah right, we were wondering what the right solution here would be https://github.com/vector-im/element-x-ios/pull/1145#discussion_r1239883612

Issue #1155, opened 2023-06-25
Disable notification filtering temporarily


manuroe:
I am missing notifications including for messages in EX-Internal. EI displays them fine.

Doing more tests, I saw this pattern: if another user sends a message in a room I have not opened for a while, it is unlikely I will get a notification for this message. The problem is the same if the room is encrypted or not.
Once I open his room, I will receive further notifications.
Note that just seeing the room in the room list with the right last message is not enough to get the display of next notifications. I really need to open the timeline. This makes me think that we may have a problem with unknown room state to verify push rules correctly.

We should disable notification filtering for the release. It seems to complex to solve.

manuroe:
I change the title to be more directive.
We should create an issue for further investigation.

manuroe:
We should rollback https://github.com/matrix-org/matrix-rust-sdk/pull/2129/files#diff-f8239996bb4c6fc5d707bbdca5b86a9ebdfb87c07bc475d337969da9e17f5167R49-R52
I am missing notifications for DMs I am pretty sure I have the keys.

Velin92:
keys or not they should be displayed... for now I am going to set a boolean in the Notification API in the rust SDK, that will decide if the check should be done or not, then I will make a new release ASAP

Velin92:
requires a new SDK release, created a branch to disable the filtering behaviour for now here:
https://github.com/matrix-org/matrix-rust-sdk/pull/2145

PR #1162, opened 2023-06-26
SDK bump to 1.0.85-alpha and disable push rules filtering

Velin92:
This will address the following issue:
https://github.com/vector-im/element-x-ios/issues/1155


pixlwave (approved):
LGTM 👍

PR #1165, opened 2023-06-26
Handle waiting list error.

### LINKED ISSUE #1154: Implement `IO_ELEMENT_X_WAIT_LIST` error code on `/login`
As a user, I want to understand why I cannot login and how I can be part of this allow list defined on my homeserver.

pixlwave:
Adds a screen for new users who end up on the matrix.org waiting list as we scale the sliding sync backend. Closes #1154.

- [ ] Confirm final strings with product.
- [ ] Add strings to Localazy.
- [ ] Generate snapshots.

The screen will refresh each time you foreground the app, and then when allowed in, the flow will continue:

https://github.com/vector-im/element-x-ios/assets/6060466/e2363cd0-af83-45e4-ab2c-a96e51b58ed8


stefanceriu (approved):
lgtm 👍

PR #1174, opened 2023-06-27
Implement all emoji text message boosting

stefanceriu:
Screenshot 2023-06-27 at 12 34 19


Velin92 (approved):
My god this is tricky, I would write some tests for this unicode shenanigans to test them

Issue #1177, opened 2023-06-27
Fix reaction button hit area


langleyd:
## Description
Tapping the top of a reaction button on timeline does not work

## Steps 
1. Send a message
2. Add a reaction
3. Tap the top of the reaction
Expected: The reaction is removed
Observed: The reaction is not removed. If you long press the top of the reaction you can see the bubble is selected. T's the long press that is stealing the touch event.

PR #1178, opened 2023-06-27
Fix reaction hit area

langleyd:
Fixes https://github.com/vector-im/element-x-ios/issues/1177

Gestures are a bit of a mess in SwiftUI. The reaction view and `TimelineReactionsView` and `messageBubble` are siblings and if the reaction view doesn't have an `onTapGesture` the `onLongPress` on `messageBubble` will steal the touch events.


pixlwave (approved):
LGTM 🙌

PR #1183, opened 2023-06-27
Update strings and screenshots for waitlist screen.

pixlwave:
Follow-up to #1165 with the strings now in Localazy and snapshots added.


stefanceriu (approved):
👍

Issue #1169, opened 2023-06-26
Opening invites locks the app solid


ara4n:
### Steps to reproduce

1. open invites on an account with ~100 pending invites
2. app ui wedges entirely solid for 30s or more
3. invite details get displayed
4. app ui wedges entirely for another 30s

#### What did you expect?

invite ui which does’t wedge

#### What happened instead?

total ui lockup for 30s +

### Application version

254

### Will you send logs?

Yes

Johennes:
@alfogrillo could you take a look into this tomorrow? Unless we can show that this problem persists for accounts with less invites, I don't think we should treat it as a release blocker. But it'll be good to figure out where all the cycles go.

alfogrillo:
> @alfogrillo could you take a look into this tomorrow? Unless we can show that this problem persists for accounts with less invites, I don't think we should treat it as a release blocker. But it'll be good to figure out where all the cycles go.

Yep, having a look already.
@ara4n did you notice if the problem has been introduced recently (approximately in the last month)?
(Trying to understand if this is related with the introduction on the room list API)

alfogrillo:
Found an optimization point implemented [here](https://github.com/vector-im/element-x-ios/pull/1191). Let's see if this makes any difference.

ara4n:
> @ara4n did you notice if the problem has been introduced recently (approximately in the last month)?

no, i don't think i'd ever actually looked at my invites on EIX before a few days ago. 257 still has it; still test once this build ships.

ara4n:
Opening invites still locks the app solid; unsure why this was closed.

Johennes:
Removing the release blocker label to avoid confusion (we're past the release now).

alfogrillo:
Possibly fixed now by https://github.com/vector-im/element-x-ios/pull/1268.
@ara4n please let us know if it works for you now.
(it will be available in the tomorrow's nightly)

ara4n:
yup, it's fixed now - thanks :)

PR #1191, opened 2023-06-28
Optimize inviters fetch using task group

alfogrillo:
This PR optimizes the fetch of inviters by using a `TaskGroup` instead of creating a `Task` manually.
It also handles cancellations of previous tasks.
Fix https://github.com/vector-im/element-x-ios/issues/1169


stefanceriu (approved):
Looks sensible to me 👍

Issue #1101, opened 2023-06-17
unsent send receipts should immediately replace ticked send receipts


ara4n:
### Steps to reproduce

1. Send two messages one after another
2. The first one will be marked as sent (with a ticked send receipt circle)
3. The second one will still be marked as sending (empty send receipt circle).

The fact the first one was sent successfully can be inferred by the fact we are now sending the second however - we don’t need to see both circles, unless the first one hasn’t yet sent.  As a result, it makes the timeline pop around and look redundant and confusing:

![IMG_0535](https://github.com/vector-im/element-x-ios/assets/1294269/a8d3588f-1059-4663-9602-7b6132db6e05)

#### What did you expect?

Only show SRs if they are the most recent message you sent (or if they are stuck trying to send). 

#### What happened instead?

Duplicated SRs telling me useless information which disappear only when the second SR arrives, causing lots of timeline pops. Instead the SR should disappear as soon as you send a new msg.

You can also get dups where both are in sent state…

![IMG_0537](https://github.com/vector-im/element-x-ios/assets/1294269/462f5bdf-38af-4d91-a264-b2711f895809)

### Application version

242

### Will you send logs?

No

manuroe:
@amshakal what do you think about this proposal? It can be a solution to the problem reported on Friday.

nadonomy:
@manuroe btw @amshakal is out this week so anything timely best to tag @callumu.

ara4n:
> We also need to consider when there are several messages queued for sending: do we want to use a temporary line for each empty circles? Every time we use this temporary line, it makes the timeline clunky when it disappears.

personally I think the negative space caused by the various unsent SRs is a feature, rather than a bug: if you're on bad connectivity and all your messages are stacked up, a) it's your fault, b) you want to know about it, c) the 'unpopping' effect is quite satisfying when they finally send.

The thing i'm objecting to here is that we include (and then unpop) *sent* SRs, not just unsent ones.  Surely we should be copying EW's semantics, which feel fine?  I have a feeling that it's probably a bug that we're not.

ara4n:
> We also need to consider when there are several messages queued for sending: do we want to use a temporary line for each empty circles? Every time we use this temporary line, it makes the timeline clunky when it disappears.

[...]

PR #1193, opened 2023-06-28
Delivery status is displayed only for last message

Velin92:
After the discussione had [here](https://github.com/vector-im/element-x-ios/issues/1101) this is the preferred solution

https://github.com/vector-im/element-x-ios/assets/34335419/1f2fdf37-3784-4e6a-83e3-18019bd8a5a9


alfogrillo (approved):
LGTM

PR #1196, opened 2023-06-28
Prevent the app from crashing when multiple virtual timelineStart ite…

stefanceriu:
…ms appear in the timeline

This is an SDK bug and should be properly handled Rust side.


nimau (approved):
👍

PR #1200, opened 2023-06-28
Fix for flipped image notification placeholder on iPhone

Velin92:
This problem was happening only on device, but was working properly on macOS and simulator.

I added a function that flips the image only on iOS devices


pixlwave (approved):
LGTM 🙃🙂

PR #1207, opened 2023-06-28
Editing a failed echo, just cancels and resends it

Velin92:
https://github.com/vector-im/element-x-ios/assets/34335419/d00cab34-f754-49fb-9fad-f26d8833a9a2


pixlwave (approved):
LGTM if product are happy with the reordering 👍

PR #1208, opened 2023-06-28
Trim newlines from message previews.

pixlwave:
- Fixes a bug with replies where the message shows on the line after the sender's name.

This is the simple fix although I we should discuss whether it would make more sense to strip markdown from the preview by building the attributed string from the formatted body to get a clean string instead.

| Before | After |
| - | - |
| Simulator Screenshot - iPhone 14 - 2023-06-28 at 18 32 12 | Simulator Screenshot - iPhone 14 - 2023-06-28 at 18 31 43 |


ElementX/Sources/Services/Room/RoomSummary/RoomEventStringBuilder.swift
```diff
@@ -81,9 +81,9 @@ struct RoomEventStringBuilder {
         if let senderDisplayName = sender.displayName,
            let attributedSenderDisplayName = try? AttributedString(markdown: "**\(senderDisplayName)**") {
             // Don't include the message body in the markdown otherwise it makes tappable links.
-            return attributedSenderDisplayName + ": " + AttributedString(eventSummary)
+            return attributedSenderDisplayName + ": " + AttributedString(eventSummary.trimmingCharacters(in: .whitespacesAndNewlines))
```
stefanceriu, line 84: Should we maybe extract this `AttributedString(eventSummary.trimmingCharacters(in: .whitespacesAndNewlines))` one level above and use in both if branches?

stefanceriu (approved):
lgtm 👍 

> we should discuss whether it would make more sense to strip markdown from the preview by building the attributed string from the formatted body to get a clean string instead

It makes sense to me but it's a product decision in the end

PR #1210, opened 2023-06-29
Fixes #954 - Implement media upload cancelling

### LINKED ISSUE #954: There is no way to cancel an image upload
### Steps to reproduce

1. Upload an image on bad connectivity
2. Observe spurious Loading spinner (which should surely be Sending)
3. Hit cancel
4. nothing happens

#### What did you expect?
 cancel button should work
#### What happened instead?
it doesn’t

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

223

### Homeserver

_No response_

### Will you send logs?

No

stefanceriu:
Requires new SDK release

### DIFF
--- a/ElementX/Sources/Screens/MediaUploadPreviewScreen/MediaUploadPreviewScreenViewModel.swift
+++ b/ElementX/Sources/Screens/MediaUploadPreviewScreen/MediaUploadPreviewScreenViewModel.swift
@@ -15,6 +15,7 @@
 //
 
 import Combine
+import MatrixRustSDK
 import SwiftUI
 
 typealias MediaUploadPreviewScreenViewModelType = StateStoreViewModel<MediaUploadPreviewScreenViewState, MediaUploadPreviewScreenViewAction>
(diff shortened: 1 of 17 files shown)


ElementX/Sources/Screens/MediaUploadPreviewScreen/MediaUploadPreviewScreenViewModel.swift
```diff
@@ -67,22 +73,28 @@ class MediaUploadPreviewScreenViewModel: MediaUploadPreviewScreenViewModelType,
...
+            requestHandle?.cancel()
             callback?(.dismiss)
         }
     }
     
     // MARK: - Private
     
     private func sendAttachment(mediaInfo: MediaInfo, progressSubject: CurrentValueSubject<Double, Never>) async -> Result<Void, RoomProxyError> {
+        let requestHandle: ((SendAttachmentJoinHandleProtocol) -> Void) = { [weak self] handle in
+            self?.requestHandle?.cancel()
```
pixlwave, line 85: Why does it call cancel when storing the handle?

pixlwave (approved):

ElementX/Sources/Screens/MediaUploadPreviewScreen/MediaUploadPreviewScreenViewModel.swift
stefanceriu: Just in case there was another one already set ¯\_(ツ)_/¯

PR #1224, opened 2023-06-30
Migration Screen

pixlwave:
This PR adds the migration screen. The following changes are made:

- Store account in `AppSettings` on login and only show the screen for accounts that haven't logged in (migration approximation for now).
- Bump version number and perform an app migration to add all accounts to the AppSettings so existing users don't see the migration screen.
- Add a convenience env var (disabled) to quickly wipe app settings at launch.
- Temporarily un-obsolete `ElementColours.brand` (will deal this this in a follow-up via Compound).
- Add a `WaitingDialog` layout and use this on both MigrationScreen and WaitlistScreen.
- Hide the bottom bar on the home screen when skeletons are shown.

| iPhone | iPad |
| - | - |
| Screenshot 2023-06-30 at 11 43 32 am | Screenshot 2023-06-30 at 11 43 57 am |

| Waitlist iPhone | Waitlist done iPad |
| - | - |
| Screenshot 2023-06-30 at 1 01 09 pm | Screenshot 2023-06-30 at 1 01 49 pm |


ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -194,6 +194,10 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
     // MARK: - Private
     
     private static func setupServiceLocator(navigationRootCoordinator: NavigationRootCoordinator) {
+        if ProcessInfo.processInfo.environment["RESET_APP_SETTINGS"].map(Bool.init) == true {
+            AppSettings.reset()
```
stefanceriu, line 198: I'm not sure this is the best place for this, feels a bit hidden. I would've put it in init next to `wipeUserData`

ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
```diff
@@ -26,9 +26,17 @@ struct HomeScreen: View {
     @ObservedObject var context: HomeScreenViewModel.Context
     
     @State private var scrollViewAdapter = ScrollViewAdapter()
-    @State private var showingBottomToolbar = true
+    @State private var lastScrollDirection: ScrollViewAdapter.ScrollDirection = .up
     @State private var isSearching = false
     
+    var bottomBarVisibility: Visibility {
```
stefanceriu, line 32: Very nice 👌

ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -211,10 +215,18 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
         guard oldVersion != newVersion else { return }
         
         if oldVersion < Version(1, 1, 0) {
+            MXLog.info("Migrating to v1.1.0, signing out the user.")
             // Version 1.1.0 switched the Rust crypto store to SQLite
             // There are no migrations in place so we need to reset everything
             wipeUserData()
         }
+        
+        if oldVersion < Version(1, 1, 7) {
```
stefanceriu, line 224: Won't the 1.1.0 migration render this one useless? 🤔 Not a bit problem even it if does.

stefanceriu (approved):
lgtm 👍

ElementX/Sources/Application/AppCoordinator.swift
pixlwave: I kind of expected wiped accounts, and then to show a migration on the next login. Not particularly necessary but feels totally Ok to me.

ElementX/Sources/Application/AppCoordinator.swift
pixlwave: It needs to happen before `AppSettings()` though I think?

pixlwave: Yeah, and for example if it happened any later than this, then the call to [`startAnalytics`](https://github.com/vector-im/element-x-ios/blob/c5d47378a47b0d6cd8c03964e2ea53dcc7f0669b/ElementX/Sources/Application/AppCoordinator.swift#L70) would be wrong.

stefanceriu: Yeah, you're right. Leave it alone 👍

callumu:
One comment: could you use the smaller spinner?

pixlwave:
My bad, the screenshots were old, the PR has the smaller spinner 😃

amshakal:
👯

PR #1232, opened 2023-07-03
Update SDK to 1.0.94-alpha

alfogrillo:
This PR:
- Updates the sdk to `1.0.94-alpha`
- Refactors the `sendLocation` to pass the new parameters

### DIFF
--- a/ElementX/Sources/Services/Room/RoomProxyProtocol.swift
+++ b/ElementX/Sources/Services/Room/RoomProxyProtocol.swift
@@ -115,7 +115,11 @@ protocol RoomProxyProtocol {
                   progressSubject: CurrentValueSubject<Double, Never>?,
                   requestHandle: (SendAttachmentJoinHandleProtocol) -> Void) async -> Result<Void, RoomProxyError>
 
-    func sendLocation(body: String, geoURI: GeoURI) async -> Result<Void, RoomProxyError>
+    func sendLocation(body: String,
+                      geoURI: GeoURI,
+                      description: String?,
+                      zoomLevel: UInt8?,
+                      assetType: AssetType?) async -> Result<Void, RoomProxyError>
 
     /// Retries sending a failed message given its transaction ID
     func retrySend(transactionID: String) async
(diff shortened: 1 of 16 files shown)


ElementX/Sources/Services/Timeline/TimelineItems/Items/Messages/LocationRoomTimelineItemContent.swift
```diff
@@ -14,7 +14,24 @@
 // limitations under the License.
 //
 
+import MatrixRustSDK
```
pixlwave, line 17: I don't think any of the view data should depend on the SDK. Could we map/proxy `AssetType` to our own type to avoid this?

ElementX/Sources/Services/Room/RoomProxyProtocol.swift
```diff
@@ -115,7 +115,11 @@ protocol RoomProxyProtocol {
                   progressSubject: CurrentValueSubject<Double, Never>?,
                   requestHandle: (SendAttachmentJoinHandleProtocol) -> Void) async -> Result<Void, RoomProxyError>
 
-    func sendLocation(body: String, geoURI: GeoURI) async -> Result<Void, RoomProxyError>
+    func sendLocation(body: String,
+                      geoURI: GeoURI,
+                      description: String?,
+                      zoomLevel: UInt8?,
```
pixlwave, line 121: A little bit of me wonders if we should simply map this to `UInt` but I guess the 8 makes it quite small and probably not a great idea to do so.

ElementX/Sources/Services/Timeline/TimelineItems/Items/Messages/LocationRoomTimelineItemContent.swift
alfogrillo: Applied -> https://github.com/vector-im/element-x-ios/pull/1232/commits/b20c3436aa59dc35d68a18b7d4b5f1daf465876f

ElementX/Sources/Services/Room/RoomProxyProtocol.swift
alfogrillo: Isn't a good idea. Mapping `UInt` to `UInt8` may cause crashes due to overflow. We may clamp values but is not worth doing it since the sdk has already this logic inside (clamp from [0,255] to [0, 20])

PR #1235, opened 2023-07-04
Refactor the NSDiffableDataSourceLayer 

Velin92:
WIP

For now the animations will still be disabled since we need a stable identification source of the timeline items from Rust (WIP). However this is a first step that would allow us to be able to keep a consistent identification of the timeline items, and not reload the cells completely on one of their changes, allowing for smoother animations and state changes in the future.


ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
```diff
@@ -226,13 +238,13 @@ class TimelineTableViewController: UIViewController {
         let previousLayout = layout()
         self.previousLayout = previousLayout
         
-        var snapshot = NSDiffableDataSourceSnapshot<TimelineSection, RoomTimelineViewProvider>()
+        var snapshot = NSDiffableDataSourceSnapshot<TimelineSection, String>()
         snapshot.appendSections([.main])
-        snapshot.appendItems(timelineItems)
+        snapshot.appendItems(timelineItemsDictionary.keys.elements)
```
pixlwave, line 243:
```suggestion
        snapshot.appendItems(timelineItemIDs)
```

pixlwave (approved):
Yes! Looks great to me. Only thing I'm not so keen on is the naming `RoomTimelineItemViewModel` but we can bikeshed that later.

PR #1241, opened 2023-07-05
Add location related analytics

alfogrillo:


flescio (approved):
LGTM!

ElementX/Sources/Services/Analytics/AnalyticsService.swift
```diff
@@ -140,10 +140,22 @@ extension AnalyticsService {
...
     ///   - isReply: whether the composer is used to reply a message
+    ///   - locationType: the type of the shared location
     ///   - startsThread: whether the composer is used to start a new thread
-    func trackComposer(inThread: Bool, isEditing: Bool, isReply: Bool, startsThread: Bool?) {
-        capture(event: AnalyticsEvent.Composer(inThread: inThread, isEditing: isEditing, isReply: isReply, startsThread: startsThread))
+    func trackComposer(inThread: Bool,
+                       isEditing: Bool,
+                       isLocation: Bool,
+                       isReply: Bool,
+                       locationType: AnalyticsEvent.Composer.LocationType? = nil,
```
pixlwave, line 151: Ideally we're trying not to expose the analytics events package outside of `AnalyticsService` (otherwise we would just be capturing the events directly). Is there an app type we can use here that we map to `AnalyticsEvent.Composer.LocationType` inside this method?

ElementX/Sources/Services/Analytics/AnalyticsService.swift
```diff
@@ -140,10 +140,22 @@ extension AnalyticsService {
...
     ///   - isEditing: whether the composer is used to edit a message
+    ///   - isLocation: whether this message it's a shared location
     ///   - isReply: whether the composer is used to reply a message
+    ///   - locationType: the type of the shared location
     ///   - startsThread: whether the composer is used to start a new thread
-    func trackComposer(inThread: Bool, isEditing: Bool, isReply: Bool, startsThread: Bool?) {
-        capture(event: AnalyticsEvent.Composer(inThread: inThread, isEditing: isEditing, isReply: isReply, startsThread: startsThread))
+    func trackComposer(inThread: Bool,
+                       isEditing: Bool,
+                       isLocation: Bool,
```
pixlwave, line 149: Do we need this parameter? Could we infer it from `locationType != nil` instead?

pixlwave (approved):
LGTM, just a couple of comments in the analytics service to address :)

ElementX/Sources/Services/Analytics/AnalyticsService.swift
alfogrillo: Sounds like a good idea, so far we don't send locations without a `locationType`! 👍🏻

ElementX/Sources/Services/Analytics/AnalyticsService.swift
alfogrillo: >  Is there an app type we can use here that we map to AnalyticsEvent.Composer.LocationType inside this method

`sendLocation` uses directly the RustSDK enum.

Added a new type [here](https://github.com/vector-im/element-x-ios/pull/1241/commits/5b80627348a2562a101c315afc668c32416ea568).

wdyt?

pixlwave: Yep yep, looks good to me thanks 🙏

PR #1248, opened 2023-07-05
Notification Placeholder Image Caching

Velin92:
This should fix the memory leak and slowdowns happening when generating a placeholder image.
Also added some logging and decreased the size of the placeholder image to 50x50 so that it's even smaller (100 was actually a bit too much for how small is the size of the icon)

### DIFF
--- a/ElementX/Sources/Other/Extensions/UNNotificationContent.swift
+++ b/ElementX/Sources/Other/Extensions/UNNotificationContent.swift
@@ -105,26 +105,31 @@ extension UNMutableNotificationContent {
     func addSenderIcon(using mediaProvider: MediaProviderProtocol?,
                        senderID: String,
                        senderName: String,
-                       icon: NotificationIcon,
-                       requiresMediaProvider: Bool) async throws -> UNMutableNotificationContent {
-        let shouldSkipPlaceholder = requiresMediaProvider && mediaProvider == nil
+                       icon: NotificationIcon) async throws -> UNMutableNotificationContent {
+        // We display the placeholder only if...
+        var needsPlaceholder = false
 
         var fetchedImage: INImage?
         let image: INImage
         if let mediaSource = icon.mediaSource {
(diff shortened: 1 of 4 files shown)


pixlwave (commented):
What was causing the leak here? E.g. just want to be sure this isn't hiding the issue if say the NSE needed to build lots of notifications for different rooms that haven't been cached, would we still see the behaviour?

Velin92:
> What was causing the leak here? E.g. just want to be sure this isn't hiding the issue if say the NSE needed to build lots of notifications for different rooms that haven't been cached, would we still see the behaviour?

From the logs

> What was causing the leak here? E.g. just want to be sure this isn't hiding the issue if say the NSE needed to build lots of notifications for different rooms that haven't been cached, would we still see the behaviour?

I can't be entirely sure given the current logs but the only major change is the placeholder generation, but I discovered that we were building the placeholder images 2 times per notifications, since we also build the notification content before trying to fetch any media content. 
At least the caching will reduce this to only building the placeholder once and only once per room.
Also added some logs so that if this happens again we can have more insight if is really caused by the placeholder images.

Actually we might even consider NOT building the placeholder image in the first iteration of generating the notification content to make things even more stable, WDYT?

ElementX/Sources/Other/Extensions/UNNotificationContent.swift
```diff
@@ -121,7 +124,8 @@ extension UNMutableNotificationContent {
 
         if let fetchedImage {
             image = fetchedImage
-        } else if let data = await getPlaceholderAvatarImageData(name: icon.groupInfo?.name ?? senderName,
+        } else if !shouldSkipPlaceholder,
```
pixlwave, line 127: Maybe call the property `needsPlaceholder` as its only used here and is slightly easier to follow that way.

pixlwave (approved):
For others following, we discussed and only generating it once definitely made sense. The logs will help identify the leak.

PR #1253, opened 2023-07-05
Show location's description when sharing to Apple Maps

alfogrillo:
**Result**

https://github.com/vector-im/element-x-ios/assets/19324622/da36936c-1e95-4a01-9a24-52f6e34509af


pixlwave (approved):
Nice 👏

PR #1254, opened 2023-07-05
isDM in Notifications

pixlwave:
Fixes DM rendering in notifications, so that a DM must be a direct room with 2 people in.

Also bumps the SDK.

### DIFF
--- a/ElementX/Sources/Services/Notification/Proxy/NotificationItemProxy.swift
+++ b/ElementX/Sources/Services/Notification/Proxy/NotificationItemProxy.swift
@@ -166,8 +166,8 @@ extension NotificationItemProxyProtocol {
     }
 
     var hasMedia: Bool {
-        if (isDirect && senderAvatarMediaSource != nil) ||
-            (!isDirect && roomAvatarMediaSource != nil) {
+        if (isDm && senderAvatarMediaSource != nil) ||
+            (!isDm && roomAvatarMediaSource != nil) {
             return true
         }
         switch event.type {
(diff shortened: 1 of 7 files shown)


ElementX/Sources/Services/Notification/Proxy/NotificationItemProxy.swift
```diff
@@ -37,15 +37,31 @@ protocol NotificationItemProxyProtocol {
...
+            default:
+                return false
+            }
+        default:
+            return false
+        }
+    }
+
+    var isDM: Bool {
+        isRoomDirect && roomJoinedMembers <= 2
```
pixlwave, line 63: Should this also include `isEncrypted`, or (more likely) is there an issue detecting this reliably in notifications?

Velin92: we are not really using it so far so... yeah we can remove it for now (also doesn't really work properly mostly for invite notificatons)

alfogrillo: Not sure it's correct to use the joined members count here.
In other contexts we use the active members count:
```swift
    var isEncryptedOneToOneRoom: Bool {
        isDirect && isEncrypted && activeMembersCount <= 2
    }
```
(I don't think "encrypted" plays a role, maybe we should remove it everywhere)
wdyt?

Velin92 (approved):

ElementX/Sources/Services/Notification/Proxy/NotificationItemProxy.swift
alfogrillo: There are several problems with membership ATM. They seem related with a dirty cache inside the rust sdk.
I tracked this issue here -> https://github.com/matrix-org/matrix-rust-sdk/issues/2078

> Also I think we want to consider a room a DM until a third person joins from a notification point of view.

Ok, then! 👍🏻

alfogrillo (approved):
LGTM

PR #1259, opened 2023-07-05
Welcome Screen

Velin92:
Simulator Screenshot - iPhone 14 - 2023-07-05 at 19 35 37


ElementX/Sources/Screens/AnalyticsPromptScreen/View/RoundedLabelItem.swift
```diff
@@ -16,33 +16,31 @@
...
-            switch self {
-            case .top:
-                return [.topLeft, .topRight]
-            case .middle:
-                return []
-            case .bottom:
-                return [.bottomLeft, .bottomRight]
-            }
+/// Represents the position of a checkmark item in a list.
+enum ListPosition {
```
pixlwave, line 20: This should definitely be nested within the `RoundedLabelItem` (or named more specifically) but can happen in a follow-up PR.

ElementX/Sources/Screens/WelcomeScreenScreen/View/WelcomeScreen.swift
```diff
@@ -0,0 +1,113 @@
...
+        }
+    }
+
+    @ViewBuilder
+    private var title: some View {
+        VStack(spacing: 12) {
+            Text(context.viewState.title)
+                .font(Font.compound.headingLGBold)
+                .foregroundColor(Color.compound.textPrimary)
+            Text(context.viewState.subtitle)
```
pixlwave, line 62: These both need a `.multilineTextAlignment(.center)` on them for dynamic type.

ElementX/Sources/Screens/WelcomeScreenScreen/WelcomeScreenScreenModels.swift
```diff
@@ -0,0 +1,35 @@
...
+    case dismiss
+}
+
+struct WelcomeScreenScreenViewState: BindableState {
+    let title = L10n.screenWelcomeTitle(InfoPlistReader.main.bundleDisplayName)
+    let subtitle = L10n.screenWelcomeSubtitle
+    let bullet1 = L10n.screenWelcomeBullet1
+    let bullet2 = L10n.screenWelcomeBullet2
+    let bullet3 = L10n.screenWelcomeBullet3
+    let buttonTitle = L10n.screenWelcomeButton
```
pixlwave, line 29: Given these are so simple I'd just put them directly in the view. The analytics screen had this weird setup because it used to use HTML parsing before the redesign. (Again fine for a follow-up PR)

ElementX/Sources/Screens/WelcomeScreenScreen/WelcomeScreenScreenViewModel.swift
```diff
@@ -0,0 +1,45 @@
...
+
+class WelcomeScreenScreenViewModel: WelcomeScreenScreenViewModelType, WelcomeScreenScreenViewModelProtocol {
+    let appSettings: AppSettings
+    private var actionsSubject: PassthroughSubject<WelcomeScreenScreenViewModelAction, Never> = .init()
+    
+    var actions: AnyPublisher<WelcomeScreenScreenViewModelAction, Never> {
+        actionsSubject.eraseToAnyPublisher()
+    }
+
+    init(appSettings: AppSettings = ServiceLocator.shared.settings) {
```
pixlwave, line 30: Stefan isn't keen on the default values here - eventually the ServiceLocator singleton is going in the bin. (Again follow up PR)

pixlwave (approved):
One comment to handle now, otherwise looks good to me!

ElementX/Sources/Screens/AnalyticsPromptScreen/View/RoundedLabelItem.swift
Velin92: the problem is that then it will require the generic when specifying the type

ElementX/Sources/Screens/WelcomeScreenScreen/WelcomeScreenScreenModels.swift
Velin92: okay will work on this in the follow up

ElementX/Sources/Screens/WelcomeScreenScreen/WelcomeScreenScreenViewModel.swift
Velin92: okay adding this also to the list of the follow up

ElementX/Sources/Screens/WelcomeScreenScreen/View/WelcomeScreen.swift
Velin92: will add this right now

Velin92: added

PR #1265, opened 2023-07-06
Notification Encryption Sync is defaulted to true

Velin92:
Will only affect those that have never changed value.

### DIFF
--- a/ElementX/Sources/Other/UserPreference.swift
+++ b/ElementX/Sources/Other/UserPreference.swift
@@ -77,10 +77,12 @@ extension UserPreference {
     }
 
     /// Convenience initializer that also immediatelly stores the provided initialValue.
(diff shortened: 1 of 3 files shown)


ElementX/Sources/Other/UserPreference.swift
```diff
@@ -77,10 +77,12 @@ extension UserPreference {
     }
 
     /// Convenience initializer that also immediatelly stores the provided initialValue.
+    /// The initial value is stored every time the app is launched.
+    /// And weill override any existing values.
```
pixlwave, line 81:
```suggestion
    /// And will override any existing values.
```

pixlwave (approved):
Yay! 🔔

That's a good clarification on the doc too 👌

Issue #1180, opened 2023-06-27
Send current user location


flescio:
Refers to https://github.com/vector-im/element-meta/issues/1682

PR #1272, opened 2023-07-06
Send current user location

flescio:
fixes https://github.com/vector-im/element-x-ios/issues/1180

add the share current user location feature and also viewing the current location in the expanded detail map from timeline.
There is a difference in this 2 modes because for the view we ask permission only after the user press the navigator icon; instead for the share screen we ask permission as soon as possible.

sharelocation


ElementX/Sources/Generated/Assets.swift
```diff
@@ -29,6 +29,10 @@ internal enum Asset {
     internal static let backgroundColor = ColorAsset(name: "colors/background-color")
   }
   internal enum Images {
+    internal static let locationMarker = ImageAsset(name: "images/location-marker")
+    internal static let locationPin = ImageAsset(name: "images/location-pin")
+    internal static let locationPointerFull = ImageAsset(name: "images/location-pointer-full")
+    internal static let locationPointer = ImageAsset(name: "images/location-pointer")
```
alfogrillo, line 35: Are they sorted by name?
If so, they look in the wrong place.
Maybe we need to generate them again?

ElementX/Sources/Other/MapLibre/MapLibreMapView.swift
```diff
@@ -86,44 +81,46 @@ struct MapLibreMapView: UIViewRepresentable {
...
     private func showUserLocation(in mapView: MGLMapView) {
-        switch showsUserLocationMode {
-        case .showAndFollow:
-            mapView.showsUserLocation = true
+        switch (showsUserLocationMode, options.annotations) {
+        case (.showAndFollow, _):
             mapView.userTrackingMode = .follow
-        case .show:
+        case (.show, let annotations) where !annotations.isEmpty:
+            guard mapView.locationManager.authorizationStatus != .notDetermined else { return }
```
alfogrillo, line 102: It would be helpful to clarify with a comment what are we doing here

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
```diff
@@ -33,26 +34,23 @@ enum StaticLocationInteractionMode: Hashable {
...
-            bindings = .init()
-        case .viewOnly(let geoURI, _):
-            bindings = .init(mapCenterLocation: .init(latitude: geoURI.latitude, longitude: geoURI.longitude))
+            bindings.showsUserLocationMode = .showAndFollow
+        case .viewOnly:
+            bindings.showsUserLocationMode = .show
         }
     }
-
+
```
alfogrillo, line 46:
```suggestion

```

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
```diff
@@ -66,16 +64,20 @@ struct StaticLocationScreenViewState: BindableState {
             return .init(latitude: geoURI.latitude, longitude: geoURI.longitude)
         }
     }
-
-    var isLocationPickerMode: Bool {
+
```
alfogrillo, line 67:
```suggestion

```

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
```diff
@@ -66,16 +64,20 @@ struct StaticLocationScreenViewState: BindableState {
...
+    
+    var initialMapCenter: CLLocationCoordinate2D {
         switch interactionMode {
         case .picker:
-            return true
-        case .viewOnly:
-            return false
+            return .init(latitude: 49.843, longitude: 9.902056)
+        case .viewOnly(let geoURI, _):
+            return .init(latitude: geoURI.latitude, longitude: geoURI.longitude)
```
alfogrillo, line 73: There is some overlap with `var mapAnnotationCoordinate`. Do we use both?

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
```diff
@@ -66,16 +64,20 @@ struct StaticLocationScreenViewState: BindableState {
...
         }
     }
 
+    var isLocationPickerMode: Bool {
+        interactionMode == .picker
```
alfogrillo, line 78: Duplicate of `var showBottomToolbar: Bool`.
Maybe delete `var showBottomToolbar: Bool`

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
```diff
@@ -66,16 +64,20 @@ struct StaticLocationScreenViewState: BindableState {
...
-
-    var isLocationPickerMode: Bool {
+    
+    var initialMapCenter: CLLocationCoordinate2D {
         switch interactionMode {
         case .picker:
-            return true
-        case .viewOnly:
-            return false
+            return .init(latitude: 49.843, longitude: 9.902056)
```
alfogrillo, line 71: Maybe put a comment to explain where this location is

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
```diff
@@ -115,6 +121,10 @@ struct StaticLocationScreenViewState: BindableState {
 
 struct StaticLocationScreenBindings {
     var mapCenterLocation: CLLocationCoordinate2D?
+    /// Behavior mode of the current user's location, can be hidden, only shown and shown following the user
```
alfogrillo, line 124: The comment doesn't feel super helpful since we already documented `ShowUserLocationMode`

ElementX/Sources/Screens/LocationSharing/StaticLocationScreenViewModel.swift
```diff
@@ -29,7 +29,7 @@ class StaticLocationScreenViewModel: StaticLocationScreenViewModelType, StaticLo
     init(interactionMode: StaticLocationInteractionMode) {
         super.init(initialViewState: .init(interactionMode: interactionMode))
     }
-    
+
```
alfogrillo, line 32:
```suggestion

```

ElementX/Sources/Screens/LocationSharing/View/StaticLocationScreen.swift
```diff
@@ -44,17 +44,21 @@ struct StaticLocationScreen: View {
...
+            .ignoresSafeArea(.all, edges: mapSafeAreaEdges)
+            
             if context.viewState.isLocationPickerMode {
                 LocationMarkerView()
             }
         }
-        .ignoresSafeArea(.all, edges: mapSafeAreaEdges)
+        .overlay(alignment: .bottomTrailing) {
+            centerToUserLocationButton
+        }
```
alfogrillo, line 61: At this point you can delete the ZStack at `44` and add another `.overlay` for the pin. Up to you.

ElementX/Sources/Other/MapLibre/MapLibreMapView.swift
```diff
@@ -20,15 +20,19 @@ import SwiftUI
...
 struct MapLibreMapView: UIViewRepresentable {
     struct Options {
-        /// The initial zoom level
         let zoomLevel: Double
+        /// The initial zoom level
+        let initialZoomLevel: Double
+        
         /// The initial map center
-        let mapCenter: CLLocationCoordinate2D?
+        let mapCenter: CLLocationCoordinate2D
```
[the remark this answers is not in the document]
flescio: That's the starting point, the client has always one to provide, it's more obvious that if the client doesn't have a starting point it will give at least the 0, 0 coordinate

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
```diff
@@ -138,5 +136,33 @@ struct StaticLocationScreenBindings {
 enum StaticLocationScreenViewAction {
     case close
     case selectLocation
+    case centerToUser
     case userDidPan
 }
+
+extension AlertInfo where T == LocationSharingViewError {
```
Velin92, line 143: I would make this fileprivate or private, if possible

ElementX/Sources/Screens/LocationSharing/StaticLocationScreenCoordinator.swift
```diff
@@ -51,6 +51,10 @@ final class StaticLocationScreenCoordinator: CoordinatorProtocol {
             switch action {
             case .close:
                 actionsSubject.send(.close)
+            case .openSystemSettings:
+                guard let url = URL(string: UIApplication.openSettingsURLString),
+                      UIApplication.shared.canOpenURL(url) else { return }
+                UIApplication.shared.open(url)
```
Velin92, line 57:
```suggestion
                guard let url = URL(string: UIApplication.openSettingsURLString),
                      UIApplication.shared.canOpenURL(url) else { 
                      return 
                }
                UIApplication.shared.open(url)
```

ElementX/Sources/Screens/LocationSharing/StaticLocationScreenViewModel.swift
```diff
@@ -38,8 +38,17 @@ class StaticLocationScreenViewModel: StaticLocationScreenViewModelType, StaticLo
...
+            state.bindings.showsUserLocationMode = .show
+        case .centerToUser:
+            switch state.bindings.isLocationAuthorized {
+            case .some(true), .none:
+                state.bindings.showsUserLocationMode = .showAndFollow
+            case .some(false):
+                state.bindings.alertInfo = .init(locationSharingViewError: .missingAuthorization,
+                                                 primaryButton: .init(title: L10n.actionNotNow, role: .cancel, action: nil),
+                                                 secondaryButton: .init(title: L10n.commonSettings,
+                                                                        action: { [weak self] in self?.actionsSubject.send(.openSystemSettings) }))
```
Velin92, line 50:
```suggestion
                let action = { [weak self] in self?.actionsSubject.send(.openSystemSettings) }
                state.bindings.alertInfo = .init(locationSharingViewError: .missingAuthorization,
                                                 primaryButton: .init(title: L10n.actionNotNow, role: .cancel, action: nil),
                                                 secondaryButton: .init(title: L10n.commonSettings,
                                                                        action: action))
```

ElementX/Sources/Screens/LocationSharing/View/StaticLocationScreen.swift
```diff
@@ -81,19 +84,22 @@ struct StaticLocationScreen: View {
...
+                         mapCenter: context.viewState.initialMapCenter)
         }
 
         return .init(zoomLevel: context.viewState.zoomLevel,
-                     mapCenter: coordinate,
-                     annotations: [LocationAnnotation(coordinate: coordinate, anchorPoint: .bottomCenter) {
+                     initialZoomLevel: context.viewState.initialZoomLevel,
+                     mapCenter: context.viewState.initialMapCenter,
+                     annotations: [LocationAnnotation(coordinate: context.viewState.initialMapCenter, anchorPoint: .bottomCenter) {
                          LocationMarkerView()
```
Velin92, line 97: I would extract the closure it's a bit hard to read while inside an element of an array that is also a parameter

ElementX/Sources/Screens/LocationSharing/View/StaticLocationScreen.swift
```diff
@@ -111,6 +117,15 @@ struct StaticLocationScreen: View {
         }
     }
     
+    private var centerToUserLocationButton: some View {
+        Button {
+            context.send(viewAction: .centerToUser)
+        } label: {
+            Image(asset: context.viewState.isSharingUserLocation ? Asset.Images.locationPointerFull : Asset.Images.locationPointer)
```
Velin92, line 124: not sure about this, but shouldn't this be
.resizable()
.scaleToFit()
.frame(width: height:)

where width and height use a @ScaledMetric ? 
@pixlwave what do you think? since this is inside a map view technically...

Velin92 (approved):
LGTM just a few suggestions

ElementX/Sources/Screens/LocationSharing/View/StaticLocationScreen.swift
pixlwave: Sounds reasonable to me. Can use the `relativeTo` parameter on it with a heading font to restrict the amount it grows by if necessary.

Should also toggle the `Preserve Vector Data` option on both of the assets too.

ElementX/Resources/Localizations/en.lproj/Localizable.strings
```diff
@@ -139,7 +139,10 @@
 "emoji_picker_category_places" = "Travel & Places";
 "emoji_picker_category_symbols" = "Symbols";
 "error_failed_creating_the_permalink" = "Failed creating the permalink";
+"error_failed_loading_map" = "Element could not load the map. Please try again later.";
 "error_failed_loading_messages" = "Failed loading messages";
+"error_failed_locating_user" = "Element could not access your location. Please try again later.";
+"error_missing_location_auth" = "Element does not have permission to access your location. You can enable access in Settings > Location";
```
pixlwave, line 145: `Element` shouldn't be hard-coded into these strings so that it updates automatically for forks.

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
alfogrillo: Unfortunately we are using this extension also outside this file :-/

ElementX/Resources/Localizations/en.lproj/Localizable.strings
alfogrillo: Addressed [here](https://github.com/vector-im/element-x-ios/pull/1272/commits/306f6c09f86076eef3cd4bd247ec4f2b6aedd8f8) 👍🏻

ElementX/Sources/Screens/LocationSharing/View/StaticLocationScreen.swift
alfogrillo: I think that for consistency it's better not to scale the pin like we are doing in the timeline after an agreement with the design team.
We change it later if we find that it's not working well.

PR #1278, opened 2023-07-06
Add xcDeferFulfillment to help with tests that have publishers.

langleyd:
## Context
Some of our unit tests that work with async code are a bit flakey, this should solve one of the cases. We had the utility in element-ios but never got moved over. This PR brings that utility over, updates it to use async/await and applies it in a couple of the tests that have been flakey.

## The Problem
In unit test where we are...

1. Creating publishers via ViewModels(or other similar artefacts)
2. Performing actions on ViewModels
3. Asserting results from the publishers

...we need to be careful to subscribe to the publishers before we perform the actions(2). If we perform the actions and then subscribe, there is a race on whether the subscription is setup before the async code completes and the publisher produces a result. If the publisher fires before the subscription is setup, an await on such a subscription will hang indefinitely. 

## The Solution
```Swift
let collectedEvents = somePublisher.collect(3).first()
// Sets up the subscription
let deferred = xcDeferFulfillment(collectedEvents)

// ...Do some other work here like sending ViewModel events that publishes to `somePublisher`

// fulfill and assert results
 XCTAssertEqual(try await deferred.fulfill(), [expected, values, here])
```


UnitTests/Sources/Extensions/XCTestCase.swift
```diff
@@ -0,0 +1,63 @@
...
+    /// let collectedEvents = somePublisher.collect(3).first()
+    /// let awaitDeferred = xcDeferFulfillment(collectedEvents)
+    /// // Do some other work that publishes to somePublisher
+    /// XCTAssertEqual(try await awaitDeferred.execute(), [expected, values, here])
+    ///  ```
+    /// - Parameters:
+    ///   - publisher: The publisher to wait on.
+    ///   - timeout: A timeout after which we give up.
+    /// - Returns: The deferred fulfilment to be executed after some actions and that returns the result of the publisher.
+    func xcDeferFulfillment<T: Publisher>(_ publisher: T, timeout: TimeInterval = 10) -> XCTDeferredFulfillment<T.Output> {
```
pixlwave, line 33: Given this is our own method, can we drop the XC prefixes?
```suggestion
    func deferFulfillment<T: Publisher>(_ publisher: T, timeout: TimeInterval = 10) -> DeferredFulfillment<T.Output> {
```

UnitTests/Sources/Extensions/XCTestCase.swift
```diff
@@ -0,0 +1,63 @@
...
+        var result: Result<T.Output, Error>?
+        let expectation = expectation(description: "Awaiting publisher")
```
pixlwave, line 35: Might be nice to allow an optional message to be passed in for the expectation.

UnitTests/Sources/Extensions/XCTestCase.swift
```diff
@@ -0,0 +1,63 @@
...
+                switch completion {
+                case .failure(let error):
+                    result = .failure(error)
+                case .finished:
+                    break
+                }
+                expectation.fulfill()
+            }, receiveValue: { value in
+                result = .success(value)
+            })
```
pixlwave, line 47: Can we use trailing closures here?

UnitTests/Sources/Extensions/XCTestCase.swift
```diff
@@ -0,0 +1,63 @@
...
+        
+        return XCTDeferredFulfillment<T.Output>(closure: {
+            await self.fulfillment(of: [expectation], timeout: timeout)
+            cancellable.cancel()
+            let unwrappedResult = try XCTUnwrap(result, "Awaited publisher did not produce any output")
+            return try unwrappedResult.get()
+        })
```
pixlwave, line 54: And here?

pixlwave (approved):
Thank you, this is a really nice and clean solution to use 👏

PR #1301, opened 2023-07-10
Performance tests

michaelkaye:
We seek to split out the performance tests into:

 * launch time
 * login time
 * (first) sync time
 * room join time

At present I'm comfortable with the timings for launch, login and first sync, but room join seems to have a 20s delay that I'm not expecting.

This is now running fairly stability on the GHA runner so if/when we merge we should let it run for a week or so to gather information on reliability of the tests.

Before merging, I would like to delete the test that just check application launch time - we can pull that metric from the main test. I tried to do so already but there was a complaint about missing files in the build, so this may be something to do with generated project configurations I'm not sure about.


fastlane/Fastfile
```diff
@@ -143,15 +143,16 @@ end
 
 lane :integration_tests do
   create_simulator_if_necessary(
-    name: "iPhone 13 Pro",
-    type: "com.apple.CoreSimulator.SimDeviceType.iPhone-13-Pro"
+    name: "iPhone 14 Pro",
+    type: "com.apple.CoreSimulator.SimDeviceType.iPhone-14-Pro"
   )
```
pixlwave, line 148: We could remove this `create_simulator_if_necessary` if there was an issue with using it with a 13 Pro, as when Xcode 15 releases we'll be in the same situation.

IntegrationTests/Sources/LoginTests.swift
```diff
@@ -100,13 +109,29 @@ class LoginTests: XCTestCase {
...
+        // we let that happen and wait
+
+        if app.staticTexts[A11yIdentifiers.migrationScreen.message].waitForExistence(timeout: 10.0) {
+            let message = app.staticTexts[A11yIdentifiers.migrationScreen.message]
+            var elapsed = 0
+            while elapsed < 300 {
+                if !message.waitForExistence(timeout: 1.0) { break }
+                sleep(10)
+                elapsed += 10
+            }
```
pixlwave, line 121: What's the rationale for the loop rather than a `waitForExistence(timeout: 300)`? Presumably terminal noise?

michaelkaye: There' no `waitForNonExistence(timeout: 300)` so instead we check every 10s for existence and break if we detect the object has gone.

fastlane/Fastfile
michaelkaye: In this case I was making the fastlane and the xcode project both use the same simulator to remove any differences when I had a working test run from xcode and a non-working fastlane build. I don't believe this was in fact a critical change and would be happy to revert it, but that's why I did this.

pixlwave: Totally fine to leave it, we can have a Dynamic Island in our integration tests! 🏝️
Just double checking :)

IntegrationTests/Sources/LoginTests.swift
pixlwave: Ah, that's because I removed the one example of this in the first commit 🙃
https://github.com/vector-im/element-x-ios/pull/1301/commits/5db9c4db5de2f1e0a3bd469f0c614712fe67184d#diff-ded1e2f70beaa851d0c37f6008ed92deab1b98bd873befac8b3b0cfa97c121a3L75

That would work for doing non-existence.

michaelkaye: Have tested this by running on the branch, it looks like it works, if you're happy with the new code.

Velin92 (approved):

PR #1305, opened 2023-07-10
Removing Delivered Notifications

Velin92:
When a room is marked as read all notifications for that room are removed.

Same for all the invites notifications when the invites screen has appeared.


ElementX/Sources/Other/NotificationName.swift
pixlwave, line 1: Maybe should go inside Sources/Other/Extensions?

pixlwave (approved):
Nice, looking forwards to this one 😎

PR #1306, opened 2023-07-11
Fix flakey tests with deferFulfillment

langleyd:
Builds on https://github.com/vector-im/element-x-ios/pull/1278

Fixes more cases where there was a race between a subscription being setup and an action being sent.


pixlwave (approved):
Thank you! Looking at UI tests right now and was going to come over to these next, but now I don't need to 👏

Issue #1237, opened 2023-07-05
App crashes when trying to view user detail in large rooms


kittykat:
### Steps to reproduce

1. In Matrix HQ timeline, click on user detail

#### What did you expect?
I see user detail view

#### What happened instead?
See spinner for a while, nothing happens, then app crashes.

### Your phone model

iPhone 11

### Operating system version

16.5

### Application version

1.1.8 (272)

### Homeserver

matrix.org

### Will you send logs?

Yes

pixlwave:
This issue is no longer present. The loading indicators were dismissing early meaning you could tap the avatar a second time (or tap a second avatar) which would attempt to push both member details resulting in a crash in the state machine on the second one. The indicator now remains until the list of members is loaded.

PR #1308, opened 2023-07-11
Fix membersPublisher thread

alfogrillo:
After this [PR](https://github.com/vector-im/element-x-ios/pull/1268) there is no guarantee on the thread where the proxy publishes an update on the `membersSubject`.
However ATM clients of the room proxy are using the `membersPublisher` as if it publishes always on the main thread.
This causes runtime UI issues.

This PR brings the `memberPublisher`'s updates back on the main queue.


Velin92:
May this actually fix this?
https://github.com/vector-im/element-x-ios/issues/1237

Velin92 (approved):

alfogrillo:
> May this actually be the fix for...
> 
> https://github.com/vector-im/element-x-ios/issues/1237
> 
> ...?

I don't think it's related.
I Merged https://github.com/vector-im/element-x-ios/pull/1268 this morning. There isn't a TF build with it yet

PR #1310, opened 2023-07-11
Bump Rust SDK version to 1.0.98-alpha

aringenbach:
Timestamps of reaction are now exposed through FFI.
Latest event has moved back to async so I restored the semaphore solution (with a slight change to use a box instead of a static property)


alfogrillo:
@aringenbach the code isn't building both on my machine and on GH.
It looks like types have been deleted from the sdk.

aringenbach:
> @aringenbach the code isn't building both on my machine and on GH. It looks like types have been deleted from the sdk.

Yeah sorry for the auto review assign, I just pre-uploaded that, it is meant to be used together with https://github.com/vector-im/element-x-ios/pull/1312

Velin92 (approved):

PR #1312, opened 2023-07-11
SDK Changes Refactor

Velin92:
This PR includes the two following refactors coming from SDK changes:

 - App Service to control at the same time the encryption sync AND the room list API
 - Internal identifiers to keep a stable identifiction system for each timeline item, this is independent from the event id which can be not available in some cases (for example between the transition from Local echo to Remote Message)
 - Refactor to properly use the right identifier for the right call, and joined the identifiers into a single  Hashable struct
 - SDK Bump to 1.0.98


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -385,85 +372,73 @@ class ClientProxy: ClientProxyProtocol {
...
+            let roomListService = appService.roomListService()
+
+            roomSummaryProvider = RoomSummaryProvider(roomListService: roomListService,
+                                                      eventStringBuilder: RoomEventStringBuilder(stateEventStringBuilder: RoomStateEventStringBuilder(userID: userID)),
+                                                      name: "AllRooms")
+
+            try await roomSummaryProvider?.setRoomList(roomListService.allRooms())
+
+            inviteSummaryProvider = RoomSummaryProvider(roomListService: roomListService,
+                                                        eventStringBuilder: RoomEventStringBuilder(stateEventStringBuilder: RoomStateEventStringBuilder(userID: userID)),
```
pixlwave, line 396: I realise this is how it was, but should we make this string builder once and pass it to both providers?

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -140,7 +140,23 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
         
         MXLog.verbose("\(name): Finished applying \(diffs.count) diffs, new room list \(rooms.compactMap { $0.id ?? "Empty" })")
     }
-        
+
+    private func fetchLastMessage(roomListItem: RoomListItemProtocol) -> EventTimelineItem? {
+        class RoomMessageBox {
```
pixlwave, line 145: Box feels very Rust and a bit alien to Swift imo…

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -385,85 +372,73 @@ class ClientProxy: ClientProxyProtocol {
...
+
+            self.appService = appService
+            self.roomListService = roomListService
+
+            appServiceUpdateTaskHandle = appService.state(listener: AppStateObserverProxy { [weak self] state in
                 guard let self else { return }
+                MXLog.info("Received app service update: \(state)")
+                switch state {
+                case .error:
+                    restartSync()
```
pixlwave, line 407: Ohhhh interesting, I'd wondered how the app was retrying requests in a quick loop. In the future we should be adding some kind of back-off on this.

pixlwave (approved):
Very nice, this is a really positive change to the code :)

ElementX.xcodeproj/xcshareddata/xcschemes/ElementX.xcscheme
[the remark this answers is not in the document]
Velin92: actually this was in @aringenbach 's PR and xcodegen does not revert it back, is it actually needed?
(I actually realised that when I compiled the code the proj was automatically doing this update too to the scheme, so I guess is probably related to Xcode updtes?)

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
Velin92: this was originally in @aringenbach 's PR and I think is related to the async method solution.
Maybe we want to call it `EventTimelineItemReference` ?

ElementX.xcodeproj/xcshareddata/xcschemes/ElementX.xcscheme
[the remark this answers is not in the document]
pixlwave: Oh ok well if xcodegen doesn't revert it, then its fine.

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
pixlwave: Given its nested could simply be `FetchResult` or similar.

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -449,17 +449,20 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
             setComposerMode(.edit(originalItemId: messageTimelineItem.id))
         case .copyPermalink:
             do {
-                let permalink = try PermalinkBuilder.permalinkTo(eventIdentifier: eventTimelineItem.id, roomIdentifier: timelineController.roomID,
+                guard let eventID = eventTimelineItem.id.eventID else {
+                    break
```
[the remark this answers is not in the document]
Velin92: or even better I can just do:
```
 displayError(.alert(L10n.errorFailedCreatingThePermalink))
 break
 ```

File: ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift [hunk 5]
```diff
@@ -121,68 +121,81 @@ class RoomTimelineController: RoomTimelineControllerProtocol {
...
+    func sendMessage(_ message: String, inReplyTo itemID: TimelineItemIdentifier?) async {
         if itemID == nil {
             MXLog.info("Send message in \(roomID)")
         } else {
             MXLog.info("Send reply in \(roomID)")
         }
-        
-        switch await roomProxy.sendMessage(message, inReplyTo: itemID) {
+
+        switch await roomProxy.sendMessage(message, inReplyTo: itemID?.eventID) {
```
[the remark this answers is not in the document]
Velin92: done

PR #1314, opened 2023-07-12
Reactions: Add more button

langleyd:
# What's in this PR
- Adds the "Add more" button to the reaction layout if there are greater than 0 already, for quickly adding more reactions.
- Create `ReactionLayoutItemType` `LayoutValueKey` to tag subviews with `ReactionLayoutItem`
- Add `LayoutValueKey` support to the layout mocks
- Make the animation of the addMore and collapse buttons consistent

# What does it look like
Simulator Screen Recording - iPhone 14 Pro - 2023-07-12 at 09 51 00


pixlwave (commented):
I still need to look at the changes in the flow layout, but finishing for today so posting this for now :)

ElementX/Resources/Assets.xcassets/images/timeline/timeline-reaction-add-more.imageset/Contents.json
```diff
@@ -0,0 +1,12 @@
+{
+  "images" : [
+    {
+      "filename" : "timeline-reaction-add-more.pdf",
+      "idiom" : "universal"
```
pixlwave, line 5: I think you want to check the `Preserves Vector Data` checkbox in the inspector to avoid this being rendered to a bitmap representation.

ElementX/Sources/Other/CollapsibleFlowLayout/CollapsibleReactionLayout.swift
pixlwave, line 1: Looks like theres a few instances of `row`/`horizontalSpacing`/`verticalSpacing` → `line`/`hSpacing`/`vSpacing` in this file, otherwise LGTM.

An unimportant thing I'm wondering (without really knowing about how it works) is that we calculate the view sizes quite a bit. Do you think the `cache` parameter could help us with that?

pixlwave (approved):
Sorry for the delay coming back to this one. LGTM 👍

Issue #1289, opened 2023-07-08
Ability to know when people sent a given emoji reaction


ara4n:
* If you long-tap on a emoji reaction, you get told who sent which emoji.  (This is very undiscoverable, but that's a separate issue)
* Weirdly, it doesn't tell you when the emoji got sent though.  If you care enough about who specifically reacted to a message, though, you'd expect to also be told *when* they did so - just as EW/ED does, given the app obviously knows that info, and it's useful to know if you're trying to track when someone handshook on a message (thinking in terms of the all-to-often-forgotten 'people coordinator' persona)

pixlwave:
Adding a note that this is already in progress.

PR #1320, opened 2023-07-12
Sort reactions on the timeline by count and then timestamp and show the date on the summary view

langleyd:
# What's in this PR
- Sort reaction aggregations on the timeline by count and then by most recent timestamp ascending(tagging new reactions on the end)
- Show the timestamp on the summary view and sort descending(newest at the top)

# What does it look like


ElementX/Sources/Services/Timeline/TimelineItemContent/AggregratedReaction.swift
```diff
@@ -23,7 +23,15 @@ struct AggregatedReaction: Hashable {
...
     let key: String
     /// The user ids of those who sent the reactions
-    let senders: [String]
+    let senders: [ReactionSender]
+}
+
+/// Details of who sent the reaction
+struct ReactionSender: Hashable {
+    /// The id of the user who sent the reaction
+    let senderId: String
```
pixlwave, line 32: We generally use `ID` on the Swift side :)
```suggestion
    let senderID: String
```

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemFactory.swift
```diff
@@ -313,13 +313,22 @@ struct RoomTimelineItemFactory: RoomTimelineItemFactoryProtocol {
...
+            return AggregatedReaction(accountOwnerID: userID, key: reaction.key, senders: senders)
         }
         .sorted { a, b in
-            // Sort by count and then by key for a consistence experience.
-            // Otherwise emojis can switch around. We can replace
-            // with timestamp as a secondary sort when it is available.
-            (a.count, a.key) > (b.count, b.key)
+            // Sort aggregated reactions by count and then by most recent reaction
+            if a.count == b.count {
+                return a.senders[0].timestamp < b.senders[0].timestamp
```
pixlwave, line 329: Should this be the other way around for most recent?
```suggestion
                return a.senders[0].timestamp > b.senders[0].timestamp
```

pixlwave (approved):
LGTM 😎

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemFactory.swift
langleyd: The comment should be clearer. It's sorted by count and then by aggregation with most recent reaction(index zero in the aggregation's list of reactions, which are in descending order from line 321) in ascending order(so new reactions are added to the end of the reaction layout).

I'll write that in a clearer way :)

pixlwave: Right, understood 👍

aringenbach:
Fixes #1289

PR #1322, opened 2023-07-13
UI Tests fixes.

pixlwave:
This PR makes various changes to UI tests:
- Fix stale snapshots.
- Fix incorrect placeholder coordinator.
- Move runs to the performance testing runner.
- Sync runs to every 12 hours with a 6 hour offset compared to integration tests.
- Reset the simulator before each run (likely unnecessary but *does* fix the keyboard being hidden)
- Zip the test results bundle to speed up uploads as it contains *lots* of little files.

Awaiting result from run [here](https://github.com/vector-im/element-x-ios/actions/runs/5542457472).

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -2,7 +2,7 @@ name: Integration tests
 
 on:
   schedule:
-    - cron:  '0 * * * *'
+    - cron:  '0 6,18 * * *'
     
   workflow_dispatch:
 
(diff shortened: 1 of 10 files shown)


michaelkaye (commented):
This looks fine from the perf timing interactions - we could even run the UI tests 7 days a week, rather than just mon-fri (1-5) - the machine will be idle otherwise.

alfogrillo (approved):
LGTM

PR #1323, opened 2023-07-13
Update SDK

pixlwave:
This PR updates the SDK and handles the async start method.


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -110,19 +110,26 @@ class ClientProxy: ClientProxyProtocol {
     }
 
     var isSyncing: Bool {
-        roomListService?.isSyncing() ?? false
+        isStartingSync || roomListService?.isSyncing() ?? false
```
Velin92, line 113: maybe at this point since the appService is composed of more than just the roomListService I think instad of relying on roomListService?.isSyncing, we should instead create a new variable, and set it to true or false according to the observation state of the app service.
WDYT?

Velin92 (approved):

ElementX/Sources/Services/Client/ClientProxy.swift
pixlwave: Sounds sensible to me, will look at this separately to get this merged ready for the Nightly build.

PR #1324, opened 2023-07-13
Redacting is now displayed for moderators in groups

Velin92:
If the room is not direct we can now check if the current user is actually able to redact a message or not.

This requires an SDK update.


ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -104,7 +106,14 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
         case .markRoomAsRead:
             Task { await markRoomAsRead() }
         case .timelineItemMenu(let itemID):
-            showTimelineItemActionMenu(for: itemID)
+            Task {
+                if case let .success(value) = await roomProxy.canUserRedact(userID: roomProxy.currentUserID) {
+                    canCurrentUserRedact = value
+                } else {
+                    canCurrentUserRedact = false
+                }
```
pixlwave, line 114: Would it make more sense to make this call just a single time early in the view model's the lifecycle, and then use the value without needing an await?

pixlwave (approved):

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
Velin92: what if the user power level has changed? We would loose the change.
the call is quite fast the only reason I am saving the value is because otherwise I need to use a semaphore in the @MainActor closure.

PR #1327, opened 2023-07-14
Updating members after an ignore/unignore action has been performed.

Velin92:
fixes https://github.com/orgs/vector-im/projects/43/views/60?pane=issue&itemId=28399782

### DIFF
--- a/ElementX/Sources/Screens/RoomMemberDetailsScreen/RoomMemberDetailsScreenViewModel.swift
+++ b/ElementX/Sources/Screens/RoomMemberDetailsScreen/RoomMemberDetailsScreenViewModel.swift
@@ -57,9 +57,7 @@ class RoomMemberDetailsScreenViewModel: RoomMemberDetailsScreenViewModelType, Ro
         switch result {
         case .success:
             state.details.isIgnored = true
-            Task.detached {
-                await self.roomProxy.updateMembers()
-            }
+            updateMembers()
         case .failure:
             state.bindings.alertInfo = .init(id: .unknown)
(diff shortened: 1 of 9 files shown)


ElementX/Sources/Screens/RoomMemberDetailsScreen/RoomMemberDetailsScreenViewModel.swift
```diff
@@ -55,6 +57,9 @@ class RoomMemberDetailsScreenViewModel: RoomMemberDetailsScreenViewModelType, Ro
         switch result {
         case .success:
             state.details.isIgnored = true
+            Task.detached {
+                await self.roomProxy.updateMembers()
+            }
```
alfogrillo, line 62: Why do you need to call `await self.roomProxy.updateMembers()` inside a detached Task?
Since you are `await`ing, blocking the main thread shouldn't be a concern right?

Velin92: I just didn't want to block the main thread for an API that is quite fast and that does not impact the current view model's UI.
I imagine this Task can run concurrently on another in a bg thread since is essentially an update of the values from the DB without slowing down the UI/main thread.
Is not really that important since as I said is quite fast, it felt more optimal to me.

But I am okay to remove it if you think is not necessary.

Ofc

alfogrillo (approved):
LGTM

pixlwave (approved):

PR #1329, opened 2023-07-14
Fix for delayed loader not working properly.

### LINKED ISSUE #1328: Delayed loading displays even if it should not
### Steps to reproduce

1. Open a room that has already been opened, or a room that you are opening for the first time with  few members.
2. immediately tap on the avatar of a user

#### What did you expect?
Since the members should already been cached in the store, the get member function is very fast, and the loading since is delayed should not appear at all.
#### What happened instead?
The loading the displays, even if the call was actually very fast, so fast that in fact that the the other view has already been pushed in the meantime, however then the loading displays.

Here is a video:
https://github.com/vector-im/element-x-ios/assets/34335419/94da4a61-8f4b-4151-a8ff-92eb6a89fa92

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

_No response_

### Homeserver

_No response_

### Will you send logs?

No

Velin92:
fixes #1328

### DIFF
--- a/ElementX.xcodeproj/project.pbxproj
+++ b/ElementX.xcodeproj/project.pbxproj
@@ -474,6 +474,7 @@
 		A680F54935A6ADEA4ED6C38F /* TimelineItemStatusView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6A4C9547BBFEEF30AA11329B /* TimelineItemStatusView.swift */; };
 		A6D4C5EEA85A6A0ABA1559D6 /* RoomDetailsEditScreenModels.swift in Sources */ = {isa = PBXBuildFile; fileRef = 16D09C79746BDCD9173EB3A7 /* RoomDetailsEditScreenModels.swift */; };
 		A6DEC1ADEC8FEEC206A0FA37 /* AttributedStringBuilderProtocol.swift in Sources */ = {isa = PBXBuildFile; fileRef = 72F37B5DA798C9AE436F2C2C /* AttributedStringBuilderProtocol.swift */; };
+		A71CAC782A617BE6003A2B54 /* Duration.swift in Sources */ = {isa = PBXBuildFile; fileRef = A71CAC772A617BE6003A2B54 /* Duration.swift */; };
 		A74438ED16F8683A4B793E6A /* AnalyticsSettingsScreenViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0BCE3FAF40932AC7C7639AC4 /* AnalyticsSettingsScreenViewModel.swift */; };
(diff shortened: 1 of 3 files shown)


ElementX.xcodeproj/project.pbxproj
```diff
@@ -3615,14 +3618,6 @@
...
 			sourceTree = "<group>";
 		};
-		"TEMP_FFE5FDBA-B4DD-4FF7-B172-18026F248E20" /* element-x-ios */ = {
-			isa = PBXGroup;
-			children = (
-				41553551C55AD59885840F0E /* secrets.xcconfig */,
-			);
-			path = "element-x-ios";
-			sourceTree = "<group>";
-		};
```
pixlwave, line 3625: @alfogrillo Does this mean the git hook to ignore the file is ignoring the checkout too?

ElementX/Sources/Other/Extensions/Duration.swift
```diff
@@ -0,0 +1,26 @@
...
+//
+
+import Foundation
+
+extension Duration {
+    /// Use this to to convert `Duration` in seconds as a `Double`.
+    /// `components.seconds` is not reliable because its type is `Int64`.
+    var seconds: Double {
+        let v = components
+        return Double(v.seconds) + Double(v.attoseconds) * 1e-18
```
pixlwave, line 24: How come not this?
```suggestion
        Double(components.seconds) + Double(components.attoseconds) * 1e-18
```

ElementX/Sources/Other/UserIndicator/UserIndicatorController.swift
```diff
@@ -58,7 +58,7 @@ class UserIndicatorController: ObservableObject, UserIndicatorControllerProtocol
         } else {
             if let delay {
                 delayedIndicators.insert(indicator.id)
-                Timer.scheduledTimer(withTimeInterval: Double(delay.components.seconds), repeats: false) { [weak self] _ in
+                Timer.scheduledTimer(withTimeInterval: Double(delay.seconds), repeats: false) { [weak self] _ in
```
pixlwave, line 61:
```suggestion
                Timer.scheduledTimer(withTimeInterval: delay.seconds, repeats: false) { [weak self] _ in
```

pixlwave (approved):
LGTM (but should double check about that removal of the secrets config).

alfogrillo (approved):
LGTM

ElementX.xcodeproj/project.pbxproj
alfogrillo: Mmmh this looks wrong.
@Velin92 can you check if targets have been configured like this?

Screenshot 2023-07-14 alle 16 09 31

If not, please generate the project again.

Velin92: now they should be, just pushed again

PR #1340, opened 2023-07-17
Sync fix

stefanceriu:
The app doesn't currently reliably restart the sync when coming back from the background. Tracked this down to `isSyncing` not reporting the correct value because of using the `roomListService` instead of the sync service `state`.

### DIFF
--- a/ElementX/Sources/Services/Client/ClientProxy.swift
+++ b/ElementX/Sources/Services/Client/ClientProxy.swift
@@ -32,7 +32,10 @@ class ClientProxy: ClientProxyProtocol {
     private var roomListStateUpdateTaskHandle: TaskHandle?
 
     private var appService: App?
-    private var appServiceUpdateTaskHandle: TaskHandle?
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -32,7 +32,10 @@ class ClientProxy: ClientProxyProtocol {
     private var roomListStateUpdateTaskHandle: TaskHandle?
 
     private var appService: App?
-    private var appServiceUpdateTaskHandle: TaskHandle?
+    private var appServiceStateUpdateTaskHandle: TaskHandle?
+    
+    #warning("This is a workaround until the RustSDK provides a sync service state getter")
+    private var appServiceStateUpdateCurrentValueSubject = CurrentValueSubject<AppState, Never>(.terminated)
```
pixlwave, line 38: Why not just a stored property here, given we only check its value?

pixlwave (approved):
Ah, smart!

PR #1343, opened 2023-07-17
Diffable fade animations

Velin92:
It includes some workarounds to allow for the animation to work properly even considering the issues we currently have with the SS proxy.

However we are now able to animate the appearance of new messages in the timeline.
It works pretty well but we might still need the fix for the timestamp id changing over time (which causes them to flicker):

https://github.com/vector-im/element-x-ios/assets/34335419/dd29818d-f6d0-42f7-972e-26621bd10f2e

For now this can however be enabled with a feature flag for some preliminary testing in Nightly

FIXED CLIENT SIDE THE DAY DIVIDER ISSUE:

https://github.com/vector-im/element-x-ios/assets/34335419/b1b3acf3-2add-4c0a-92bf-c2a2fa1d9f64

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
+++ b/ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
@@ -170,11 +170,11 @@ class TimelineTableViewController: UIViewController {
         guard !hasAppearedOnce else { return }
         scrollToBottom(animated: false)
         hasAppearedOnce = true
-            // This is to prevent the SS proxy issue that forces a full refresh of all the timeline items
-            // the first time the timeline is opened
-            DispatchQueue.main.asyncAfter(deadline: .now() + 2) {
-                self.shouldAnimate = ServiceLocator.shared.settings.timelineDiffableAnimationsEnabled
-            }
+        // This is to prevent the SS proxy issue that forces a full refresh of all the timeline items
+        // the first time the timeline is opened
+        DispatchQueue.main.asyncAfter(deadline: .now() + 2) {
+            self.shouldAnimate = ServiceLocator.shared.settings.timelineDiffableAnimationsEnabled
+        }
     }
     
     override func didMove(toParent parent: UIViewController?) {


ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
```diff
@@ -97,6 +97,8 @@ class TimelineTableViewController: UIViewController {
     private var previousLayout: LayoutDescriptor?
     /// Whether or not the view has been shown on screen yet.
     private var hasAppearedOnce = false
+    /// Wether the scroll and the animations should happen
```
alfogrillo, line 100:
```suggestion
    /// Whether the scroll and the animations should happen
```

ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
```diff
@@ -168,6 +170,13 @@ class TimelineTableViewController: UIViewController {
         guard !hasAppearedOnce else { return }
         scrollToBottom(animated: false)
         hasAppearedOnce = true
+        if ServiceLocator.shared.settings.timelineDiffableAnimationsEnabled {
+            // This is to prevent the SS proxy issue that forces a full refresh of all the timeline items
+            // the first time the timeline is opened
+            DispatchQueue.main.asyncAfter(deadline: .now() + 2) {
+                self.shouldAnimate = true
+            }
+        }
```
alfogrillo, line 179:
```suggestion
            // This is to prevent the SS proxy issue that forces a full refresh of all the timeline items
            // the first time the timeline is opened
            DispatchQueue.main.asyncAfter(deadline: .now() + 2) {
                self.shouldAnimate = ServiceLocator.shared.settings.timelineDiffableAnimationsEnabled
            }
```

Can explain me better why the need to dispatch after 2 seconds?

ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
```diff
@@ -231,8 +240,7 @@ class TimelineTableViewController: UIViewController {
             return cell
         }
 
-//        dataSource?.defaultRowAnimation = .automatic
-        
+        dataSource?.defaultRowAnimation = .fade
```
alfogrillo, line 243: Curiosity: did you try `.bottom`?

ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
Velin92: yeah there is a proxy issue... essentially sometimes (not always) when you enter in the timeline of a room, we confirm that the existing messages with the server by receiving them again. However since there is a proxy issue with the txn_id what happens is that the sdk does not recognize such items, and just completely flushes the existing timeline and regenerates all of them, with a new identifier (so they are essentially considered new messages) which of course removes existing messages and replaces with new ones that exactly look the same creating a blinking animation

ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
Velin92: yeah but it looks bad when the sliding sync proxy doesn't work properly when sending a lot of messages very fast, while .fade makes that issue difficult to notice UX wise.

ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
Velin92: Good news btw Jonas has found an SDK fix for this issue, so I am probably removing this delay in the next SDK update.

ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
Velin92: Actually  Jonas has found a fix for this too maybe! So with the next sdk update we will show different animations and have design pick the best

PR #1351, opened 2023-07-18
Various content pasting issues

stefanceriu:
* fix copy pasting from notes trying to upload a text file instead of pating into the composer
* fix copy pasting image from safari pasting url into the composer instead of only triggering the upload dialog


pixlwave (approved):
👍

PR #1354, opened 2023-07-18
Update "user's location" dot color

alfogrillo:
This PR changes the user's location dot color, so that **doesn't** change in dark mode:

| Before | After |
| --- | --- |
| before|after|


callumu:
This looks good to me 👍

PR #1357, opened 2023-07-18
Fix visibility of navigation bar when searching the forwarding screen.

pixlwave:
- Update Introspect and replace deprecated calls with new ones.
- Make a re-usable method to configure the search controller.
- Use this method in the invite users and message forwarding screens.
- Remove obsoleted colours from DesignKit.

Note: Discussed with design about the double Cancel buttons on the forwarding screen and concluded this is good enough for now.


stefanceriu (approved):
lgtm 👍

PR #1358, opened 2023-07-19
Delete Location Sharing feature flags

alfogrillo:
This PR deletes the location sharing related feature flags.


alfogrillo:
> Nice 🚀🚀
> 
> Should this have a changelog to tell users that location sharing is now available?

Yep, better to add one.
Also trying to understand if we want to change the copy in the welcome screen. 🤔

PR #1361, opened 2023-07-19
Better identification system for virtual items

Velin92:
The issue with recreated timeline date dividers was fixed, however we handled them wrongly by using a chunk index value.

Now we can reliably use the timelineID provided by the SDK.


ElementX/Sources/Screens/RoomScreen/View/Timeline/ReadMarkerRoomTimelineView.swift
```diff
@@ -38,7 +38,8 @@ struct ReadMarkerRoomTimelineView: View {
 struct ReadMarkerRoomTimelineView_Previews: PreviewProvider {
     static let viewModel = RoomScreenViewModel.mock
 
-    static let item = ReadMarkerRoomTimelineItem()
+    static let item = ReadMarkerRoomTimelineItem(id: .init(timelineID: .init(UUID().uuidString)))
```
alfogrillo, line 41: To avoid this verbosity maybe you can add an extension like this:
```swift
extension TimelineItemIdentifier {
    static var random: TimelineItemIdentifier {
        .init(timelineID: UUID().uuidString)
    }
}
```
and just do:
```swift
static let item = ReadMarkerRoomTimelineItem(id: .random)
```

Velin92: lovely! thanks! will do that immediately

PR #1364, opened 2023-07-19
Refactor map styles for forking

alfogrillo:
This PR makes more explicit how to replace a map style (eg. when forking). 
Furthermore it adds UTs for map tiler url builders.

### DIFF
--- a/ElementX/Sources/Other/MapLibre/MapTilerStyle.swift
+++ b/ElementX/Sources/Other/MapLibre/MapTilerStyle.swift
@@ -16,6 +16,7 @@
 
 /// The style for a map.
 /// Values should be Map Libre style IDs generated with an account where the API key belongs to.
+/// For more information read [FORKING.md](https://github.com/vector-im/element-x-ios/blob/develop/docs/FORKING.md#setup-the-location-sharing).
 enum MapTilerStyle: String {
     case light = "9bc819c8-e627-474a-a348-ec144fe3d810"
     case dark = "dea61faf-292b-4774-9660-58fcef89a7f3"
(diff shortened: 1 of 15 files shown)


ElementX/Sources/Other/MapLibre/MapTilerStyle.swift
```diff
@@ -0,0 +1,22 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+/// The style for a map.
+/// Values should be Map Libre style IDs generated with an account where the API key belongs to.
+enum MapTilerStyle: String {
+    case light = "9bc819c8-e627-474a-a348-ec144fe3d810"
+    case dark = "dea61faf-292b-4774-9660-58fcef89a7f3"
+}
```
Velin92, line 22: So I imagine these could be changed according to the user's API settings?

Velin92: Maybe I would just make the doc a bit more cleare by writing something like:

`// FOR FORKS ONY: replace with your color keys
`
or something like that

Velin92 (approved):
LGTM

ElementX/Sources/Other/MapLibre/MapTilerStyle.swift
alfogrillo: This are map styles identifiers.

They aren't supposed to be for forks only, we may decide to change them later.

People can:
- Use public styles like `basic-v2` (reported in the doc like an example)
- Using custom styles (like us). If this is the case, the api key and the custom style should belong to the same account.

Does this clarify?

PR #1366, opened 2023-07-20
Flipping the tableView

Velin92:
I know you have read this title and you are wondering what is this nonsense?

But by observing how Telegram and Whatsapp work, and by checking online:
https://stackoverflow.com/questions/40157645/how-to-add-new-rows-at-bottom-of-tableview-chat-messages
https://www.reddit.com/r/iOSProgramming/comments/vexs68/how_to_create_reverse_tableview/
This is probably an industry standard trick for chat apps.

Usually table views with a pagination system always have its pagination happen at the bottom, however a chat always paginates backwards... which creates a lot of extra complexities to handle... so what about flipping the tableView and its cells?

This does not only work very well, but literally reduced half of the code needed and the overall complexity of handling scrolling, padding, pagination, restoring scroll positions and so on... is all gone, the table view is very light now.

It probably requires some more testing, but the result is amazing, and by checking WhatsApp behaviour this is likely how they made it work.

https://github.com/vector-im/element-x-ios/assets/34335419/20e11191-d7fc-407f-bf3b-422673570ab9

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
+++ b/ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
@@ -188,7 +188,7 @@ class TimelineTableViewController: UIViewController {
             .minSize(height: 1)
             .background(Color.clear)
 
-            cell.transform = CGAffineTransform(scaleX: 1, y: -1)
+            cell.contentView.transform = CGAffineTransform(scaleX: 1, y: -1)
             return cell
         }
 
--- a/ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
+++ b/ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
@@ -35,6 +35,7 @@ class TimelineItemCell: UITableViewCell {
 ///
 /// This class subclasses `UIViewController` as `UITableViewController` adds some
 /// extra keyboard handling magic that wasn't playing well with SwiftUI (as of iOS 16.1).
+/// Also this TableViewController uses a **flipped tableview**
 class TimelineTableViewController: UIViewController {
     private let coordinator: TimelineView.Coordinator
     private let tableView = UITableView(frame: .zero, style: .plain)
@@ -84,8 +85,6 @@ class TimelineTableViewController: UIViewController {
     /// Our view actions get wrapped in a `Task` so it is possible that a second call in
     /// quick succession can execute before ``isBackPaginating`` becomes `true`.
     private let paginateBackwardsPublisher = PassthroughSubject<Void, Never>()
-    /// Whether or not the ``timelineItems`` value should be applied when scrolling stops.
-//    private var hasPendingUpdates = false
     /// Whether or not the view has been shown on screen yet.
     private var hasAppearedOnce = false
     /// Whether the scroll and the animations should happen
@@ -142,6 +141,7 @@ class TimelineTableViewController: UIViewController {
         guard !hasAppearedOnce else { return }
         scrollToBottom(animated: false)
         hasAppearedOnce = true
+        paginateBackwardsPublisher.send()
     }
     
     override func viewWillLayoutSubviews() {
@@ -188,6 +188,7 @@ class TimelineTableViewController: UIViewController {
             .minSize(height: 1)
             .background(Color.clear)
 
+            // Flipping the cell can create some issues with cell resizing, so flip the content View
             cell.contentView.transform = CGAffineTransform(scaleX: 1, y: -1)
             return cell
         }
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
```diff
@@ -246,101 +201,25 @@ class TimelineTableViewController: UIViewController {
...
         MXLog.verbose("DIFF: \(snapshot.itemIdentifiers.difference(from: currentSnapshot.itemIdentifiers))")
 
-        // We only animate if the last item has changed
-        // We don't care to animate backpagination since we want to keep the scrolling position when that happens
-        let animated = shouldAnimate && snapshot.itemIdentifiers.last != currentSnapshot.itemIdentifiers.last
+        // We need to decide if we want to use animations only for the appearence at the bottom or in general
+        // however I noticed that while paginating since some remote messages may lag a bit behind in the first
+        // collected update, this may cause some moving around while scrolling very fast towards the top which is more
+        // noticeable by the animation. However it looks way cooler.
+        let animated = shouldAnimate && snapshot.itemIdentifiers.first != currentSnapshot.itemIdentifiers.first
```
Velin92, line 216: @pixlwave the video in the PR description uses shouldAnimate directly.
It's way cooler but in some very big rooms where some messages that are in the "middle" of a paginations are inserted later by the SS it may result in some moving around. 

This removes the animation when backpaginating but also hides this issue.

Velin92: Maybe we could solve this by actually animating if either lasr != previousLast || first != previousFirst ?

stefanceriu:
We've known about this solution for quite some time, manu having tried it years ago, before element x was even on the board. The original problem with it was that the context menus would be inverted as well (I have a vague feeling that there we others but I can't remember what)

Now, seeing as we don't use context menus anymore and how complex the current solution is, I honestly don't have anything against it but we need to make sure there's no other funky regressions anywhere.

Otherwise, and possibly more importantly, with this in hand do we even need to drop down to UIKit anymore or can we just go full SwiftUI?

stefanceriu:
Case in point: tapping on the status bar scrolls timeline to the bottom 🤪

Velin92:
> Case in point: tapping on the status bar scrolls timeline to the bottom 🤪
> 
> We should try to make a list of regressions and see if we can work around them so we can keep the inverted scroll.

i kinda fixed this, by using the delegate function, however the problem is that the gesture that triggers that delegate function only works when you are not already at the top of the table view, but there is likely a workaround for that too.

ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
```diff
@@ -367,11 +247,11 @@ extension TimelineTableViewController: UITableViewDelegate {
         DispatchQueue.main.async { [weak self] in
             guard let self else { return }
             
-            let isAtBottom = self.isAtBottom()
+            let scrollToBottomButtonVisible = tableView.contentOffset.y > 15
```
pixlwave, line 250: I'm not strongly opinionated about this but worth pointing out:

Its possible we don't need the `15` anymore as scrolling to the top of the table view won't have the same issues of sometimes missing the last few pixels like it did when scrolling to the bottom.

ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
```diff
@@ -399,6 +279,11 @@ extension TimelineTableViewController: UITableViewDelegate {
     func scrollViewDidScrollToTop(_ scrollView: UIScrollView) {
         scrollAdapter.scrollViewDidScrollToTop(scrollView)
     }
+
+    func scrollViewShouldScrollToTop(_ scrollView: UIScrollView) -> Bool {
```
pixlwave, line 283: Maybe lets add a comment here explaining that the system gesture scrolls to the bottom because of flipping and we're fixing this to scroll to the start by scrolling to the bottom.

pixlwave (approved):
LGTM ↕️

ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
Velin92: This means that as soon there will be a minimal scroll that button will appear, if that's okay yeah no problem.

ElementX/Sources/Screens/RoomScreen/View/TimelineTableViewController.swift
```diff
@@ -349,12 +226,15 @@ class TimelineTableViewController: UIViewController {
...
-              tableView.contentOffset.y < tableView.visibleSize.height * 2.0
+              tableView.contentOffset.y > tableView.contentSize.height - tableView.visibleSize.height * 2.0
         else { return }
         
         coordinator.send(viewAction: .paginateBackwards)
     }
+
+    private func scrollToTop(animated: Bool) {
+        tableView.scrollToRow(at: IndexPath(item: timelineItemsIDs.count - 1, section: 0), at: .bottom, animated: animated)
+    }
```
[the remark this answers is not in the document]
Velin92: yes

PR #1367, opened 2023-07-20
Fixes #1262 - Compute correct sizes for portrait videos

### LINKED ISSUE #1262: Portrait videos are shown with distorted landscape aspect thumbnail
### Steps to reproduce

1. take a portrait aspect video
2. send it to an (e2ee) room
3. observe horrible distorted landscape aspect thumbnail
IMG_0800

#### What did you expect?

thumbnails should honour orientation metadata in videos

#### What happened instead?

orientation metadata ignored => awful thumbnail

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

276

### Homeserver

_No response_

### Will you send logs?

No

stefanceriu:
When uploading videos the `naturalSize` does not take the `preferredTransform` into consideration resulting in portrait videos reporting inverted values.


ElementX/Sources/Services/Media/MediaUploadingPreprocessor.swift
```diff
@@ -432,4 +437,15 @@ struct MediaUploadingPreprocessor {
...
+    private func isVideoTransformPortrait(_ transform: CGAffineTransform) -> Bool {
+        switch (transform.a, transform.b, transform.c, transform.d) {
+        case (0, 1, -1, 0):
+            return true
+        case (0, -1, 1, 0):
+            return true
+        default:
+            return false
+        }
+    }
```
alfogrillo, line 450: I think, it would be more clean moving this code into a separate extension like this:

```swift
private extension AVAssetTrack {
    var size: CGSize {
        get async throws {
            let naturalSize = try await load(.naturalSize)
            guard mediaType == .video else {
                return naturalSize
            }

            let transform = try await load(.preferredTransform)

            switch (transform.a, transform.b, transform.c, transform.d) {
            case (0, 1, -1, 0), (0, -1, 1, 0):
                return CGSize(width: naturalSize.height, height: naturalSize.width)
            default:
                return CGSize(width: naturalSize.width, height: naturalSize.height)
            }
        }
    }
}
```

alfogrillo (approved):
LGTM 👍🏻
(left an optional suggestion to improve code readability)

ElementX/Sources/Services/Media/MediaUploadingPreprocessor.swift
stefanceriu: Very nice, I like it!

alfogrillo (approved):
LGTM! 👍🏻

PR #1368, opened 2023-07-20
Location event body representation

alfogrillo:
This PR:
- Updates the representation of a location event which is now represented with a localized string (instead of the event's `body`)
- Changes the `body` format when sending location events to other people

**Result**
| Before | After |
| --- | --- |
|b1|a1|
|b2|a2|


ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -675,6 +675,6 @@ private extension RoomFlowCoordinator {
 
 private extension GeoURI {
     var bodyMessage: String {
-        String(format: "Location was shared at %@ as of %@", string, Date().description)
+        "Location was shared at \(string)"
```
stefanceriu, line 678: Is this not being localised intentional? 🤔

stefanceriu (approved):
lgtm 👍

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
alfogrillo: Yep yep, this is the fallback representation for a location event for clients not supporting the map rendering.

callumu:
Looks good 👍

PR #1370, opened 2023-07-20
Fix tint colour on alerts and dialogs.

pixlwave:
I honestly thought Apple would fix this as it seems a pretty bad bug but as they haven't and the latest Introspect makes it easy to introspect a window, this fixes the behaviour:

| Old behaviour | Fixed behaviour |
| - | - |
| Simulator Screen Recording - iPhone 14 - 2023-07-20 at 12 02 13 | Simulator Screen Recording - iPhone 14 - 2023-07-20 at 12 03 55 |


alfogrillo (approved):
LGTM 🎉

PR #1373, opened 2023-07-20
Disable tilt and rotate interactions on maps

alfogrillo:
This PR disables:

- Pinch to rotate
- Pan with two fingers to tilt

on maps


stefanceriu (approved):
I personally like rotating maps but if this is what product wants ¯\_(ツ)_/¯

alfogrillo:
> I personally like rotating maps but if this is what product wants ¯_(ツ)_/¯

I personally like it as well. 😕

PR #1375, opened 2023-07-20
Fix for incorrect icon in file replies

### LINKED ISSUE #1319: Incorrect icon for attachments in reply
### Steps to reproduce

1. Upload a file (txt, doc, pdf, etc)
2. Reply to the message containing the attachment
3. Note the icon used

#### What did you expect?
Icon in the reply should be the file icon

#### What happened instead?
Icon in the reply is that of an audio file

### Your phone model

iPhone 12 mini

### Operating system version

iOS 16.5.1

### Application version

Element 1.2.1 (281)

### Homeserver

matrix.org

### Will you send logs?

Yes

Velin92:
fixes #1319


alfogrillo (approved):
LGTM
(just check if there are impacted snapshot tests)

PR #1376, opened 2023-07-21
Fix tint colour in dark mode.

pixlwave:
The tint colour is broken in dark mode following #1370. This PR fixes it.

### DIFF
--- a/ElementX/Sources/Application/Application.swift
+++ b/ElementX/Sources/Application/Application.swift
@@ -37,7 +37,7 @@ struct Application: App {
                 .statusBarHidden(shouldHideStatusBar)
                 .introspect(.window, on: .iOS(.v16)) { window in
                     // Workaround for SwiftUI not consistently applying the tint colour to Alerts/Confirmation Dialogs.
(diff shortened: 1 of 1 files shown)


ElementX/Sources/Application/Application.swift
```diff
@@ -37,7 +37,9 @@ struct Application: App {
                 .statusBarHidden(shouldHideStatusBar)
                 .introspect(.window, on: .iOS(.v16)) { window in
                     // Workaround for SwiftUI not consistently applying the tint colour to Alerts/Confirmation Dialogs.
-                    window.tintColor = UIColor(Color.compound.textActionPrimary)
+                    window.tintColor = UIColor(named: "colorTextActionPrimary",
+                                               in: Bundle(identifier: "CompoundDesignTokens-CompoundDesignTokens-resources"),
+                                               compatibleWith: nil)
```
pixlwave, line 42: I definitely don't like this, but its good enough for a fix for now. I'll see about getting the Compound colours generated as `UIColour` too.

Velin92 (approved):

alfogrillo (approved):
LGTM

PR #1377, opened 2023-07-21
Enable animations + delay + possible crash fix

Velin92:
@manuroe let me know if we are happy merging scroll animations fully.
- Removes the animation FF and sets the animations true by default
- This PR also enabled them after a 2 seconds delay to hide the SS proxy issue (however this depends a lot on the user's connection 2 seconds might not be enough sometimes)
- Fixes a possible crash


pixlwave (approved):
LGTM 👍

manuroe (approved):
I have not looked at the code but the behavior of the Element PR build is great!

PR #1378, opened 2023-07-21
Reduce boilerplate in Developer Options Screen

alfogrillo:
This PR refactors the developer options screen to minimize the boilerplate.
Now most of the source of truth come directly from the `AppSettings` type.

### DIFF
--- a/ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/DeveloperOptionsScreenModels.swift
+++ b/ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/DeveloperOptionsScreenModels.swift
@@ -48,7 +48,6 @@ protocol DeveloperOptions: AnyObject {
     var readReceiptsEnabled: Bool { get set }
     var isEncryptionSyncEnabled: Bool { get set }
     var notificationSettingsEnabled: Bool { get set }
-    var timelineDiffableAnimationsEnabled: Bool { get set }
 }
 
 extension AppSettings: DeveloperOptions { }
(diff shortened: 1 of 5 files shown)


ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/DeveloperOptionsScreenModels.swift
```diff
@@ -24,19 +24,30 @@ struct DeveloperOptionsScreenViewState: BindableState {
...
 enum DeveloperOptionsScreenViewAction {
-    case changedShouldCollapseRoomStateEvents
-    case changedUserSuggestionsEnabled
-    case changedReadReceiptsEnabled
-    case changedIsEncryptionSyncEnabled
-    case changedNotificationSettingsEnabled
     case clearCache
 }
+
+protocol DeveloperOptions: AnyObject {
```
stefanceriu, line 45: Can we call this `DeveloperOptionsProtocol` so that it matches the naming convention used throughout the app please?

stefanceriu (approved):
Very very nice, love it! 👏

ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/DeveloperOptionsScreenModels.swift
alfogrillo: Right, applied [here](https://github.com/vector-im/element-x-ios/pull/1378/commits/c82da03814ad7156a5493083deaf21a0856eb045)!

PR #1383, opened 2023-07-23
Tweaks for macOS

### LINKED ISSUE #1280: Compose button is moving around the screen on macOS
### Steps to reproduce

1. Open the Element X on macOS
2. Look at the compose button

#### What did you expect?

The compose button to stay still.

#### What happened instead?

The compose button was moving around.

### Your phone model

_No response_

### Operating system version

13.4.1 (22F82)

### Application version

1.1.8 (54)

### Homeserver

matrix.org

### Will you send logs?

Yes

pixlwave:
This PR makes 3 tweaks the use on macOS:
- Fixes #1280.
- Allows right-clicking on timeline items again by reusing the timeline menu actions to compute a context menu when on macOS. This is a more expected gesture on macOS, but long clicking is still available for those who want that.
- Uses an inline preview when uploading a media file as the floating QuickLook panel was obscuring the cancel/send buttons which was confusing
    - This change has one caveat: as noted in the [docs](https://developer.apple.com/documentation/quicklook/qlpreviewcontroller#3668757), previews are no longer interactive on Mac, so you can't play a video, zoom into an image or scroll through a document. The limitation seems worthwhile to me given that you can actually see the send button though.

**Context Menu**
Context

https://github.com/vector-im/element-x-ios/assets/6060466/1367a10c-b870-42e9-8b62-cf0a8ed48f7e

### DIFF
--- a/ElementX/Sources/Screens/MediaUploadPreviewScreen/View/MediaUploadPreviewScreen.swift
+++ b/ElementX/Sources/Screens/MediaUploadPreviewScreen/View/MediaUploadPreviewScreen.swift
@@ -20,13 +20,19 @@ import SwiftUI
 struct MediaUploadPreviewScreen: View {
     @ObservedObject var context: MediaUploadPreviewScreenViewModel.Context
     
+    var title: String {
+        ProcessInfo.processInfo.isiOSAppOnMac ? context.viewState.title ?? "" : ""
+    }
+    
     var body: some View {
         PreviewView(context: context,
                     fileURL: context.viewState.url,
                     title: context.viewState.title)
             .id(UUID())
+            .navigationTitle(title)
+            .navigationBarTitleDisplayMode(.inline)
             .disabled(context.viewState.shouldDisableInteraction)
-            .ignoresSafeArea(edges: .bottom)
+            .ignoresSafeArea(edges: [.horizontal, .bottom])
             .toolbar { toolbar }
             .interactiveDismissDisabled()
     }
@@ -52,15 +58,19 @@ private struct PreviewView: UIViewControllerRepresentable {
     let fileURL: URL
     let title: String?
 
-    func makeUIViewController(context: Context) -> UINavigationController {
+    func makeUIViewController(context: Context) -> UIViewController {
         let previewController = QLPreviewController()
         previewController.dataSource = context.coordinator
         previewController.delegate = context.coordinator
-
-        return UINavigationController(rootViewController: previewController)
+        
+        if ProcessInfo.processInfo.isiOSAppOnMac {
+            return previewController
+        } else {
+            return UINavigationController(rootViewController: previewController)
+        }
     }
 
-    func updateUIViewController(_ uiViewController: UINavigationController, context: Context) { }
+    func updateUIViewController(_ uiViewController: UIViewController, context: Context) { }
 
     func makeCoordinator() -> Coordinator {
         Coordinator(view: self)
--- a/ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
+++ b/ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
@@ -95,11 +95,37 @@ struct RoomScreenViewState: BindableState {
     var readReceiptsEnabled: Bool
     var isEncryptedOneToOneRoom = false
     
+    var composerMode: RoomScreenComposerMode = .default
+    let scrollToBottomPublisher = PassthroughSubject<Void, Never>()
+    
     var bindings: RoomScreenViewStateBindings
     
+    /// A closure providing the actions to show when long pressing on an item in the timeline.
     var timelineItemMenuActionProvider: (@MainActor (_ itemId: TimelineItemIdentifier) -> TimelineItemMenuActions?)?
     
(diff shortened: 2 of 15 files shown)


ElementX/Sources/Screens/MediaUploadPreviewScreen/View/MediaUploadPreviewScreen.swift
```diff
@@ -20,13 +20,19 @@ import SwiftUI
...
     var body: some View {
         PreviewView(context: context,
                     fileURL: context.viewState.url,
                     title: context.viewState.title)
             .id(UUID())
+            .navigationTitle(title)
+            .navigationBarTitleDisplayMode(.inline)
             .disabled(context.viewState.shouldDisableInteraction)
-            .ignoresSafeArea(edges: .bottom)
+            .ignoresSafeArea(edges: [.horizontal, .bottom])
```
pixlwave, line 35: This is the only change that affects iPadOS/iOS - iPhone in landscape would put white bars either side of the preview when you go fullscreen, which this fixes.

ElementX/Sources/Screens/RoomScreen/View/TimelineItemMenu.swift
```diff
@@ -79,20 +83,37 @@ enum TimelineItemMenuAction: Identifiable, Hashable {
...
+        case .copy: return Label(L10n.actionCopy, systemImage: "doc.on.doc")
+        case .edit: return Label(L10n.actionEdit, systemImage: "pencil.line")
+        case .copyPermalink: return Label(L10n.actionCopyLinkToMessage, systemImage: "link")
+        case .reply: return Label(L10n.actionReply, systemImage: "arrowshape.turn.up.left")
+        case .forward: return Label(L10n.actionForward, systemImage: "arrowshape.turn.up.right")
+        case .redact: return Label(L10n.actionRemove, systemImage: "trash")
+        case .viewSource: return Label(L10n.actionViewSource, systemImage: "doc.text.below.ecg")
+        case .retryDecryption: return Label(L10n.actionRetryDecryption, systemImage: "arrow.down.message")
+        case .report: return Label(L10n.actionReportContent, systemImage: "exclamationmark.bubble")
+        case .react: return Label(L10n.screenRoomTimelineAddReaction, systemImage: "hand.thumbsup")
```
pixlwave, line 99: Not sure if either of these are correct, the image isn't shown anyway but the title is "Add emoji" 🤔

ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
```diff
@@ -95,11 +95,37 @@ struct RoomScreenViewState: BindableState {
...
     
+    /// A closure providing the actions to show when long pressing on an item in the timeline.
     var timelineItemMenuActionProvider: (@MainActor (_ itemId: TimelineItemIdentifier) -> TimelineItemMenuActions?)?
     
-    var composerMode: RoomScreenComposerMode = .default
+    /// Builds the contents of the context menu shown when right clicking an item in the timeline on a Mac.
+    @MainActor @ViewBuilder
+    func macContextMenu(item: RoomTimelineItemProtocol, send: @escaping (TimelineItemMenuAction) -> Void) -> some View {
+        if ProcessInfo.processInfo.isiOSAppOnMac {
+            if let menu = timelineItemMenuActionProvider?(item.id) {
```
stefanceriu, line 110: This isn't techincally a menu, just a list of actions

ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
```diff
@@ -95,11 +95,37 @@ struct RoomScreenViewState: BindableState {
...
+    
     var bindings: RoomScreenViewStateBindings
     
+    /// A closure providing the actions to show when long pressing on an item in the timeline.
     var timelineItemMenuActionProvider: (@MainActor (_ itemId: TimelineItemIdentifier) -> TimelineItemMenuActions?)?
     
-    var composerMode: RoomScreenComposerMode = .default
+    /// Builds the contents of the context menu shown when right clicking an item in the timeline on a Mac.
+    @MainActor @ViewBuilder
+    func macContextMenu(item: RoomTimelineItemProtocol, send: @escaping (TimelineItemMenuAction) -> Void) -> some View {
```
stefanceriu, line 108: Are you sure this View belongs here? I would've expected it to be to be within a separate file withing the View subfolder

stefanceriu (approved):
Very nice, one step closer to using this as a desktop daily driver 🙌

ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
pixlwave: Fair, I wasn't really sure where to put it and was trying to use it in a similar was to the action provider, but I'll split it out 👍

ElementX/Sources/Screens/RoomScreen/View/TimelineItemMenu.swift
pixlwave: Updated to `React` after feedback from Design.

PR #1385, opened 2023-07-24
Default media upload file mime types to `application/octet-stream` so…

stefanceriu:
… that uploading doesn't fail rust side

Fixes problems in uploading less common file types like dmg, apk etc.


pixlwave (approved):
Ahhhh! Wonder if we should make mimeType non-optional on the Rust side then? Or nil coalesce to octet-stream in Rust? Seems pointless allowing it to accept something that will fail.

stefanceriu:
> Seems pointless allowing it to accept something that will fail.
The main reason for not doing it like that to being with was that the MediaInfo types are reused for received media too and those do have these fields as optionals.

PR #1389, opened 2023-07-24
Make the app version and the device ID copyable in the Settings screen.

### LINKED ISSUE #623: can't copypaste version or device id
the app version and device id in settings is not copyable, which is needlessly frustrating.

pixlwave:
Fixes #623

### DIFF
--- a/ElementX/Sources/Screens/Settings/SettingsScreen/View/SettingsScreen.swift
+++ b/ElementX/Sources/Screens/Settings/SettingsScreen/View/SettingsScreen.swift
@@ -204,7 +204,8 @@ private extension TimelineStyle {
 
 struct SettingsScreen_Previews: PreviewProvider {
     static let viewModel = {
-        let userSession = MockUserSession(clientProxy: MockClientProxy(userID: "@userid:example.com"),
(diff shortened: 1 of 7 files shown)


ElementX/Sources/Screens/Settings/SettingsScreen/View/SettingsScreen.swift
```diff
@@ -164,8 +164,10 @@ struct SettingsScreen: View {
             VStack {
                 versionText
                     .frame(maxWidth: .infinity)
+                    .textSelection(.enabled)
                 
                 context.viewState.deviceID.map(Text.init)
```
Velin92, line 169: I'll be honest... I don't like this.
I would rather have:

```suggestion
                Text(context.viewState.deviceID)
```

pixlwave: Ah, but its mapping an optional, so its the single line version of 

```
if let deviceID = context.viewState.deviceID {
    Text(deviceID)
}
```

pixlwave: Updated 👍

PR #1392, opened 2023-07-24
SwiftUI timeline refactor + ObservedObject

Velin92:
This is the best approach I could achieve so far.

The performances are great, the timeline has never been so fast, however there are some issues with any cell animation that changes the vertical size of a cell like when the delivery status disappears (creating an ugly jump) or when opening or closing a colalpsible item.

I have no idea why this happens since reaction animations work great.

https://github.com/vector-im/element-x-ios/assets/34335419/e1a37f

https://github.com/vector-im/element-x-ios/assets/34335419/9e4c7f05-c70d-458c-928e-46b7debd67e009-b661-4ded-8e5b-a68d3c63eaed


ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
```diff
@@ -24,7 +24,7 @@ struct RoomScreen: View {
     
     var body: some View {
         timeline
-            .background(Color.compound.bgCanvasDefault.ignoresSafeArea()) // Kills the toolbar translucency.
```
stefanceriu, line 27: We should probably keep this comment, no?

ElementX/Sources/Screens/RoomScreen/View/Timeline/CollapsibleRoomTimelineView.swift
```diff
@@ -18,34 +18,35 @@ import SwiftUI
...
+                ForEach(groupedViewStates) { viewState in
+                    RoomTimelineItemView(viewState: viewState)
                 }
-            }.transition(.opacity.animation(.elementDefault))
+            }
         }
         .disclosureGroupStyle(CollapsibleRoomTimelineItemDisclosureGroupStyle())
-        .transaction { transaction in
-            transaction.animation = .noAnimation // Fixes weird animations on the disclosure indicator
-        }
```
stefanceriu, line 41: 👏

ElementX/Sources/Screens/RoomScreen/View/TimelineView.swift
```diff
@@ -14,68 +14,146 @@
...
-        let context: RoomScreenViewModel.Context
-        
-        init(viewModelContext: RoomScreenViewModel.Context) {
-            context = viewModelContext
-            
-            if viewModelContext.viewState.itemViewModels.isEmpty {
-                viewModelContext.send(viewAction: .paginateBackwards)
+
+    private let bottomID = "bottomID"
+    private let topID = "topID"
```
stefanceriu, line 27: I would name these something harder to clash on e.g. `RoomTimelineTopPinIdentifier`, `RoomTimelineBottomPinIdentifier`. You never know what might come out of the SDK in the future

stefanceriu (approved):
Left some minor comments but otherwise I absolutely love it! 👏👏👏

ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineItemStatusView.swift
```diff
@@ -23,11 +23,15 @@ struct TimelineItemStatusView: View {
     @EnvironmentObject private var context: RoomScreenViewModel.Context
 
     private var isLastOutgoingMessage: Bool {
-        context.viewState.timelineIDs.last == timelineItem.id.timelineID &&
-            timelineItem.isOutgoing
+        timelineItem.isOutgoing && context.viewState.timelineViewState.timelineIDs.last == timelineItem.id.timelineID
     }
 
     var body: some View {
+        mainContent
```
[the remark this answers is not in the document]
Velin92: at first I wanted to add a custom transition but in the end is not up to me to decide, I prepared the code to do so, I can revert but is not really necessary

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
Velin92: i remember checking with Doug and since technically speaking now the timeline is reversed the toolbar transparency is not an issue anymore... actually I wonder, can we remove this line entirely? @pixlwave

pixlwave: Yeah, it wasn't very clear but the comment was pointing out that it broke the translucency rather than intentionally disabling it. This re-write now works perfectly thanks to content insets being applied.

The screen needs this background colour for our grey background in dark mode. Could move it to the timeline, but I'd say its probably nicer to keep it theming at the top-level of the body here.

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemView.swift
```diff
@@ -16,17 +16,27 @@
 import SwiftUI
 
 struct RoomTimelineItemView: View {
-    @ObservedObject var viewModel: RoomTimelineItemViewModel
+    @EnvironmentObject private var context: RoomScreenViewModel.Context
```
pixlwave, line 19: I don't think it matters given that both stylers observe this, but if this was passed in as a `let` to both TimelineView and this view, you could remove another source of observation.

Again for a follow-up PR, maybe we should audit the use of the context within the timeline items as some are observing it purely to send an action. We could possibly get some more performance through trickery like creating an environment value and scoping the use with key paths.

ElementX/Sources/Screens/RoomScreen/View/TimelineView.swift
```diff
@@ -14,68 +14,146 @@
...
+
+                topPin
             }
-        }
-        
-        /// Updates the specified table view's properties from the current view state.
-        func update(tableViewController: TimelineTableViewController, timelineStyle: TimelineStyle) {
-            if tableViewController.timelineStyle != timelineStyle {
-                tableViewController.timelineStyle = timelineStyle
+            .introspect(.scrollView, on: .iOS(.v16)) { uiKitScrollView in
```
pixlwave, line 50: Seems totally reasonable to match the type name here.
```suggestion
            .introspect(.scrollView, on: .iOS(.v16)) { uiScrollView in
```

ElementX/Sources/Screens/RoomScreen/View/TimelineView.swift
```diff
@@ -14,68 +14,146 @@
...
-                tableViewController.timelineItemsDictionary = context.viewState.itemsDictionary
+            .scaleEffect(x: 1, y: -1)
+            .onReceive(scrollToBottomPublisher) { _ in
+                withAnimation {
+                    scrollView.scrollTo(bottomID)
+                }
             }
-            if tableViewController.canBackPaginate != context.viewState.canBackPaginate {
-                tableViewController.canBackPaginate = context.viewState.canBackPaginate
+            .scrollDismissesKeyboard(.interactively)
```
pixlwave, line 72: Woah wait?! Does this finally work properly?!! 🤯

ElementX/Sources/Screens/RoomScreen/View/TimelineView.swift
```diff
@@ -14,68 +14,146 @@
...
+                        .shadow(color: .primary.opacity(0.33), radius: 2.0)
+                }
+                .padding()
+        }
+        .opacity(scrollToBottomButtonVisible ? 1.0 : 0.0)
+        .accessibilityHidden(!scrollToBottomButtonVisible)
+        .animation(.elementDefault, value: scrollToBottomButtonVisible)
+    }
+
+    private func tryPaginateBackwards() {
```
pixlwave, line 139: Totally subjective, but to my mind `paginateBackwardsIfNeeded()` is a more fitting name here, indicating its safe to call this regularly because it will only paginate if there is more history available and there isn't enough content available offscreen.

pixlwave (approved):
This looks amazing, love it! 🤩

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -245,7 +256,7 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
     }
         
     private func buildTimelineViews() {
-        var timelineItemsDictionary = OrderedDictionary<String, RoomTimelineItemViewModel>()
+        var timelineItemsDictionary = OrderedDictionary<String, RoomTimelineItemViewState>()
```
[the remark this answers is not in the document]
Velin92: I still like the dictionary because sometimes we need to do a lot of item look ups given IDs, and searching for a value in the array is way less performant than just accessing the ID directly in a dictionary.

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -666,7 +667,7 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
     // MARK: - Reactions
     
     private func showEmojiPicker(for itemID: TimelineItemIdentifier) {
-        guard let item = state.itemsDictionary[itemID.timelineID], item.isReactable else { return }
+        guard let item = state.timelineViewState.itemsDictionary[itemID.timelineID], item.isReactable else { return }
```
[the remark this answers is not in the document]
Velin92: exactly that's why I would keep it. It has the best of both worlds.

ElementX/Sources/Screens/RoomScreen/View/TimelineView.swift
Velin92: works pretty well for me, feels great, there is even a bit of animation

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemView.swift
Velin92: the let here without the observation would create some issues with MainActor conformance and the context.send usage... Maybe the fix is as easy as adding MainActor to the view, but I don't know if this could mess or slow down anything, honestly is a can of worms I don't want to open for now, but yeah we should definitely revisit it.

ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
```diff
@@ -184,3 +173,19 @@ struct RoomMemberState {
...
+
+    var timelineIDs: [String] {
+        itemsDictionary.keys.elements
+    }
+
+    var itemViewStates: [RoomTimelineItemViewState] {
+        itemsDictionary.values.elements
+    }
+
+    var paginateAction: (() -> Void)?
```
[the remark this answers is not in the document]
Velin92: dunno, don't really like implicitly unwrapped closures... should be safe to do, but is just a thing I don't like

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemViewState.swift
```diff
@@ -16,32 +16,29 @@
...
 
-    convenience init(item: RoomTimelineItemProtocol, groupStyle: TimelineGroupStyle) {
-        self.init(type: .init(item: item), groupStyle: groupStyle)
-    }
-
-    init(type: RoomTimelineItemType, groupStyle: TimelineGroupStyle) {
-        self.type = type
-        self.groupStyle = groupStyle
+    var id: String {
+        identifier.timelineID
```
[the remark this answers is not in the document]
Velin92: however we never use the viewState anywhere for business logic, this is only used at a view level.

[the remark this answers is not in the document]
Velin92: Also the make things more consistent that's why every time we go through any business logic function we always constrained the parameter type to be `TimelineIdentifier`

ElementX/Sources/Screens/RoomScreen/View/Timeline/CollapsibleRoomTimelineView.swift
```diff
@@ -18,34 +18,35 @@ import SwiftUI
...
 }
 
 private struct CollapsibleRoomTimelineItemDisclosureGroupStyle: DisclosureGroupStyle {
     func makeBody(configuration: Configuration) -> some View {
         VStack(spacing: 0.0) {
-            Button { configuration.isExpanded.toggle() } label: {
+            Button {
+                withAnimation {
+                    configuration.isExpanded.toggle()
+                }
```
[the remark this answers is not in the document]
Velin92: sadly animation .animation(.elementDefault, value: configuration.isExpanded) doesn't work for some reason (i think is because this would rely on being used on a internal state, while the configuration makes the body so it won't be able to listen to changes).

However withElementAnimation works just fine.

PR #1396, opened 2023-07-25
Fix LocationRoomTimelineView clipping

alfogrillo:
Fix the clipping on `LocationRoomTimelineView`. It was previously visible just with an invalid api key and the modern style.

| Before | After |
| --- | --- |
|b|a|


pixlwave (approved):
👍

PR #1398, opened 2023-07-25
Specify ElementX target for coverage in integrations plan.

pixlwave:
This PR update coverage collecting to match UI tests (specify only the EX target) along with clearing derived data on each run as the Integration tests appeared to be uploading coverage from the UI tests.


stefanceriu (approved):
🚀 it!

PR #1401, opened 2023-07-25
Fix problem with disclosure groups randomly not disclosing by replaci…

stefanceriu:
…ng with custom component

- 99% of the code is the same as we were already massively customising it

Seems to be the same problem as here following our move to lazyVStack https://stackoverflow.com/questions/74120571/swiftui-disclosuregroup-wont-expand-inside-lazyvstack


Velin92 (approved):
LGTM

PR #1403, opened 2023-07-25
Two Sync Loop always enabled by default and removed the feature flag

Velin92:
We have this in nightly since a month, seems to be quite stable, so we can start rolling it out.


stefanceriu (approved):
lgtm 👍

alfogrillo (approved):
LGTM

PR #1405, opened 2023-07-25
Use generated UIColors everywhere.

pixlwave:
This PR uses the generated `UIColor`s from Compound instead of wrapping SwiftUI colours (which caused an issue in dark mode when setting the window's tint colour).

Additionally fixes a bug introduced by #1383 which lost the destructive role for some timeline action buttons.


Velin92 (approved):
LGTM

PR #1406, opened 2023-07-26
Fix more unit tests.

pixlwave:
- Randomise test order.
- Use deferred fulfilment in more places.
- Make sure to change something before awaiting the fulfilment.
- Build a fresh AppSettings after reset().


stefanceriu (approved):
nicely done 👍

PR #1407, opened 2023-07-26
Reverting to TableView but with SwiftUI compatibility

Velin92:
Also if we really want to test and play around in nightly with SwiftUI I have set up a feature flag that allows us to change the timeline to SwiftUI whenever we wish.

### DIFF
--- a/ElementX/Sources/Application/AppSettings.swift
+++ b/ElementX/Sources/Application/AppSettings.swift
@@ -215,6 +215,6 @@ final class AppSettings {
     @UserPreference(key: UserDefaultsKeys.notificationSettingsEnabled, defaultValue: false, storageType: .userDefaults(store))
     var notificationSettingsEnabled
 
-    @UserPreference(key: UserDefaultsKeys.swiftUITimelineEnabled, defaultValue: false, storageType: .userDefaults(store))
+    @UserPreference(key: UserDefaultsKeys.swiftUITimelineEnabled, defaultValue: false, storageType: .volatile)
     var swiftUITimelineEnabled
 }
--- a/ElementX/Sources/Application/AppSettings.swift
+++ b/ElementX/Sources/Application/AppSettings.swift
@@ -34,6 +34,7 @@ final class AppSettings {
         case readReceiptsEnabled
         case hasShownWelcomeScreen
         case notificationSettingsEnabled
+        case swiftUITimelineEnabled
     }
     
     private static var suiteName: String = InfoPlistReader.main.appGroupIdentifier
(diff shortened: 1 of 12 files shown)


ElementX/Sources/Application/AppSettings.swift
```diff
@@ -213,4 +214,7 @@ final class AppSettings {
     
     @UserPreference(key: UserDefaultsKeys.notificationSettingsEnabled, defaultValue: false, storageType: .userDefaults(store))
     var notificationSettingsEnabled
+
+    @UserPreference(key: UserDefaultsKeys.swiftUITimelineEnabled, defaultValue: false, storageType: .userDefaults(store))
```
pixlwave, line 218: Maybe we shouldn't persist this one so we know bugs reported are for whatever is active?

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
```diff
@@ -82,13 +82,48 @@ struct RoomScreen: View {
...
+    @ViewBuilder
+    private var timelineSwitch: some View {
+        if context.viewState.swiftUITimelineEnabled {
+            TimelineView(viewState: context.viewState.timelineViewState,
+                         scrollToBottomButtonVisible: $context.scrollToBottomButtonVisible)
+        } else {
+            UITimelineView()
+                .overlay(alignment: .bottomTrailing) {
+                    scrollToBottomButton
+                }
```
pixlwave, line 105: Double checking: Does this overlay in the right place when the keyboard is shown? I know the order matters between the `overlay` and the `safeAreaInset` but can't remember which way round it is meant to be.

ElementX/Sources/Screens/RoomScreen/View/TimelineView.swift
```diff
@@ -164,14 +164,14 @@ struct TimelineView: View {
 
 // MARK: - Previews
 
-struct TimelineTableView_Previews: PreviewProvider {
+struct UITimelineTableView_Previews: PreviewProvider {
```
pixlwave, line 168: I think this should be?
```suggestion
struct TimelineView_Previews: PreviewProvider {
```

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
```diff
@@ -82,13 +82,50 @@ struct RoomScreen: View {
             }
     }
 
+    @ViewBuilder
```
pixlwave, line 85: Does this one need to be `@ViewBuilder`?

pixlwave (approved):
*waves bye bye to smoothness for now* 😢

ElementX/Sources/Application/AppSettings.swift
Velin92: oh yeah better to keep it volatile I guess

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
Velin92: looks good on my side, also i checked the previous code, this is the exact position it was present

PR #1412, opened 2023-07-27
FF to enable filtering notifications by push rules

Velin92:
Now that notifications do a small sliding sync when received, they should be able to get the push rules and filter automatically.

Since this is experimental, this needs to be enabled as a FF.

### DIFF
--- a/ElementX/Sources/Application/AppSettings.swift
+++ b/ElementX/Sources/Application/AppSettings.swift
@@ -190,7 +190,7 @@ final class AppSettings {
     @UserPreference(key: UserDefaultsKeys.pusherProfileTag, storageType: .userDefaults(store))
     var pusherProfileTag: String?
 
-    /// Alows notifications to he filtered based on the push context of the room
+    /// Allows notifications to be filtered based on the push context of the room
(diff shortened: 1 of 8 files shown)


ElementX/Sources/Application/AppSettings.swift
```diff
@@ -188,6 +188,10 @@ final class AppSettings {
     /// Tag describing which set of device specific rules a pusher executes.
     @UserPreference(key: UserDefaultsKeys.pusherProfileTag, storageType: .userDefaults(store))
     var pusherProfileTag: String?
+
+    /// Alows notifications to he filtered based on the push context of the room
```
pixlwave, line 192:
```suggestion
    /// Allows notifications to be filtered based on the push context of the room
```

NSE/Sources/Other/NSESettings.swift
```diff
@@ -21,4 +21,8 @@ final class NSESettings {
 
     /// UserDefaults to be used on reads and writes.
     private static var store: UserDefaults! = UserDefaults(suiteName: suiteName)
+
+    /// Alows notifications to he filtered based on the push context of the room
```
pixlwave, line 25: Same here

Velin92:
```suggestion
    /// Allows notifications to be filtered based on the push context of the room
```

Issue #1402, opened 2023-07-25
Message Reactions Design Review


callumu:
### Steps to reproduce

1. Add a 4px offset from the first emoji to the side of the bubble:

<img width="878" alt="Screenshot 2023-07-25 at 16 18 23" src="https://github.com/vector-im/element-x-ios/assets/110109933/1b78b310-9531-4966-9e96-4a5c046a8b47">
<img width="1215" alt="Screenshot 2023-07-25 at 16 18 10" src="https://github.com/vector-im/element-x-ios/assets/110109933/930ba8c7-3b60-45d0-8f6a-140de1e36a43">

2. The Show more / show less animation is jerky - it's probably best to be removed for now:

https://github.com/vector-im/element-x-ios/assets/110109933/92f61037-eb9d-411f-b1d3-f997919ca808

This jerky animation also shows when there are lots of emojis and some are removed.

3. Can we add a haptic for selecting an emoji?

4. Emoji bottom sheet - when launched with a carousel of emojis, the carousel launches in the incorrect position:

https://github.com/vector-im/element-x-ios/assets/110109933/83fc21b9-f4e2-4315-92c1-a72b53325605

5. Emoji bottom sheet - can the timestamp be vertically aligned with the user name? Can the size of the avatar, spacing between avatar and padding between name and matrix ID be checked against the designs? Can the padding of a carousel item between the number and the right of the selected state be increased if a number shows next to the emoji?

6. Animation when adding an emoji - can this be sped up x2?

7. Emoji icon in dark mode - check that it's dark/color/icon/secondary

### Outcome

...

### Will you send logs?

Yes

PR #1417, opened 2023-07-28
Reactions - Design review tweaks

langleyd:
Fixes https://github.com/vector-im/element-x-ios/issues/1402
- truncate reactions as per spec
- add haptic feedback on reacting
- Remove all collapse/expand animations on reactions
- tweak reaction and summary padding
- Fix alignment of time on reaction summary view
- Made sure reaction summary button in horizontal scroller is on screen on appear
- update screenshots


ElementX/Resources/Assets.xcassets/images/timeline/timeline-reaction-add-more.imageset/Contents.json
pixlwave, line 1: I have a feeling it would make more sense to make this asset render as a template by default, and apply the correct `foregroundColor` from Compound here instead of adding a dark variation of the same icon?

ElementX/Sources/Other/Extensions/String.swift
```diff
@@ -78,3 +78,12 @@ extension String {
         return whiteSpaces + "\u{2800}"
     }
 }
+
+extension String {
+    func ellipsize(length: Int) -> String {
+        guard count > length else {
+            return self
+        }
+        return "\(prefix(length))…"
```
pixlwave, line 87: I'm not sure I know how, but is it worth considering what do for for RTL strings here?

UITests/Sources/__Snapshots__/Application/en-GB-iPad-9th-generation.roomEncryptedWithAvatar.png
pixlwave, line 1: The new screenshot shows that adding reactions slightly reduces the padding of the trailing edge of outgoing bubbles to the screen's edge:
Screenshot 2023-07-28 at 3 29 39 pm

Weirdly, this only seems to be happening on iPad, on iPhone it looks fine.

pixlwave (approved):
LGTM, although there's a slight issue with the layout on iPad.

ElementX/Sources/Other/Extensions/String.swift
langleyd: Turns out it's handled for us, i.e. the text will flip for an RTL language.
image

This is the same behaviour for an iOS component like a Button(also uses an ellipsis).

UITests/Sources/__Snapshots__/Application/en-GB-iPad-9th-generation.roomEncryptedWithAvatar.png
langleyd: hrmm strange and looks ok in the ipad preview in xcode and on the simulator when I try it.

ElementX/Sources/Other/Extensions/String.swift
pixlwave: Oh nice! 🙌

UITests/Sources/__Snapshots__/Application/en-GB-iPad-9th-generation.roomEncryptedWithAvatar.png
langleyd: I actually had the padding incorrect, it should only have been applied on the leading side as per designs, not leading + trailing.

Not exactly sure why applying on both cause this but its not happening now at least.

PR #1425, opened 2023-07-31
SwiftUI timeline tweaks

stefanceriu:
Delay rendering bottom timeline items when scrolled up and avoid content shifts
- builds on top of langleyd/timeline_defer_new_items
- improve animations

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
+++ b/ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
@@ -299,7 +299,7 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol 
             }
         }
         
-        // The SwiftUI scroll view needs special handling, see `selectivellyUpdateTimelineItems`
+        // The SwiftUI scroll view needs special handling, see `selectivelyUpdateTimelineItems`
         if state.swiftUITimelineEnabled {
             selectivelyUpdateTimelineItems(timelineItemsDictionary: timelineItemsDictionary)
         } else {
--- a/ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
+++ b/ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
@@ -304,6 +304,7 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol 
             selectivelyUpdateTimelineItems(timelineItemsDictionary: timelineItemsDictionary)
         } else {
             state.timelineViewState.itemsDictionary = timelineItemsDictionary
+            state.timelineViewState.renderedTimelineIDs = Array(timelineItemsDictionary.keys)
         }
     }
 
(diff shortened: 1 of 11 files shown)


ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -276,29 +280,34 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
...
-                        timelineItemsDictionary.updateValue(updateViewstate(item: item, groupStyle: .middle),
+                        timelineItemsDictionary.updateValue(updateViewState(item: item, groupStyle: .middle),
                                                             forKey: item.id.timelineID)
                     }
                 }
             }
         }
         
-        state.timelineViewState.itemsDictionary = timelineItemsDictionary
+        // The SwiftUI scroll view needs special handling, see `selectivellyUpdateTimelineItems`
```
pixlwave, line 302:
```suggestion
        // The SwiftUI scroll view needs special handling, see `selectivelyUpdateTimelineItems`
```

ElementX/Sources/Screens/RoomScreen/View/TimelineView.swift
```diff
@@ -61,17 +61,17 @@ struct TimelineView: View {
...
             guard let scrollView = scrollViewAdapter.scrollView else {
                 return
             }
             let offset = scrollView.contentOffset.y + scrollView.contentInset.top
 
-            // We give it a bit of tollerance which solves the issue when of it being displayed when the keyboard appears
-            let scrollToBottomButtonVisibleValue = offset > 5
-            if scrollToBottomButtonVisibleValue != scrollToBottomButtonVisible {
-                scrollToBottomButtonVisible = scrollToBottomButtonVisibleValue
+            let isScrolledToBottom = offset <= 0
```
pixlwave, line 70: Pointing out this went from a tolerance of 5pt to 0. Maybe that's fine, just wanted to make sure its intended.

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -307,6 +316,41 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
...
     }
+    
+    /// With the timeline scroll being reversed, introducing items at it's top (i.e. bottom now) will make the content move upwards, which is unwanted when
+    /// reading history. Delay rendering new items until it reaches the bottom again.
+    private func selectivelyUpdateTimelineItems(timelineItemsDictionary: OrderedDictionary<String, RoomTimelineItemViewState>) {
+        var timelineViewState = state.timelineViewState
+        
+        let newItemIdentifiers = Array(timelineItemsDictionary.keys)
+        
+        if !state.bindings.isScrolledToBottom,
```
pixlwave, line 327: Personally I would flip this `if` for readability.

pixlwave (approved):
Looks reasonable to me 👍

ElementX/Sources/Screens/RoomScreen/View/TimelineView.swift
stefanceriu: I tested it with both timelines and it seemed fine, 5 felt to magical ¯\_(ツ)_/¯

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
stefanceriu: Switching it would mean to have 2 different places where we `render everything normally` to account for the other conditions failing. I don't think it will make it more readable so I would prefer to keep it like this.

PR #1426, opened 2023-07-31
Send read receipts as messages are displayed.

### LINKED ISSUE #639: Self-sent msgs mark rooms as unread
### Steps to reproduce

1. go into an unread room
2. send a message
3. go out; observe that the room is still marked as unread
4. go in again
5. go out again; observer thst the room is now marked as read

#### What did you expect?

reliable read state

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

1.0.23 (42)

### Homeserver

_No response_

### Will you send logs?

No

pixlwave:
This PR changes read receipt sending to match Android, tracking the ID of the last sent read receipt, and sending if needed as items are added and the timeline is scrolled.

Note: This isn't plummed into the SwiftUI timeline as we really need iOS 17's `scrollPosition(id:anchor:)` to determine visible items given we can't use the hacks from the home screen due to cells having unique sizes.

Fixes #639.

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
+++ b/ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
@@ -53,7 +53,7 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol 
         self.userIndicatorController = userIndicatorController
         self.notificationCenterProtocol = notificationCenterProtocol
         
-        super.init(initialViewState: RoomScreenViewState(roomId: timelineController.roomID,
+        super.init(initialViewState: RoomScreenViewState(roomID: timelineController.roomID,
                                                          roomTitle: roomProxy.roomTitle,
                                                          roomAvatarURL: roomProxy.avatarURL,
                                                          timelineStyle: appSettings.timelineStyle,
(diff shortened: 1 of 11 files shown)


ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -245,9 +245,46 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
         }
     }
     
-    private func markRoomAsRead() async {
-        notificationCenterProtocol.post(name: .roomMarkedAsRead, object: roomProxy.id)
-        _ = await timelineController.markRoomAsRead()
+    /// The ID of the newest item in the room that the user has seen.
+    /// This includes both event based items and virtual items.
+    private var lastReadItemID: TimelineItemIdentifier?
```
stefanceriu, line 250: Let's say I just open the app and deep link into a particular message. Would this `lastReadItemID` being nil cause any problems? 
I think it might move the read receipt to the wrong place, no? 

I have a feeling that deeplinks wouldn't be the only edge case. Is there any way in which we can fetch the current read receipt identifier from rust instead of storing it in memory?

pixlwave: We discussed this offline, it shouldn't be an issue in practice as you can't move the read receipt backwards, but long term we should look to Rust for an indication of `isRead` on timeline items.

PR #1428, opened 2023-07-31
Fix a bug when dismissing a media preview.

pixlwave:
Currently if you tap on a media item when the keyboard is visible, dismissing the item sometimes returns you to a view where the composer is pushed up by a big white empty space where the keyboard would normally be. This PR makes sure the composer resigns focus before displaying the preview.


stefanceriu (approved):
👍

PR #1430, opened 2023-07-31
Reactions UX updates

langleyd:
- Highlight selected emoji in full picker
- Fix the add more button in message actions bottom sheet
- Fix the order of count/key label on outgoing messages with bubble style(it was backward).
- Fix the size and padding of emoji in reactions


stefanceriu (approved):
lgtm 👍

PR #1431, opened 2023-08-01
Various tweaks

stefanceriu:
* Fix warnings on now non-async methods
* Made SwiftUI timeline feature flag non volatile


pixlwave (approved):
LGTM and adding for visibility that we discussed earlier about reverting the feature flag's volatility.

PR #1432, opened 2023-08-01
Send the SDK's short git SHA to rageshakes

stefanceriu:
Implemented this Rust side a long time ago but forgot to actually use it anywhere


pixlwave (approved):
Very nice. (Looks like unit tests need a bit of attention though)

Issue #1045, opened 2023-06-07
Add a link to the global settings in the room Notifications settings


giomfo:
## Description ##
Following https://github.com/vector-im/element-x-ios/issues/1004
In the room Notifications settings, when the setting is the default one, the sub-title "You can change it in your Global Settings" is clickable. The global Notification settings screen is opened on a click

## Suggested Design ##
<img width="237" alt="image" src="https://github.com/vector-im/element-x-ios/assets/8969772/9cc44ce0-db1a-4d80-85d5-c7f567b2626b">

PR #1433, opened 2023-08-01
Add a link to the global settings in the room notifications settings

nimau:
This PR implements #1045 

On the room notification screen, when the user taps on "You can change it in your global settings", the notification settings screen opens modally.

### DIFF
--- a/ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenCoordinator.swift
+++ b/ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenCoordinator.swift
@@ -189,7 +189,8 @@ final class RoomDetailsScreenCoordinator: CoordinatorProtocol {
     }
     
     private func presentNotificationSettings() {
-        let roomNotificationSettingsParameters = RoomNotificationSettingsScreenCoordinatorParameters(notificationSettingsProxy: parameters.notificationSettings,
+        let roomNotificationSettingsParameters = RoomNotificationSettingsScreenCoordinatorParameters(navigationStackCoordinator: parameters.navigationStackCoordinator,
+                                                                                                     notificationSettingsProxy: parameters.notificationSettings,
                                                                                                      roomProxy: parameters.roomProxy)
         let roomNotificationSettingsCoordinator = RoomNotificationSettingsScreenCoordinator(parameters: roomNotificationSettingsParameters)
         navigationStackCoordinator?.push(roomNotificationSettingsCoordinator)
--- a/ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenCoordinator.swift
+++ b/ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenCoordinator.swift
@@ -29,19 +29,24 @@ struct RoomDetailsScreenCoordinatorParameters {
 
 enum RoomDetailsScreenCoordinatorAction {
     case leftRoom
+    case presentNotificationSettingsScreen
 }
 
 final class RoomDetailsScreenCoordinator: CoordinatorProtocol {
     private let parameters: RoomDetailsScreenCoordinatorParameters
     private var viewModel: RoomDetailsScreenViewModelProtocol
-    private var cancellables: Set<AnyCancellable> = .init()
     private let selectedUsers: CurrentValueSubject<[UserProfileProxy], Never> = .init([])
     private var navigationStackCoordinator: NavigationStackCoordinator? {
         parameters.navigationStackCoordinator
     }
     
-    var callback: ((RoomDetailsScreenCoordinatorAction) -> Void)?
-    
+    private let actionsSubject: PassthroughSubject<RoomDetailsScreenCoordinatorAction, Never> = .init()
+    private var cancellables: Set<AnyCancellable> = .init()
+        
+    var actions: AnyPublisher<RoomDetailsScreenCoordinatorAction, Never> {
+        actionsSubject.eraseToAnyPublisher()
+    }
+        
     init(parameters: RoomDetailsScreenCoordinatorParameters) {
         self.parameters = parameters
         
@@ -65,7 +70,7 @@ final class RoomDetailsScreenCoordinator: CoordinatorProtocol {
             case .requestInvitePeoplePresentation:
                 self.presentInviteUsersScreen()
             case .leftRoom:
-                self.callback?(.leftRoom)
+                self.actionsSubject.send(.leftRoom)
             case .requestEditDetailsPresentation(let accountOwner):
                 self.presentRoomDetailsEditScreen(accountOwner: accountOwner)
             case .requestNotificationSettingsPresentation:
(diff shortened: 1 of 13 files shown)


ElementX/Sources/Screens/Settings/NotificationSettingsScreen/NotificationSettingsScreenCoordinator.swift
```diff
@@ -39,11 +46,16 @@ final class NotificationSettingsScreenCoordinator: CoordinatorProtocol {
...
-                                                        notificationSettingsProxy: parameters.notificationSettings)
+                                                        notificationSettingsProxy: parameters.notificationSettings,
+                                                        isModallyPresented: parameters.isModallyPresented)
     }
     
     func start() {
         viewModel.fetchInitialContent()
+        
+        viewModel.actions.sink { [weak self] _ in
+            self?.completion?(.close)
```
pixlwave, line 57: Would probably be sensible to switch on the action (even if there is only one right now), so that if a second is added in the future, it causes a compile error instead of unexpected behaviour.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/NotificationSettingsScreenCoordinator.swift
```diff
@@ -39,11 +46,16 @@ final class NotificationSettingsScreenCoordinator: CoordinatorProtocol {
...
+        }.store(in: &cancellables)
```
pixlwave, line 58:
```suggestion
        }
        .store(in: &cancellables)
```

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/NotificationSettingsScreenCoordinator.swift
```diff
@@ -30,6 +35,8 @@ final class NotificationSettingsScreenCoordinator: CoordinatorProtocol {
     private let actionsSubject: PassthroughSubject<NotificationSettingsScreenCoordinatorAction, Never> = .init()
     private var cancellables: Set<AnyCancellable> = .init()
     
+    var completion: ((NotificationSettingsScreenCoordinatorResult) -> Void)?
```
pixlwave, line 38: We have been slowly replacing this optional `completion` with the actions publisher below. Could we use that instead please rather than adding the old one back?

ElementX/Sources/Screens/RoomNotificationSettingsScreen/View/RoomNotificationSettingsScreen.swift
```diff
@@ -71,6 +71,9 @@ struct RoomNotificationSettingsScreen: View {
                 .compoundFormSectionHeader()
         } footer: {
             Text(context.viewState.strings.customSettingFootnote)
+                .onTapGesture {
+                    context.send(viewAction: .globalSettingsTapped)
+                }
```
pixlwave, line 76: It would be more correct here to add a link to the string and use an OpenURLAction to send the view action instead of a custom tap gesture.

ElementX/Sources/Screens/RoomNotificationSettingsScreen/RoomNotificationSettingsScreenCoordinator.swift
```diff
@@ -18,23 +18,51 @@ import Combine
...
+        let navigationStackCoordinator = NavigationStackCoordinator()
+        let notificationSettingsScreenParameters = NotificationSettingsScreenCoordinatorParameters(userNotificationCenter: UNUserNotificationCenter.current(),
+                                                                                                   notificationSettings: parameters.notificationSettingsProxy,
+                                                                                                   isModallyPresented: true)
+        let notificationSettingsScreenCoordinator = NotificationSettingsScreenCoordinator(parameters: notificationSettingsScreenParameters)
+        notificationSettingsScreenCoordinator.completion = { [weak self] _ in
+            self?.navigationStackCoordinator?.setSheetCoordinator(nil)
+        }
+        navigationStackCoordinator.setRootCoordinator(notificationSettingsScreenCoordinator)
+        self.navigationStackCoordinator?.setSheetCoordinator(navigationStackCoordinator)
```
pixlwave, line 66: Given this navigation as adding a sheet, it really shouldn't be done by this coordinator. The flow coordinator needs to know about any modals so that if a user tapped on a notification whilst this screen modal was presented, the flow coordinator would know the navigation state and that it needs to tear down the modal when pushing to the room.

Can you update the navigation so that this coordinator sends an action to the flow coordinator requesting the notification settings screen to be presented so it can all go through the state machine?

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/NotificationSettingsScreenModels.swift
```diff
@@ -17,11 +17,10 @@
 import Foundation
 import UIKit
 
-enum NotificationSettingsScreenViewModelAction { }
-
```
pixlwave, line 21: Whilst I see the logic of moving this next to the view actions, it makes the structure of this file different to all of the others where the vm actions are at the top and the view actions are at the bottom. Maybe if this is something we want to do, it should be a separate PR to update all the files consistently.

pixlwave (commented):
Very nice, one thought about navigation, otherwise all the other comments are minor.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/NotificationSettingsScreenModels.swift
nimau: Done.

ElementX/Sources/Screens/RoomNotificationSettingsScreen/RoomNotificationSettingsScreenCoordinator.swift
nimau: Oh ok, the `NotificationSettingsScreen` is now opened modally by the `RoomFlowCoordinator`

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/NotificationSettingsScreenCoordinator.swift
nimau: done.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/NotificationSettingsScreenCoordinator.swift
nimau: 👍

ElementX/Sources/Screens/RoomNotificationSettingsScreen/View/RoomNotificationSettingsScreen.swift
nimau: On this point, I understand the logic, it's just that I'm not sure what this link should look like.

nimau: Ok, done, thanks for the help :)

ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenCoordinator.swift
```diff
@@ -29,19 +29,24 @@ struct RoomDetailsScreenCoordinatorParameters {
...
     private var navigationStackCoordinator: NavigationStackCoordinator? {
         parameters.navigationStackCoordinator
     }
     
-    var callback: ((RoomDetailsScreenCoordinatorAction) -> Void)?
-    
+    private let actionsSubject: PassthroughSubject<RoomDetailsScreenCoordinatorAction, Never> = .init()
+    private var cancellables: Set<AnyCancellable> = .init()
+        
+    var actions: AnyPublisher<RoomDetailsScreenCoordinatorAction, Never> {
```
pixlwave, line 46: Oh nice, thanks for adding it here too 👍

pixlwave (approved):
Perfect, looks great to me! 😎

PR #1449, opened 2023-08-04
Integrate emojibase

langleyd:
### Context
At the moment web, android and iOS all use different datasources for their emoji picker. This means inconsistencies in what emojis are available(the iOS one is a bit out of date so is missing the new emojis), order they are displayed, different tags etc.

It can also result in bugs like this where the data sent by each client may vary(incorrect unicode points):
image

https://github.com/milesj/emojibase that web uses seems to make most sense, it provides just the json and is spec compliant(generated directly from http://unicode.org/Public/emoji/). It also has translations which we could take advantage of for tags, so that the picker search works for people in non-english languages.

https://github.com/matrix-org/emojibase-bindings is a new repo that will provide bindings for the 3 platforms.

We now have moose!


ElementX/Sources/Services/Emojis/EmojiProvider.swift
```diff
@@ -74,3 +75,26 @@ class EmojiProvider: EmojiProviderProtocol {
...
+
+extension EmojibaseDatasource: EmojiLoaderProtocol {
+    func load() async -> [EmojiCategory] {
+        do {
+            let store: EmojibaseStore = try await load()
+            return EmojibaseCategory.allCases.map { category in
+                let emojis = store.categories[category.rawValue] ?? []
+                return EmojiCategory(id: category.rawValue, emojis: emojis.compactMap(EmojiItem.init(from:)))
+            }
+        } catch { }
```
stefanceriu, line 87: We should probably do something with these errors, at least log them if nothing else

stefanceriu (approved):
Looks good to me, very nice 🤩

stefanceriu:
> searching `wizard` on web doesn't return 🧙, instead you need to search for `mage` which is totally undiscoverable).

I think that's more of a problem with the current implementation than with the data itself. If you look at the generated json you see this

```

group: 1,
hexcode: "1F9D9",
label: "mage",
order: 1723,
tags: [
"sorcerer",
"sorceress",
"witch",
"wizard"
],
unicode: "🧙",
```

pixlwave:
Ok nice, then bring on the Moose!

pixlwave (approved):
🚢 it!

edit: Although the EmojiProviderTests are failing.

PR #1456, opened 2023-08-07
Prevent xcodegen from creating temporary references in the project file

stefanceriu:
i.e. things like this

Screenshot 2023-08-07 at 10 41 02


alfogrillo:
> NB I haven't made any changes to the post checkout hook as it still applies to secrets.xcconfig

I think we need to update the script as well, otherwise I see changes in the working area.

```
git update-index --assume-unchanged ElementX/SupportingFiles/secrets.xcconfig
```

Works for me

stefanceriu:
> I think we need to update the script as well, otherwise I see changes in the working area.

Strange, I swear I tested it locally and it still seemed to work, doesn't anymore 🤔 Anyways, I'll update the path.

alfogrillo (approved):
LGTM 👍🏻

PR #1458, opened 2023-08-08
Timeline item long press tweaks

### LINKED ISSUE #1455: Long-press on a bubble (if you miss slightly?) can end up selecting the whole timeline
### Steps to reproduce

I've had this at least 5 times - 3 of which were in the last 24 hours.  I also thought i'd submitted it at least twice already (once a few days ago!), but i can't find any evidence of it, so sorry if this is a dup:

1. Long-press on a bubble to pull up its context menu (in this instance to edit to fix a typo)
2. Miss slightly (unsure how; presumably a few pixels above/below the bubble)
3. Entire timeline ends up selected and greyed-out/depressed instead, looking like this:

IMG_1564

You then have to either force quit the app to get out of the state it's in, or possibly bg, fg, then vertical swipe.

#### What did you expect?

Not to accidentally select the whole timeline and get stuck in limbo

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

323

### Homeserver

_No response_

### Will you send logs?

No

stefanceriu:
* reduces the time it takes for the timeline item long press gesture to trigger
* correctly identifies timeline items based on their stable timelineID so that actions are always correctly computed (fixes #1455)
* tweaks swiftlint to remove various warnings

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -24,7 +24,7 @@ line_length:
   error: 1000
 
 file_length:
-  warning: 800
+  warning: 1000
   error: 1000
 
 type_name:
@@ -38,8 +38,11 @@ type_body_length:
   error: 1000
 
 function_body_length:
-  warning: 50
+  warning: 100
   error: 100
+
+cyclomatic_complexity:
+  ignores_case_statements: true
 
 nesting:
   type_level:
--- a/ElementX/Sources/Screens/RoomScreen/View/Style/LongPressWithFeedback.swift
+++ b/ElementX/Sources/Screens/RoomScreen/View/Style/LongPressWithFeedback.swift
@@ -33,7 +33,7 @@ struct LongPressWithFeedback: ViewModifier {
             .animation(isLongPressing ? .spring(response: 0.5).delay(0.1) : .spring(response: 0.5),
                        value: isLongPressing)
             // The minimum duration here doesn't actually invoke the perform block when elapsed (thus
-            // the implementation bellow) but it does cancel other system gestures e.g. swipe to reply
+            // the implementation below) but it does cancel other system gestures e.g. swipe to reply
             .onLongPressGesture(minimumDuration: 0.25) { } onPressingChanged: { isPressing in
                 isLongPressing = isPressing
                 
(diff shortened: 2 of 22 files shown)


.swiftlint.yml
```diff
@@ -24,8 +24,8 @@ line_length:
   error: 1000
 
 file_length:
-  warning: 800
-  error: 1000
+  warning: 1000
+  error: 1500
```
pixlwave, line 28: Should we just 1000, 1000 these? I think its reasonable to error at 1000 lines.

.swiftlint.yml
```diff
@@ -38,8 +38,11 @@ type_body_length:
   error: 1000
 
 function_body_length:
-  warning: 50
-  error: 100
+  warning: 100
+  error: 150
```
pixlwave, line 42: Haven't thought about this one as much but potentially the same. As we don't want warnings we should just set error to the same value imo.

.swiftlint.yml
```diff
@@ -38,8 +38,11 @@ type_body_length:
...
+
+cyclomatic_complexity:
+  ignores_case_statements: true
```
pixlwave, line 45: Ohhhhh! Nice 👏

ElementX/Sources/Screens/RoomScreen/View/Style/LongPressWithFeedback.swift
```diff
@@ -29,15 +30,27 @@ struct LongPressWithFeedback: ViewModifier {
...
                          y: isLongPressing ? 1.05 : 1)
-            .animation(isLongPressing ? .spring(response: 1.5).delay(0.2) : .spring(response: 0.5),
+            .animation(isLongPressing ? .spring(response: 0.5).delay(0.1) : .spring(response: 0.5),
                        value: isLongPressing)
-            .onLongPressGesture(minimumDuration: 0.25) {
-                action()
-                feedbackGenerator.impactOccurred()
-            } onPressingChanged: { isPressing in
+            // The minimum duration here doesn't actually invoke the perform block when elapsed (thus
+            // the implementation bellow) but it does cancel other system gestures e.g. swipe to reply
```
pixlwave, line 36:
```suggestion
            // the implementation below) but it does cancel other system gestures e.g. swipe to reply
```

pixlwave (approved):
Nice 👍

.swiftlint.yml
stefanceriu: Fair 👍

PR #1462, opened 2023-08-08
Re-enable and (hopefuly) simplify background app refresh handling

stefanceriu:
This was previously disabled by rust not being able to recover correctly after stopping a sync and processing the stop signal on app foregrounding. 
That has been since fixed and the original implementation should work but I also wanted to simplify it a bit.


pixlwave (approved):
Ok, lets see I guess 🤷‍♂️

Issue #1029, opened 2023-06-06
Edit the default notification settings for the Direct chats


giomfo:
## Description ##
Let the user configure the default setting for direct chats. Use the 2 Underride rules: `.m.rule.encrypted_room_one_to_one` and `.m.rule.room_one_to_one` to handle this configuration

## Suggested Design ##
Replace here in the design "Group" with "Direct"
<img width="184" alt="image" src="https://github.com/vector-im/element-x-ios/assets/8969772/b1e965d7-1423-4d7d-8838-9d1bde917320">

giomfo:
Done (behind the Developer Option "Show notification settings")

Issue #1026, opened 2023-06-06
Edit the default notification settings for the Group chats


giomfo:
## Description ##
Let the user configure the default setting for group chats. Use the 2 Underride rules: `.m.rule.message` and `.m.rule.encrypted` to handle this configuration

## Suggested Design ##
<img width="283" alt="image" src="https://github.com/vector-im/element-x-ios/assets/8969772/88f8ac68-7cb0-4632-852a-47f23fa31292">

giomfo:
Done (behind the Developer Option "Show notification settings")

PR #1468, opened 2023-08-09
Edit the default notification settings

nimau:
This PR implements #1026 and #1029 to select the default notification mode for group chats and direct chats.

### DIFF
--- a/ElementX/Sources/Screens/Settings/NotificationSettingsEditScreen/NotificationSettingsEditScreenViewModel.swift
+++ b/ElementX/Sources/Screens/Settings/NotificationSettingsEditScreen/NotificationSettingsEditScreenViewModel.swift
@@ -71,11 +71,12 @@ class NotificationSettingsEditScreenViewModel: NotificationSettingsEditScreenVie
     
     private func fetchSettings() {
         fetchSettingsTask = Task {
-            let mode: RoomNotificationModeProxy
+            var mode: RoomNotificationModeProxy?
             let encrypted_mode = await notificationSettingsProxy.getDefaultRoomNotificationMode(isEncrypted: true, isOneToOne: isDirect)
             let unencrypted_mode = await notificationSettingsProxy.getDefaultRoomNotificationMode(isEncrypted: false, isOneToOne: isDirect)
-            mode = encrypted_mode != unencrypted_mode ? .allMessages : encrypted_mode
-            
+            if encrypted_mode == unencrypted_mode {
+                mode = encrypted_mode
+            }
             guard !Task.isCancelled else { return }
             
             switch mode {
@@ -90,7 +91,7 @@ class NotificationSettingsEditScreenViewModel: NotificationSettingsEditScreenVie
     }
     
     private func setMode(_ mode: NotificationSettingsEditScreenDefaultMode) {
-        guard state.pendingMode == nil else { return }
+        guard state.pendingMode == nil, !state.isSelected(mode: mode) else { return }
         let roomNotificationModeProxy: RoomNotificationModeProxy
         switch mode {
         case .allMessages:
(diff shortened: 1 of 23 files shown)


ElementX/Sources/Services/NotificationSettings/NotificationSettingsProxy.swift
```diff
@@ -55,23 +55,24 @@ final class NotificationSettingsProxy: NotificationSettingsProxyProtocol {
...
         let roomNotificationMode = await notificationSettings.getDefaultRoomNotificationMode(isEncrypted: isEncrypted, activeMembersCount: activeMembersCount)
         return RoomNotificationModeProxy.from(roomNotificationMode: roomNotificationMode)
     }
+
+    func setDefaultRoomNotificationMode(isEncrypted: Bool, isOneToOne: Bool, mode: RoomNotificationModeProxy) async throws {
+        let backgroundTask = await backgroundTaskService?.startBackgroundTask(withName: "setDefaultRoomNotificationMode")
+        defer { backgroundTask?.stop() }
+        
+        try await notificationSettings.setDefaultRoomNotificationMode(isEncrypted: isEncrypted, isOneToOne: isOneToOne, mode: mode.roomNotificationMode)
+        await updatedSettings()
```
stefanceriu, line 74: What does this actually do? 🤔

ElementX/Sources/Screens/Settings/NotificationSettingsEditScreen/NotificationSettingsEditScreenViewModel.swift
```diff
@@ -0,0 +1,112 @@
...
+        case .allMessages:
+            roomNotificationModeProxy = .allMessages
+        case .mentionsAndKeywordsOnly:
+            roomNotificationModeProxy = .mentionsAndKeywordsOnly
+        }
+        state.pendingMode = mode
+        Task {
+            do {
+                try await notificationSettingsProxy.setDefaultRoomNotificationMode(isEncrypted: true, isOneToOne: isDirect, mode: roomNotificationModeProxy)
+                try await notificationSettingsProxy.setDefaultRoomNotificationMode(isEncrypted: false, isOneToOne: isDirect, mode: roomNotificationModeProxy)
```
stefanceriu, line 105: Can we run into weird issues here if only the second one fails?

stefanceriu (approved):
Left some comments inline but otherwise looks good to me 👍

ElementX/Sources/Services/NotificationSettings/NotificationSettingsProxy.swift
```diff
@@ -55,23 +55,24 @@ final class NotificationSettingsProxy: NotificationSettingsProxyProtocol {
...
-        }
-        try await notificationSettings.setRoomNotificationMode(roomId: roomId, mode: roomNotificationMode)
+        try await notificationSettings.setRoomNotificationMode(roomId: roomId, mode: mode.roomNotificationMode)
         await updatedSettings()
     }
     
-    func getDefaultNotificationRoomMode(isEncrypted: Bool, activeMembersCount: UInt64) async -> RoomNotificationModeProxy {
+    func getDefaultRoomNotificationMode(isEncrypted: Bool, isOneToOne: Bool) async -> RoomNotificationModeProxy {
+        // A one-to-one room is a room with exactly two active members
+        let activeMembersCount: UInt64 = isOneToOne ? 2 : 3
```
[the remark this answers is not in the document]
nimau: @stefanceriu From the point of view of push rules, a `one-to-one` room is a room with exactly 2 members because of the following condition: `{ "kind": "room_member_count", "is": "2" }` for push rules `.m.rule.room_one_to_one` and `.m.rule.encrypted_room_one_to_one`.
Here, this function retrieves the default mode for a given room type (`group chat` or `direct chat`), but because there are 4 different push rules (`.m.rule.room_one_to_one`, `.m.rule.encrypted_room_one_to_one`, `.m.rule.message` and `.m.rule.encrypted`) we need to pass the encryption status and the room members count.

When setting the default notification mode for `direct chats` or `group chats`, we don't have an actual room, so we need to pass a fake `activeMembersCount` which will correspond to the correct push rule (here `2` for one-to-one and `3` for group chats). I'll update the RustSDK to avoid this by passing a boolean parameter `is_one_to_one` instead of the number of active members.

In addition, the parameter is named `isOneToOne` instead of `isDM` to avoid confusion with what we call a DM from the application's point of view.

ElementX/Sources/Services/NotificationSettings/NotificationSettingsProxy.swift
nimau: It sets the room's default notification mode (`all messages`, `mentions and keywords only`) for a room type (`direct chat` or `group chat`)
`await updatedSettings()` waits for the next `.settingsDidChange` callback (triggered on the RustSDK side when a `PushRuleEvent` is received)

ElementX/Sources/Services/NotificationSettings/NotificationSettingsProxy.swift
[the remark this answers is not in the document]
nimau: PR created: https://github.com/matrix-org/matrix-rust-sdk/pull/2401 to stop passing the active members count to the notification settings ffi layer.

ElementX/Sources/Screens/Settings/NotificationSettingsEditScreen/NotificationSettingsEditScreenViewModel.swift
nimau: Ok, I've added an retry button on the alert:


The banner on the previous screen will be implemented on a later PR.

ElementX/Sources/Services/NotificationSettings/NotificationSettingsProxy.swift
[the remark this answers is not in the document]
nimau: The RustSDK has been updated so that the `activeMembersCount` parameter is no longer required.

PR #1471, opened 2023-08-10
Fixes #1140 - Dismiss all modals when presenting a room

### LINKED ISSUE #1140: push notifs can break view stack
### Steps to reproduce

1. start inviting someone to a room
2. push notif comes in
3. accidentally tap the notif
4. room in the background invisibly changes to show the push notif
5. finish inviting the user
6. hit invite button
7. nothing happens
8. cancel out and discover you’re in the room from the push notif

#### What did you expect?

opening a push should cancel other UI open in the app, not break the UI.

#### What happened instead?

the room view changed to the push but UI on too didn’t

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

244

### Homeserver

_No response_

### Will you send logs?

No

stefanceriu:


ElementX/Sources/Screens/FilePreviewScreen/InteractiveQuickLook.swift
```diff
@@ -28,21 +29,27 @@ extension View {
...
+    @State private var dismissalPublisher = PassthroughSubject<Void, Never>()
+    
     func body(content: Content) -> some View {
         content.background {
             if let item {
-                MediaPreviewViewController(previewItem: item) { self.item = nil }
+                MediaPreviewViewController(previewItem: item, dismissalPublisher: dismissalPublisher) { self.item = nil }
+            } else {
+                // Work around QLPreviewController dismissal issues, see below.
+                let _ = dismissalPublisher.send(())
```
pixlwave, line 40: How come you went with this in the end instead of the `onChange`?

ElementX/Sources/Screens/FilePreviewScreen/InteractiveQuickLook.swift
```diff
@@ -70,7 +88,7 @@ private struct MediaPreviewViewController: UIViewControllerRepresentable {
         }
         
         override func viewDidLoad() {
-            view.backgroundColor = .clear
+            view.backgroundColor = .red
```
pixlwave, line 91: Should probably revert this change 😉

ElementX/Sources/Screens/FilePreviewScreen/InteractiveQuickLook.swift
```diff
@@ -85,7 +103,7 @@ private struct MediaPreviewViewController: UIViewControllerRepresentable {
             
             self.previewController = previewController
         }
-        
+
```
pixlwave, line 106: And this whitespace

ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenCoordinator.swift
```diff
@@ -79,6 +79,10 @@ final class RoomDetailsScreenCoordinator: CoordinatorProtocol {
         }
     }
     
+    func stop() {
+        viewModel.stop()
```
pixlwave, line 83: Oh nice, good to see this is called in time.

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -259,6 +259,10 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
     }
     
     private func asyncPresentRoom(_ roomID: String, animated: Bool, destinationRoomProxy: RoomProxyProtocol? = nil) async {
+        // If any sheets are presented dismiss them, rely on their dismissal callbacks to transition the state machine
+        // through the correct states before presenting the room
+        navigationStackCoordinator.setSheetCoordinator(nil)
```
pixlwave, line 264: Maybe a dumb question but I feel like this might be more correctly handled in `clearRoute()` instead of here?

pixlwave (approved):

ElementX/Sources/Screens/FilePreviewScreen/InteractiveQuickLook.swift
stefanceriu: I tried that and it didn't work at all for some reason ¯\_(ツ)_/¯

ElementX/Sources/Screens/FilePreviewScreen/InteractiveQuickLook.swift
stefanceriu: Oupsie 🙈

ElementX/Sources/Screens/FilePreviewScreen/InteractiveQuickLook.swift
stefanceriu: Oupsie 🙈

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
stefanceriu: That's a very good point, didn't even occur to me. Will try it out to see if works

stefanceriu: It does actually make things a bit more complicated as `presentRoom` might just want to popToRoot instead of to the roomList. Doing it separately gives us more control so I guess I'll leave it alone (for now at least)

PR #1472, opened 2023-08-10
Fix reply to media crash caused by missing room context (and implicit…

stefanceriu:
…ly image provider) on the new composer toolbar


pixlwave (approved):
LGTM 👍

PR #1475, opened 2023-08-10
Disable timeline in the middle test again.

pixlwave:
After running happily for a few days, it has been failing for the last week, hiding signs of other tests failing.

Additionally fixes the room details avatar changing from an image to a button.

Run going on here: https://github.com/vector-im/element-x-ios/actions/runs/5823415696


alfogrillo (approved):
LGTM!

PR #1480, opened 2023-08-11
Update SDK to 1.0.106-alpha

nimau:
This PR updates the RustSDK to 1.0.106-alpha

### DIFF
--- a/ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenViewModel.swift
+++ b/ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenViewModel.swift
@@ -195,7 +195,7 @@ class RoomDetailsScreenViewModel: RoomDetailsScreenViewModelType, RoomDetailsScr
         do {
             let notificationMode = try await notificationSettingsProxy.getNotificationSettings(roomId: roomProxy.id,
                                                                                                isEncrypted: roomProxy.isEncrypted,
-                                                                                               activeMembersCount: UInt64(roomProxy.activeMembersCount))
+                                                                                               isOneToOne: roomProxy.activeMembersCount == 2)
             state.notificationSettingsState = .loaded(settings: notificationMode)
         } catch {
             state.notificationSettingsState = .error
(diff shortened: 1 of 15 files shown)


ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenViewModel.swift
```diff
@@ -195,7 +195,7 @@ class RoomDetailsScreenViewModel: RoomDetailsScreenViewModelType, RoomDetailsScr
         do {
             let notificationMode = try await notificationSettingsProxy.getNotificationSettings(roomId: roomProxy.id,
                                                                                                isEncrypted: roomProxy.isEncrypted,
-                                                                                               activeMembersCount: UInt64(roomProxy.activeMembersCount))
+                                                                                               isOneToOne: roomProxy.activeMembersCount == 2)
```
alfogrillo, line 198: I'm not sure this is correct.
If you want to check if a room **is direct** and it has two people in it we need something different.

We already have in the codebase:
```swift
extension RoomProxyProtocol {
    var isEncryptedOneToOneRoom: Bool {
        isDirect && isEncrypted && activeMembersCount <= 2
    }
}
```

maybe we also need:
```swift
extension RoomProxyProtocol {
    var isOneToOneRoom: Bool {
        isDirect && activeMembersCount <= 2
    }
}
```
?

PS: We opted for `activeMembersCount <= 2` and not `activeMembersCount == 2` because the other person may leave after a while and we still want to show the room as a direct one in the UI. 
Not sure it is helpful in your case.

alfogrillo: (the comment is on the logic `activeMembersCount == 2`)

nimau: From the point of view of push rules, a one-to-one room is a room with exactly 2 members because of the following condition: `{ "kind": "room_member_count", "is": "2" }` for push rules `.m.rule.room_one_to_one` and `.m.rule.encrypted_room_one_to_one`

alfogrillo: Right, thanks for the clarification

alfogrillo (approved):
LGTM!

Issue #1372, opened 2023-07-20
Location share jumps when reaction is added to it from another session


kittykat:
### Steps to reproduce

1. Share static location
2. It's easier to reproduce this issue if you then send other messages in the room so that the static location is not on the most recent screen
3. Someone (or yourself) reacts to the static location share from another session

#### What did you expect?
Reaction appears smoothly

#### What happened instead?
Things jump in the timeline, see video

https://github.com/vector-im/element-x-ios/assets/51663/4daf6d00-50ac-49d3-8c92-47d03a1e3f90

### Your phone model

iPhone 11

### Operating system version

16.5.1

### Application version

1.2.1 (294)

### Homeserver

matrix.org

### Will you send logs?

No

langleyd:
I think the problem is more how the the timeline/location sharing interact/animate when something causes the size of the cell to change(in this case reactions).

I think there are two parts tot he jankyness:
1. The size of the location view changes and animates when an emjoi is added, but doesn't happen when it's removed. The animation looks a bit strange, the container seems to jump to a larger size and then the location view animates bigger.
2. When the size changes a different url is requested from api.maptiler.com and for a brief moment the location share displays in the error state(a cancellation error is returned).

These two can be seen in this screenshot from the video above.
<img width="300" alt="Screenshot 2023-08-10 at 22 00 21" src="https://github.com/vector-im/element-x-ios/assets/6135282/e65e5002-2bec-4fd1-85f0-118c24bc89b1">

@alfogrillo Maybe it would be worth figuring out the max image size you will need, request that and then just resize locally on the device to avoid the reloading of the url on size changes?

Also might look less janky for now if we just remove the animations?

alfogrillo:
@langleyd @kittykat 
This is a more general problem affecting also other kind of bubbles (see the video).
However I managed reduce it a bit by avoiding the reload of the map when the size changes ([PR](https://github.com/vector-im/element-x-ios/pull/1481)).
This isn't ideal but it seems to work.

https://github.com/vector-im/element-x-ios/assets/19324622/c2bc6219-33a3-4332-afa2-2d6d94acb9a9

langleyd:
As @alfogrillo mentions @kittykat the real source of this problem(and most of the timeline animation glitches) is the timeline implementation. We are moving to the SwiftUI based timeline in the coming month or two which will address this problem, and most of the animation issues that occur due to a conflict of UIKit + SwiftUI.

PR #1481, opened 2023-08-11
Fix refresh of location events

alfogrillo:
Fix https://github.com/vector-im/element-x-ios/issues/1372
ATM bubbles in the timeline change size (in general, also for other event types) on the UITableView based timeline.
This cause a problems on the location view since the url we fetch is related with the bubble size.
This PR is a workaround for the problem until we switch to the SwiftUI timeline where the problem seems not happening.


stefanceriu (approved):
👍

Issue #1470, opened 2023-08-09
Add reaction button overlaps with the next location message bubble


amshakal:
![image](https://github.com/vector-im/element-x-ios/assets/85162587/de7aa7c4-6459-4c8d-851a-b0b15c5e1e76)

It should just behave like any other message bubble with reactions

PR #1482, opened 2023-08-11
Fix reaction height issue

langleyd:
Fixes https://github.com/vector-im/element-x-ios/issues/1470

- Fixes reaction height in issue above. The layout makes all the items have the same height. We had this logic in the placing of the views but didn't account for it in the overall size calculation.
- Adds a regression test fot it.

Also fixes fixes reaction summary button color for text based reactions:
image
->
image


ElementX/Sources/Other/CollapsibleFlowLayout/CollapsibleReactionLayout.swift
```diff
@@ -141,13 +141,21 @@ struct CollapsibleReactionLayout: Layout {
...
     private func sizeThatFits(proposal: ProposedViewSize, rows: [[FlowLayoutSubview]]) -> CGSize {
-        rows.enumerated().reduce(CGSize.zero) { partialResult, rowItem in
+        let sizes = rows.map { row in
+            row.map { subview in
+                subview.sizeThatFits(.unspecified)
+            }
+        }
+        let maxHeight = sizes.joined().reduce(0) { partialResult, size in
+            max(partialResult, size.height)
+        }
```
alfogrillo, line 151:
```suggestion
     let maxHeight = sizes.joined().max(by: { $0.height < $1.height })?.height ?? 0
```

alfogrillo (approved):
LGTM

ElementX/Sources/Other/CollapsibleFlowLayout/CollapsibleReactionLayout.swift
langleyd: To be honest I find `max(by:)` more opaque to read vs. using a reduce here, as you have to go read the definition to know that the closure indicates a check on `areInIncreasingOrder`.

PR #1483, opened 2023-08-11
Adopt fuzzy searching through the RoomListService's dynamic filter

stefanceriu:


ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
```diff
@@ -80,7 +80,11 @@ struct HomeScreen: View {
...
             }
             .onChange(of: context.searchQuery) { searchQuery in
-                guard searchQuery.isEmpty else { return }
+                context.send(viewAction: .updatedSearchQuery)
+                
+                guard searchQuery.isEmpty else {
+                    return
+                }
                 // Dispatch allows the view to update after changing the query
                 DispatchQueue.main.async { updateVisibleRange() }
```
pixlwave, line 89: Is this line still needed?

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -113,6 +110,15 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
         }
     }
     
+    func updateFilterPattern(_ pattern: String?) {
+        guard let pattern, !pattern.isEmpty else {
+            _ = listUpdatesSubscriptionResult?.dynamicFilter.set(kind: .all)
+            return
+        }
+        
+        _ = listUpdatesSubscriptionResult?.dynamicFilter.set(kind: .fuzzyMatchRoomName(pattern: pattern.lowercased()))
```
pixlwave, line 119: Not part of this specifically, but it would be nice if the `lowercases()` happened on the Rust side instead of client side.

pixlwave (approved):
Oh this is going to be nice! 🔎

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
stefanceriu: @Hywan This is a workaround as I noticed the rust side doesn't take care of it. Do you think we can move it?

ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
stefanceriu: You're right, it's not. Ignoring view range updates while a search query is used is already happening in the viewModel's visibleItemRangePublisher

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
Hywan: The fuzzy match uses smart case: if everything is in lower-case, the case is ignored, otherwise if at least one character is in uppercase, the case isn't ignored. So I would suggest to **not** put the search pattern in lowercase here.

ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
Hywan: This line hasn't been deleted. Is it expected?

PR #1484, opened 2023-08-11
Use Compound ListRow instead of Compound form styles.

pixlwave:
This PR adopts the new `ListRow` component replacing `.xyzStyle(.compoundForm)` in all the screens it was used. There are still remaining screens using the older EX form styles which I will look at in a separate PR.

Additionally there are refinements to make to the component (as noted in the snapshots commit), but this gets everything merged to try and avoid any conflicts with other branches.


UITests/Sources/__Snapshots__/Application/en-GB-iPad-9th-generation.templateScreen.png
Velin92, line 1: what is this screen about?

Velin92 (approved):
LGTM

PR #1490, opened 2023-08-14
Enable fuzzy search behind a feature flag.

pixlwave:
Re-enables #1483 but with a feature flag (defaulted to on) to disable it if issues are encountered.

### DIFF
--- a/ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
+++ b/ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
@@ -22,12 +22,14 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
     private let roomListService: RoomListServiceProtocol
     private let eventStringBuilder: RoomEventStringBuilder
     private let name: String
+    private var appSettings: AppSettings
     
     private let serialDispatchQueue: DispatchQueue
     
     private var roomList: RoomListProtocol?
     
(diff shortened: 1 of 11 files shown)


ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -113,6 +113,17 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
...
     
+    func updateFilterPattern(_ pattern: String?) {
+        guard let pattern, !pattern.isEmpty else {
+            _ = listUpdatesSubscriptionResult?.dynamicFilter.set(kind: .all)
+            return
+        }
+        
+        guard appSettings.fuzzySearchEnabled else { return }
+        
+        _ = listUpdatesSubscriptionResult?.dynamicFilter.set(kind: .fuzzyMatchRoomName(pattern: pattern.lowercased()))
```
Hywan, line 124: Don't lower the case please, the Rust SDK has a smart case approach.

stefanceriu: The search box has autocapitalization enabled, which text fields tend to do on iOS. Not lowercasing it will be mega annoying e.g. searching for Kt, won't find kittykat

stefanceriu (approved):
Awesome, thank you! 🚀

PR #1497, opened 2023-08-16
Speed up the room event string builder, bold sender names directly in…

stefanceriu:
…stead of parsing markdown


pixlwave (approved):
Didn't you want to remove the range on the string inside `bold()` too?

stefanceriu:
Talked about it offline: we prefer leaving the range instead of adding a swiftlint exception

PR #1499, opened 2023-08-16
Subscription fix

stefanceriu:
The `InvitesScreenViewModel` fetches inviters through RoomProxies which has the side effect of adding subscriptions for all of them. This PR move subscriptions to a different call in order to prevent this from happening in the future

### DIFF
--- a/ElementX.xcodeproj/xcshareddata/xcschemes/ElementX.xcscheme
+++ b/ElementX.xcodeproj/xcshareddata/xcschemes/ElementX.xcscheme
@@ -89,7 +89,7 @@
          </EnvironmentVariable>
          <EnvironmentVariable
             key = "HTTPS_PROXY"
-            value = "192.168.0.111:9090"
+            value = "localhost:9090"
             isEnabled = "NO">
          </EnvironmentVariable>
          <EnvironmentVariable
--- a/ElementX/SupportingFiles/target.yml
+++ b/ElementX/SupportingFiles/target.yml
@@ -23,7 +23,7 @@ schemes:
           value: full
           isEnabled: false
         - variable: HTTPS_PROXY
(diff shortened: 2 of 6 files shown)


ElementX.xcodeproj/xcshareddata/xcschemes/ElementX.xcscheme
```diff
@@ -89,7 +89,7 @@
          </EnvironmentVariable>
          <EnvironmentVariable
             key = "HTTPS_PROXY"
-            value = "192.168.0.111:9090"
+            value = "localhost:9090"
```
Velin92, line 92: what is this?

ElementX/SupportingFiles/target.yml
```diff
@@ -23,7 +23,7 @@ schemes:
           value: full
           isEnabled: false
         - variable: HTTPS_PROXY
-          value: 192.168.0.111:9090
+          value: localhost:9090
```
Velin92, line 26: same here

Velin92 (approved):
LGTM

ElementX.xcodeproj/xcshareddata/xcschemes/ElementX.xcscheme
stefanceriu: No no, this is a placeholder for when you want to get the app to go through a http proxy. Previously we used to have a hardcoded IP in there but localhost works just fine

pixlwave: Woooo finally!

stefanceriu: 😅

PR #1503, opened 2023-08-16
Prevent local echoes from being stuck behind read receipt and reactio…

stefanceriu:
…n sending


stefanceriu:
> am wondering if the send read receipt shouldn't also be dispatched in the low priority/utility queue

Good point, as it's not really something the user initiated or is waiting on

PR #1507, opened 2023-08-17
NSE uses UserAgentBuilder

### LINKED ISSUE #1504: Element X sometime presenting as 'matrix-rust-sdk'
### Steps to reproduce

From the perspective of Synapse, Element X sometimes presents itself correctly as 'Element X/...' and other times as the rust-sdk default 'matrix-rust-sdk'.

I'm not sure which contexts lead to which outcome but it is common enough that for a given user like me there will be examples of 'matrix-rust-sdk' daily.

#### What did you expect?
I expect Element X to correctly set the user agent so that we can determine the client type server side.

#### What happened instead?
The user agent is sometimes 'matrix-rust-sdk'

@jplatte suggests that this could be due to EX not overriding the user_agent properly when constructing an SDK client object.

### Your phone model

n/a

### Operating system version

n/a

### Application version

n/a

### Homeserver

matrix.org

### Will you send logs?

No

Velin92:
fixes #1504


NSE/Sources/Other/NSEUserSession.swift
```diff
@@ -30,6 +30,7 @@ final class NSEUserSession {
         baseClient = try ClientBuilder()
             .basePath(path: URL.sessionsBaseDirectory.path)
             .username(username: credentials.userID)
+            .userAgent(userAgent: UserAgentBuilder.makeASCIIUserAgent() ?? "unknown")
```
stefanceriu, line 33: Actually one comment, this ^ is an optional properly, I would leave it as nil so it uses the rust-sdks agent instead of "unknown"

Velin92: I see that this is also the current behaviour we use in the main app UserSessionStore.

Also because the userAgent function takes a String and not a String? as argument type.

stefanceriu: Okay, talked about it offline, decide to move "unknown" to within the UserAgentBuilder so it's all in one place

Issue #1434, opened 2023-08-01
Invite list doesn’t update when sync updates


ara4n:
### Steps to reproduce

1. Receive a push for an invite
2. Tap on it
3. Get taken to the invite list
4. Watch the sync spinner for a while as the all syncs
5. Sync completes
6. The invite isn’t shown
7. Exit list and reenter
8. Now the invite is shown and you can accept it

#### What did you expect?

reliable invite list

### Application version

315

### Will you send logs?

No

PR #1508, opened 2023-08-17
Start using the new (faster) RoomInfo record to populate the room and…

stefanceriu:
… invite lists

This brings performance improvements as we no longer need to go over FFI for each of the room's properties. It also exposes the inviter so that we can directly populate the invites list without having to fetch them separately

Fixes https://github.com/vector-im/element-x-ios/issues/1434


Velin92 (approved):
LGTM

PR #1513, opened 2023-08-18
Fixes #1282 - Switch composer text vield to a @FocusState

### LINKED ISSUE #1282: Room list -> Room animation stutters badly whenever the keyboard is open
### Steps to reproduce

1. Go to room
2. Open keyboard
3. Go back to roomlist
4. Observe the animation judder nastily back and forth as the keyboard gets hidden

Compare the two transitions below:

https://github.com/vector-im/element-x-ios/assets/1294269/a869e900-5ec4-45b1-805d-dad9ac3a3d9b

#### What did you expect?

smooth transitions

#### What happened instead?

juddering janky transition when keyboard is open in a room

can we change the timing of the keyboard hide anim?

Turns out this may have been like this forever, although i only noticed on release day.

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

276

### Homeserver

_No response_

### Will you send logs?

No

stefanceriu:


Velin92 (approved):
LGTM, however when the timeline is popped, the keyboard is still dismissed and then represented if the search bar was focused before. We still want to keep also this behaviour?

stefanceriu:
> he keyboard is still dismissed and then represented if the search bar was focused before

Yes, I think it's fine to keep whatever the default system behavior is

PR #1463, opened 2023-08-08
Fixes #1418 - Preserve new lines when parsing html strings

stefanceriu:
Trick DTCoreText into preserving `\n`s by replacing them with `<br>` tags before processing.

PR #1521, opened 2023-08-18
`MessageText` component

Velin92:
This is going to replace SwiftUI's Text inside `FormattedBodyText`, since this allows to include custom views as `NSTextAttachments`, into `AttributedString` objects

What's left to fix/adjust:

- [x] Dynamic Type support
- [x] Remove the weird extra margin that appears in quote blocks
- [x] Sometimes when the message contains an AttributedString but the text is actually simple plain text, the font is not in body format.
- [x] Fix a bug where the timestamp additional spacing is ignored
- [x] Boost emoji size not working
- [x] UI tests for the room need to be regenerated (for the human eye they look the same but the Text and TextView have minuscule differences that affect the UI tests making them fail, about 96% precision instead of the required 98%)


ElementX/Sources/Screens/RoomScreen/View/Timeline/FormattedBodyText.swift
```diff
@@ -24,29 +24,37 @@ struct FormattedBodyText: View {
     private let additionalWhitespacesCount: Int
     private let boostEmojiSize: Bool
 
+    private let defaultFontContainer: AttributeContainer = {
+        var container = AttributeContainer()
+        container.font = UIFont.preferredFont(forTextStyle: .body)
```
pixlwave, line 29: Can you add a comment on this that we should use bodyLG from Compound here just for consistency in the future (it doesn't yet exist as a UIFont, so there's no need to do it right now).

ElementX/Sources/Screens/RoomScreen/View/Timeline/FormattedBodyText.swift
```diff
@@ -90,19 +97,19 @@ struct FormattedBodyText: View {
                         }
                         .layoutPriority(TimelineBubbleLayout.Priority.visibleQuote)
                 } else {
-                    Text(component.attributedString)
+                    MessageText(attributedString: component.attributedString)
                         .padding(.horizontal, timelineStyle == .bubbles ? 4 : 0)
                         .fixedSize(horizontal: false, vertical: true)
                         .foregroundColor(.compound.textPrimary)
```
pixlwave, line 103: Am I right in thinking we can remove this `foregroundColor` now as it does nothing?

ElementX/Sources/Screens/RoomScreen/View/Timeline/FormattedBodyText.swift
```diff
@@ -121,12 +128,12 @@ struct FormattedBodyText: View {
...
+                        MessageText(attributedString: component.attributedString)
                             .foregroundColor(.compound.textPrimary)
                     }
                     .fixedSize(horizontal: false, vertical: true)
                 } else {
-                    Text(component.attributedString)
+                    MessageText(attributedString: component.attributedString)
                         .padding(.horizontal, timelineStyle == .bubbles ? 4 : 0)
                         .fixedSize(horizontal: false, vertical: true)
                         .foregroundColor(.compound.textPrimary)
```
pixlwave, line 139: Same here.

ElementX/Sources/Screens/RoomScreen/View/Timeline/FormattedBodyText.swift
pixlwave: Also, should this include the colour of `.compound.textPrimary`? Sadly this one should really result in the snapshots being re-generated as they're currently using the system's primary label colour.

ElementX/Sources/Screens/RoomScreen/View/Timeline/FormattedBodyText.swift
```diff
@@ -121,12 +128,12 @@ struct FormattedBodyText: View {
                         Rectangle()
                             .foregroundColor(Color.red)
                             .frame(width: 4.0)
-                        Text(component.attributedString)
+                        MessageText(attributedString: component.attributedString)
                             .foregroundColor(.compound.textPrimary)
```
pixlwave, line 132: And here.

ElementX/Sources/Other/Pills/PillAttachmentViewProvider.swift
```diff
@@ -0,0 +1,35 @@
...
+    override func loadView() {
+        super.loadView()
+
+        guard textAttachment is PillTextAttachment else {
+            MXLog.failure("[PillAttachmentViewProvider]: attachment is missing or not of expected class")
+            return
+        }
+
+        let view = PillView()
+        let controller = UIHostingController(rootView: view)
```
pixlwave, line 32: Genuine question: should the hosting controller be retained by the provider? Or does its view strongly reference it? I have no idea of the consequences here, just thinking out loud.

ElementX/Sources/Other/InfoPlistReader.swift
```diff
@@ -85,6 +85,16 @@ struct InfoPlistReader {
         infoPlistStringValue(forKey: Keys.mapLibreAPIKey)
     }
 
+    /// Mention Pills UTType
+    var pillsUTType: String {
+        guard let values = bundle.object(forInfoDictionaryKey: "UTExportedTypeDeclarations") as? [[String: Any]],
+              let firstValue = values.first,
+              let utTypeIdentifier = firstValue["UTTypeIdentifier"] as? String else {
```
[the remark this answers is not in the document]
pixlwave: Also, I wonder given this fatalerrors, would it make sense to use `values.first(where:)` to allow for new types to be added before this one in the array.

ElementX/Sources/Other/Pills/MessageText.swift
```diff
@@ -0,0 +1,101 @@
...
+
+    func sizeThatFits(_ proposal: ProposedViewSize, uiView: UITextView, context: Context) -> CGSize? {
+        uiView.sizeThatFits(CGSize(width: proposal.width ?? UIView.layoutFittingExpandedSize.width, height: UIView.layoutFittingCompressedSize.height))
+    }
+
+    func makeCoordinator() -> Coordinator {
+        Coordinator()
+    }
+
+    final class Coordinator: NSObject, UITextViewDelegate {
```
pixlwave, line 57: Do you think we should implement `textView(_:shouldInteractWith:in:interaction:)` too, and use that to forward the URL up to SwiftUI's `openURLAction` from `MessageText`'s environment?

pixlwave (approved):
Very nice 👏

Some minor comments, otherwise looks good to me. I'm a tiny bit curious about what causes the shift in layout of the string within the bubble without changing the size of the bubble, but its not particularly important.

ElementX/Sources/Screens/RoomScreen/View/Timeline/FormattedBodyText.swift
Velin92: the thing is that the color was never set anywhere in the Text and in the AttributedString so... I think they were already using primary label colour.
They should be using the compound one?

Velin92: oh missed that, no problem, should take 15 mins to regenerate them

ElementX/Sources/Other/Pills/PillAttachmentViewProvider.swift
Velin92: No idea, but as we keep developing this thing we will see. So seems to work well as it is.

ElementX/Sources/Other/Pills/MessageText.swift
Velin92: oh yeah definitely! Actually do we use openURLAction in formatted body? (or somewhere else) or is it something for the future?

pixlwave: We definitely don't use it correctly, but its set [here](https://github.com/vector-im/element-x-ios/blob/e1d41ad8a3160e6107699bafd4f8c624d730cb01/ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemView.swift#L34) and [here](https://github.com/vector-im/element-x-ios/blob/e1d41ad8a3160e6107699bafd4f8c624d730cb01/ElementX/Sources/Screens/RoomScreen/View/Timeline/TimelineTableViewController.swift#L178). Also I think that's a bug right? The timeline table view controller is duplicating some of the calls on the view (inc appear/disappear).

pixlwave:
Adding a few things I found during a test (and already shared examples of)
- [ ] Long pressing a message on top of a link shows the context menu and then a link preview jumps in afterwards
- [ ] Long pressing a message shows the context menu but with a little magnified area if you drag around 🔍
- [ ] Lists have a *lot* of spacing between each item. Possibly can revert https://github.com/vector-im/element-x-ios/pull/1463 given EI is using UIKit and didn't need that workaround to behave correctly.
- (Unimportant given you can't change the size with the app visible on iOS) but might highlight something useful: Messages with HTML animate when changing Dynamic Type size

ElementX/Sources/Other/Pills/MessageText.swift
Velin92: yeah we definitely are doing one too many times, I'll keep it in the RoomTimelineItemView

ElementX/Sources/Other/InfoPlistReader.swift
```diff
@@ -42,61 +49,61 @@ struct InfoPlistReader {
...
+              let utType = mentionPills[Keys.utTypeIdentifierKey] as? String else {
+            fatalError("Add properly \(Values.mentionPills) exported type into your target's Info.plst")
         }
-        return utTypeIdentifier
+        return utType
     }
 
-    private func infoPlistStringValue(forKey key: String) -> String {
-        guard let result = bundle.object(forInfoDictionaryKey: key) as? String else {
+    private func infoPlistValue<T>(forKey key: String) -> T {
```
pixlwave, line 105: Nice, I almost asked if we should have made this generic 👏

ElementX/Sources/Other/Pills/MessageText.swift
```diff
@@ -17,11 +17,23 @@
 import SwiftUI
 import UIKit
 
+final class MessageTextView: UITextView {
+    override func addGestureRecognizer(_ gestureRecognizer: UIGestureRecognizer) {
+        // Prevent long press to show the magnifying glass
+        if gestureRecognizer is UILongPressGestureRecognizer {
+            gestureRecognizer.isEnabled = false
```
pixlwave, line 24: Smart 😎

ElementX/Sources/Other/Pills/MessageText.swift
```diff
@@ -42,22 +57,36 @@ struct MessageText: UIViewRepresentable {
...
+            self.openURLAction = openURLAction
+        }
+        
         func textViewDidChangeSelection(_ textView: UITextView) {
             textView.selectedTextRange = nil
         }
+        
+        func textView(_ textView: UITextView, shouldInteractWith URL: URL, in characterRange: NSRange, interaction: UITextItemInteraction) -> Bool {
+            if interaction == .invokeDefaultAction {
+                openURLAction.callAsFunction(URL)
```
pixlwave, line 86: Not for this PR, but I wonder now, maybe we should be reading the return value from `openURLAction` and using that to compute what we return from this delegate method. Then we could have SwiftUI let the text view open the URL in some scenarios.

pixlwave (approved):
👏

PR #1534, opened 2023-08-21
Feature flagged dynamic entries

stefanceriu:
This is all so that we can track down startup performance problems between `roomList.entries` and `roomList.entriesWithDynamicFilter`


pixlwave (approved):
LGTM 👍

Issue #1030, opened 2023-06-06
Edit the custom settings from the Global Notification settings


giomfo:
## Description ##
Following https://github.com/vector-im/element-x-ios/issues/1027
In the global Group chats (or Direct chats) Notification settings:
- Let the user click of one of the listed chats to change the custom or remove it

## Suggested Design ##
<img width="620" alt="image" src="https://github.com/vector-im/element-x-ios/assets/8969772/11f2ce5b-eb55-4b67-804b-d6aed76ef5cc">

giomfo:
Done (behind the Developer Option "Show notification settings")

Issue #1027, opened 2023-06-06
List the chats with a custom notification setting


giomfo:
## Description ##
- In the global Group chats Notification settings screen, list in alphabetic order all the group chats (room with members count != 2) where a custom push rule has been created
- In the global Direct chats Notification settings screen, list in alphabetic order all the direct chats (rooms with 2 members) where a custom push rule has been created
- The edition of these custom settings with be managed by https://github.com/vector-im/element-x-ios/issues/1030

## Suggested Design ##
<img width="287" alt="image" src="https://github.com/vector-im/element-x-ios/assets/8969772/fff01738-7828-4606-a9be-dd1d385d77d3">

PR #1535, opened 2023-08-21
Display the list of rooms with user-defined notification settings

nimau:
The PR implements #1027 and #1030 

From the notification settings screen, when editing the default notification mode for direct chats or a group chats, a list of room with user-defined notification settings is now displayed.

If there is no room with a user-defined mode, the section is not displayed:

If some rooms contain a user-defined mode, they are listed in alphabetical order:

It is possible to change the mode for a specific room and to delete the custom setting:


stefanceriu (approved):
Looks and works great, nicely done! 👏

PR #1538, opened 2023-08-22
Fix for voice over reading the braille dot

Velin92:
We use a braille dot in app to get some extra spacing for the timestamp, however I just discovered that the voice over reads it.

This small fix solves the problem by overriding the accessibility with the original attributed string instead of the adjusted one.


stefanceriu (approved):
Very nice! 🙌

PR #1539, opened 2023-08-22
Fix timestamp not rendering properly when RTL and LTR are mixed together

Velin92:
Screenshot 2023-08-22 at 15 08 50


pixlwave (approved):
👌

PR #1540, opened 2023-08-22
Split incoming large .resets into a reset + append in order to quicke…

stefanceriu:
…r update the UI

### DIFF
--- a/ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
+++ b/ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
@@ -133,24 +133,35 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
         
         MXLog.info("\(name): Received \(diffs.count) diffs, current room list \(rooms.compactMap { $0.id ?? "Empty" })")
         
-        rooms = diffs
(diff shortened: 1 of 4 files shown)


ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -131,26 +131,37 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
...
-                    return currentItems
-                }
-                
-                guard let updatedItems = currentItems.applying(collectionDiff) else {
-                    MXLog.error("\(name): Failed applying diff: \(collectionDiff)")
-                    return currentItems
-                }
-                
-                return updatedItems
+        func processDiff(_ diff: RoomListEntriesUpdate) {
```
Velin92, line 136: can we make this a non-nested private func?

Velin92 (approved):
LGTM

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
stefanceriu: can do 👍

PR #1541, opened 2023-08-22
Enable OIDC support

pixlwave:
This PR updates the Rust SDK and enables OIDC support.

Additionally it adds basic support for displaying the OIDC account URL from the Settings screen and fixes a crash when you aren't signed in.

### DIFF
--- a/ElementX/Sources/Services/Authentication/AuthenticationServiceProxyProtocol.swift
+++ b/ElementX/Sources/Services/Authentication/AuthenticationServiceProxyProtocol.swift
@@ -66,16 +66,15 @@ enum OIDCError: Error {
 }
 
 struct OIDCAuthenticationDataProxy: Equatable {
-//    let underlyingData: OidcAuthenticationData
-//
-//    var url: URL {
-//        URL(string: underlyingData.loginUrl())!
-//    }
-    let url: URL = "https://theroadtonowhere"
+    let underlyingData: OidcAuthenticationData
+    
+    var url: URL {
+        URL(string: underlyingData.loginUrl())!
+    }
 }
 
-// extension OidcAuthenticationData: Equatable {
-//    public static func == (lhs: MatrixRustSDK.OidcAuthenticationData, rhs: MatrixRustSDK.OidcAuthenticationData) -> Bool {
-//        lhs.loginUrl() == rhs.loginUrl()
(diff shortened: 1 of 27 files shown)


NSE/Sources/NotificationServiceExtension.swift
```diff
@@ -40,6 +40,12 @@ class NotificationServiceExtension: UNNotificationServiceExtension {
             // - NotificationID could not be resolved
             return contentHandler(request.content)
         }
+        
+        if credentials.restorationToken.session.oidcData != nil {
+            // Notification content is disabled for OIDC sessions
+            // until token refresh is multi-process aware.
+            return contentHandler(request.content)
+        }
```
pixlwave, line 48: @Velin92 Can you confirm that this is the correct place to do this?

ElementX/Sources/Services/Authentication/AuthenticationServiceProxyProtocol.swift
```diff
@@ -66,16 +66,15 @@ enum OIDCError: Error {
...
-//    let underlyingData: OidcAuthenticationData
-//
-//    var url: URL {
-//        URL(string: underlyingData.loginUrl())!
-//    }
-    let url: URL = "https://theroadtonowhere"
+    let underlyingData: OidcAuthenticationData
+    
+    var url: URL {
+        URL(string: underlyingData.loginUrl())!
```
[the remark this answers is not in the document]
stefanceriu: oh oh 😅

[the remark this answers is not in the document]
pixlwave: 🙈

NSE/Sources/NotificationServiceExtension.swift
Velin92: Do we want to discard the notification or just present it in a generic way?
This should just present it as a generic notification. Which I guess is okay.

pixlwave: This is basically only for element.io use, I'm happier to say that we don't support fetching the content rather than notifications are completely disabled.

stefanceriu (approved):
Looks great, mega exciting! 👏

PR #1549, opened 2023-08-23
Log level settings

stefanceriu:
Expose the log level in the developer options and set it by default to .info.


ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -57,11 +57,11 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
     @Consumable private var storedAppRoute: AppRoute?
 
     init() {
-        MXLog.configure()
-
```
pixlwave, line 61: I don't think there's an issue, but just want to highlight it in case: What happens if setupServiceLocator does anything that calls `MXLog.whatever()`?

ElementX/Sources/Other/Logging/RustTracing.swift
```diff
@@ -59,11 +55,11 @@ struct TracingConfiguration {
     
     static let targets: OrderedDictionary<Target, LogLevel> = [
         .common: .info,
-        .hyper: .warn,
-        .matrix_sdk_crypto: .debug,
-        .matrix_sdk_http_client: .debug,
-        .matrix_sdk_sliding_sync: .debug,
-        .matrix_sdk_base_sliding_sync: .debug,
+        .hyper: .info,
```
pixlwave, line 58: I got the impression that `hyper` was super noisy on info which is why it was on `warn`?

ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/View/DeveloperOptionsScreen.swift
```diff
@@ -22,6 +22,16 @@ struct DeveloperOptionsScreen: View {
...
         Form {
+            Section {
+                Picker("LogLevel", selection: $context.logLevel) {
+                    ForEach(TracingConfiguration.LogLevel.allCases, id: \.self) { logLevel in
+                        Text(logLevel.rawValue.capitalized)
+                    }
+                }
+            } footer: {
+                Text("Requires app reboot")
+            }
```
pixlwave, line 33: You can remove the footer on it by putting the picker inside a 
```swift
LabeledContent {
    Picker …
}  label: {
    Text("Log level")
    Text("Requires app reboot)
}
```

pixlwave (approved):

ElementX/Sources/Other/Logging/RustTracing.swift
stefanceriu: I thought so too but I tested it and not a peep from hyper ¯\_(ツ)_/¯

ElementX/Sources/Application/AppCoordinator.swift
stefanceriu: They won't work. I tried doing a `MXLog.configure(logLevel: .info)` at the very top but configuring the Rust side logging more than once crashes the app (we prevent against that within MXLog.didConfigureOnce)

PR #1551, opened 2023-08-23
Diffing performance fix

stefanceriu:
Fixed the problems we've been seeing with room list hangs
* on the previous `Split incoming large .resets into a reset + append` we introduced a behavioral change which would make the room summary provider inform the UI layer about room changes after every single processed diff
* now we go through the diffs one by one, collecting them, and **if** we encounter a large release we chunk it, process it and then directly inform the UI. The rest of the diffs will continue being collected and the UI will be informed once more at the end

### DIFF
--- a/ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
+++ b/ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
@@ -133,24 +133,35 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
         
         MXLog.verbose("\(name): Received \(diffs.count) diffs, current room list \(rooms.compactMap { $0.id ?? "Empty" })")
         
-        rooms = diffs
-            .reduce(rooms) { currentItems, diff in
-                guard let collectionDiff = buildDiff(from: diff, on: currentItems) else {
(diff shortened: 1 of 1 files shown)


ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -133,25 +133,42 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
...
-                }
-                
-                return updatedItems
+        var updatedItems = rooms
+        for diff in diffs {
+            let resetDiffChunkingThreshhold = 50
+            if case .reset(let values) = diff, values.count > resetDiffChunkingThreshhold {
+                // Special case resets in order to prevent large updates from blocking the UI
+                // Render the first resetDiffChunkingThreshhold as a reset and then append the rest to give the UI time to update
+                updatedItems = processDiff(RoomListEntriesUpdate.reset(values: Array(values[..<resetDiffChunkingThreshhold])), on: updatedItems)
```
pixlwave, line 142: Could we used the inferred type so there's less to visually parse when reading these manual diffs?
```suggestion
                updatedItems = processDiff(.reset(values: Array(values[..<resetDiffChunkingThreshhold])), on: updatedItems)
```

pixlwave (approved):

PR #1553, opened 2023-08-24
Feature flag to OTLP tracing

stefanceriu:
This allows nightly users to enable OTLP tracing from the developer options menu. Configuration and credentials are read from the `secrets.xcconfig` and have been added as envars to all Xcode Cloud workflows.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -164,7 +164,7 @@ lane :config_nightly do
   data["settings"]["BASE_APP_GROUP_IDENTIFIER"] = "io.element.nightly"
   data["settings"]["BASE_BUNDLE_IDENTIFIER"] = "io.element.elementx.nightly"
 
-  config_maplibre()
+  config_secrets()
   
   File.open(target_file_path, 'w') { |f| YAML.dump(data, f) }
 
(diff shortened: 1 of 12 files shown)


kegsay (approved):
As someone who knows no swift, this looks fine. It's seemingly threading through the right vars which is good :D - it's worth emphasising here that we are well aware the user/pass here are not secret in that they get put into the final binary. We have other safeguards to prevent misuse, and have a user/pass is just one layer to stop random probe requests.

fastlane/Fastfile
```diff
@@ -414,13 +414,43 @@ private_lane :create_simulator_if_necessary do |options|
...
   set_xcconfig_value(
     path: './ElementX/SupportingFiles/secrets.xcconfig',
     name: 'MAPLIBRE_API_KEY',
-    value: api_key
+    value: maplibre_api_key
+  )
+
+  # URLs need special treatment to work properly in xcconfig files
+  # https://stackoverflow.com/a/36297483/730924
+  # i.e. use `https:/$()/` in the scheme
```
pixlwave, line 438: Feels like it might be more visible to also put this comment in the `secrets.xcconfig` file?

pixlwave (approved):

Issue #1418, opened 2023-07-28
Pasted content does not respect line breaks


germain-gg:
### Steps to reproduce

1. Play [Wordle](https://www.nytimes.com/games/wordle/index.html)
2. Copy your results
3. Paste them in the EX composer
4. Hit send

#### What did you expect?

<img width="175" alt="Screenshot 2023-07-28 at 10 57 49" src="https://github.com/vector-im/element-x-ios/assets/769871/226d3ec3-2616-4590-a990-a3f2548fd04d">

#### What happened instead?

<img width="389" alt="Screenshot 2023-07-28 at 10 58 05" src="https://github.com/vector-im/element-x-ios/assets/769871/e6d956c0-667c-4c56-bd50-4702dfe28726">

### Will you send logs?

No

amshakal:
Happens to me as well!

stefanceriu:
I'm not sure if this is because of something we changed recently or not but it seems to be working properly now. @germain-gg or @amshakal, can you confirm please?

![Screenshot 2023-08-08 at 13 52 55](https://github.com/vector-im/element-x-ios/assets/637564/627c2d2a-4b45-4785-a31c-dd070989519e)

germain-gg:
@stefanceriu I have tested and now when pasted in the composer the content respects the line breaks, but if you hit send, it inlines each guess from your Wordle result

bnjbvr:
How is the message created?

Hywan:
So…

The following Markdown:

```
Wordle 793 4/6

⬛⬛🟨⬛⬛
🟨⬛⬛⬛⬛
🟩🟩⬛⬛⬛
🟩🟩🟩🟩🟩
```

is compiled to the following HTML:

```html
<p>Wordle 793 4/6</p>
<p>⬛⬛🟨⬛⬛
🟨⬛⬛⬛⬛
🟩🟩⬛⬛⬛
🟩🟩🟩🟩🟩</p>

```

[...]

PR #1556, opened 2023-08-24
fix: Do not transform `\n` into `<br />` in the app

Hywan:
https://github.com/vector-im/element-x-ios/pull/1463 tried to fix https://github.com/vector-im/element-x-ios/issues/1418. However, the solution wasn't working properly as Markdown is difficult.

A solution has been provided by the Rust SDK [1] itself, via `ruma` [2]. Markdown soft line breaks are transformed into hard line breaks, when compiling Markdown to HTML. Doing that inside the Markdown parser itself is the correct place, as it's particularly difficult and tricky to detect soft line breaks correctly in Markdown.

This patch reverts #1463, but the Rust SDK must be updated to correctly experience the fix.

[1]: https://github.com/matrix-org/matrix-rust-sdk/pull/2452
[2]: https://github.com/ruma/ruma/pull/1636

### DIFF
--- a/UnitTests/Sources/AttributedStringBuilderTests.swift
+++ b/UnitTests/Sources/AttributedStringBuilderTests.swift
@@ -395,7 +395,7 @@ class AttributedStringBuilderTests: XCTestCase {
         
         XCTAssertEqual(numberOfBlockquotes, 3, "Couldn't find all the blockquotes")
     }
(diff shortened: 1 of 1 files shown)


Hywan:
Ah, it's been already removed, I can close the PR :-).

PR #1557, opened 2023-08-24
Signposting tweaks

### LINKED ISSUE #1529: Move performance markers to reflect what the user sees.
### Steps to reproduce

1. The "time to launch" perf graphs claim time to launch has been completely static for weeks:

Screenshot 2023-08-19 at 17 18 29

2. This is not true; specifically, there should be a shifts of some kind in the perf graphs for launch time at:

 * 339 when stefan reenabled the room cache (could be better or worse; hopefully better, but in the past it made things much worse due to being O(N) rather than O(1) perf)
 * a worsening of 1-2 seconds at 342 due to fuzzy search (https://github.com/vector-im/element-x-ios/issues/1514
 * a major shift at 346 due to https://github.com/matrix-org/matrix-rust-sdk/pull/2421 - which made launch time on my account simply not complete any more (30s+)

3. Need to move https://github.com/vector-im/element-x-ios/blob/a14c995686836efe482a22e72d4b78492837a1e9/ElementX/Sources/Application/AppCoordinator.swift#L572 or similar to more sensible places to actually reflect what the user sees, rather than just measuring the performance of a /sync request (which i think is what it's currently doing)?

#### What did you expect?

Accurate perf stats, otherwise they're useless.

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

345

### Homeserver

_No response_

### Will you send logs?

No

stefanceriu:
Split `sync` into `initialSync` and `firstRooms` which will track how long the very first sync takes separately from how long the first rooms appear in the list

Fixes #1529, goes in hand with #1558

### DIFF
--- a/ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
+++ b/ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
@@ -164,7 +164,7 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol 
                 
                 // End the initial sync performance timing once the room list is about to be displayed.
                 if roomListMode == .rooms, self.state.roomListMode == .skeletons {
-                    analytics.signpost.endSync()
+                    analytics.signpost.endFirstRooms()
                 }
                 
                 self.state.roomListMode = roomListMode
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -67,6 +67,8 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
         navigationRootCoordinator = NavigationRootCoordinator()
         
         Self.setupServiceLocator(navigationRootCoordinator: navigationRootCoordinator, appSettings: appSettings)
+        
+        ServiceLocator.shared.analytics.signpost.beginFirstRooms()
 
         ServiceLocator.shared.analytics.startIfEnabled()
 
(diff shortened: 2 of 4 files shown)


ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -163,8 +163,8 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
                 }
                 
                 // End the initial sync performance timing once the room list is about to be displayed.
-                if roomListMode == .rooms, self.state.roomListMode == .skeletons {
-                    analytics.signpost.endSync()
+                if roomListMode == .rooms {
+                    analytics.signpost.endFirstRooms()
```
pixlwave, line 167: 2 optional thoughts:

- Isn't going from skeletons -> rooms specifically a requirement for this signpost?
- Maybe `displayRooms` might be a more obvious name to use here?

pixlwave (approved):
LGTM 👍

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
stefanceriu: We agreed over the call this morning that we're interested in how long it takes to render the rooms from when you actually start the app. My thinking there is that it includes everything up to there, like restoring the session or even the previous behavior where the cache would block the app loading.

As for the second one: we might want to track other room list behavior in the future, like tracking how long it takes to render them after the first sync update finishes so for now I think `firstRooms` works.

michaelkaye (commented):
LGTM: You found that we needed to update the parsePerformanceMetrics.sh as well, which was missing from the other PR.

PR #1558, opened 2023-08-24
Signposting - app support: split `sync` into `initialSync` and `first…

### LINKED ISSUE #1529: Move performance markers to reflect what the user sees.

stefanceriu:
split `sync` into `initialSync` and `firstRooms` which will track how long the very first sync takes separately from how long the first rooms appear in the list

Fixes #1529, goes in hand with https://github.com/vector-im/element-x-ios/pull/1557


pixlwave (approved):
LGTM, although if suggested renaming makes sense in #1557, that presumably has an implication here.

PR #1565, opened 2023-08-25
Integration test tweaks

stefanceriu:


pixlwave (approved):
Nice, curious to see the next run and the change in run time 👍

PR #1566, opened 2023-08-25
Logging tweaks: keep default log levels for common and hyper no matte…

stefanceriu:
…r what the feature flag says


Velin92 (approved):
LGTM

PR #1579, opened 2023-08-28
Disable background app refreshes

stefanceriu:
- intentionally disabled until we get to the bottom of the performance issues we've been seeing
- it looks like the process is severely throttled when this runs causing slow syncs


Hywan (approved):
:+1:

PR #1588, opened 2023-08-29
Move send button out of composer's text field

aringenbach:
Move the send button outside of the field to follow latest designs. 
This enables for a better transition towards text field + formatting toolbar.


pixlwave:
Not a review, just looking at the changed file list in the email, don't forget to update the UserSessionFlow snapshots too :)

aringenbach:
> Not a review, just looking at the changed file list in the email, don't forget to update the UserSessionFlow snapshots too :)

Whoops totally missed these 🙈, thanks !

stefanceriu (approved):
This looks good to me but I think I found a bug: entering a space or tabs in the composer doesn't enable the send button. Seems like the `WysiwygComposerViewModel`s `$isContentEmpty` isn't firing with updates in that case although the textView reports the correct string `[TextView] Sel(att): {1, 0} | Text: "	" | textViewDidChange(_:)`

PR #1591, opened 2023-08-29
Use Safari for OIDC account URL.

pixlwave:
Following on from https://github.com/matrix-org/matrix-authentication-service/pull/1598, cookies are now correctly shared from the `ASWebAuthenticationSession` to Safari. As requested by the auth team, we now open the account URL in Safari. The WAS is still used in the follow scenarios:

- On the simulator - it doesn't get access to the cookie in Safari for some reason.
- When Safari is disabled behind Screen Time (by checking for canOpenURL).

Additionally, this PR will present RP-Initiated Logout to the user when returned by the authentication issuer (untested).


stefanceriu (approved):
lgtm 👍

PR #1604, opened 2023-08-31
New Room button has been moved to the top

### LINKED ISSUE #1602: Move the compose new message button to the top
https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X?type=design&node-id=5-7383&mode=design&t=AKD0Pq0Cb38RnlFh-0

Velin92:
fixes #1602

Moved to the new room button to the top.
(Need to do other changes to the UI, so I will generate the UI tests when all the PRs are merged)

Simulator Screenshot - iPhone 14 - 2023-08-31 at 15 44 48

### DIFF
--- a/ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
+++ b/ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
@@ -163,8 +163,7 @@ struct HomeScreen: View {
             HomeScreenUserMenuButton(context: context)
         }
         
-        ToolbarItemGroup(placement: .bottomBar) {
-            Spacer()
+        ToolbarItemGroup(placement: .navigationBarTrailing) {
             newRoomButton
         }
     }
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
```diff
@@ -163,8 +163,7 @@ struct HomeScreen: View {
             HomeScreenUserMenuButton(context: context)
         }
         
-        ToolbarItemGroup(placement: .bottomBar) {
-            Spacer()
+        ToolbarItemGroup(placement: .navigationBarTrailing) {
```
pixlwave, line 166: Should probably use `.primaryAction` here fwiw. It *might* even mean you don't need the fontWeight below, I can't remember if that's only the case for `.confirmationAction`.

pixlwave (commented):
We can now remove all that code about changing the bottom bar visibility too.

ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
Velin92: seems that the fontWeight is still required but changed this!

PR #1606, opened 2023-08-31
Fix some accessibility modifiers

nimau:
This PR fixes some accessibility modifiers (`.accessibilityLabel()` was used instead of `.accessibilityIdentifier()`)


pixlwave (approved):
Gosh there's a lot of mistakes 😳 Hoping Xcode 15's accessibility audit is going to help us with things like this.

Thank you 🙏

PR #1607, opened 2023-08-31
Move the invites summary provider configuration the underlying slidin…

stefanceriu:
…g sync list is availablle sooner now


pixlwave (approved):
Nice 👌

PR #1608, opened 2023-08-31
Polls refinements

alfogrillo:
This PR:
- Deletes the feature flags for polls
- Updates the "end poll" UI in the timeline
- Refines the focus logics in the "create poll" form
- Adds an alert when the user discards a "draft poll"


ElementX/Sources/Screens/CreatePollScreen/View/CreatePollScreen.swift
```diff
@@ -39,6 +39,8 @@ struct CreatePollScreen: View {
         .navigationBarTitleDisplayMode(.inline)
         .toolbar { toolbar }
         .animation(.elementDefault, value: context.options)
+        .interactiveDismissDisabled(context.viewState.bindings.hasDraftContent)
```
stefanceriu, line 42: Shouldn't this be always disabled? I don't want to lose even my draft content by mistake 🤔

UITests/Sources/CreatePollScreenUITests.swift
```diff
@@ -56,6 +56,9 @@ class CreatePollScreenUITests: XCTestCase {
             addOption.tap()
         }
 
+        if app.keyboards.count > 0 {
```
stefanceriu, line 59: 🙊

stefanceriu (approved):
lgtm 👍

ElementX/Sources/Screens/CreatePollScreen/View/CreatePollScreen.swift
alfogrillo: The idea is to disable the interaction if the user wrote at least some character somewhere. 
If the screen is completely empty we still give the possibility to drag down the sheet. Does it make sense?
Maybe just `hasContent` clarifies better the intent?

UITests/Sources/CreatePollScreenUITests.swift
alfogrillo: Playing with `@FocusState` apparently doesn't work like expected on iPad.
But I fell like is a minor problem.

stefanceriu: Yeah, don't worry about it too much

ElementX/Sources/Screens/CreatePollScreen/View/CreatePollScreen.swift
alfogrillo: Applied!
https://github.com/vector-im/element-x-ios/pull/1608/commits/04e63f5d0dd2781ba8944955e966ead71b138ecc

stefanceriu: Thanks!

PR #1609, opened 2023-08-31
Update RTE version to 2.8.0

aringenbach:
Only currently significant change for this version: 
OS logs are now configurable. By default only errors will show in the console.


Velin92 (approved):
LGTM

PR #1612, opened 2023-09-01
Be more lenient with fatalErrors coming out of equal transitions, cau…

stefanceriu:
…sed by asynchronicity on the UI layer


pixlwave (approved):
Very nice. Am I right in thinking it makes this redundant now:
https://github.com/vector-im/element-x-ios/blob/38d6ffb3632a838c85bdd2b6a68f1f823b674491/ElementX/Sources/Application/AppCoordinator.swift#L264-L267

stefanceriu:
> Very nice. Am I right in thinking it makes this redundant now:

Well sort of, it's still nice not to log an error for it 🤔

pixlwave:
Ok right, sorry I get the intention now, its to find more cases like that and handle them however 👍

PR #1613, opened 2023-09-01
Fixes #980 - Generated thumbnail size when uploading media

### LINKED ISSUE #980: Sending photos from EIX results in weird vertical padding on EW
I assume we're sending the wrong dimensions somehow on the thumbnail/image on the metadata sent by EIX?  Images from other clients seem to be fine:

Screenshot 2023-05-29 at 15 19 28

n.b. empty space above & below the image.

stefanceriu:


Velin92 (approved):

pixlwave (approved):
Good catch!

PR #1614, opened 2023-09-01
Add RTE formatting buttons

aringenbach:
Add an option to display formatting buttons when RTE is active


ElementX/Sources/Screens/ComposerToolbar/ComposerToolbarViewModel.swift
```diff
@@ -128,4 +163,121 @@ final class ComposerToolbarViewModel: ComposerToolbarViewModelType, ComposerTool
...
+                  title: L10n.richTextEditorEditLink,
+                  primaryButton: AlertInfo<UUID>.AlertButton(title: L10n.actionRemove, role: .destructive, action: {
+                      self.restoreComposerSelectedRange()
+                      self.removeLinks()
+                  }),
+                  verticalButtons: [AlertInfo<UUID>.AlertButton(title: L10n.actionEdit, action: {
+                      self.state.bindings.alertInfo = nil
+                      DispatchQueue.main.async {
+                          self.state.bindings.alertInfo = self.makeSetUrlAlertInfo(urlBinding: urlBinding, isEdit: true)
+                      }
```
aringenbach, line 244: reaaaally not ideal but it works

ElementX/Sources/Other/Extensions/Alert.swift
```diff
@@ -52,10 +52,24 @@ extension View {
...
 struct AlertInfo<T: Hashable>: Identifiable, AlertProtocol {
-    struct AlertButton {
+    struct AlertButton: Identifiable {
         let title: String
         var role: ButtonRole?
         let action: (() -> Void)?
+
+        var id: UUID {
+            UUID()
+        }
```
alfogrillo, line 62: Should be this a stored property?
Otherwise the id would be different each time 🤔

ElementX/Sources/Other/Extensions/Alert.swift
```diff
@@ -52,10 +52,24 @@ extension View {
...
+    }
+
+    struct AlertTextField: Identifiable {
+        let placeholder: String
+        let text: Binding<String>
+        let autoCapitalization: TextInputAutocapitalization
+
+        var id: UUID {
+            UUID()
+        }
```
alfogrillo, line 72: same here

ElementX/Sources/Other/Extensions/Alert.swift
aringenbach: It doesn't break anything since its used as a one shot in the ForEach, but yeah definitely. Fixed in [a6d41e2](https://github.com/vector-im/element-x-ios/pull/1614/commits/a6d41e2e098395f88612ae26790fbcef709169ed)

ElementX/Sources/Other/AccessibilityIdentifiers.swift
```diff
@@ -41,6 +41,7 @@ struct A11yIdentifiers {
     static let notificationSettingsScreen = NotificationSettingsScreen()
     static let notificationSettingsEditScreen = NotificationSettingsEditScreen()
     static let createPollScreen = CreatePollScreen()
+    static let composerToolbar = ComposerToolbar()
```
stefanceriu, line 44: I've been trying to keep these alphabetically sorted but I'm not sure that's worth it anymore

ElementX/Sources/Other/AccessibilityIdentifiers.swift
```diff
@@ -213,4 +215,21 @@ struct A11yIdentifiers {
             "\(roomNamePrefix):\(name)"
         }
     }
+
+    struct ComposerToolbar {
```
stefanceriu, line 219: I wonder if we should move this to within the RoomScreen, all the other identifiers are per room not component

ElementX/Sources/Screens/ComposerToolbar/View/FormattingToolbar.swift
```diff
@@ -0,0 +1,61 @@
...
+import SwiftUI
+
+struct FormattingToolbar: View {
+    /// The list of items to render in the toolbar
+    var formatItems: [FormatItem]
+    /// The action when an item is selected
+    var formatAction: (FormatType) -> Void
+
+    var body: some View {
+        ScrollView(.horizontal) {
```
stefanceriu, line 26: This scroll view is very hidden, didn't even realise we can scroll the formatting toolbar until I saw the code

ElementX/Sources/Screens/ComposerToolbar/View/FormattingToolbar.swift
```diff
@@ -0,0 +1,61 @@
...
+                    .background(getBackgroundColor(for: item))
+                    .cornerRadius(8)
+                    .accessibilityIdentifier(item.accessibilityIdentifier)
+                    .accessibilityLabel(item.accessibilityLabel)
+                }
+            }
+        }
+    }
+
+    private func getForegroundColor(for item: FormatItem) -> Color {
```
stefanceriu, line 47: `getSomething` doesn't feel very SwiftUI-ish, should we drop the `get` from these methods? 🤔

ElementX/Sources/Screens/ComposerToolbar/ComposerToolbarViewModel.swift
```diff
@@ -65,6 +90,7 @@ final class ComposerToolbarViewModel: ComposerToolbarViewModelType, ComposerTool
                 actionsSubject.send(.sendPlainTextMessage(message: context.composerPlainText,
                                                           mode: state.composerMode))
             }
+            state.bindings.composerActionsEnabled = false
```
stefanceriu, line 93: Feels very strange to me that the formatting options don't stick around after sending a message, or coming back to a room for that matter

ElementX/Sources/Screens/ComposerToolbar/View/ComposerToolbar.swift
```diff
@@ -26,29 +26,65 @@ struct ComposerToolbar: View {
...
+            }
+        }
+        .alert(item: $context.alertInfo)
+    }
+
+    private var topBar: some View {
         HStack(alignment: .bottom, spacing: 10) {
-            RoomAttachmentPicker(context: context)
-                .padding(.bottom, 5) // centre align with the send button
+            if !context.composerActionsEnabled {
```
stefanceriu, line 40: Why don't we allow this? 🤔

stefanceriu (approved):
This looks amazing 👏 I personally think the UX can be improved but that's something to pick up with the design team and doesn't reflect on this PR

ElementX/Sources/Other/AccessibilityIdentifiers.swift
alfogrillo: It looks like we are adding to the the end nowadays.

ElementX/Sources/Screens/ComposerToolbar/View/FormattingToolbar.swift
alfogrillo: Agreed. Applied: https://github.com/vector-im/element-x-ios/pull/1614/commits/5186834c19cd185e3229f2d1f2b0e035518be7f7

ElementX/Sources/Screens/ComposerToolbar/View/ComposerToolbar.swift
alfogrillo: My understating is that when the RTE bar shows, we hide the attachment button which is "replaced" with the "cross" button in the RTE bar.
This is how it would look like if we show both:

Screenshot 2023-09-05 alle 09 49 09

ElementX/Sources/Screens/ComposerToolbar/ComposerToolbarViewModel.swift
alfogrillo: Yep, this seems odd to me as well. Someone raised this flag already, let's see if people want to change it in the future...

ElementX/Sources/Other/AccessibilityIdentifiers.swift
alfogrillo: Are you suggesting to nest `ComposerToolbar` under `RoomScreen`? 
It looks good to me having a namespace for all the RTE stuff.

stefanceriu: Yep, that's what I was suggesting but no strong feelings

PR #1625, opened 2023-09-04
Limit the number of rooms the summary provider handles to 500

stefanceriu:
In order to avoid overloading the room summary serial diff processing queue with diffs the user most likely is not interested in, this PR introduced a hard limit on the amount of rooms that we show in the room list. It does so by simply ignoring diffs outside of that range.
Searching and the rest of the functionality is not affected.

P.S. Tried unit testing but couldn't find a way around private constructors or closed classes in the various uniffi components needed


stefanceriu:
Closing this in favor of a rust side batching solution

PR #1628, opened 2023-09-05
Add client and OIDC tracing targets.

pixlwave:
This PR adds sdk/client and sdk/oidc as tracing targets and includes them (plus the ffi target which was missing) in the default config.


stefanceriu (approved):
👍 🚀

PR #1630, opened 2023-09-05
Avatar Bloom View in home screen navigation bar

Velin92:
Simulator Screenshot - iPhone 14 - 2023-09-07 at 16 24 43

https://github.com/vector-im/element-x-ios/assets/34335419/67d0a9c9-66ba-44b4-aee2-f9b00745fa0b

Simulator Screenshot - iPhone 14 - 2023-09-07 at 16 35 35

### DIFF
--- a/ElementX/Sources/Other/ScrollViewAdapter.swift
+++ b/ElementX/Sources/Other/ScrollViewAdapter.swift
@@ -49,8 +49,8 @@ class ScrollViewAdapter: NSObject, UIScrollViewDelegate {
             .eraseToAnyPublisher()
     }
     
-    private let scrollViewIsAtTopEdge: CurrentValueSubject<Bool, Never> = .init(false)
-    var scrollViewIsAtTopEdgePublisher: CurrentValuePublisher<Bool, Never> {
+    private let isAtTopEdgeSubject: CurrentValueSubject<Bool, Never> = .init(false)
+    var isAtTopEdge: CurrentValuePublisher<Bool, Never> {
         scrollViewIsAtTopEdge
             .asCurrentValuePublisher()
     }
--- a/ElementX/Sources/Other/ScrollViewAdapter.swift
+++ b/ElementX/Sources/Other/ScrollViewAdapter.swift
@@ -49,11 +49,19 @@ class ScrollViewAdapter: NSObject, UIScrollViewDelegate {
             .eraseToAnyPublisher()
     }
     
+    private let scrollViewIsAtTopEdge: CurrentValueSubject<Bool, Never> = .init(false)
+    var scrollViewIsAtTopEdgePublisher: CurrentValuePublisher<Bool, Never> {
+        scrollViewIsAtTopEdge
+            .asCurrentValuePublisher()
+    }
+    
     // MARK: - UIScrollViewDelegate
     
     func scrollViewDidScroll(_ scrollView: UIScrollView) {
         didScrollSubject.send(())
         updateScrollDirection(scrollView)
+        let contentInsetOffset = scrollView.contentOffset.y + scrollView.contentInset.top
+        scrollViewIsAtTopEdge.send(contentInsetOffset >= 3)
     }
     
     func scrollViewWillBeginDragging(_ scrollView: UIScrollView) {
--- a/ElementX/Sources/Other/ScrollViewAdapter.swift
+++ b/ElementX/Sources/Other/ScrollViewAdapter.swift
@@ -60,7 +60,7 @@ class ScrollViewAdapter: NSObject, UIScrollViewDelegate {
     func scrollViewDidScroll(_ scrollView: UIScrollView) {
         didScrollSubject.send(())
         updateScrollDirection(scrollView)
-        let contentInsetOffset = scrollView.contentOffset.y + scrollView.contentInset.top
+        let insetContentOffset = scrollView.contentOffset.y + scrollView.contentInset.top
         scrollViewIsAtTopEdge.send(contentInsetOffset >= 3)
     }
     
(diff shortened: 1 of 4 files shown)


janogarcia (changes requested):
Thanks for the demos and test build, Mauro! We're getting super close to the final implementation.

- Blending modes: We updated the values as follows. Light theme has a `Hard Light` bottom layer @ 20% opacity, and a top `Color` layer @ 80% opacity. Dark theme has a `Exclusion` bottom layer @ 50% opacity, and a top `Color` layer @ 20% opacity
- Soft edge mask: We settled 40px/pt for the linear gradient height.
- Collapsed navigation bar: The bloom shouldn't change its position, if possible, and just be cropped instead (e.g., [behavior on Android](https://matrix.to/#/!tuoNvqSfUFMgFtnlKz:matrix.org/$AB-j8TtBuWRQtizUnJH_dvkz_JCWKTVTm9uHE7Fvb_8?via=element.io&via=matrix.org&via=jki.re)).

janogarcia (changes requested):
Based on the latest screenshots that @Velin92 just shared over DM.

- When the Navigation Bar is not collapsed, the linear gradient mask for the "soft edge" effect seems to not be properly placed, there's a small visible gap.
- When the Navigation Bar is collapsed, we don't have to render the linear gradient mask for the "soft edge" effect.
- We should restore the hairline border on the Navigation Bar bottom.

ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
```diff
@@ -87,10 +90,68 @@ struct HomeScreen: View {
...
+    }
+            
+    private func setBloomView(controller: UIViewController) {
+        guard let navigationBarContainer = controller.navigationController?.navigationBar.subviews.first,
+              let leftBarButtonView = controller.navigationItem.leadingItemGroups.first?.barButtonItems.first?.customView else {
+            return
+        }
+        
+        navigationBarContainer.clipsToBounds = true
+        let hostingController = UIHostingController(rootView: bloom)
```
pixlwave, line 124: Might be nice to call this `bloomController` now there's a `gradientController`

ElementX/Sources/Other/ScrollViewAdapter.swift
```diff
@@ -49,11 +49,19 @@ class ScrollViewAdapter: NSObject, UIScrollViewDelegate {
             .eraseToAnyPublisher()
     }
     
+    private let scrollViewIsAtTopEdge: CurrentValueSubject<Bool, Never> = .init(false)
+    var scrollViewIsAtTopEdgePublisher: CurrentValuePublisher<Bool, Never> {
```
pixlwave, line 53: Looking at the ones above, this would be more aligned naming
```suggestion
    private let isAtTopEdgeSubject: CurrentValueSubject<Bool, Never> = .init(false)
    var isAtTopEdge: CurrentValuePublisher<Bool, Never> {
```

ElementX/Sources/Other/ScrollViewAdapter.swift
```diff
@@ -49,11 +49,19 @@ class ScrollViewAdapter: NSObject, UIScrollViewDelegate {
...
+        scrollViewIsAtTopEdge
+            .asCurrentValuePublisher()
+    }
+    
     // MARK: - UIScrollViewDelegate
     
     func scrollViewDidScroll(_ scrollView: UIScrollView) {
         didScrollSubject.send(())
         updateScrollDirection(scrollView)
+        let contentInsetOffset = scrollView.contentOffset.y + scrollView.contentInset.top
```
pixlwave, line 63: Bah, I finally realised the ordering that makes sense 😅
```suggestion
        let insetContentOffset = scrollView.contentOffset.y + scrollView.contentInset.top
```

ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
```diff
@@ -20,10 +20,16 @@ import SwiftUI
 import SwiftUIIntrospect
 
 struct HomeScreen: View {
+    @Environment(\.colorScheme) private var colorScheme
```
pixlwave, line 23: This one is unused now :)

ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
```diff
@@ -20,10 +20,16 @@ import SwiftUI
...
     @ObservedObject var context: HomeScreenViewModel.Context
     
     @State private var scrollViewAdapter = ScrollViewAdapter()
     @State private var isSearching = false
+    @State private var bloomView: UIView?
```
pixlwave, line 28: Maybe useful to group these together now there is so many of them 🙃
```suggestion
    
    // Bloom components
    @State private var bloomView: UIView?
```

ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
```diff
@@ -87,10 +93,96 @@ struct HomeScreen: View {
...
     }
     
     // MARK: - Private
     
+    private var bloomGradient: some View {
+        LinearGradient(colors: [.clear, .compound.bgCanvasDefault], startPoint: .top, endPoint: .bottom)
+            .mask {
+                LinearGradient(stops:
+                    [.init(color: .white, location: 0.75), .init(color: .clear, location: 1.0)],
+                    startPoint: .leading, endPoint: .trailing)
```
pixlwave, line 133: Should probably be like this (I thought SwiftFormat would be upset in some way about this).
```suggestion
                LinearGradient(stops: [.init(color: .white, location: 0.75), .init(color: .clear, location: 1.0)],
                               startPoint: .leading,
                               endPoint: .trailing)
```

ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
```diff
@@ -87,10 +93,96 @@ struct HomeScreen: View {
...
+            }
+            .ignoresSafeArea(edges: .all)
+    }
+            
+    private func setBloomView(controller: UIViewController) {
```
pixlwave, line 138: Might be a bit clearer as `set` could be interpreted that `controller` is the bloom view controller.
```suggestion
    private func makeBloomView(controller: UIViewController) {
```

pixlwave (approved):
👏👏👏

PR #1638, opened 2023-09-05
Initial skeleton for Universal Link handling.

pixlwave:
This PR adds the initial implementation for receiving Universal Links into the App Coordinator. It doesn't implement parsing `AppRoute`s from the links or do anything with them yet though.

Additionally it removes all use of `ServiceLocator.shared.settings` in the App Coordinator.

@stefanceriu You're welcome to merge this tomorrow morning if you're happy with it and need to use it.


ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -132,6 +135,15 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
         )
     }
     
+    func handleUniversalLink(_ url: URL) {
```
stefanceriu, line 138: I wonder if we should rename this to just `handleURL` as it should probably handle non-universal links too

stefanceriu (approved):
lgtm 👍

PR #1643, opened 2023-09-06
Generic element call link support

stefanceriu:
This PR adds support for Element Call universal links
* removes `swift-url-routing` as it wasn't used and doesn't bring much value on app.element.io links (fragments)
* replaces ^ with a custom (for the moment trivial) solution
* parses Call links into AppRoutes and handles them accordingly
* enable extra background modes to allow the calls to run properly (pending web changes)


ElementX/Sources/Application/Navigation/AppRoutes.swift
```diff
@@ -0,0 +1,60 @@
...
+    case roomDetails(roomID: String)
+    case invites
+    case genericCallLink(url: URL)
+}
+
+enum AppRouteURLParser {
+    private enum KnownHosts: String, CaseIterable {
+        case elementIo = "element.io"
+        case appElementIo = "app.element.io"
+        case statingElementIo = "staging.element.io"
```
pixlwave, line 31:
```suggestion
        case stagingElementIo = "staging.element.io"
```

UnitTests/Sources/AppRoutesTests.swift
```diff
@@ -0,0 +1,30 @@
...
+// limitations under the License.
+//
+
+import XCTest
+
+@testable import ElementX
+
+class AppRouterTests: XCTestCase {
+    func testRoutes() {
+        guard let url = URL(string: "https://call.element.io/test") else {
```
pixlwave, line 23: Could we add the URL `ExpressibleByStringLiteral` extension to the unit tests target?

ElementX/Sources/Screens/Other/GenericCallLinkCoordinator.swift
```diff
@@ -0,0 +1,77 @@
...
+            configuration.allowsPictureInPictureMediaPlayback = true
+            
+            webView = WKWebView(frame: .zero, configuration: configuration)
+            webView.uiDelegate = self
+        }
+
+        // MARK: - WKUIDelegate
+        
+        func webView(_ webView: WKWebView, decideMediaCapturePermissionsFor origin: WKSecurityOrigin, initiatedBy frame: WKFrameInfo, type: WKMediaCaptureType) async -> WKPermissionDecision {
+            .grant
```
pixlwave, line 74: *please* can we check the origin before blindly streaming the users camera to the page presented here.
```suggestion
            guard origin.host == appSettings.elementCallURL.host else { return .deny }
            .grant
```

pixlwave: Could also add the navigation delegate to make sure you don't end up with a link on the page taking you outside the call:

https://github.com/pixlwave/Element-Call-macOS/blob/6407a23458ae95b402922e4ebfdb357b4ecef819/Sources/VOIPCoordinator.swift#L118-L131

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -135,6 +130,15 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
 
     // MARK: - Private
     
+    private func clearPresentedSheets(animated: Bool, completion: @escaping () -> Void) {
```
pixlwave, line 133: Should we make this async with a continuation instead? Or indeed just a `Task.sleep`

ElementX/Sources/Application/Navigation/NavigationRootCoordinator.swift
```diff
@@ -45,6 +65,25 @@ class NavigationRootCoordinator: ObservableObject, CoordinatorProtocol, CustomSt
         
         rootModule = NavigationModule(coordinator, dismissalCallback: dismissalCallback)
     }
+    
+    ///   - dismissalCallback: called when the sheet has been dismissed, programatically or otherwise
```
pixlwave, line 69: This is either missing more docs or the doc was unintentionally copied 🙈

ElementX/Sources/Application/Navigation/AppRoutes.swift
```diff
@@ -0,0 +1,60 @@
...
+
+enum AppRoute: Equatable {
+    case roomList
+    case room(roomID: String)
+    case roomDetails(roomID: String)
+    case invites
+    case genericCallLink(url: URL)
+}
+
+enum AppRouteURLParser {
```
pixlwave, line 27: I wonder if the architecture of this should be more modular. E.g. you supply it with an array of `LinkParserProtocol` which can be set a build time, so forks can use their own universal link domains. Then you do a `parsers.first { $0.host == url.host }` and let each parser have its own logic per domain?

pixlwave (approved):
LGTM, although some comments in there might be quite big changes.

ElementX/Sources/Screens/Other/GenericCallLinkCoordinator.swift
stefanceriu: Good idea 👍

ElementX/Sources/Application/Navigation/NavigationRootCoordinator.swift
stefanceriu: No no, that's correct 😁

UnitTests/Sources/AppRoutesTests.swift
stefanceriu: We would need to import a bunch of other files too (MXLog, FileManager etc) , don't think it's worth it

ElementX/Sources/Application/Navigation/AppRoutes.swift
stefanceriu: I though about that (was thinking of how we handle push rules in the old app) but honestly, I don't think its worth it until we actually go implement deep linking.

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
stefanceriu: We could but that complicates things within `handleAppRoute` which is not async at the moment. Let's tackle that when we decide what to do with actual deeplinking and handling propagation through the various flow coordinators

ElementX/Sources/Screens/Other/GenericCallLinkCoordinator.swift
pixlwave: Make sense 👍

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
pixlwave:
```
Task {
    await clear()
    doStuff()
}
```

is fundamentally the same as

```
clear() {
    doStuff()
}
```

but easier to read imo, but I hear you 👍

ElementX/Sources/Screens/Other/GenericCallLinkCoordinator.swift
```diff
@@ -71,7 +73,29 @@ private struct WebView: UIViewRepresentable {
...
+                return .allow
+            }
+            
+            // Additionally allow any embedded content such as captchas.
+            if let targetFrame = navigationAction.targetFrame, !targetFrame.isMainFrame {
+                return .allow
+            }
+            
+            // Otherwise the request is invalid.
+            return .cancel
```
pixlwave, line 98: We could in the future do a `UIApplication.shared.open(navigationAction.request.url) at this stage, but probably not worth thinking about right now.

pixlwave (approved):
LGTM 😎

PR #1650, opened 2023-09-07
Bump the RustSDK to v1.1.10

stefanceriu:


pixlwave (approved):
LGTM, should probably double check those unit test failures in case its related.

stefanceriu:
I did, there's working locally

PR #1652, opened 2023-09-07
Adopt the new room list service syncIndicatorListener

stefanceriu:
* expose sync service loading state through a client proxy current value publisher and use it accordingly
* a current value publisher because the UI might not be instantiated by the time it finishes loading

### DIFF
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -598,33 +598,25 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
         ServiceLocator.shared.analytics.signpost.beginFirstSync()
         userSession.clientProxy.startSync()
         
-        let identifier = "StaleDataIndicator"
-        
-        func showLoadingIndicator() {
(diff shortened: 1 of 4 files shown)


ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -598,33 +598,25 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
...
         clientProxyObserver = userSession.clientProxy
-            .callbacks
+            .loadingStatePublisher
+            .removeDuplicates()
             .receive(on: DispatchQueue.main)
-            .sink { action in
-                switch action {
-                case .startedUpdating:
-                    showLoadingIndicator()
-                case .receivedSyncUpdate:
```
pixlwave, line 623: These still exist right? Double checking that the MigrationScreen will still behave as expected.

pixlwave (approved):

ElementX/Sources/Application/AppCoordinator.swift
stefanceriu: `receivedSyncUpdate` still exists because it's used in the MigrationScreen and SessionVerification but the other one I removed

PR #1654, opened 2023-09-07
Stop the sync service constantly attempting to restart on auth error.

pixlwave:
Additionally removes a retain cycle that was preventing the client from being released.

This really isn't the ultimate solution. The sync service should provide us with more context as to what error occurred, and more specifically, what conditions need to change before a restart is triggered (e.g. network conditions/some user input/none, its fatal)

### DIFF
--- a/ElementX/Sources/Services/Client/ClientProxy.swift
+++ b/ElementX/Sources/Services/Client/ClientProxy.swift
@@ -131,7 +131,10 @@ class ClientProxy: ClientProxyProtocol {
     }
 
     func startSync() {
-        guard !hasEncounteredAuthError else { return }
+        guard !hasEncounteredAuthError else {
+            MXLog.warning("Ignoring request, this client has an unknown token.")
+            return
+        }
         
         MXLog.info("Starting sync")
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -372,6 +380,7 @@ class ClientProxy: ClientProxyProtocol {
                 MXLog.error("Failed restarting the sync service with error: \(error)")
             }
             
+            guard !hasEncounteredAuthError else { return }
```
stefanceriu, line 383: Should we move this one higher above so it doesn't even print anything or attempt to stop the service? Also, should we maybe log something when we're in this state?

stefanceriu (approved):
lgtm 👍

ElementX/Sources/Services/Client/ClientProxy.swift
pixlwave: Discussed this and we concluded that restartSync is no longer a necessary method.

PR #1657, opened 2023-09-08
Enable RUST_BACKTRACE by default

stefanceriu:


ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -230,6 +232,10 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
     
     // MARK: - Private
     
+    private static func setupEnvironmentVariables() {
+        setenv("RUST_BACKTRACE", "1", 1)
```
pixlwave, line 236: Just so I understand, the last `1` is telling it to overwrite an existing value?

pixlwave (approved):

ElementX/Sources/Application/AppCoordinator.swift
stefanceriu: Yep, do you think I should extrat that into a small function just to make the parameters clearer?

pixlwave: Only if you want. Took me 2 minutes to find the docs and figure it out.

stefanceriu: Cool, then no I guess, doesn't bring a lot of value

PR #1660, opened 2023-09-08
Top level deeplink handling

stefanceriu:
* move link handling to the top most level of the application. It will then invoke the AppCoordinator, parse the route and trickle it throughout the application
* add support for a custom app url scheme


ElementX/SupportingFiles/target.yml
```diff
@@ -54,6 +54,15 @@ targets:
         CFBundleDisplayName: $(APP_DISPLAY_NAME)
         CFBundleShortVersionString: $(MARKETING_VERSION)
         CFBundleVersion: $(CURRENT_PROJECT_VERSION)
+        CFBundleURLTypes: [
+          {
+            CFBundleTypeRole: Editor,
+            CFBundleURLName: $(BASE_BUNDLE_IDENTIFIER),
```
pixlwave, line 60: This is the only bit I can't say I understand. The name for `vector://` is `io.element.elementx`?

pixlwave (approved):
Looks good to me! Kind of wish we could use `elementCall://` though rather than the old `vector` name.

PR #1678, opened 2023-09-11
Fix button and make mxid copyable.

### LINKED ISSUE #1669: There is no way to copy a user’s matrix id from their member info page
### Steps to reproduce

1. Need to access a user’s matrix id (eg to enter it into moderation tooling)
2. Open their member info by tapping on their avatar or from the member list (yay!)
3. See their mxid shown
4. Try to long-tap or tap on it to select it to copy it
5. …Instead see a zoomed form of their avatar.
6. …wince, especially if you were trying to ban an abusive user.
7. have to manually remember their mxid and type it back in, instead

#### What did you expect?

Copyable text field; only tapping on the avatar should zoom it, not the whole member info.

#### What happened instead?

Zoomed avatar

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

380

### Homeserver

_No response_

### Will you send logs?

No

pixlwave:
On the room/member details screens, the avatar button was applied to the entire row, this PR fixes that and makes the MXID copyable.

Fixes #1669

### DIFF
--- a/ElementX/Sources/Other/SwiftUI/Views/AvatarHeaderView.swift
+++ b/ElementX/Sources/Other/SwiftUI/Views/AvatarHeaderView.swift
@@ -37,6 +37,7 @@ struct AvatarHeaderView<Footer: View>: View {
                                     avatarSize: avatarSize,
                                     imageProvider: imageProvider)
             }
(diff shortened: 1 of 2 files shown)


stefanceriu (approved):
🐑 it!

PR #1680, opened 2023-09-11
Polish polls UX

alfogrillo:
This PR polishes the poll experience by:

- Updating the "final votes" copy
- Adding the haptic feedback on voting
- Fixing the "empty bar" color for an option

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
+++ b/ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
@@ -503,10 +503,15 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol 
 
         var debugActions: [TimelineItemMenuAction] = appSettings.canShowDeveloperOptions ? [.viewSource] : []
 
-        if let encryptedItem = timelineItem as? EncryptedRoomTimelineItem,
-           case let .megolmV1AesSha2(sessionID) = encryptedItem.encryptionType {
-            debugActions.append(.retryDecryption(sessionID: sessionID))
-            return .init(actions: [], debugActions: debugActions)
+        if let encryptedItem = timelineItem as? EncryptedRoomTimelineItem {
+            switch encryptedItem.encryptionType {
+            case .megolmV1AesSha2(let sessionID):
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Screens/RoomScreen/View/Timeline/PollOptionView.swift
```diff
@@ -74,7 +71,7 @@ private struct PollProgressView: View {
         GeometryReader { geometry in
             ZStack(alignment: .leading) {
                 Capsule()
-                    .foregroundColor(.compound.borderDisabled)
```
alfogrillo, line 77: @pixlwave  `colorAlphaGray500` isn't available under `CompoundColors`.
 I'm not sure if I need to open a PR for adding it in Compound. 
Don't know which name should I give to it in case.
wdyt?

pixlwave: You'll need to add it as a deprecated token like [these](https://github.com/vector-im/compound-ios/blob/71b0c887880de639c854600a9ada6b5bcc93c719/Sources/Compound/Colors/CompoundColors.swift#L117-L119) until design has exposed it as a semantic token.

alfogrillo: Added one [here](https://github.com/vector-im/compound-ios/pull/24). 
I'm updating the package as soon as it gets merged.

ElementX/Sources/Screens/RoomScreen/View/Timeline/PollRoomTimelineView.swift
```diff
@@ -23,15 +23,18 @@ struct PollRoomTimelineView: View {
     @ScaledMetric private var summaryPadding = 32
     @ScaledMetric private var iconSize = 22
 
+    private let feedbackGenerator = UIImpactFeedbackGenerator(style: .heavy)
```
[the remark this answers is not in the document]
alfogrillo: We didn't discuss it explicitly, but to me `UIImpactFeedbackGenerator` feels better also because it seems it's the only one we are using on EX. I would keep it for now, and allow the designers to think on it.

PR #1688, opened 2023-09-12
Parse markdown and html from push notifications

### LINKED ISSUE #1679: Push notifications leak HTML + Markdown in rich messages
### Steps to reproduce

1. Receive a push with a rich message
2. Observe that it shows the raw body of the message, which includes MD (and by extension HTML) formatting, which looks ugly and unprofessional.  The recipient doesn't care what markup the sender used to write the message; they care about reading its contents instead.

e.g.

IMG_2858
IMG_2857

#### What did you expect?

HTML of rich msgs should be 'flattened' to plaintext to be rendered in notifications, stripping out the tags.

(It's interesting that in this example with bnjbvr, the sense of the message would have been lost, however, by stripping out the <del> tags. Perhaps we leave semantically important tags in place.)

#### What happened instead?

Ugly mix of MD & tags in notifs.

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

380

### Homeserver

_No response_

### Will you send logs?

No

stefanceriu:
Fixes #1679 

This PR:
* extracts the logic that generates the latest room message text into the `RoomMessageEventStringBuilder` and reuses that for push notifications
* it simplifies the existing NSE logic and moves content generation outside of a `NotificationItemProxyProtocol` extension and into the `NotificationContentBuilder`
* as part of these changes notifications are not built from the parsed version of the formatted message body


ElementX/Sources/Screens/Settings/NotificationSettingsScreen/NotificationSettingsScreenViewModel.swift
```diff
@@ -209,3 +209,9 @@ class NotificationSettingsScreenViewModel: NotificationSettingsScreenViewModelTy
         state.applyingChange = false
     }
 }
+
+extension UNUserNotificationCenter {
+    func authorizationStatus() async -> UNAuthorizationStatus {
+        await notificationSettings().authorizationStatus
+    }
+}
```
alfogrillo, line 217: why not a computed async property?
```swift
extension UNUserNotificationCenter {
    var authorizationStatus: UNAuthorizationStatus {
        get async {
            await notificationSettings().authorizationStatus
        }
    }
}
```

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -398,8 +398,10 @@ class ClientProxy: ClientProxyProtocol {
                 .withCrossProcessLock(appIdentifier: "MainApp")
                 .finish()
             let roomListService = syncService.roomListService()
-
-            let eventStringBuilder = RoomEventStringBuilder(stateEventStringBuilder: RoomStateEventStringBuilder(userID: userID))
+            
+            let roomMessageEventStringBuilder = RoomMessageEventStringBuilder(attributedStringBuilder: AttributedStringBuilder(permalinkBaseURL: .homeDirectory))
```
alfogrillo, line 402: I imagine the `permalinkBaseURL: .homeDirectory` is fake for now?

ElementX/Sources/Services/Room/RoomSummary/RoomEventStringBuilder.swift
```diff
@@ -19,70 +19,38 @@ import MatrixRustSDK
 
 struct RoomEventStringBuilder {
     private let stateEventStringBuilder: RoomStateEventStringBuilder
+    private let messageEventStringBuilder: RoomMessageEventStringBuilder
     
-    init(stateEventStringBuilder: RoomStateEventStringBuilder) {
+    init(stateEventStringBuilder: RoomStateEventStringBuilder, messageEventStringBuilder: RoomMessageEventStringBuilder) {
```
alfogrillo, line 24: What's the plan for non "message based" events like polls? We build another `EventStringBuilder` or we collapse them under `RoomMessageEventStringBuilder`? 
In fact they are still [message events](https://spec.matrix.org/v1.8/client-server-api/#types-of-room-events) but they don't have a message type.

NSE/Sources/NotificationServiceExtension.swift
```diff
@@ -85,10 +86,10 @@ class NotificationServiceExtension: UNNotificationServiceExtension {
                 MXLog.info("\(tag) no notification for the event, discard")
                 return discard()
             }
-
+
```
alfogrillo, line 89:
```suggestion

```

NSE/Sources/NotificationServiceExtension.swift
```diff
@@ -85,10 +86,10 @@ class NotificationServiceExtension: UNNotificationServiceExtension {
                 MXLog.info("\(tag) no notification for the event, discard")
                 return discard()
             }
-
+            
             // After the first processing, update the modified content
-            modifiedContent = try await itemProxy.process(mediaProvider: nil)
-
+            modifiedContent = try await notificationContentBuilder.content(for: itemProxy, mediaProvider: nil)
+
```
alfogrillo, line 92:
```suggestion

```

alfogrillo (approved):
LGTM, just left minor comments and doubts.

How hard is to add unit tests for `RoomEventStringBuilder`?
(maybe as a separate PR?)

ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu: Yep, it's irellevant for places where the text isn't interactive like the room list last message or notifications

ElementX/Sources/Services/Room/RoomSummary/RoomEventStringBuilder.swift
stefanceriu: I guess we should handle them separately in the `RoomMessageEventStringBuilder` and `NotificationContentBuilder`. Extracting out `MessageType` into its own builder was the best I could do at the moment as the other types are completely different (TimelineItem vs NotificationItem)

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/NotificationSettingsScreenViewModel.swift
stefanceriu: I didn't actually write it, just moved it from another file. Probably because you can't have that in a protocol, or at least I can't figure out how.

PR #1690, opened 2023-09-12
Fixes #1640 - Prevent loading skeletons from being dismissed too soon…

### LINKED ISSUE #1640: Loading skeleton in the room list doesn't work properly
### Steps to reproduce

1. Login fresh
2. Observe the loading skeleton in the room list
3. The loading skeleton goes away before the content is ready

https://github.com/vector-im/element-x-ios/assets/19324622/3c278444-4757-4855-b5dc-84cae7832f25

#### What did you expect?
The skeleton goes away **after** the content is ready

#### What happened instead?
The skeleton disappeared **before** the content was ready

### Your phone model

iPhone 14 simulator

### Operating system version

_No response_

### Application version

Commit fef3077f95b6e77f6e985fcb252c5160dd7fe760

### Homeserver

_No response_

### Will you send logs?

No

stefanceriu:
…. Avoid weird animations caused by the session verification banner.

### DIFF
--- a/ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
+++ b/ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
@@ -148,8 +148,10 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol 
             .sink { [weak self] state, rooms in
                 guard let self else { return }
                 
-                let isLoadingData = !state.isLoaded && rooms.isEmpty
-                let hasNoRooms = (state.isLoaded && state.totalNumberOfRooms == 0 && rooms.isEmpty)
+                // Rust sends back loaded even when no rooms are loaded yet, check room for empty to be sure
+                let isLoadingData = !state.isLoaded || (state.totalNumberOfRooms != 0 && rooms.isEmpty)
+                
+                let hasNoRooms = state.isLoaded && state.totalNumberOfRooms == 0 && rooms.isEmpty
                 
                 var roomListMode = self.state.roomListMode
                 if isLoadingData {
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -148,7 +148,9 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
             .sink { [weak self] state, rooms in
                 guard let self else { return }
                 
-                let isLoadingData = !state.isLoaded && rooms.isEmpty
+                // Rust sends back loaded even when no rooms are loaded yet, check room for empty to be sure
+                let isLoadingData = !state.isLoaded || rooms.isEmpty
```
pixlwave, line 152: This won't work, it will prevent `hasNoRooms` from being used to show the empty state. Does it work for this case in combination with `state.totalNumberOfRooms == 0`?

stefanceriu: Bah, of course 🤦 Might work, let me try it out

stefanceriu: Yep, that seems fine

Screenshot 2023-09-13 at 08 33 41

pixlwave (approved):
LGTM 👍

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
pixlwave: Nice, thanks.

PR #1691, opened 2023-09-12
Add polls analytics

alfogrillo:
Implements the poll related analytics added [here](https://github.com/matrix-org/matrix-analytics-events/pull/85/).
It also refactors the location analytics since the api has changed.


stefanceriu (approved):
lgtm 👍

Issue #1198, opened 2023-06-28
The only way to cancel sending a rageshake is to kill the app.


ara4n:
### Steps to reproduce

1. Send a rageshake on slowish data
2. Realise that it's going to take minutes to send the logs
3. Give up by pressing cancel
4. nothing happens.
5. the only way to stop it is to force-quit the app.

(As an extra bonus, perhaps due to https://github.com/vector-im/element-x-ios/issues/1140, if you tap on a push notif for the app while it's sending the rageshake, the rageshake uploader UI disappears.  looks like the RS completes in the background though)

#### What did you expect?

Cancel button should cancel

#### What happened instead?

Cancel button didn't cancel.

### Application version

257

### Will you send logs?

No

PR #1692, opened 2023-09-12
Fixe #1198 - Allow cancelling feeback submissions while they're uploa…

stefanceriu:
…ding

### DIFF
--- a/ElementX/Sources/Screens/BugReportScreen/BugReportScreenViewModel.swift
+++ b/ElementX/Sources/Screens/BugReportScreen/BugReportScreenViewModel.swift
@@ -24,7 +24,7 @@ class BugReportScreenViewModel: BugReportScreenViewModelType, BugReportScreenVie
     private let userID: String
     private let deviceID: String?
     private let actionsSubject: PassthroughSubject<BugReportScreenViewModelAction, Never> = .init()
-    private var uploadTask: Task<Void, Never>?
+    @CancellableTask private var uploadTask: Task<Void, Never>?
 
     var actions: AnyPublisher<BugReportScreenViewModelAction, Never> {
         actionsSubject.eraseToAnyPublisher()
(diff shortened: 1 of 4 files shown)


ElementX/Sources/Screens/BugReportScreen/View/BugReportScreen.swift
```diff
@@ -131,6 +132,7 @@ struct BugReportScreen: View {
                 context.send(viewAction: .submit)
             }
             .disabled(context.reportText.count < 5)
+            .disabled(context.viewState.shouldDisableInteraction)
```
alfogrillo, line 135: Shouldn't line 35 propagate to all the sections? 🤔

ElementX/Sources/Screens/BugReportScreen/BugReportScreenViewModel.swift
```diff
@@ -24,6 +24,7 @@ class BugReportScreenViewModel: BugReportScreenViewModelType, BugReportScreenVie
     private let userID: String
     private let deviceID: String?
     private let actionsSubject: PassthroughSubject<BugReportScreenViewModelAction, Never> = .init()
+    private var uploadTask: Task<Void, Never>?
```
alfogrillo, line 27: Maybe using `@CancellableTask` would be better since you can cancel the pending task just assigning a new one or nil.

alfogrillo (approved):
LGTM
just left a couple of minor comments

ElementX/Sources/Screens/BugReportScreen/BugReportScreenViewModel.swift
stefanceriu: Good idea 👍

ElementX/Sources/Screens/BugReportScreen/View/BugReportScreen.swift
alfogrillo: Right, I missed we were in the toolbar! 👍🏻

alfogrillo (approved):

PR #1697, opened 2023-09-13
Fixes #1342 - Remove custom last message for rooms that don't have one

### LINKED ISSUE #1342: Room list shows "Message" as the placeholder for a new room
### Steps to reproduce

1. Create a new room
2. Look at timeline

#### What did you expect?
Nothing or a placeholder that says "no messages"

#### What happened instead?
A message preview that says "Message"

IMG_0259

### Your phone model

iPhone 11

### Operating system version

16.5.1

### Application version

Testflight 1.2.1 (291)

### Homeserver

matrix.org

### Will you send logs?

No

stefanceriu:


ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
```diff
@@ -105,23 +105,7 @@ struct HomeScreenViewStateBindings {
 
 struct HomeScreenRoom: Identifiable, Equatable {
     static let placeholderLastMessage = AttributedString("Hidden last message")
```
pixlwave, line 107: Can probably make this non-static now FWIW.

pixlwave (approved):
LGTM 👌

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
stefanceriu: It's actually nice this way as the placeholder function is static too ¯\_(ツ)_/¯

PR #1700, opened 2023-09-13
Fixes #1699 - Introduce a new advanced settings screen

### LINKED ISSUE #1699: Show advanced settings for external users
### Story

As an external user that wants to configure their element client beyond the default setting I would love to access certain feature flags in a separated settings view.

- Create an "Advanced settings view" for external users. In the settings view.

The following options should be configurable:
- Message Layout (Bubbles vs. Modern)
-- Use the existing setting and move it here
- Enable Rich Text Editor
- - Use the existing setting from developer options and move it here @langleyd Good idea or should this be an other one?
- View event source option
- - Needs to be build. Enables to see the event source option on long pressing an event.

stefanceriu:
Screenshot 2023-09-13 at 16 52 45

Screenshot 2023-09-13 at 16 52 47

### DIFF
--- a/ElementX/Sources/Screens/Settings/AvancedOptionsScreen/View/AdvancedSettingsScreen.swift
+++ b/ElementX/Sources/Screens/Settings/AvancedOptionsScreen/View/AdvancedSettingsScreen.swift
@@ -32,7 +32,7 @@ struct AdvancedSettingsScreen: View {
                 ListRow(label: .default(title: L10n.actionViewSource, systemIcon: .mailAndTextMagnifyingglass),
                         kind: .toggle($context.viewSourceEnabled))
                 
-                ListRow(label: .default(title: L10n.commonRichTextEditor, systemIcon: .squareFillTextGrid1x2),
+                ListRow(label: .default(title: L10n.commonRichTextEditor, icon: Image(asset: Asset.Images.textFormat)),
                         kind: .toggle($context.richTextEditorEnabled))
             }
         }
--- a/ElementX/Sources/Screens/Settings/AvancedOptionsScreen/View/AdvancedSettingsScreen.swift
+++ b/ElementX/Sources/Screens/Settings/AvancedOptionsScreen/View/AdvancedSettingsScreen.swift
@@ -1,0 +1,65 @@
+//
+// Copyright 2022 New Vector Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Compound
+import SwiftUI
+
+struct AdvancedSettingsScreen: View {
+    @ObservedObject var context: AdvancedSettingsScreenViewModel.Context
+    @State private var showConfetti = false
+    
+    var body: some View {
+        Form {
+            Section {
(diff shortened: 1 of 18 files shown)


ElementX/Sources/Screens/Settings/AvancedOptionsScreen/View/AdvancedSettingsScreen.swift
```diff
@@ -0,0 +1,65 @@
...
+            Section {
+                ListRow(label: .default(title: L10n.commonMessageLayout,
+                                        systemIcon: .rectangleGrid1x2),
+                        kind: .picker(selection: $context.timelineStyle,
+                                      items: TimelineStyle.allCases.map { (title: $0.name, tag: $0) }))
+                
+                ListRow(label: .default(title: L10n.actionViewSource, systemIcon: .mailAndTextMagnifyingglass),
+                        kind: .toggle($context.viewSourceEnabled))
+                
+                ListRow(label: .default(title: L10n.commonRichTextEditor, systemIcon: .squareFillTextGrid1x2),
```
pixlwave, line 35: Seeing this symbol name, I think its more semantically intended for e.g. a list of articles with a thumbnail. Maybe we should use one of the `list…` symbols instead? But probably more a question for @amshakal.

ElementX/Sources/Screens/Settings/SettingsScreen/View/SettingsScreen.swift
```diff
@@ -113,48 +98,52 @@ struct SettingsScreen: View {
...
+            
+            if context.viewState.showDeveloperOptions {
+                ListRow(label: .default(title: L10n.commonDeveloperOptions,
+                                        systemIcon: .hammerCircle),
+                        kind: .navigationLink {
+                            context.send(viewAction: .developerOptions)
+                        })
+                        .accessibilityIdentifier(A11yIdentifiers.settingsScreen.developerOptions)
+            }
+        }
```
pixlwave, line 146: This shouldn't be inside `simplifiedSection`, can we extract it into its own property like the others?

pixlwave (approved):
Nice 👍

ElementX/Sources/Screens/Settings/AvancedOptionsScreen/View/AdvancedSettingsScreen.swift
stefanceriu: It's okay, we replaced it with the same icon we use in the attachments menu

ElementX/Sources/Screens/Settings/SettingsScreen/View/SettingsScreen.swift
stefanceriu: Ended up moving both to an `advancedOptionsSection`

ElementX/Sources/Screens/Settings/AvancedOptionsScreen/View/AdvancedSettingsScreen.swift
```diff
@@ -0,0 +1,65 @@
...
+    var body: some View {
+        Form {
+            Section {
+                ListRow(label: .default(title: L10n.commonMessageLayout,
+                                        systemIcon: .rectangleGrid1x2),
+                        kind: .picker(selection: $context.timelineStyle,
+                                      items: TimelineStyle.allCases.map { (title: $0.name, tag: $0) }))
+                
+                ListRow(label: .default(title: L10n.actionViewSource, systemIcon: .mailAndTextMagnifyingglass),
+                        kind: .toggle($context.viewSourceEnabled))
```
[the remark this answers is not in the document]
stefanceriu: We could but I would prefer to leave it alone

bmarty:
"View Source" wording is maybe a bit confusing here. Maybe something like "Enable View Source action", or "Be able to view the source of Event" would fit better. Not sure what is best (probably something else!). On Element Android, there is no such option, but when "Developer mode" is on, this action become visible. It is a bit better I think...

image

amshakal:
LGTM

amshakal (approved):

PR #1711, opened 2023-09-14
Enable OIDC token refresh in the NSE.

### LINKED ISSUE #1712: Enable token refresh in the NSE (and notifications for OIDC accounts).

pixlwave:
Updates the SDK to enable the cross-process lock for token refreshing to work in the NSE. Notifications are now allowed for accounts using OIDC.

Fixes #1712.

### DIFF
--- a/ElementX/Sources/Services/Client/ClientProxy.swift
+++ b/ElementX/Sources/Services/Client/ClientProxy.swift
@@ -86,8 +86,6 @@ class ClientProxy: ClientProxyProtocol {
         client.setDelegate(delegate: ClientDelegateWrapper { [weak self] isSoftLogout in
             self?.hasEncounteredAuthError = true
             self?.callbacks.send(.receivedAuthError(isSoftLogout: isSoftLogout))
-        } tokenRefreshCallback: { [weak self] in
-            self?.callbacks.send(.updateRestorationToken)
         })
         
         await configureAppService()
@@ -536,12 +534,9 @@ private class RoomListServiceSyncIndicatorListenerProxy: RoomListServiceSyncIndi
 
 private class ClientDelegateWrapper: ClientDelegate {
     private let authErrorCallback: (Bool) -> Void
-    private let tokenRefreshCallback: () -> Void
     
-    init(authErrorCallback: @escaping (Bool) -> Void,
-         tokenRefreshCallback: @escaping () -> Void) {
+    init(authErrorCallback: @escaping (Bool) -> Void) {
         self.authErrorCallback = authErrorCallback
-        self.tokenRefreshCallback = tokenRefreshCallback
     }
     
     // MARK: - ClientDelegate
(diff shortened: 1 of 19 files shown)


NSE/Sources/Other/NSEUserSession.swift
```diff
@@ -60,3 +72,15 @@ final class NSEUserSession {
         }
     }
 }
+
+private class ClientDelegateWrapper: ClientDelegate {
```
pixlwave, line 76: Added for the logs

ElementX/Sources/Services/Keychain/KeychainController.swift
```diff
@@ -86,4 +87,22 @@ class KeychainController: KeychainControllerProtocol {
             MXLog.error("Failed removing all tokens")
         }
     }
+    
+    // MARK: - ClientSessionDelegate
+    
+    #warning("Should there be an extra level of indirection here?")
```
pixlwave, line 93: I think I'm happy with the below methods living directly on the `KeychainController`, just flagging that here in case I've missed something.

stefanceriu: I too think this is fine

stefanceriu (approved):
lgtm 🚀

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -552,7 +547,6 @@ private class ClientDelegateWrapper: ClientDelegate {
     }
     
     func didRefreshTokens() {
-        MXLog.info("The session has updated tokens.")
-        tokenRefreshCallback()
+        MXLog.info("Delegating session updates to the ClientSessionDelegate.")
```
bnjbvr, line 550: As a small note: right now the client doesn't use the new `ClientSessionDelegate` mechanism if using the Matrix Auth scheme (i.e. not OIDC). Until it's properly supported, we might need to use both mechanisms: this for the Matrix Auth scheme, the `ClientSessionDelegate` for OIDC. Will fix asap in the SDK.

pixlwave: Ok perfect. It won't introduce any regressions anyway as we never handled token refreshed with Matrix Auth before we enabled OIDC anyway, so we'll merge this with the expectation of updating the SDK later today.

ElementX/Sources/Services/Notification/Proxy/NotificationItemProxyProtocol.swift
```diff
@@ -65,7 +65,7 @@ extension NotificationItemProxyProtocol {
                 return false
             case let .messageLike(content):
                 switch content {
-                case let .roomMessage(messageType):
+                case let .roomMessage(messageType, _):
```
pixlwave, line 68: Leaving the `inReplyTo` property unused for now.

NSE/Sources/NotificationContentBuilder.swift
```diff
@@ -36,7 +36,7 @@ struct NotificationContentBuilder {
             switch try? event.eventType() {
             case let .messageLike(content):
                 switch content {
-                case .roomMessage(messageType: let messageType):
+                case .roomMessage(let messageType, _):
```
pixlwave, line 39: Same here.

PR #1721, opened 2023-09-15
Fix room notification mode decoration with unread indicator

nimau:
This PR fixes the following issue:

If the notification mode is set to `.mentionsAndKeywordsOnly` or `.mute` while a room has unread notifications, we should not display the unread indicator.

`.allMessages` with unread notifications:
Screenshot 2023-09-15 at 14 30 26

`.mentionsAndKeywordsOnly` with unread notifications:
Screenshot 2023-09-15 at 14 36 01

`.mute` with unread notifications:
Screenshot 2023-09-15 at 14 30 52


pixlwave (approved):
Looks good to me, although I'm not sure I fully understand when a room would be muted but still receive notifications.

nimau:
I'm closing this PR as it has been decided that it will be corrected in another way.

PR #1731, opened 2023-09-18
Fix #1728 - Rooms not loading in offline mode because the room list s…

### LINKED ISSUE #1728: Roomlist won't load if the client is offline
### Steps to reproduce

1. Enable airplane mode
2. Launch app
3. Expect roomlist to load from room cache - it doesn't.

#### What did you expect?

Now we have a room cache, i'd expect the roomlist to load from it to give basic offline support (even if it can only show the most recent msg in each timeline).

#### What happened instead?

Blank room list.

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

386

### Homeserver

_No response_

### Will you send logs?

No

stefanceriu:
…ervice doesn't go into .loaded


stefanceriu:
This PR is actually incomplete and introduces other problems, as did the change previous to it. 

Will handle it separately as part of a new SDK release and https://github.com/matrix-org/matrix-rust-sdk/pull/2564/

PR #1732, opened 2023-09-18
Avoid publishing changes from a background thread, dispatch upload ta…

stefanceriu:
…sk handles on the main queue


pixlwave (approved):
LGTM. Did you consider using `@MainActor` on the closure instead of `@escaping` + manual dispatches?

stefanceriu:
> LGTM. Did you consider using `@MainActor` on the closure instead of `@escaping` + manual dispatches?

No, never crossed my mind, but I switched to it and works great 👏

PR #1733, opened 2023-09-18
Update room notification settings mismatch screen

nimau:
This PR updates the room notification settings mismatch screen to match the updated design.

Before:

After:


stefanceriu (approved):
lgtm 👍

PR #1735, opened 2023-09-18
Dynamic room summary provider limits

stefanceriu:
* adopts the new limited room list api which limits both the number of rooms display but also the diffs that come in
* fixes room list state handling after rust changes
* removes debouncing from searching as everything is just faster now and there's no point in artificially delaying it
* adopts new room member apis
* adopts new sync indicator delays

### DIFF
--- a/ElementX/Sources/Services/Client/ClientProxy.swift
+++ b/ElementX/Sources/Services/Client/ClientProxy.swift
@@ -507,7 +507,7 @@ class ClientProxy: ClientProxyProtocol {
     }
     
     private func createRoomListLoadingStateUpdateObserver(_ roomListService: RoomListService) -> TaskHandle {
-        roomListService.syncIndicator(listener: RoomListServiceSyncIndicatorListenerProxy { [weak self] state in
+        roomListService.syncIndicator(delayBeforeShowingInMs: 1000, delayBeforeHidingInMs: 0, listener: RoomListServiceSyncIndicatorListenerProxy { [weak self] state in
             guard let self else { return }
             
             switch state {
--- a/ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
+++ b/ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
@@ -334,10 +334,8 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
             changes.append(.remove(offset: Int(index), element: summary, associatedWith: nil))
             changes.append(.insert(offset: Int(index), element: summary, associatedWith: nil))
         case .truncate(let length):
-            let entriesToKeep = rooms.prefix(Int(length))
-        
             for (index, value) in rooms.enumerated() {
-                if entriesToKeep.contains(value) {
+                if index < length {
                     continue
                 }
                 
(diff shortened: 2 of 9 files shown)


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -457,7 +457,7 @@ class ClientProxy: ClientProxyProtocol {
     }
     
     private func createRoomListLoadingStateUpdateObserver(_ roomListService: RoomListService) -> TaskHandle {
-        roomListService.syncIndicator(listener: RoomListServiceSyncIndicatorListenerProxy { [weak self] state in
+        roomListService.syncIndicator(delayBeforeShowingInMs: 1000, delayBeforeHidingInMs: 0, listener: RoomListServiceSyncIndicatorListenerProxy { [weak self] state in
```
pixlwave, line 460: Maybe put 100/200ms delay on the hiding too just so it avoids the awkward shows and immediately hides type animations?

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -124,15 +126,21 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
                 MXLog.error("Failed updating visible range with error: \(error)")
             }
         }
+        
+        if range.upperBound >= rooms.count {
```
pixlwave, line 130: I have no idea but do we need to add a `guard rooms.count > 1` here or something? Also, if I only have 20 rooms, will it keep calling `addOnePage` each time I scroll to the bottom, and will that do anything bad?

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -309,22 +328,21 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
...
+        case .set(let index, let value):
+            MXLog.verbose("\(name): Update \(value.debugIdentifier) at \(index)")
+            let summary = buildSummaryForRoomListEntry(value)
+            changes.append(.remove(offset: Int(index), element: summary, associatedWith: nil))
+            changes.append(.insert(offset: Int(index), element: summary, associatedWith: nil))
+        case .truncate(let length):
+            let entriesToKeep = rooms.prefix(Int(length))
+        
             for (index, value) in rooms.enumerated() {
+                if entriesToKeep.contains(value) {
```
pixlwave, line 340: I wonder if `index < length` would be a faster check to make?

pixlwave (approved):
LGTM 👍

ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu: I copied what rust used to do (minus the 200 delayBeforeShowingInMs) which I didn't notice cause any issues

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
stefanceriu: From my testing no, it doesn't have any effect and I would prefer to keep it simple than risk any extra logic

PR #1736, opened 2023-09-18
Use TextKit 2 for MessageText

Velin92:
Since we were accessing the old NSLayoutManager instead of the new NSTextLayoutManager, we were forcing the TextView to be refreshed as a TextKit 1 version instead of the TextKit 2 version (which is more performant)

This removes the warning and should increase the performance slightly


stefanceriu (approved):
👏

PR #1742, opened 2023-09-18
Improve message edit with RTE

alfogrillo:
This PR changes the edit behavior to fit the current rich text editor flag:

- RTE on: the editing text in the composer looks visually formatted
- RTE off: the editing text in the composer is presented as raw markdown (if there is style applied)

| RTE on | RTE off|
| --- | --- |
| on|off|


ElementX/Sources/Services/Timeline/TimelineItems/Items/Messages/TextRoomTimelineItemContent.swift
```diff
@@ -19,4 +19,6 @@ import UIKit
 struct TextRoomTimelineItemContent: Hashable {
     let body: String
     var formattedBody: AttributedString?
+    // The textual representation of the formatted body (usually HTML code)
```
pixlwave, line 22: Not suggesting to word it exactly like this, but I find this helpful clarify the meaning for me:
```suggestion
    // The original textual representation of the formatted body directly from the event (usually HTML code)
```

ElementX/Sources/Services/Timeline/TimelineItems/Items/Messages/TextRoomTimelineItemContent.swift
```diff
@@ -19,4 +19,6 @@ import UIKit
...
+    var rawFormattedBody: String?
```
pixlwave, line 23: Actually, thinking about this more, calling it `htmlBody` or `formattedBodyHTMLString` might be even easier to approach, given we use these elsewhere and the spec doesn't allow any other format right now anyway.

pixlwave (approved):
👍

ElementX/Sources/Services/Timeline/TimelineItems/Items/Messages/TextRoomTimelineItemContent.swift
alfogrillo: It looks sensible to me.
Applied [here](https://github.com/vector-im/element-x-ios/pull/1742/commits/61d328c0ea8bd8e97e4360f986127a15db85a20c)

PR #1743, opened 2023-09-18
Prefire basic implementation

Velin92:
For now this includes a basic but tweakable configuration of Prefire with one single Preview test.

- The preview tests run in our existing UnitTests suite
- They assume that the tests will always be run on an iPhone 14 light mode with iOS 16.

We probably need to decide if we want to have a different test suite just for the preview testing or if we are okay running unit tests with this configuration (iPhone 14 iOS 16) on the CI and also locally always.

After this is merged I'll proceed into making another PR were I make all our Previews testable, which should bump also the code coverage by a lot and increase resilience to breaking view changes


.prefire.yml
```diff
@@ -0,0 +1,3 @@
+test_configuration:
+  - test_file_path: UnitTests/Sources/PreviewTests.generated.swift
```
pixlwave, line 2: Have you added this path to Git LFS?

UnitTests/Sources/__Snapshots__/PreviewTests.generated/test_timelineItemStyler.Plain.png
pixlwave, line 1: Could we add some horizontal padding to this preview?

UnitTests/SupportingFiles/target.yml
```diff
@@ -44,6 +49,8 @@ targets:
 
     sources:
     - path: ../Sources
+      excludes:
+        - "**/__Snapshots__/**"
```
pixlwave, line 53: Have you added this path to Git LFS?

.github/workflows/unit_tests.yml
```diff
@@ -41,6 +41,16 @@ jobs:
 
       - name: Run tests
         run: bundle exec fastlane unit_tests
+        
+      - name: Archive artifacts
+        uses: actions/upload-artifact@v3
+        # We only care about artifcats if the tests fail
+        if: failure()
```
stefanceriu, line 48: 👌

.swiftlint.yml
```diff
@@ -47,6 +47,9 @@ cyclomatic_complexity:
 nesting:
   type_level:
     warning: 5
+    
+type_name:
+  allowed_symbols: "_"
```
stefanceriu, line 52: Can we maybe add this as an exception in that generated file?

stefanceriu (approved):
Looks amazing, awesome job! 👏

.swiftlint.yml
Velin92: oh is not for the generated file, sadly is for all the previews because this exception is essentially working by default from swiftlint only if the` X_Preview` struct is deriving from `PreviewProvider`, since I made a custom protocol that merges both `PreviewProvider` and `PrefireProvider ` (to avoid importing Prefire everywhere in the code) that exception does not work anymore, so we need to force it everywhere here.

fastlane/Fastfile
```diff
@@ -77,8 +77,11 @@ end
 lane :unit_tests do
   run_tests(
     scheme: "UnitTests",
+    device: 'iPhone 14 (16.4)',
+    ensure_devices_found: true,
     result_bundle: true,
     number_of_retries: 3,
+    xcargs: '-skipPackagePluginValidation',
```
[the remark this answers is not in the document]
Velin92: required to make prefire plugin work on CI, however since we run this on GH CI, and not on an internal CI system I am pretty sure we don't need to worry too much about security issues for GH systems.

PR #1749, opened 2023-09-19
Media upload fixes

stefanceriu:


alfogrillo (approved):
LGTM!

Just a curiosity: in which case the current codebase failed to work properly?

stefanceriu:
> Just a curiosity: in which case the current codebase failed to work properly?

Files with whitespaces in the name 😅 The size thing was just an extra precaution on top.

Issue #1707, opened 2023-09-13
Low hanging improvements to the message send, edit & reaction animations


nadonomy:
I understand we're waiting on iOS17/some SwiftUI updates to properly update the timeline animations. However, from recording and slow-mo debugging, I think we can materially improve what we have today just by tweaking tween values, easing and overall choreography with todays implementation.

I'd like us to timebox a working session to see how much we can improve this, ahead of recording launch materials.

pixlwave:
Closed by https://github.com/vector-im/element-x-ios/pull/1751

PR #1751, opened 2023-09-19
Change timeline animations to use .top (flipped).

pixlwave:
This PR changes the timeline animations (which are only applied when the last message ID changes).

**Before**

https://github.com/vector-im/element-x-ios/assets/6060466/1084f8f6-ef69-44b7-bdb9-9101049c4ec3

**After**

https://github.com/vector-im/element-x-ios/assets/6060466/00571e04-8e7a-4209-a1cf-c9531cd1ac95


stefanceriu (approved):
👏

pixlwave:
There is potentially some future tweaks we can make to the `TimelineItemStatusView` for further polish, but this solves #1707 well-enough for the video.

PR #1757, opened 2023-09-20
Member optimisations

stefanceriu:
* Prevent mapping members on the main queue inside the room screen view model
* Avoid too much back and forth through FFI by lazily fetching and storing member proxy properties


Velin92 (approved):
LGTM

PR #1765, opened 2023-09-20
Add polls "creator view"

alfogrillo:
This PR adds a customization in the timeline for polls.
The customization for a given poll is visible just for the user that started that poll.

**Result**
Simulator Screenshot - iPhone 15 - 2023-09-20 at 15 04 37


Velin92 (approved):
LGTM, however could you wait this https://github.com/vector-im/element-x-ios/pull/1773 to be merged first? Since the previews will now be generated in a different folder, so you may need to move them into the new folder or just regenerate them.

PR #1769, opened 2023-09-20
Release script fixes: prevent sed from incorrectly matching other ver…

stefanceriu:
…sions, correctly increment the path version past .9 which would previously increment the minor

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -1,4 +1,5 @@
 require 'yaml'
+require 'semantic'
 require_relative 'changelog'
 
 before_all do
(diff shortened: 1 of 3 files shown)


fastlane/Fastfile
```diff
@@ -256,11 +257,14 @@ lane :prepare_next_release do
...
   current_version = data["settings"]["MARKETING_VERSION"]
-  new_version = current_version.next
+
+  version = Semantic::Version.new(current_version)
+
+  new_version = version.increment!(:patch)
   
   # Bump the patch version. The empty string after -i is so that sed doesn't 
   # create a backup file on macOS
-  sh("sed -i '' 's/#{current_version}/#{new_version}/g' #{target_file_path}")
```
pixlwave, line 263: Ohhhh lol I see what was happening 😆

pixlwave (approved):
😎

fastlane/Fastfile
stefanceriu: 🤫

PR #1773, opened 2023-09-20
Prefire compatibility with Xcode 15

Velin92:
- This will generate the tests in the DerivedData (since in Xcode 15 IS NOT possible to have SPM plugins generate code in the project)
- I also improved the stencil, it won't fail when it does not find the file, it will instead succeed and generate the file like in UI tests
- I also added a check that will make the preview tests fail automatically if we run them with a dark mode device.
- Sadly all the previews files had to be moved into another folder since they are now generated in a different way since we are using the derived datas
- Also added some delay to some views that require some loading time like the home screen
- The bloom effect is broken in snapshot tests since is generated through introspect which creates some issues when doing some specific system UI customisations like in that case, but is not a big deal, since that instead works in UI tests)


Tools/Prefire/PreviewTests.stencil
```diff
@@ -42,8 +42,8 @@ class PreviewTests: XCTestCase {
     private func assertSnapshots(matching preview: _Preview, testName: String = #function) {
         let isScreen = preview.layout == .device
         let device = preview.device?.snapshotDevice() ?? deviceConfig
-        var delay: TimeInterval = 1.0
-        var precision: Float = 0.99
+        var delay: TimeInterval = 0
+        var precision: Float = 1
```
alfogrillo, line 46: Is it safe? I remember that in UI tests we have less than 1

alfogrillo (approved):
LGTM!

Tools/Prefire/PreviewTests.stencil
Velin92: yeah they just passed without a problem, the thing is that I discovered that these value do not matter, these are customisable by adding 
.snapshot(delay: precision:) at the bottom of a preview to specify a custom precision/delay.

Also I still kept the perceptualPrecision at 0.98 below in the stencil which is what really matters.

PR #1777, opened 2023-09-21
Replace timeline animation delay with a check on the item count.

pixlwave:
The delay was to avoid animations on initial timeline load. Now that we only animate when a new message arrives at the bottom, combining that with a check for an item count > 0 appears to be enough.

https://github.com/vector-im/element-x-ios/assets/6060466/ee064c2d-3db8-45cd-ab6b-45da38dd17a1


stefanceriu (approved):
👏

PR #1783, opened 2023-09-22
Fix various flakey unit tests

stefanceriu:
This PR tries to address unit test flakiness. It does so by replacing the old `nextViewState` and `deferFulfillment` with a block based `deferFulfillment` that should give the user more control over the outcome. 

It also removes the automatic fastlane unit test retry to avoid hiding problematic tests.

Not done:

- [ ] Cleanup old tests that use plain expectations
- [ ] Cleanup tests that just assert the old thing or look at the UI
- [ ] Make the UnitTests target throw proper syntax errors instead of failing within the swiftlint or swiftformat main target phases


ElementX/Sources/Screens/RoomMemberListScreen/RoomMembersListScreenViewModel.swift
```diff
@@ -83,7 +83,8 @@ class RoomMembersListScreenViewModel: RoomMembersListScreenViewModelType, RoomMe
             self.members = members
             self.state = .init(joinedMembersCount: roomProxy.joinedMembersCount,
                                joinedMembers: roomMembersDetails.joinedMembers,
-                               invitedMembers: roomMembersDetails.invitedMembers)
+                               invitedMembers: roomMembersDetails.invitedMembers,
+                               bindings: state.bindings)
```
pixlwave, line 87: Good spot

UnitTests/Sources/AnalyticsSettingsScreenViewModelTests.swift
```diff
@@ -24,6 +24,10 @@ class AnalyticsSettingsScreenViewModelTests: XCTestCase {
     private var viewModel: AnalyticsSettingsScreenViewModelProtocol!
     private var context: AnalyticsSettingsScreenViewModelType.Context!
     
+    override func tearDown() {
```
pixlwave, line 27: Unimportant, but shouldn't this come after setup? Its like writing deinit before init to me.

UnitTests/Sources/EmojiProviderTests.swift
```diff
@@ -19,45 +19,53 @@ import XCTest
 @testable import ElementX
 
 final class EmojiProviderTests: XCTestCase {
```
pixlwave, line 21: Maybe the tests should be `@MainActor` now that `EmojiProvider.init` is called per test?

UnitTests/Sources/EmojiProviderTests.swift
```diff
@@ -19,45 +19,53 @@ import XCTest
...
         let item = EmojiItem(label: "test", unicode: "test", keywords: ["1", "2"], shortcodes: ["1", "2"], skins: ["🙂"])
         let category = EmojiCategory(id: "test", emojis: [item])
+        
+        let emojiLoaderMock = EmojiLoaderMock()
         emojiLoaderMock.categories = [category]
-        let categories = await sut.getCategories()
+        
+        let emojiProvider = await EmojiProvider(loader: emojiLoaderMock)
+        
+        let categories = await emojiProvider.getCategories()
```
pixlwave, line 31: Whilst we're here, can we do?
```suggestion
        let categories = await emojiProvider.categories()
```

fastlane/Fastfile
```diff
@@ -81,7 +81,6 @@ lane :unit_tests do
     device: 'iPhone 14 (16.4)',
     ensure_devices_found: true,
     result_bundle: true,
-    number_of_retries: 3,
```
pixlwave, line 84: 👏

UnitTests/Sources/RoomScreenViewModelTests.swift
```diff
@@ -309,7 +312,7 @@ class RoomScreenViewModelTests: XCTestCase {
         // Test
         viewModel.context.send(viewAction: .retrySend(itemID: .init(timelineID: UUID().uuidString, transactionID: "test retry send id")))
         await Task.yield()
```
pixlwave, line 314: I think this one can probably be removed 🙈

UnitTests/Sources/NotificationSettingsEditScreenViewModelTests.swift
```diff
@@ -74,21 +77,28 @@ class NotificationSettingsEditScreenViewModelTests: XCTestCase {
...
-        let pendingModes = try await deferredViewState.fulfill()
         
-        XCTAssertEqual(pendingModes, [nil, .allMessages, nil])
+        try await deferredViewState.fulfill()
+        
+        deferredViewState = deferFulfillment(viewModel.context.$viewState) { state in
+            state.pendingMode == nil
+        }
         
+        try await deferredViewState.fulfill()
```
pixlwave, line 100: This happens a few times in this file. Shouldn't it always go

- create deferred fulfilment
- trigger some change
- await fulfilment

Otherwise it all feels inherently racey.

UnitTests/Sources/RoomDetailsViewModelTests.swift
```diff
@@ -126,36 +131,56 @@ class RoomDetailsScreenViewModelTests: XCTestCase {
                                                mediaProvider: MockMediaProvider(),
                                                userIndicatorController: ServiceLocator.shared.userIndicatorController,
                                                notificationSettingsProxy: NotificationSettingsProxyMock(with: NotificationSettingsProxyMockConfiguration()))
-        await context.nextViewState()
+        
+        let deferred = deferFulfillment(viewModel.context.$viewState) { state in
+            state.dmRecipient != nil
+        }
+        
+        try await deferred.fulfill()
```
pixlwave, line 139: Same thought here (and below).

UnitTests/Sources/RoomDetailsViewModelTests.swift
```diff
@@ -222,20 +273,33 @@ class RoomDetailsScreenViewModelTests: XCTestCase {
...
+        deferred = deferFulfillment(context.$viewState) { state in
+            state.isProcessingIgnoreRequest == false
+        }
+
+        try await deferred.fulfill()
+        
         XCTAssert(context.viewState.dmRecipient?.isIgnored == true)
         XCTAssertNotNil(context.alertInfo)
     }
-
```
pixlwave, line 238: Maybe re-indent this file?

UnitTests/Sources/NotificationSettingsScreenViewModelTests.swift
```diff
@@ -27,41 +27,41 @@ class NotificationSettingsScreenViewModelTests: XCTestCase {
     private var userSession: UserSessionProtocol!
     private var userNotificationCenter: UserNotificationCenterMock!
     private var notificationSettingsProxy: NotificationSettingsProxyMock!
-
```
pixlwave, line 30: Again re-indent this file probably :)

UnitTests/Sources/NotificationSettingsScreenViewModelTests.swift
```diff
@@ -140,33 +150,40 @@ class NotificationSettingsScreenViewModelTests: XCTestCase {
...
+
         context.send(viewAction: .fixConfigurationMismatchTapped)
-        let fixingStates = try await deferredState.fulfill()
         
-        XCTAssertEqual(fixingStates, [false, true, false])
+        try await deferred.fulfill()
         
+        deferred = deferFulfillment(viewModel.context.$viewState) { state in
+            state.fixingConfigurationMismatch == false
+        }
```
pixlwave, line 175: I see this pattern a lot. Do you think we could solve for it inside of `deferFulfillment`?

UnitTests/Sources/RoomNotificationSettingsScreenViewModelTests.swift
```diff
@@ -22,99 +22,118 @@ import XCTest
...
-        context.send(viewAction: .changedAllowCustomSettings)
-        let states = try await deferredIsRestoringDefaultSettings.fulfill()
-        XCTAssertEqual(states, [false, true, false])
+        viewModel.context.send(viewAction: .changedAllowCustomSettings)
+        
+        let deferredIsRestoringDefaultSettings = deferFulfillment(viewModel.context.$viewState) { state in
+            state.isRestoringDefaultSetting == false
+        }
+        
+        try await deferredIsRestoringDefaultSettings.fulfill()
```
pixlwave, line 109: More instances in here that don't follow the create deferred fulfilment, trigger change, await fulfilment.

pixlwave (commented):
Very nice, glad to see this. Some thoughts inline :)

UnitTests/Sources/AnalyticsSettingsScreenViewModelTests.swift
stefanceriu: That's exactly how both of them should be 😅 This is probably the ObjC developer in me but the idea is that you can have a ton of inits but just one deinit. It makes sense for it to come on top of everything.

UnitTests/Sources/RoomNotificationSettingsScreenViewModelTests.swift
stefanceriu: Yep yep, sorry about that. I change the way I was going about it while I was fixing the tests and forgot to come back and fix these.

UnitTests/Sources/NotificationSettingsScreenViewModelTests.swift
stefanceriu: Honestly I thought about it but I'm not entirely sure how to make it happen. I figured I'm going to write all of them to explicitly do what you expect and then add niceties on top later. I'm thinking some sort of `deferFulfillment(publisher, keyPath, untilOrderReached: [States]` but I need to have a proper play with it

UnitTests/Sources/RoomDetailsViewModelTests.swift
stefanceriu: This one is just waiting for the members to load. Not much I can do about it as I have no other entry point

PR #1785, opened 2023-09-22
Fix for reversed placeholder on notifications on iOS 17

Velin92:
Fix to avoid reversing placeholders on iOS 17 and above

### DIFF
--- a/ElementX/Sources/Other/Extensions/UNNotificationContent.swift
+++ b/ElementX/Sources/Other/Extensions/UNNotificationContent.swift
@@ -19,6 +19,8 @@ import Intents
 import SwiftUI
 import UserNotifications
 
+import Version
+
 struct NotificationIcon {
     struct GroupInfo {
         let name: String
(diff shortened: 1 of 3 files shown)


ElementX/Sources/Other/Extensions/UNNotificationContent.swift
```diff
@@ -185,7 +187,7 @@ extension UNMutableNotificationContent {
 
     private func getPlaceholderAvatarImageData(name: String, id: String) async -> Data? {
         // The version value is used in case the design of the placeholder is updated to force a replacement
-        let fileName = "notification_placeholderV2_\(name)_\(id).png"
+        let fileName = "notification_placeholderV3_\(name)_\(id).png"
```
pixlwave, line 190: This one won't quite be that simple. We would need to only use V3 if the device is on iOS 17 as not everyone has upgraded yet.

Velin92: uhm why? we just invalidate caching for everyone from the next release

Velin92: done

pixlwave (approved):

PR #1791, opened 2023-09-25
Element call tweaks

stefanceriu:
* Only allow `https` call links to be passed through the custom app scheme
* Pass in custom call URL query parameters
* Add more tests for the AppRouteURLParser


Velin92 (approved):
LGTM

PR #1792, opened 2023-09-25
Fix replying in UI tests.

pixlwave:
The mocks were missing an `eventID` so all considered to be local echoes. This PR fixes that.

UI tests run is here: https://github.com/vector-im/element-x-ios/actions/runs/6297508527


stefanceriu (approved):
👏

PR #1795, opened 2023-09-25
remove `reply in thread` in the swipe action

Velin92:
Requested by @callumu in https://github.com/vector-im/element-x-android/issues/1422#event-10462995639


ElementX/Sources/Screens/RoomScreen/View/Style/SwipeToReplyView.swift
```diff
@@ -23,12 +23,6 @@ struct SwipeToReplyView: View {
         VStack(spacing: 4) {
```
pixlwave, line 23: I think you can remove the `VStack` now too?

pixlwave (approved):
👍

PR #1800, opened 2023-09-26
Room list searching tweaks

### LINKED ISSUE #1772: Flash of incorrect room items when filtering
### Steps to reproduce

1. Focus the "search" input to filter the room list
2. Type the letter "Z"

#### What did you expect?

For the items that match `Z` to appear

#### What happened instead?

The unfiltered room list appears for a split seconds, then gets replaced by the filtered down version
Can reproduce both on iOS and on Desktop

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

_No response_

### Homeserver

_No response_

### Will you send logs?

No

stefanceriu:
This PR improves a couple of things around the room list:

1) It starts using the new .none filter so that when the search bar is focused we stop displaying any rooms but also avoid flashing them before the real search results come in - Fixes #1772 
2) Adopts the new async room fetching API and stops the app from crashing on rust dbg builds
3) It removes the reset diff chunking "tweak" as it's no longer necessary

### DIFF
--- a/ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
+++ b/ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
@@ -88,7 +88,17 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol 
             .map(\.bindings.searchQuery)
             .removeDuplicates()
             .sink { [weak self] searchQuery in
-                self?.roomSummaryProvider?.updateFilterPattern(searchQuery)
+                guard let self else { return }
+                updateFilter(isSearchFieldFocused: state.bindings.isSearchFieldFocused, searchQuery: searchQuery)
+            }
+            .store(in: &cancellables)
+        
+        context.$viewState
+            .map(\.bindings.isSearchFieldFocused)
+            .removeDuplicates()
+            .sink { [weak self] isSearchFieldFocused in
+                guard let self else { return }
+                updateFilter(isSearchFieldFocused: isSearchFieldFocused, searchQuery: state.bindings.searchQuery)
             }
             .store(in: &cancellables)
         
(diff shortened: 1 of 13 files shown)


ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -88,7 +88,17 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
...
+            }
+            .store(in: &cancellables)
+        
+        context.$viewState
+            .map(\.bindings.isSearchFieldFocused)
+            .removeDuplicates()
+            .sink { [weak self] isSearchFieldFocused in
+                guard let self else { return }
+                updateFilter(isSearchFieldFocused: isSearchFieldFocused, searchQuery: state.bindings.searchQuery)
             }
```
alfogrillo, line 102: Maybe we can use "combine latest" and build a single chain here?

alfogrillo (approved):
LGTM!

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
stefanceriu: I would but I have no idea how to chain 2 of those `Publishers.MapKeyPath` thingies 🤔

stefanceriu: I'm going to merge this PR now as I have another one that depends on it. Let's fix this later if you know how.

alfogrillo: Sure, np.
If you want to do this next, you can do it like this:

```swift
        let isSearchFieldFocusedPublisher = context.$viewState.map(\.bindings.isSearchFieldFocused).removeDuplicates()
        let searchQueryPublisher = context.$viewState.map(\.bindings.searchQuery).removeDuplicates()

        isSearchFieldFocusedPublisher
            .combineLatest(searchQueryPublisher)
            .sink { [weak self] in
                updateFilter(isSearchFieldFocused: $0, searchQuery: $1)
            }
            .store(in: &cancellables)
```

stefanceriu: Oh I see, my mistake was not keeping the `removeDuplicates` which resulted in that weird MapKeyPath 👍

PR #1801, opened 2023-09-26
Prevent ElementX from intercepting links that it can't handle. Redire…

### LINKED ISSUE #1794: When installed as a macOS app, EXI intercepts EC URLs, but can't render them.
### Steps to reproduce

1. Install EXI as a macOS app
2. Click on an call.element.io link in your calendar
3. It opens in EXI, which say "WebRTC unavailable"

#### What did you expect?

Either EX should work on macOS (given Safari supports WebRTC!) - or we shouldn't try to intercept EC URLs on macOS until it does.

#### What happened instead?

Unexpected breakage.  Given we're not stopping users from installing EXI on macOS (and nor should we, as it works really very well, especially relative to the lumbering beast that is EW), i think we should at least support basic use cases (e.g. not breaking EC calls)

### Your phone model

Macbook Pro :P

### Operating system version

macOS Ventura 13.6 (22G120)

### Application version

393

### Homeserver

_No response_

### Will you send logs?

No

stefanceriu:
…ct them to the system browser

Fixes #1794

### DIFF
--- a/ElementX/Sources/Application/Application.swift
+++ b/ElementX/Sources/Application/Application.swift
@@ -19,6 +19,7 @@ import SwiftUI
 @main
 struct Application: App {
     @UIApplicationDelegateAdaptor(AppDelegate.self) private var applicationDelegate
+    @Environment(\.openURL) private var openURL
     private let appCoordinator: AppCoordinatorProtocol
 
     init() {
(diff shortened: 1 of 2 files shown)


pixlwave (approved):
Nice 👌

PR #1802, opened 2023-09-26
Add support for replying to events different than room messages

stefanceriu:
This PR adopts the new `canBeRepliedTo` timeline item property and starts using the `InReplyToDetails` that are now available for events other than room messages


Velin92 (approved):
LGTM, however how some of these messages will look when replied to in the reply view?

stefanceriu:
> LGTM, however how some of these messages will look when replied to in the reply view?

They will look like a reply to a text in which we use parts of their body e.g. for polls the question

PR #1827, opened 2023-09-28
Update strings that will be removed.

pixlwave:
Syncing up with https://github.com/vector-im/element-x-android/issues/1356 which will remove some strings, this PR handles any that affect iOS.


alfogrillo (approved):
LGTM

PR #1834, opened 2023-09-29
@room pill

Velin92:
This also introduces the red color for when the mentions is of self, and adds more testing.

Simulator Screenshot - iPhone 14 - 2023-09-29 at 14 13 54


stefanceriu (approved):
Looks great! 👏

PR #1835, opened 2023-09-29
Fixed a bug that prevented the mentions to be shown as plain text in the room list cell

Velin92:
This was caused by the fact that we render them as Plain text only, so we need to use a different mention builder for that case.
However at the same time we also needed to split the cache so that the room list cache does not invalidate or override the default cache.


stefanceriu (approved):
lgtm 👍

Issue #1107, opened 2023-06-18
No /me command


ara4n:
### Steps to reproduce

1. Try to use `/me`
2. `/me` sent literally

#### What did you expect?

`/me` should work.

#### What happened instead?

We can display emotes, but not send them.

### Application version

242

### Will you send logs?

No

PR #1841, opened 2023-10-01
implement emotes (/me)

ara4n:
requires https://github.com/matrix-org/matrix-rust-sdk/pull/2648

closes https://github.com/vector-im/element-x-ios/issues/1107

Arguably this could be handled by a slash-command engine in matrix-sdk-ui, but given it's not clear EX is going to even have slash commands (given whatsapp & imessage & signal etc don't), I'm adding it as a quick & dirty easter egg in the application layer. Given the logic is so simple, I'm not convinced it's worth the complexity of shoving it into the SDK at this point.

Handles:
 * [x] plaintext/markdown msg which begins `/me `
 * [x] richtext editor msg whose plaintext/markdown representation begins `/me `
 * [x] editing existing plaintext msgs
 * [x] editing existing richtext msgs
 * [x] replying with a /me (plaintext)
 * [x] replying with a /me (richtext)

### Pull Request Checklist

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
+++ b/ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
@@ -21,7 +21,6 @@ import SwiftUI
 
 typealias RoomScreenViewModelType = StateStoreViewModel<RoomScreenViewState, RoomScreenViewAction>
 
-// swiftlint:disable type_body_length
 class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol {
     private enum Constants {
         static let backPaginationEventLimit: UInt = 20
@@ -888,8 +887,6 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol 
         timelineController.playbackViewState(for: itemID)
     }
 }
-
-// swiftlint:enable type_body_length
 
 private extension RoomProxyProtocol {
     /// Checks if the other person left the room in a direct chat
--- a/ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
+++ b/ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
@@ -21,6 +21,7 @@ import SwiftUI
 
 typealias RoomScreenViewModelType = StateStoreViewModel<RoomScreenViewState, RoomScreenViewAction>
 
+// swiftlint:disable type_body_length
 class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol {
     private enum Constants {
         static let backPaginationEventLimit: UInt = 20
@@ -615,10 +616,16 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol 
                 } else {
                     text = messageTimelineItem.body
                 }
+            case .emote(let emoteItem):
+                if ServiceLocator.shared.settings.richTextEditorEnabled, let formattedBodyHTMLString = emoteItem.formattedBodyHTMLString {
+                    text = "/me " + formattedBodyHTMLString
+                } else {
+                    text = "/me " + messageTimelineItem.body
+                }
             default:
                 text = messageTimelineItem.body
             }
-
+            
             actionsSubject.send(.composer(action: .setText(text: text)))
             actionsSubject.send(.composer(action: .setMode(mode: .edit(originalItemId: messageTimelineItem.id))))
         case .copyPermalink:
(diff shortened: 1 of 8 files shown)


ara4n:
How do I make the tests run against the right sdk branch name? (EW would at least automatically try to build against the branch of the same name in the SDK...)

ElementX/Sources/Services/Room/RoomProxy.swift
```diff
@@ -253,17 +253,34 @@ class RoomProxy: RoomProxyProtocol {
         }
     }
     
+    func checkEmote(body: inout String, htmlBody: inout String?) -> Bool {
```
stefanceriu, line 256: This needs to be private and moved down to the right section but also the name is off as it doesn't just "check" but also modifies the original strings. 
I would split it into 2 clear methods `checkIsEmote` and `buildEmoteContent`. I'm not a fan of `inout` parameters so let's make `buildEmoteContent` return a tuple.

ElementX/Sources/Services/Room/RoomProxy.swift
```diff
@@ -253,17 +253,34 @@ class RoomProxy: RoomProxyProtocol {
...
+    
     func sendMessage(_ message: String, html: String?, inReplyTo eventID: String? = nil) async -> Result<Void, RoomProxyError> {
         sendMessageBackgroundTask = await backgroundTaskService.startBackgroundTask(withName: backgroundTaskName, isReusable: true)
         defer {
             sendMessageBackgroundTask?.stop()
         }
         
+        var body: String = message
+        var htmlBody: String? = html
+        let emote: Bool = checkEmote(body: &body, htmlBody: &htmlBody)
```
stefanceriu, line 277: Let's call this `isEmote`

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -615,10 +615,16 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
                 } else {
                     text = messageTimelineItem.body
                 }
+            case .emote(let emoteItem):
+                if ServiceLocator.shared.settings.richTextEditorEnabled, let formattedBodyHTMLString = emoteItem.formattedBodyHTMLString {
```
stefanceriu, line 619: Are you sure we want to give users the option to edit the original `/me`? What happens if they just delete it? Would the event turn into a plain room message text? 🤔

stefanceriu (changes requested):
Left comments inline for a couple of small tweaks that need to happen. Otherwise looks sensible to me.

@Velin92 you were going to look into user suggestions this week. They should probably be based on `CompletionSuggestion` from the old codebase which also supports slash commands so please take this PR under your wing.

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
ara4n: yup, which is desirable, so you can change whether you meant to do a /me or not.

stefanceriu:
> run against the right sdk branch name?

There is nothing like that atm. We agreed to look into building the sdk on the CI after the next release and this can very well be a part of that work.

manuroe:
> > run against the right sdk branch name?
> 
> There is nothing like that atm. We agreed to look into building the sdk on the CI after the next release and this can very well be a part of that work.

We will also need to improve the infra. GH action runners are slow to build the SDK bindings. They take [4h](https://github.com/matrix-org/matrix-rust-components-kotlin/actions/runs/6312986182) to build them for Android. This would be way too long to validate a PR on EAX or EIX.

ara4n:
ptal

ElementX/Sources/Services/Room/RoomProxy.swift
```diff
@@ -252,18 +252,33 @@ class RoomProxy: RoomProxyProtocol {
...
+            } else {
+                messageContent = messageEventContentFromHtml(body: body, htmlBody: htmlBody)
+            }
         } else {
-            messageContent = messageEventContentFromMarkdown(md: message)
+            if isEmote {
+                messageContent = messageEventContentFromMarkdownAsEmote(md: body)
+            } else {
+                messageContent = messageEventContentFromMarkdown(md: body)
+            }
```
Velin92, line 281: Can we a private func that does this? I also see that this code is quite similiar to the code below, so maybe we can actually have 1 single function and avoid replication?

ElementX/Sources/Services/Room/RoomProxy.swift
```diff
@@ -441,13 +456,27 @@ class RoomProxy: RoomProxyProtocol {
...
+            } else {
+                newMessageContent = messageEventContentFromHtml(body: body, htmlBody: htmlBody)
+            }
         } else {
-            newMessageContent = messageEventContentFromMarkdown(md: newMessage)
+            if isEmote {
+                newMessageContent = messageEventContentFromMarkdownAsEmote(md: body)
+            } else {
+                newMessageContent = messageEventContentFromMarkdown(md: body)
+            }
```
Velin92, line 478: As I said in the comment above, this is replicated and we can extract it into a function.

Velin92 (approved):

stefanceriu (approved):

PR #1850, opened 2023-10-02
Handle tap on user mentions

Velin92:
User mentions are now tappable and will send to the member details of the mentioned user in the given room.

Also added more mentions tests


stefanceriu (approved):
lgtm, very nice! 👏

PR #1856, opened 2023-10-05
Fix the room summary and reply view for voice messages

nimau:
This PR fixes:
- The room summary when the last message is a voice message

_Before_
Screenshot 2023-10-05 at 10 55 16

_After_
Screenshot 2023-10-05 at 10 55 16

- The reply view when replying to a voice message

Screenshot 2023-10-05 at 10 48 51


stefanceriu (approved):
Nice! 👍

PR #1862, opened 2023-10-06
Report rust tracing configuration filter in rageshakes

stefanceriu:


pixlwave (approved):
Looks good to me. Thinking longer term, maybe it would make sense to have a general `rust` field, that we dump in the contents of some kind of `rageshake_configuration` property from the SDK, so that we don't need to do app specific work for these kind of requests.

@jplatte WDYT?

jplatte:
Let's consider that if for data like this enters the picture. For now it wouldn't simplify things much, right?

pixlwave:
Yep, sounds like a plan to me 👍

PR #1863, opened 2023-10-06
Mention in RTE (but without the pill yet)

Velin92:
Now tapping on a suggestion makes the suggestion appear in RTE as a link, however this is not yet rendered as a Pill, which I will do in the next PR.

https://github.com/vector-im/element-x-ios/assets/34335419/387e2dab-678d-496d-9c90-1220efd5db6e

### DIFF
--- a/ElementX/Sources/Screens/ComposerToolbar/ComposerToolbarViewModel.swift
+++ b/ElementX/Sources/Screens/ComposerToolbar/ComposerToolbarViewModel.swift
@@ -22,9 +22,10 @@ import WysiwygComposer
 typealias ComposerToolbarViewModelType = StateStoreViewModel<ComposerToolbarViewState, ComposerToolbarViewAction>
 
 final class ComposerToolbarViewModel: ComposerToolbarViewModelType, ComposerToolbarViewModelProtocol {
-    private let appSettings: AppSettings
     private let wysiwygViewModel: WysiwygComposerViewModel
     private let completionSuggestionService: CompletionSuggestionServiceProtocol
+    private let appSettings: AppSettings
+
     private let actionsSubject: PassthroughSubject<ComposerToolbarViewModelAction, Never> = .init()
     var actions: AnyPublisher<ComposerToolbarViewModelAction, Never> {
         actionsSubject.eraseToAnyPublisher()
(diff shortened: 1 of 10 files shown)


ElementX/Sources/Screens/ComposerToolbar/ComposerToolbarViewModel.swift
```diff
@@ -22,6 +22,7 @@ import WysiwygComposer
 typealias ComposerToolbarViewModelType = StateStoreViewModel<ComposerToolbarViewState, ComposerToolbarViewAction>
 
 final class ComposerToolbarViewModel: ComposerToolbarViewModelType, ComposerToolbarViewModelProtocol {
+    private let appSettings: AppSettings
```
pixlwave, line 25: Tiny nit: Can we keep these ordered to match the init?

ElementX/Sources/Screens/ComposerToolbar/ComposerToolbarViewModel.swift
```diff
@@ -38,7 +39,8 @@ final class ComposerToolbarViewModel: ComposerToolbarViewModelType, ComposerTool
 
     private var currentLinkData: WysiwygLinkData?
 
-    init(wysiwygViewModel: WysiwygComposerViewModel, completionSuggestionService: CompletionSuggestionServiceProtocol, mediaProvider: MediaProviderProtocol) {
+    init(wysiwygViewModel: WysiwygComposerViewModel, completionSuggestionService: CompletionSuggestionServiceProtocol, mediaProvider: MediaProviderProtocol, appSetting: AppSettings) {
+        appSettings = appSetting
```
pixlwave, line 43: Same here

ElementX/Sources/Screens/ComposerToolbar/View/CompletionSuggestionView.swift
```diff
@@ -63,7 +64,9 @@ struct CompletionSuggestionView: View {
 
     private func list() -> some View {
```
pixlwave, line 65: Totally unrelated, but shouldn't this be a var?

ElementX/Sources/Screens/RoomScreen/RoomScreenCoordinator.swift
```diff
@@ -59,15 +60,18 @@ final class RoomScreenCoordinator: CoordinatorProtocol {
...
 
         wysiwygViewModel = WysiwygComposerViewModel(minHeight: ComposerConstant.minHeight,
                                                     maxCompressedHeight: ComposerConstant.maxHeight,
                                                     maxExpandedHeight: ComposerConstant.maxHeight,
                                                     parserStyle: .elementX)
-        composerViewModel = ComposerToolbarViewModel(wysiwygViewModel: wysiwygViewModel, completionSuggestionService: parameters.completionSuggestionService, mediaProvider: parameters.mediaProvider)
+        composerViewModel = ComposerToolbarViewModel(wysiwygViewModel: wysiwygViewModel,
+                                                     completionSuggestionService: parameters.completionSuggestionService,
+                                                     mediaProvider: parameters.mediaProvider,
+                                                     appSetting: parameters.appSettings)
```
pixlwave, line 74:
```suggestion
                                                     appSettings: parameters.appSettings)
```

pixlwave (approved):
Oooo completions! 😎

ElementX/Sources/Screens/ComposerToolbar/View/CompletionSuggestionView.swift
Velin92: I guess... this code was copied from El-iOS to match the same intrinsic behaviour list + max height based on number of cell from there, so didn't look too much into improving it. Also this is probably going to be completely remade by Design, so is likely going to completely change in the future. This is just to have something that works that can be tested functionality wise

PR #1867, opened 2023-10-09
Fixes 1864 - Message forwarding room list filtering and pagination problems

stefanceriu:
* use a separate room summary provider instance just for message forwarding
* filter rooms through the rust sdk
* add pagination support

### DIFF
--- a/ElementX/Sources/Other/HTMLParsing/DTHTMLElement+AttributedStringBuilder.swift
+++ b/ElementX/Sources/Other/HTMLParsing/DTHTMLElement+AttributedStringBuilder.swift
@@ -64,13 +64,11 @@ public extension DTHTMLElement {
         }
     }
     
-    private static var allowedHTMLTags = {
-        ["font", // custom to matrix for IRC-style font coloring
-         "del", // for markdown
-         "body", // added internally by DTCoreText
-         "mx-reply",
-         "h1", "h2", "h3", "h4", "h5", "h6", "blockquote", "p", "a", "ul", "ol",
-         "nl", "li", "b", "i", "u", "strong", "em", "strike", "code", "hr", "br", "div",
-         "table", "thead", "caption", "tbody", "tr", "th", "td", "pre"]
-    }()
+    private static var allowedHTMLTags = ["font", // custom to matrix for IRC-style font coloring
+                                          "del", // for markdown
+                                          "body", // added internally by DTCoreText
+                                          "mx-reply",
+                                          "h1", "h2", "h3", "h4", "h5", "h6", "blockquote", "p", "a", "ul", "ol",
+                                          "nl", "li", "b", "i", "u", "strong", "em", "strike", "code", "hr", "br", "div",
+                                          "table", "thead", "caption", "tbody", "tr", "th", "td", "pre"]
 }
--- a/ElementX/Sources/Other/HTMLParsing/DTHTMLElement+AttributedStringBuilder.swift
+++ b/ElementX/Sources/Other/HTMLParsing/DTHTMLElement+AttributedStringBuilder.swift
@@ -64,7 +64,7 @@ public extension DTHTMLElement {
         }
     }
     
-    private static var allowedHTMLTags = ["font", // custom to matrix for IRC-style font coloring
+    private static let allowedHTMLTags = ["font", // custom to matrix for IRC-style font coloring
                                           "del", // for markdown
                                           "body", // added internally by DTCoreText
                                           "mx-reply",
--- a/ElementX/Sources/Screens/MessageForwardingScreen/MessageForwardingScreenViewModel.swift
+++ b/ElementX/Sources/Screens/MessageForwardingScreen/MessageForwardingScreenViewModel.swift
@@ -43,6 +43,15 @@ class MessageForwardingScreenViewModel: MessageForwardingScreenViewModelType, Me
             }
             .store(in: &cancellables)
         
+        context.$viewState
+            .map(\.bindings.searchQuery)
+            .removeDuplicates()
+            .sink { [weak self] searchQuery in
+                guard let self else { return }
+                self.roomSummaryProvider?.setFilter(.normalizedMatchRoomName(searchQuery))
+            }
+            .store(in: &cancellables)
+        
         updateRooms()
     }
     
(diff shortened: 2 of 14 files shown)


ElementX/Sources/Other/HTMLParsing/DTHTMLElement+AttributedStringBuilder.swift
```diff
@@ -64,13 +64,11 @@ public extension DTHTMLElement {
...
-    private static var allowedHTMLTags = {
-        ["font", // custom to matrix for IRC-style font coloring
-         "del", // for markdown
-         "body", // added internally by DTCoreText
-         "mx-reply",
-         "h1", "h2", "h3", "h4", "h5", "h6", "blockquote", "p", "a", "ul", "ol",
-         "nl", "li", "b", "i", "u", "strong", "em", "strike", "code", "hr", "br", "div",
-         "table", "thead", "caption", "tbody", "tr", "th", "td", "pre"]
-    }()
+    private static var allowedHTMLTags = ["font", // custom to matrix for IRC-style font coloring
```
pixlwave, line 67: Good spot. I think this can be a `let` too?
```suggestion
    private static let allowedHTMLTags = ["font", // custom to matrix for IRC-style font coloring
```

ElementX/Sources/Screens/MessageForwardingScreen/View/MessageForwardingScreen.swift
```diff
@@ -22,10 +22,18 @@ struct MessageForwardingScreen: View {
     var body: some View {
         Form {
             Section {
-                ForEach(context.viewState.visibleRooms) { room in
+                ForEach(context.viewState.rooms) { room in
                     MessageForwardingRoomCell(room: room, context: context)
                         .buttonStyle(FormButtonStyle(accessory: .singleSelection(isSelected: context.viewState.selectedRoomID == room.id)))
                 }
+            } header: {
```
pixlwave, line 29: Is it worth adding a note for the future that whenever we drop iOS 16, these can be replaced by the new scrollview APIs?

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -35,10 +35,11 @@ class ClientProxy: ClientProxyProtocol {
     private var syncService: SyncService?
     private var syncServiceStateUpdateTaskHandle: TaskHandle?
     
-    var roomSummaryProvider: RoomSummaryProviderProtocol?
-    var inviteSummaryProvider: RoomSummaryProviderProtocol?
+    private(set) var roomSummaryProvider: RoomSummaryProviderProtocol?
+    private(set) var messageForwardingRoomSummaryProvider: RoomSummaryProviderProtocol?
```
pixlwave, line 39: Maybe a comment on this one that explains why we have both using the same list?

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -22,10 +22,11 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
     private let roomListService: RoomListServiceProtocol
     private let eventStringBuilder: RoomEventStringBuilder
     private let name: String
+    private let shouldUpdateVisibleRange: Bool
     private let notificationSettings: NotificationSettingsProxyProtocol
     private let backgroundTaskService: BackgroundTaskServiceProtocol
     
-    private let roomListPageSize = 200
+    private let roomListPageSize = 40
```
pixlwave, line 29: Was this change intentional??

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -59,12 +60,14 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
     init(roomListService: RoomListServiceProtocol,
          eventStringBuilder: RoomEventStringBuilder,
          name: String,
+         shouldUpdateVisibleRange: Bool = false,
```
pixlwave, line 63: Two thoughts:
- Might also be useful to document why you would pass this in as `true`/`false`?
- Shouldn't the default value be `true` here?

ElementX/Sources/Screens/MessageForwardingScreen/MessageForwardingScreenViewModel.swift
```diff
@@ -91,4 +104,22 @@ class MessageForwardingScreenViewModel: MessageForwardingScreenViewModelType, Me
         
         MXLog.verbose("Finished updating rooms")
     }
+    
+    /// The actual range values don't matter as long as they contain the lower
+    /// or upper bounds
```
pixlwave, line 109: Should we add some more detail to this comment? Coming to it fresh, it doesn't particularly explain why this is the case.

pixlwave (approved):
LGTM

ElementX/Sources/Other/HTMLParsing/DTHTMLElement+AttributedStringBuilder.swift
stefanceriu: Thanks, all SwiftFormat's doing 😅

ElementX/Sources/Screens/MessageForwardingScreen/View/MessageForwardingScreen.swift
stefanceriu: Yep, good idea, let's keep it consistent

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
stefanceriu: Definitely not 🙈

ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu: Do you think that's really necessary, they different instances of the same list which I assumed makes perfect sense without any extra comments 🤔

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
stefanceriu: > Might also be useful to document why you would pass this in as true/false?

Will do

> Shouldn't the default value be true here?

The only summary provider that should actually worry about this is the room list one so I figured I wouldn't bother all the rest

ElementX/Sources/Services/Client/ClientProxy.swift
pixlwave: If you hadn't seen this PR it would be a valid question to ask *why* though? Just a short comment like so would be enough
```
/// A separate provider for message forwarding to avoid showing a filtered room list.
```

stefanceriu: Fair

PR #1873, opened 2023-10-10
Improved display of voice messages in the timeline

nimau:
This PR:
- makes the progression indicator animation smoother (previously, the progression indicator was jumping from one sample to the next)
- fixes the waveform rendering
- fixes the duration string displayed when the duration of the voice message exceeds 10 minutes
- fixes the accessibility for voice messages in the timeline

### DIFF
--- a/ElementX/Sources/Services/AudioPlayer/AudioPlayerState.swift
+++ b/ElementX/Sources/Services/AudioPlayer/AudioPlayerState.swift
@@ -35,7 +35,7 @@ class AudioPlayerState: ObservableObject {
 
     private var audioPlayer: AudioPlayerProtocol?
     private var cancellables: Set<AnyCancellable> = []
-    private var cancellableTimer: AnyCancellable?
+    private var displayLink: CADisplayLink?
 
     var isAttached: Bool {
         audioPlayer != nil
@@ -50,6 +50,10 @@ class AudioPlayerState: ObservableObject {
         self.waveform = waveform ?? Waveform(data: [])
         self.progress = progress
         playbackState = .stopped
+    }
+    
+    deinit {
+        displayLink?.invalidate()
     }
     
     func updateState(progress: Double) async {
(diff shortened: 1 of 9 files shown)


ElementX/Sources/Services/AudioPlayer/AudioPlayerState.swift
```diff
@@ -122,22 +127,23 @@ class AudioPlayerState: ObservableObject {
...
-                guard let self else { return }
-                if let currentTime = self.audioPlayer?.currentTime, self.duration > 0 {
-                    self.progress = currentTime / self.duration
-                }
-            })
+        if displayLink != nil {
+            displayLink?.invalidate()
+        }
+        displayLink = CADisplayLink(target: self, selector: #selector(updateProgress))
+        displayLink?.preferredFrameRateRange = .init(minimum: 10, maximum: 20)
```
pixlwave, line 134: TIL 😄

ElementX/Sources/Services/AudioPlayer/AudioPlayerState.swift
```diff
@@ -122,22 +127,23 @@ class AudioPlayerState: ObservableObject {
...
-            .autoconnect()
-            .receive(on: DispatchQueue.main)
-            .sink(receiveValue: { [weak self] _ in
-                guard let self else { return }
-                if let currentTime = self.audioPlayer?.currentTime, self.duration > 0 {
-                    self.progress = currentTime / self.duration
-                }
-            })
+        if displayLink != nil {
+            displayLink?.invalidate()
```
pixlwave, line 131: Just for cleanness, might be good to call `stopPublishProgress` here instead?

ElementX/Sources/Services/Timeline/TimelineItems/Items/Messages/VoiceMessages/WaveformView.swift
```diff
@@ -56,43 +70,49 @@ struct WaveformView: View {
...
-                        let drawingAmplitude = max(minimumGraphAmplitude, sample * (height - 2))
-
-                        path.move(to: CGPoint(x: xOffset, y: centerY - drawingAmplitude / 2))
-                        path.addLine(to: CGPoint(x: xOffset, y: centerY + drawingAmplitude / 2))
-                        xOffset += lineWidth + linePadding
-                        index += 1
-                    }
-                }
-                .stroke(Color.compound.iconSecondary, style: StrokeStyle(lineWidth: lineWidth, lineCap: .round))
+                WaveformShape(lineWidth: lineWidth,
```
pixlwave, line 73: 👌

ElementX/Sources/Services/Timeline/TimelineItems/Items/Messages/VoiceMessages/WaveformView.swift
```diff
@@ -25,10 +25,21 @@ extension Waveform {
...
+            for index in 0..<count {
+                let targetIndex = (Double(index) * (originalCount / expectedCount)).rounded()
+                filteredData.append(UInt16(data[Int(targetIndex)]))
+            }
+        } else {
+            filteredData = data
+        }
+        // Normalize the sample
+        let max = max(1.0, filteredData.max().flatMap { Float($0) } ?? 1.0)
+        return filteredData.map { Float($0) / max }
```
pixlwave, line 42: So this might be early optimisation but I worry that this is going to be called frequently now and that it's a lot of redundant work. Two thoughts:
- Should we pre-compute this in the `init` and store it inside of `Waveform`?
- How many samples are inside `data`? Is it already a reduced count for display or could it be the entire file?

pixlwave (approved):

ElementX/Sources/Services/AudioPlayer/AudioPlayerState.swift
nimau: 👍

ElementX/Sources/Services/Timeline/TimelineItems/Items/Messages/VoiceMessages/WaveformView.swift
nimau: Ok, so I've updated the code so that the waveform data is only computed when the geometryreader size changes (we need the geometryready size to know how many samples we can render).
I think the maximum data length is 100, but for short voice message, it can be less.  If the width of the waveform view is 150, lineWidth = 2 and linePadding = 2, we will only display 37 samples.

pixlwave: Right, that makes sense 👍

pixlwave (approved):

PR #1874, opened 2023-10-10
Render pills in RTE

Velin92:
Pills can now be rendered in RTE.

This requires an incoming update from the RTE package, for now I have commitred the project that is using the local state but  I am going to update this soon.

Simulator Screenshot - iPhone 14 - 2023-10-10 at 17 03 29

https://github.com/vector-im/element-x-ios/assets/34335419/c7ef0cdf-f540-4fa7-bb9e-6abbb4e5de19


ElementX/Sources/Screens/ComposerToolbar/ComposerToolbarViewModel.swift
```diff
@@ -40,11 +40,11 @@ final class ComposerToolbarViewModel: ComposerToolbarViewModelType, ComposerTool
 
     private var currentLinkData: WysiwygLinkData?
 
-    init(wysiwygViewModel: WysiwygComposerViewModel, completionSuggestionService: CompletionSuggestionServiceProtocol, mediaProvider: MediaProviderProtocol, appSettings: AppSettings) {
+    init(wysiwygViewModel: WysiwygComposerViewModel, completionSuggestionService: CompletionSuggestionServiceProtocol, mediaProvider: MediaProviderProtocol, appSettings: AppSettings, roomContext: RoomScreenViewModel.Context) {
```
pixlwave, line 43: Can we pass in a `MentionDisplayHelper` here and create the `ComposerMentionDisplayHelper` outside of this VM? We should never give one VM a dependency on an entirely separate context.

ElementX/Sources/Other/HTMLParsing/AttributedStringBuilder.swift
pixlwave, line 1: I've probably missed it, but I can't see where these are being used that needed the change to make them public?

ElementX/Sources/Screens/ComposerToolbar/ComposerToolbarViewModel.swift
```diff
@@ -293,3 +320,33 @@ private extension LinkAction {
...
+    func restoreMarkdown(in attributedString: NSAttributedString) -> String {
+        attributedString.string
+    }
+    
+    func replacementForMention(_ url: String, text: String) -> NSAttributedString? {
+        replacementForMentionClosure(url, text)
+    }
+}
+
+final class ComposerMentionDisplayHelper: MentionDisplayHelper {
```
pixlwave, line 346: Rather than exposing the roomContext from this class, could we instead add all the accessors needed to it so its scoped to only what is needed for pills? Where do I find the definition of `MentionDisplayHelper`? Can that be used to set the requirements of the accessors needed?

pixlwave (changes requested):
I think we need to refine the access a little bit more:

ElementX/Sources/Screens/ComposerToolbar/ComposerToolbarViewModel.swift
Velin92: it's internal to the RTE and is only a markdown protocol I have made to be able to inject an object that gives back to the provider any context required for displaying a pill, it is essentially an empty protocol used to recreat the way in which  the TextView can pass a context onto the provider.

ElementX/Sources/Other/HTMLParsing/AttributedStringBuilder.swift
Velin92: line 168 of ComposerToolbarViewModel

ElementX/Sources/Other/Pills/PillAttachmentViewProvider.swift
```diff
@@ -61,18 +64,37 @@ final class PillAttachmentViewProvider: NSTextAttachmentViewProvider {
...
+    init(roomContext: RoomScreenViewModel.Context) {
+        self.roomContext = roomContext
+    }
+    
+    private init() {
+        roomContext = nil
+    }
+    
+    static var mock: Self {
+        Self()
```
pixlwave, line 90: Could use `Self(roomContext: RoomScreenViewModel.mock.context)` here I think, if it provides any nice benefits.

ElementX/Sources/Screens/ComposerToolbar/ComposerToolbarViewModel.swift
```diff
@@ -293,3 +319,25 @@ private extension LinkAction {
         return url
     }
 }
+
+private final class ComposerMentionReplacer: MentionReplacer {
+    let replacementForMentionClosure: (String, String) -> (NSAttributedString?)
+    
+    init(replacementForMentionClosure: @escaping (String, String) -> (NSAttributedString?)) {
```
pixlwave, line 326: Is it possible to add labels on the 2 `String`s to identify them (I can never remember in closures). Otherwise can we add doc comments just to make it super easy to identify :)

pixlwave (approved):
Nice, looks good to me 😎

ElementX/Sources/Other/Pills/PillAttachmentViewProvider.swift
Velin92: oh yeah you are right makes way more sense

PR #1876, opened 2023-10-11
Initial setup ready for PIN/Biometric app lock.

pixlwave:
The PR lays the initial groundwork for adding PIN/biometric unlock to the app. It makes the following changes:
- Adds a WindowManager which can switch between the main window and an alternate window.
    - Additionally includes an overlay window on the main window which we can use for global user indicators in the future.
- Adds a AppLockFlowCoordinator and placeholder AppLockScreen.
- Displays this flow in the alternate window when the app is backgrounded (placeholder view) and foregrounded (app lock screen).
- Adds a feature flag for the flow.

https://github.com/vector-im/element-x-ios/assets/6060466/36a79992-d3ea-4f68-bffe-660f03c0042c

### DIFF
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -129,6 +129,7 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
 
         observeApplicationState()
         observeNetworkState()
+        observeAppLockChanges()
         
         registerBackgroundAppRefresh()
     }
@@ -194,16 +195,6 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
     
     func windowManager(_ windowManager: WindowManager, didConfigureWith windowScene: UIWindowScene) {
         windowManager.alternateWindow.rootViewController = UIHostingController(rootView: appLockFlowCoordinator.toPresentable())
-        
-        appLockFlowCoordinator.actions.sink { action in
-            switch action {
-            case .lockApp:
-                windowManager.switchToAlternate()
-            case .unlockApp:
-                windowManager.switchToMain()
-            }
-        }
-        .store(in: &cancellables)
     }
     
     // MARK: - NotificationManagerDelegate
@@ -579,6 +570,19 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
                 }
             }
             .store(in: &cancellables)
+    }
+    
+    private func observeAppLockChanges() {
+        appLockFlowCoordinator.actions.sink { [weak self] action in
+            guard let self else { return }
+            switch action {
+            case .lockApp:
+                windowManager.switchToAlternate()
+            case .unlockApp:
+                windowManager.switchToMain()
+            }
+        }
+        .store(in: &cancellables)
     }
     
     private func handleAppRoute(_ appRoute: AppRoute) {
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -20,7 +20,7 @@ import MatrixRustSDK
 import SwiftUI
 import Version
 
-class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate, NotificationManagerDelegate {
+class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate, NotificationManagerDelegate, WindowManagerDelegate {
     private let stateMachine: AppCoordinatorStateMachine
     private let navigationRootCoordinator: NavigationRootCoordinator
     private let userSessionStore: UserSessionStoreProtocol
@@ -43,8 +43,9 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
         }
     }
     
-    private var userSessionFlowCoordinator: UserSessionFlowCoordinator?
     private var authenticationCoordinator: AuthenticationCoordinator?
+    private let appLockFlowCoordinator: AppLockFlowCoordinator
+    private var userSessionFlowCoordinator: UserSessionFlowCoordinator?
     private var softLogoutCoordinator: SoftLogoutScreenCoordinator?
     
     private let backgroundTaskService: BackgroundTaskServiceProtocol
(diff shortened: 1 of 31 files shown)


ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -178,6 +190,22 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
...
+        appLockFlowCoordinator.actions.sink { action in
+            switch action {
+            case .lockApp:
+                windowManager.switchToAlternate()
+            case .unlockApp:
+                windowManager.switchToMain()
+            }
+        }
+        .store(in: &cancellables)
+    }
```
Velin92, line 207: shouldn't this function have a check to ensure that is called only once?

ElementX/Sources/Application/AppSettings.swift
```diff
@@ -257,4 +258,7 @@ final class AppSettings {
     
     @UserPreference(key: UserDefaultsKeys.mentionsEnabled, defaultValue: false, storageType: .userDefaults(store))
     var mentionsEnabled
+    
+    @UserPreference(key: UserDefaultsKeys.appLockFlowEnabled, defaultValue: false, storageType: .volatile)
```
Velin92, line 262: why volatile?

ElementX/Sources/Application/Windowing/WindowManager.swift
```diff
@@ -0,0 +1,67 @@
...
+/// an alternate window to switch contexts whilst also preserving the main view hierarchy.
+class WindowManager {
+    weak var delegate: WindowManagerDelegate?
+    
+    /// The app's main window (we only support a single scene).
+    private(set) var mainWindow: UIWindow!
+    /// Presented on top of the main window, to display e.g. user indicators.
+    private(set) var overlayWindow: UIWindow!
+    /// A secondary window that can be presented instead of the main/overlay window combo.
+    private(set) var alternateWindow: UIWindow!
```
Velin92, line 36: can't we have these as normal optionals for safety?

ElementX/Sources/FlowCoordinators/AppLockFlowCoordinator.swift
```diff
@@ -0,0 +1,87 @@
...
+        
+        let coordinator = AppLockScreenCoordinator(parameters: .init(appLockService: appLockService))
+        coordinator.actions.sink { [weak self] action in
+            guard let self else { return }
+            switch action {
+            case .appUnlocked:
+                actionsSubject.send(.unlockApp)
+            }
+        }
+        .store(in: &cancellables)
```
Velin92, line 82: do we need to sink this multiple times? we are subscribing to the actions publisher every time this is called.
Maybe I would at least store this into a single known AnyCancellable that gets overwritten everytime.

ElementX/Sources/Screens/AppLockScreen/AppLockScreenModels.swift
```diff
@@ -0,0 +1,40 @@
...
+struct AppLockScreenViewStateBindings { }
+
+enum AppLockScreenViewAction: CustomStringConvertible {
+    /// Attempt to unlock the app with the supplied PIN code.
+    case submitPINCode(String)
+    
+    var description: String {
+        switch self {
+        case .submitPINCode:
+            return "submitPINCode"
```
Velin92, line 37: localised?

Velin92 (approved):
small comments but generally LGTM

ElementX/Sources/Screens/AppLockScreen/AppLockScreenModels.swift
pixlwave: This is the debug description to avoid logging the user's PIN code, it will never be shown to the user.

That said, I haven't localised the strings in the view yet as there aren't any designs. I'll add them to untranslated to be clear.

ElementX/Sources/Application/AppCoordinator.swift
Velin92: sorry actualy not this function, the `configure` function on `WindowManager` that calls this one should

ElementX/Sources/Application/AppSettings.swift
pixlwave: Because if I make a mistake, I don't want to lock people out of the app in a way that they can't disable the flag again 😂

ElementX/Sources/Application/AppCoordinator.swift
pixlwave: Yeah good point, I've moved the actions sink into a method called from `init` so that if for some reason it's called again, we still reconfigure the alternate window.

ElementX/Sources/Application/Windowing/WindowManager.swift
pixlwave: I don't think we should. These must be configured before the app use used and if they're not, we want to know about that and not allow access to the app (otherwise it would bypass the app lock feature).

ElementX/Sources/FlowCoordinators/AppLockFlowCoordinator.swift
pixlwave: This is the same behaviour that we do in all the flow coordinators. We collect old cancellables [everywhere](https://github.com/vector-im/element-x-ios/blob/b35bee5d8fb1e88a645d266fa3bb66ca78a125dd/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift#L415), but they will never be fired after the presented coordinator has been removed from the navigation stack and torn down.

ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -178,6 +191,12 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
         stateMachine.processEvent(.createdUserSession)
     }
     
+    // MARK: - WindowManagerDelegate
+    
+    func windowManager(_ windowManager: WindowManager, didConfigureWith windowScene: UIWindowScene) {
```
stefanceriu, line 196: Do we even need to expose the scene?

ElementX/Sources/Application/Windowing/WindowManager.swift
```diff
@@ -0,0 +1,67 @@
...
+    
+    /// Configures the window manager to operate on the supplied scene.
+    func configure(with windowScene: UIWindowScene) {
+        mainWindow = windowScene.keyWindow
+        
+        overlayWindow = UIWindow(windowScene: windowScene)
+        overlayWindow.backgroundColor = .clear
+        // We don't support user interaction on our indicators so disable interaction, to pass
+        // touches through to the main window. If this changes, there's another solution here:
```
stefanceriu, line 45: 👌

stefanceriu (approved):
Awesome stuff, can't wait to start using that new `overlayWindow`

ElementX/Sources/Application/AppCoordinator.swift
pixlwave: Probably not, removed.

PR #1878, opened 2023-10-11
Various optimisations

stefanceriu:
* Cache TimelineBubbleLayout subview sizes
* Cache MessageText sizes, avoid extra updates
* Only use the `CollapsibleReactionLayout` if there's more than 5 reactions on a particular message
* Upgrade Sentry to 8.13.0, disable various options as they're not useful and impact performance


ElementX/Sources/Services/BugReport/BugReportService.swift
```diff
@@ -69,6 +69,14 @@ class BugReportService: NSObject, BugReportServiceProtocol {
             
             options.dsn = self.sentryURL.absoluteString
             
+            // Sentry swizzling shows up quite often as the heaviest stack trace when profiling
+            // We don't need any of the features it powers (see docs)
+            options.enableSwizzling = false
+            
+            // WatchdogTermination is currently the top issue but we've had zero complaints
+            // so it might very well just all be false positives
+            options.enableWatchdogTerminationTracking = false
```
pixlwave, line 78: Just a thought: Did you check the release notes for the newer versions to see if there were any fixed bugs around this?

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineBubbleLayout.swift
```diff
@@ -36,30 +40,41 @@ struct TimelineBubbleLayout: Layout {
...
     
-    func placeSubviews(in bounds: CGRect, proposal: ProposedViewSize, subviews: Subviews, cache: inout ()) {
+    func placeSubviews(in bounds: CGRect, proposal: ProposedViewSize, subviews: Subviews, cache: inout Cache) {
         guard !subviews.isEmpty else { return }
         
         // Calculate the width using the regular text and the non-greedy quote bubbles.
         let layoutSubviews = subviews.filter { $0.priority != Priority.visibleQuote }
-        let maxWidth = layoutSubviews.map { $0.sizeThatFits(proposal).width }.reduce(0, max)
         
-        // Place the regular text and greedy quote bubbles using the calculated width.
```
pixlwave, line 60: Was this removal necessary? 😢

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineBubbleLayout.swift
```diff
@@ -70,4 +85,25 @@ struct TimelineBubbleLayout: Layout {
             y += height + spacing
         }
     }
+    
+    // MARK: - Private
+    
+    private func cachedSizeFor(subview: LayoutSubview, subviews: LayoutSubviews, proposedSize: ProposedViewSize, cache: inout Cache) -> CGSize {
```
pixlwave, line 91: This name is slightly confusing at the call site as it implies the size *only* comes from the cache. Its not a great suggestion, but `sizeThatFitsUsingCache` or something would be clearer.

ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineReactionsView.swift
```diff
@@ -34,6 +34,40 @@ struct TimelineReactionsView: View {
...
+            Button {
+                context.send(viewAction: .displayEmojiPicker(itemID: itemID))
+            } label: {
+                TimelineReactionAddMoreButtonLabel()
+            }
+        }
+        .environment(\.layoutDirection, reactionsLayoutDirection)
+        .animation(.easeInOut(duration: 0.1).disabledDuringTests(), value: reactions)
+        .padding(.leading, 4)
+    }
```
pixlwave, line 68: A neater solution would be to use `AnyLayout` in the collapsibleBody so we don't duplicate any code.

pixlwave (approved):
I'm not massively convinced reverting to an `HStack` is better than figuring out the Cache in the reactions layout, but the code lgtm.

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineBubbleLayout.swift
stefanceriu: Oh, sorry, no, complete mistake 👍

ElementX/Sources/Services/BugReport/BugReportService.swift
stefanceriu: I did yes, nothing for that. The only interesting bit was `Symbolicate locally only when debug is enabled` in `8.9.0`

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineBubbleLayout.swift
stefanceriu: How about just `sizeFor`? :D

pixlwave: How about `size(for:)` then 😉

PR #1880, opened 2023-10-11
@room in user suggestions

### LINKED ISSUE #1875: Mentions: add @room mention to the suggestions
For now the user suggestions only present the users, but they should also include an extra suggestion to mention all the users in the room.

Velin92:
fixes #1875

Requires a new RTE version.

Allows to select the @room mention in the completion suggestion and add it to the RTE, but only if the user has the rights to trigger room notifications.


ElementX/Sources/Other/Pills/PillConstants.swift
```diff
@@ -0,0 +1,23 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+
+enum PillConstants {
+    static let atRoom = "@room"
+    static let everyone = "Everyone"
```
pixlwave, line 21: Shouldn't this be localised rather than a constant?

ElementX/Sources/Other/Pills/PillConstants.swift
```diff
@@ -0,0 +1,23 @@
...
+    static let composerAtRoomURLString = "#"
```
pixlwave, line 22: Probably add a doc comment on this one, as I have no idea how "#" is related to `@room` without looking at the code.

ElementX/Sources/Screens/ComposerToolbar/CompletionSuggestionModels.swift
```diff
@@ -33,6 +45,10 @@ struct MentionSuggestionItem: Identifiable, Equatable {
     let id: String
     let displayName: String?
     let avatarURL: URL?
+    
+    static func allUsersMention(roomAvatar: URL?) -> Self {
+        MentionSuggestionItem(id: PillConstants.atRoom, displayName: "Everyone", avatarURL: roomAvatar)
```
pixlwave, line 50: Should this use PillConstants.everyone (or L10n.commonEveryone if localising)?

ElementX/Sources/Screens/ComposerToolbar/CompletionSuggestionService.swift
```diff
@@ -18,47 +18,55 @@ import Combine
...
                 switch suggestionTrigger.type {
                 case .user:
-                    return members.filter { member in
-                        guard !member.isAccountOwner && member.membership == .join else {
-                            return false
+                    return members
+                        .filter { !$0.isAccountOwner && $0.membership == .join }
+                        .map { SuggestionItem.user(item: .init(id: $0.userID, displayName: $0.displayName, avatarURL: $0.avatarURL)) }
+                        .appending(self?.canMentionAllUsers ?? false ? [SuggestionItem.allUsers(item: .allUsersMention(roomAvatar: roomProxy.avatarURL))] : [])
+                        .filter { member in
```
pixlwave, line 45: This is going to iterate over the array 3 times, which presumably could be slow in large rooms. Can we use a `compactMap` to filter & map in a single iteration and then make the append at the end?

UnitTests/Sources/CompletionSuggestionServiceTests.swift
```diff
@@ -31,7 +31,7 @@ final class CompletionSuggestionServiceTests: XCTestCase {
         let members: [RoomMemberProxyMock] = [alice, .mockBob, .mockCharlie, .mockMe]
         let roomProxyMock = RoomProxyMock(with: .init(displayName: "test", members: members))
         let service = CompletionSuggestionService(roomProxy: roomProxyMock, areSuggestionsEnabled: true)
-        
+
```
pixlwave, line 34: Probably unintentional

pixlwave (approved):
LGTM 👍

ElementX/Sources/Screens/ComposerToolbar/CompletionSuggestionModels.swift
Velin92: we have something like that? I guess we can use that one, even if on El-iOS is not localised.

ElementX/Sources/Other/Pills/PillConstants.swift
Velin92: asking design, since they are not localised on El-iOS.

pixlwave (approved):
👌

PR #1886, opened 2023-10-12
Remove DesignKit package

pixlwave:
This PR deletes the old DesignKit package and removes the use of the legacy element-design-tokens dependency.
- The button styles are now part of Compound.
- The text style has been moved to the Authentication module as that is its only use.
- The brand colour has been added locally for now until its included in Compound.


stefanceriu (approved):
Yey! 👏

PR #1889, opened 2023-10-13
Remove ‘copy’ action for "media like" types

alfogrillo:
This PR removes the "copy" actions for "media like" events.
This is because the "copy" action currently copies the textual body of the event and not the actual media like users probably expect-.


nimau (approved):
SGTM. I wasn't sure whether you could copy a `.notice' but it seems possible under El-iOS.

PR #1891, opened 2023-10-13
Pills performance improvements

Velin92:
The main issue was caused by decoding the UIFont, now instead we only take the properties we care from the font and encode and decode them from a simple struct.
Also some minor improvements in how the attachment is created, in how the pills are added and removed to the view, and a way to inject the room context without making it necessarily observable

### DIFF
--- a/ElementX/Sources/Other/Pills/MessageText.swift
+++ b/ElementX/Sources/Other/Pills/MessageText.swift
@@ -16,8 +16,6 @@
 
 import SwiftUI
 
-import Collections
-
 final class MessageTextView: UITextView, PillAttachmentViewProviderDelegate {
     var roomContext: RoomScreenViewModel.Context?
     var updateClosure: (() -> Void)?
--- a/ElementX/Sources/Other/Pills/MessageText.swift
+++ b/ElementX/Sources/Other/Pills/MessageText.swift
@@ -16,9 +16,12 @@
 
 import SwiftUI
 
+import Collections
+
 final class MessageTextView: UITextView, PillAttachmentViewProviderDelegate {
     var roomContext: RoomScreenViewModel.Context?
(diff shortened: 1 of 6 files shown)


ElementX/Sources/Other/Pills/MessageText.swift
```diff
@@ -16,9 +16,12 @@
 
 import SwiftUI
 
+import Collections
```
pixlwave, line 19: I think this one is unused now?

ElementX/Sources/Other/Pills/PillTextAttachmentData.swift
```diff
@@ -24,43 +24,22 @@ enum PillType: Codable, Equatable {
     case allUsers
 }
 
-struct PillTextAttachmentData {
-    // MARK: - Properties
+struct EnclosingFontData: Codable, Equatable {
```
pixlwave, line 27: Should probably call this `PillTextAttachmentFontData` or nest it as `PillTextAttachmentData.Font`.

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -925,3 +925,15 @@ private struct ReplyInfo {
...
     let isThread: Bool
 }
+
+private struct RoomContextKey: EnvironmentKey {
+    @MainActor
+    static let defaultValue = RoomScreenViewModel.mock.context
+}
+
+extension EnvironmentValues {
+    var roomContext: RoomScreenViewModel.Context {
```
pixlwave, line 935: Maybe we should add a doc comment saying this is for access to the context for sending actions without observing the room state?

pixlwave (approved):
LGTM 👍

PR #1893, opened 2023-10-13
Stop voice message on redaction

alfogrillo:
This PR adds the logic to stop playing a voice message when it gets redacted.


nimau (approved):
SGTM!

stefanceriu (approved):
Do we need to be smarter here and actually delete the converted file?

alfogrillo:
> Do we need to be smarted here and actually delete the converted file?

We need to think in general how to manage caching for files/voice messages.
In fact we never evict cache items, and this will be a problem.
We will analyze better the problem soon. 👍🏻

PR #1894, opened 2023-10-13
Pill View rendering, performance improvement part 2

Velin92:
Calculating the bound is extremely intensive, so we only calculate them the first time or if necessary after a text change


pixlwave (approved):
👌

Issue #660, opened 2023-03-01
VoiceOver rotor on messages 


sempruijs:
## problem

VoiceOver users navigate allot with the rotor.
With the rotor, you can select secondary options of the element that is selected. For visual people, you can see it like a right mouse click.

Currently, this is not implemented on messages, which is frustrating because VoiceOver users are used to navigate this way. (messengers like WhatsApp, iMessage and signal already have this and VoiceOver users love this feature)

## solution

Implement the rotor on messages. With 1 swipe they can delete messages, reply, forward, etc.

## what should happen

1. Turn VoiceOver on
2. View a conversation (so you can compose a message that view)
3. put the VoiceOver cursor on a message. (That is already send by you or someone else)
4. Swipe vertically to select secondary options. VoiceOver users hear options like: Reply, Remove, forward.
5. When the users hears the option that they want to select, they select it by double tapping. 
6. The action is performed.

**Describe alternatives you've considered**

One option is to keep it as it is. Currently the workflow to delete a message for instance is holding the cursor on a message that they want to delete, simulate a long press by triple tapping and clicking on the delete button.

For VoiceOver users that know allot about the subject, they know the triple tap exist. But an average VoiceOver user doesn't know this and feel frustrated that they cannot delete a message or reply to a message.

An average VoiceOver user does know about the rotor and when they revalidate the tutors teach about the rotor and not about the triple tap. So the rotor it is!

## links

[explenation of what the rotor is and how to use it](https://support.apple.com/en-us/HT204783)

[swiftui documentation for implementing secondary actions for swiftUI](https://developer.apple.com/documentation/swiftui/view/accessibilityactions(_:))

kittykat:
@sempruijs Would you mind reaching out to me over Matrix to chat about the issues you've filed? You can reach me at @kittykat:matrix.org

kittykat:
Each item in the bottom sheet should correspond to an item in the rotor. For example, for a bottom sheet containing:
```
reply
delete
reply in thread
react
forward
```

The rotor should offer:
```
[message bubble rotor actions] {
reply
delete
reply in thread
react
forward
}
```

stefanceriu:
Hi, thanks for raising this. Unfortunately, the current APIs don't allow accessibility actions to be computed on the fly when highlighting a particular item and require them available upfront. That would mean computing them for every message displayed in the timeline which is too computationally expensive.

[...]

PR #1907, opened 2023-10-16
Enable VoiceOver in timeline menu

alfogrillo:
Enabled the voice over on the timeline menu to align to EX Android


pixlwave:
Ohhhh interesting. So whilst screen was completely broken (🙈), I'm not sure these changes end up helping the user that much as you also can't access this sheet via VoiceOver. If it's within scope for you, we should probably be adding support for all the actions in the rotor. I don't think it would end up being too different from how we build the mac context menu in terms of how it fits in.

Related issue: https://github.com/vector-im/element-x-ios/issues/660

pixlwave (approved):
Whoops I clicked the wrong one before.

alfogrillo:
> Ohhhh interesting. So whilst screen was completely broken (🙈), I'm not sure these changes end up helping the user that much as you also can't access this sheet via VoiceOver. If it's within scope for you, we should probably be adding support for all the actions in the rotor. I don't think it would end up being too different from how we build the mac context menu in terms of how it fits in.
> 
> Related issue: #660


> I'm not sure these changes end up helping the user that much as you also can't access this sheet via VoiceOver.

I think we technically can (I did it), but the trigger (the long press) will be still difficult to action for blind people. 😕

>  If it's within scope for you, we should probably be adding support for all the actions in the rotor.

Nope, I think it's out of scope. I wanted at least to have the VoiceOver reading the options in the menu like EXA is doing.

PR #1912, opened 2023-10-17
Initial service implementation for using a PIN code.

pixlwave:
This PR implements the AppLockService for PIN codes, storing them in the keychain and allowing a grace period for when the app should be unlocked. It does not include TouchID/FaceID support yet, nor is there any UI yet.

### DIFF
--- a/ElementX/Sources/Application/AppSettings.swift
+++ b/ElementX/Sources/Application/AppSettings.swift
@@ -117,6 +117,13 @@ final class AppSettings {
     /// An email address that should be used for support requests.
     let supportEmailAddress = "support@element.io"
     
+    // MARK: - Security
+    
+    /// The amount of time the app can remain in the background for without requesting the PIN/TouchID/FaceID.
+    let appLockGracePeriod: TimeInterval = 180
+    /// Any codes that the user isn't allowed to use for their PIN.
+    let appLockPINCodeBlockList = ["0000", "1234"]
+    
     // MARK: - Authentication
     
     /// The URL that is opened when tapping the Learn more button on the sliding sync alert during authentication.
--- a/ElementX/Sources/Services/AppLock/AppLockService.swift
+++ b/ElementX/Sources/Services/AppLock/AppLockService.swift
@@ -16,14 +16,31 @@
 
 import LocalAuthentication
 
+enum AppLockServiceError: Error {
+    /// The operation failed to access the keychain.
+    case keychainError
+    /// The PIN code was rejected as an insecure choice.
+    case weakPIN
+}
+
 @MainActor
 protocol AppLockServiceProtocol {
     /// The app has been configured to automatically lock with a PIN code.
     var isEnabled: Bool { get }
-    /// The app can additionally be unlocked using FaceID or TouchID.
-    var supportsBiometrics: Bool { get }
+    /// The type of biometric authentication supported by the device.
+    var biometryType: LABiometryType { get }
+    /// Whether or not the user has enabled unlock via TouchID, FaceID or (possibly) OpticID.
+    var biometricUnlockEnabled: Bool { get set }
     /// The app should be unlocked with a PIN code/biometrics before being presented.
     var needsUnlock: Bool { get }
+    
+    /// Sets the user's PIN code used to unlock the app.
+    func setupPINCode(_ pinCode: String) -> Result<Void, AppLockServiceError>
+    /// Disables the App Lock feature, removing the user's stored PIN code.
+    func disable()
+    
+    /// Informs the service that the app has entered the background.
+    func applicationDidEnterBackground()
     
     /// Attempt to unlock the app with the supplied PIN code.
     func unlock(with pinCode: String) -> Bool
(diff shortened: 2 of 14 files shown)


ElementX/Sources/Services/AppLock/AppLockTimer.swift
```diff
@@ -0,0 +1,62 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+
+/// A timer that adds a grace-period to the app before locking it.
+struct AppLockTimer {
```
pixlwave, line 20: I'm split on whether this should really be a class, so very much open to feedback on that.

ElementX/Sources/Services/AppLock/AppLockService.swift
```diff
@@ -16,40 +16,105 @@
...
     
     init(keychainController: KeychainControllerProtocol, appSettings: AppSettings) {
         self.keychainController = keychainController
         self.appSettings = appSettings
+        timer = AppLockTimer(gracePeriod: appSettings.appLockGracePeriod)
+    }
+    
+    func setupPINCode(_ pinCode: String) -> Result<Void, AppLockServiceError> {
+        do {
+            guard !appSettings.appLockPINCodeBlockList.contains(pinCode) else { return .failure(.weakPIN) }
```
stefanceriu, line 85: You might want more validations here, length, isEmpty, isNumeric etc etc

ElementX/Sources/Services/AppLock/AppLockTimer.swift
stefanceriu: I don't think it matters either way. What a better question I think is whether this even needs to live in its file of its own or it can just be part of the Service.

stefanceriu: On the other hand, if you do want to keep it separate, maybe it should be completely generic: `AppBackgroundGraceTimer` or something, that just times how long since the app was in the backround.

ElementX/Sources/Application/AppSettings.swift
```diff
@@ -117,6 +117,13 @@ final class AppSettings {
     /// An email address that should be used for support requests.
     let supportEmailAddress = "support@element.io"
     
+    // MARK: - Security
+    
+    /// The amount of time the app can remain in the background for without requesting the PIN/TouchID/FaceID.
+    let appLockGracePeriod: TimeInterval = 180
+    /// Any codes that the user isn't allowed to use for their PIN.
+    let appLockPINCodeBlockList = ["0000", "1234"]
```
Velin92, line 125: I would make this not a stored check but a computed check.

I think what constitutes a pin as not being safe is when:
- It contains at least 3 occurrences of the same number (3x33 or 222x and so on), this is to avoid the chances of guessing being 1/40 
- it contains a full sequence of number descending or ascending (1234, 2345, 9876), this is to avoid the chance of guessing to be 1/16
- It should not contain 2 occurrences of the same number that are adjacent (22xx, x33x), this is quite strict, so I don't think is necessary but worth mentioning this makes the guessing chance to 1/300

ElementX/Sources/Services/AppLock/AppLockService.swift
Velin92, line 85: comment above, this should be computed using an algorithm to detect the patterns explained above

ElementX/Sources/Services/AppLock/AppLockTimer.swift
Velin92: Since this contains a state I would prefer it to be a class instead of a struct

ElementX/Sources/Services/AppLock/AppLockTimer.swift
```diff
@@ -0,0 +1,62 @@
...
+    
+    /// Signals to the timer to track how long the app will be backgrounded for.
+    mutating func applicationDidEnterBackground(date: Date = .now) {
+        // Only update the last background date if the app is currently unlocked.
+        guard !isLocked else { return }
+        lastUnlockedBackground = date
+    }
+    
+    /// Asks the timer whether the app should be unlocked to be accessed.
+    mutating func needsUnlock(date: Date = .now) -> Bool {
```
Velin92, line 45: I don't like a mutating func that also returns a value, also the name does not imply an internal change of the state.

ElementX/Sources/Services/AppLock/AppLockTimer.swift
pixlwave: Ok class it is. I split it out so it could be tested on its own, it can't be generic, because it needs to track how long the app was backgrounded for since using an unlocked session. Which would be different to background to background checking.

ElementX/Sources/Services/AppLock/AppLockService.swift
pixlwave: Ok fair, I kind of assumed the input screens should handle validation like that, but it certainly won't hurt.

ElementX/Sources/Application/AppSettings.swift
pixlwave: Unfortunately this is part of the requirements to have a list: https://github.com/vector-im/element-meta/issues/2129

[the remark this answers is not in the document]
pixlwave: These are just placeholder values for now.

Velin92: I think this is a bit dumb, however if the list is regex based this should be more robust, since we can compute way more edge cases all at ones.

ElementX/Sources/Services/AppLock/AppLockTimer.swift
pixlwave: Updated with clearer naming.

ElementX/Sources/Application/AppSettings.swift
pixlwave: Sounds like a reasonable idea. Comment in that linked issue if you want to, the end solution will ultimately need to allow for the list to be set on the homeserver so there's more to this than just client side stuff.

Velin92 (approved):

PR #1916, opened 2023-10-18
Debug is no longer crashing as of matrix-rust-sdk/pull/2595, time to …

stefanceriu:
…make it the default again so that debugging works properly

https://github.com/matrix-org/matrix-rust-sdk/pull/2595


nimau (approved):
SGTM

PR #1917, opened 2023-10-18
Add App Lock settings screen

pixlwave:
Adds the App Lock settings screen. This is the screen only, more screens are needed before the flow will work. This branch is based upon #1912.

Screenshot 2023-10-18 at 10 05 14 am


ElementX/Sources/Screens/AppLock/AppLockSettingsScreen/AppLockSettingsScreenViewModel.swift
```diff
@@ -0,0 +1,69 @@
...
+        state.bindings.alertInfo = .init(id: .confirmRemovePINCode,
+                                         title: UntranslatedL10n.screenAppLockSettingsRemovePinAlertTitle,
+                                         message: UntranslatedL10n.screenAppLockSettingsRemovePinAlertMessage,
+                                         primaryButton: .init(title: L10n.actionYes) { self.completeRemovePIN() },
+                                         secondaryButton: .init(title: L10n.actionCancel, role: .cancel, action: nil))
+    }
+    
+    /// Removes the user's PIN code, disabling the App Lock feature.
+    private func completeRemovePIN() {
+        appLockService.disable()
```
stefanceriu, line 66: Shouldn't we ask the user to input the pin again here?

stefanceriu (approved):
lgtm 👍

ElementX/Sources/Screens/AppLock/AppLockSettingsScreen/AppLockSettingsScreenViewModel.swift
pixlwave: So in the flow, this whole screen will be protected by the user first re-entering their PIN. I need to build that screen before I can insert it into the flow though 🙈

SpyHoodle:
Shouldn't this page be titled "PIN settings" or something, instead of just Settings, as you'd be going back to Settings?
image

PR #1919, opened 2023-10-18
Voice message waveform from audio files

alfogrillo:
This PR introduces the dependency [DSWaveformImage](https://github.com/dmrschmidt/DSWaveformImage) for rendering waveforms from audio files.

The waveform in the timeline will work like this:
- An estimated waveform (inside the matrix event) is shown in the timeline first.
- As soon as the user plays the voice message, a more accurate waveform is displayed using the actual audio file.
  
**Result**
poc


Velin92 (approved):
LGTM

PR #1929, opened 2023-10-20
Send Intentional Mentions

Velin92:
This PR also updates the Rust SDK and RTE.

Now we can fetch the current state of mentions from RTE and when sending a message we can fetch this state, and use it also add intentional message to the content.

### DIFF
--- a/ElementX/Sources/Services/Room/RoomProxyProtocol.swift
+++ b/ElementX/Sources/Services/Room/RoomProxyProtocol.swift
@@ -96,7 +96,10 @@ protocol RoomProxyProtocol {
     
     func sendMessageEventContent(_ messageContent: RoomMessageEventContentWithoutRelation) async -> Result<Void, RoomProxyError>
     
-    func sendMessage(_ message: String, html: String?, inReplyTo eventID: String?) async -> Result<Void, RoomProxyError>
+    func sendMessage(_ message: String,
+                     html: String?,
+                     inReplyTo eventID: String?,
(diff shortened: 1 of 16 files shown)


ElementX/Sources/Services/Room/RoomProxyProtocol.swift
```diff
@@ -96,7 +96,10 @@ protocol RoomProxyProtocol {
     
     func sendMessageEventContent(_ messageContent: RoomMessageEventContentWithoutRelation) async -> Result<Void, RoomProxyError>
     
-    func sendMessage(_ message: String, html: String?, inReplyTo eventID: String?) async -> Result<Void, RoomProxyError>
+    func sendMessage(_ message: String,
+                     html: String?,
+                     inReplyTo eventID: String?,
+                     intentionalMentions: Mentions) async -> Result<Void, RoomProxyError>
```
stefanceriu, line 102: Shouldn't we keep our own `IntentionalMentions` at this level and only drop to Rust's internally?

stefanceriu (approved):
Noice! 👌

ElementX/Sources/Services/Room/RoomProxyProtocol.swift
Velin92: I imagined the RoomProxy being the lowest possible level should have used the rust types?
I see it uses rust types as `MediaInfo` `SendAttachmentJoinHandleProtocol` and `RoomMessageEventContentWithoutRelation`

Happy to change that ofc, I just thought we wanted to keep the rust types in the proxy

PR #1930, opened 2023-10-20
Add AppLockSetupPINScreen.

pixlwave:
This PR creates the AppLockSetupPINScreen. Not connected to a flow yet.

test_appLockSetupPINScreen Create

### DIFF
--- a/ElementX/Sources/Screens/AppLock/AppLockSetupPINScreen/View/PINTextField.swift
+++ b/ElementX/Sources/Screens/AppLock/AppLockSetupPINScreen/View/PINTextField.swift
@@ -1,0 +1,124 @@
+//
+// Copyright 2023 New Vector Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import SwiftUI
+
+/// A text field that enables secure entry of a numerical PIN code.
+/// The view itself handles validation and the base text field type.
+struct PINTextField: View {
+    @Binding var pinCode: String
+    var isSecure = false
+    
+    var body: some View {
+        textField
+            .textFieldStyle(PINTextFieldStyle(pinCode: pinCode, isSecure: isSecure))
+            .keyboardType(.numberPad)
+            .onChange(of: pinCode) { newValue in
+                // add some tests.
+                let sanitized = sanitize(newValue)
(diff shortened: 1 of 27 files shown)


ElementX/Sources/Screens/AppLock/AppLockSetupPINScreen/View/PINTextField.swift
```diff
@@ -0,0 +1,124 @@
...
+    var isSecure = false
+    
+    var body: some View {
+        textField
+            .textFieldStyle(PINTextFieldStyle(pinCode: pinCode, isSecure: isSecure))
+            .keyboardType(.numberPad)
+            .onChange(of: pinCode) { newValue in
+                // add some tests.
+                let sanitized = sanitize(newValue)
+                if sanitized != newValue { pinCode = sanitized }
```
stefanceriu, line 32: Is this really necessary, would it trigger a change if they're equal?

ElementX/Sources/Screens/AppLock/AppLockSetupPINScreen/View/PINTextField.swift
```diff
@@ -0,0 +1,124 @@
...
+    @ViewBuilder
+    var textField: some View {
+        if isSecure {
+            SecureField("", text: $pinCode)
+        } else {
+            TextField("", text: $pinCode)
+        }
+    }
+    
+    func sanitize(_ pinCode: String) -> String {
```
stefanceriu, line 45: This reminds me, should we disable pasting on this text field?

ElementX/Sources/Screens/AppLock/AppLockSetupPINScreen/View/PINTextField.swift
```diff
@@ -0,0 +1,124 @@
...
+struct PINTextField: View {
+    @Binding var pinCode: String
+    var isSecure = false
+    
+    var body: some View {
+        textField
+            .textFieldStyle(PINTextFieldStyle(pinCode: pinCode, isSecure: isSecure))
+            .keyboardType(.numberPad)
+            .onChange(of: pinCode) { newValue in
+                // add some tests.
```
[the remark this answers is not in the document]
pixlwave: Oh yeah, already done 😁

ElementX/Sources/Screens/AppLock/AppLockSetupPINScreen/View/PINTextField.swift
pixlwave: Its not possible to paste via the long press menu. I hadn't thought about cmd + c / v until just now. Seems it isn't possible in SwiftUI, you need a custom `UITextView` which is probably overkill here. I think I'm happy sanitising the input if someone does try to paste with a shortcut.

ElementX/Sources/Screens/AppLock/AppLockSetupPINScreen/View/PINTextField.swift
pixlwave: Well it *would* recompute the view needlessly if they're equal but wouldn't trigger a change in here. I think its fine to leave as is. I might add a warning log though that invalid characters were detected.

stefanceriu (approved):

PR #1935, opened 2023-10-23
Update DSWaveformImage to 14.1.1

alfogrillo:
This PR updates DSWaveformImage to 14.1.1 which is not compatible with Xcode 14 (undoing this [fix](https://github.com/vector-im/element-x-ios/pull/1928)).


stefanceriu (approved):
👏

Issue #1936, opened 2023-10-23
OIDC redirect sometimes fails to be picked up.


pixlwave:
### Steps to reproduce

1. Sign out of the app
2. Choose a homeserver with OIDC configured
3. Try to sign in to the app.

#### What did you expect?

Sign in would complete.

#### What happened instead?

Sign in almost finishes, but instead of the redirect being picked up by the app, the `ASWebAuthenticationSession` navigates to the page (which doesn't exists so it 404s).

#### How to fix

- Initially lets revert back to using `io.element:/oidc` as the redirect URL, as the WAS handles the redirection itself and we know it works reliably
- As we would really like to use the universal link, we should host a fallback HTML page on http://mobile.element.io/ which opens up the app with the custom URI scheme. This will need additional parsing added to the URL routes on the app side.

### Your phone model

N/A

### Operating system version

iOS 17

### Application version

1.3.4

### Will you send logs?

No

pixlwave:
Adding a note that iOS 17.4 deprecated the `ASWebAuthenticationSession` init with a scheme, replacing it with a URL: https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession/init(url:callback:completionhandler:)-6nut7

We should try this again as it sounds like we won't need the fallback page and can simply use the scheme on older devices and the URL on newer ones.

PR #1681, opened 2023-09-11
Use a universal link for the OIDC callback.

pixlwave:
This PR updates OIDC the OIDC callback to use a universal link of `mobile.element.io/oidc/appvariant`. Changes and notes:
- For some reason using `element.io/mobile/oidc/appvariant` was unreliable.
- `/appvariant` is added to avoid e.g. the Nightly build opening the link from a PR build (see MRs on `mobile-web-fallback` for more information).
- Universal links aren't captured by the ASWebAuthenticationSession, so the OIDC presenter was tweaked to accept completion externally.
- The state machine is tweaked so that the AppCoordinator can tell if it should forward the universal link to the authentication flow or the soft logout coordinator (its quite possible that OIDC isn't configured to signal a soft logout through Rust, and there may be a more efficient way to re-authenticate when this occurs).

Marking this PR as ready although it will need another deployment of mobile.element.io and a final round of testing to make sure EI doesn't hijack the callback links.

Fixes #1734.

PR #1937, opened 2023-10-23
Revert the OIDC callback to use a custom scheme for now.

pixlwave:
Universal links are still slightly unreliable for the OIDC callback. This PR is the first part of #1936 and a partial revert of #1681. Will follow this up in November once we have time to test hosting a page with a redirect as a fallback.


stefanceriu (approved):
👍

PR #1939, opened 2023-10-23
Fix EstimatedWaveform amplitudes

alfogrillo:
This PR fixes the normalization computation for an `EstimatedWaveform`.


nimau (approved):
Nice, the waveform looks much better that way.

PR #1941, opened 2023-10-23
Make the reldbg profile default for SDK builds again

stefanceriu:
Following https://github.com/matrix-org/matrix-rust-sdk/pull/2765 we need to switch back to reldbg which not contains debug symbols but also doesn't crash

Revert "Debug is no longer crashing as of matrix-rust-sdk/pull/2595time to make it the default again so that debugging works properly"

This reverts commit eb9c69ce20a05ebabee7a9580e1a8a7f8cc3ebb9.


alfogrillo (approved):
LGTM

PR #1943, opened 2023-10-23
Fix Suggestion Pattern when text is empty

Velin92:
The user suggestions should appear even when there is not text yet after the @ character. This PR fixes this issue. Also it improves debouncing, by only doing that when the suggestion pattern is not nil.


stefanceriu (approved):
👏

PR #1948, opened 2023-10-24
Fix playback and recording of voice messages

nimau:
This PR fixes
- a problem where switching from one voice message playback to another did not update the user interface accordingly
- an issue where briefly pressing the microphone button while a playback was in progress did not cancel the recording.


ElementX/Sources/Screens/ComposerToolbar/View/ComposerToolbar.swift
```diff
@@ -229,10 +229,10 @@ struct ComposerToolbar: View {
             context.send(viewAction: .startVoiceMessageRecording)
         } stopRecording: {
             if let voiceMessageRecordingStartTime, Date.now.timeIntervalSince(voiceMessageRecordingStartTime) < voiceMessageMinimumRecordingDuration {
-                context.send(viewAction: .cancelVoiceMessageRecording)
                 withAnimation {
                     showVoiceMessageRecordingTooltip = true
                 }
+                context.send(viewAction: .cancelVoiceMessageRecording)
```
alfogrillo, line 235: Is it part of the fix or you are just cleaning up the code?
Unrelated: we should use `withElementAnimation` and not `withAnimation` whenever we can (on line 232).

ElementX/Sources/Services/Audio/Recorder/AudioRecorder.swift
```diff
@@ -42,7 +42,7 @@ class AudioRecorder: NSObject, AudioRecorderProtocol, AVAudioRecorderDelegate {
         audioRecorder?.isRecording ?? false
     }
     
-    func record(with recordID: AudioRecordingIdentifier) async -> Result<Void, AudioRecorderError> {
+    @MainActor func record(with recordID: AudioRecordingIdentifier) async -> Result<Void, AudioRecorderError> {
```
alfogrillo, line 45: At this point: is it better to apply the `@MainActor` to the class `AudioRecorder` instead of repeating it each time?

ElementX/Sources/Services/MediaPlayer/MediaPlayerProvider.swift
```diff
@@ -48,30 +48,28 @@ class MediaPlayerProvider: MediaPlayerProviderProtocol {
         return audioPlayerStates[audioPlayerStateID]
     }
     
-    @MainActor
-    func register(audioPlayerState: AudioPlayerState) {
+    @MainActor func register(audioPlayerState: AudioPlayerState) {
```
alfogrillo, line 51: Same `@MainActor` consideration for the whole class (or maybe for the protocol `MediaPlayerProviderProtocol`) here. wdyt?

ElementX/Sources/Screens/ComposerToolbar/View/ComposerToolbar.swift
nimau: 👍

ElementX/Sources/Services/Audio/Recorder/AudioRecorder.swift
nimau: Ok, instead of using the `@MainActor` to fix concurrency issues, I've added an internal `DispatchQueue` for the `AudioRecorder`

ElementX/Sources/Services/MediaPlayer/MediaPlayerProvider.swift
nimau: Yes, I think it's a good idea to define the `MediaPlayerProviderProtocol` `@MainActor`.

ElementX/Sources/Services/Audio/Recorder/AudioRecorder.swift
```diff
@@ -42,42 +42,45 @@ class AudioRecorder: NSObject, AudioRecorderProtocol, AVAudioRecorderDelegate {
...
-            audioRecorder = try AVAudioRecorder(url: url, settings: settings)
-            audioRecorder?.delegate = self
-            audioRecorder?.isMeteringEnabled = true
-            audioRecorder?.record()
-            actionsSubject.send(.didStartRecording)
-        } catch {
-            MXLog.error("audio recording failed: \(error)")
-            actionsSubject.send(.didFailWithError(error: error))
+        return await withCheckedContinuation { continuation in
+            startRecording(with: recordID) { [weak self] result in
```
alfogrillo, line 54: If you write another `async` overload for `startRecording` the code will be more compact.
I mean create something like this
```swift
    private func startRecording(with recordID: AudioRecordingIdentifier) async -> Result<Void, AudioRecorderError> {
        await withCheckedContinuation { continuation in
            startRecording(with: recordID) { result in
                continuation.resume(returning: result)
            }
        }
    }
```

And call it here.
Wdyt?

alfogrillo (approved):
Looks good, I left just a comment to simplify the code.

stefanceriu:
[Fix EncryptedHistoryRoomTimelineView test preview](https://github.com/vector-im/element-x-ios/pull/1948/commits/3bdf9aacd83b702168892b3e05ef0a8fdad141a0)

Oupsie, thanks Nicolas! 😅

PR #1949, opened 2023-10-24
Initial implementation for AppLockSetupFlowCoordinator.

pixlwave:
This PR contains 2 commits:
- Add an `AppLockSetupFlowCoordinator` which is presented either as part of the authentication flow (when the PIN is mandatory) or from the Settings screen (when the PIN is optional). This commit also renames AppLockSettingsScreen to AppLockSetupSettingsScreen.
- Implement the forceLogout action to when the user has entered their PIN incorrectly 3 times.


nimau (approved):
LGTM 👍

PR #1953, opened 2023-10-24
Update permission alerts

alfogrillo:
This PR update system and custom alert for asking permission for the microphone and location.

| First time asking permission | Asking the permission again |
| --- | --- |
| mic-first|mic-later|
|location-first|location-later|


callumu:
A few comments:

- Can we use black as the text for all alerts instead of the system blue?
- Can you update the copy of the alert with the map [to this](https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X?type=design&node-id=13665%3A29230&mode=design&t=OhDPeO5zlHazekA2-1)
- I've also changed all instances of 'provide' to 'grant' as I feel like it's better English for the use case. Can you update across all four?

alfogrillo:
> * Can we use black as the text for all alerts instead of the system blue?

Nope. These are owned by iOS. We **can't** change the alert title (in bold) as well.

> * Can you update the copy of the alert with the map [to this](https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X?type=design&node-id=13665%3A29230&mode=design&t=OhDPeO5zlHazekA2-1)

Done. I updated the screenshots in this PR as well.

> * I've also changed all instances of 'provide' to 'grant' as I feel like it's better English for the use case. Can you update across all four?

Done. I updated the screenshots in this PR as well.

callumu:
Great, thanks. This all looks good now

PR #1954, opened 2023-10-24
Session verification fixes

### LINKED ISSUE #1868: Incorrectly prompted to verify session
### Steps to reproduce

1. Open the app and see the prompt to complete verification
2. I look at my list of sessions in Element Desktop and see that the X session is already verified.
4. I tap on the prompt to verify session anyway in order to try and get rid of the prompt
5. I attempt to proceed with emoji verification against Element Desktop
6. I get a message saying Cancelled

#### What did you expect?

Not to be prompted to verify my session as already verified.

#### What happened instead?

I was prompted to verify my session and when I attempted to complete the steps I got an error saying verification was cancelled.

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

1.3.1 (74)

### Homeserver

matrix.org

### Will you send logs?

Yes

stefanceriu:
Fixes #1868: Incorrect `is_verified` flag after successfully running verification flow
- the inner user_identity isn't automatically updated when the flow finishes, needs to be fetched again from encryption
- also replaces `UserIdentity.is_verified` with `Device.is_cross_signed_by_owner`
- depends on matrix-org/matrix-rust-sdk/pull/2775

### DIFF
--- a/ElementX/Sources/Services/Session/UserSession.swift
+++ b/ElementX/Sources/Services/Session/UserSession.swift
@@ -57,14 +57,19 @@ class UserSession: UserSessionProtocol {
         Task {
             switch await clientProxy.sessionVerificationControllerProxy() {
             case .success(let sessionVerificationController):
-                tearDownSessionVerificationControllerWatchdog()
-                
-                if !sessionVerificationController.isVerified {
-                    callbacks.send(.sessionVerificationNeeded)
+                guard case let .success(isVerified) = await sessionVerificationController.isVerified() else {
+                    MXLog.error("Failed checking verification state. Will retry on the next sync update.")
+                    return
                 }
                 
-                self.sessionVerificationController = sessionVerificationController
+                tearDownSessionVerificationControllerWatchdog()
                 
+                if isVerified == false {
+                    callbacks.send(.sessionVerificationNeeded)
+                }
+
+                self.sessionVerificationController = sessionVerificationController
+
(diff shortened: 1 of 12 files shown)


ElementX/Sources/Services/Session/UserSession.swift
```diff
@@ -57,14 +57,19 @@ class UserSession: UserSessionProtocol {
...
                 
-                if !sessionVerificationController.isVerified {
+                if isVerified == false {
                     callbacks.send(.sessionVerificationNeeded)
                 }
-                
+
                 self.sessionVerificationController = sessionVerificationController
-                
+
```
pixlwave, line 72: Probably unintentional.

ElementX/Sources/Services/Session/UserSession.swift
```diff
@@ -57,14 +57,19 @@ class UserSession: UserSessionProtocol {
...
             case .success(let sessionVerificationController):
+                guard case let .success(isVerified) = await sessionVerificationController.isVerified() else {
+                    MXLog.error("Failed checking verification state. Will retry on the next sync update.")
+                    return
+                }
+                
                 tearDownSessionVerificationControllerWatchdog()
                 
-                if !sessionVerificationController.isVerified {
+                if isVerified == false {
```
pixlwave, line 67: Would have thought some tool would pick this up if its not optional which I don't think it is.
```suggestion
                if !isVerified {
```

PR #1955, opened 2023-10-24
Suggestions view design improvements

Velin92:
NOTE: the snapshot tests incorrectly render the last divider even if is hidden in the real preview and when the app is built, likely a snapshotting bug.

This PR also sets the max width of a pill using the line fragment of the text view.

Simulator Screenshot - iPhone 14 - 2023-10-24 at 17 48 58

### DIFF
--- a/ElementX/Sources/Screens/ComposerToolbar/View/CompletionSuggestionView.swift
+++ b/ElementX/Sources/Screens/ComposerToolbar/View/CompletionSuggestionView.swift
@@ -68,9 +68,8 @@ struct CompletionSuggestionView: View {
             .listRowInsets(.init(top: 0, leading: Constants.leadingPadding, bottom: 0, trailing: 0))
         }
         .listStyle(PlainListStyle())
-        .frame(height:
-            min(contentHeightForRowCount(Constants.maxVisibleRows),
-                contentHeightForRowCount(items.count)))
+        .frame(height: min(contentHeightForRowCount(Constants.maxVisibleRows),
+                           contentHeightForRowCount(items.count)))
         .background(Color.compound.bgCanvasDefault)
     }
     
--- a/ElementX/Sources/Screens/ComposerToolbar/View/CompletionSuggestionView.swift
+++ b/ElementX/Sources/Screens/ComposerToolbar/View/CompletionSuggestionView.swift
@@ -25,7 +25,7 @@ struct CompletionSuggestionView: View {
         static let topPadding: CGFloat = 8.0
         static let listItemPadding: CGFloat = 6.0
         // added by the list itself when presenting the divider
-        static let linePadding: CGFloat = 4.0
+        static let listItemSpacing: CGFloat = 4.0
         static let leadingPadding: CGFloat = 16.0
         static let maxVisibleRows = 4
     }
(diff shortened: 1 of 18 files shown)


pixlwave (approved):
📏

ElementX/Sources/Screens/ComposerToolbar/View/MentionSuggestionItemView.swift
```diff
@@ -21,13 +21,13 @@ struct MentionSuggestionItemView: View {
     let item: MentionSuggestionItem
     
     var body: some View {
-        HStack(alignment: .center) {
+        HStack(alignment: .center, spacing: 16) {
             LoadableAvatarImage(url: item.avatarURL,
                                 name: item.displayName,
                                 contentID: item.id,
-                                avatarSize: .custom(42),
+                                avatarSize: .custom(32),
```
pixlwave, line 28: Maybe we should define this one as not custom?

ElementX/Sources/Screens/ComposerToolbar/View/CompletionSuggestionView.swift
```diff
@@ -23,19 +23,11 @@ struct CompletionSuggestionView: View {
     
     private enum Constants {
         static let topPadding: CGFloat = 8.0
-        static let listItemPadding: CGFloat = 4.0
-        static let lineSpacing: CGFloat = 10.0
-        static let maxHeight: CGFloat = 300.0
+        static let listItemPadding: CGFloat = 6.0
+        // added by the list itself when presenting the divider
+        static let linePadding: CGFloat = 4.0
```
pixlwave, line 28: Maybe we could call this one `listItemSpacing`, but its great you added the comment 👍

ElementX/Sources/Screens/ComposerToolbar/View/CompletionSuggestionView.swift
```diff
@@ -73,16 +65,17 @@ struct CompletionSuggestionView: View {
...
             .modifier(ListItemPaddingModifier(isFirst: items.first?.id == item.id))
+            .listRowInsets(.init(top: 0, leading: Constants.leadingPadding, bottom: 0, trailing: 0))
         }
         .listStyle(PlainListStyle())
-        .frame(height: min(Constants.maxHeight,
-                           min(contentHeightForRowCount(Constants.maxVisibleRows),
-                               contentHeightForRowCount(items.count))))
+        .frame(height:
+            min(contentHeightForRowCount(Constants.maxVisibleRows),
+                contentHeightForRowCount(items.count)))
```
pixlwave, line 73:
```suggestion
        .frame(height: min(contentHeightForRowCount(Constants.maxVisibleRows),
                           contentHeightForRowCount(items.count)))
```

ElementX/Sources/Screens/ComposerToolbar/View/CompletionSuggestionView.swift
```diff
@@ -23,19 +23,11 @@ struct CompletionSuggestionView: View {
     
     private enum Constants {
```
pixlwave, line 24: I'd be up for extracting these from this enum as simple stored `private let`s like we do elsewhere if you want to?

pixlwave (approved):
Sorry first review only showed me the last commit. Looks good to me 👍

ElementX/Sources/Screens/ComposerToolbar/View/CompletionSuggestionView.swift
Velin92: I think is because they are used by other internal private classes in the file

PR #1956, opened 2023-10-24
Add a progress indicator to voice message preview

nimau:
This PR adds a progress indicator  to `VoiceMessagePreviewComposer`

This indicator is displayed during playback and scrubbing.

Screenshot 2023-10-24 at 19 09 21


ElementX/Sources/Other/VoiceMessage/WaveformProgressCursorModifier.swift
```diff
@@ -0,0 +1,57 @@
...
+//
+
+import Compound
+import SwiftUI
+
+extension View {
+    func waveformProgressCursor(progress: CGFloat,
+                                color: Color = .compound.iconAccentTertiary,
+                                width: CGFloat = 2,
+                                opacity: CGFloat = 1.0) -> some View {
```
alfogrillo, line 24: Cool!
In theory this could have been even more generic taking a `@ViewBuilder` closure for the actual cursor view.
At that point the name could have been `func progressCursor` since the more generic purpose.
But I don't know if this will be really helpful in practice.

Furthermore: instead of having a progress and an opacity we may have a `progress: CGFloat?` and use `nil` to hide the cursor?

nimau: Good idea for passing a cursor view. I've updated the code, so we can now use a custom CursorView if we want.
Regarding the progress being optional, I think it is easier and clearer to keep the opacity instead.

ElementX/Sources/Other/VoiceMessage/WaveformProgressCursorModifier.swift
```diff
@@ -0,0 +1,45 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Compound
+import SwiftUI
+
+extension View {
+    func waveformProgressCursor<CursorView: View>(progress: CGFloat,
```
alfogrillo, line 21:
```suggestion
    func progressCursor<CursorView: View>(progress: CGFloat,
```

This isn't now really linked to waveforms, it can be used anywhere! ;)

ElementX/Sources/Other/VoiceMessage/WaveformProgressCursorModifier.swift
```diff
@@ -0,0 +1,45 @@
...
+                                                  width: CGFloat = 2,
+                                                  cursorView: @escaping () -> CursorView) -> some View {
+        modifier(WaveformProgressCursorModifier(progress: progress,
+                                                width: width,
+                                                cursorView: cursorView))
+    }
+}
+
+private struct WaveformProgressCursorModifier<CursorView: View>: ViewModifier {
```
alfogrillo, line 30:
```suggestion
private struct ProgressCursorModifier<CursorView: View>: ViewModifier {
```

ElementX/Sources/Other/VoiceMessage/WaveformProgressCursorModifier.swift
```diff
@@ -0,0 +1,45 @@
...
+    let progress: CGFloat
+    let width: CGFloat
+    @ViewBuilder var cursorView: () -> CursorView
+    
+    func body(content: Content) -> some View {
+        GeometryReader { geometry in
+            content
+                .overlay(alignment: .leading) {
+                    cursorView()
+                        .offset(CGSize(width: (progress * geometry.size.width) - width / 2, height: 0.0))
```
alfogrillo, line 40: There is no need to handle the `width/2` internally.
You can offset directly the `CursorView` you pass in and remove `width` entirely from this modifier.
wdyt?

nimau: Ok, actually the offset is not needed (without it, the cursor is placed correctly in both the `EstimatedWaveformView` and the `WaveformLiveCanvas`), so I've removed it.

alfogrillo (approved):
LGTM!

PR #1957, opened 2023-10-25
Replies now display mentions with @displayname

### LINKED ISSUE #1831: Mentions: Display mentions in replies
The mention should also be displayed in the timeline for replies, but not as a tappable link (since it should not override the reply tap behaviour).

Velin92:
fixes #1831 

If the display name is not available yet they will use the user id instead


ElementX/Sources/Other/Pills/MessageText.swift
```diff
@@ -67,7 +67,11 @@ struct MessageText: UIViewRepresentable {
         let textView = MessageTextView(usingTextLayoutManager: false)
         textView.roomContext = viewModel
         textView.updateClosure = {
-            attributedString = AttributedString(textView.attributedText)
+            do {
+                attributedString = try AttributedString(textView.attributedText, including: \.elementX)
+            } catch {
+                return
```
pixlwave, line 73: Perhaps we should log the error here. It was useful debugging Xcode 15 URLs. And the same for below too.

ElementX/Sources/Screens/RoomScreen/View/Replies/TimelineReplyView.swift
```diff
@@ -126,8 +126,26 @@ struct TimelineReplyView: View {
...
+                        attributedString.replaceCharacters(in: range, with: "@\(displayName)")
+                    } else {
+                        attributedString.replaceCharacters(in: range, with: userID)
+                    }
+                }
+                
+                if attributes[.MatrixAllUsersMention] as? Bool == true {
+                    attributedString.replaceCharacters(in: range, with: PillConstants.atRoom)
+                }
+            }
```
pixlwave, line 147: Optional: Would be nice to use stick to AttributedString here and use `runs` instead of dropping down to NSAttributedString.

ElementX/Sources/Other/HTMLParsing/AttributedStringBuilder.swift
```diff
@@ -208,22 +211,18 @@ struct AttributedStringBuilder: AttributedStringBuilderProtocol {
...
                 attributedString.addAttribute(.MatrixAllUsersMention, value: true, range: match.range)
+            } else {
+                var link = String(string[matchRange])
+                
+                if linkMatches.contains(match), !link.contains("://") {
+                    link.insert(contentsOf: "https://", at: link.startIndex)
+                }
+                
+                if let url = URL(string: link) {
+                    attributedString.addAttribute(.link, value: url, range: match.range)
```
pixlwave, line 224: If you have time and want to do this in a separate PR, maybe you could try attaching the Matrix IDs as permalinks instead here?

pixlwave (approved):
Nice 👍

ElementX/Sources/Screens/RoomScreen/View/Replies/TimelineReplyView.swift
Velin92: however runs do not have the power to consistetly keep the NSRange when changing piece of the attributed string. While instead the NSString being an object, when enumerated and changed is able to keep consistency between different changes of the string in its enumerated range, does it make sense?

Honestly this AttributedString API is quite useless if you want to do string replacing, is only good mostly for attributed merging/removal/extending

ElementX/Sources/Other/HTMLParsing/AttributedStringBuilder.swift
pixlwave: Yeah, definitely not part of this PR, just thinking if you're looking for something to do whilst waiting for those designs. Because we agreed we should pillify these anyway right? So could well implement it for Xcode 14 now.

Velin92: will try to look at it however tomorrow I'll have the designs and next week I am off for the first two days, so not sure if I'll have the time to look at this

PR #1958, opened 2023-10-25
Fix: do not hide keyboard when recording a voice message

nimau:
With this PR, the keyboard is no longer hidden when the user starts to record a voice message. This prevents the record button from moving when recording starts.

### DIFF
--- a/ElementX/Sources/Screens/ComposerToolbar/View/ComposerToolbar.swift
+++ b/ElementX/Sources/Screens/ComposerToolbar/View/ComposerToolbar.swift
@@ -73,7 +73,7 @@ struct ComposerToolbar: View {
     }
     
     private var topBar: some View {
-        mainContentLayout {
+        topBarLayout {
             mainTopBarContent
 
             if !context.composerActionsEnabled {
@@ -106,14 +106,14 @@ struct ComposerToolbar: View {
         }
     }
     
-    private var mainContentLayout: some Layout {
+    private var topBarLayout: some Layout {
         HStackLayout(alignment: .bottom, spacing: 5)
     }
 
     @ViewBuilder
     private var mainTopBarContent: some View {
         ZStack {
-            mainContentLayout {
+            topBarLayout {
                 if !context.composerActionsEnabled {
                     RoomAttachmentPicker(context: context)
                 }
@@ -246,7 +246,7 @@ struct ComposerToolbar: View {
             VoiceMessageRecordingComposer(recorderState: state)
                 .padding(.leading, 12)
         case .previewVoiceMessage(let state, let waveform, let isUploading) where context.viewState.enableVoiceMessageComposer:
-            mainContentLayout {
+            topBarLayout {
                 voiceMessageTrashButton
                 voiceMessagePreviewComposer(audioPlayerState: state, waveform: waveform)
             }


ElementX/Sources/Screens/ComposerToolbar/View/ComposerToolbar.swift
```diff
@@ -108,28 +108,36 @@ struct ComposerToolbar: View {
...
-            Group {
-                voiceMessageTrashButton
-                voiceMessagePreviewComposer(audioPlayerState: state, waveform: waveform)
-            }
-            .disabled(isUploading)
-        default:
-            if !context.composerActionsEnabled {
-                RoomAttachmentPicker(context: context)
+        ZStack {
+            HStack {
```
alfogrillo, line 112: If you don't match parameters of the HStack in `private var topBar` the layout breaks (try to write multiple lines).
To avoid copy and paste I suggest to create a single layout and use it in both places:
```swift
    private var mainContentLayout: some Layout {
        HStackLayout(alignment: .bottom, spacing: 5)
    }
```

and then you can use it like this
```swift
mainContentLayout {
   viewA
   viewB
}
```

ElementX/Sources/Screens/ComposerToolbar/View/ComposerToolbar.swift
```diff
@@ -108,28 +108,36 @@ struct ComposerToolbar: View {
...
+                HStack {
+                    voiceMessageTrashButton
+                    voiceMessagePreviewComposer(audioPlayerState: state, waveform: waveform)
                 }
-                .padding(.leading, context.composerActionsEnabled ? 7 : 0)
-                .padding(.trailing, context.composerActionsEnabled ? 4 : 0)
+                .disabled(isUploading)
+            default:
+                EmptyView()
+            }
```
alfogrillo, line 140: Maybe it's time to create another private var like this?

```swift
    @ViewBuilder 
    var voiceMessageContent: some View {
        switch context.viewState.composerMode {
        case .recordVoiceMessage(let state) where context.viewState.enableVoiceMessageComposer:
            VoiceMessageRecordingComposer(recorderState: state)
                .padding(.leading, 12)
        case .previewVoiceMessage(let state, let waveform, let isUploading) where context.viewState.enableVoiceMessageComposer:
            HStack {
                voiceMessageTrashButton
                voiceMessagePreviewComposer(audioPlayerState: state, waveform: waveform)
            }
            .disabled(isUploading)
        default:
            EmptyView()
        }
    }
```

ElementX/Sources/Screens/ComposerToolbar/View/ComposerToolbar.swift
nimau: 👍

ElementX/Sources/Screens/ComposerToolbar/View/ComposerToolbar.swift
nimau: 👍

ElementX/Sources/Screens/ComposerToolbar/View/ComposerToolbar.swift
```diff
@@ -105,31 +105,32 @@ struct ComposerToolbar: View {
                 .padding(.leading, 7)
         }
     }
+    
+    private var mainContentLayout: some Layout {
```
alfogrillo, line 109: This name is maybe better (my bad)
```suggestion
    private var topBarLayout: some Layout {
```

alfogrillo (approved):
LGTM!

ElementX/Sources/Screens/ComposerToolbar/View/ComposerToolbar.swift
nimau: 👍

PR #1961, opened 2023-10-25
Fix: Voice recording tooltip location

nimau:
This PR fixes the location of the tooltip (specifically the position of the cursor).


ElementX/Sources/Screens/ComposerToolbar/View/ComposerToolbar.swift
```diff
@@ -86,6 +87,9 @@ struct ComposerToolbar: View {
                         .padding(.leading, 3)
                 } else if context.viewState.enableVoiceMessageComposer {
                     voiceMessageRecordingButton
+                        .background(
+                            ViewFrameReader(frame: $voiceMessageRecordingButtonFrame, coordinateSpace: .global)
+                        )
```
pixlwave, line 92:
```suggestion
                        .background {
                            ViewFrameReader(frame: $voiceMessageRecordingButtonFrame, coordinateSpace: .global)
                        }
```

pixlwave (approved):

PR #1967, opened 2023-10-26
Add some logs to allow debugging of voice messages in rageshake

nimau:
This PR adds some logs to allow debugging of voice messages in rageshake.


pixlwave (approved):
👍

PR #1968, opened 2023-10-26
Removed Pill Avatar + @room is not displayed among suggestions for DM

Velin92:
Also @room is always the first suggestion if available.


nimau (approved):
LGTM 😀

PR #1972, opened 2023-10-27
Handle invalid PIN input in the settings flow.

pixlwave:
This PR records any invalid PIN unlock attempts in the settings screen, syncing this value with the unlock screen, and forcing a logout if 3 failures occur.

Additionally it will restart the create PIN flow if the confirmation PIN fails to match 3 times.


nimau (approved):
LGTM 👍

PR #1980, opened 2023-10-30
Set voice messages max width

alfogrillo:
Useful for limiting the width of voice messages on iPad or landscape mode.


pixlwave (approved):
LGTM 👍

PR #1981, opened 2023-10-30
Fix presentation when not enrolled in Touch/Face ID.

pixlwave:
This PR makes a small tweak to the app lock setup state machine to differentiate between creating a PIN or updating a PIN.


alfogrillo (approved):
LGTM!

PR #1983, opened 2023-10-30
Enable Element Call FF by default.

pixlwave:
- Enabled by default rather than removed for now. Aligns with the same change on Android: https://github.com/vector-im/element-x-android/pull/1685
- Remove the unused FF for mention pills.


nimau (approved):
LGTM

PR #1985, opened 2023-10-30
Fix the delay when switching between voice messages

nimau:
This PR fixes the delay observed when switching between voice messages.

### DIFF
--- a/ElementX/Sources/Services/Audio/Player/AudioPlayer.swift
+++ b/ElementX/Sources/Services/Audio/Player/AudioPlayer.swift
@@ -119,7 +119,6 @@ class AudioPlayer: NSObject, AudioPlayerProtocol {
         isStopped = true
         internalAudioPlayer?.pause()
         internalAudioPlayer?.seek(to: .zero)
-        try? AVAudioSession.sharedInstance().setActive(false)
     }
     
     func seek(to progress: Double) async {
--- a/ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift
+++ b/ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift
@@ -284,16 +284,15 @@ class RoomTimelineController: RoomTimelineControllerProtocol {
         
         let audioPlayerState = audioPlayerState(for: itemID)
         
-        // Detach all other states
-        await mediaPlayerProvider.detachAllStates(except: audioPlayerState)
         // Ensure this one is attached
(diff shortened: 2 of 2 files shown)


ElementX/Sources/Services/Audio/Player/AudioPlayer.swift
```diff
@@ -119,7 +119,6 @@ class AudioPlayer: NSObject, AudioPlayerProtocol {
         isStopped = true
         internalAudioPlayer?.pause()
         internalAudioPlayer?.seek(to: .zero)
-        try? AVAudioSession.sharedInstance().setActive(false)
```
alfogrillo, line 122: Why we don't need this line anymore?

ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift
```diff
@@ -284,16 +284,15 @@ class RoomTimelineController: RoomTimelineControllerProtocol {
...
         // Ensure this one is attached
         if !audioPlayerState.isAttached {
             audioPlayerState.attachAudioPlayer(audioPlayer)
         }
 
+        // Detach all other states
+        await mediaPlayerProvider.detachAllStates(except: audioPlayerState)
+
         guard audioPlayer.mediaSource == source, audioPlayer.state != .error else {
-            audioPlayer.stop()
```
alfogrillo, line 295: Why we don't need this line anymore?

nimau: Because the player is stopped when the current state is detached. So here it is already stopped.

ElementX/Sources/Services/Audio/Player/AudioPlayer.swift
nimau: This is the line that causes the delay, and it works just fine without it.

alfogrillo (approved):
LGTM!

ElementX/Sources/Services/Audio/Player/AudioPlayer.swift
pixlwave: Would it maybe make sense to deactivate the session in a cancellable task with a second or 2 sleep on it? Then if you play again you can cancel the task. I ask because I'm under the impression that leaving this enabled can mess with things like the now playing screen and other apps playing bg audio etc?

PR #1992, opened 2023-10-31
Fix: release audio session after playback ends

nimau:
This PR releases the audio session using a cancellable task when the playback ends

### DIFF
--- a/ElementX/Sources/Services/Audio/Player/AudioPlayer.swift
+++ b/ElementX/Sources/Services/Audio/Player/AudioPlayer.swift
@@ -50,7 +50,13 @@ class AudioPlayer: NSObject, AudioPlayerProtocol {
     private var appBackgroundObserver: NSObjectProtocol?
     private var autoplay = false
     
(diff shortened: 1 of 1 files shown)


ElementX/Sources/Services/Audio/Player/AudioPlayer.swift
```diff
@@ -129,6 +133,38 @@ class AudioPlayer: NSObject, AudioPlayerProtocol {
...
+            try? await Task.sleep(for: .seconds(timeInterval))
+            guard !Task.isCancelled else { return }
+            guard let self else { return }
+            self.releaseAudioSession()
+        }
+    }
+    
+    private func releaseAudioSession() {
+        releaseAudioSessionTask?.cancel()
+        releaseAudioSessionTask = nil
```
pixlwave, line 161: As its using `@CancellableTask` we don't need to call cancel, that will happen automatically.
```suggestion
        releaseAudioSessionTask = nil
```

pixlwave (approved):
LGTM 👍

PR #1993, opened 2023-10-31
Log the app version number at startup.

pixlwave:
Tiny PR, just adds an extra couple of logs of the app version (and the previous version) to help track behaviour across different versions in the rageshakes.


nimau (approved):
LGTM

PR #1963, opened 2023-10-25
Converting detected plain text mentions into permalinks

Velin92:
Works perfectly now

PR #2007, opened 2023-11-01
Adopt Xcode 15

pixlwave:
This PR updates the project to be built with Xcode 15. The following has been done:

- [x] AttributedStringBuilder works with room IDs now (fixed in #1963).
- [x] Media processing tests fixed.
- [x] Preview snapshot tests re-generated, layout fixes applied as necessary. A few caveats:
    - The room screen header is snapshotted as horizontally centred, but in the app its still leading aligned. Not sure I fully understand what is going on here.
    - The waveform icon for an audio message is snapshotted as white, but always shows as iconPrimary in previews.
    - Capsule outlines render with vertical lines in snapshots but are fine on device. Capsule fills have a slight point at either end in snapshots but are fine on device. This is related to rounded corners now defaulting to the `.continuous` style, but why it renders like this in snapshots is beyond me.
- [x] Xcode updated in Fastfile
- [x] UI Tests snapshots re-generated.
    - Same issue with the room screen header, although our tolerances are looser so it was only caught by a single test.
- [x] Integration tests validated.

### Pull Request Checklist

- [ ] Voiceover enabled.


pixlwave:
Closing this PR as stale. CI was too slow with the iOS 17.0.1 simulator to merge this.

PR #2013, opened 2023-11-02
Voice message record button hapitcs

alfogrillo:
This PR:
- Moves voice message specific logics from `ComposerToolbar` to `VoiceMessageRecordingButton`
- Adds haptic feedbacks in `VoiceMessageRecordingButton`

### DIFF
--- a/ElementX/Sources/Other/SwiftUI/Layout/ViewFrameReader.swift
+++ b/ElementX/Sources/Other/SwiftUI/Layout/ViewFrameReader.swift
@@ -42,3 +42,9 @@ struct ViewFrameReader: View {
         }
     }
 }
+
+extension View {
+    func readFrame(_ frame: Binding<CGRect>, in coordinateSpace: CoordinateSpace = .local) -> some View {
+        background(ViewFrameReader(frame: frame, coordinateSpace: coordinateSpace))
+    }
+}
--- a/ElementX/Sources/Screens/ComposerToolbar/View/ComposerToolbar.swift
+++ b/ElementX/Sources/Screens/ComposerToolbar/View/ComposerToolbar.swift
@@ -22,14 +22,14 @@ struct ComposerToolbar: View {
     @ObservedObject var context: ComposerToolbarViewModel.Context
     let wysiwygViewModel: WysiwygComposerViewModel
     let keyCommandHandler: KeyCommandHandler
-
+    
     @FocusState private var composerFocused: Bool
     @ScaledMetric private var sendButtonIconSize = 16
     @ScaledMetric(relativeTo: .title) private var spinnerSize = 44
     @ScaledMetric(relativeTo: .title) private var closeRTEButtonSize = 30
     @State private var frame: CGRect = .zero
     @Environment(\.verticalSizeClass) private var verticalSizeClass
-
+    
     var body: some View {
         VStack(spacing: 8) {
             topBar
@@ -69,7 +69,7 @@ struct ComposerToolbar: View {
     private var topBar: some View {
         topBarLayout {
             mainTopBarContent
-
+            
             if !context.composerActionsEnabled {
                 if context.viewState.isUploading {
                     ProgressView()
@@ -103,7 +103,7 @@ struct ComposerToolbar: View {
     private var topBarLayout: some Layout {
         HStackLayout(alignment: .bottom, spacing: 5)
     }
-
+    
     @ViewBuilder
     private var mainTopBarContent: some View {
         ZStack(alignment: .bottom) {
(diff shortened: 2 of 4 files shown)


ElementX/Sources/Services/Audio/Recorder/AudioRecorder.swift
```diff
@@ -141,8 +141,10 @@ class AudioRecorder: NSObject, AudioRecorderProtocol, AVAudioRecorderDelegate {
                             AVEncoderAudioQualityKey: AVAudioQuality.high.rawValue]
             
             do {
-                try AVAudioSession.sharedInstance().setCategory(.playAndRecord, mode: .default)
-                try AVAudioSession.sharedInstance().setActive(true)
+                let audioSession = AVAudioSession.sharedInstance()
+                try audioSession.setAllowHapticsAndSystemSoundsDuringRecording(true)
```
pixlwave, line 145: Do the haptics get picked up by the mic?

ElementX/Sources/Other/SwiftUI/Layout/ViewFrameReader.swift
```diff
@@ -42,3 +42,9 @@ struct ViewFrameReader: View {
         }
     }
 }
+
+extension View {
+    func readFrame(_ frame: Binding<CGRect>, in coordinateSpace: CoordinateSpace = .local) -> some View {
```
pixlwave, line 47: Good idea 👍

ElementX/Sources/Screens/ComposerToolbar/View/ComposerToolbar.swift
```diff
@@ -22,23 +22,14 @@ struct ComposerToolbar: View {
     @ObservedObject var context: ComposerToolbarViewModel.Context
     let wysiwygViewModel: WysiwygComposerViewModel
     let keyCommandHandler: KeyCommandHandler
-    
+
```
pixlwave, line 25: Nit: These lines should remain indented (this and the one below)

pixlwave (approved):
Much clearer 👏

callumu:
This works well 👍

ElementX/Sources/Screens/ComposerToolbar/View/ComposerToolbar.swift
alfogrillo: What's the rationale?
I thought that as a team we were aligned to the same Xcode settings.

trim

ElementX/Sources/Other/SwiftUI/Layout/ViewFrameReader.swift
alfogrillo: Maybe `struct ViewFrameReader` can become private in the future so that we are forced to use the modifier in the supposed way?

ElementX/Sources/Screens/ComposerToolbar/View/ComposerToolbar.swift
pixlwave: We are, but these are the expected settings:
Screenshot 2023-11-02 at 11 07 04 am

Hence this swiftlint [rule](https://github.com/vector-im/element-x-ios/blob/4af8ef2a5f37a3f80f17a152a6e0806cb4288393/.swiftlint.yml#L2) and this swiftformat [rule](https://github.com/vector-im/element-x-ios/blob/4af8ef2a5f37a3f80f17a152a6e0806cb4288393/.swiftformat#L12).

ElementX/Sources/Other/SwiftUI/Layout/ViewFrameReader.swift
pixlwave: Makes perfect sense to me.

ElementX/Sources/Services/Audio/Recorder/AudioRecorder.swift
alfogrillo: I'm not sure.
There is small noise in the end, but I can hear it even if I comment the release haptic and I write `try audioSession.setAllowHapticsAndSystemSoundsDuringRecording(false)`.
It's not huge, but maybe worth an investigation in a separate PR.

ElementX/Sources/Screens/ComposerToolbar/View/ComposerToolbar.swift
alfogrillo: Cool, I will deselect it for whitespace-only lines 👍🏻

ElementX/Sources/Services/Audio/Recorder/AudioRecorder.swift
alfogrillo: The noise is maybe related with debugging from Xcode. I can't hear it when running the app without the debugger.

ElementX/Sources/Screens/ComposerToolbar/View/ComposerToolbar.swift
pixlwave: Thank you 🙏

ElementX/Sources/Services/Audio/Recorder/AudioRecorder.swift
pixlwave: Nice. It was purely a curiosity but sounds like we're good 👍

PR #2018, opened 2023-11-02
Add deferred spinner for voice messages

alfogrillo:
This PR adds a logic to delay the voice message spinner by two seconds. 
In this way spinners don't show at all when the loading time is less than two seconds.
It also fixes a run time issues caused by the redaction of voice messages.

Tried also to address the problem all in the UI layer [here](https://github.com/vector-im/element-x-ios/compare/alfogrillo/deferred-loader?expand=1).
It works, but maybe the complexity is bigger.

### DIFF
--- a/ElementX/Sources/Other/VoiceMessage/VoiceMessageButton.swift
+++ b/ElementX/Sources/Other/VoiceMessage/VoiceMessageButton.swift
@@ -88,13 +88,13 @@ private struct VoiceMessageButtonStyle: ButtonStyle {
 }
 
 extension VoiceMessageButton.State {
-    init(state: AudioPlayerPlaybackState) {
+    init(_ state: AudioPlayerPlaybackState) {
         switch state {
         case .loading:
             self = .loading
         case .playing:
             self = .playing
-        default:
+        case .stopped, .error, .readyToPlay:
             self = .paused
         }
     }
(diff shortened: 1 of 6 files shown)


pixlwave:
I'll review this properly in the morning, but for now having looked at the second implementation I'm wondering - would it be cleaner to add a `.startedLoaded` state (or similar) that makes the intent clear in the model, and then instead of setting `.loading` directly in `AudioPlayerState`, create a Task with a sleep that starts with `startedLoading` and transitions to `loading` if it hasn't been cancelled?

alfogrillo:
> I'll review this properly in the morning, but for now having looked at the second implementation I'm wondering - would it be cleaner to add a `.startedLoaded` state (or similar) that makes the intent clear in the model, and then instead of setting `.loading` directly in `AudioPlayerState`, create a Task with a sleep that starts with `startedLoading` and transitions to `loading` if it hasn't been cancelled?

> would it be cleaner to add a `.startedLoaded` state (or similar) that makes the intent clear in the model, and then instead of setting `.loading` directly in `AudioPlayerState`, create a Task with a sleep that starts with `startedLoading` and transitions to `loading` if it hasn't been cancelled?

Do you mean a new case in `AudioPlayerPlaybackState`? And the Task in `AudioPlayerPlaybackState`?

My reasoning is that `AudioPlayerState` should just publish the state updates as usual since this is a UX requirement that shouldn't affect it.
This more evident in the [second proposal](https://github.com/vector-im/element-x-ios/compare/alfogrillo/deferred-loader?expand=1) where I don't touch it at all.
In this PR, I touch it adding a new property (`delayedLoaderPlaybackState`) preserving the behavior of the current API.
Does it make sense?

pixlwave (approved):
LGTM, we discussed naming of `delayedLoaderPlaybackState` in a DM.

PR #2020, opened 2023-11-03
Use the custom log level inside the NSE too.

pixlwave:
We weren't reading the log level available in Developer Options in the NSE, and instead had it hardcoded to `.info`. This PR fixes that.


alfogrillo (approved):
LGTM

PR #2021, opened 2023-11-03
Voice over for the bubble style 

alfogrillo:
Element X now has interactive events:
- Polls 
- Voice messages

combining those bubbles as single accessibility element makes this interaction not accessible.

This PR is a proposal for addressing this problem.

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
+++ b/ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
@@ -87,6 +87,7 @@ struct TimelineItemBubbledStylerView<Content: View>: View {
                     .lineLimit(1)
                     .padding(.vertical, senderNameVerticalPadding)
             }
+            // sender info are read inside the `TimelineAccessibilityModifier`
             .accessibilityHidden(true)
             .onTapGesture {
                 context.send(viewAction: .tappedOnUser(userID: timelineItem.sender.id))
@@ -99,14 +100,6 @@ struct TimelineItemBubbledStylerView<Content: View>: View {
         // Figma overlaps reactions by 3
         VStack(alignment: alignment, spacing: -3) {
             messageBubble
-                .accessibilityRepresentation {
-                    VStack(spacing: 0) {
-                        Text(timelineItem.sender.displayName ?? timelineItem.sender.id)
-                            // Workaround: fixes VoiceOver frames offset when children accessibility element aren't combined
-                            .frame(height: 0)
-                        messageBubble
-                    }
-                }
                 .timelineAccessibility(timelineItem)
             
             if !timelineItem.properties.reactions.isEmpty {
--- a/ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
+++ b/ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
@@ -105,7 +105,6 @@ struct TimelineItemBubbledStylerView<Content: View>: View {
                         messageBubble
                     }
                 }
-                .accessibilityElement(children: .combine)
             
             if !timelineItem.properties.reactions.isEmpty {
                 TimelineReactionsView(itemID: timelineItem.id,
--- a/ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
+++ b/ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
@@ -100,11 +100,14 @@ struct TimelineItemBubbledStylerView<Content: View>: View {
         VStack(alignment: alignment, spacing: -3) {
             messageBubble
                 .accessibilityRepresentation {
-                    VStack {
+                    VStack(spacing: 0) {
                         Text(timelineItem.sender.displayName ?? timelineItem.sender.id)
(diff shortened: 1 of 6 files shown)


ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
```diff
@@ -105,7 +105,6 @@ struct TimelineItemBubbledStylerView<Content: View>: View {
                         messageBubble
```
alfogrillo, line 105: Since we now don't combine the children to make a single accessibility element, `.accessibilityRepresentation` is problematic since VoiceOver frames are offset (see the photo below).
One idea to communicate the sender info can be to leverage `.accessibilityHint`.
What do people think?

example

Velin92 (approved):
I think the accessibility hint idea is fine as of right now, but maybe is better to double check with design, to see how actually we want to read the sender info and in what order. As we talked about privately, we don't really have clear voice over guidelines as of right now

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
```diff
@@ -105,7 +105,6 @@ struct TimelineItemBubbledStylerView<Content: View>: View {
...
                     }
                 }
-                .accessibilityElement(children: .combine)
```
pixlwave, line 108: Maybe we could do this conditionally depending on the message type?

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
pixlwave: Accessibility hint would read it in the wrong order I think. We want the sender, and then the message.

alfogrillo: > Accessibility hint would read it in the wrong order I think. We want the sender, and then the message.

Yeah, in this case we would have : "Message", pause, "sent by sender name"

alfogrillo: > I have no idea if this helps, but what if you put a `.frame(height: 0)` on the Text, would that fix the offset?

it does, but we still have frames for the `Text` inside `.accessibilityRepresentation` 

IMG_1409

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
alfogrillo: Don't know, is it helpful to combine accessibility element for non interactive bubbles?
We don't do it apparently for the modern style.

pixlwave: I would say so. For non-interactive messages I would expect that swiping to scroll through the timeline would move one message at a time. Otherwise thinks like the timestamp, Edited, etc will be extra items to have to swipe through. The inconsistencies between bubbles and plain timelines purely comes from us not regularly testing the plain style due to it being an after thought.

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
pixlwave: > Yeah, in this case we would have : "Message", pause, "sent by sender name"

We went with the VStack because the ordering matches what Apple does in iMessage. I can see why it makes sense. If there's a long message in a group chat, it's going to be a lot easier to follow the meaning if you know who sent it before you hear the message.

alfogrillo: I mean: I agree it's better to have the sender read before the message.
But It's not good that the VoiceOver shows this ghost frames.
But maybe we can accept them for now and think better about accessibility later on?

pixlwave: What happens if you pass `EmptyView()` in the accessibility representation? If we conditionally combine the accessibility elements for non-interactive messages, maybe we could also pass an empty view in for the interactive messages if that makes it use the underlying view?

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
alfogrillo: Added a modifier [here](https://github.com/vector-im/element-x-ios/pull/2021/commits/35861317eecff8e532a602011995f6f745311483) which should preserve the current behavior for other message types.

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
alfogrillo: > What happens if you pass EmptyView() in the accessibility representation?

It doesn't read anything apparently.
Maybe I can move `.accessibilityRepresentation` into `TimelineAccessibilityModifier` and apply it just for non interactive items (all but polls, voice messages)?
The flip side is that polls and voice messages wouldn't have speech for the sender.
But I feel it is good enough for now.

Velin92 (approved):

PR #2022, opened 2023-11-03
Fix NSE Leak

### LINKED ISSUE #1923: Missing notifications because of a memory leak in NSE
### Steps to reproduce

- Do not open the app
- Received10 messages that should trigger a notification for each
- Count notifications you receive

#### What did you expect?
I should get 10 notifications

#### What happened instead?
Sometimes I get only 2

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

_No response_

### Homeserver

_No response_

### Will you send logs?

Yes

Velin92:
WIP: Requires updated SDK

fixes #1923

### DIFF
--- a/ElementX/Sources/Services/Media/Provider/MediaFileHandleProxy.swift
+++ b/ElementX/Sources/Services/Media/Provider/MediaFileHandleProxy.swift
@@ -59,8 +59,12 @@ extension MediaFileHandleProxy: Hashable {
 /// An unmanaged file handle that can be created direct from a URL.
 ///
 /// This type allows for mocking but doesn't provide the automatic clean-up mechanism provided by the SDK.
-private struct UnmanagedMediaFileHandle: MediaFileHandleProtocol {
+private class UnmanagedMediaFileHandle: MediaFileHandleProtocol {
     let url: URL
+    
+    init(url: URL) {
+        self.url = url
+    }
     
     func path() -> String {
         url.path()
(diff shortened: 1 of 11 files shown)


NSE/Sources/NotificationServiceExtension.swift
```diff
@@ -18,15 +18,38 @@ import Intents
...
+//      notifications it needs to
+//  4)  Extension notifies its work is complete by calling
+//      the contentHandler
+//  5)  If the extension takes too long to perform its work
+//      (more than 30s), it will be notified and immediately
+//      terminated
+//
+// Note that the NSE does *not* always spawn a new process to
+// handle a new notification and will also try and process notifications
+// in parallel. `didReceive` could be called twice for the same process,
```
pixlwave, line 35: Are you sure about in parallel? I thought it always queued the next one up until the contentHandler is called 🤔

Otherwise you could receive notifications out of order, no?

ElementX/Sources/Services/Media/Provider/MediaFileHandleProxy.swift
pixlwave, line 1: What's are these two changes about??

pixlwave (approved):
Nice 👌

NSE/Sources/NotificationServiceExtension.swift
Velin92: I think the parallel is referring to the parallel threads that can handle the didReceive function, but is still possible that the different threads still serve only one notification per run during the lifecycle of the didReceive function making the other threads wait the first one that has come. But not 100% sure

ElementX/Sources/Services/Media/Provider/MediaFileHandleProxy.swift
Velin92: the protocol now requires to be using a AnyObject.

PR #2034, opened 2023-11-06
Don't allow the Placeholder Screen layout to be influenced by the keyboard

### LINKED ISSUE #2032: Lockscreen has the wrong layout if you lock the app with the keybord open.
### Steps to reproduce

1. Enable PIN
2. Open a timeline (thus displaying the keyboard)
3. Lock the app by bging it
4. Fg it
5. Observe that the animation of app launch blends between a correct(?) lock screen graphic (with the centered Element X logo) and an incorrect(?) one (with the logo offset by the height of the keyboard). As a result you see doubled up logos during the launch animation, which looks really unpolished and wrecks what would otherwise be a really nice visual polish.

IMG_3913

6.  Compare it when you do the same operation with the keyboard shut (eg by bging in the unfiltered roomlist), and see how nice it looks without the bug (ignoring the FaceID overlay):

IMG_3915

#### What did you expect?

pretty lock screen

#### What happened instead?

IMG_3916

### Your phone model

iphone 12 pro max

### Operating system version

16.4

### Application version

428

### Homeserver

_No response_

### Will you send logs?

No

pixlwave:
Whilst I was unable to reproduce the issue reported, this fixes #2032.

| Forced layout to visualise the fix |
| - |
| Simulator Screenshot - iPhone 14 - 2023-11-06 at 10 37 28 |


alfogrillo (approved):
LGTM!

PR #2037, opened 2023-11-06
Secure backup fixes

stefanceriu:
* fixed wrong logout alert title when secure backup is enabled
* disable corrections on the recovery key text field
* allow the recovery key confirmation screen to scroll on small screens


ElementX/Sources/Screens/SecureBackup/SecureBackupRecoveryKeyScreen/View/SecureBackupRecoveryKeyScreen.swift
```diff
@@ -20,15 +20,31 @@ import SwiftUI
...
+                    .padding()
+                    .onChange(of: focused) { newValue in
+                        guard newValue == true else { return }
+                        reader.scrollTo(textFieldIdentifier)
+                    }
+            }
+        }
+        .safeAreaInset(edge: .bottom) {
+            footer
+                .padding([.horizontal, .bottom])
```
pixlwave, line 39: FWIW, should probably be explicit and put 16 in here

ElementX/Sources/Screens/SecureBackup/SecureBackupRecoveryKeyScreen/View/SecureBackupRecoveryKeyScreen.swift
```diff
@@ -20,15 +20,31 @@ import SwiftUI
...
-            .padding()
-            .interactiveDismissDisabled()
-            .toolbar { toolbar }
-            .toolbar(.visible, for: .navigationBar)
-            .background(Color.compound.bgCanvasDefault.ignoresSafeArea())
-            .alert(item: $context.alertInfo)
+        ScrollView {
+            ScrollViewReader { reader in
+                mainContent
+                    .padding()
```
pixlwave, line 30: And here.

pixlwave (approved):
LGTM. FWIW, maybe using `FullscreenDialog` would be sensible here because it adds affordances for AX1 and larger font sizes.

stefanceriu:
> wondering if using FullscreenDialog would be sensible here

I probably would but that's not what design expects so I'm not going to bother with it too much.

PR #2039, opened 2023-11-07
Expose Element Call settings in the developer options. Start using th…

stefanceriu:
…e encryption widget parameter.

### DIFF
--- a/ElementX/Sources/Application/AppSettings.swift
+++ b/ElementX/Sources/Application/AppSettings.swift
@@ -37,6 +37,9 @@ final class AppSettings {
         case viewSourceEnabled
         case richTextEditorEnabled
         
+        case elementCallBaseURL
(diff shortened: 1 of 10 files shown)


ElementX/Sources/Application/AppSettings.swift
```diff
@@ -215,7 +218,11 @@ final class AppSettings {
     
     // MARK: - Element Call
     
-    let elementCallBaseURL: URL = "https://call.element.dev"
+    @UserPreference(key: UserDefaultsKeys.elementCallBaseURL, defaultValue: "https://call.element.io", storageType: .userDefaults(store))
```
stefanceriu, line 221: This was an intentional change, with the setting available in developer options we no longer need to default to .dev

nimau (approved):
SGTM

nimau (approved):

PR #2041, opened 2023-11-07
Fixes #2040 - Prevent the room list from staying empty after cancelli…

### LINKED ISSUE #2040: Room list empty after cancelling search
Original report here https://github.com/vector-im/element-x-ios-rageshakes/issues/1018

```
Empty room list after a room search

To reproduce

start a room search
type a letter
cancel the search
The room list is empty NOK
```

stefanceriu:
…ng a search

### DIFF
--- a/ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
+++ b/ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
@@ -98,19 +98,20 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol 
             .weakAssign(to: \.state.selectedRoomID, on: self)
             .store(in: &cancellables)
         
-        struct SearchBarState: Equatable {
-            let searchQuery: String
-            let isSearchFieldFocused: Bool
-        }
-        
-        context.$viewState
-            .map { SearchBarState(searchQuery: $0.bindings.searchQuery, isSearchFieldFocused: $0.bindings.isSearchFieldFocused) }
-            .removeDuplicates()
-            .sink { [weak self] _ in
+        let isSearchFieldFocused = context.$viewState.map(\.bindings.isSearchFieldFocused)
+        let searchQuery = context.$viewState.map(\.bindings.searchQuery)
+        isSearchFieldFocused
+            .combineLatest(searchQuery)
+            .removeDuplicates { $0 == $1 }
+            .map { _ in () }
+            .sink { [weak self] in
                 guard let self else { return }
                 // Don't capture the values here as combine behaves incorrectly and `isSearchFieldFocused` is sometimes
-                // turning to true after cancelling the search. Read them directly from the state in the updateFilter method instead.
-                updateFilter()
+                // turning to true after cancelling the search. Read them directly from the state in the updateFilter
+                // method instead on the next run loop to make sure they're up to date.
+                DispatchQueue.main.async {
+                    self.updateFilter()
+                }
             }
             .store(in: &cancellables)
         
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -98,21 +98,19 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
...
-            .map(\.bindings.isSearchFieldFocused)
+            .map { SearchBarState(searchQuery: $0.bindings.searchQuery, isSearchFieldFocused: $0.bindings.isSearchFieldFocused) }
             .removeDuplicates()
-            .sink { [weak self] isSearchFieldFocused in
+            .sink { [weak self] _ in
                 guard let self else { return }
-                updateFilter(isSearchFieldFocused: isSearchFieldFocused, searchQuery: state.bindings.searchQuery)
+                // Don't capture the values here as combine behaves incorrectly and `isSearchFieldFocused` is sometimes
+                // turning to true after cancelling the search. Read them directly from the state in the updateFilter method instead.
+                updateFilter()
```
stefanceriu, line 113: fwiw doug and I tried various combine solutions but it still wasn't behaving correctly, no idea what the underlying problem actually is. Sometimes `isSearchFieldFocused` just turns to true without anybody changing it.

alfogrillo: I think the problem in the previous implementation was that the two combine chains were reading from `state.bindings` one value and receiving the other one from the publisher.
Did you try to read both from the publishers?
One idea can be this:
```swift
        let isSearchFieldFocused = context.$viewState.map(\.bindings.isSearchFieldFocused).removeDuplicates()
        let searchQuery = context.$viewState.map(\.bindings.searchQuery).removeDuplicates()
        isSearchFieldFocused
            .combineLatest(searchQuery)
            .sink { [weak self] isSearchFieldFocused, searchQuery in
                updateFilter(isSearchFieldFocused: isSearchFieldFocused, searchQuery: searchQuery)
            }
            .store(in: &cancellables)
```
If it works we could delete `struct SearchBarState: Equatable`

stefanceriu: That what we figured as well but we did actually try this ^ and it was still wrong. Even the plain .map which gets the valure directly from bindings at every change was wrong at times. The only thing that did work was to fetch it from the state outside of the sink.

alfogrillo: My memories on [objectwillchange](https://developer.apple.com/documentation/combine/observableobject/objectwillchange-2oa5v) are that published  values are always updated, but the corresponding values in the state may still be "old" if devs try to read them directly.
This is why I thought that reading both from a Combine chain is better.
So given the current implementation it's like inside `updateFilter()` we read the old values 🤔
Tbh I'm not sure why this works.
wdyt?

stefanceriu: Yeah, I know exactly what you mean, I have absolutely no idea either but empirically it does 😅

alfogrillo:
LGTM!

alfogrillo (approved):

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
stefanceriu: Talked about it offline, pushed another that uses the same approach but with an extra async dispatch and simpler implementation.

PR #2042, opened 2023-11-07
Voice message recording enhancements

nimau:
This PR adds some enhancements to voice message recording:
- The `AudioRecorder` is now using `AVAudioEngine` instead of `AVAudioRecorder`, this allows to have a more accurate recording duration, to detect configuration change (switching to and from AirPods) and to record a voice message using AirPods.
- For the live waveform, the average power is now calculated using the samples provided by the `AudioEngine`.
- The `AudioRecorder` class now reports its state using `Combine` actions instead of returning a result. This makes the `VoiceMessageRecorder` class more testable and robust.

### DIFF
--- a/ElementX/Sources/Services/Audio/Recorder/AudioRecorder.swift
+++ b/ElementX/Sources/Services/Audio/Recorder/AudioRecorder.swift
@@ -14,6 +14,7 @@
 // limitations under the License.
 //
 
+import Accelerate
 import AVFoundation
 import Combine
 import Foundation
@@ -373,17 +374,10 @@ class AudioRecorder: AudioRecorderProtocol {
             return
         }
         
-        // Convert to an array of Float
-        let channelDataValue = channelData.pointee
-        let channelDataValueArray = stride(from: 0,
-                                           to: Int(buffer.frameLength),
-                                           by: buffer.stride)
-            .map { channelDataValue[$0] }
-        
         // Compute RMS
-        let rms = sqrt(channelDataValueArray.map { $0 * $0 }
-            .reduce(0, +) / Float(buffer.frameLength))
-
+        var rms: Float = .nan
+        vDSP_rmsqv(channelData.pointee, buffer.stride, &rms, vDSP_Length(buffer.frameLength))
+        
         // Convert to decibels
         let avgPower = 20 * log10(rms)
         
--- a/ElementX/Sources/Services/Audio/Recorder/AudioRecorder.swift
+++ b/ElementX/Sources/Services/Audio/Recorder/AudioRecorder.swift
@@ -19,11 +19,18 @@ import Combine
 import Foundation
 import UIKit
 
+private enum InternalAudioRecorderState: Equatable {
+    case recording
+    case suspended
+    case stopped
+    case error(AudioRecorderError)
+}
+
 class AudioRecorder: AudioRecorderProtocol {
     private var audioEngine: AVAudioEngine!
     private var audioFile: AVAudioFile?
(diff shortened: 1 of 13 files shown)


Velin92 (approved):

Velin92 (approved):
LGTM!

ElementX/Sources/Services/Audio/Recorder/AudioRecorder.swift
```diff
@@ -183,29 +225,168 @@ class AudioRecorder: NSObject, AudioRecorderProtocol, AVAudioRecorderDelegate {
...
+        // Convert to an array of Float
+        let channelDataValue = channelData.pointee
+        let channelDataValueArray = stride(from: 0,
+                                           to: Int(buffer.frameLength),
+                                           by: buffer.stride)
+            .map { channelDataValue[$0] }
+        
+        // Compute RMS
+        let rms = sqrt(channelDataValueArray.map { $0 * $0 }
+            .reduce(0, +) / Float(buffer.frameLength))
```
pixlwave, line 385: Should we maybe start using Accelerate now, given this is going to be called a *lot*?

ElementX/Sources/Services/Audio/Recorder/AudioRecorder.swift
```diff
@@ -128,37 +131,64 @@ class AudioRecorder: NSObject, AudioRecorderProtocol, AVAudioRecorderDelegate {
...
-                            AVNumberOfChannelsKey: 1,
-                            AVEncoderAudioQualityKey: AVAudioQuality.high.rawValue]
             
+            setupAudioSession()
+            let audioEngine = AVAudioEngine()
+            self.audioEngine = audioEngine
+
+            // The sample rate must match the hardware sample rate for the audio engine to work.
+            let sampleRate = audioEngine.inputNode.inputFormat(forBus: 0).sampleRate
+            let recordingFormat = AVAudioFormat(commonFormat: .pcmFormatFloat32,
```
pixlwave, line 157: Just a thought, did you try `AVAudioFormat(settings:)`? We should be able to set `kAudioFormatMPEG4AAC` for `AVFormatIDKey` to record to that directly. If that works, seeing as we've switched to AVAudioEngine, if its within scope, I seriously think we should consider recording directly to `kAudioFormatOpus` and sending the data to an OGG container ourselves rather than doing the transcoding dance.

pixlwave (commented):
A couple of comments because seeing the use of `AVAudioEngine` piqued my curiosity.

ElementX/Sources/Services/Audio/Recorder/AudioRecorder.swift
nimau: Oh nice 😀 I've updated the code.

ElementX/Sources/Services/Audio/Recorder/AudioRecorder.swift
nimau: The `AVAudioFile` is created with `kAudioFormatMPEG4AAC`, so when a buffer is written, the AVAudioFile takes care of compressing the file. This way, the sample given by the audio engine is in .pcm, easy to process and it's written in a compressed format.
For the `kAudioFormatOpus`, the DSWaveformImage package doesn't seem to be happy with an OGG file, so it will take a little more work to record directly to opus.

pixlwave: That'll teach me for skimming a PR quickly whilst I'm switching off for the day 😅.

PR #2045, opened 2023-11-08
Extract room screen action handling into separate component

stefanceriu:
This is hopefuly the first step of many that will simplify the room screen VM, clarify the differences between it and the room timeline controller and make our life easier


ElementX/Sources/Screens/RoomScreen/RoomScreenActionsHandler.swift
```diff
@@ -0,0 +1,479 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Combine
+import UIKit
+
+enum RoomScreenActionsHandlerAction {
```
pixlwave, line 20: I think the repetition of Action here is enough for me to suggest this file should be something more like `RoomScreenInteractionHandler` or similar?

pixlwave (approved):

ElementX/Sources/Screens/RoomScreen/RoomScreenActionsHandler.swift
stefanceriu: Yeah, I like that 👍

nimau (approved):
LGTM 👍

alfogrillo (approved):
LGTM

pixlwave:
Oh I forgot to mention, should we also revert `swiftlint.yml` back to erroring at 1000 lines again?

stefanceriu:
> should we also revert `swiftlint.yml` back to erroring at 1000 lines again?

yep yep, all done 👍

PR #2046, opened 2023-11-08
Add UI tests for optional app lock setup.

pixlwave:
This PR runs through the create and unlock flows for the app lock feature when presented from settings.

Will follow this PR up with another one for the mandatory flow shown during auth.


ElementX/Sources/Other/AccessibilityIdentifiers.swift
```diff
@@ -41,6 +45,25 @@ struct A11yIdentifiers {
     static let notificationSettingsScreen = NotificationSettingsScreen()
     static let notificationSettingsEditScreen = NotificationSettingsEditScreen()
     static let createPollScreen = CreatePollScreen()
+    
+    struct AlertInfo {
+        let primaryButton = "alert_info-primary_button"
+        let secondaryButton = "alert_info-secondary_button"
+    }
```
pixlwave, line 52: I kind of hoped there was a way to detect this from `app.alerts[0]` but I just got a generic array of buttons, so hard-coded these identifiers to fix it.

ElementX/Sources/UITests/BlanckFormCoordinator.swift
pixlwave, line 1: Weird diff, but ok 🙈

ElementX/Sources/UITests/UITestsAppCoordinator.swift
```diff
@@ -42,12 +42,12 @@ class UITestsAppCoordinator: AppCoordinatorProtocol {
     }
     
     func start() {
-        // disabling CA animations
+        // Fix the app tint colour.
         UIApplication.shared.connectedScenes.forEach { scene in
             guard let delegate = scene.delegate as? UIWindowSceneDelegate else {
                 return
             }
-            delegate.window??.layer.speed = 0
```
pixlwave, line 50: This caused a very specific problem: Presenting a sheet, and then focussing a text field on that sheet, would completely disable all interaction on that sheet. I have no idea why, but running the UI tests on the PR to check if it is needed now that we use `Transaction`.

pixlwave: The run is here: https://github.com/vector-im/element-x-ios/actions/runs/6798348298

UITests/Sources/AppLockSetupUITests.swift
```diff
@@ -0,0 +1,130 @@
...
+
+import XCTest
+
+@testable import ElementX
+
+@MainActor
+class AppLockSetupUITests: XCTestCase {
+    var app: XCUIApplication!
+    
+    @MainActor enum Step {
```
alfogrillo, line 25: Is `@MainActor` really needed here for the `Step`?

alfogrillo (approved):
LGTM 👏🏻

UITests/Sources/AppLockSetupUITests.swift
pixlwave: Yeah its stupid: [`UIDevice`](https://developer.apple.com/documentation/uikit/uidevice) is MainActor so to keep the `isPhone` property synchronous (and therefore the 2 change pin properties), this was the simplest way to do it.

alfogrillo: Maybe you could have moved the annotation just to that property (`isPhone`) to make this clear, but it doesn't really make any difference I guess.

ElementX/Sources/UITests/UITestsAppCoordinator.swift
pixlwave: Seems to have worked.

PR #2049, opened 2023-11-08
Add UI tests for mandatory PIN setup flow.

pixlwave:
Adds the tests for the flow coordinator configuration where the presenting flow is `.onboarding`.


UITests/Sources/__Snapshots__/Application/pseudo-iPad-9th-generation.appLockSetupFlowMandatory-0.png
Velin92, line 1: is it normal that the iPad displays a numeric keyboard with also symbols?

pixlwave: Yeah sadly when you ask for a number pad on iPad this is what you get. The text field already [filters out](https://github.com/vector-im/element-x-ios/blob/abe2080ee2faddff738093e71ec87706ea367580/ElementX/Sources/Screens/AppLock/AppLockSetupPINScreen/View/PINTextField.swift#L50) anything that isn't a digit as it also supports input from a real keyboard (in the setup flow, that isn't implemented in the unlock flow) so it isn't *that* bad.

PR #2050, opened 2023-11-08
Custom copy for notifications that contain mentions

### LINKED ISSUE #1825: Mentions: Push notifications display user mentions differently
As of right now they are displayed as links, but they probably should be converted into text

Mentions should be displayed with @username in the notification (however since the username is usually not available in the notification context, is also okay to prefix with @ the hyperlink text just to make it clear that is a mention from the notification)

Also if you have been mentioned in the notification we should prefix the notification body with:
"Username mentioned you"
or
"You have been mentioned" if the username of the sender is not available

Velin92:
fixes #1825 

ima_0ed5ddd


ElementX/Sources/Services/Media/Provider/MediaFileHandleProxy.swift
```diff
@@ -60,6 +66,10 @@ extension MediaFileHandleProxy: Hashable {
 ///
 /// This type allows for mocking but doesn't provide the automatic clean-up mechanism provided by the SDK.
 private class UnmanagedMediaFileHandle: MediaFileHandleProtocol {
+    func persist(path: String) throws -> Bool {
+        false
+    }
+
```
Velin92, line 72: I have no idea what to do with this

ElementX/Sources/Services/Media/Provider/MediaLoader.swift
```diff
@@ -48,7 +48,7 @@ actor MediaLoader: MediaLoaderProtocol {
     
     func loadMediaFileForSource(_ source: MediaSourceProxy, body: String?) async throws -> MediaFileHandleProxy {
         let result = try await Task.dispatch(on: clientQueue) {
-            try self.client.getMediaFile(mediaSource: source.underlyingSource, body: body, mimeType: source.mimeType ?? "application/octet-stream", tempDir: nil)
+            try self.client.getMediaFile(mediaSource: source.underlyingSource, body: body, mimeType: source.mimeType ?? "application/octet-stream", useCache: false, tempDir: nil)
```
Velin92, line 51: same here

ElementX/Sources/Services/SecureBackup/SecureBackupController.swift
```diff
@@ -195,8 +195,6 @@ extension BackupState {
             return .enabled
         case .disabling:
             return .disabling
-        case .disabled:
-            return .disabled
```
Velin92, line 199: seems that we don't have a disabled case anymore here

NSE/Sources/Other/NSESettings.swift
```diff
@@ -17,6 +17,8 @@
 import Foundation
 
 final class NSESettings {
+    let permalinkBaseURL: URL = "https://matrix.to"
```
Velin92, line 20: I would like this value to be shared among NSESettings and the normal AppSettings... I was thinking of maybe creating a common protocol for it and expose it a common extension for both of them?

ElementX/Sources/Services/SecureBackup/SecureBackupController.swift
Velin92: @pixlwave

ElementX/Sources/Services/Media/Provider/MediaLoader.swift
Velin92: @stefanceriu

ElementX/Sources/Services/Media/Provider/MediaFileHandleProxy.swift
Velin92: @stefanceriu

ElementX/Sources/Application/CommonConstantSettings.swift
pixlwave, line 1: This is all getting too over-engineered. Discussed it with @stefanceriu  and the whole point of `AppSettings` is to have a single file where you can find all of the build settings and default values for user settings. Please can we just use that directly in the NSE so we keep that behaviour.

ElementX/Sources/Services/Media/Provider/MediaLoader.swift
nimau: I think we should use `useCache: true` because it seems it's was true by default on [Rust side](https://github.com/matrix-org/matrix-rust-sdk/commit/57136247bf018e1c00b00ad54bb84e306a16b189)

ElementX/Sources/Services/Media/Provider/MediaFileHandleProxy.swift
```diff
@@ -37,8 +37,14 @@ class MediaFileHandleProxy {
     }
     
     /// The media file's location on disk.
-    var url: URL {
-        URL(filePath: handle.path())
+    var url: URL? {
```
pixlwave, line 40: Given we don't expose or call persist, this could be `URL!`.

ElementX/Sources/Other/Extensions/UNNotificationContent.swift
```diff
@@ -86,8 +86,12 @@ extension UNMutableNotificationContent {
         switch await mediaProvider.loadFileFromSource(mediaSource) {
         case .success(let file):
             do {
+                guard let url = file.url else {
+                    MXLog.error("Couldn't add media attachment: URL is nil")
+                    return self
+                }
                 let identifier = ProcessInfo.processInfo.globallyUniqueString
-                let newURL = try FileManager.default.copyFileToTemporaryDirectory(file: file.url, with: "\(identifier).\(file.url.pathExtension)")
+                let newURL = try FileManager.default.copyFileToTemporaryDirectory(file: url, with: "\(identifier).\(url.pathExtension)")
```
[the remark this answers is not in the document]
Velin92: but the url is value type, I should have copied it.

ElementX/Sources/Services/SecureBackup/SecureBackupController.swift
stefanceriu: Ah right, disabled has been merged into unknown so let's just do `case .unknown: return .disabled`.
I'll come back to all of this later, when damir is closer to finishing

ElementX/Sources/Other/Extensions/UNNotificationContent.swift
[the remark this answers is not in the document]
Velin92: also swift will keep alive the object at least until the end of the scope of the function, so I don't see any issue here

ElementX/Sources/Application/CommonConstantSettings.swift
Velin92: the problem is that we will need to replicate this value twice which doesn't make a lot of sense and can create issues if both are not updated a the same time, and importing the whole AppSettings would add a lot of useless imports.

Another route to make things easier and to avoid the use of a protocol and use instead a shared enum and set the values for permalinkBaseURL to be equal to the static values of the enum:

```
enum CommonSettings {
static let permalinkBaseURL: URL = whatever
}

NSESettings: {
let permalinkBaseURL = CommonSettings.permalinkBaseURL
}

AppSettings: {
let permalinkBaseURL = CommonSettings.permalinkBaseURL
}
```

pixlwave (approved):
LGTM 😎

PR #2052, opened 2023-11-09
Have the UserIndicatorController use a separate window, switch to usi…

stefanceriu:
…ng one single instance throughout the application


nimau (approved):
LGTM

PR #2055, opened 2023-11-09
Add UI tests for AppLockFlowCoordinator.

pixlwave:
The PR adds UI tests for the unlock flow of the App Lock feature along with the following changes:
- Add a `UITestsNotificationCenter` for sending notifications from the test runner into the app.
- Simplify the UITests signalling with Codable.
- Use the window manager in all 3 app coordinators.
- Pass the window manager into the mock screen for controlling as needed.
- Remove all `import ElementX` statements from UITests as they're unused, and also not needed (we add any files directly to the target to avoid symbol issues).


ElementX/Sources/Other/AccessibilityIdentifiers.swift
```diff
@@ -51,6 +52,10 @@ enum A11yIdentifiers {
         let secondaryButton = "alert_info-secondary_button"
     }
     
+    struct AppLockScreen {
+        func numpad(_ digit: Int) -> String { "app_lock-numpad_\(digit)" }
+    }
```
pixlwave, line 57: I think this is fine?

ElementX/Sources/UITests/UITestsAppCoordinator.swift
```diff
@@ -42,14 +48,6 @@ class UITestsAppCoordinator: AppCoordinatorProtocol {
     }
     
     func start() {
-        // Fix the app tint colour.
-        UIApplication.shared.connectedScenes.forEach { scene in
-            guard let delegate = scene.delegate as? UIWindowSceneDelegate else {
-                return
-            }
-            delegate.window??.tintColor = .compound.textActionPrimary
-        }
```
pixlwave, line 51: This is all done by the window manager now.

nimau (approved):
LGTM 👍

PR #2057, opened 2023-11-10
Better compiler flag for AppSettings

Velin92:
Since we are now sharing the AppSettings with NSE we need to avoid compiling stuff that is not available in the NSE context.
This fixes an issue that made the NSE crash when trying to access the log level.


pixlwave (approved):
We add the `baseBundleIdentifier` and the `appGroupIdentifier` to each target, so these could happily remain on `main` if you wanted, but yep good spot 👍

pixlwave (approved):

PR #2061, opened 2023-11-10
Fixes #660 - Expose message actions to voice over rotor

### LINKED ISSUE #660: VoiceOver rotor on messages

stefanceriu:
Expose a timeline item action on the voice over rotor that opens up the timeline item actions menu.

Ideally we would've exposed the actions directly but the available APIs don't allow computing the a11y actions on the fly and computing them for all messages is intractable

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemAccessibilityModifier.swift
+++ b/ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemAccessibilityModifier.swift
@@ -25,6 +25,11 @@ private struct TimelineItemAccessibilityModifier: ViewModifier {
         switch timelineItem {
         case is PollRoomTimelineItem, is VoiceMessageRoomTimelineItem:
             content
+                .accessibilityActions {
+                    Button(L10n.commonMessageActions) {
+                        action()
+                    }
+                }
         case let timelineItem as EventBasedTimelineItemProtocol:
             content
                 .accessibilityRepresentation {
--- a/ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemAccessibilityModifier.swift
+++ b/ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemAccessibilityModifier.swift
@@ -16,8 +16,9 @@
 
 import SwiftUI
 
-private struct TimelineAccessibilityModifier: ViewModifier {
+private struct TimelineItemAccessibilityModifier: ViewModifier {
     let timelineItem: RoomTimelineItemProtocol
+    let action: () -> Void
     
     @ViewBuilder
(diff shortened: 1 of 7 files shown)


ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemAccessibilityModifier.swift
```diff
@@ -33,6 +34,11 @@ private struct TimelineAccessibilityModifier: ViewModifier {
                     }
                 }
                 .accessibilityElement(children: .combine)
+                .accessibilityActions {
+                    Button(L10n.commonMessageActions) {
+                        action()
+                    }
```
alfogrillo, line 40: Maybe we should also have it on the previous case (`case is PollRoomTimelineItem, is VoiceMessageRoomTimelineItem`)?

alfogrillo (commented):
Nice idea!

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemAccessibilityModifier.swift
stefanceriu: Oh oupsie, of course 👍

alfogrillo (approved):
LGTM!

PR #2063, opened 2023-11-10
Fix a regression where the forced logout indicator was in the hidden overlay window.

pixlwave:
Given we don't use indicators for anything else on the lock screen, we decided to overlay it directly on top of the screen.

### DIFF
--- a/ElementX/Sources/FlowCoordinators/AppLockFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/AppLockFlowCoordinator.swift
@@ -49,12 +49,6 @@ class AppLockFlowCoordinator: CoordinatorProtocol {
         // Set the initial background state.
         showPlaceholder()
         
-        appLockService.disabledPublisher
-            .sink {
-                // When the service is disabled via a force logout, we need to remove the activity indicator.
-            }
-            .store(in: &cancellables)
-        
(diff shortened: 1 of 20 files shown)


ElementX/Sources/FlowCoordinators/AppLockFlowCoordinator.swift
```diff
@@ -52,7 +52,6 @@ class AppLockFlowCoordinator: CoordinatorProtocol {
         appLockService.disabledPublisher
             .sink {
                 // When the service is disabled via a force logout, we need to remove the activity indicator.
-                ServiceLocator.shared.userIndicatorController.retractAllIndicators()
```
pixlwave, line 55: Ooops, there's more to remove here.

Velin92 (approved):

ElementX/Sources/Screens/AppLock/AppLockScreen/View/AppLockScreen.swift
```diff
@@ -58,6 +58,11 @@ struct AppLockScreen: View {
         }
         .background()
         .environment(\.backgroundStyle, AnyShapeStyle(Color.compound.bgCanvasDefault))
+        .disabled(context.viewState.forcedLogoutIndicator != nil)
```
pixlwave, line 61: The overlay blocks the screen, but just in case :)

PR #2064, opened 2023-11-10
Don't show the call button on macOS

pixlwave:
Calls don't work in the web view on macOS and currently clicking the button shows you both the error, and for some reason, redirects you to FaceTime 🙃

Note: CI failures handled in #2067.


nimau (approved):
LGTM

stefanceriu (approved):
lgtm too 👍

PR #2065, opened 2023-11-10
Fix for UI tests now showing the status bar.

pixlwave:
Now that we've hooked up the window manager, the overlay window is shown in UI tests. This has resulted in all of the screens being presented with the status bar visible. This PR fixes that.

Note: CI failures handled in #2067.


stefanceriu (approved):
lgtm 👍

PR #2067, opened 2023-11-10
Update the welcome screen with the latest designs.

### LINKED ISSUE #2053: Iterate on onboarding screen
The Onbaording screen should look like the following:

image

https://www.figma.com/file/0MMNu7cTOzLOlWb7ctTkv3/Element-X?type=design&node-id=12221%3A142790&mode=design&t=za7VBS1tD1upttSj-1

Context: https://github.com/vector-im/element-x-android/issues/1776#issuecomment-1803719411

pixlwave:
Fixes both a failing PreviewTest (the string was updated) and fixes #2053.

I haven't switched to the Compound Icons in the design because the lock with a dot doesn't exist (cc @callumu).

| Before | After |
| - | - |
| test_welcomeScreen 1 | test_welcomeScreen-1 1 |

### DIFF
--- a/ElementX/Sources/Screens/AnalyticsPromptScreen/View/AnalyticsPromptScreen.swift
+++ b/ElementX/Sources/Screens/AnalyticsPromptScreen/View/AnalyticsPromptScreen.swift
@@ -76,6 +76,7 @@ struct AnalyticsPromptScreen: View {
         }
         .fixedSize(horizontal: false, vertical: true)
         .frame(maxWidth: .infinity)
+        .environment(\.backgroundStyle, AnyShapeStyle(.compound.bgSubtleSecondary))
     }
 
     @ViewBuilder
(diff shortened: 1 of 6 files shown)


callumu:
> Fixes both a failing PreviewTest (the string was updated) and fixes #2053.
> 
> I haven't switched to the Compound Icons in the design because the lock with a dot doesn't exist (cc @callumu).
> 
> Before	After
> test_welcomeScreen 1	test_welcomeScreen-1 1

Okay thanks. The new outline lock icon should will be added to the Compound repo soon - is this icon reusing the one that's in place for PIN?

alfogrillo (approved):
LGTM!

pixlwave:
> is this icon reusing the one that's in place for PIN?

@callumu No, that one is a solid icon rather than outlined and looks way too prominent against the outlined message bubble. They're both just using SF Symbols like the original screen did.

PR #2068, opened 2023-11-11
Display whether a call is ongoing in any particular room on the room …

stefanceriu:
…list


ElementX/Sources/Screens/RoomScreen/View/Timeline/SeparatorRoomTimelineView.swift
```diff
@@ -24,6 +24,7 @@ struct SeparatorRoomTimelineView: View {
             .font(.compound.bodySMSemibold)
             .foregroundColor(.compound.textPrimary)
             .frame(maxWidth: .infinity)
+            .multilineTextAlignment(.center)
```
pixlwave, line 27: 👍

UnitTests/__Snapshots__/PreviewTests/test_homeScreen.Loaded.png
pixlwave, line 1: I think we should add a 4th room to this preview that has no decoration.

pixlwave (approved):
👍

PR #2069, opened 2023-11-11
Use the unread count to update the application badge on all flow bran…

stefanceriu:
…ches

### DIFF
--- a/NSE/Sources/NotificationServiceExtension.swift
+++ b/NSE/Sources/NotificationServiceExtension.swift
@@ -88,7 +88,7 @@ class NotificationServiceExtension: UNNotificationServiceExtension {
         // Called just before the extension will be terminated by the system.
         // Use this as an opportunity to deliver your "best attempt" at modified content, otherwise the original push payload will be used.
         MXLog.warning("\(tag) serviceExtensionTimeWillExpire")
-        notify()
+        notify(unreadCount: nil)
     }
 
     private func run(with credentials: KeychainCredentials,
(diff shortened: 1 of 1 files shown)


NSE/Sources/NotificationServiceExtension.swift
```diff
@@ -130,34 +130,39 @@ class NotificationServiceExtension: UNNotificationServiceExtension {
...
     }
 
-    private func discard() {
+    private func discard(unreadCount: Int?) {
         MXLog.info("\(tag) discard")
+        
+        let content = UNMutableNotificationContent()
+        
+        if let unreadCount {
+            content.badge = NSNumber(value: unreadCount)
```
Velin92, line 162: not too sure about this, for discarding the UNMutableNotificationContent should be completely empty... if you test this by triggering a notification that is delivered by the server but discarded in the client, with an unreadCount != nil, does it still prevent the notification from being completely displayed in app?
Maybe try to force in the code an unreadCount != nil, and then try to trigger a discard notification like Element Web Call to El-X? (it sends some extra events notifications that are discarded)

Velin92 (approved):

NSE/Sources/NotificationServiceExtension.swift
stefanceriu: I just tested it now by returning a random value from `UNNotificationRequest.unreadCount` and `discarding` every single notification. The badge is update and there's no notification shows, so I think we're good 👍

PR #2071, opened 2023-11-13
Fixes 2070 - Prevent crashes when inserting mention pills in the comp…

stefanceriu:
…osers on Mac

### DIFF
--- a/ElementX/Sources/Other/Pills/PillAttachmentViewProvider.swift
+++ b/ElementX/Sources/Other/Pills/PillAttachmentViewProvider.swift
@@ -74,6 +74,19 @@ final class PillAttachmentViewProvider: NSTextAttachmentViewProvider {
         self.view = controller.view
         delegate?.registerPillView(controller.view)
     }
+    
+    // MARK: - NSSecureCoding
+    
+    // Fixes crashes when inserting mention pills in the composer on Mac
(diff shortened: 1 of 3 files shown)


Velin92 (approved):

ElementX/Sources/Other/Pills/PillAttachmentViewProvider.swift
```diff
@@ -30,8 +30,6 @@ protocol PillAttachmentViewProviderDelegate: AnyObject {
 final class PillAttachmentViewProvider: NSTextAttachmentViewProvider {
     private weak var delegate: PillAttachmentViewProviderDelegate?
     
-    // MARK: - Override
```
Velin92, line 33: I guess we can keep this line?

stefanceriu: If you really have strong feelings about it 😅 I figured all methods in there start with `override`, I might as well remove it

Velin92: well we know have the NSSecureCoding mark, so maybe we can have the override mark too to separate both of them? not really important, just point this out

stefanceriu: No no, that's fair, I'll add it back 👍

PR #2072, opened 2023-11-13
Fixes #1619 - Remove push notification registration on logout

### LINKED ISSUE #1619: Remove notifications registration on auth error.
The callback `didReceiveAuthError` always triggers a sign out. However when this occurs, the device continues to receive notifications (for which content is unavailable as there is no longer an access token).

Not exactly sure how we would do this without a working access token and considering the possibility of multi-account support in the future, but we should unregister somewhere so that this stops happening. Semi-related to #898 I guess.

stefanceriu:
I believe this should be enough. The app already checks for permissions and re-registers every time it becomes active and I believe that should be more than enough for this edge case

### DIFF
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -527,10 +527,12 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
             return
         }
         
-        // The user will log out, clear any existing notifications
+        // The user will log out, clear any existing notifications and unregister from receving new ones
         UNUserNotificationCenter.current().removeAllPendingNotificationRequests()
         UNUserNotificationCenter.current().removeAllDeliveredNotifications()
         UIApplication.shared.applicationIconBadgeNumber = 0
+        
+        unregisterForRemoteNotifications()
         
         Task {
             // First log out from the server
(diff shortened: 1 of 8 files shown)


Velin92 (approved):

ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -527,11 +527,13 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationCoordinatorDelegate,
             return
         }
         
-        // The user will log out, clear any existing notifications
+        // The user will log out, clear any existing notifications and unregister from receving new ones
         UNUserNotificationCenter.current().removeAllPendingNotificationRequests()
         UNUserNotificationCenter.current().removeAllDeliveredNotifications()
         UIApplication.shared.applicationIconBadgeNumber = 0
         
+        unregisterForRemoteNotifications()
```
pixlwave, line 535: Two thoughts:
- Do you think we should do this above the `isSoft` guard?
- Is it worth thinking about multi-account support in any of this?

pixlwave (approved):
Yep looks reasonable to me too 👍

ElementX/Sources/Application/AppCoordinator.swift
stefanceriu: I wouldn't bother with either of them. They're both mega edge cases and the app would auto-recover pretty quickly.

PR #2079, opened 2023-11-13
Add scaledFrame/scaledPadding APIs

alfogrillo:
This PR adds the `.scaledFrame` and `.scaledPadding` APIs with the goal of reducing the number of `@ScaledMetric` variables in the views, so making the code more concise and readable.


ElementX/Sources/Screens/ComposerToolbar/View/VoiceMessageRecordingButton.swift
```diff
@@ -48,20 +43,18 @@ struct VoiceMessageRecordingButton: View {
...
-                    .padding(idleImagePadding)
-                    .padding(4)
+                    .scaledPadding(10, relativeTo: .title)
             case .recording:
                 Asset.Images.stopRecording.swiftUIImage
                     .resizable()
                     .foregroundColor(.compound.iconOnSolidPrimary)
-                    .frame(width: recordingImageSize, height: recordingImageSize)
-                    .padding(recordingImagePadding)
+                    .scaledFrame(width: 24, height: 24, relativeTo: .title)
```
pixlwave, line 51: FWIW, any of these where the asset is square (generally 24x24) could be replaced with

```
CompoundIcon(customImage: Asset.Images.stopRecording.swiftUIImage) // default 24x24 relative to body.
CompoundIcon(customImage: Asset.Images.stopRecording.swiftUIImage, size: .medium, relativeTo: .title)
```

and not need the resizable & scaled frame. Doesn't have to happen now as, as everything gets migrated it'll happen naturally anyway, just pointing it out for completeness.

UnitTests/__Snapshots__/PreviewTests/test_reactionsSummaryView.1.png
pixlwave, line 1: This looks like it has switched to UK dates instead of US.

ElementX/Sources/Other/SwiftUI/Layout/ScaledFrameModifier.swift
```diff
@@ -0,0 +1,45 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import SwiftUI
+
+extension View {
+    func scaledFrame(width: CGFloat, height: CGFloat, alignment: Alignment = .center, relativeTo textStyle: Font.TextStyle? = nil) -> some View {
```
Velin92, line 20: I would change this API a bit, because in this way is not possible to provide a `scaledWidth` with a normal `height` or a `scaledHeight` with a normal `width`.
I would either add other two functions that offer in signature the hybrid functionality
Or the function could have 4 width and height paraemters, 2 for the normal and 2 for the scaled (however I imagine we should decide which takes precedence)
Another solution is to just call it `frame` as the original function and have widht and height be provided as an enum that can either be:
`scaled(CGFloat, relativeTo: Font.TextStyle)`
`absolute(CGFloat)`

This is because while chaining padding is okay, chaining frames can create weird behaviour so hybrid cases where we have a scaled with and a scaled height might be difficult to implement.

Velin92 (approved):
LGTM
Added one comment, is not something however that we need to do know, might be useful for the future however

ElementX/Sources/Other/SwiftUI/Layout/ScaledFrameModifier.swift
```diff
@@ -0,0 +1,45 @@
...
+    }
+    
+    func body(content: Content) -> some View {
+        content.frame(width: width, height: height, alignment: alignment)
+    }
+}
+
+extension ScaledMetric {
+    init(wrappedValue: Value, relativeTo textStyle: Font.TextStyle?) {
+        self = textStyle.map { .init(wrappedValue: wrappedValue, relativeTo: $0) } ?? .init(wrappedValue: wrappedValue)
```
alfogrillo, line 43: The [documentation](https://developer.apple.com/documentation/swiftui/scaledmetric/init(wrappedvalue:)) isn't clear about what is the "default scaling" (people [say](https://useyourloaf.com/blog/the-@scaledmetric-property-wrapper/) is `.body`).
This code doesn't make any assumption on the default (`nil` is mapped to the "default" calling the other init).
Another option is removing optionality from `Font.TextStyle?` and give `.body` as default value.

pixlwave (approved):
Looks great to me! It would be amazing to have a `.scaledFrame(size: 16)` that sets width & height together as 95% of these are square.

Also its a bit surprising to see so many snapshots have changed. Is that expected given this is meant to be a change that has 0 impact on the layout?

UnitTests/__Snapshots__/PreviewTests/test_reactionsSummaryView.1.png
alfogrillo: Thanks for noticing! 
Fixed hopefully [here](https://github.com/vector-im/element-x-ios/pull/2079/commits/01b4f77bcbf68893f5f238d0c29f61606055df0e).

ElementX/Sources/Screens/ComposerToolbar/View/VoiceMessageRecordingButton.swift
alfogrillo: Addressed [here](https://github.com/vector-im/element-x-ios/pull/2079/commits/e1421e647a3019e320587a000b7cb3df9c2e88db).

alfogrillo:
> Looks great to me! It would be amazing to have a `.scaledFrame(size: 16)` that sets width & height together as 95% of these are square.

Yep, I thought about this as well.
I will add an overload. 👍🏻

> Also its a bit surprising to see so many snapshots have changed. Is that expected given this is meant to be a change that has 0 impact on the layout?

I'm surprised as well, I expected most of the snapshots to stay the same.
One of the most surprising was the preview of `RoomAttachmentPicker` given it's a single button.
I'm not sure what's happening, but I can tell you changes are very slight offsets.
Given the example above: do you have any insight?

ElementX/Sources/Other/SwiftUI/Layout/ScaledFrameModifier.swift
alfogrillo: I know, you can't do everything we do with `.frame` using this API.
One of my attempt was to have both `width` and `height` optional so that the API can accept just one of them.
The problem was the `@ScaledMetric` doesn't work with optionals, and instantiating it directly (type `ScaledMetric<Value>?`) didn't work as well.
Maybe we keep it simple for now and expand it in the future if needed.

stefanceriu (approved):
Very nice, love it! 👌

ElementX/Sources/Other/SwiftUI/Layout/ScaledFrameModifier.swift
alfogrillo: We decided privately to put a fallback to `.body`.
Applied [here](https://github.com/vector-im/element-x-ios/pull/2079/commits/47669d8bebed69fb7e448027800187f62c5abb24)

PR #2080, opened 2023-11-14
Add SwiftLint rule to enforce stack spacing.

pixlwave:
As part of #2007 I noticed that nested stack behaviour has changed when building with the new SDK. Whilst we're still waiting for a solution on the CI side, this PR adds SwiftLint rules to ensure all `HStack` and `VStack`s have the spacing parameter (and then goes through the codebase and fixes them by adding spacing and watching the preview to get the right value).

Some snapshots have changed as a result of preview stacks and so don't reflect changes in the app. There are some layout changes to the plain timeline style too:
- There was random trailing spacing meaning that items didn't line up with the trailing edge of the timestamp. This is now fixed.
- The spacing between the thread decorator, reply preview and message has ended up collapsing slightly. We don't have designs so there's nothing to go on here, but it helps pull all the content from the same message together to me.


ElementX/Sources/Other/MapLibre/MapLibreStaticMapView.swift
```diff
@@ -86,7 +86,7 @@ struct MapLibreStaticMapView<PinAnnotation: View>: View {
         } label: {
             placeholderImage
                 .overlay {
-                    VStack {
+                    VStack(spacing: 0) {
```
Velin92, line 89: isn't the default value a specing of 10?

Velin92 (approved):

ElementX/Sources/Other/MapLibre/MapLibreStaticMapView.swift
pixlwave: Honestly its random to the maxxx. I think it takes into account a mixture of factors such as 
- what kind of views are being spaced.
- what is the spacing of the parent if nested.
- potentially how much space is available in some cases too.

pixlwave: But generally I've found it to be 8.

PR #2081, opened 2023-11-14
Added a disclaimer in Notification Settings for "Mentions Only"

### LINKED ISSUE #2078: Mentions: Add a warning on servers where MSC 4028 is not supported
We check through the API I exposed https://github.com/matrix-org/matrix-rust-sdk/pull/2840 if the homeserver supports MSC 4027 and if not, an alert saying something like "be aware that your homeserver does not support the Mentions only options for ecrypted rooms which may result in the loss of all pushes if enabled" (final copy TBD) is shown.

This should be shown in Advanced Settings -> Notifications -> Group or Direct, as soon as the view is loaded
This should also be shown in Room Details -> Notifications -> Custom Settings on, **but only and only if the room is encrypted** otherwise we can avoid that

We still allow the user to do any change, we just inform them that it might affect encrypted rooms.

Velin92:
fixes #2078
The disclaimer is slightly different between the global settings and an encrypted room custom setting.
For a non ecrypted room custom settings the disclaimer will not appear (doesn't matter if accessed from the global setting or from the room settings)
If the homeserver supports MSC 4028 the disclaimer won't appear at all in any case, an API is exposed on the SDK to allow that

Requires the following [PR](https://github.com/matrix-org/matrix-rust-sdk/pull/2840) to be merged in the Rust SDK

All the following cases are tested on a sever that does not support 4028
custom setting for a non encrypted room (no change)
Simulator Screenshot - iPhone 14 - 2023-11-14 at 16 13 38
custom setting for an encrypted room
Simulator Screenshot - iPhone 14 - 2023-11-14 at 16 13 32
custom setting for an encrypted room, from room details
Simulator Screenshot - iPhone 14 - 2023-11-14 at 16 12 56
global settings
Simulator Screenshot - iPhone 14 - 2023-11-14 at 16 12 35


ElementX/Sources/Screens/Settings/NotificationSettingsEditScreen/NotificationSettingsEditScreenModels.swift
```diff
@@ -33,11 +33,20 @@ struct NotificationSettingsEditScreenViewState: BindableState {
     var defaultMode: NotificationSettingsEditScreenDefaultMode?
     var pendingMode: NotificationSettingsEditScreenDefaultMode?
     var roomsWithUserDefinedMode: [NotificationSettingsEditScreenRoom] = []
+    var canHomeServerPushEncryptedEvents = false
```
nimau, line 36:
```suggestion
    var canHomeserverPushEncryptedEvents = false
```
because it's called `xxxHomeserverXxx` everywhere else

PR #2091, opened 2023-11-15
Artificial delay when a message fails to send

### LINKED ISSUE #1352: Hitting resend on failed msg doesn’t reset its send status
### Steps to reproduce

1. Tried to send two msgs
2. For whatever reason (bad connectivity? server problems?) they failed.
3. Conversation moved on
4. I hit retry on the first one
5. It moved to the bottom of the timeline, but was still marked as failed-send (i guess it might have immediately failed?). i expected it to move back to sending/unsent state, not failed
6. I then did the second one; same result. Plus very annoying to have to individually resend them.

#### What did you expect?

resending should be shown as resending, not failed.

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

292

### Homeserver

_No response_

### Will you send logs?

Yes

Velin92:
fixes #1352 

Since in poor internet conditions (or when completely offline) the resend action fails pretty much immediately there is no visual indication that it has failed again. so the idea is to create an artificial delay of about 0.5 seconds when the status transitions to `sendingFailed` to make it look like for a brief moment that is being sent to give a visual cue that the user that the app has actively tried to resend the message

https://github.com/vector-im/element-x-ios/assets/34335419/c15c6dfa-9a64-4acd-b327-9acbe9bbad4c

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/View/Style/TimelineStyler.swift
+++ b/ElementX/Sources/Screens/RoomScreen/View/Style/TimelineStyler.swift
@@ -24,13 +24,43 @@ struct TimelineStyler<Content: View>: View {
 
     let timelineItem: EventBasedTimelineItemProtocol
     @ViewBuilder let content: () -> Content
+    
+    @State private var status: TimelineItemDeliveryStatus?
+    @State private var task: Task<Void, Never>?
 
     var body: some View {
+        mainContent
+            .onChange(of: timelineItem.properties.deliveryStatus) { newStatus in
+                MXLog.info("Delivery status: \(newStatus)")
+                if newStatus == .sendingFailed {
+                    guard task == nil else {
+                        return
+                    }
+                    task = Task {
+                        if status == nil {
+                            status = .sending
+                        }
+                        try? await Task.sleep(for: .milliseconds(500))
+                        if !Task.isCancelled {
+                            status = newStatus
+                        }
+                        task = nil
+                    }
+                } else {
+                    task?.cancel()
+                    task = nil
+                    status = newStatus
+                }
+            }
+    }
+    
+    @ViewBuilder
+    var mainContent: some View {
         switch style {
         case .plain:
-            TimelineItemPlainStylerView(timelineItem: timelineItem, content: content)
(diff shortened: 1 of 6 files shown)


changelog.d/2078.feature
stefanceriu, line 1: Hmm.. what's this about?

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineStyler.swift
```diff
@@ -24,13 +24,46 @@ struct TimelineStyler<Content: View>: View {
 
     let timelineItem: EventBasedTimelineItemProtocol
     @ViewBuilder let content: () -> Content
+    
+    @State private var deliveryStatus: TimelineItemDeliveryStatus?
```
stefanceriu, line 28: We should probably rename this, it's very easy to get confused between it and the one from the timelineItem's properties. Took me a while to figure out why we don't just use the properties everywhere. I'm thinking `transientDeliveryStatus` or maybe `adjustedDeliveryStatus`

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineStyler.swift
```diff
@@ -24,13 +24,46 @@ struct TimelineStyler<Content: View>: View {
...
 
     var body: some View {
+        mainContent
+            .onChange(of: timelineItem.properties.deliveryStatus) { newStatus in
+                if newStatus == .sendingFailed {
+                    guard task == nil else {
+                        return
+                    }
+                    task = Task {
+                        try? await Task.sleep(for: .milliseconds(500))
```
stefanceriu, line 45: I would make this even slower to be honest

stefanceriu (approved):
Good stuff! 👍

changelog.d/2078.feature
Velin92: previous PR forgot to add the changelog

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineStyler.swift
Velin92: Checked with @amshakal she feels 0.5 is okay

PR #2095, opened 2023-11-16
Secure backup tweaks

stefanceriu:
* Show a loading indicator when confirming the recovery key
* Hide the passphrase characters when confirming the recovery key
* Don't show the settings security section until session verification state is retrieved (async)


nimau (approved):
LGTM

PR #2113, opened 2023-11-17
Switch to Xcode 15 (with the iOS 16.4 Simulator for tests still)

pixlwave:
Follow up from #2007, this switches the project to use Xcode 15, but with the iOS 16.4 simulator [linked from Xcode 14](https://github.com/actions/runner-images/issues/8402#issuecomment-1797918209) on CI to work around the performance issues.

Updated Preview snapshots are mainly about the default change of continuous corners on rounded rectangles and capsules. There are some rendering errors but those don't appear in real life.

UI tests and integration tests both work locally too and the precision is lower on UI tests so it doesn't pick up the corner shape change.


UnitTests/__Snapshots__/PreviewTests/test_invitesScreenCell.1.png
pixlwave, line 1: These do look stupid but I can't see them anywhere in the real app (or in UI test snapshots).

stefanceriu (approved):
🕺

alfogrillo (approved):
LGTM 🎉

Velin92 (approved):
LGTM!

PR #2118, opened 2023-11-20
Translations update

RiotRobot:
- Translations update


ElementX/Sources/Generated/Strings.swift
```diff
@@ -10,10 +10,6 @@ import Foundation
 // swiftlint:disable explicit_type_interface function_parameter_count identifier_name line_length
 // swiftlint:disable nesting type_body_length type_name vertical_whitespace_opening_braces
 public enum L10n {
-  /// Read by %1$@ and %2$@
-  public static func a11ReadReceiptsMultiple(_ p1: Any, _ p2: Any) -> String {
```
stefanceriu, line 14: @Velin92 Seems we use this in code, what's the new version supposed to be?

Velin92: checked with @bmarty and this was canged to a11**y**ReadReceiptsMultiple

PR #2131, opened 2023-11-21
Fix: Long replies overlay the replied message (#1832)

### LINKED ISSUE #1832: Long replies overlay the replied message.
### Steps to reproduce

I'm on the timeline for a room, I click reply, I write a long reply. While still composing, if I scroll up and down the text of my reply is overlaid on the text of the original message

#### What did you expect?
I would have expected the original message to not be overlaid with text on text rendering both my reply and the original message illegible. 

#### What happened instead?
I could not read my reply or the original message (this is especially problematic since I am still composing my reply).

I could send a screen shot, but thus far all examples have been in private rooms so I don't want to attach to the issue. 

### Your phone model

iPhone 11

### Operating system version

16.6.1

### Application version

1.3.2(398)

### Homeserver

matrix.org

### Will you send logs?

No

nimau:
This PR fixes #1832

### DIFF
--- a/ElementX/Sources/Screens/ComposerToolbar/View/MessageComposer.swift
+++ b/ElementX/Sources/Screens/ComposerToolbar/View/MessageComposer.swift
@@ -70,6 +70,7 @@ struct MessageComposer: View {
             Color.clear
                 .overlay(alignment: .top) {
                     composerView
+                        .clipped()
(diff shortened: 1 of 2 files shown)


stefanceriu (approved):
🚀

PR #2135, opened 2023-11-21
Update CONTRIBUTING.md

pixlwave:
Small follow-up to #2113 which updates some relevant docs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,7 +21,7 @@ This will:
 
 ### Xcode
 
-We suggest using an Xcode version later than 13.2.1.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -55,12 +55,13 @@ Git LFS is used to store UI test snapshots. `swift run tools setup-project` will
 
 ```
 git lfs install
+ln -s "$(which git-lfs)" "$(git --exec-path)/git-lfs"
```
pixlwave, line 58: I think this is a necessity again for some obscure reason, but for now let's leave it here as a separate step until we confirm

stefanceriu (approved):

Issue #1999, opened 2023-11-01
Mention pills fall back to plain text in markdown mode


Jaffex:
### Steps to reproduce

Disable the Rich-Text-Editor in the settings.

Go to a chat room and mention another user by beginning to type "@" and their username.

Select the user to mention from the appearing dropdown.

The message composer will now contain a mention pill and you can compose the rest of your message. Hit send.

#### What did you expect?
The mention pill to be visible and interactive in the timeline after sending the message, regardless of which style of message composer I use

#### What happened instead?
Your message in the timeline will show up as plain text, and is not interactive as a mention pill usually is.

### Additional information
With the Rich-Text-Editor enabled, the same procedure will result in a working interactive mention pill in the timeline.

### Your phone model

iPhone 12 mini

### Operating system version

iOS 17.1

### Application version

1.4.0

### Homeserver

Synapse

### Will you send logs?

No

manuroe:
@Velin92  can we close this issue thanks to https://github.com/element-hq/element-x-ios/pull/2246?

Velin92:
More than a fix we have prevented the user suggestions to appear entirely when in plain text mode since pills are only supported in RTE as of right now. We have this tracked in the RTE repo:
https://github.com/matrix-org/matrix-rich-text-editor/issues/708

PR #2141, opened 2023-11-22
Always send formatted content

langleyd:
Fixes https://github.com/vector-im/element-x-ios/issues/1999

I can understand the rational behind why the code that removes formatted_body for non-RTE mode might have been added but whether we send 'formatted_body' should not be tied to the RTE, we should always send it(e.g. the web client sends it even on the legacy client which is markdown based, as do legacy mobile markdown clients). 

It is the preferred field for clients that support displaying rich content, whereas `body` field is a plain text representation of what is in `formatted_body` and can be used for clients that don't support displaying rich content.

`formatted_body` is required for mentions to work and render correctly in the timeline, hence the bug.

### DIFF
--- a/ElementX.xcodeproj/xcshareddata/xcschemes/ElementX.xcscheme
+++ b/ElementX.xcodeproj/xcshareddata/xcschemes/ElementX.xcscheme
@@ -4,8 +4,7 @@
    version = "1.7">
    <BuildAction
       parallelizeBuildables = "YES"
-      buildImplicitDependencies = "YES"
-      runPostActionsOnFailure = "NO">
+      buildImplicitDependencies = "YES">
       <BuildActionEntries>
          <BuildActionEntry
             buildForTesting = "YES"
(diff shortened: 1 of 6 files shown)


pixlwave:
> I can understand the rational behind why the code that removes formatted_body for non-RTE mode might have been added but whether we send 'formatted_body' should not be tied to the RTE, we should always send it(e.g. the web client sends it even on the legacy client which is markdown based, as do legacy mobile markdown clients).

I don't think its that simple. We always send the `formatted_body` when it isn't included, but Rust builds it from the Markdown in the `body`. So with this change, sending a message with Markdown will no longer work unless the RTE parses it first.

pixlwave (changes requested):
An example, this is from `develop`:
```
"content": {
    "body": "This is **bold**.",
    "format": "org.matrix.custom.html",
    "formatted_body": "<p>This is <strong>bold</strong>.</p>\n",
    "m.mentions": {},
    "msgtype": "m.text"
},
```

And this is with the changes:
```
"content": {
    "body": "This is **bold**.",
    "format": "org.matrix.custom.html",
    "formatted_body": "This is **bold**.",
    "m.mentions": {},
    "msgtype": "m.text"
}
```

Presumably we either need to fix the pills some other way or enable MD input support in the RTE.

stefanceriu:
Handled in https://github.com/element-hq/element-x-ios/pull/2246

PR #2142, opened 2023-11-22
Using thumbnails for notification icons

### LINKED ISSUE #2136: Profile picture is sideways in notification
### Steps to reproduce

Send a message to someone using EX
See their avatar is rotated in the notification 90 degrees left

The original profile photo was uploaded from an iPhone in app

Build 1.4.2
iOS 17.0.3
iPhone 12 Pro

#### What did you expect?
Profile photo in notification to be the right way up 
#### What happened instead?
It was rotated 

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

_No response_

### Homeserver

Synapse 

### Will you send logs?

No

Velin92:
It is technically more efficient to use small thumbnails of 50x50 size for notification icons instead of loading the original image data, since their size would be way less but most importantly the server is also going to handle things like orientation, which seems that with iOS 17 is handled poorly by the Notification Center system.

fixes #2136

### DIFF
--- a/ElementX/Sources/Other/Extensions/UNNotificationContent.swift
+++ b/ElementX/Sources/Other/Extensions/UNNotificationContent.swift
@@ -117,7 +117,7 @@ extension UNMutableNotificationContent {
         var fetchedImage: INImage?
         let image: INImage
         if let mediaSource = icon.mediaSource {
-            switch await mediaProvider?.loadImageDataFromSource(mediaSource) {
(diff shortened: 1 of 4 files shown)


stefanceriu (approved):
Nice one, good catch and fix 👍

wrjlewis:
Thanks @Velin92 for the super speedy fix to my issue!

PR #2146, opened 2023-11-22
Fixes #2145 - Hide pin code in setup mode

### LINKED ISSUE #2145: Hide pin code in setup mode
When setting up a pin code the numbers should be hidden (black dot)

stefanceriu:


pixlwave (approved):
👍

Changelog perhaps?

Issue #1915, opened 2023-10-18
Markdown lists at the beginning of messages are rendered incorrectly


marwing:
### Steps to reproduce

1. Send a message containing only a markdown list

#### What did you expect?

All list items to be formatted the same (as is happening in Element iOS and Web). Similar to this (but without the A):
![IMG_3507](https://github.com/vector-im/element-x-ios/assets/29716248/c8a0814d-03e4-4329-a7de-3d4f7a45f772)

#### What happened instead?

The first list item was rendered looking like an indented paragraph instead of a list item.
![IMG_3504](https://github.com/vector-im/element-x-ios/assets/29716248/70bee7f3-f48d-4d58-9c2f-7b7539d4effd)
This also happens with short lists with very short list items:
![IMG_3506](https://github.com/vector-im/element-x-ios/assets/29716248/f2b12c2f-d05d-4309-a55f-9713b50bc6ed)

### Your phone model

iPhone 13 mini / iPad Pro

### Operating system version

iOS/iPadOS 17.0.3

### Application version

1.3.3 (76)

### Homeserver

Synapse 1.94.0

### Will you send logs?

No

langleyd:
Fixes with https://github.com/vector-im/element-x-ios/pull/2152

PR #2152, opened 2023-11-22
Fix a rendering issue when a message starts with a list (#1915)

nimau:
This PR fixes an issue with the display of messages when they begin with a list (#1915)

Before:
Screenshot 2023-11-22 at 16 16 18

After:
Screenshot 2023-11-22 at 16 16 45

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/View/Timeline/FormattedBodyText.swift
+++ b/ElementX/Sources/Screens/RoomScreen/View/Timeline/FormattedBodyText.swift
@@ -33,7 +33,12 @@ struct FormattedBodyText: View {
     }()
         
     private var attributedComponents: [AttributedStringBuilderComponent] {
-        var adjustedAttributedString = AttributedString(layoutDirection.isolateLayoutUnicodeString) + attributedString + AttributedString(additionalWhitespacesSuffix)
+        var adjustedAttributedString = attributedString + AttributedString(additionalWhitespacesSuffix)
+        
+        // If this is not a list, force the writing direction by adding the correct unicode character.
+        if !String(attributedString.characters).starts(with: "\t") {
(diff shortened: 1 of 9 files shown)


stefanceriu (commented):
Looks sensible to me but I would like mauro's oppinion on it 👍

ElementX/Sources/Screens/RoomScreen/View/Timeline/FormattedBodyText.swift
```diff
@@ -33,7 +33,12 @@ struct FormattedBodyText: View {
     }()
         
     private var attributedComponents: [AttributedStringBuilderComponent] {
-        var adjustedAttributedString = AttributedString(layoutDirection.isolateLayoutUnicodeString) + attributedString + AttributedString(additionalWhitespacesSuffix)
+        var adjustedAttributedString = attributedString + AttributedString(additionalWhitespacesSuffix)
+        
+        // If this is not a list, force the writing direction by adding the correct unicode character.
+        if !String(attributedString.characters).starts(with: "\t") {
+            adjustedAttributedString = AttributedString(layoutDirection.isolateLayoutUnicodeString) + adjustedAttributedString
```
Velin92, line 40: It is possible we want to add the isolate for RTL always and add it on LTR only when it does not start with a list, since this might break RTL languages, but let's a check it.
If if there is an issue, it's okay we can always fix it later, definitely let's give priority to LTR since most of our supported languages are LTR.

nimau: If we add the isolate for RTL, even if it starts with a list, we have the same problem where the first element got an offset.

Velin92 (approved):
I think that fixing the issue for LTR is enough, we can revisit RTL later when we have translation that support it

PR #2155, opened 2023-11-22
Fix a bug where the Screen lock placeholder could be slow.

pixlwave:
Seems the timing was off with the `Task` when locking the device with the app open.

**Before:**

https://github.com/vector-im/element-x-ios/assets/6060466/83ae35fa-c105-433f-b74a-568e7dc98af1

**After:**

https://github.com/vector-im/element-x-ios/assets/6060466/5264f896-6840-4c7e-b65b-91ddc9b54402


stefanceriu (approved):
looks sensible to me 👍

PR #2160, opened 2023-11-23
Fixing a crash that happened when leaving the room through the details screen

Velin92:


pixlwave (approved):
LGTM, maybe worth asking someone else too given we came at this from the same direction.

Also I'm wondering about that comment that's removed. Is this code called if you push a new room from a notification when already in the room details of a different room?

PR #2162, opened 2023-11-24
Fixes #2005 - Remove rooms from the room list after leaving them

### LINKED ISSUE #2005: DM is still in room list after leaving it
### Steps to reproduce

1. Leave a DM

#### What did you expect?
DM is no longer in the room list

#### What happened instead?
DM is still in the room list

There are three scenarios:

1. User leaves room
If a user leaves a room the room should disappear from the room-list. 

2. User gets kicked from a room
If a user gets kicked from a room the room should stay in the room list.
- The user should still have access to the history that they have been a part of (UX - is that right?)
- The room should not be updated.
- The user cannot write in the room anymore
- The user can join the room again. To join a room the user needs to go through the same invitation flow as a user that has never been a member.

3. User is banned from a room
The room should disappear from the roomlist.
The user can only re-join when they are un-banned by a moderator.

stefanceriu:
Left rooms will still be available if seached by name

### DIFF
--- a/ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
+++ b/ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
@@ -151,7 +151,7 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
         case .none:
             _ = listUpdatesSubscriptionResult?.controller.setFilter(kind: .none)
         case .all:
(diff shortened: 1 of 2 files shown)


manuroe:
> Left room swill still be available is seached by name

@bmarty can we have the same behavior on Android?

ganfra:
> > Left room swill still be available is seached by name
> 
> @bmarty can we have the same behavior on Android?

Not with the current behavior

pixlwave (approved):

PR #2165, opened 2023-11-24
vector-im/element-meta/issues/2213 - Automatically pop the invites li…

stefanceriu:
…st when acting on the last one. Coming back from the room will go directly back to the room list

### DIFF
--- a/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
@@ -97,6 +97,12 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
             switch action {
             case .presentedRoom(let roomID):
                 analytics.signpost.beginRoomFlow(roomID)
+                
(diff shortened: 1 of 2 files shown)


ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -97,6 +97,12 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
             switch action {
             case .presentedRoom(let roomID):
                 analytics.signpost.beginRoomFlow(roomID)
+                
+                let availableInvitesCount = userSession.clientProxy.inviteSummaryProvider?.roomListPublisher.value.count ?? 0
+                if availableInvitesCount == 1 {
```
pixlwave, line 102: Should this also have a `case .invitesScreen = stateMachine.state` on it too? Otherwise what happens if I open a room when I have a single invite (which I'm ignoring).

stefanceriu: Yeah, very good point, broke this just before committing

PR #2166, opened 2023-11-24
Add a state machine to AppLockFlowCoordinator.

pixlwave:
Previously, cancelling Face ID would loop back round and present the Face ID prompt again when becoming active. Fixing that revealed a few more edge cases one-by-one so I added a state machine to drive the flow instead.

https://github.com/vector-im/element-x-ios/assets/6060466/ba52e5cf-a561-4b5a-997d-1071322a7c8a

### DIFF
--- a/ElementX/Sources/FlowCoordinators/AppLockFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/AppLockFlowCoordinator.swift
@@ -90,14 +90,15 @@ class AppLockFlowCoordinator: CoordinatorProtocol {
         actionsSubject.eraseToAnyPublisher()
     }
     
-    init(appLockService: AppLockServiceProtocol,
+    init(initialState: State = .initial,
+         appLockService: AppLockServiceProtocol,
          navigationCoordinator: NavigationRootCoordinator,
          notificationCenter: NotificationCenter = .default) {
         self.appLockService = appLockService
         self.navigationCoordinator = navigationCoordinator
         
-        // The app starts off in the background (with the placeholder screen in the flow.)
-        stateMachine = .init(state: .initial)
+        // Set the initial state and start with the placeholder screen as the root view.
+        stateMachine = .init(state: initialState)
         showPlaceholder()
         
         notificationCenter.publisher(for: UIApplication.willResignActiveNotification)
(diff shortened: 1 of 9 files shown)


ElementX/Sources/FlowCoordinators/AppLockFlowCoordinator.swift
```diff
@@ -31,111 +32,203 @@ class AppLockFlowCoordinator: CoordinatorProtocol {
...
+        /// The app is resigning active (going into the app switcher, showing system UI like Face ID, permissions prompt etc).
+        case willResignActive
+        /// The app is now backgrounded and not visible to the user.
+        case didEnterBackground
+        /// The app is in the foreground and has been given focus.
+        case didBecomeActive
+        /// Biometric unlock has completed with the following result.
+        case biometricResult(AppLockServiceBiometricResult)
+        /// The entered PIN code was accepted.
+        case pinSuccess
```
stefanceriu, line 68: Pretty much the same comment as above. I don't want to be pedantic but it would be nice for them to be the same if possible.

ElementX/Sources/FlowCoordinators/AppLockFlowCoordinator.swift
```diff
@@ -31,111 +32,203 @@ class AppLockFlowCoordinator: CoordinatorProtocol {
...
+    }
+    
+    private let stateMachine: StateMachine<State, Event>
+    
     /// A task used to await biometric unlock before showing the PIN screen.
     @CancellableTask private var unlockTask: Task<Void, Never>?
     private var cancellables: Set<AnyCancellable> = []
     
+    /// Whether or not biometric unlock should be attempted instead of asking for a PIN.
+    private var biometricUnlockIsAvailable: Bool {
```
stefanceriu, line 84: `isBiometricUnlockAvailable` maybe?

ElementX/Sources/FlowCoordinators/AppLockFlowCoordinator.swift
```diff
@@ -31,111 +32,203 @@ class AppLockFlowCoordinator: CoordinatorProtocol {
...
+            case (_, .didEnterBackground):
+                return .backgrounded
+            case (.backgrounded, .didBecomeActive), (.initial, .didBecomeActive):
+                guard appLockService.computeNeedsUnlock(didBecomeActiveAt: .now) else { return .unlocked }
+                return biometricUnlockIsAvailable ? .biometricUnlock : .pinCodeUnlock
+            case (.biometricUnlock, .biometricResult(let result)):
+                return .biometricUnlockDismissing(result) // Transitional state until the app becomes active again.
+            case (.biometricUnlockDismissing(let result), .didBecomeActive):
+                return switch result {
+                case .unlocked: .unlocked
```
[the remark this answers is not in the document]
pixlwave: Errrr, I think maybe Danger's SwiftLint is old and doesn't know about if/switch expressions?

stefanceriu (approved):
Very nice addition, everything is clearer to follow now 👏

PR #2167, opened 2023-11-24
Update poll form localization keys

alfogrillo:
Updates a couple of keys for the poll form alert

### DIFF
--- a/ElementX/Resources/Localizations/en.lproj/Localizable.strings
+++ b/ElementX/Resources/Localizations/en.lproj/Localizable.strings
@@ -360,8 +360,9 @@
 "screen_create_poll_anonymous_desc" = "Show results only after poll ends";
 "screen_create_poll_anonymous_headline" = "Hide votes";
 "screen_create_poll_answer_hint" = "Option %1$d";
-"screen_create_poll_discard_confirmation" = "Your changes won’t be saved";
-"screen_create_poll_discard_confirmation_title" = "Cancel Poll";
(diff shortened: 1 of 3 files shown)


ElementX/Resources/Localizations/en.lproj/Localizable.strings
```diff
@@ -360,8 +360,9 @@
 "screen_create_poll_anonymous_desc" = "Show results only after poll ends";
 "screen_create_poll_anonymous_headline" = "Hide votes";
 "screen_create_poll_answer_hint" = "Option %1$d";
-"screen_create_poll_discard_confirmation" = "Your changes won’t be saved";
-"screen_create_poll_discard_confirmation_title" = "Cancel Poll";
+"screen_create_poll_cancel_confirmation_android_content" = "Your changes have not been saved. Are you sure you want to go back?";
+"screen_create_poll_cancel_confirmation_ios_content" = "Your changes won’t be saved";
+"screen_create_poll_cancel_confirmation_ios_title" = "Cancel Poll";
```
pixlwave, line 365: If these keys end in `_ios` and `_android` then they'll be filtered out from the opposite platform during download.

pixlwave (approved):
LGTM, but might want to rename the keys so they're filtered.

ElementX/Resources/Localizations/en.lproj/Localizable.strings
alfogrillo: We didn't know!
Applied [here](https://github.com/vector-im/element-x-ios/pull/2167/commits/60f4f3880b5c45c64132f882cd44cb2fd5b05f5f)

PR #2170, opened 2023-11-27
Translations update

RiotRobot:
- Translations update


ElementX/Resources/Localizations/en.lproj/Localizable.strings
```diff
@@ -238,7 +238,6 @@
 "notification_invitation_action_join" = "Join";
 "notification_invitation_action_reject" = "Reject";
 "notification_invite_body" = "Invited you to chat";
-"notification_mentioned_you_body" = "Mentioned you: %1$@";
```
nimau, line 241: @Velin92 I don't find this string in Localazy, should we remove its use from [here](https://github.com/vector-im/element-x-ios/blob/3d76f06b0d79f690a1ad3772f6cff7115b41776c/NSE/Sources/NotificationContentBuilder.swift#L105) ?

nimau (commented):
There a string missing from Localazy.

nimau (approved):
Missing string has been restored on Localazy

PR #2174, opened 2023-11-27
Fix the touch area size for the voice message button.

### LINKED ISSUE #2038: Hitmask on voice message buttons is too small
### Steps to reproduce

1. Try to click play on a voice message in the timeline on macOS
2. Discover that only the icon itself is a hit target; the rest of the target ignores you:

https://github.com/vector-im/element-x-ios/assets/1294269/0f0f826d-fa15-4e50-8f3a-23f1d4f58315

I assume the same is true on iOS too, but is harder to trigger due to literal fat fingers.

#### What did you expect?

Buttons which click when you press them

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

429

### Homeserver

_No response_

### Will you send logs?

No

nimau:
This PR fixes #2038

### DIFF
--- a/ElementX/Sources/Other/VoiceMessage/VoiceMessageButton.swift
+++ b/ElementX/Sources/Other/VoiceMessage/VoiceMessageButton.swift
@@ -49,7 +49,6 @@ struct VoiceMessageButton: View {
         Button(action: action) {
             buttonLabel
                 .frame(width: buttonSize, height: buttonSize)
-                .background(Circle().foregroundColor(.compound.bgCanvasDefault))
         }
         .animation(nil, value: state)
         .buttonStyle(VoiceMessageButtonStyle())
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Other/VoiceMessage/VoiceMessageButton.swift
```diff
@@ -49,11 +49,11 @@ struct VoiceMessageButton: View {
         Button(action: action) {
             buttonLabel
                 .frame(width: buttonSize, height: buttonSize)
+                .background(Circle().foregroundColor(.compound.bgCanvasDefault))
```
pixlwave, line 52: It might be nice to apply this background in the `VoiceMessageButtonStyle`, that way you can change the colour when it's pressed.

pixlwave (approved):
LGTM, one optional idea for the pressed state.

ElementX/Sources/Other/VoiceMessage/VoiceMessageButton.swift
nimau: I've updated the code.

PR #2186, opened 2023-11-29
Access direct messages directly from the member details screen

stefanceriu:
This PR adds a button on the room member's detail screen to open an existing direct message (or create it). In order to do so and not break the navigation coordination a lot of methods have been moved from the room detiail screen coordinator and into the RoomFlowCoordinator.

This PR can and should be reviewed commit by commit

### DIFF
--- a/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
@@ -124,92 +124,92 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
     // swiftlint:disable:next function_body_length
     private func setupStateMachine() {
         stateMachine.addRouteMapping { event, fromState, _ in
-            switch (event, fromState) {
-            case (.presentRoom(let roomID), _):
+            switch (fromState, event) {
+            case (_, .presentRoom(let roomID)):
                 return .room(roomID: roomID)
-            case (.dismissRoom, .room):
+            case (.room, .dismissRoom):
                 return .initial
                 
-            case (.presentRoomDetails(let roomID), .initial):
+            case (.initial, .presentRoomDetails(let roomID)):
                 return .roomDetails(roomID: roomID, isRoot: true)
-            case (.presentRoomDetails(let roomID), .room(let currentRoomID)):
+            case (.room(let currentRoomID), .presentRoomDetails(let roomID)):
                 return .roomDetails(roomID: roomID, isRoot: roomID != currentRoomID)
-            case (.presentRoomDetails(let roomID), .roomDetails(let currentRoomID, _)):
+            case (.roomDetails(let currentRoomID, _), .presentRoomDetails(let roomID)):
                 return .roomDetails(roomID: roomID, isRoot: roomID != currentRoomID)
-            case (.dismissRoomDetails, .roomDetails(let roomID, _)):
+            case (.roomDetails(let roomID, _), .dismissRoomDetails):
                 return .room(roomID: roomID)
-            case (.dismissRoom, .roomDetails):
+            case (.roomDetails, .dismissRoom):
                 return .initial
                 
-            case (.presentRoomDetailsEditScreen, .roomDetails(let roomID, _)):
+            case (.roomDetails(let roomID, _), .presentRoomDetailsEditScreen):
                 return .roomDetailsEditScreen(roomID: roomID)
-            case (.dismissRoomDetailsEditScreen, .roomDetailsEditScreen(let roomID)):
+            case (.roomDetailsEditScreen(let roomID), .dismissRoomDetailsEditScreen):
                 return .roomDetails(roomID: roomID, isRoot: false)
                 
-            case (.presentNotificationSettingsScreen, .roomDetails(let roomID, _)):
+            case (.roomDetails(let roomID, _), .presentNotificationSettingsScreen):
                 return .notificationSettings(roomID: roomID)
-            case (.dismissNotificationSettingsScreen, .notificationSettings(let roomID)):
+            case (.notificationSettings(let roomID), .dismissNotificationSettingsScreen):
                 return .roomDetails(roomID: roomID, isRoot: false)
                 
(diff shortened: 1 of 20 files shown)


ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -820,7 +961,79 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
...
+            case .toggleUser(let user):
+                var selectedUsers = selectedUsersSubject.value
+                
+                if let index = selectedUsers.firstIndex(where: { $0.userID == user.userID }) {
+                    selectedUsers.remove(at: index)
+                } else {
+                    selectedUsers.append(user)
+                }
+                
+                selectedUsersSubject.send(selectedUsers)
```
pixlwave, line 1040: I'm not sure I understand why these 2 are exposed to the flow coordinator. They don't perform navigation. Shouldn't this stay internal to the screen and only signal upwards when something is complete?

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -841,15 +1054,19 @@ private extension RoomFlowCoordinator {
...
         case initial
         case room(roomID: String)
-        case reportContent(roomID: String, itemID: TimelineItemIdentifier, senderID: String)
         case roomDetails(roomID: String, isRoot: Bool)
+        case roomDetailsEditScreen(roomID: String)
+        case notificationSettings(roomID: String)
+        case globalNotificationSettings(roomID: String)
+        case roomMembersList(roomID: String)
+        case roomMemberDetails(roomID: String, member: HashableRoomMemberWrapper, fromRoomMembersList: Bool)
+        case inviteUsersScreen(roomID: String, fromRoomMembersList: Bool)
```
pixlwave, line 1099: Do we need to preserve `roomDetailsIsRoot` when navigating into and out of these screens?

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -867,6 +1084,24 @@ private extension RoomFlowCoordinator {
         
         case presentRoomDetails(roomID: String)
         case dismissRoomDetails
+        
+        case presentRoomDetailsEditScreen(accountOwner: HashableRoomMemberWrapper)
+        case dismissRoomDetailsEditScreen
```
pixlwave, line 1125: Whilst where here… Maybe we could call all these `dismissedXYZ`, because it very much sounds like the event will dismiss the screen rather than the event is tidying up after the screen has been dismissed.

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -853,6 +853,42 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
...
+                    case .success(.some(let roomID)):
+                        stateMachine.tryEvent(.presentRoom(roomID: roomID))
+                    case .success(nil):
+                        switch await userSession.clientProxy.createDirectRoom(with: member.userID, expectedRoomName: member.displayName) {
+                        case .success(let roomID):
+                            analytics.trackCreatedRoom(isDM: true)
+                            stateMachine.tryEvent(.presentRoom(roomID: roomID))
+                        case .failure:
+                            userIndicatorController.alertInfo = .init(id: UUID())
+                        }
```
pixlwave, line 882: Is this shared with the create DM flow? Maybe we should have a `createIfNone: Bool` parameter on `clientProxy.directRoomForUserID`?

pixlwave (approved):
LGTM 👍

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
pixlwave: Which seems to be what we use elsewhere now I've seen those too.

stefanceriu: There are situations in which the event actually cleans up the stack e.g. when deleting the custom notification options for a room. The way it is now caters to both imho.

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
stefanceriu: This is coming from the start chat flows in which the selected users are shared between 2 different screens

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -38,6 +38,7 @@ enum RoomFlowCoordinatorAction: Equatable {
     }
 }
 
+// swiftlint:disable file_length
```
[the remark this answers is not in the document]
stefanceriu: 😅

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
pixlwave: (The inviteUsers seems more curious to me in this regard).

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
stefanceriu: Not really, no. It's only useful when first presenting the roomDetailsScreen

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
stefanceriu: They are shared but they do slightly different things with the results and also call analytics. We could refactor it but let's tackle that outside this PR

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
stefanceriu: Talked about this offline, we'll tackle a refactoring separate from this PR

PR #2187, opened 2023-11-29
Element Call permission tweaks

stefanceriu:
vector-im/element-meta/issues/2230 - Specify `call.member` default po…wer levels when creating a room
vector-im/element-meta/issues/2229 - Take into account permissions when configuring the room screen call button

### DIFF
--- a/ElementX/Sources/Services/Client/ClientProxy.swift
+++ b/ElementX/Sources/Services/Client/ClientProxy.swift
@@ -208,7 +208,15 @@ class ClientProxy: ClientProxyProtocol {
     func createDirectRoom(with userID: String, expectedRoomName: String?) async -> Result<String, ClientProxyError> {
         let result: Result<String, ClientProxyError> = await Task.dispatch(on: clientQueue) {
             do {
-                let parameters = CreateRoomParameters(name: nil, topic: nil, isEncrypted: true, isDirect: true, visibility: .private, preset: .trustedPrivateChat, invite: [userID], avatar: nil)
+                let parameters = CreateRoomParameters(name: nil,
+                                                      topic: nil,
+                                                      isEncrypted: true,
+                                                      isDirect: true,
+                                                      visibility: .private,
+                                                      preset: .trustedPrivateChat,
+                                                      invite: [userID],
+                                                      avatar: nil,
+                                                      powerLevelContentOverride: Self.defaultRoomCreationPowerLevels)
                 let result = try self.client.createRoom(request: parameters)
                 return .success(result)
             } catch {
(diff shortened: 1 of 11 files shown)


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -575,6 +584,22 @@ class ClientProxy: ClientProxyProtocol {
...
+              stateDefault: nil,
+              ban: nil,
+              kick: nil,
+              redact: nil,
+              invite: nil,
+              notifications: nil,
+              users: [:],
+              events: [
+                  "m.call.member": Int32(0),
+                  "org.matrix.msc3401.call.member": Int32(0)
```
toger5, line 600: This is really good, that we have both here (the unstable prefix + the final one)

toger5 (approved):
Looks reasonable. It does make sense and is easy to follow even though I am not familiar with the codebase.

Have you tested this (There are no rust sdk unit tests for this part since we dont have them for the FFI and most of this is FFI code)

Just one sanity test should be enough (see if a new room is having the correct entry in power_levels, there are not many cases but this would guarantee that there is no insane typo somewhere...)

ElementX/Sources/Services/Room/RoomProxy.swift
```diff
@@ -396,6 +396,15 @@ class RoomProxy: RoomProxyProtocol {
         }
     }
     
+    func canUserJoinCall(userID: String) async -> Result<Bool, RoomProxyError> {
+        do {
+            return try await .success(room.canUserSendState(userId: userID, stateEvent: .roomMemberEvent))
```
pixlwave, line 401: Can we add a comment about why I need to be able to send a room member event in order to join a call? It feels very counter-intuitive just looking at it.

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -575,6 +584,22 @@ class ClientProxy: ClientProxyProtocol {
             return (nil, nil)
         }
     }
+    
+    private static var defaultRoomCreationPowerLevels: PowerLevels {
```
pixlwave, line 588: I kind of expected to find this static var at the top of the file. Not a biggy, just unexpected.

pixlwave: Also, a second thought for a follow-up PR: could this exist in the SDK? So we could simply pass `powerLevelContentOverride: .callOverride` or similar to the creation parameters?

pixlwave (approved):
Nice 👌

ElementX/Sources/Services/Room/RoomProxy.swift
toger5: Sorry for missing this in my review...

PR #2191, opened 2023-11-30
Fix scrambled screen when searchinfg

Velin92:
fixes the issue on iOS 17.1.1 of the scrambled screen while keeping the navigation bar not hidden on previous versions. However this won't allow the bar to display on 17.1.1 yet.

### DIFF
--- a/ElementX/Sources/Screens/MessageForwardingScreen/View/MessageForwardingScreen.swift
+++ b/ElementX/Sources/Screens/MessageForwardingScreen/View/MessageForwardingScreen.swift
@@ -28,13 +28,15 @@ struct MessageForwardingScreen: View {
                 }
                 // Replace these with ScrollView's `scrollPosition` when dropping iOS 16.
             } header: {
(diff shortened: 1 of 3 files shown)


ElementX/Sources/Screens/MessageForwardingScreen/View/MessageForwardingScreen.swift
```diff
@@ -59,6 +61,12 @@ struct MessageForwardingScreen: View {
         .compoundSearchField()
         .disableAutocorrection(true)
     }
+    
+    // The greedy size of Rectangle can create an issue with the navigation bar when the search is highlighted, so is best to use a fixed frame instead of hidden() or EmptyView()
```
pixlwave, line 65: Maybe?
```suggestion
    /// The greedy size of Rectangle can create an issue with the navigation bar when the search is highlighted, so is best to use a fixed frame instead of hidden() or EmptyView()
```

pixlwave (approved):

PR #2202, opened 2023-12-04
Add an extra call to obscure the app on willEnterForeground.

pixlwave:
It shouldn't be necessary given the double call by the time the app is backgrounded, but we're still seeing occasional room list flashes. So this PR adds a new `launching` state to the state machine that is used from between `willEnterForeground` and `didBecomeActive`.

### DIFF
--- a/ElementX/Sources/FlowCoordinators/AppLockFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/AppLockFlowCoordinator.swift
@@ -44,6 +44,8 @@ class AppLockFlowCoordinator: CoordinatorProtocol {
         case appObscured
         /// The app is in the background.
         case backgrounded
+        /// The app is returning to the foreground.
+        case launching
         /// The app is presenting biometric unlock to the user.
         case attemptingBiometricUnlock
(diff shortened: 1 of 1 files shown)


ElementX/Sources/FlowCoordinators/AppLockFlowCoordinator.swift
```diff
@@ -188,8 +200,10 @@ class AppLockFlowCoordinator: CoordinatorProtocol {
             case (_, .backgrounded):
                 appLockService.applicationDidEnterBackground()
                 showPlaceholder() // Double call but just to be safe. Useful at app launch.
+            case (_, .launching):
+                showPlaceholder() // Triple call but necessary after being suspended.
             case (_, .attemptingBiometricUnlock):
-                showPlaceholder() // For the unlock background. Triple call but just to be safe.
+                showPlaceholder() // For the unlock background. Quadruple call but just to be safe.
```
stefanceriu, line 206: When in doubt 😅

stefanceriu (approved):
Looks sensible to me, let's see what happens.

PR #2203, opened 2023-12-05
Expand all sections on the view source screen by default.

pixlwave:
When you wanted to look at the JSON it was annoying to have to scroll down, expand the section and then scroll again to see it.


stefanceriu (approved):
Ship it! 😉

nimau (approved):
LGTM

PR #2207, opened 2023-12-06
Add notifications toggle invitations.

### LINKED ISSUE #2205: Notification setting for invites
Add a setting with a toggle in the notifications setting screen to enable / disable the notification for invites, in the Additional settings section which has to be unhidden (currently hidden, since it only contains call setting which is hidden - the call settings stay hidden), with the wording Invitations, and enabled by default (note that we read the value from the SDK, and it should be true by default, that's the default push rules).

Using https://github.com/matrix-org/matrix-rust-sdk/pull/2908. Android reference https://github.com/vector-im/element-x-android/pull/1955

pixlwave:
Fixes #2205.

notifications@2x


stefanceriu (approved):
lgtm 👍

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/View/NotificationSettingsScreen.swift
```diff
@@ -242,7 +260,9 @@ struct NotificationSettingsScreen_Previews: PreviewProvider, TestablePreview {
 
     static var previews: some View {
         NotificationSettingsScreen(context: viewModel.context)
+            .snapshot(delay: 0.1)
         NotificationSettingsScreen(context: viewModelConfigurationMismatch.context)
+            .snapshot(delay: 0.1)
```
pixlwave, line 265: The snapshots were still in a loading state so added this. Weirdly 2 of the rows are still loading, but even at 2.5 seconds they didn't update so left it like this.

PR #2209, opened 2023-12-06
Replace searchable modifiers with a custom implementation using UISearchController.

pixlwave:
From iOS 17.1 the search configuration was unreliable (see #2189, #2191). Instead of introspecting the search field, this PR adds a custom implementation that covers what `searchable` plus `searchableConfiguration`, `disableInteractiveDismissOnSearch` and `dismissSearchOnDisappear` provided. This is used in the start chat flow and message forwarding screen with the same configuration as originally used for iOS 16 (showing the nav bar and hiding the cancel button).

### DIFF
--- a/ElementX/Sources/Other/SwiftUI/Search.swift
+++ b/ElementX/Sources/Other/SwiftUI/Search.swift
@@ -30,18 +30,25 @@ extension View {
         modifier(DismissSearchOnDisappear())
     }
     
-    /// Configures a searchable's underlying search controller.
-    /// - Parameters:
-    ///   - hidesNavigationBar: A Boolean indicating whether to hide the navigation bar when searching.
-    ///   - showsCancelButton: A Boolean indicating whether the search controller manages the visibility of the search bar’s cancel button.
+    /// A custom replacement for searchable that allows more precise configuration of the underlying search controller.
     ///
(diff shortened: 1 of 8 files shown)


ElementX/Sources/Other/SwiftUI/Search.swift
```diff
@@ -18,53 +18,151 @@ import SwiftUI
...
+            .interactiveDismissDisabled(!searchQuery.isEmpty && disablesInteractiveDismiss)
+            .background {
+                SearchController(searchQuery: $searchQuery,
+                                 placeholder: placeholder,
+                                 hidesNavigationBarDuringPresentation: hidesNavigationBarDuringPresentation,
+                                 automaticallyShowsCancelButton: automaticallyShowsCancelButton,
+                                 isSearching: $isSearching)
+            }
             .onDisappear {
+                // Dismiss search when the view disappears to tidy up appearance when popping back to the view.
```
pixlwave, line 68: I didn't see the point of adding a parameter for this, it seems like behaviour we would always want.

nimau (approved):
LGTM 👍

pixlwave:
Something weird has happened in the PreviewTests, the search field's font is slightly larger than it should be. I checked the UI tests and the app itself, and it hasn't changed so must be another one of those weird snapshotting bugs.

PR #2212, opened 2023-12-07
More secure backup tweaks

stefanceriu:
* fix indicator not going away when disabling backups
* simplify rust to swift key backup state mapping
* add more logs


pixlwave (approved):
LGTM 👍

PR #2222, opened 2023-12-11
Fix Spacing

Velin92:
This will probably end up breaking support for bidi text (text that contains both LTR and RTL at the same time) with the timestamp, but will make line spacing consistent


pixlwave (approved):
LGTM, seems a reasonable change until we figure something else out 👍

UnitTests/__Snapshots__/PreviewTests/test_timelineItemBubbledStylerView.Mock-Timeline-RTL.png
Velin92, line 1: @pixlwave however I am not very convinced about this... this will break RTL display when it contains LTR messages

PR #2234, opened 2023-12-12
Removing Dead Code part 1

Velin92:


ElementX/Sources/Other/BlurHashDecode.swift
pixlwave, line 1: Maybe we should ignore both this and `Encode` given they're essentially external dependencies.

ElementX/Sources/Other/Extensions/ConfirmationDialog.swift
pixlwave, line 1: I dunno, this and `Alert.swift` seem to both remove potentially useful convenience funcs.

ElementX/Sources/Other/SwiftUI/Layout/CustomLayoutLabelStyle.swift
pixlwave, line 1: I think this is another nice utility that would be nice to leave for convenience. Maybe add a preview of a trailing icon layout to a) keep it around and b) make sure it's validated.

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
```diff
@@ -23,13 +23,10 @@ typealias CallScreenViewModelType = StateStoreViewModel<CallScreenViewState, Cal
 
 class CallScreenViewModel: CallScreenViewModelType, CallScreenViewModelProtocol {
     private let roomProxy: RoomProxyProtocol
-    private let callBaseURL: URL
-    private let clientID: String
     
     private let widgetDriver: ElementCallWidgetDriverProtocol
     
     private let callController = CXCallController()
-    private let callProvider = CXProvider(configuration: .init())
```
pixlwave, line 32: Ok, no, now this one, it's some weird Apple magic going on here. We need it irrespective of how unused it appears to be.

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
```diff
@@ -130,18 +125,6 @@ class CallScreenViewModel: CallScreenViewModelType, CallScreenViewModelProtocol
...
-    func callObserver(_ callObserver: CXCallObserver, callChanged call: CXCall) {
-        MXLog.info("Call changed: \(call)")
-    }
-    
-    // MARK: - CXProviderDelegate
-    
-    func providerDidReset(_ provider: CXProvider) {
-        MXLog.info("Call provider did reset: \(provider)")
-    }
-
```
pixlwave, line 144: And the same with these. Don't ask me how they work, but without them background calls don't last forever.

ElementX/Sources/Services/Analytics/AnalyticsClientProtocol.swift
```diff
@@ -31,24 +31,13 @@ protocol AnalyticsClientProtocol {
     /// Stop the analytics client reporting data.
     func stop()
     
-    /// Send any queued events immediately.
-    func flush()
```
pixlwave, line 35: Ok, right yeah this one is useful for debugging. But tbf no reason why you can't manually add it back if we're ever doing stuff to PostHog again.

ElementX/Sources/Services/Audio/Player/AudioPlayerState.swift
```diff
@@ -167,7 +167,7 @@ class AudioPlayerState: ObservableObject, Identifiable {
         displayLink?.add(to: .current, forMode: .common)
     }
     
-    @objc private func updateProgress(displayLink: CADisplayLink) {
+    @objc private func updateProgress() {
```
pixlwave, line 170: Stupid question, but is the @objc dynamic access still happy without the parameter?

ElementX/Sources/Services/Audio/Recorder/AudioRecorderState.swift
```diff
@@ -98,7 +98,7 @@ class AudioRecorderState: ObservableObject, Identifiable {
         displayLink?.add(to: .current, forMode: .common)
     }
     
-    @objc private func publishUpdate(displayLink: CADisplayLink) {
+    @objc private func publishUpdate() {
```
pixlwave, line 101: Same question here.

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -413,18 +412,6 @@ class ClientProxy: ClientProxyProtocol {
...
-            .failure(.failedRetrievingAccountData)
-        }
-    }
-    
-    func setAccountData<Content: Encodable>(content: Content, type: String) async -> Result<Void, ClientProxyError> {
-        await Task.dispatch(on: clientQueue) {
-            .failure(.failedSettingAccountData)
-        }
-    }
-
```
pixlwave, line 427: Hmmm interesting. I'm mixed on removing methods from a proxy. It will make it harder for someone to find out how to get account data in the future if they need it. But then these are also not well typed, so passing strings around isn't great anyway. Yeah I've talked me self into it, lets get rid of them

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
```diff
@@ -164,19 +147,6 @@ class CallScreenViewModel: CallScreenViewModelType, CallScreenViewModelProtocol
...
-        }
-        
-        do {
-            return try await evaluator(script) as? String
-        } catch {
-            MXLog.error("Failed evaluating javaScript with error: \(error)")
-            return nil
-        }
-    }
-
```
pixlwave, line 179: And I really can't remember about this one, but I got mega confused in review and Stefan convinced me it was needed. Maybe we can confirm this one with him?

pixlwave (changes requested):
Overall looks good to me, but we need to be very careful with the CallKit magic.

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
Velin92: sure, as of right now this function is private and is not used in the code @stefanceriu do we need it?

ElementX/Sources/Services/Audio/Player/AudioPlayerState.swift
Velin92: how is this function being used? is it used externally in objc in some way?

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
Velin92: but these delegates where are they even implemented? I don't see them being enforced anywhere

ElementX/Sources/Other/BlurHashDecode.swift
Velin92: I mean we probably don't need to use all of it, since this subscript in particular is not used (while the one it is).

ElementX/Sources/Other/Extensions/ConfirmationDialog.swift
Velin92: true actually the confirmaiton dialog and that one were quite nice, I guess I can put an ignore on top of them

ElementX/Sources/Other/SwiftUI/Layout/CustomLayoutLabelStyle.swift
Velin92: it only needed to be used as a switch statement

ElementX/Sources/Services/Audio/Player/AudioPlayerState.swift
pixlwave: Discussed offline and found this [requirement](https://developer.apple.com/documentation/quartzcore/cadisplaylink/1621228-init)

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
pixlwave: Maaaagic!

pixlwave (approved):
Cool, LGTM 👍

PR #2237, opened 2023-12-13
Update codecov links.

pixlwave:
This PR updates Codecov links which are currently broken to the new organisation. I'm not seeing coverage reports working properly yet so there may be something more for us to do. Let's see what this PR does as I think Codecov only updated the organisation name overnight so it might have fixed itself. 👀


pixlwave:
Codecov is indeed happy once again 🎉

Screenshot 2023-12-13 at 10 10 21 am

nimau (approved):
👍

PR #2244, opened 2023-12-13
Polls history

nimau:
This PR adds a new polls history screen, accessible from the room details screen. (closes https://github.com/element-hq/element-x-ios/issues/2230)

It doesn't use the new `pollHistoryTimeline` available on `RoomProxy` because once a poll is edited, the `timeline` and the `pollHistoryTimeline` seem to be out of sync.

Notes:

- The `PollView` has been extracted from `PollRoomTimelineView` in order to be reused for the history.
- The polls interaction handler functions have been extracted to `PollInteractionHandler` in order to be reused for the history.


ElementX/Sources/Screens/RoomPollsHistoryScreen/RoomPollsHistoryScreenCoordinator.swift
```diff
@@ -0,0 +1,63 @@
...
+    let pollInteractionHandler: PollInteractionHandlerProtocol
+    let roomTimelineController: RoomTimelineControllerProtocol
+}
+
+enum RoomPollsHistoryScreenCoordinatorAction {
+    case editPoll(pollStartID: String, poll: Poll)
+}
+
+final class RoomPollsHistoryScreenCoordinator: CoordinatorProtocol {
+    private let parameters: RoomPollsHistoryScreenCoordinatorParameters
```
Velin92, line 31: we can remove this, we don't need to store them

ElementX/Sources/Screens/RoomPollsHistoryScreen/RoomPollsHistoryScreenCoordinator.swift
```diff
@@ -0,0 +1,63 @@
...
+    private var viewModel: RoomPollsHistoryScreenViewModelProtocol
+    private let actionsSubject: PassthroughSubject<RoomPollsHistoryScreenCoordinatorAction, Never> = .init()
+    private var cancellables = Set<AnyCancellable>()
+    
+    var actions: AnyPublisher<RoomPollsHistoryScreenCoordinatorAction, Never> {
+        actionsSubject.eraseToAnyPublisher()
+    }
+    
+    init(parameters: RoomPollsHistoryScreenCoordinatorParameters) {
+        self.parameters = parameters
```
Velin92, line 41: also here

ElementX/Sources/Screens/RoomPollsHistoryScreen/RoomPollsHistoryScreenViewModel.swift
```diff
@@ -0,0 +1,195 @@
...
+    
+    private func updateBackPaginatingState(_ value: Bool) {
+        state.isBackPaginating = value
+        if value {
+            userIndicatorController.submitIndicator(.init(id: isPaginatingIndicatorID, type: .modal(progress: .indeterminate, interactiveDismissDisabled: true, allowsInteraction: false), title: L10n.commonLoading))
+        } else {
+            userIndicatorController.retractIndicatorWithId(isPaginatingIndicatorID)
+        }
+    }
+}
```
Velin92, line 186: maybe instead of a function, would be better to use have a sink on the viewstate that checks for the value of isBackPaginating and removes duplicates to control if this should be displayed or not.

Velin92 (approved):

ElementX/Sources/UITests/UITestsAppCoordinator.swift
```diff
@@ -877,6 +877,38 @@ class MockScreen: Identifiable {
             let coordinator = PollFormScreenCoordinator(parameters: .init(mode: .new))
             navigationStackCoordinator.setRootCoordinator(coordinator)
             return navigationStackCoordinator
+        case .roomPollsHistoryEmptyLoadMore:
+            let navigationStackCoordinator = NavigationStackCoordinator()
+            let roomProxy = RoomProxyMock(with: .init(displayName: "Some room name", avatarURL: nil))
```
Velin92, line 882: I think this is not used

ElementX/Sources/UITests/UITestsAppCoordinator.swift
```diff
@@ -877,6 +877,38 @@ class MockScreen: Identifiable {
...
+            ]
+            let parameters = RoomPollsHistoryScreenCoordinatorParameters(roomProxy: RoomProxyMock(),
+                                                                         pollInteractionHandler: interactionHandler,
+                                                                         roomTimelineController: roomTimelineController)
+            let coordinator = RoomPollsHistoryScreenCoordinator(parameters: parameters)
+            navigationStackCoordinator.setRootCoordinator(coordinator)
+            return navigationStackCoordinator
+        case .roomPollsHistoryLoadMore:
+            let navigationStackCoordinator = NavigationStackCoordinator()
+            let roomProxy = RoomProxyMock(with: .init(displayName: "Some room name", avatarURL: nil))
```
Velin92, line 897: same this one

nimau:
@stefanceriu can you tell me if the last commit works for you? It solves an issue where the back pagination state is not seen as `.timelineStartReached` when a second timelineController is built using the same timeline (which is the case for `PollHistoryScreenViewModel`).

stefanceriu:
> @stefanceriu can you tell me if the last commit works for you?

yep yep, makes sense to me 👍

nimau:
@amshakal here are some screenshots, does it work for you?

New section in room details


History with 1 poll not editable, and 1 editable:





Empty history with back pagination available:





Not editable poll:

amshakal (approved):
These look good to me, thank you!!

PR #2246, opened 2023-12-14
Send html in plain text mode

Velin92:
Requires 2.22.0 version of RTE


ElementX/Sources/Screens/ComposerToolbar/ComposerToolbarViewModel.swift
```diff
@@ -107,9 +111,8 @@ final class ComposerToolbarViewModel: ComposerToolbarViewModelType, ComposerTool
             case .previewVoiceMessage:
                 actionsSubject.send(.voiceMessage(.send))
             default:
-                let sendHTML = ServiceLocator.shared.settings.richTextEditorEnabled
                 actionsSubject.send(.sendMessage(plain: wysiwygViewModel.content.markdown,
-                                                 html: sendHTML ? wysiwygViewModel.content.html : nil,
+                                                 html: wysiwygViewModel.content.html,
```
pixlwave, line 115: Should we go through and make this `html` non-optional everywhere and then we can remove the call to the SDK too?

pixlwave (approved):
👏

PR #2247, opened 2023-12-14
Remove NCE as a target

Velin92:
temporarily removing NCE since we are not really using as of right now


pixlwave (approved):
LGTM 👍

Do you think it would it be better to remove everything (inc the files) rather than commenting out the target? We'll have a nice commit that's easy to revert when we want it back.

nimau (approved):

PR #2249, opened 2023-12-14
Process InfoPlist.strings in Localazy config.

### LINKED ISSUE #2100: Setup support for multiple files from Localazy
On iOS we need to localise the Info.plist file. Localazy supports this, but we need to make sure our download tasks are set up correctly so that each platform downloads all the files it needs as right now they assume only a single file iirc.

pixlwave:
Also merge the 2 configs into one using a group to download all strings.

Closes #2100

| It works 🎉 |
| - |
| Screenshot 2023-12-14 at 6 01 08 pm |


Velin92 (approved):
LGTM

PR #2251, opened 2023-12-15
Fix poll form UI tests

pixlwave:
This is slower but is more reliable so I hope this is the tests done. Annoying that they work before merging and then fail once merged 🙈


stefanceriu:
Wrote a different fix for it [here] which seems to run on the CI just fine

PR #2260, opened 2023-12-18
Fixes #2259 - Fix flipped notification placeholder avatars on iOS 17.2+

### LINKED ISSUE #2259: Placeholder avatars have flipped again on iOS 17.2
### Steps to reproduce

1. Receive push from a room without an explicit avatar on EX 465 on iOS 17.2
2. Observe that some of the avatars are vertically flipped:

image

3. Also, observe they are horribly low resolution (much less than retina) and have strange aliasing artefacts at the compass points.

#### What did you expect?

Room avatars which are the right way up

#### What happened instead?

The upside down

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

465

### Homeserver

_No response_

### Will you send logs?

No

stefanceriu:


ElementX/Sources/Other/Extensions/UNNotificationContent.swift
```diff
@@ -235,12 +236,26 @@ extension UNMutableNotificationContent {
...
+        if ProcessInfo.processInfo.isiOSAppOnMac {
+            return false
+        }
+        
         guard let version = Version(UIDevice.current.systemVersion) else {
             return false
         }
         
-        return version.major >= 17
+        if version >= Version(17, 0, 0), version < Version(17, 2, 0) {
```
Velin92, line 253: I am not sure but maybe and I say maybe this could actually also happen on iOS 17.1 but sadly this needs to be checked on a physical device

stefanceriu: As I understand it **don't** need to flip the icon for iOS 17.0, 17.0.1, 17.0.2, 17.0.3, 17.1, 17.1.1 and 17.1.2 which this check should handle, right?

ElementX/Sources/Other/Extensions/UNNotificationContent.swift
```diff
@@ -189,10 +189,11 @@ extension UNMutableNotificationContent {
         return updatedContent.mutableCopy() as! UNMutableNotificationContent
     }
 
+    @MainActor
     private func getPlaceholderAvatarImageData(name: String, id: String) async -> Data? {
         // The version value is used in case the design of the placeholder is updated to force a replacement
-        let isIOS17Available = isIOS17Available()
-        let prefix = "notification_placeholder\(isIOS17Available ? "V3" : "V2")"
+        let shouldFlipAvatar = shouldFlipAvatar()
+        let prefix = "notification_placeholder\(shouldFlipAvatar ? "V3" : "V2")"
```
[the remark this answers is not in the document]
stefanceriu: I initially figured that's unnecessary but you're right, between now and when people update they might have a bunch of wrongly cached avatars 👍

ElementX/Sources/Other/Extensions/UNNotificationContent.swift
Velin92: But I mean... it's okay in the end eventually people WILL update to iOS 17.2, so if they are on 17.1 and they have the bug, when they update everything will be fixed

Velin92 (approved):

PR #2264, opened 2023-12-18
Removing dead code part 3

Velin92:
Removed all the dead code that is not related to assets and strings, next step is to find a way to detect only the non used strings and the non used assets from the generated code, without triggering false positives generated by some helper functions of the generated code, but this is for another PR


ElementX/Sources/UITests/UITestsNotificationCenter.swift
```diff
@@ -25,6 +25,7 @@ import SwiftUI
 /// - Start the app from the tests and call `client.waitForApp()` to establish communication.
 /// - Send the notification from the tests you would like posted in the app.
 class UITestsNotificationCenter: NotificationCenter {
+    // periphery:ignore - retaining purpose
```
Velin92, line 28: not sure about this one... @stefanceriu  i think even if we assign but we don't use it it is required for some retaining instance purpose right?

ElementX/Sources/Other/ShareToMapsAppActivity.swift
```diff
@@ -43,10 +43,6 @@ final class ShareToMapsAppActivity: UIActivity {
         type.activityTitle
     }
 
-    var activityCategory: UIActivity.Category {
-        .action
-    }
-
```
stefanceriu, line 49: I wonder if this one is just missing an `override`. On the other hand the `default is UIActivityCategoryAction` so that's why it worked to begin with 😁

ElementX/Sources/Screens/ComposerToolbar/View/VoiceMessagePreviewComposer.swift
```diff
@@ -95,20 +94,6 @@ struct VoiceMessagePreviewComposer: View {
             onPlay()
         }
     }
-    
-    private func onScrubbing(_ scrubbing: Bool) {
```
stefanceriu, line 99: Took me a while to untangle all of these 😅

ElementX/Sources/Screens/Settings/SettingsScreen/SettingsScreenCoordinator.swift
```diff
@@ -41,6 +41,7 @@ final class SettingsScreenCoordinator: CoordinatorProtocol {
     private let parameters: SettingsScreenCoordinatorParameters
     private var viewModel: SettingsScreenViewModelProtocol
     
+    // periphery:ignore - retaining purpose
     private var appLockSetupFlowCoordinator: AppLockSetupFlowCoordinator?
```
stefanceriu, line 45: Are these basically just false positives? They're clearly used, even within the same object in this case 🤔

ElementX/Sources/Screens/Settings/UserDetailsEditScreen/UserDetailsEditScreenCoordinator.swift
```diff
@@ -24,18 +24,11 @@ struct UserDetailsEditScreenCoordinatorParameters {
...
 final class UserDetailsEditScreenCoordinator: CoordinatorProtocol {
     private let parameters: UserDetailsEditScreenCoordinatorParameters
     private var viewModel: UserDetailsEditScreenViewModelProtocol
-    private let actionsSubject: PassthroughSubject<UserDetailsEditScreenCoordinatorAction, Never> = .init()
     private var cancellables = Set<AnyCancellable>()
     
-    var actions: AnyPublisher<UserDetailsEditScreenCoordinatorAction, Never> {
-        actionsSubject.eraseToAnyPublisher()
-    }
-
```
stefanceriu, line 38: I'm loving all the deletions but can't help to wonder if we're making things harder on ourselves in the long run. Is re-writing all of this boiler plate worth it if we ever need to add actions to these Screens/VMs/Coordinators?

ElementX/Sources/Services/SecureBackup/SecureBackupControllerProtocol.swift
```diff
@@ -64,11 +64,3 @@ protocol SecureBackupControllerProtocol {
...
     func waitForKeyBackupUpload() async -> Result<Void, SecureBackupControllerError>
 }
-
-extension SecureBackupControllerMock {
-    convenience init(keyBackupState: SecureBackupKeyBackupState, recoveryKeyState: SecureBackupRecoveryKeyState) {
-        self.init()
-        underlyingKeyBackupState = CurrentValueSubject<SecureBackupKeyBackupState, Never>(keyBackupState).asCurrentValuePublisher()
-        underlyingRecoveryKeyState = CurrentValueSubject<SecureBackupRecoveryKeyState, Never>(recoveryKeyState).asCurrentValuePublisher()
-    }
-}
```
stefanceriu, line 74: 👌

ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift
```diff
@@ -259,11 +254,10 @@ class RoomTimelineController: RoomTimelineControllerProtocol {
             
             // Try building a stable identifier for items that don't have one
             // We need to avoid duplicates otherwise the diffable datasource will crash
```
stefanceriu, line 256: Should we get rid of this comment as well?

ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift
```diff
@@ -27,11 +27,6 @@ class RoomTimelineController: RoomTimelineControllerProtocol {
     private let serialDispatchQueue: DispatchQueue
     
     private var cancellables = Set<AnyCancellable>()
-    private var timelineItemsUpdateTask: Task<Void, Never>? {
```
stefanceriu, line 30: Some of these things I don't even remember adding 🙈

ElementX/Sources/UITests/UITestsNotificationCenter.swift
stefanceriu: Yep yep, that seems to be the case 👍

Tools/Scripts/Templates/SimpleScreenExample/ElementX/TemplateScreenCoordinator.swift
```diff
@@ -14,6 +14,8 @@
 // limitations under the License.
 //
 
+// periphery:ignore:all - this is just a template remove this comment once generating the final file
+
```
stefanceriu, line 18: This feels dangerous 😅

UnitTests/Sources/AnalyticsTests.swift
```diff
@@ -134,10 +134,9 @@ class AnalyticsTests: XCTestCase {
         XCTAssertNil(client.pendingUserProperties, "No user properties should have been set yet.")
         
         // When updating the user properties
-        client.updateUserProperties(AnalyticsEvent.UserProperties(ftueUseCaseSelection: .PersonalMessaging,
+        client.updateUserProperties(AnalyticsEvent.UserProperties(allChatsActiveFilter: nil, ftueUseCaseSelection: .PersonalMessaging,
```
stefanceriu, line 137: Let's move these guys to the next line please

stefanceriu (approved):
Love it! 👏

ElementX/Sources/Screens/Settings/SettingsScreen/SettingsScreenCoordinator.swift
Velin92: **they are assigned but never used**, periphery is even capable of telling this to us. Ofc we need to be aware that sometimes we do this just for retaining purpose, luckily then comment ignores the flag for the next run and provides context.

ElementX/Sources/Screens/Settings/UserDetailsEditScreen/UserDetailsEditScreenCoordinator.swift
Velin92: we still have the template as an example if we ever want to reconstruct that boilerplate code (I added a full ignore in the template screen)

Tools/Scripts/Templates/SimpleScreenExample/ElementX/TemplateScreenCoordinator.swift
Velin92: the ignore all ?
Is just a way to not flag the whole template file as dead code.

stefanceriu: Yeah, okay, fair enough 👍

PR #2268, opened 2023-12-19
Fix a crash that can occur when updating timeline items in a background thread

nimau:
This PR attempts to fix a crash when a timeline item ID is not found:
`Fatal error > TimelineItem TimelineItemIdentifier(....) not found`

The simplest reason for this is if the `RoomTimelineController.timelineItems` array is updated during use.

This PR also ensures that `updateTimelineItems()` is always executed using the same serialQueue (which was not the case with the `contentSizeCategoryChanged` notification).


stefanceriu (approved):
🚀

PR #2281, opened 2023-12-21
MentionBadge FF

Velin92:
Requires a new SDK release with the notificationCount fix


ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -238,14 +238,14 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
         }
         
         let notificationMode = roomInfo.userDefinedNotificationMode.flatMap { RoomNotificationModeProxy.from(roomNotificationMode: $0) }
-        
+        MXLog.info("roomInfo: \(roomInfo)")
```
stefanceriu, line 241: Let's get rid of this please, too much info, especially for very large accounts

stefanceriu (approved):
lgtm but it's still a question whether it's the right way to go

Velin92:
> lgtm but it's still a question whether it's the right way to go

let's wait, for me and Benoit this doesn't work on this SDK version sadly so maybe not...

stefanceriu (approved):
lgtm 👍

PR #2307, opened 2024-01-08
Fixes #2306 - Reporting content screen not being displayed

### LINKED ISSUE #2306: Reporting content doesn't present report screen
https://github.com/element-hq/element-x-ios-rageshakes/issues/1292

Reporting content from the timeline doesn't present the report content screen. Subsequent transitions fails because the state is the same. 

This is a regression introduced with the `RoomScreenInteractionHandler` and it's basically a typo.

stefanceriu:


Velin92:
what is the actual change of the PR? so far seems to be only a rename of the stack coordinator?

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -646,8 +646,8 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
             }
             .store(in: &cancellables)
         
-        navigationStackCoordinator.setRootCoordinator(coordinator)
-        navigationStackCoordinator.setSheetCoordinator(navigationStackCoordinator) { [weak self] in
+        stackCoordinator.setRootCoordinator(coordinator)
+        navigationStackCoordinator.setSheetCoordinator(stackCoordinator) { [weak self] in
```
stefanceriu, line 650: @Velin92 this bit here ^ is key, where `navigationStackCoordinator` was used to present itself

Velin92 (approved):

PR #2309, opened 2024-01-08
UI test fixes

stefanceriu:
Various tweaks for getting the UI tests working again


UITests/Sources/PollFormScreenUITests.swift
```diff
@@ -56,9 +59,7 @@ class PollFormScreenUITests: XCTestCase {
             addOption.tap()
         }
         
-        if app.hasFocus {
-            app.typeText("\n") // Dismiss the keyboard so the Add button is always visible.
-        }
+        app.swipeDown() // Dismiss the keyboard so the Add button is always visible.
```
pixlwave, line 62: At one stage this test had a double `app.swipeUp()` (one for the keyboard and one to get to the bottom). If that works, its probably better just to be 100% sure we're really right at the bottom.

ElementX/Sources/Screens/CreatePollScreen/View/PollFormScreen.swift
```diff
@@ -74,11 +74,18 @@ struct PollFormScreen: View {
...
+                        context.options[index].text.removeLast()
+                        maybeFocusNextElementWith(currentOptionIndex: index)
+                    }
                     .onSubmit {
-                        let nextOptionIndex = index == context.options.endIndex - 1 ? nil : index + 1
-                        focus = nextOptionIndex.map { .option(index: $0) }
+                        maybeFocusNextElementWith(currentOptionIndex: index)
                     }
-                    .submitLabel(index == context.options.endIndex - 1 ? .done : .next)
+                    .submitLabel(index == context.options.endIndex - 1 ? .done : .return)
```
pixlwave, line 88: I don't think this change is necessary with your code above

UITests/Sources/PollFormScreenUITests.swift
stefanceriu: Seems this doesn't work on an iPhone, the latest run of the tests failed because of it. I'll raise another PR to fix it

stefanceriu: https://github.com/element-hq/element-x-ios/pull/2312

PR #2315, opened 2024-01-10
Fixes #2314 - Cache room and member avatars to disk

### LINKED ISSUE #2314: Cache avatars
### Steps to reproduce

1. Launch app
2. Scroll room list
3. Observe that avatars incrementally load and haven’t been cached

#### What did you expect?

Maintain a cache of room & user avatars

#### What happened instead?

Looks like they get pulled in from scratch from the server every time, wasting network and server resources and making the app feel sluggish/amnesiac.

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

484

### Homeserver

_No response_

### Will you send logs?

No

stefanceriu:


ElementX/Sources/Services/UserSession/UserSessionStore.swift
```diff
@@ -91,11 +91,8 @@ class UserSessionStore: UserSessionStoreProtocol {
     // MARK: - Private
     
     private func buildUserSessionWithClient(_ clientProxy: ClientProxyProtocol) -> UserSessionProtocol {
-        let imageCache = ImageCache.onlyInMemory
-        imageCache.memoryStorage.config.keepWhenEnteringBackground = true
```
pixlwave, line 95: Is this set to true on the default one? Otherwise we'll need to re-load timeline content right?

pixlwave (approved):
The code looks fine to me, but personally I'm sure this is the right direction to close the issue. I would have thought:
- Either the SDK should be caching these things to stop the network requests

or

- If the slowness isn't actual network requests but is just them loading them from the SDK store, then double caching is going to result in un-necessary disk usage of the app, and maybe we should be figuring out how to speed up access to the store.

stefanceriu:
> The code looks fine to me, but personally I'm not sure this is the right direction to close the issue. I would have thought:
> 
> * Either the SDK should be caching these things to stop the network requests
> 
> or
> 
> * If the slowness isn't actual network requests but is just them loading them from the SDK store, then double caching is going to result in un-necessary disk usage of the app, and maybe we should be figuring out how to speed up access to the store.

Actually as it turns out neither of these are the problem. No network requests run, everything is properly cached and fast to retrieve. It just feels slow because they start loading on appear and have an animation. I'll raise a separate PR that handles the animation part

PR #2317, opened 2024-01-10
Various integration test tweaks following test homeserver and account…

stefanceriu:
… change

Switches the integration tests to matrix.org and the big tests account and tweaks certain steps to make them more reliable.

Test run here https://github.com/element-hq/element-x-ios/actions/runs/7474683118


pixlwave (approved):
Nice, they're finally running again! 👏👏👏

PR #2324, opened 2024-01-11
updateMembers now is chain of membersNoSync + membersWithSync

### LINKED ISSUE #2304: Read receipt avatars don't load reliably
### Steps to reproduce

Much of the time when you open a room, the RRs are shown with placeholder initials rather than the user's avatars.

See also https://github.com/element-hq/element-x-android/issues/2176

#### What did you expect?

Reliable avatars on RRs

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

482

### Homeserver

_No response_

### Will you send logs?

No

Velin92:
fixes #2304 

This will make the loading of members be faster if the members have already been stored locally.
This PR also updates the SDK and renames some variables and functions related to an API name change in the SDK

### DIFF
--- a/ElementX/Sources/Services/Room/RoomProxy.swift
+++ b/ElementX/Sources/Services/Room/RoomProxy.swift
@@ -175,23 +175,31 @@ class RoomProxy: RoomProxyProtocol {
             }
         }
     }
-
+    
     func updateMembers() async {
         // We always update members first using the no sync API in case internet is not readily available
         // To get the members stored on disk first, this API call is very fast.
-        if let membersNoSyncIterator = try? await room.membersNoSync(),
-           let members = membersNoSyncIterator.nextChunk(chunkSize: membersNoSyncIterator.len()) {
-            membersSubject.value = members.map {
-                RoomMemberProxy(member: $0, backgroundTaskService: self.backgroundTaskService)
+        do {
+            let membersNoSyncIterator = try await room.membersNoSync()
+            if let members = membersNoSyncIterator.nextChunk(chunkSize: membersNoSyncIterator.len()) {
+                membersSubject.value = members.map {
+                    RoomMemberProxy(member: $0, backgroundTaskService: self.backgroundTaskService)
+                }
             }
+        } catch {
+            MXLog.error("[RoomProxy] Failed to update members using no sync API: \(error)")
         }
-            
-        // Then we update members using the sync API, this is slower but will get us the latest members
-        if let membersIterator = try? await room.members(),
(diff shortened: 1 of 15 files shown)


ElementX/Sources/Services/Room/RoomProxy.swift
```diff
@@ -65,10 +65,6 @@ class RoomProxy: RoomProxyProtocol {
         pollHistoryTimeline = await TimelineProxy(timeline: room.pollHistory(), backgroundTaskService: backgroundTaskService)
         
         Task {
-            // Force the timeline to load member details so it can populate sender profiles whenever we add a timeline listener
-            // This should become automatic on the RustSDK side at some point
-            await room.timeline().fetchMembers()
-
```
Velin92, line 71: this function was actually not needed anymore and did nothing

ElementX/Sources/Services/Room/RoomProxy.swift
```diff
@@ -181,19 +177,21 @@ class RoomProxy: RoomProxyProtocol {
...
+        // Then we update members using the sync API, this is slower but will get us the latest members
+        if let membersIterator = try? await room.members(),
+           let members = membersIterator.nextChunk(chunkSize: membersIterator.len()) {
+            membersSubject.value = members.map {
                 RoomMemberProxy(member: $0, backgroundTaskService: self.backgroundTaskService)
             }
-            
-            membersSubject.value = roomMembersProxies
-        } catch {
-            return
```
pixlwave, line 196: Maybe might have been nice to keep the do catch and add a log that the request failed, but not super important.

ElementX/Sources/Services/Room/RoomProxy.swift
```diff
@@ -181,19 +177,21 @@ class RoomProxy: RoomProxyProtocol {
     }
 
     func updateMembers() async {
```
pixlwave, line 179: `InviteUsersScreenViewModel` and `RoomMembersListScreenViewModel` both call this method wrapped by a show/hide loading indicator call. I wonder if we need to revisit those (possibly just changing the indicators to be toasts rather than blocking to indicate that the loaded members might change?

pixlwave (approved):
Nice 👍

ElementX/Sources/Services/Room/RoomProxy.swift
Velin92: I think for these two specific views we want to actually display the up to date members so having the UI wait for also the sync call to finish is intended. I would suggest to leave as it is for now.

PR #2327, opened 2024-01-12
Reverting RTE to use SDK markdown when in markdown mode

### LINKED ISSUE #2265: Crash when pasting XML into the RTE
### Steps to reproduce

1. Was running EXI on macOS
2. RTE disabled
3. Paste XML (see rageshake)
4. App crashes

#### What did you expect?

no crash

#### What happened instead?

big badda boom

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

465

### Homeserver

_No response_

### Will you send logs?

Yes

### LINKED ISSUE #2303: Rich text editor fails parsing lists
Original report here https://github.com/element-hq/element-x-ios-rageshakes/issues/1289

```
It happens if you type a message like:
Djdjs
1. Jdjdj
2. Jdjdj

So i think 2 line returns make the trick
```

The composer remains empty and then we fatal error when sending the message

```
ElementX/RoomScreenViewModel.swift:466: Fatal error: This message should never be empty
```

Velin92:
Also added a way to prevent user suggestions to display when RTE is disabled

fixes #2303 
fixes #2265

### DIFF
--- a/ElementX/Sources/Services/Timeline/TimelineProxy.swift
+++ b/ElementX/Sources/Services/Timeline/TimelineProxy.swift
@@ -99,6 +99,8 @@ final class TimelineProxy: TimelineProxyProtocol {
                      html: String?,
                      original eventID: String,
                      intentionalMentions: IntentionalMentions) async -> Result<Void, TimelineProxyError> {
+        MXLog.info("Editing message with original event ID: \(eventID)")
+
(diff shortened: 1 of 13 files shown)


ElementX/Sources/Services/Timeline/TimelineProxy.swift
pixlwave, line 1: Maybe useful to keep those logs?

pixlwave (approved):
LGTM 👍

PR #2328, opened 2024-01-12
Enable database encryption for new logins on Nightly/PR builds.

### LINKED ISSUE #441: Encrypt local databases
We have to replace `None` in the [code](https://github.com/matrix-org/matrix-rust-sdk/blob/badacdcaa668988b6b08d29177913db14b32f796/bindings/matrix-sdk-ffi/src/client_builder.rs#L77) with a secret.

We should manage this secret like on EI where we have a protocol and different implementations to manage secrets.

pixlwave:
This PR essentially closes #441, although we're going to test on Nightly first to measure any potential performance impact. The following approach has been taken:
- The `RestorationToken` now contains an optional `passphrase` string which is given to the SDK when restoring a session. Existing users won't have one set, so their accounts will continue to operate without encryption on the db.
- When running a dev build (inc Nightly and PRs) the `AuthenticationServiceProxy` will generate a passphrase in the init (which we do for each login so a fresh key will be generated for each login). This passphrase is handed to the SDK which will create the stores, and when a sucessfull login is made, it is handed to the session store to add to the `RestorationToken`
- OIDC sessions update the `RestorationToken` each time the restoration token is refreshed. We now need to fetch the token to preserve the passphrase and if it is missing the app will crash (which will likely result in the user being signed out on next launch but if this happened there wouldn't be much we could do app-side to fix it anyway).
- I've tweaked the location where the `pusherNotificationClientIdentifier` is generated, as getting it from the `ClientProxy` through the restoration token is no longer possible, due to the proxy not knowing what the passphrase is.

### DIFF
--- a/ElementX/Sources/Services/UserSession/RestorationToken.swift
+++ b/ElementX/Sources/Services/UserSession/RestorationToken.swift
@@ -14,24 +14,13 @@
 // limitations under the License.
 //
 
-import CryptoKit
 import Foundation
-
 import MatrixRustSDK
 
 struct RestorationToken: Codable, Equatable {
     let session: MatrixRustSDK.Session
+    let passphrase: String?
     let pusherNotificationClientIdentifier: String?
-
-    init(session: MatrixRustSDK.Session) {
-        self.session = session
-        if let data = session.userId.data(using: .utf8) {
-            let digest = SHA256.hash(data: data)
-            pusherNotificationClientIdentifier = digest.compactMap { String(format: "%02x", $0) }.joined()
-        } else {
-            pusherNotificationClientIdentifier = nil
-        }
-    }
 }
 
 extension MatrixRustSDK.Session: Codable {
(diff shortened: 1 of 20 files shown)


ElementX/Sources/Services/UserSession/RestorationToken.swift
```diff
@@ -14,24 +14,13 @@
...
-
-    init(session: MatrixRustSDK.Session) {
-        self.session = session
-        if let data = session.userId.data(using: .utf8) {
-            let digest = SHA256.hash(data: data)
-            pusherNotificationClientIdentifier = digest.compactMap { String(format: "%02x", $0) }.joined()
-        } else {
-            pusherNotificationClientIdentifier = nil
-        }
-    }
```
pixlwave, line 34: Another option if we wanted to keep this code here would be to give the NotificationManager access to the keychainController. I'm split on this.

Velin92 (approved):

Velin92 (changes requested):
I don't understand a lot of the new passphrase system works but... if we restore the encrypted db from the NSE... how does the NSE decrypt such db?

pixlwave:
> I don't understand a lot of the new passphrase system works but... if we restore the encrypted db from the NSE... how does the NSE decrypt such db? I think we should decrypt it also in the NSE

Thank you 🙌 - I totally forgot about the NSE 🤦‍♂️

PR #2332, opened 2024-01-12
User ID is shown also in the room members list

Velin92:
This will avoid ambiguity when users share the same display name in the list


pixlwave:
Looks good to me, although I would imagine there are some UI test snapshots to update too.

pixlwave (approved):

PR #2342, opened 2024-01-16
Fixes #2325 - Prevent the same WebView from being reused for multiple…

### LINKED ISSUE #2325: Switching between EC SPA calls doesn’t work
### Steps to reproduce

1. Click on a link for an EC call; it opens in EX
2. Close the call when it’s over
3. Click on a link for another call; it opens in EX… but dials into the previous SPA, not the new one.

#### What did you expect?

SPA calling to take you to the right call.

#### What happened instead?

SPA calling took us to the wrong call.

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

1.5.1

### Homeserver

_No response_

### Will you send logs?

No

stefanceriu:
… Element Call links

### DIFF
--- a/ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
+++ b/ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
@@ -23,6 +23,7 @@ struct CallScreen: View {
     
     var body: some View {
         WebView(url: context.viewState.url, viewModelContext: context)
+            .id(UUID())
             .ignoresSafeArea(edges: .bottom)
             .presentationDragIndicator(.visible)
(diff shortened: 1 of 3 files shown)


ElementX/Sources/Mocks/Generated/GeneratedMocks.swift
```diff
@@ -1,4 +1,4 @@
-// Generated using Sourcery 2.1.3 — https://github.com/krzysztofzablocki/Sourcery
```
Velin92, line 1: `brew upgrade sourcery`

Velin92 (approved):

PR #2351, opened 2024-01-17
Subscription fixes

stefanceriu:
Makes room proxy unsubscriptions explicit. Fixes problems in which creating and deallocating room proxies would unexpectedly unsubscribe other proxies.


ElementX/Sources/Mocks/Generated/GeneratedMocks.swift
```diff
@@ -1,4 +1,4 @@
-// Generated using Sourcery 2.1.3 — https://github.com/krzysztofzablocki/Sourcery
```
Velin92, line 1: time to update sourcery xD

Velin92 (approved):

ElementX/Sources/Mocks/Generated/GeneratedMocks.swift
stefanceriu: My god, I saw that like 10 times and I always read it as my version being newer 🤦‍♂️

PR #2363, opened 2024-01-20
Add support for quick room search through the Cmd+K shortcut

stefanceriu:
This PR leverages the new WindowManager, existing room list filtering and UI to implement a quick room search window that appears on Cmd+k and has full keyboard support for both the Mac and iPad OS (whenever a hardware keyboard is connected).

Goes hand in hand with https://github.com/element-hq/element-x-ios/pull/2343 to make finding, opening and starting typing in a room a breeze.

https://github.com/element-hq/element-x-ios/assets/637564/e4b70745-d07a-4320-be7b-3c9c6094f623


ElementX/Sources/Application/Windowing/WindowManager.swift
```diff
@@ -83,11 +92,16 @@ class WindowManager: WindowManagerProtocol {
...
-            overlayWindow.isHidden = true
             mainWindow.isHidden = true
+            overlayWindow.isHidden = true
+            globalSearchWindow.isHidden = true
         }
     }
     
+    func switchToGlobalSearch() {
+        globalSearchWindow.makeKey()
+    }
```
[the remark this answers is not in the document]
pixlwave: Actually no wait, I see what this is doing now. I think we need to leave `switchToMain` and `switchToAlternate` alone as the "lock/unlock" mechanism. And instead we should have `show`/`hideGlobalSearch` with that guard on both so we only alter visibility when unlocked.

ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
```diff
@@ -78,6 +78,12 @@ struct HomeScreen: View {
             .onChange(of: context.viewState.visibleRooms) { _ in
                 updateVisibleRange()
             }
+            .background(
+                Button("", action: {
+                    context.send(viewAction: .globalSearch)
+                })
+                .keyboardShortcut(KeyEquivalent("k"), modifiers: [.command])
+            )
```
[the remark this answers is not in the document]
pixlwave: Given you also do the same in `GlobalSearchScreen`, I wonder if we should make a custom `keyboardShortcut(_:modifiers:action:)` modifier that wraps this nicely in a reusable way too.

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -333,7 +335,14 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
             if room.isPublic {
                 state.bindings.leaveRoomAlertItem = LeaveRoomAlertItem(roomId: roomId, isDM: room.isEncryptedOneToOneRoom, state: .public)
             } else {
-                state.bindings.leaveRoomAlertItem = room.joinedMembersCount > 1 ? LeaveRoomAlertItem(roomId: roomId, isDM: room.isEncryptedOneToOneRoom, state: .private) : LeaveRoomAlertItem(roomId: roomId, isDM: room.isEncryptedOneToOneRoom, state: .empty)
+                let alertItem: LeaveRoomAlertItem
+                if room.joinedMembersCount > 1 {
+                    alertItem = LeaveRoomAlertItem(roomId: roomId, isDM: room.isEncryptedOneToOneRoom, state: .private)
+                } else {
+                    alertItem = LeaveRoomAlertItem(roomId: roomId, isDM: room.isEncryptedOneToOneRoom, state: .empty)
+                }
```
pixlwave, line 343: 😉
```suggestion
                let alertItem = if room.joinedMembersCount > 1 {
                    LeaveRoomAlertItem(roomId: roomId, isDM: room.isEncryptedOneToOneRoom, state: .private)
                } else {
                    LeaveRoomAlertItem(roomId: roomId, isDM: room.isEncryptedOneToOneRoom, state: .empty)
                }
```

UnitTests/Sources/GlobalSearchScreenViewModelTests.swift
```diff
@@ -0,0 +1,22 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import XCTest
+
+@testable import ElementX
+
+@MainActor
+class GlobalSearchScreenViewModelTests: XCTestCase { }
```
pixlwave, line 22: Probably worth adding a basic test here that changing the search query binding filers some rooms away.

ElementX/Sources/Screens/GlobalSearchScreen/GlobalSearchScreenViewModel.swift
```diff
@@ -0,0 +1,104 @@
...
+    }
+    
+    // MARK: - Public
+    
+    override func process(viewAction: GlobalSearchScreenViewAction) {
+        MXLog.info("View model: received view action: \(viewAction)")
+        
+        switch viewAction {
+        case .dismiss:
+            actionsSubject.send(.dismiss)
```
pixlwave, line 61: This should clear the filter right?

ElementX/Sources/Screens/GlobalSearchScreen/View/GlobalSearchScreen.swift
```diff
@@ -0,0 +1,219 @@
...
+import SwiftUI
+
+struct GlobalSearchScreen: View {
+    @ObservedObject var context: GlobalSearchScreenViewModel.Context
+    
+    @State private var selectedRoom: GlobalSearchRoom?
+    @FocusState private var searchFieldFocus
+    
+    var body: some View {
+        ZStack {
```
pixlwave, line 27: I don't think you need this `ZStack`. Doing `.frame(700).bg(.white).frame(.inifinity).bg(.black.opacity(0.5))` should be enough to get the same effect.

ElementX/Sources/Screens/GlobalSearchScreen/View/GlobalSearchScreen.swift
```diff
@@ -0,0 +1,219 @@
...
+                            }
+                    }
+                }
+            }
+            .listStyle(.plain)
+            .background(.compound.bgCanvasDefault)
+            .frame(maxWidth: 700, maxHeight: 800)
+            .clipShape(RoundedRectangle(cornerRadius: 10))
+        }
+        .listStyle(PlainListStyle())
```
pixlwave, line 54: This one is a dupe

ElementX/Sources/Screens/GlobalSearchScreen/View/GlobalSearchScreen.swift
```diff
@@ -0,0 +1,219 @@
...
+                header
+                
+                Section {
+                    ForEach(context.viewState.rooms) { room in
+                        GlobalSearchScreenListRow(room: room, context: context)
+                            .listRowBackground(backgroundColor(for: room))
+                            .listRowInsets(.init())
+                            .onTapGesture {
+                                context.send(viewAction: .select(roomID: room.id))
+                            }
```
pixlwave, line 38: Should really be a button ideally.

ElementX/Sources/Screens/GlobalSearchScreen/View/GlobalSearchScreen.swift
```diff
@@ -0,0 +1,219 @@
...
+    func updateUIView(_ uiView: UITextField, context: Context) {
+        uiView.text = text
+    }
+
+    func makeCoordinator() -> Coordinator {
+        Coordinator(parent: self)
+    }
+
+    class Coordinator: NSObject, UITextFieldDelegate {
+        var parent: GlobalSearchTextFieldRepresentable
```
pixlwave, line 164: Might be a bit cleaner to pass the handler and binding in rather than the parent?

ElementX/Sources/Screens/GlobalSearchScreen/View/GlobalSearchScreen.swift
```diff
@@ -0,0 +1,219 @@
...
+        fatalError()
+    }
+    
+    init(keyPressHandler: @escaping (UIKeyboardHIDUsage) -> Bool) {
+        self.keyPressHandler = keyPressHandler
+        super.init(frame: .zero)
+    }
+
+    override func pressesBegan(_ presses: Set<UIPress>, with event: UIPressesEvent?) {
+        guard let key = presses.first?.key else { return }
```
pixlwave, line 198: Super obscure edge case, just wondering if this should this check that there's only a single press incase someone is playing keyboard gymnastics in the message composer?

ElementX/Sources/Screens/GlobalSearchScreen/View/GlobalSearchScreenCell.swift
```diff
@@ -0,0 +1,62 @@
...
+import SwiftUI
+
+struct GlobalSearchScreenListRow: View {
+    @Environment(\.dynamicTypeSize) var dynamicTypeSize
+    
+    let room: GlobalSearchRoom
+    let context: GlobalSearchScreenViewModel.Context
+    
+    var body: some View {
+        ZStack { // The list row swallows listRowBackgrounds for some reason
```
pixlwave, line 27: This is cheating beyond belief! 😅

ElementX/Sources/Screens/GlobalSearchScreen/View/GlobalSearchScreen.swift
stefanceriu: Making it a button break highlighting ¯\_(ツ)_/¯

ElementX/Sources/Screens/GlobalSearchScreen/View/GlobalSearchScreenCell.swift
stefanceriu: Whatchagonnado 😁

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -974,7 +974,7 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
     }
     
     private func presentMessageForwarding(for itemID: TimelineItemIdentifier) {
-        guard let roomProxy, let roomSummaryProvider = userSession.clientProxy.messageForwardingRoomSummaryProvider, let eventID = itemID.eventID else {
+        guard let roomProxy, let roomSummaryProvider = userSession.clientProxy.alternateRoomSummaryProvider, let eventID = itemID.eventID else {
```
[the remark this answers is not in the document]
stefanceriu: I added a manual clear for the filters but we're good anyway because the default value of the query gets sank into the view model on creation

ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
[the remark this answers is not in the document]
stefanceriu: I'm not sure how useful that would be. If we don't pass an array of modifiers and actions then we would have to introduce multiple backgrounds and you might not even want that on background but on a button instead ¯\_(ツ)_/¯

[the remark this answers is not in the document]
stefanceriu: I couldn't find it in the accessibility inspector so I think were good

ElementX/Sources/Screens/GlobalSearchScreen/View/GlobalSearchScreen.swift
stefanceriu: These only kick in when the global search is displayed but and I trust that the SwiftUI home screen shortcut does the right. I wouldn't be able to check it anyway.

ElementX/Sources/Application/Windowing/WindowManager.swift
```diff
@@ -50,6 +51,11 @@ class WindowManager: WindowManagerProtocol {
         overlayWindow.backgroundColor = .clear
         overlayWindow.isHidden = false
         
+        globalSearchWindow = PassthroughWindow(windowScene: windowScene)
```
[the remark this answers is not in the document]
stefanceriu: > we should make sure there's a way to dismiss the search

Already took care of that, there's a tap outside of the search box that dismissed the whole thing

ElementX/Sources/Application/Windowing/WindowManager.swift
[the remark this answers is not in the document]
stefanceriu: > Does makeKey do anything if the window is hidden
Well making a window key does not affect its visibility but it does make it receive all non-touch input

UITests/Sources/GlobalSearchScreenUITests.swift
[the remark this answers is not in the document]
stefanceriu: I'm going to have a look at integration tests overall later, I need to fix some bugs there anyway

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -578,4 +582,44 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
...
+        let hostingController = UIHostingController(rootView: coordinator.toPresentable())
+        hostingController.view.backgroundColor = .clear
+        windowManager.globalSearchWindow.rootViewController = hostingController
+
+        windowManager.switchToGlobalSearch()
+    }
+    
+    private func dismissGlobalSearch() {
+        windowManager.globalSearchWindow.rootViewController = nil
+        windowManager.switchToMain()
```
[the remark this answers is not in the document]
stefanceriu: Honestly I don't think it's worth the trouble, flow coordinators should know what they're doing generally

ElementX/Sources/Application/Windowing/WindowManager.swift
```diff
@@ -98,10 +99,20 @@ class WindowManager: WindowManagerProtocol {
...
+        guard alternateWindow.isHidden else {
+            return
+        }
+        
+        globalSearchWindow.isHidden = false
         globalSearchWindow.makeKey()
     }
-
+    
+    func hideGlobalSearch() {
```
pixlwave, line 111: Can this one also guard that the alternateWindow is hidden too?

pixlwave (approved):
Ok LGTM, with that 1 comment above.

PR #2368, opened 2024-01-22
App now prevents redacting own messages if there is no permission

Velin92:
requires:
https://github.com/element-hq/element-x-ios/pull/2367


stefanceriu:
Do you maybe need to change this PR's base? doesn't look quite right 🤔

ElementX/Sources/Screens/RoomScreen/RoomScreenInteractionHandler.swift
```diff
@@ -604,7 +611,7 @@ class RoomScreenInteractionHandler {
     // MARK: - Private
     
     private func canRedactItem(_ item: EventBasedTimelineItemProtocol) -> Bool {
-        item.isOutgoing || (canCurrentUserRedactOthers && !roomProxy.isDirect)
+        item.isOutgoing ? canCurrentUserRedactSelf : canCurrentUserRedactOthers && !roomProxy.isDirect
```
stefanceriu, line 614: I wonder if this `roomProxy.isDirect` should be part of `canCurrentUserRedactOthers`. This is the first time I'm realising this is the behavior and feels quite hidden..

stefanceriu (approved):
lgtm 👍

ElementX/Sources/Screens/RoomScreen/RoomScreenInteractionHandler.swift
Velin92: technically you can redact  other messages in direct rooms, but from a product perspective we don't want such behaviour in el-x

stefanceriu: Sure, but can we maybe make part of the `RoomProxy`'s `canUserRedactOther(userID: String)` implementation?

Velin92: this has been  discussed privately and solved since we want the proxy to be source of the actual data, while doing this product related business in the handler

alexander-potemkin: > technically you can redact other messages in direct rooms, but from a product perspective we don't want such behaviour in el-x

@Velin92 , if you don't mind me asking - why you don't want such behavior in Element X? It exists in Element (Classic) and in Element Web and it doesn't save anyone from getting messages deleted, people get used to do that, they still can - just not in Element X.

So... Why?

stefanceriu: @alexander-potemkin I'm curious why you're so set on this. In a DM there's 2 participants and both tend to know what they wrote. How does redacting your counterpart's messages help anything at any point

alexander-potemkin: @stefanceriu I have no ideas how and why, to be honest, but the people used to have that functionality and seems like they used to use it.

And now it's taken away with no clear reasons. And it's not like it's saving anyone from anything: if one wants - he/she will just do it via desktop/classic app.

On top of that, Telegram, for example, let you do that.

So, overall, it looks like: technically it is possible in Element X; it is possible on all other Elements and many messengers. But all of the sudden it's taken out of Element X.

And I can't explain why. Looks like it was done because "why would anyone need that". Which makes sense to me. But then, I believe, it shall be consistent across all apps and restricted by a server (otherwise, I will just delete it anyway using compatible client).

It also might create a (false) sense that 1:1 messages can't be deleted.

With this in mind, it feels like it makes sense to revert the behavior to what it was earlier and what it is across other clients.

Happy to create a PR, if it might be considered.

alexander-potemkin: @stefanceriu , happy holidays!

I'm getting more and more feedback from the users, that ability to delete messages in 1:1 rooms is something they used to see on all messengers and inability to do so in Element X is quite surprising.

Is there something I can do to get this functionality (kind of back) into Element X?

alexander-potemkin:
@stefanceriu , I'm sorry - I would guess [my previous message](https://github.com/element-hq/element-x-ios/pull/2368/files/c0026b51161e1e031782d2d7a2f3105ac1dc6c8a#r2668358240) might get lost, so I'm posting my arguments to return ability to delete messages in 1:1 here:

People used to have that functionality in Element and all other messengers and seems like they used to use it.
And now it's taken away with no clear reasons. And it's not like it's saving anyone from anything: if one wants - he/she will just do it via desktop/classic app.

On top of that, Telegram, for example, let you do that.
So, overall, it looks like: technically it is possible in Element X; it is possible on all other Elements and many messengers. But all of the sudden it's taken out of Element X.

And I can't explain why. Looks like it was done because "why would anyone need that". Which makes sense to me. But then, I believe, it shall be consistent across all apps and restricted by a server (otherwise, I will just delete it anyway using compatible client).

It also might create a (false) sense that 1:1 messages can't be deleted.

With this in mind, it feels like it makes sense to revert the behavior to what it was earlier and what it is across other clients.

Given your confirmation, I'm happy to create a PR

stefanceriu:
> Given your confirmation, I'm happy to create a PR

While I appreciate your passion on the matter I'm not the one that gets to chose how this works, you'll have to pick it up with the product team on https://github.com/element-hq/element-meta.

alexander-potemkin:
> > Given your confirmation, I'm happy to create a PR
> 
> While I appreciate your passion on the matter I'm not the one that gets to chose how this works, you'll have to pick it up with the product team on https://github.com/element-hq/element-meta.

Thank you. I have some experience raising questions of the kind at element-meta and it's usually something that might take years to be even considered. 

And in this specific case it seems like it was decided between you - @stefanceriu and @Velin92 (screenshot attached) , that's why I'm raising it here.

Please, correct me if I've got it wrong!

Screenshot from 2026-01-28 13-56-42

stefanceriu:
@Velin92 Do you happen to remember anything about this or if it ended up being implemented on the other platforms?

stefanceriu:
Okay, tell you what, seeing as this behavior hasn't made it to any of our other clients I'm perfectly happy to go ahead and revert it. If product ever wants it back we can align cross-platform again.

alexander-potemkin:
> Okay, tell you what, seeing as this behavior hasn't made it to any of our other clients I'm perfectly happy to go ahead and revert it. If product ever wants it back we can align cross-platform again.

It's a relatively private space (as it's a PR), so I guess I can express myself without any hesitation: "Thank you, @stefanceriu - I love you! ☺️"

Or, in a bit more formal way: thank you, much appreciated! 🙏

stefanceriu:
Ha, no worries, thanks for pointing this out to begin with! 🙏

PR #2371, opened 2024-01-23
Fix: Grey dot appears when you have unread messages that does not trigger a notification

Velin92:
Grey dot is back

Simulator Screenshot - iPhone 14 - 2024-01-23 at 13 13 59


UnitTests/__Snapshots__/PreviewTests/test_homeScreenRoomCell.Notifications-State.png
Velin92, line 1: @pixlwave Is possible that I covered some cases in the test that are not possible to come from the SDK, however I hope that the main gist makes sense:
Design wants the mention badge to always come alongside the green dot, otherwise it should not appear.
The green dot -> highlight of the timestamp
Muted -> notifications for mentions are ignored and appear as a simple grey dot.

ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomCell.swift
```diff
@@ -128,34 +128,33 @@ struct HomeScreenRoomCell: View {
                         .foregroundColor(.compound.iconQuaternary)
                 }
                 
-                if room.hasMentions, room.notificationMode != .mute {
+                if room.hasMentions, isHighlighted {
```
stefanceriu, line 131: What's the difference between isHighlighted and hasMentions? Naming is confusing..

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
```diff
@@ -145,10 +145,12 @@ struct HomeScreenRoom: Identifiable, Equatable {
     
     var name = ""
     
-    var hasUnreads = false
+    var hasMessages = false
```
stefanceriu, line 148: Why change it to `hasMesages` if it ends up being `details.unreadMessagesCount > 0` anyway?

pixlwave, line 148: Personally I think hadUnreads was fine, but if we need messages in the name, then should it be `hasUnreadMessages` or `hasNewMessages` or something? Otherwise it would be easy to interpret this var as "is it a new room that hasn't been posted into"

ElementX/Sources/Services/Room/RoomSummary/MockRoomSummaryProvider.swift
```diff
@@ -155,74 +161,54 @@ extension Array where Element == RoomSummary {
...
-        .filled(details: RoomSummaryDetails(id: "11",
-                                            settingsMode: .mute,
-                                            hasUnreadNotifications: true,
-                                            hasUnreadMentions: true)),
-        .filled(details: RoomSummaryDetails(id: "12",
-                                            settingsMode: .mute,
-                                            hasUnreadNotifications: false,
-                                            hasUnreadMentions: true))
-    ]
+    static let mockRoomsWithNotificationsState: [Element] = {
```
stefanceriu, line 174: 👌

ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomCell.swift
Velin92: isHiglighted means that the view should highlight = green dot + timestamp text is green and semibold

This snapshot test is based on some the guidelines Amsha gave some time ago, which I had to expand upon given the new API for notifications which is actually very useful. Ofc this huge snapshot is probably covering some very niche edge cases + some impossible ones that can't come from the SDK, but better be safe than sorry:
https://github.com/element-hq/element-x-ios/pull/2371/files#r1463294900

ElementX/Sources/Services/Room/RoomSummary/MockRoomSummaryProvider.swift
Velin92: Screenshot 2024-01-23 at 14 47 59

pixlwave, line 174: Very nice 👏

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
Velin92: because all the other are called `details.unreadSomethingCount` so to avoid confusion maybe these should refer to that something

ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomCell.swift
```diff
@@ -128,34 +128,33 @@ struct HomeScreenRoomCell: View {
...
-                if hasNewContent {
+                if isHighlighted {
                     Circle()
                         .frame(width: 12, height: 12)
-                        .foregroundColor(isHighlighted ? .compound.iconAccentTertiary : .compound.iconQuaternary)
+                        .foregroundColor(.compound.iconAccentTertiary)
+                } else if hasNewContent {
+                    Circle()
+                        .frame(width: 12, height: 12)
+                        .foregroundColor(.compound.iconQuaternary)
```
pixlwave, line 143: If I'm not mistaking things, wouldn't the old version have worked just the same here? Only asking as the ternary is technically more efficient when it comes to SwiftUI managing the underlying views.

pixlwave (approved):
Very nice

ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomCell.swift
Velin92: No because highlighted depends also on the settings and placeholder state...

PR #2372, opened 2024-01-23
Switch Unit and Integration to iOS 17 and iPhone 15

stefanceriu:


stefanceriu:
Dropping this for now as it's not working properly, neither locally nor on CI

PR #2373, opened 2024-01-23
Stefan/2104

stefanceriu:
Automatically generate what to test notes out of the commits since the last tag


ci_scripts/ci_common.sh
```diff
@@ -56,4 +56,42 @@ setup_github_actions_translations_environment() {
...
+
+        # Xcode Cloud shallow clones the repo, we need to manually fetch the tags
+        git fetch --tags
+        
+        TESTFLIGHT_DIR_PATH=TestFlight
+
+        LATEST_TAG=""
+        if [ "$CI_WORKFLOW" = "Release" ]; then
+            # Use -v to invert grep, searching for non-nightlies
+            LATEST_TAG=$(git tag --sort=-creatordate | grep -v 'nightly' | head -n1)
```
pixlwave, line 75: Shouldn't we be using the release notes here instead of the commits? It'll be much more approachable for normal users.

ci_scripts/ci_common.sh
```diff
@@ -56,4 +56,42 @@ setup_github_actions_translations_environment() {
...
+
+        echo "generate_what_to_test_notes: latest tag is $LATEST_TAG"
+
+        mkdir $TESTFLIGHT_DIR_PATH
+
+        NOTES=$(git log --pretty='- %an: %s' "$LATEST_TAG"..HEAD)
+
+        echo "generate_what_to_test_notes: Generated notes:\n$NOTES"
+
+        echo $NOTES > $TESTFLIGHT_DIR_PATH/WhatToTest.en-US.txt
```
pixlwave, line 93: Maybe worth creating a var for `WhatToTest.en-US.txt` given its used 3 times, and typos.

pixlwave (approved):
Hurray! So good to have this implemented 👏

ci_scripts/ci_common.sh
stefanceriu: I think people that look at test flight release notes (especially for nightlies) need a bit more information than we provide in the release notes. For release builds we'd have both, git diff in testflight, proper release notes in github.

ci_scripts/ci_common.sh
```diff
@@ -60,15 +60,12 @@ setup_github_actions_translations_environment() {
...
     if [[ -d "$CI_APP_STORE_SIGNED_APP_PATH" ]]; then
-        echo "generate_what_to_test_notes: App signed, notes for $CI_WORKFLOW"
-
-        echo "generate_what_to_test_notes: current path is $(pwd)"
+        TESTFLIGHT_DIR_PATH=TestFlight
+        TESTFLIGHT_NOTES_FILE_NAME=WhatToTest.en-US.txt
 
         # Xcode Cloud shallow clones the repo, we need to manually fetch the tags
-        git fetch --tags
+        git fetch --tags --quiet
```
pixlwave, line 67: 🙏

pixlwave (approved):

PR #2377, opened 2024-01-24
Failure when finding a refresh token in a non OIDC login

### LINKED ISSUE #2365: Refresh tokens aren't supported and break badly (without OIDC)
### Steps to reproduce

Based on [doug's msg](https://matrix.to/#/!amwlFMejXQcoCxNYHH:matrix.org/$Z2FO2QLsS7zARvGrYIqAJrcDKvpqludnXiMxL8-N7ks?via=matrix.org&via=beeper.com&via=element.io) it sounds like EX doesn't support refresh tokens [MSC2918](https://github.com/sandhose/matrix-doc/blob/sandhose/msc/refresh-token/proposals/2918-refreshtokens.md) unless the server is talking full OIDC.  So if you enable it in synapse at https://element-hq.github.io/synapse/latest/usage/configuration/user_authentication/refresh_tokens.html then all sorts of things break in mysterious ways.

#### What did you expect?

Break hard, or not at all, if the server has refresh tokens configured.

#### What happened instead?

Break in mysterious ways.

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

_No response_

### Homeserver

_No response_

### Will you send logs?

No

Velin92:
fixes #2365


ElementX/Sources/Services/Authentication/AuthenticationServiceProxy.swift
```diff
@@ -122,6 +122,18 @@ class AuthenticationServiceProxy: AuthenticationServiceProxyProtocol {
...
             
+            let refreshToken = try? await Task.dispatch(on: .global()) {
+                try client.session().refreshToken
+            }
+            
+            if refreshToken != nil {
+                MXLog.warning("Refresh token found for a non oidc session, can't restore session, logging out")
+                _ = try? await Task.dispatch(on: .global()) {
+                    try client.logout()
+                }
```
pixlwave, line 133: Not at all important, but may as well simplify it to this.
```suggestion
                _ = await Task.dispatch(on: .global()) {
                    try? client.logout()
                }
```

ElementX/Sources/Services/Authentication/AuthenticationServiceProxy.swift
```diff
@@ -122,6 +122,18 @@ class AuthenticationServiceProxy: AuthenticationServiceProxyProtocol {
                                                      deviceId: deviceID)
             }
             
+            let refreshToken = try? await Task.dispatch(on: .global()) {
+                try client.session().refreshToken
+            }
+            
+            if refreshToken != nil {
+                MXLog.warning("Refresh token found for a non oidc session, can't restore session, logging out")
```
stefanceriu, line 130: Shouldn't this be an error?

ElementX/Sources/Screens/Authentication/SoftLogoutScreen/SoftLogoutScreenCoordinator.swift
```diff
@@ -186,6 +186,9 @@ final class SoftLogoutScreenCoordinator: CoordinatorProtocol {
         case .oidcError(.userCancellation):
             // No need to show an error, the user cancelled authentication.
             break
+        case .cantUseRefreshToken:
+            // We should display a specific error saying that we do not support this kind of login
```
stefanceriu, line 190: Why not do it now? 😄

ElementX/Sources/Services/Authentication/AuthenticationServiceProxyProtocol.swift
```diff
@@ -34,8 +34,8 @@ enum AuthenticationServiceError: Error {
     case slidingSyncNotAvailable
     case accountDeactivated
     case failedLoggingIn
-    
     case isOnWaitlist
+    case cantUseRefreshToken
```
stefanceriu, line 38: `cantUseRefreshToken` to me implies that there's something wrong with the token. A better name would be something along the lines of `sessionTokenRefreshNotSupported`

stefanceriu:
Also, we should remove the `default` in the LoginScreenCoordinator's `handleError` to make sure we handle all current and future errors accordingly. On the same not I find it a bit weird that `.isOnWaitlist` is treated separately from `handleError`

ElementX/Sources/Screens/Authentication/SoftLogoutScreen/SoftLogoutScreenCoordinator.swift
pixlwave: I'm going to include it when going over all the new errors we want to display with Volker.

ElementX/Sources/Services/Authentication/AuthenticationServiceProxy.swift
Velin92: Since is more of a feature we don't support yet than an actual error, I felt the warning was a better log leve, but that was just my opinion at the moment, if you feel the error would be better I can change it.

stefanceriu: No strong feelings really, up to you.

PR #2382, opened 2024-01-25
UI POC for room list filters

### LINKED ISSUE #2339: [RoomList Filters] Implement the filters horizontal scroll view
- Create a horizzontal scroll view that only appears if the room list setting/FF is turned on below the search bar
- This should include the following filters: "Rooms" "People" "Unreads" "Low Priority" "Favourites" (check design for the ordering and naming)
- These filters can be selected and unselected and should change their color according to the selection state
- An X button should appear when at least one filter is selected, to unselected them all at once.

### LINKED ISSUE #2340: [RoomList Filters] Implement filters display/hiding logic
Certain filters are mutually exclusive with other filters:
- You can't have a "Rooms" and "People" filter turned on at the same time
- You can't have a "Favourite" and a "Low priority" filter turned on at the same time
Instead filters that are not mutually exclusive can be chained together, and then can be both selected at the same time.

If two filters are mutually exclusive, and one of them is selected, the other one should hide from the filters list, and as soon as it is unselected the other excluded filter should appear back.
We need to implement this logic.

### LINKED ISSUE #2341: Add a feature flag to the developer options

Velin92:
This is just a UI preview, the filters won't do anything yet to the room list

fixes #2339 
fixes #2340 
fixes #2341

https://github.com/element-hq/element-x-ios/assets/34335419/dabd2f05-678b-4ec2-ac6f-635470d4e047


ElementX/Resources/Localizations/en.lproj/Localizable.strings
```diff
@@ -120,6 +120,11 @@
 "common_everyone" = "Everyone";
 "common_face_id_ios" = "Face ID";
 "common_file" = "File";
+"common_filter_favourites" = "Favourites";
+"common_filter_low_priority" = "Low Priority";
+"common_filter_people" = "People";
+"common_filter_rooms" = "Rooms";
+"common_filter_unreads" = "Unreads";
```
pixlwave, line 127: These should either be `common_favourites` or `screen_roomlist_filter_favourites`. A string for a specific feature can't really be a common string.

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
```diff
@@ -175,3 +178,88 @@ struct HomeScreenRoom: Identifiable, Equatable {
...
+        case .unreads:
+            return L10n.commonFilterUnreads
+        case .favourites:
+            return L10n.commonFilterFavourites
+        case .lowPriority:
+            return L10n.commonFilterLowPriority
+        }
+    }
+    
+    var complementaryFilter: RoomListFilter? {
```
pixlwave, line 208: 👏

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
```diff
@@ -175,3 +178,88 @@ struct HomeScreenRoom: Identifiable, Equatable {
...
+        rawValue
+    }
+    
+    case people
+    case rooms
+    case unreads
+    case favourites
+    case lowPriority
+    
+    var localisedName: String {
```
pixlwave, line 193: As much as I prefer the UK spelling, we should be using US in code.
```suggestion
    var localizedName: String {
```

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
```diff
@@ -175,3 +178,88 @@ struct HomeScreenRoom: Identifiable, Equatable {
...
+        for filter in enabledFilters {
+            availableFilters.remove(filter)
+            if let complementaryFilter = filter.complementaryFilter {
+                availableFilters.remove(complementaryFilter)
+            }
+        }
+        return availableFilters.sorted(by: { $0.rawValue < $1.rawValue })
+    }
+    
+    var isEmpty: Bool {
```
pixlwave, line 246: From the outside this might be nicer to call if it was inverted and named `isFiltering` or similar.

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
```diff
@@ -175,3 +178,88 @@ struct HomeScreenRoom: Identifiable, Equatable {
...
+        enabledFilters.isEmpty
+    }
+    
+    func setIsEnabled(_ isEnabled: Bool, for filter: RoomListFilter) {
```
pixlwave, line 250: Here we say `isEnabled`, but below we check for `isSelected`. Would be nice to use the same for both. Also might be slight nicer to change this to

```suggestion
    func set(_ filter: RoomListFilter, isEnabled: Bool) {
```

So from the outside it is `filter.set(.rooms, isEnabled: true)`

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -333,7 +337,9 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
             if room.isPublic {
                 state.bindings.leaveRoomAlertItem = LeaveRoomAlertItem(roomId: roomId, isDM: room.isEncryptedOneToOneRoom, state: .public)
             } else {
-                state.bindings.leaveRoomAlertItem = room.joinedMembersCount > 1 ? LeaveRoomAlertItem(roomId: roomId, isDM: room.isEncryptedOneToOneRoom, state: .private) : LeaveRoomAlertItem(roomId: roomId, isDM: room.isEncryptedOneToOneRoom, state: .empty)
+                state.bindings.leaveRoomAlertItem = room.joinedMembersCount > 1 ?
+                    LeaveRoomAlertItem(roomId: roomId, isDM: room.isEncryptedOneToOneRoom, state: .private) :
+                    LeaveRoomAlertItem(roomId: roomId, isDM: room.isEncryptedOneToOneRoom, state: .empty)
```
pixlwave, line 342: Unimportant, but might be nice to use an if expression instead of a multiline ternary. WDYT?
```suggestion
                state.bindings.leaveRoomAlertItem = if room.joinedMembersCount > 1 {
                    LeaveRoomAlertItem(roomId: roomId, isDM: room.isEncryptedOneToOneRoom, state: .private)
                } else {
                    LeaveRoomAlertItem(roomId: roomId, isDM: room.isEncryptedOneToOneRoom, state: .empty)
                }
```

ElementX/Sources/Screens/HomeScreen/View/Filters/FilterView.swift
```diff
@@ -0,0 +1,74 @@
...
+}
+
+struct FilterView_Previews: PreviewProvider, TestablePreview {
+    static var previews: some View {
+        FilterView(filter: .people, state: .init())
+        FilterView(filter: .people, state: .init(enabledFilters: [.people]))
+    }
+}
+
+private struct FilterToggleStyle: ToggleStyle {
```
pixlwave, line 43: If you wanted, this could become a common public `ChipToggleStyle` that's reusable elsewhere. I think this would be a nice candidate to move into Compound eventually. Can also have a static accessor too so that we can `.toggleStyle(.chip)` :)

ElementX/Sources/Screens/HomeScreen/View/Filters/FilterView.swift
```diff
@@ -0,0 +1,74 @@
...
+        let shape = RoundedRectangle(cornerRadius: 20)
+        configuration.label
+            .font(.compound.bodyLG)
+            .foregroundColor(foregroundColor(isOn: configuration.isOn))
+            .padding(.horizontal, 16)
+            .padding(.vertical, 8)
+            .background(shape.fill(backgroundColor(isOn: configuration.isOn)))
+            .overlay(shape
+                .inset(by: 0.5)
+                .stroke(strokeColor(isOn: configuration.isOn)))
```
pixlwave, line 66: Tiny nit: I find using a view builder easier to read than a multiline parameter.
```suggestion
            .overlay {
                shape
                    .inset(by: 0.5)
                    .stroke(strokeColor(isOn: configuration.isOn))
            }
```

ElementX/Sources/Screens/HomeScreen/View/Filters/FilterView.swift
```diff
@@ -0,0 +1,74 @@
...
+            .drawingGroup()
+            .onTapGesture {
```
pixlwave, line 68: Can you add a comment here about why we went with a tap gesture instead of a button please. Otherwise someone annoying (aka me) will come along in the future and try tidy that up 🙈

changelog.d/pr-2382.feature
pixlwave, line 1: Side note: This should probably be a `.wip` otherwise it goes in the release notes as a new feature.

ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
```diff
@@ -191,20 +193,30 @@ struct HomeScreen: View {
...
+            if context.viewState.hasPendingInvitations, !context.isSearchFieldFocused {
+                HomeScreenInvitesButton(title: L10n.actionInvitesList, hasBadge: context.viewState.hasUnreadPendingInvitations) {
+                    context.send(viewAction: .selectInvites)
+                }
+                .accessibilityIdentifier(A11yIdentifiers.homeScreen.invites)
+                .frame(maxWidth: .infinity, alignment: .trailing)
             }
-            .accessibilityIdentifier(A11yIdentifiers.homeScreen.invites)
-            .frame(maxWidth: .infinity, alignment: .trailing)
-            .padding(.vertical, -8.0)
```
pixlwave, line 206: I seem to remember lots of conversations about this buttons tap area. Not sure if this changes it.

Maybe we should add a couple of preview snapshots with the Invites button shown if we don't already?

ElementX/Sources/Screens/HomeScreen/View/Filters/RoomListFiltersView.swift
```diff
@@ -0,0 +1,67 @@
...
+                }
+            }
+            .padding(.leading, state.isEmpty ? 8 : 16)
+            .padding(.vertical, 12)
+        }
+        .scrollIndicators(.hidden)
+        .background(Color.compound.bgCanvasDefault)
+    }
+    
+    private var clearButton: some View {
```
pixlwave, line 48: This one can definitely be a button please for Voiceover :)

ElementX/Sources/Screens/HomeScreen/View/Filters/RoomListFiltersView.swift
```diff
@@ -0,0 +1,67 @@
...
+                }
+                ForEach(state.sortedAvailableFilters) { filter in
+                    FilterView(filter: filter, state: state)
+                }
+            }
+            .padding(.leading, state.isEmpty ? 8 : 16)
+            .padding(.vertical, 12)
+        }
+        .scrollIndicators(.hidden)
+        .background(Color.compound.bgCanvasDefault)
```
pixlwave, line 45: Maybe remove this now we put the background on the entire header?

pixlwave (commented):
Woooo nice 👌

ElementX/Sources/Screens/HomeScreen/View/Filters/FilterView.swift
```diff
@@ -0,0 +1,74 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import SwiftUI
+
+struct FilterView: View {
```
stefanceriu, line 19: With this within the HomeScren folder should we name it something a tad less generic?

ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
Velin92: Adding snapshots with the invites button is actually quite complicated, because is not easy to mock this. It's actually quite a problem that I think we should address in the future.

Btw i think I can just remove th vertical 8 padding that we have internally, and this will keep is size consistent as it was before.

PR #2384, opened 2024-01-25
Settings to enable and disable sending public RRs

### LINKED ISSUE #2319: Advanced Settings: Enable/Disable Sending Read Receipts
We would like to add new advanced setting options that allows users to turn off and on public read receipts.
When they are on we just send the normal `read` type of read receipts, when the setting is off, we instead resort to sending the `privateRead` type of read receipts, so that rooms can still be seen as read but other users will not be able to see our read receipts.

The copies of the setting should be:
title: "Send read receipts"
description: "On disabling, others won't see your read receipt."

Velin92:
fixes #2319


stefanceriu (approved):
lgtm 👍

PR #2389, opened 2024-01-26
Move the account migration screen to within the room list

stefanceriu:
This PR removes the previous standalone modal migration screen and instead rebuilds it within the room list, giving the users access the rest of the application (especially logs) while that is happening

mac_large
iphone_max_landscape
iphone_portrait
ipad_landscape


ElementX/Sources/Screens/HomeScreen/View/HomeScreenContent.swift
```diff
@@ -0,0 +1,189 @@
...
+            
+            Text(L10n.screenMigrationMessage)
+                .minimumScaleFactor(0.01)
+                .font(.compound.bodyLG)
+                .multilineTextAlignment(.center)
+                .foregroundColor(.compound.textPrimary)
+                .accessibilityIdentifier(A11yIdentifiers.migrationScreen.message)
+        }
+        .padding(.horizontal)
+    }
```
pixlwave, line 161: Could extract this bit out into a `HomeScreenMigrationView.swift` file to keep this one clean, although not super important.

pixlwave (approved):
Very nice 👏

stefanceriu:
Unit tests are broken but they're broken on develop so I'll just try to fix them separately.

PR #2412, opened 2024-02-01
Add a notification setting to hide unread messages badges.

pixlwave:
Adds a setting that allows users to focus more by hiding grey unread dots on rooms that are set to only notify for mentions & keywords or not at all. Current strings are a placeholder, this PR isn't ready to merge until they've been added to Localazy.

https://github.com/element-hq/element-x-ios/assets/6060466/2f030ff2-ee05-4e70-901d-929c58805653

### DIFF
--- a/ElementX/Resources/Localizations/en.lproj/Untranslated.strings
+++ b/ElementX/Resources/Localizations/en.lproj/Untranslated.strings
@@ -4,10 +4,6 @@
 /* Used for testing */
 "untranslated" = "Untranslated";
 
-// MARK: - Notification Settings
-"screen_notification_settings_hide_unread_badges" = "Sanity mode";
-"screen_notification_settings_hide_unread_badges_description" = "Hide unread badges for rooms set to mute or mentions and keywords.";
-
 // MARK: - Soft logout
 
 "soft_logout_signin_title" = "Sign in";
--- a/ElementX/Sources/Generated/Strings+Untranslated.swift
+++ b/ElementX/Sources/Generated/Strings+Untranslated.swift
@@ -10,10 +10,6 @@ import Foundation
 // swiftlint:disable explicit_type_interface function_parameter_count identifier_name line_length
 // swiftlint:disable nesting type_body_length type_name vertical_whitespace_opening_braces
 public enum UntranslatedL10n {
-  /// Sanity mode
-  public static var screenNotificationSettingsHideUnreadBadges: String { return UntranslatedL10n.tr("Untranslated", "screen_notification_settings_hide_unread_badges") }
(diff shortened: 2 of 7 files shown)


ElementX/Resources/Localizations/en.lproj/Untranslated.strings
```diff
@@ -4,6 +4,10 @@
 /* Used for testing */
 "untranslated" = "Untranslated";
 
+// MARK: - Notification Settings
+"screen_notification_settings_hide_unread_badges" = "Sanity mode";
+"screen_notification_settings_hide_unread_badges_description" = "Hide unread badges for rooms set to mute or mentions and keywords.";
```
pixlwave, line 9: Reminder to self: Waiting on real strings before this PR can be merged.

amshakal:
🥇 Sanity mode :'D

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -327,7 +331,7 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
         return HomeScreenRoom(id: identifier,
                               roomId: details.id,
                               name: details.name,
-                              hasUnreadMessages: details.unreadMessagesCount > 0,
+                              hasUnreadMessages: appSettings.hideUnreadMessagesBadge ? false : details.unreadMessagesCount > 0,
                               hasUnreadMentions: details.unreadMentionsCount > 0,
```
Velin92, line 335: I think you should handle this in the view and not here because the grey dot might also appear for a muted room, that has a mention and no unreads

Velin92: Or maybe is worth refactoring this HomeScreenRoom model so that it includes directly the logic that is used in the view to decide wether to display the dots, and in which color.

Velin92 (changes requested):
Yeah would probably refactor HomeScreenRoom at this point to include the view logic directly

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
pixlwave: Makes sense to me, updated 👍

Velin92 (approved):

ElementX/Resources/Localizations/en.lproj/Untranslated.strings
pixlwave: The setting has been moved into Developer Options, and these strings have been removed.

PR #2419, opened 2024-02-05
Update compound and remove all of the temporary icons

pixlwave:
This PR updates the Compound package following the latest release of our Design Tokens. It removes the temporary icons added on the run up to the October release, using the ones from Compound instead. Probably easiest to review this one commit by commit.

Note: Snapshot tests show a layout change for the `HeroImage` component, but in reality this is a testing issue, the UI tests keep the same layout (none of them failed with the change of foreground colour though so I haven't regenerated those).


stefanceriu (approved):
Whoa this is awesome!! Thank you! 👏👏👏

PR #2420, opened 2024-02-05
Fixes #2416 - Allow text selection on the timeline item menu header

### LINKED ISSUE #2416: Can’t copy mxid or go to profile from context menu
### Steps to reproduce

1. See spam to moderate
2. Long tap on a msg to get its sender (as multiple messages get their sender suppressed and you don’t want to keep scrolling)
3. See the mxid of the sender
4. Can’t do anything with it
5. Work around via View Source

#### What did you expect?

At the least, tapping on the sender name in the context menu should take you to their member view

#### What happened instead?

painful moderation

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

513

### Homeserver

_No response_

### Will you send logs?

No

stefanceriu:


pixlwave (approved):
LGTM, although it doesn't seem to me that it would fix issue if the sender has a displayname set.

stefanceriu:
> LGTM, although it doesn't seem to me that it would fix issue if the sender has a displayname set.

Yeah, it won't, but I guess spammers won't have a display name ¯\_(ツ)_/¯

PR #2423, opened 2024-02-06
WIP RoomList Filter implementation

Velin92:
Don't review yet. WIP


ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift
```diff
@@ -92,8 +92,9 @@ class RoomTimelineController: RoomTimelineControllerProtocol {
     }
     
     func sendReadReceipt(for itemID: TimelineItemIdentifier) async -> Result<Void, RoomTimelineControllerError> {
-        guard let eventID = itemID.eventID
-        else { return .success(()) }
+        guard let eventID = itemID.eventID else {
+            return .failure(.generic)
```
Velin92, line 96: actually why it was returning .success here? I imagine if the identifier is missing we should return this as an error

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -151,12 +151,16 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
...
     func setFilter(_ filter: RoomSummaryProviderFilter) {
         switch filter {
-        case .none:
+        case .excludeAll:
             _ = listUpdatesSubscriptionResult?.controller.setFilter(kind: .none)
-        case .all:
-            _ = listUpdatesSubscriptionResult?.controller.setFilter(kind: .allNonLeft)
-        case .normalizedMatchRoomName(let query):
-            _ = listUpdatesSubscriptionResult?.controller.setFilter(kind: .normalizedMatchRoomName(pattern: query.lowercased()))
+        case let .include(filterLogic):
```
stefanceriu, line 156:
```suggestion
        case let .include(predicate):
```

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -438,3 +442,19 @@ private class RoomListStateObserver: RoomListLoadingStateListener {
...
     }
 }
+
+private extension RoomListFilter {
+    var rustFilter: RoomListEntriesDynamicFilterKind? {
+        switch self {
+        case .people:
+            return .kind(expectedKind: .directMessages)
+        case .rooms:
+            return .kind(expectedKind: .group)
```
stefanceriu, line 452: @Hywan why `.group` instead of `.room`? I don't think we use that naming anywhere else 🤔

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -438,3 +442,19 @@ private class RoomListStateObserver: RoomListLoadingStateListener {
         onUpdateClosure(state)
     }
 }
+
+private extension RoomListFilter {
+    var rustFilter: RoomListEntriesDynamicFilterKind? {
+        switch self {
+        case .people:
```
stefanceriu, line 449: Why people and not `directMessage`? The way we render it in the UI shouldn't affect the underlying terminology.

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProviderProtocol.swift
```diff
@@ -87,10 +87,28 @@ enum RoomSummary: CustomStringConvertible, Equatable {
     }
 }
 
-enum RoomSummaryProviderFilter {
-    case none
-    case all
-    case normalizedMatchRoomName(String)
+enum RoomSummaryProviderFilter: Equatable {
```
stefanceriu, line 90: 👌

Hywan:
> NOTE: in the video unreads work only for unread message not for unread notifications only yet, this is because we are missing the last bit of the implementation in the SDK

A notification or a mention will always increase the number of unread messages, see [this test](https://github.com/matrix-org/matrix-rust-sdk/blob/452cf0f2698c81bf782d5494a5f561261a933ccc/crates/matrix-sdk-base/src/read_receipts.rs#L762-L831).

Right now, the `unread` filter checks for `num_unread`, which looks correct. Do you expect another behaviour?

UnitTests/Sources/RoomListFiltersStateTests.swift
```diff
@@ -0,0 +1,74 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import XCTest
+
+@testable import ElementX
+
+final class RoomListFiltersStateTests: XCTestCase {
```
stefanceriu, line 21: 👌

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -438,3 +442,19 @@ private class RoomListStateObserver: RoomListLoadingStateListener {
...
+            return .kind(expectedKind: .directMessages)
```
Hywan, line 450: It should now be `.category(expect: .people)`.

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
Hywan, line 452: `.category(expect: .group)`

stefanceriu (commented):
This is looking awesome, nicely done! 👏 
Left a bunch of comments but most of them are about naming and moving things around.

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
stefanceriu: Why are we changing the terminology and not using DMs and rooms?

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
```diff
@@ -230,15 +227,14 @@ enum RoomListFilter: Int, CaseIterable, Identifiable {
         case .unreads:
             return nil
         case .favourites:
-            return .lowPriority
-        case .lowPriority:
-            return .favourites
+            // When we will have Low Priority we may need to return it here
```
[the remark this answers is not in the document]
Velin92: Actually I worte wrong the word `complement` (sorry this is my english getting confused with my italian)
Since `complement` in [set theory](https://en.wikipedia.org/wiki/Complement_(set_theory)) is the defined as the set of the elements not included in the first set, which well reflects the concept of a mtutally exclusive filter.

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
```diff
@@ -204,7 +204,6 @@ enum RoomListFilter: Int, CaseIterable, Identifiable {
     case rooms
     case unreads
     case favourites
-    case lowPriority
```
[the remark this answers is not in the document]
Velin92: I mean yeah, but is more work to keep them around and not using, than just simply adding it in the future

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
```diff
@@ -265,6 +261,11 @@ final class RoomListFiltersState: ObservableObject {
     
     func set(_ filter: RoomListFilter, isEnabled: Bool) {
```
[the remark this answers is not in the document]
Velin92: This API naming was actually suggested by @pixlwave I am okay to change it however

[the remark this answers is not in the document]
pixlwave: Yep don't block on me, if separate enable/disable methods are clearer for everyone, lets do it 👍

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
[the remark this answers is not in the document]
stefanceriu: > which well reflects the concept of a mtutally exclusive filter

Well it doesn't really, a set theory complement would give you **all** the filters outside of your current set.

Please go with `incompatible` so nobody else gets confused by this.

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -313,9 +313,9 @@ class ClientProxy: ClientProxyProtocol {
         var (roomListItem, room) = await roomTupleForIdentifier(identifier)
         
         if let roomListItem, let room {
-            return await RoomProxy(roomListItem: roomListItem,
-                                   room: room,
-                                   backgroundTaskService: backgroundTaskService)
+            return try await RoomProxy(roomListItem: roomListItem,
```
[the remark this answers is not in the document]
Velin92: oh good catch, this is probably something that came from the merge conflict since I had prepare this branch before the sdk bump, will remove it

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
Velin92: because directMessage doesn't imply `members == 2`

Velin92:
> > NOTE: in the video unreads work only for unread message not for unread notifications only yet, this is because we are missing the last bit of the implementation in the SDK
> 
> A notification or a mention will always increase the number of unread messages, see [this test](https://github.com/matrix-org/matrix-rust-sdk/blob/452cf0f2698c81bf782d5494a5f561261a933ccc/crates/matrix-sdk-base/src/read_receipts.rs#L762-L831).
> 
> Right now, the `unread` filter checks for `num_unread`, which looks correct. Do you expect another behaviour?

We want the unreads filter to essentially be displayed only when there is a green badge. So the filtering must comply with the push rules of the room.

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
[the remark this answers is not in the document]
Velin92: No.
it would give you all the elements that do not belong to the set.
The set of rooms has as its own complement as the set of people.
The set of people has its own complement as the set of rooms.
So people is the `complementFilter `of rooms because by activating both filter you would end up with an empty set.
But the fact that we are discussing this, means that this is not very clear, so let's just change it to incompatible.

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
```diff
@@ -230,15 +227,14 @@ enum RoomListFilter: Int, CaseIterable, Identifiable {
...
-            return .lowPriority
-        case .lowPriority:
-            return .favourites
+            // When we will have Low Priority we may need to return it here
+            return nil
         }
     }
 }
 
 final class RoomListFiltersState: ObservableObject {
```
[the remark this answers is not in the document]
Velin92: We actually use the unsorted enabledFilters, but yeah can remove it from the available I guess

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
```diff
@@ -230,15 +227,14 @@ enum RoomListFilter: Int, CaseIterable, Identifiable {
...
-    @Published private var enabledFilters: Set<RoomListFilter>
+    @Published private(set) var enabledFilters: Set<RoomListFilter>
     
     init(enabledFilters: Set<RoomListFilter> = []) {
```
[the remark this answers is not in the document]
Velin92: I thought about this at the start of designing this API, but in the end the only filter that will not be available for now is just 1, and is more easy to just add it in the future, than having the API have a list of the filters that can be displayed or not. If we are going to have way more filters in the future however we can definitely do this, I just don't want to overcomplicate things.

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
stefanceriu: This one doesn't seem to be handled

stefanceriu (approved):
🐑 it!

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
Hywan: Because it's not a DM: it's `is_direct` + 2 members + `is_encrypted` :-(.
I wanted to use a different vocabulary here to make it explicit that it's different from a simple DM.

Hywan: See https://github.com/element-hq/element-x-ios/pull/2423#discussion_r1481495895.

PR #2425, opened 2024-02-06
Fixes #2414 - Move member loading to the room member detail screen, a…

### LINKED ISSUE #2414: Loading profiles can block the app for tens of seconds before loading
### Steps to reproduce

1. Tap on a user’s avatar who’s on a small distant server
2. Watch a “loading…” spinner block the whole app for 20s or so

#### What did you expect?

Don’t block the UI. load a membership page with spinner while it loads that you can back out of.

#### What happened instead?

blocked the UI; you have to force quit the app to cancel it

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

511

### Homeserver

_No response_

### Will you send logs?

Yes

stefanceriu:
…void blocking the whole application

https://github.com/element-hq/element-x-ios/assets/637564/6a3e9a06-c603-48e6-ab5c-7d76d523e0a8


ElementX/Sources/Screens/RoomMemberDetailsScreen/View/RoomMemberDetailsScreen.swift
```diff
@@ -71,32 +87,27 @@ struct RoomMemberDetailsScreen: View {
...
-                    details: .isWaiting(context.viewState.isProcessingIgnoreRequest),
-                    kind: .button {
-                        context.send(viewAction: blockUserButtonAction)
-                    })
-                    .accessibilityIdentifier(blockUserButtonAccessibilityIdentifier)
-                    .disabled(context.viewState.isProcessingIgnoreRequest)
+        if let memberDetails = context.viewState.memberDetails {
+            let title = memberDetails.isIgnored ? L10n.screenRoomMemberDetailsUnblockUser : L10n.screenRoomMemberDetailsBlockUser
+            let action: RoomMemberDetailsScreenViewAction = memberDetails.isIgnored ? .showUnignoreAlert : .showIgnoreAlert
+            let accessibilityIdentifier = memberDetails.isIgnored ? A11yIdentifiers.roomMemberDetailsScreen.unignore : A11yIdentifiers.roomMemberDetailsScreen.ignore
```
Velin92, line 95: maybe these could be computed vars?

UnitTests/Sources/RoomMemberDetailsViewModelTests.swift
```diff
@@ -48,23 +55,26 @@ class RoomMemberDetailsViewModelTests: XCTestCase {
...
         let deferred = deferFulfillment(context.$viewState) { state in
-            state.details.isIgnored
+            state.memberDetails?.isIgnored == true
         }
         
         try await deferred.fulfill()
         
         XCTAssertFalse(context.viewState.isProcessingIgnoreRequest)
-        XCTAssertTrue(context.viewState.details.isIgnored)
+        XCTAssertTrue(context.viewState.memberDetails?.isIgnored ?? false)
```
Velin92, line 77: just a nitpick, but maybe XCTAssertEqual(context.viewState.memberDetails?.isIgnored, false) feels nicer to read to me

UnitTests/Sources/RoomMemberDetailsViewModelTests.swift
```diff
@@ -91,7 +105,7 @@ class RoomMemberDetailsViewModelTests: XCTestCase {
         try await deferred.fulfill()
         
         XCTAssertNotNil(context.alertInfo)
-        XCTAssertFalse(context.viewState.details.isIgnored)
+        XCTAssertFalse(context.viewState.memberDetails?.isIgnored ?? false)
```
Velin92, line 108: as above

UnitTests/Sources/RoomMemberDetailsViewModelTests.swift
```diff
@@ -102,38 +116,44 @@ class RoomMemberDetailsViewModelTests: XCTestCase {
...
         
         let deferred = deferFulfillment(context.$viewState) { state in
-            state.details.isIgnored == false
+            state.memberDetails?.isIgnored == false
         }
         
         try await deferred.fulfill()
         
-        XCTAssertFalse(context.viewState.details.isIgnored)
+        XCTAssertFalse(context.viewState.memberDetails?.isIgnored ?? false)
```
Velin92, line 139: as above

ElementX/Sources/Screens/RoomMemberDetailsScreen/View/RoomMemberDetailsScreen.swift
stefanceriu: They were initially but with the memberDetails now being an optional that code became too complicated, so I just moved them in here instead.

UnitTests/Sources/RoomMemberDetailsViewModelTests.swift
stefanceriu: Good point, I guess I'm just going to guard those member details to make it cleaner

PR #2427, opened 2024-02-06
Enable trace logging for the Olm Account

poljar:
This should help us to track down UTD sources, it ensures that we log the state of a Olm session when we try to decrypt an `m.olm.*` to-device message.

This should be applied after we update to https://github.com/matrix-org/matrix-rust-sdk/pull/3100.

### DIFF
--- a/ElementX/Sources/Other/Logging/RustTracing.swift
+++ b/ElementX/Sources/Other/Logging/RustTracing.swift
@@ -74,6 +74,7 @@ struct TracingConfiguration {
         case hyper, matrix_sdk_ffi, matrix_sdk_crypto
         
         case matrix_sdk_client = "matrix_sdk::client"
+        case matrix_sdk_crypto_account = "matrix_sdk_crypto::olm::account"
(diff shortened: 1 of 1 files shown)


Velin92 (approved):

ElementX/Sources/Other/Logging/RustTracing.swift
```diff
@@ -89,6 +90,7 @@ struct TracingConfiguration {
         .matrix_sdk_ffi: .info,
         .matrix_sdk_client: .trace,
         .matrix_sdk_crypto: .debug,
+        .matrix_sdk_crypto_account: .trace
```
Velin92, line 93: there is a missing `,` here

poljar: Trying to install `swift-format` on Linux :smiling_face_with_tear:

poljar: Well that ain't happening, at least not today... Added the `,` and force pushed.

PR #2431, opened 2024-02-07
Infer joined and active member count from the member list, default t…

stefanceriu:
…he members to all mock members

### DIFF
--- a/ElementX/Sources/Mocks/RoomMemberProxyMock.swift
+++ b/ElementX/Sources/Mocks/RoomMemberProxyMock.swift
@@ -42,6 +42,15 @@ extension RoomMemberProxyMock {
     }
 
     // Mocks
+    static var mockMe: RoomMemberProxyMock {
+        RoomMemberProxyMock(with: .init(userID: "@me:matrix.org",
+                                        displayName: "Me",
+                                        avatarURL: URL.picturesDirectory,
+                                        membership: .join,
+                                        isAccountOwner: true,
+                                        canInviteUsers: true))
+    }
+    
     static var mockAlice: RoomMemberProxyMock {
         RoomMemberProxyMock(with: .init(userID: "@alice:matrix.org",
                                         displayName: "Alice",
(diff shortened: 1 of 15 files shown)


ElementX/Sources/Mocks/RoomMemberProxyMock.swift
```diff
@@ -103,3 +111,15 @@ extension RoomMemberProxyMock {
                                         canSendStateEvent: { allowedStateEvents.contains($0) }))
     }
 }
+
+extension Array where Element == RoomMemberProxyMock {
+    static let allMembers: [RoomMemberProxyMock] = [
```
pixlwave, line 116: Nice 👌

UnitTests/__Snapshots__/PreviewTests/test_roomDetailsScreen.DM-Room.png
pixlwave, line 1: It's not an issue with this PR, but is this test actually configured correctly? Looks like it's showing room details and not the alternative DM details configuration 🤔

ElementX/Sources/Mocks/RoomProxyMock.swift
```diff
@@ -66,15 +63,13 @@ extension RoomProxyMock {
...
+        members = CurrentValueSubject(configuration.members).asCurrentValuePublisher()
 
-        if let configuredMembers = configuration.members {
-            members = CurrentValueSubject(configuredMembers).asCurrentValuePublisher()
-        } else {
-            members = CurrentValueSubject([]).asCurrentValuePublisher()
-        }
+        let joinedMembers = configuration.members.filter { $0.membership == .join }
+        joinedMembersCount = joinedMembers.count
+        activeMembersCount = joinedMembers.count
```
pixlwave, line 72: To be accurate shouldn't this filter to include `.join` and `.invited`? (I think that's the definition of active right).

pixlwave (approved):
Nice tidy-up here 👍

ElementX/Sources/Mocks/RoomProxyMock.swift
stefanceriu: Oh, I didn't know that 😅 I thought it's members that have been active recently

PR #2432, opened 2024-02-08
Fix room list cell highlighting issues, add unit tests

stefanceriu:
Fixes the problem reported by Benoit [here](https://github.com/element-hq/element-x-ios/pull/2408#pullrequestreview-1864576124)
* highlights badges when the room is marked as unread but muted
* adds unit tests to prevent us from breaking this rather complex logic

### DIFF
--- a/ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
+++ b/ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
@@ -197,7 +197,7 @@ extension HomeScreenRoom {
         let isMentionShown = details.hasUnreadMentions && !details.isMuted
         let isMuteShown = details.isMuted
         let isCallShown = details.hasOngoingCall
-        let isHighlighted = !details.isMuted && (details.hasUnreadNotifications || details.hasUnreadMentions || details.isMarkedUnread)
+        let isHighlighted = details.isMarkedUnread || (!details.isMuted && (details.hasUnreadNotifications || details.hasUnreadMentions))
         
         self.init(id: identifier,
                   roomId: details.id,
--- a/ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomCell.swift
+++ b/ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomCell.swift
@@ -119,26 +119,23 @@ struct HomeScreenRoomCell: View {
             HStack(spacing: 8) {
                 if room.badges.isCallShown {
                     CompoundIcon(\.videoCallSolid, size: .xSmall, relativeTo: .compound.bodySM)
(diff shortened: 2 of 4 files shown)


ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
```diff
@@ -197,7 +197,7 @@ extension HomeScreenRoom {
         let isMentionShown = details.hasUnreadMentions && !details.isMuted
         let isMuteShown = details.isMuted
         let isCallShown = details.hasOngoingCall
-        let isHighlighted = !details.isMuted && (details.hasUnreadNotifications || details.hasUnreadMentions || details.isMarkedUnread)
+        let isHighlighted = details.isMarkedUnread || (!details.isMuted && (details.hasUnreadNotifications || details.hasUnreadMentions))
```
Velin92, line 200: we want it to appear as green EVEN if the room is muted?

ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomCell.swift
```diff
@@ -119,26 +119,23 @@ struct HomeScreenRoomCell: View {
...
-                        .foregroundColor(.compound.iconAccentTertiary)
                 }
                 
                 if room.badges.isDotShown {
                     Circle()
                         .frame(width: 12, height: 12)
-                        .foregroundColor(room.isHighlighted ? .compound.iconAccentTertiary : .compound.iconQuaternary)
                 }
             }
+            .foregroundColor(room.isHighlighted ? .compound.iconAccentTertiary : .compound.iconQuaternary)
```
stefanceriu, line 138: It's been decided that everything will be green when the room is highlighted, either normally or through being marked as unread.

Velin92: Wait.
I think in the call has been the decided that what matters is only the green dot, not the mute icon.
If the room is marked as unread but is muted, it's okay if there is a green dot, the mute icon can stay grey.

stefanceriu: Didn't we agree to go with the simplest solution and not have any special cases, make everything green if it's highlighted and everything gray if not?

Velin92: let's get one final written confirmation from @amshakal

amshakal: Yes, lets stick to green. Test it out, if it feels weird, we can make learn and make changes accordingly. Since this is a new feature, it's hard to say what will look strange and what wouldnt until we try it. :)

Velin92 (approved):
LGTM

PR #2438, opened 2024-02-08
Use new decorative colours.

pixlwave:
Adopts new decorative colours from Compound which gives us consistent avatar colours compared to develop.element.io. Requires https://github.com/element-hq/compound-ios/pull/59


stefanceriu (approved):
🚀

PR #2442, opened 2024-02-09
Sending RRs logic will be managed by the SDK

Velin92:


ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -284,30 +284,13 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
         }
     }
     
-    /// The ID of the newest item in the room that the user has seen.
-    /// This includes both event based items and virtual items.
-    private var lastReadItemID: TimelineItemIdentifier?
     private func sendReadReceiptIfNeeded(for lastVisibleItemID: TimelineItemIdentifier) async -> Result<Void, RoomTimelineControllerError> {
-        guard lastReadItemID != lastVisibleItemID,
```
Velin92, line 291: @stefanceriu 
We used to check also if the current state of the application was active, now we don't do that anymore... is that okay?

pixlwave: Ooof no, we should keep that. Otherwise on macOS we keep sending RRs when you "quit" the app.

pixlwave: Wait, don't we do that at a different level?

stefanceriu: Oh I don't know, the problem there was if you were running the mac app on a room everything would get read even if the app wasn't active 🤔

pixlwave (approved):
LGTM 👍

PR #2443, opened 2024-02-09
Sending all RR when the app is active client side.

Velin92:
The SDK will check if that's the case or not.


bnjbvr (commented):
Thanks for fixing this :heart: 

(sorry for the drive-by, but it seems that `eventBasedItem` is dead code now and could be removed?)

PR #2447, opened 2024-02-09
Removed the 50ms delay from updating the filters.

Velin92:


pixlwave (approved):
Feels weird to me that this time was added when not needed, but if it doesn't break anything then I'm happy 👍

Issue #2355, opened 2024-01-19
[Task] Show sections for roles in room-member list


pixlwave:
Please see and discuss the details in the meta issue.

PR #2448, opened 2024-02-09
Sort members in the member list by power level and show mods and admins.

pixlwave:
First part of #2355, this PR makes 2 changes:
- Sort the members list by Power Level (and alphabetically as a secondary sort)
- Shows a label when the user's role is mod or admin.

Screenshot 2024-02-09 at 4 59 47 pm


ElementX/Sources/Services/Room/RoomMember/RoomMemberProxyProtocol.swift
```diff
@@ -30,6 +30,8 @@ protocol RoomMemberProxyProtocol: AnyObject {
     var membership: MembershipState { get }
     var isAccountOwner: Bool { get }
     var isIgnored: Bool { get }
+    var powerLevel: Int { get }
+    var role: RoomMemberRole { get }
```
pixlwave, line 34: I was undecided if this should be the Rust type, or mapped to our own type. We seem to have other simple enums from Rust at this stage so I followed that.

ElementX/Sources/Screens/RoomMemberListScreen/RoomMembersListScreenViewModel.swift
```diff
@@ -141,3 +142,22 @@ private struct RoomMembersDetails {
...
+    func sorted() -> Self {
+        sorted { lhs, rhs in
+            if lhs.powerLevel != rhs.powerLevel {
+                lhs.powerLevel > rhs.powerLevel
+            } else {
+                lhs.sortingName < rhs.sortingName
+            }
+        }
+    }
+}
```
Velin92, line 156: I would actually add this into the RoomMemberProxyProtocol extension directly.

ElementX/Sources/Screens/RoomMemberListScreen/RoomMembersListScreenViewModel.swift
```diff
@@ -141,3 +142,22 @@ private struct RoomMembersDetails {
...
+
+private extension RoomMemberProxyProtocol {
+    var sortingName: String {
+        // If there isn't a displayname we sort by the userID without the @.
+        displayName ?? String(userID.dropFirst())
+    }
+}
```
Velin92, line 163: same

Velin92 (approved):
LGTM

ElementX/Sources/Screens/RoomMemberListScreen/RoomMembersListScreenViewModel.swift
pixlwave: Ok will do 👍 (I hesitated as wasn't sure if it was specific to this screen).

ElementX/Sources/Screens/RoomMemberListScreen/RoomMembersListScreenViewModel.swift
pixlwave: To me, `sortedName` would be an existing name that has been sorted. Whereas this is the name used for sorting.

PR #2449, opened 2024-02-09
Restored Link long pressing

### LINKED ISSUE #2440: No way to copy links from msgs
### Steps to reproduce

1. Someone sends you a url buried in a msg
2. you want to copy it to put it somewhere else (eg refer to it in a bug report)
3. you long tap on it
4. the context menu gives you options to copy the whole msg or copy link to msg - but not copy the thing you just selected in the long press.

#### What did you expect?

add an option to the context menu to copy link?

#### What happened instead?

you have to tap on the link to open it in a browser and copy the url from the url bar

or you have to copy the whole msg text and then manually trim out the contents which wasn’t the url

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

516

### Homeserver

_No response_

### Will you send logs?

No

Velin92:
Fixes #2440 

This requires however that our long press matches the system default so it has been slightly increased from 0.35 to 0.5, however to not create a feeling of delay between the tap and the appearance  of the sheet, I slowed down slightly the spring animation

https://github.com/element-hq/element-x-ios/assets/34335419/547e19b8-ca2b-4831-a095-d2622ae050b6

https://github.com/element-hq/element-x-ios/assets/34335419/852c5d61-b33f-4009-8fda-e3b09a94ce8d


ElementX/Sources/Other/Pills/MessageText.swift
```diff
@@ -128,6 +128,7 @@ struct MessageText: UIViewRepresentable {
 
     final class Coordinator: NSObject, UITextViewDelegate {
         var openURLAction: OpenURLAction
+        let permalinkBaseURL = ServiceLocator.shared.settings.permalinkBaseURL
```
stefanceriu, line 131: I'm not a big fan of this but I also don't have any ideas at the moment on how to improve it so ¯\_(ツ)_/¯

stefanceriu (approved):
lgtm 👍

PR #2455, opened 2024-02-12
Typing notifications

stefanceriu:
This PR will
- send typing notifications based on the message composer's content
- listen to and render other people's typing notifications as a "virtual" timeline item at the bottom of the items list
- introduce a new advanced settings flag called `Share presence` which if
    - enabled: will send public read receipts and typing notifications as well as render others'
    - disabled: will send private read receipts and no typing notifications as well as not render others' read receipts or typing notifications

Fixes https://github.com/element-hq/element-x-ios/issues/2347, https://github.com/element-hq/element-x-ios/issues/2348 and https://github.com/element-hq/element-x-ios/issues/2349

P.S. there's a screenshot here of all the cases https://github.com/element-hq/element-x-ios/blob/df7980320548be995b95160388893d8a8314f3ab/UnitTests/__Snapshots__/PreviewTests/test_typingIndicatorView.1.png

Screenshot 2024-02-12 at 15 47 03

<details> <summary> Old video </summary>

https://github.com/element-hq/element-x-ios/assets/637564/f32c64b4-b991-4054-b3a6-2895260c0b75

</details>

https://github.com/element-hq/element-x-ios/assets/637564/31c6ee8d-f77b-462e-863f-a79897091880

### DIFF
--- a/ElementX/Sources/Application/AppSettings.swift
+++ b/ElementX/Sources/Application/AppSettings.swift
@@ -37,7 +37,7 @@ final class AppSettings {
         case viewSourceEnabled
         case richTextEditorEnabled
         case appAppearance
-        case sendReadReceiptsEnabled
+        case sharePresence
         case hideUnreadMessagesBadge
         
         case elementCallBaseURL
@@ -266,10 +266,10 @@ final class AppSettings {
     // maptiler api key
     let mapTilerApiKey = InfoPlistReader.main.mapLibreAPIKey
     
-    // MARK: - Read Receipts
+    // MARK: - Presence
 
-    @UserPreference(key: UserDefaultsKeys.sendReadReceiptsEnabled, defaultValue: true, storageType: .userDefaults(store))
-    var sendReadReceiptsEnabled
+    @UserPreference(key: UserDefaultsKeys.sharePresence, defaultValue: true, storageType: .userDefaults(store))
+    var sharePresence
     
     // MARK: - Feature Flags
     
--- a/ElementX/Sources/Screens/ComposerToolbar/ComposerToolbarViewModel.swift
+++ b/ElementX/Sources/Screens/ComposerToolbar/ComposerToolbarViewModel.swift
@@ -68,7 +68,10 @@ final class ComposerToolbarViewModel: ComposerToolbarViewModelType, ComposerTool
             .store(in: &cancellables)
 
         wysiwygViewModel.$isContentEmpty
-            .weakAssign(to: \.state.composerEmpty, on: self)
+            .sink { [weak self] isEmpty in
+                self?.state.composerEmpty = isEmpty
+                self?.actionsSubject.send(.contentChanged(isEmpty: isEmpty))
+            }
             .store(in: &cancellables)
 
         wysiwygViewModel.$actionStates
--- a/ElementX/Sources/Screens/ComposerToolbar/ComposerToolbarViewModel.swift
+++ b/ElementX/Sources/Screens/ComposerToolbar/ComposerToolbarViewModel.swift
@@ -68,6 +68,7 @@ final class ComposerToolbarViewModel: ComposerToolbarViewModelType, ComposerTool
             .store(in: &cancellables)
 
         wysiwygViewModel.$isContentEmpty
+            .removeDuplicates()
             .sink { [weak self] isEmpty in
                 self?.state.composerEmpty = isEmpty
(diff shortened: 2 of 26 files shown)


ElementX/Sources/Application/AppSettings.swift
```diff
@@ -37,7 +37,7 @@ final class AppSettings {
         case viewSourceEnabled
         case richTextEditorEnabled
         case appAppearance
-        case sendReadReceiptsEnabled
```
Velin92, line 40: Just a note, we should definitely remove the old copies from localazy for the old setting.

ElementX/Sources/Screens/ComposerToolbar/ComposerToolbarViewModel.swift
```diff
@@ -68,7 +68,10 @@ final class ComposerToolbarViewModel: ComposerToolbarViewModelType, ComposerTool
             .store(in: &cancellables)
 
         wysiwygViewModel.$isContentEmpty
-            .weakAssign(to: \.state.composerEmpty, on: self)
```
Velin92, line 71: let's add a removeDuplicates

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -317,6 +329,12 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
             .weakAssign(to: \.state.members, on: self)
             .store(in: &cancellables)
         
+        roomProxy.typingMembers
+            .receive(on: DispatchQueue.main)
+            .filter { [weak self] _ in self?.appSettings.sharePresence ?? false }
```
Velin92, line 334: if you are not sharing your presence, we are disabling also the sharing of others' presence with you?

ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineItemStatusView.swift
```diff
@@ -33,7 +33,7 @@ struct TimelineItemStatusView: View {
 
     @ViewBuilder
     private var mainContent: some View {
-        if !timelineItem.properties.orderedReadReceipts.isEmpty {
+        if context.viewState.showReadReceipts, !timelineItem.properties.orderedReadReceipts.isEmpty {
```
Velin92, line 36: same also for RRs I imagine, like WA essentially?

stefanceriu: yep yep, if you don't share you don't get, quid pro quo 😁

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
stefanceriu: Well sort of, we can't stop them from sending it but we can stop it from being rendered

ElementX/Sources/Application/AppSettings.swift
stefanceriu: I'm not sure we can just delete them, at least not until we make sure other clients stop using them too 🤔

ara4n:
I'm a bit worried about the design here (but i don't know if this is the impl or the design's fault):

> it feels a bit weird that the timeline pops as the notification disappears, and then pops again as the message appears
in EW we put a lot of effort into ensuring that the notification height was the same as a (one line) text bubble height
and that we wouldn't pop the timeline down between typing stopping and message sending
(and put a delay after typing stops before hiding the gap in the timeline created by the typing notification, to avoid someone stuttering causing the timeline to bounce/pop)
is the video a feature or a bug?

amshakal:
I agree with you, I don't think the message bubbles should move up and down so much between typing and actually sending the message

stefanceriu:
The original implementation wasn't quite correct as the typing indicator would get removed without an animation. I tweaked the code and updated the video in the description above.

The issue reported by Matthew is still there but if we don't want the timeline to move with typing notifications then I believe this approach is fundamentally wrong. Instead of putting a lot of effort into forcing it into that position and being subject to network lag and random people typing we should just move it to somewhere where it doesn't affect the timeline layout.

stefanceriu:
Here's a new video following recent discussions and tweaks

https://github.com/element-hq/element-x-ios/assets/637564/139fd070-0dc0-4414-b187-514b2a0f5981

Velin92 (approved):
LGTM

PR #2456, opened 2024-02-12
Fix for isSearchFieldFocus not always updating properly

Velin92:
On some slower devices this can create an issue where we are still hiding the rooms for the search, even if it has been dismissed.


stefanceriu (commented):
Let's try it out 👍

PR #2459, opened 2024-02-12
Fix filters being jumpy when scrolling them

Velin92:
The fix also improved the overall code by resorting to use `Binding` in place of `StateObject` for the state and matched geometry and HStack to create the scrollable view

### DIFF
--- a/ElementX/Sources/Screens/HomeScreen/View/Filters/RoomListFiltersView.swift
+++ b/ElementX/Sources/Screens/HomeScreen/View/Filters/RoomListFiltersView.swift
@@ -21,24 +21,21 @@ struct RoomListFiltersView: View {
     
     var body: some View {
         ScrollView(.horizontal) {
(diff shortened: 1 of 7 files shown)


ElementX/Sources/Screens/HomeScreen/View/Filters/RoomListFiltersView.swift
```diff
@@ -17,28 +17,30 @@
 import SwiftUI
 
 struct RoomListFiltersView: View {
-    @StateObject var state: RoomListFiltersState
+    @Binding var state: RoomListFiltersState
+    @Namespace var namespace
```
pixlwave, line 21: I never really know how to name these in a meaningful way. I'm sure this is fine, but should probably be private given it makes no sense outside of the body of this view?

pixlwave (approved):
Nice! 👌

PR #2460, opened 2024-02-12
Show banned members in the members list.

### LINKED ISSUE #2355: [Task] Show sections for roles in room-member list

pixlwave:
This PR adds a segmented picker to the room members list for anyone who has the power to ban users. The picker is only shown when there are banned users in the room and lets them toggle between current members and banned users.

Additionally it re-works some display name handling where we were forcing the display name to be the MXID at the model level even though it's valid to `nil` (on top of which at the view level we were also doing `displayname ?? ""`). All of this is now handled at the view level, along with hiding the profiles for banned users. I've added cases in snapshot tests that cover everything that we handle.

And there's a bug fix for member sorting that was sorting upper-case first, then lower-case then Ștefan at the bottom 🥲

Closes #2355.

Screenshot 2024-02-15 at 5 40 32 pm

### DIFF
--- a/ElementX/Sources/Screens/RoomMemberListScreen/View/RoomMembersListScreen.swift
+++ b/ElementX/Sources/Screens/RoomMemberListScreen/View/RoomMembersListScreen.swift
@@ -107,6 +107,7 @@ struct RoomMembersListScreen: View {
 
 struct RoomMembersListScreen_Previews: PreviewProvider, TestablePreview {
     static let viewModel = makeViewModel()
+    static let invitesViewModel = makeViewModel(withInvites: true)
     static let adminViewModel = makeViewModel(isAdmin: true, initialMode: .members)
     static let bannedViewModel = makeViewModel(isAdmin: true, initialMode: .banned)
     
@@ -116,6 +117,12 @@ struct RoomMembersListScreen_Previews: PreviewProvider, TestablePreview {
         }
         .snapshot(delay: 1.0)
         .previewDisplayName("Member")
+        
+        NavigationStack {
+            RoomMembersListScreen(context: invitesViewModel.context)
+        }
+        .snapshot(delay: 1.0)
+        .previewDisplayName("Invites")
         
         NavigationStack {
             RoomMembersListScreen(context: adminViewModel.context)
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Screens/RoomMemberListScreen/View/RoomMembersListScreenMemberCell.swift
```diff
@@ -67,22 +70,59 @@ struct RoomMembersListScreenMemberCell: View {
...
+    }
 }
 
 struct RoomMembersListMemberCell_Previews: PreviewProvider, TestablePreview {
     static let members: [RoomMemberProxyMock] = [
         .mockAlice,
-        .mockBob,
-        .mockCharlie,
-        .mockModerator
+        .mockAdmin,
```
stefanceriu, line 97: Should we add invitedAlice in here as well? It would be nice to show the 2 sections in the snapshots.

ElementX/Sources/Screens/RoomMemberListScreen/RoomMembersListScreenModels.swift
```diff
@@ -21,41 +21,59 @@ enum RoomMembersListScreenViewModelAction {
     case invite
 }
 
+/// The different modes that the screen can be in.
+enum RoomMembersListScreenMode {
```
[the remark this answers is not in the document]
pixlwave: That's what the product requirements asked for yes. @VolkerJunginger are you still happy with that decision. I think one of the pros of doing this would be that some users are banned based on offensive names/user IDs etc.

ElementX/Sources/Screens/RoomMemberListScreen/View/RoomMembersListScreenMemberCell.swift
```diff
@@ -67,22 +70,59 @@ struct RoomMembersListScreenMemberCell: View {
             nil
         }
     }
+    
+    // Computed properties to hide the user's profile when banned.
```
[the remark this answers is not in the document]
pixlwave: To reduce the places where admins/mods would be shown offensive names/avatars.

ElementX/Sources/Screens/RoomMemberListScreen/View/RoomMembersListScreenMemberCell.swift
pixlwave: There isn't any difference between the cell configuration for invited/joined users so I didn't see the use case. We show them properly in the snapshots for the screen, this is just looking at the cell.

stefanceriu: Oh sorry, I thought we have different sections in the list for invited vs joined

pixlwave: Actually, whilst I don't seen an advantage here, you're right, we should add an invited member to the screen snapshot 👍

PR #2466, opened 2024-02-14
Fixed the link color and improved link dection in code blocks

### LINKED ISSUE #2379: Links in HTML are almost illegible on dark theme
### Steps to reproduce

1. Try to read this msg:

image

2. on an iphone screen the dark blue on dark grey is almost illegible

The source is:

```json
    "body" : "In [`Room::encrypt_directly`](https:\/\/github.com\/matrix-org\/matrix-rust-sdk-project-neo\/blob\/071df28a4ec0b3ce030c2993290d9425372819e4\/crates\/matrix-sdk\/src\/room\/mod.rs#L1373-L1374)",
    "formatted_body" : "In <a href=\"https:\/\/github.com\/matrix-org\/matrix-rust-sdk-project-neo\/blob\/071df28a4ec0b3ce030c2993290d9425372819e4\/crates\/matrix-sdk\/src\/room\/mod.rs#L1373-L1374\"><code>Room::encrypt_directly<\/code><\/a>",
```

#### What did you expect?

legible links

#### What happened instead?

blue on black.

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

503

### Homeserver

_No response_

### Will you send logs?

No

Velin92:
fixes #2379


ElementX/Sources/Other/HTMLParsing/AttributedStringBuilder.swift
```diff
@@ -52,8 +52,8 @@ struct AttributedStringBuilder: AttributedStringBuilderProtocol {
 
         let mutableAttributedString = NSMutableAttributedString(string: string)
         addLinksAndMentions(mutableAttributedString)
-        detectPermalinks(mutableAttributedString)
         removeLinkColors(mutableAttributedString)
+        detectPermalinks(mutableAttributedString)
```
pixlwave, line 56: How come the change in order here? Might be good to add a comment, as to the eye it looks like we remove link colouring before we add the links. I guess we're only removing the colours for links added in HTML, but then that's why I'm curious about the change :)

pixlwave (approved):
LGTM, (could use a changelog)

ElementX/Sources/Other/HTMLParsing/AttributedStringBuilder.swift
Velin92: Because detect permalink is also the place where all schemes as set as https://

Velin92: Just something I left out while playing around with the fix, reverting this change, is not needed

PR #2469, opened 2024-02-15
Fix localization bundle and attributed string builder crashes on unsy…

stefanceriu:
…nchronised access to their internal caches

We've had sentry reports about this for quite a while now. I've decided to go with plain DispatchQueues instead of an Actor so that I wouldn't have to modify existing interfaces and make them async.

This accounts for almost all of the crashes reported on Sentry

### DIFF
--- a/ElementX/Sources/Other/Extensions/Bundle.swift
+++ b/ElementX/Sources/Other/Extensions/Bundle.swift
@@ -32,13 +32,17 @@ public extension Bundle {
     
     // MARK: - Localisation
     
-    private static var cachedLocalizationBundles = [String: Bundle]()
+    /// Overrides `Bundle.app.preferredLocalizations` for testing translations.
+    static var overrideLocalizations: [String]?
+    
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Other/Extensions/Bundle.swift
```diff
@@ -48,13 +52,25 @@ public extension Bundle {
...
         return bundle
     }
-
-    /// Overrides `Bundle.app.preferredLocalizations` for testing translations.
-    static var overrideLocalizations: [String]?
+    
+    // MARK: - Private
+    
+    private static func cacheValue(_ value: Bundle?, forKey key: String) {
+        cacheDispatchQueue.sync {
```
pixlwave, line 63: So if this is called from the main thread, when a bg thread is also calling this, won't it block the main thread?

pixlwave (approved):
👍

ElementX/Sources/Other/Extensions/Bundle.swift
stefanceriu: It will yes, but it shouldn't be noticeable

PR #2471, opened 2024-02-15
Last session recovery

stefanceriu:
Best reviewd commit by commit.

Allow the user to recover their last session through entering their recovery key. Fixes https://github.com/element-hq/element-x-ios/issues/2470 & https://github.com/element-hq/element-x-ios/issues/2424

This PR reworks how we handle session verification and session recovery. The userSession now publishes them combined after fetching the data internally from the session verification controller and secure backup controller.

On both the home and settings screens unverified/incompletely recovered sessions are now treated very similarly: listen to the `UserSession.sessionSecurity` callbacks, switch on session verification and recovery states, decide what UI to show. 

The UI is also very similar between them with both showing badges on the avatar and settings entry, and the view model being responsible for all the logic.

### DIFF
--- a/ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
+++ b/ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
@@ -75,33 +75,21 @@ enum HomeScreenRoomListMode: CustomStringConvertible {
     }
 }
 
+enum SecurityBannerMode {
+    case none
+    case dismissed
+    case sessionVerification
+    case recoveryKeyConfirmation
+}
+
 struct HomeScreenViewState: BindableState {
     let userID: String
     var userDisplayName: String?
     var userAvatarURL: URL?
     
-    var isSessionVerified: Bool?
-    var hasSessionVerificationBannerBeenDismissed = false
-    var showSessionVerificationBanner: Bool {
-        guard let isSessionVerified else {
-            return false
-        }
+    var securityBannerMode = SecurityBannerMode.none
+    var requiresExtraAccountSetup = false
         
-        return !isSessionVerified && !hasSessionVerificationBannerBeenDismissed
-    }
-    
-    var requiresSecureBackupSetup = false
-
-    var needsRecoveryKeyConfirmation = false
-    var hasRecoveryKeyConfirmationBannerBeenDismissed = false
-    var showRecoveryKeyConfirmationBanner: Bool {
-        guard let isSessionVerified else {
-            return false
-        }
-        
-        return isSessionVerified && needsRecoveryKeyConfirmation && !hasRecoveryKeyConfirmationBannerBeenDismissed
-    }
-    
     var rooms: [HomeScreenRoom] = []
     var roomListMode: HomeScreenRoomListMode = .skeletons
     
--- a/ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
+++ b/ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
@@ -61,7 +61,7 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol 
             .weakAssign(to: \.state.userDisplayName, on: self)
             .store(in: &cancellables)
         
-        userSession.sessionSecurityState
+        userSession.sessionSecurityStatePublisher
             .receive(on: DispatchQueue.main)
             .sink { [weak self] securityState in
                 guard let self else { return }
--- a/ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
+++ b/ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
@@ -61,20 +61,35 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol 
             .weakAssign(to: \.state.userDisplayName, on: self)
             .store(in: &cancellables)
         
-        userSession.sessionVerificationState
+        userSession.sessionSecurityState
             .receive(on: DispatchQueue.main)
-            .weakAssign(to: \.state.isSessionVerified, on: self)
-            .store(in: &cancellables)
-        
-        userSession.clientProxy.secureBackupController.recoveryKeyState
-            .receive(on: DispatchQueue.main)
-            .sink { [weak self] recoveryKeyState in
+            .sink { [weak self] securityState in
                 guard let self else { return }
                 
-                let requiresSecureBackupSetup = recoveryKeyState == .disabled || recoveryKeyState == .incomplete
-                state.requiresSecureBackupSetup = requiresSecureBackupSetup
-                
-                state.needsRecoveryKeyConfirmation = recoveryKeyState == .incomplete
+                switch (securityState.verificationState, securityState.recoveryState) {
+                case (.unverified, _):
+                    state.requiresExtraAccountSetup = true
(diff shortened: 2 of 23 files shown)


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -166,6 +166,16 @@ class ClientProxy: ClientProxyProtocol {
         let digest = SHA256.hash(data: data)
         return digest.compactMap { String(format: "%02x", $0) }.joined()
     }()
+    
+    func isLastDevice() async -> Result<Bool, ClientProxyError> {
```
Velin92, line 170: I find this name a bit confusing... from what I undersand this means that this is the only device left right?
But the naming "last" kinda implies for me of a order of something, like this is the "last" device that has established a session.

We don't need to change the SDK name, even if I would prefer, but could this be called `isLastDeviceLeft` or `isOnlyDeviceLeft` ?
Something that specifies more the fact that "is the only one left".

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
```diff
@@ -75,33 +75,21 @@ enum HomeScreenRoomListMode: CustomStringConvertible {
     }
 }
 
+enum SecurityBannerMode {
+    case none
+    case dismissed
+    case sessionVerification
+    case recoveryKeyConfirmation
```
Velin92, line 82: way better now!

ElementX/Sources/Screens/SecureBackup/SecureBackupRecoveryKeyScreen/View/SecureBackupRecoveryKeyScreen.swift
```diff
@@ -65,23 +65,25 @@ struct SecureBackupRecoveryKeyScreen: View {
...
+                Button {
+                    context.send(viewAction: .done)
+                } label: {
+                    Text(L10n.actionDone)
                 }
                 .buttonStyle(.compound(.primary))
-                .simultaneousGesture(TapGesture().onEnded { _ in
-                    context.send(viewAction: .keySaved)
-                })
+                .disabled(context.viewState.recoveryKey == nil || context.viewState.doneButtonEnabled == false)
```
Velin92, line 85: can we incorporate this into a `private var some View` up to you to chose the naming

ElementX/Sources/Screens/Settings/SettingsScreen/SettingsScreenModels.swift
```diff
@@ -35,15 +35,22 @@ enum SettingsScreenViewModelAction {
 }
 
 struct SettingsScreenViewState: BindableState {
+    enum SecuritySectionMode {
+        case none
+        case sessionVerification
+        case secureBackup
+    }
```
Velin92, line 42: maybe I would extract this from this struct

ElementX/Sources/Services/Session/UserSession.swift
```diff
@@ -49,89 +51,105 @@ class UserSession: UserSessionProtocol {
         }
     }
     
-    private var sessionVerificationStateSubject: CurrentValueSubject<Bool?, Never> = .init(nil)
-    var sessionVerificationState: CurrentValuePublisher<Bool?, Never> {
-        sessionVerificationStateSubject.asCurrentValuePublisher()
+    private var sessionSecurityStateSubject: CurrentValueSubject<SessionSecurityState, Never> = .init(.init(verificationState: .unknown, recoveryState: .unknown))
```
Velin92, line 54: Actually I don't think this needs to be CurrentValueSubject. 
Since it's created in the init, and we never need to read its value directly, but we only listen for its changes, I think this could easily be a `AnyPublisher<SessionSecurityState,Never>` 

When this is initialised we just do:
```
sessionSecurityStatePublisher = Publishers.CombineLatest(sessionVerificationStateSubject, clientProxy.secureBackupController.recoveryState)
            .removeDuplicates { $0 == $1 }
            .map(SessionSecurityState.init)
            .eraseToAnyPublisher()
```

Ofc we should change its name to `sessionSecurityStatePublisher`

Velin92: If you want to keep the logging part that can be done in the map

ElementX/Sources/Services/Session/UserSession.swift
```diff
@@ -49,89 +51,105 @@ class UserSession: UserSessionProtocol {
...
+        Publishers.CombineLatest(sessionVerificationStateSubject, clientProxy.secureBackupController.recoveryState)
+            .removeDuplicates { $0 == $1 }
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] verificationState, recoveryState in
+                
+                MXLog.info("Session security state changed, verificationState: \(verificationState), recoveryState: \(recoveryState)")
+                
+                self?.sessionSecurityStateSubject.send(.init(verificationState: verificationState, recoveryState: recoveryState))
+            }
+            .store(in: &cancellables)
```
Velin92, line 95: the part I am referring above, with that we don't even need to store this extra cancellable

Velin92 (approved):
LGTM just a naming suggestion, and a small publisher improvement which would lead to use one less subscriber

ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu: Yeah, I can see how this is confusing. I went with `lastDevice` just to keep in sync with the SDK naming but I have nothing against chaging it. I like `isOnlyDeviceLeft` 👍

ElementX/Sources/Screens/SecureBackup/SecureBackupRecoveryKeyScreen/View/SecureBackupRecoveryKeyScreen.swift
stefanceriu: I went with `recoveryCreatedActionButtons` ¯\_(ツ)_/¯

ElementX/Sources/Services/Session/UserSession.swift
stefanceriu: Love it!

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -61,20 +61,35 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
...
+                    if state.securityBannerMode != .dismissed {
+                        state.securityBannerMode = .sessionVerification
+                    }
+                case (.unverifiedLastSession, .incomplete):
+                    state.requiresExtraAccountSetup = true
+                    if state.securityBannerMode != .dismissed {
+                        state.securityBannerMode = .recoveryKeyConfirmation
+                    }
+                case (.verified, .disabled):
+                    state.requiresExtraAccountSetup = true
```
bmarty, line 81: @stefanceriu why `requiresExtraAccountSetup` is set to true in this case?

bmarty: Actually we are currently doing the same on Android. but it looks weird to have the indicator *always* be displayed when the user has explicitly disabled the key backup.

stefanceriu: I believe this is about session recovery not key backup, we really want you to setup a recovery key.

PR #2474, opened 2024-02-16
Replace share location icon with the iOS share location icon

### LINKED ISSUE #1486: It's not at all clear that the "Share my location" button is a button
### Steps to reproduce

1. Select "location" from the (+) button on the composer.
2. A popup card shows you the map with your location indicated with a (massive) pin-inside-a-pin. (sidenote: this feels way too big, and hides way too much surrounding map data; for instance, right now I can't even read the name of the street i'm on because the huge circle covers it. it feels like we have visibility-impaired mode enabled)
3. Look around for a button to press to share your location.
4. Notice there's a "Share my location" text at the bottom of the card. It looks a bit like a legend describing what the big pin means.
5. Look around some more
6. Realise that this text is actually a button you're meant to press in order to perform the share.

#### What did you expect?

I expected there to be some hint that the "Share this location" text at the bottom is actually a button you're meant to interact with, rather than a legend or some other informative text:

image

That could be via a green 'action' highlight of some kind, or making it look like a button (by putting it on a lozenge or whatever), or some other affordance.  But right now there ain't no affordance.

#### What happened instead?

This is the 3rd time or so i've tried to share my location, and found myself flummoxed by the UI.

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

340

### Homeserver

_No response_

### Will you send logs?

No

Velin92:
fixes #1486


ElementX/Sources/Screens/LocationSharing/View/StaticLocationScreen.swift
```diff
@@ -110,7 +110,7 @@ struct StaticLocationScreen: View {
             context.send(viewAction: .selectLocation)
         } label: {
             HStack(spacing: 8) {
-                LocationMarkerView(mode: .button)
```
pixlwave, line 113: I may be mistaken, but off the top of my head I have a feeling this `mode` is no longer used and we could simplify `LocationMarkerView` to be much simpler.

pixlwave (approved):
LGTM 👍

ElementX/Sources/Screens/LocationSharing/View/StaticLocationScreen.swift
Velin92: removed the mode and the button configuration for this view

PR #2476, opened 2024-02-16
Fix a retain cycle on MessageText.

pixlwave:
This was generating warnings about an excessive number of gesture recognisers.


stefanceriu (approved):
Niiice!

PR #2483, opened 2024-02-19
Session recovery verification

stefanceriu:
This PR allows recovering your session by entering your passkey starting from the session verification screen (if recovery was previously set up)
It also updates translations following the copy change from "Confirm recovery key" to "Enter key"

Please see attached video for flows and details:

https://github.com/element-hq/element-x-ios/assets/637564/bcf3e68c-eda9-421f-94a0-8b28ac7028f0


pixlwave (approved):
LGTM 👍

PR #2492, opened 2024-02-21
Move PreviewTests into their own scheme.

pixlwave:
This PR moves the preview tests into their own scheme so it is easier to run them without having to run all the unit tests for the plugin's generation step to be called.

We didn't go with a shared scheme and multiple targets in the end as we want to run the preview tests in 2 languages at some point, and having a single scheme with multiple plans didn't seem to be all that different in usage to 2 schemes (and slathering the coverage may have ended up being more tricky).

The Fastfile is updated to run both schemes in the unit_tests lane and slather the output of both. It doesn't appear that we need an update to the workflow to upload both of these coverage files, but lets see what happens on CI to make sure 🤞


stefanceriu (approved):
Looks good to me! 👏  Let's see how it goes on the CI

PR #2493, opened 2024-02-21
Configure SDK to use the global/OS proxy

### LINKED ISSUE #2489: [Task] Configure the SDK to use OS proxy settings
It is now possible with https://github.com/matrix-org/matrix-rust-sdk/pull/3145.

Velin92:
fixes #2489

### DIFF
--- a/ElementX/Sources/Other/Extensions/URL.swift
+++ b/ElementX/Sources/Other/Extensions/URL.swift
@@ -67,4 +67,17 @@ extension URL: ExpressibleByStringLiteral {
 
         return url
     }
+    
+    var globalProxy: String? {
+        if let proxySettingsUnmanaged = CFNetworkCopySystemProxySettings() {
+            let proxySettings = proxySettingsUnmanaged.takeRetainedValue()
+            let proxiesUnmanaged = CFNetworkCopyProxiesForURL(self as CFURL, proxySettings)
+            if let proxy = (proxiesUnmanaged.takeRetainedValue() as? [[AnyHashable: Any]])?.first,
+               let hostname = proxy[kCFProxyHostNameKey] as? String,
+               let port = proxy[kCFProxyPortNumberKey] as? Int {
+                return "\(hostname):\(port)"
+            }
+        }
+        return nil
+    }
 }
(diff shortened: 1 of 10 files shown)


ElementX/Sources/Other/Extensions/URL.swift
```diff
@@ -67,4 +67,17 @@ extension URL: ExpressibleByStringLiteral {
 
         return url
     }
+    
+    var globalProxy: String? {
+        if let proxySettingsUnmanaged = CFNetworkCopySystemProxySettings() {
+            let proxySettings = proxySettingsUnmanaged.takeRetainedValue()
+            let proxiesUnmanaged = CFNetworkCopyProxiesForURL(self as CFURL, proxySettings)
+            if let proxy = (proxiesUnmanaged.takeRetainedValue() as? [[AnyHashable: Any]])?.first,
```
Velin92, line 75: honestly there may be multiple proxies, so I just take the first one... since we should assume that the first is the highest in priority.

ElementX/Sources/Services/UserSession/UserSessionStore.swift
```diff
@@ -115,19 +115,28 @@ class UserSessionStore: UserSessionStoreProtocol {
...
             .enableCrossProcessRefreshLock(processId: InfoPlistReader.main.bundleIdentifier,
                                            sessionDelegate: keychainController)
             .serverVersions(versions: ["v1.0", "v1.1", "v1.2", "v1.3", "v1.4", "v1.5"]) // FIXME: Quick and dirty fix for stopping version requests on startup https://github.com/matrix-org/matrix-rust-sdk/pull/1376
-
+        
+        if let homeserverURL = URL(string: homeserverURL),
+           let proxy = homeserverURL.globalProxy {
+            builder = builder.proxy(url: proxy)
+        }
+        let completeBuilder = builder
```
pixlwave, line 135: I'm not sure I see what this line is about?

pixlwave (approved):
👍

ElementX/Sources/Services/UserSession/UserSessionStore.swift
pixlwave: Ahhhh, ok makes sense now 👍

PR #2494, opened 2024-02-22
Attempt to fix multiline Testflight changelog

stefanceriu:


pixlwave (approved):

ci_scripts/ci_common.sh
```diff
@@ -85,8 +85,8 @@ generate_what_to_test_notes() {
 
         NOTES=$(git log --pretty='- %an: %s' "$LATEST_TAG"..HEAD)
 
-        echo "generate_what_to_test_notes: Generated notes:\n$NOTES"
+        echo "generate_what_to_test_notes: Generated notes:\n"$NOTES""
```
pixlwave, line 88: Actually, wait… Are the `"` nested inside the `"` ok here?

stefanceriu: I tested them locally and seem fine

PR #2495, opened 2024-02-22
Handover log file management to Rust.

pixlwave:
The PR lets Rust do the log file management which gives us the following:
- No longer need to choose between console *or* file output, we get both 🥳
- Automatic file rotation (hourly) so there's no need to chunk the uploads.
- Chronologically sorted logs in rageshakes (see linked RS)
- MXLogger can go 🙌

<del>One caveat is that logs files are named e.g. `console.2024-02-22-12` meaning that we can't QuickLook them directly. So I've had to add a `copyLogs(to:)` method that copies them to a directory whilst appending the `.log` extension. Because of this, the bug report view model copies them to a temporary directory, passes these temporary logs to the view logs screen and on deinit deletes the temporary directory.</del>

The individual commits are maybe useful to look at, but not 100% necessary.


UnitTests/Sources/LoggingTests.swift
pixlwave, line 1: These tests all work when run individually, but not when run together. I've disabled them for now.

pixlwave:
Going to update this following https://github.com/matrix-org/matrix-rust-sdk/pull/3157 as Rust *can* add the `.log` file extension so that should simplify everything even more 👏

UnitTests/Sources/LoggingTests.swift
pixlwave: Re-enabled, but merging all the test cases that use live logs into a single test. Which isn't ideal, but I've marked it in sections and at least they're now running.

UnitTests/Sources/LoggingTests.swift
```diff
@@ -19,207 +19,60 @@
 import XCTest
 
 class LoggingTests: XCTestCase {
+    let tempDir = URL.temporaryDirectory.appending(path: "Logs")
```
Velin92, line 22: where is this used?

UnitTests/Sources/LoggingTests.swift
```diff
@@ -19,207 +19,60 @@
...
     private enum Constants {
         static let genericFailure = "Test failed"
     }
 
     override func setUpWithError() throws {
-        // Force MXLogger to flush stderr and clean up after itself
-        MXLog.configure(logLevel: .info, redirectToFiles: false)
-        
-        MXLogger.deleteLogFiles()
+        RustTracing.deleteLogFiles()
```
Velin92, line 29: I see that we delete the log files in the setup, why then the tests faill when run one after another?

UnitTests/Sources/LoggingTests.swift
pixlwave: Oh yeah, my bad. Removed.

Velin92 (approved):
LGTM, be sure to split the functions into several ones so we can restore individual tests if the deleteLogs() issue is solved in the future.

UnitTests/Sources/LoggingTests.swift
pixlwave: We discussed this and I've added a comment. It seems that Rust's file writer handle isn't aware when the file it is writing to gets deleted, and so any tests that are run after the first one in this file end up writing the data to the wrong place. Have split this up into smaller functions though.

PR #2504, opened 2024-02-27
Blocked users

stefanceriu:
Fixes https://github.com/element-hq/element-x-ios/issues/2486

This PR adds a new blocked users section in the application settings from where users can see who they blocked and can unblock them

I should add a snapshot for the empty state but that's blocked behind refactoring the MockClientProxy and rewriting it on top of a ClientProxyMock + Configuration

Screenshot 2024-02-27 at 12 32 42
Screenshot 2024-02-27 at 12 32 46

### DIFF
--- a/ElementX/Sources/Screens/BlockedUsersScreen/BlockedUsersScreenViewModel.swift
+++ b/ElementX/Sources/Screens/BlockedUsersScreen/BlockedUsersScreenViewModel.swift
@@ -65,9 +65,14 @@ class BlockedUsersScreenViewModel: BlockedUsersScreenViewModelType, BlockedUsers
         }
     }
     
+    func stop() {
+        hideLoadingIndicator()
+    }
+    
     // MARK: - Private
     
     private func unblockUser(_ userID: String) {
+        showLoadingIndicator()
         state.processingUserID = userID
         
         Task {
@@ -76,6 +81,7 @@ class BlockedUsersScreenViewModel: BlockedUsersScreenViewModelType, BlockedUsers
             }
             
             state.processingUserID = nil
+            hideLoadingIndicator()
         }
     }
     
--- a/ElementX/Sources/Screens/BlockedUsersScreen/BlockedUsersScreenViewModel.swift
+++ b/ElementX/Sources/Screens/BlockedUsersScreen/BlockedUsersScreenViewModel.swift
@@ -1,0 +1,97 @@
+//
+// Copyright 2022 New Vector Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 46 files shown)


ElementX/Sources/Screens/BlockedUsersScreen/BlockedUsersScreenViewModel.swift
```diff
@@ -0,0 +1,97 @@
...
+        
+        showLoadingIndicator()
+        
+        clientProxy.ignoredUsersPublisher
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] blockedUsers in
+                guard let self else { return }
+                
+                if let blockedUsers {
+                    hideLoadingIndicator()
```
Velin92, line 46: I think hideLoadingIndicator should be called regardless of if let blockedUsers

ElementX/Sources/Screens/BlockedUsersScreen/BlockedUsersScreenViewModel.swift
```diff
@@ -0,0 +1,97 @@
...
+        state.processingUserID = userID
+        
+        Task {
+            if case .failure = await clientProxy.unignoreUser(userID) {
+                state.bindings.alertInfo = .init(id: .error)
+            }
+            
+            state.processingUserID = nil
+        }
+    }
```
Velin92, line 80: what if I tap on multiple users at the same time? since we keep track of only one processed user at a time, won't this remove the spinner on a task that is currently being served, even if is not completed yet?
Maybe by adding a sleep of 5-10 seconds you could try to this, by tapping multiple different users, and if this changes the state even if the request has not been served yet.

Velin92 (approved):
Setting the processing user ID may need a bit more testing, however we could make it an array/Set so that we can handle also multiple requests at once! (or maybe we can just avoid having more than one request at a time, and make all the buttons wait while 1 user is getting ignored)

ElementX/Sources/Screens/BlockedUsersScreen/View/BlockedUsersScreen.swift
```diff
@@ -0,0 +1,71 @@
...
+    private var content: some View {
+        if context.viewState.blockedUsers.isEmpty {
+            Text(L10n.screenBlockedUsersEmpty)
+                .font(.compound.bodyMD)
+                .foregroundColor(.compound.textSecondary)
+        } else {
+            Form {
+                ForEach(context.viewState.blockedUsers, id: \.self) { userID in
+                    ListRow(label: .avatar(title: userID, icon: avatar(for: userID)),
+                            details: .isWaiting(context.viewState.processingUserID == userID),
```
Velin92, line 42: I see we only block the action for that single specific user, but given the code above, it would be in theory possible to tap multiple of them before each request is completed resetting this state.

ElementX/Sources/Screens/BlockedUsersScreen/BlockedUsersScreenViewModel.swift
stefanceriu: Well if I do that then it's gonna disappear when the publisher sends it's current value, which would be nil.

ElementX/Sources/Screens/BlockedUsersScreen/BlockedUsersScreenViewModel.swift
stefanceriu: Oh, very good catch! I think I'm going to keep it simple and just add a full screen waiting indicator.

PR #2505, opened 2024-02-27
Add roles and permissions screen

pixlwave:
This only implements the screen, I'll implement the other screens and then assemble them together with a flow coordinator once they're all done.

Screenshot 2024-02-27 at 1 06 50 pm


ElementX/Sources/Screens/RoomRolesAndPermissionsScreen/View/RoomRolesAndPermissionsScreen.swift
```diff
@@ -0,0 +1,117 @@
...
+            permissionsSection
+            
+            resetSection
+        }
+        .compoundList()
+        .navigationTitle(L10n.screenRoomRolesAndPermissionsTitle)
+        .navigationBarTitleDisplayMode(.inline)
+    }
+    
+    var rolesSection: some View {
```
stefanceriu, line 35: These vars should probably be private

stefanceriu (approved):
lgtm 👍

PR #2506, opened 2024-02-27
Don't pass a URL as a log destination to rust

ara4n:
On macOS (catalyst) `URL.appGroupContainerDirectory` returns:

`/Users/matthew/Library/Group%20Containers/group.io.element/`

So you need to decode the %20 back into a space, otherwise logging fails with a permissions error.

### Pull Request Checklist

- [x] I read the [contributing guide](https://github.com/element-hq/element-ios/blob/develop/CONTRIBUTING.md)


ElementX/Sources/Other/Logging/RustTracing.swift
```diff
@@ -38,21 +38,25 @@ enum RustTracing {
...
+
         if let otlpConfiguration {
             setupOtlpTracing(config: .init(clientName: "ElementX-iOS",
                                            user: otlpConfiguration.username,
                                            password: otlpConfiguration.password,
                                            otlpEndpoint: otlpConfiguration.url,
                                            filter: configuration.filter,
                                            writeToStdoutOrSystem: true,
-                                           writeToFiles: .init(path: logsDirectory.path(),
+                                           writeToFiles: .init(path: path,
```
pixlwave, line 52: We can also do it like this directly: `logsDirectory.path(percentEncoded: false)` which is preferable as it isn't an optional this way.

pixlwave (approved):
LGTM, although would be nice to use `path(percentEncoded:)` as suggested.

ElementX/Sources/Other/Logging/RustTracing.swift
pixlwave: I've updated and set to auto-merge with this change.

PR #2508, opened 2024-02-28
Replace the existing `MockClientProxy` with a generated one + special…

stefanceriu:
… configuration options

This will help with consolidating how we handle mocks and provide better configuration options


Velin92 (approved):
finally! LGTM

PR #2510, opened 2024-02-28
Use xcresultparser in all workflows.

pixlwave:
After using it in #2492 we saw it provided more reliable coverage reports so let's use it everywhere! Additionally this PR:
- Removes slather as a dependency.
- Fixes a bug where the UI test results would never be zipped after a failure by moving the zipping into the workflow.
- Also zips the results from integration tests.
- Sets both UI and Integration tests to only upload the results on failure.

[UI Tests run](https://github.com/element-hq/element-x-ios/actions/runs/8083588048)
[Integration Tests run](https://github.com/element-hq/element-x-ios/actions/runs/8084522027)

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -36,14 +36,19 @@ jobs:
           INTEGRATION_TESTS_HOST: ${{ secrets.INTEGRATION_TESTS_HOST }}
           INTEGRATION_TESTS_USERNAME: ${{ secrets.INTEGRATION_TESTS_USERNAME }}
           INTEGRATION_TESTS_PASSWORD: ${{ secrets.INTEGRATION_TESTS_PASSWORD }}
+
+      - name: Zip results # for faster upload
+        if: failure()
+        working-directory: fastlane/test_output
+        run: zip -r IntegrationTests.xcresult.zip IntegrationTests.xcresult
       
       - name: Archive artifacts
         uses: actions/upload-artifact@v3
-        if: always()
+        if: failure()
         with:
-          name: test-output
-          path: fastlane/test_output
-          retention-days: 7
+          name: Results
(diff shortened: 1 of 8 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -36,14 +36,19 @@ jobs:
...
         uses: actions/upload-artifact@v3
-        if: always()
+        if: failure()
         with:
-          name: test-output
-          path: fastlane/test_output
-          retention-days: 7
+          name: Results
+          path: fastlane/test_output/IntegrationTests.xcresult.zip
+          retention-days: 1
```
pixlwave, line 51: Maybe controversial? Just wasn't sure if 7 days was completely necessary seeing as these run every 12 hours, but happy to revert if wanted. Same for UI tests too.

stefanceriu: I tend to forget about these tests and look at them when I remember. It would be really annoying not finding anything there.. ¯\_(ツ)_/¯

stefanceriu (approved):
Love it, great change! 👏

.github/workflows/integration-tests.yml
pixlwave: Ah. My thinking was, you're only going care if the last couple of runs failed, and if they passed, you won't look at the older failures.

stefanceriu: When the integration tests were failing I went and downloaded the failures for the past 10 or so and then just fixed them all at once. Not saying that's the right approach but that's what happened 😅

stefanceriu: Let's maybe meet in the middle and go with 3 days?

pixlwave: We discussed and decided to leave it at 7.

PR #2511, opened 2024-02-28
Replace the existing `RoomSummaryProviderMock` with a generated one

stefanceriu:


Velin92 (approved):
LGTM, more generated mocks are good

PR #2514, opened 2024-02-29
Update Compound.

pixlwave:
This fixes a bug introduced in #2438 where placeholder avatars in notifications looked a bit odd for new rooms due to using a colour with an alpha channel. Requires https://github.com/element-hq/compound-ios/pull/60

notification@2x

### DIFF
--- a/ElementX/Sources/Other/Extensions/UNNotificationContent.swift
+++ b/ElementX/Sources/Other/Extensions/UNNotificationContent.swift
@@ -180,7 +180,7 @@ extension UNMutableNotificationContent {
     private func getPlaceholderAvatarImageData(name: String, id: String) async -> Data? {
         // The version value is used in case the design of the placeholder is updated to force a replacement
         let shouldFlipAvatar = shouldFlipAvatar()
-        let prefix = "notification_placeholder\(shouldFlipAvatar ? "V7" : "V6")"
+        let prefix = "notification_placeholder\(shouldFlipAvatar ? "V8F" : "V8")"
         let fileName = "\(prefix)_\(name)_\(id).png"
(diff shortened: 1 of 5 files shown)


project.yml
```diff
@@ -52,7 +52,7 @@ packages:
     # path: ../matrix-rust-sdk
   Compound:
     url: https://github.com/element-hq/compound-ios
-    revision: 98bd0ed412c29bd5b3762b673ca5720b6fb9d713
+    revision: d38528736c3abe9dc739d96391d1ffd372cd802d
```
pixlwave, line 55: Needs updating once https://github.com/element-hq/compound-ios/pull/60 is merged.

ElementX/Sources/Other/Extensions/UNNotificationContent.swift
```diff
@@ -180,7 +180,7 @@ extension UNMutableNotificationContent {
     private func getPlaceholderAvatarImageData(name: String, id: String) async -> Data? {
         // The version value is used in case the design of the placeholder is updated to force a replacement
         let shouldFlipAvatar = shouldFlipAvatar()
-        let prefix = "notification_placeholder\(shouldFlipAvatar ? "V7" : "V6")"
+        let prefix = "notification_placeholder\(shouldFlipAvatar ? "V8F" : "V8")"
```
pixlwave, line 183: I forgot to do this when they first changed so only new rooms showed the bug. Decided we probably don't need to maintain 2 versions here so added the `F`.

project.yml
pixlwave: Done

Velin92 (approved):

PR #2521, opened 2024-03-04
Filters UI Update and ordering

Velin92:
https://github.com/element-hq/element-x-ios/assets/34335419/e21335ff-e07d-4870-98fa-72be0a201735

https://github.com/element-hq/element-x-ios/assets/34335419/d534ea56-fb43-480f-a56a-f385d7122232


ElementX/Sources/Screens/HomeScreen/View/Filters/RoomListFilterModels.swift
```diff
@@ -101,15 +99,21 @@ struct RoomListFiltersState {
            activeFilters.contains(incompatibleFilter) {
             fatalError("[RoomListFiltersState] adding mutually exclusive filters is not allowed")
         }
-        activeFilters.insert(filter)
+        activeFilters.append(filter)
```
pixlwave, line 102: Might be nice to have a comment here similar to below, that states that we want the most recent filter to be enabled to be shown as the last active filters.

pixlwave (approved):
Very nice 👌

PR #2533, opened 2024-03-06
Fix for a potential crash when building a removed timeline item.

pixlwave:
We were seeing the first of these 2 fatal errors being hit in Sentry, which seems likely due to building an item whilst a timeline reset comes in. So making it optional and we can fatal error on actually trying to play the item instead.


stefanceriu (approved):
👌

PR #2534, opened 2024-03-06
Add RoomChangeRolesScreen.

pixlwave:
Similarly to #2513, this screen isn't part of a flow yet, just the screen being added. Will do a flow coordinator next.

| Administrators | Moderators |
| - | - |
| Screenshot 2024-03-06 at 1 13 33 pm | Screenshot 2024-03-06 at 1 13 42 pm |

### DIFF
--- a/ElementX/Sources/Screens/RoomChangeRolesScreen/RoomChangeRolesScreenModels.swift
+++ b/ElementX/Sources/Screens/RoomChangeRolesScreen/RoomChangeRolesScreenModels.swift
@@ -1,0 +1,100 @@
+//
+// Copyright 2022 New Vector Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Collections
+import Foundation
+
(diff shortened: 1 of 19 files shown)


ElementX/Sources/Screens/RoomChangeRolesScreen/RoomChangeRolesScreenModels.swift
```diff
@@ -0,0 +1,100 @@
...
+    
+    /// The screen's title.
+    var title: String {
+        switch mode {
+        case .administrator:
+            L10n.screenRoomChangeRoleAdministratorsTitle
+        case .moderator:
+            L10n.screenRoomChangeRoleModeratorsTitle
+        case .user:
+            ""
```
Velin92, line 47: is this an occurence that can't happen or we just don't have a copy for this case?

ElementX/Sources/Screens/RoomChangeRolesScreen/RoomChangeRolesScreenModels.swift
```diff
@@ -0,0 +1,100 @@
...
+    }
+    
+    /// The visible members in the screen (after searching).
+    var visibleMembers: [RoomMemberDetails] {
+        guard !bindings.searchQuery.isEmpty else { return members }
+        
+        return members.filter {
+            $0.name?.localizedStandardContains(bindings.searchQuery) == true || $0.id.localizedStandardContains(bindings.searchQuery)
+        }
+    }
```
Velin92, line 58: this is a bit long can we use `element in` instead of `$0` and have it on its own line?

Velin92 (approved):
LGTM

ElementX/Sources/Screens/RoomChangeRolesScreen/RoomChangeRolesScreenModels.swift
pixlwave: Good point, it can't happen. I'll put a comment on it.

PR #2540, opened 2024-03-07
Prevent the sync service from looping so tightly on failure.

pixlwave:
When the sliding sync proxy is misconfigured in the well-known, EX will tightloop trying to restart the service easily generating log files that are multi-gigabytes in size. Once the user is bored of the migration screen, signing out doesn't help either as the client continues trying to restart (but this time fails due to the db not being readable).

This PR adds a 250ms delay when restarting the sync service so we generate less logs in this situation and additionally adds a task that can be cancelled to prevent the service from being restarting if `stopSync` has been called.

Note: This is just a workaround until we have https://github.com/matrix-org/matrix-rust-sdk/issues/2822

### DIFF
--- a/ElementX/Sources/Services/Client/ClientProxy.swift
+++ b/ElementX/Sources/Services/Client/ClientProxy.swift
@@ -211,12 +211,38 @@ class ClientProxy: ClientProxyProtocol {
         }
     }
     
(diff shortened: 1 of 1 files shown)


stefanceriu (approved):
Noice!

PR #2541, opened 2024-03-07
Filters, Mark Unread/Read/Favourites FF removals + code and tests clean up

Velin92:
The title says it all.
We also removed the user suggestions, since it's a feature that was not completed for quite some time.

I have also run periphery to remove some unused code


ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -24,9 +24,7 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
     private let name: String
     private let shouldUpdateVisibleRange: Bool
     private let notificationSettings: NotificationSettingsProxyProtocol
-    private let backgroundTaskService: BackgroundTaskServiceProtocol
```
Velin92, line 27: Did this thing have any reason to be here at all? Periphery spotted it has assigned but unused.

ElementX/Sources/Services/Room/RoomMember/RoomMemberProxy.swift
```diff
@@ -18,16 +18,9 @@ import Foundation
 import MatrixRustSDK
 
 final class RoomMemberProxy: RoomMemberProxyProtocol {
-    private let backgroundTaskService: BackgroundTaskServiceProtocol
```
Velin92, line 21: same comment as the other one

ElementX/Sources/Services/Room/RoomMember/RoomMemberProxy.swift
pixlwave, line 1: My bad, forgot about these 🙈

pixlwave (approved):
LGTM 👍

ElementX/Sources/Services/Room/RoomMember/RoomMemberProxy.swift
pixlwave: Yeah this one did, it was for ignoring members when the methods were on the proxy.

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
pixlwave: Possibly for back pagination before that was moved to the timeline?

ElementX/Sources/Services/Room/RoomMember/RoomMemberProxy.swift
Velin92: np, periphery is always here to guide us towards cleaner code

PR #2542, opened 2024-03-07
Zip results from Unit Test failures

pixlwave:
Uploading the unit test results can be slow. E.g. for what ends up as a 41MB download:

Screenshot 2024-03-07 at 3 32 59 pm

So let's try zip them like we do for UI and Integration tests.

Test result, this looks more reasonable to me:
Screenshot 2024-03-07 at 4 02 25 pm


stefanceriu (approved):
lgtm 👌

PR #2545, opened 2024-03-08
Update the SDK adopting new power levels APIs.

pixlwave:
We simplified the SDK to always get `RoomPowerLevels` and always apply `RoomPowerLevelChanges`. This means a load of gymnastics around optionals from #2513 is no longer necessary.


stefanceriu (approved):
Gotta love simplifications!

PR #2547, opened 2024-03-08
Disable auto completion

### LINKED ISSUE #2008: Hitting send unexpectedly triggers predictive text completion
### Steps to reproduce

With predictive text turned on in iOS 17+ (Settings > General > Keyboard > Predictive Text), try writing some text that gets predicted (indicated by a grey suggestion inside the message composer)

#### What did you expect?
I would expect similar behavior to iMessage and other messaging apps.
Coming to mind are:

1. Sending a message while having begun to type a word that gets predicted, it gets completed when hitting send (This is working already)
2. A word you didn't even begin to type, but got predicted, should **not** be completed upon sending
3. When hitting backspace, predicted text should not be completed

#### What happened instead?
1. Worked as expected

2. The word "time" in the below screen capture was predicted, but the user did not show intent to actually want it in the message. iMessage would only complete the word if you at least had entered a "t"

https://github.com/vector-im/element-x-ios/assets/31133207/1de12ab3-d561-4244-8585-62845489b6bb

3. Seen below is the most confusing interaction: When a text is predicted, but the user changes their mind and hits backspace to correct their already written text, the predicted text gets completed in the composer and gets mixed up with whatever remained after the user's correction

https://github.com/vector-im/element-x-ios/assets/31133207/d294be57-2d9d-4e40-b7fe-0115b2d1359e

Another example:

https://github.com/vector-im/element-x-ios/assets/31133207/58b989d4-2be0-4dfe-8d43-a6056f978d90

### Your phone model

iPhone 12 mini

### Operating system version

iOS 17.1

### Application version

1.4.0

### Homeserver

Synapse

### Will you send logs?

No

### LINKED ISSUE #2059: Autocomplete in composer on macOS sonoma corrupts text
### Steps to reproduce

- MBP 14.0 (23A344)
- Advanced composer on
- Version 1.4.0 (77)  (I wish this text were copyable on the Element X settings page)
- Inline predictive text enabled on OS
    - image

When you type a word for which the OS has a predictive suggestion (I think anything longer than a few characters), the next key you press after the inline suggestion has appeared will vape most of the word.

I'm talking here about the inline suggestion (the greyed out end of the suggested word appearing after the characters you've already typed), not the pill suggestion that also sometimes appears (what are you doing Apple I don't need one way to autocomplete words on a computer never mind two).

#### What did you expect?
OS typing suggestions to work as they do natively elsewhere

#### What happened instead?
Deleting the word I'm halfway through typing

### Your phone model

MBP 14.0 (23A344)

### Operating system version

MBP 14.0 (23A344)

### Application version

Version 1.4.0 (77)

### Homeserver

lant.uk

### Will you send logs?

Yes

Velin92:
Given the current issues with predictive text, until we find a way to solve them we disabled auto completion

fixes #2059 
fixes #2008

### DIFF
--- a/ElementX.xcodeproj/project.pbxproj
+++ b/ElementX.xcodeproj/project.pbxproj
@@ -7136,7 +7136,7 @@
 			repositoryURL = "https://github.com/matrix-org/matrix-wysiwyg-composer-swift";
 			requirement = {
 				kind = exactVersion;
-				version = 2.29.0;
+				version = 2.31.0;
 			};
 		};
 		96495DD8554E2F39D3954354 /* XCRemoteSwiftPackageReference "posthog-ios" */ = {
--- a/ElementX.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/ElementX.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -139,8 +139,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/matrix-org/matrix-wysiwyg-composer-swift",
       "state" : {
-        "revision" : "f788fe2482c0b89019f679a1f43dccf9c25a0782",
-        "version" : "2.29.0"
+        "revision" : "e1cb970b144508a13108762651d141f3488e69ae",
+        "version" : "2.31.0"
       }
     },
     {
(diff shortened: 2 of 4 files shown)


pixlwave (approved):
LGTM 👍

PR #2548, opened 2024-03-08
Moderation tweaks

pixlwave:
A selection of tweaks to the moderation implementation:

- Only allow admins to see the roles and permissions screen.
- Hide the selection checkbox on Admins when changing roles.
- Show an empty state for banned users.
- Add separate actions for ban and remove.
- Implement alerts for reset permissions and demote self.
- Add a warning when promoting someone to administrator.

This PR can be reviewed by commit.


Velin92 (approved):
LGTM but for safety also do a Periphery run and see if there is any code that needs to be cleaned up.

PR #2553, opened 2024-03-11
Fix crashes when blocking/unblocking users

stefanceriu:
- SwiftUI crashes when mutating optional state properties in place
- workaround it buy mutating a local reference first


Velin92 (approved):
LGTM

PR #2556, opened 2024-03-12
In preparation of FTUE changes...

stefanceriu:
This PR is mostly preparation for changing the FTUE flows and introducing mandatory verification
* it removes the welcome screen
* lets the UserSessionFlowCoordinator control the whole app hierarchy
* transform the AuthenticationCoordinator into a FlowCoordinator
* updates the session verification to expose the [newly added Rust publishers](https://github.com/matrix-org/matrix-rust-sdk/pull/3193)
* plus various other tweaks


ElementX/Sources/Screens/Authentication/UIConstants.swift
```diff
@@ -28,7 +28,7 @@ enum UIConstants {
     static let actionButtonBottomPadding: CGFloat = 24
     
     /// The padding used to the top of the view for breaker screens that don't have a navigation bar.
-    static let onboardingBreakerScreenTopPadding: CGFloat = 80
+    static let startScreenBreakerScreenTopPadding: CGFloat = 80
```
pixlwave, line 31: Not for this PR but I think this one is probably needs a better name anyway as its used in a bunch of places now.

ElementX/Sources/Screens/AuthenticationStartScreen/View/AuthenticationStartLogo.swift
```diff
@@ -17,7 +17,7 @@
 import SwiftUI
 
 /// The app's logo styled to fit on various launch pages.
-struct OnboardingLogo: View {
+struct AuthenticationStartLogo: View {
```
pixlwave, line 20: This one needs a better name. It's used for the placeholder too.

ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
```diff
@@ -41,30 +42,36 @@ class AuthenticationCoordinator: CoordinatorProtocol {
...
     }
     
     func start() {
         showStartScreen()
     }
     
-    func stop() {
-        stopLoading()
+    func handleAppRoute(_ appRoute: AppRoute, animated: Bool) {
+        fatalError()
```
pixlwave, line 70: We should probably consolidate this with `handleOIDCRedirectURL`. Can't say I'm super keen on fatalErrors for these 2.

ElementX/Sources/Screens/AuthenticationStartScreen/View/AuthenticationStartScreen.swift
```diff
@@ -49,6 +49,7 @@ struct AuthenticationStartScreen: View {
                     Text(L10n.commonReportAProblem)
                         .font(.compound.bodySM)
                         .foregroundColor(.compound.textSecondary)
+                        .padding()
```
pixlwave, line 52: That's mighty generous of you! Could we make this bottom padding only? Just so it doesn't push the continue button up more than it needs to.

pixlwave (approved):
Overall looks good to me, but now I've seen it I wonder if this would be safer:

- AppCoordinator's root nav is never passed to any child flows.
- AuthFlowCoordinator goes back to how it was before, create a nav stack and pass it in to the flow's constructor.
- UserFlowCoordinator stays the same, but we pass in a root nav that we set to the AppCoordinator's root's root. If you get my meaning.

Only thinking that this way the AppCoordinator still has the final say and a rogue flow coordinator couldn't decided to switch away from auth e.g. after a failed sign out or something silly.

ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
stefanceriu: Let's handle all of these changes separately from this PR. For now, seeing as this flow coordinator should never handle any routes I think we're good with fatal errors.

ElementX/Sources/Screens/AuthenticationStartScreen/View/AuthenticationStartScreen.swift
stefanceriu: Sure thing 👍

stefanceriu:
> Overall looks good to me, but now I've seen it I wonder if this would be safer

I gave this a lot of thought but I really don't see any reason why we shouldn't trust our top level flow coordinators. Also, I find it icky for the AppCoordinator to create random UI components depending on what each flow coordinator needs. 

Otherwise, in the UserSessionFlowCoordinator one for example we used to pass in a splitView, which would mean that we aren't able to replace the whole hierarchy with the onboarding flow (like we talked about)

I would to ahead with the approach in this PR and reduce access later, if it proves to be a problem.

PR #2570, opened 2024-03-13
Possible fix for filters view sticking in the middle of the room list

### LINKED ISSUE #2567: Filter bar ends up halfway down the room list
### Steps to reproduce

Unsure what the repro steps are, but the filter bar often ends up at least one or two rooms too 'low' in the room list.  May be related to keybaord interactions when filtering the room list.

### Outcome

Filter bar to stay at the top, not go drifting down the page

### Your phone model

iPh

### Operating system version

iOS 17.4

### Application version

549

### Homeserver

_No response_

### Will you send logs?

Yes

Velin92:
fixes #2567


pixlwave (approved):
🤞🤞

PR #2575, opened 2024-03-14
Send UTD occurences to Posthog

Velin92:
this uses UTD tracking to allow better logging of any UTD issues and send them to posthog. Also bumps the SDK to version 1.1.50

### DIFF
--- a/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
@@ -139,6 +139,17 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
             }
         }
         .store(in: &cancellables)
+        
+        userSession.clientProxy.actionsPublisher
+            .receive(on: DispatchQueue.main)
+            .sink { action in
+                guard case .receivedDecryptionError = action else {
+                    return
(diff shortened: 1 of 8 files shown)


ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -139,6 +139,17 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
...
         .store(in: &cancellables)
+        
+        userSession.clientProxy.actionsPublisher
+            .receive(on: DispatchQueue.main)
+            .sink { action in
+                guard case .receivedDecryptionError = action else {
+                    return
+                }
+                
+                analytics.trackError(context: nil, domain: .E2EE, name: .OlmKeysNotSentError)
```
pixlwave, line 150: I think we need to use the `UnableToDecryptInfo` in here to provide the correct name?

stefanceriu: We don't actually have the info to do that atm, keys not sent was the compromise to unblock this.

pixlwave (approved):
LGTM, although there are conflicts from my updating the analytics package earlier to send the moderation event.

PR #2585, opened 2024-03-20
[WIP] Public Search View

Velin92:
requires using the branch from this SDK PR:  https://github.com/matrix-org/matrix-rust-sdk/pull/3169


ElementX/Sources/Screens/RoomDirectorySearchScreen/View/RoomDirectoryCell.swift
```diff
@@ -0,0 +1,67 @@
...
+    
+    @ViewBuilder @MainActor
+    var avatar: some View {
+        if dynamicTypeSize < .accessibility3 {
+            LoadableAvatarImage(url: result.avatarURL,
+                                name: result.name,
+                                contentID: result.id,
+                                avatarSize: .room(on: .roomDirectorySearch),
+                                imageProvider: imageProvider)
+                .dynamicTypeSize(dynamicTypeSize < .accessibility1 ? dynamicTypeSize : .accessibility1)
```
pixlwave, line 55: Are these 2 dynamic type conditions copied from the home screen? Maybe (but not sure) we don't need them here given the avatar is much smaller?

ElementX/Sources/Screens/RoomDirectorySearchScreen/View/RoomDirectoryCell.swift
```diff
@@ -0,0 +1,67 @@
...
+        
+        return result.roomID
+    }
+
+    var body: some View {
+        ListRow(label: .avatar(title: result.name ?? result.roomID,
+                               description: description,
+                               icon: avatar), kind: .label)
+            .padding(.horizontal, 16)
+            .padding(.vertical, 12)
```
pixlwave, line 44: The ListRow provides this padding already which is why the rows look taller in the snapshots than the designs :)

ElementX/Sources/Services/RoomDirectorySearch/RoomDirectorySearchProxyProtocol.swift
```diff
@@ -0,0 +1,42 @@
...
+    func nextPage() async -> Result<Void, RoomDirectorySearchError>
+}
+
+enum RoomDirectorySearchError: Error {
+    case searchFailed
+    case nextPageQueryFailed
+}
+
+struct RoomDirectorySearchResult: Identifiable {
+    let roomID: String
```
pixlwave, line 33: Could we rename this to `id` and drop the computed property at the bottom?

pixlwave (approved):
LGTM! Just a few minor comments :)

ElementX/Sources/Screens/RoomDirectorySearchScreen/RoomDirectorySearchScreenViewModel.swift
```diff
@@ -0,0 +1,91 @@
...
+            case .success:
+                break
+            case .failure:
+                userIndicatorController.submitIndicator(UserIndicator(id: Self.errorID,
+                                                                      type: .toast,
+                                                                      title: L10n.screenRoomDirectorySearchLoadingError,
+                                                                      iconName: "xmark"))
+            }
+            state.isLoading = false
+        }
```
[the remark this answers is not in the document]
pixlwave: Hmmm, now that I've read the Proxy I see that the results are published by Rust. Do the requests get queued on the Rust side?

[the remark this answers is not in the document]
Velin92: yeah exactly so the `isCanceled` would not help that much, to solve the issue I am just listening to also the changes on the isLoading state, in the combine listener, so that searches are stopped until the loading state is false again, and only done again if the actual term has changed.

ElementX/Sources/Screens/RoomDirectorySearchScreen/View/RoomDirectoryCell.swift
Velin92: yeah, I took the conditions from that, so I can safely remove it?

ElementX/Sources/Screens/RoomDirectorySearchScreen/View/RoomDirectoryCell.swift
```diff
@@ -0,0 +1,67 @@
...
+    var description: String? {
+        if let topic = result.topic {
+            return topic
+        }
+        
+        if result.name == nil {
+            return nil
+        }
+        
+        return result.roomID
```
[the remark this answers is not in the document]
Velin92: @americanrefugee @ganfra 

@pixlwave here raised a good point, the description should probably be the room alias if the topic is missing, and then the id only as a last resort

ElementX/Sources/Screens/RoomDirectorySearchScreen/View/RoomDirectoryCell.swift
pixlwave: Yeah I would say so. We can define something on Compound itself if we actually want this behaviour - then it will be shared across all the screens.

ElementX/Sources/Screens/RoomDirectorySearchScreen/RoomDirectorySearchScreenViewModel.swift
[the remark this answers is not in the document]
pixlwave: Riiiight, I get it now 👍

PR #2589, opened 2024-03-21
Remove pinned views from the home screen room list header, potential …

stefanceriu:
…fix for various layout issues.

We've been getting various reports of broken home screen layouts, either the filters appearing in the middle of the screen or the list spasming when scrolling upwards. We have no good way of reproducing it but I believe it's caused by the recently introduced room list filters and the way they're pinned to the top.

This PR removes the pinning so that we can test it out in the wild and confirm (or deny) they are the cause.


Velin92 (approved):
LGTM, but remember to remove `topSectionFrame` since is not necessary anymore

stefanceriu:
> remember to remove `topSectionFrame`

Oh nice, thank you 👍

PR #2578, opened 2024-03-15
Onboarding flow coordinator and FTUE changes

stefanceriu:
This PR introduces the following **significant** changes:
* session verification is mandatory for all users, the app shouldn't be usable without a verified session
* session verification / recovery, the (optional) app pin lock setup, the analytics consent prompt and the push notifications permissions requesting are now handled by the newly introduced OnboardingFlowCoordinator
* all traces of these flows have been removed from the AuthenticationFlowCoordinator and the main AppCoordinator, and consolidated in the Onboarding

Designs here https://www.figma.com/file/qjs3RKuofeKzBRPnOmLo2a/OIDC-sign-in-with-QR-code?type=design&node-id=19-144144&mode=design&t=7MXZzQhzq8C3h2uv-0

https://github.com/element-hq/element-x-ios/assets/637564/6c8919f6-d5b8-41e3-a037-0764ccd2be17

PR #2596, opened 2024-03-21
Don't require verification on debug builds.

pixlwave:
Small follow-up to #2578 that bypasses verification for debug builds.


stefanceriu (approved):
lgtm 👍

PR #2598, opened 2024-03-22
UI test cleanup

stefanceriu:
With the move to more unit test snapshotting a lot of the UI tests no longer make sense. This PR deletes everything that's not part of a flow and all the pseudo language tests.


ElementX/Sources/UITests/UITestsScreenIdentifier.swift
```diff
@@ -17,75 +17,42 @@
...
-    case invites
-    case invitesWithBadges
-    case invitesNoInvites
-    case inviteUsers
-    case createRoom
-    case createRoomNoUsers
-    case createPoll
-    case roomPollsHistoryEmptyLoadMore
-    case roomPollsHistoryLoadMore
+    case login
```
pixlwave, line 55: Lol why is this one all the way at the bottom now?

UITests/Sources/BugReportUITests.swift
pixlwave, line 1: FWIW, this could just as well be handled by view model tests and preview tests if we wanted.

UITests/Sources/__Snapshots__/Application/appLockSetupFlow-3-iPad-10th-generation-en-GB.1.png
pixlwave, line 1: Something's gone wrong here…

`appLockSetupFlow-3-iPad-10th-generation-en-GB.1.png` and `appLockSetupFlow-3-iPad-10th-generation-en-GB.2.png` both exist and are identical. Should we even have the last index when we already have the first one?

fastlane/Fastfile
```diff
@@ -133,14 +133,14 @@ lane :integration_tests do
   clear_derived_data()
   
   create_simulator_if_necessary(
-    name: "iPhone 14 Pro",
-    type: "com.apple.CoreSimulator.SimDeviceType.iPhone-14-Pro",
-    runtime: "com.apple.CoreSimulator.SimRuntime.iOS-16-4"
+    name: "iPhone 15 Pro",
+    type: "com.apple.CoreSimulator.SimDeviceType.iPhone-15-Pro",
+    runtime: "com.apple.CoreSimulator.SimRuntime.iOS-15-2"
```
pixlwave, line 138: Errr, lets go forwards perhaps 😅
```suggestion
    runtime: "com.apple.CoreSimulator.SimRuntime.iOS-17-2"
```

pixlwave (commented):
Looking forward to this change so much 👏

.gitignore
```diff
@@ -32,18 +32,8 @@ towncrier.toml.tmp
...
 ._*
-
-## Temporary Screenshots
-# ignore all
-/UITests/Sources/__Snapshots__/Application/*
-# but keep the references
-!/UITests/Sources/__Snapshots__/Application/en-GB-iPad-9th-generation.*.png
-!/UITests/Sources/__Snapshots__/Application/pseudo-iPad-9th-generation.*.png
-!/UITests/Sources/__Snapshots__/Application/en-GB-iPhone-14.*.png
-!/UITests/Sources/__Snapshots__/Application/pseudo-iPhone-14.*.png
```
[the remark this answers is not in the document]
stefanceriu: I mean.. the fact that you're uploading new files should be notice enough. The flip side is what just happened to me where I spent half an hour trying to figure out why files weren't getting updated. Similarly, we don't do this for snapshot tests

ElementX/Sources/UITests/UITestsScreenIdentifier.swift
stefanceriu: Good question 🙈

fastlane/Fastfile
stefanceriu: 🤦‍♂️

UITests/Sources/BugReportUITests.swift
stefanceriu: Yep yep, I think we can improve a lot of the tests but for now I figured it's simpler to just delete what's not needed anymore

.gitignore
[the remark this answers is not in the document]
pixlwave: There's a big difference for snapshot tests: the simulated devices are hard coded and (I think it's still the case that) you can only run them on a specific device or it fails to run. I think that's a big enough difference that it's worth remembering to update this once a year.

pixlwave (approved):
LGTM 👏

PR #2599, opened 2024-03-22
Hard code the room flow to a single room proxy.

pixlwave:
This PR scopes the responsibility of `RoomFlowCoordinator` down to a single room. Rather than letting it change rooms, it becomes the job of `UserSessionFlowCoordinator` to replace the room flow when necessary.

### DIFF
--- a/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
@@ -44,7 +44,6 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
     private let userSession: UserSessionProtocol
     private let roomTimelineControllerFactory: RoomTimelineControllerFactoryProtocol
     private let navigationStackCoordinator: NavigationStackCoordinator
-    private let navigationSplitCoordinator: NavigationSplitCoordinator
     private let emojiProvider: EmojiProviderProtocol
     private let appSettings: AppSettings
     private let analytics: AnalyticsService
@@ -73,7 +72,6 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
          userSession: UserSessionProtocol,
          roomTimelineControllerFactory: RoomTimelineControllerFactoryProtocol,
          navigationStackCoordinator: NavigationStackCoordinator,
-         navigationSplitCoordinator: NavigationSplitCoordinator,
          emojiProvider: EmojiProviderProtocol,
          appSettings: AppSettings,
          analytics: AnalyticsService,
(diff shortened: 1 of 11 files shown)


ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -62,34 +62,39 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
...
         self.navigationStackCoordinator = navigationStackCoordinator
-        self.navigationSplitCoordinator = navigationSplitCoordinator
         self.emojiProvider = emojiProvider
         self.appSettings = appSettings
         self.analytics = analytics
         self.userIndicatorController = userIndicatorController
         self.orientationManager = orientationManager
         
+        // Maybe this should happen outside, not sure?
+        let subscriptionTask = Task { await roomProxy.subscribeForUpdates() }
```
stefanceriu, line 91: I think I better liked the previous approach when we would subscribe/unsubscribe when presenting the room instead of always. This feels dangerous, especially the deinit, we might run into race condition.

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -489,49 +448,21 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
...
-        }
-        
-        if navigationSplitCoordinator.detailCoordinator == nil {
-            navigationSplitCoordinator.setDetailCoordinator(navigationStackCoordinator, animated: animated)
+            self?.stateMachine.tryEvent(.dismissRoom)
         }
     }
     
-    private func dismissRoom(animated: Bool) {
-        // DON'T CHANGE THE ORDER IN WHICH POP AND SET ARE DONE, IT CAN CAUSE A CRASH
```
stefanceriu, line 506: Does this comment still apply but somewhere else?

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -489,49 +448,21 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
...
         timelineController = nil
         
-        actionsSubject.send(.dismissedRoom)
-    }
-    
-    private func presentRoomDetails(roomID: String, isRoot: Bool, animated: Bool) {
-        Task {
-            await asyncPresentRoomDetails(roomID: roomID, isRoot: isRoot, animated: animated)
-        }
+        actionsSubject.send(.complete)
```
stefanceriu, line 461: I wonder if `finished` is a better name for this, imperatives don't really work in this context imho.

stefanceriu (approved):
It's hard to statically analyse these changes but I love the idea and it's a great addition to the codebase 👏

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
pixlwave: Oh yeah, I thought I moved the comment into `UserSessionFlowCoordinator`, have put it in there.

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
pixlwave: Discussed and for now, we've moved the deinit to `dismissFlow`. Will discuss with SDK team about multiple subscriptions for child flows.

PR #2601, opened 2024-03-22
Cache Xcode derived data and SPM packages for unit tests

stefanceriu:


.github/workflows/unit_tests.yml
```diff
@@ -14,16 +14,20 @@ jobs:
     runs-on: macos-14
 
     concurrency:
-      # When running on develop, use the sha to allow all runs of this workflow to run concurrently.
-      # Otherwise only allow a single run of this workflow on each branch, automatically cancelling older runs.
-      group: ${{ github.ref == 'refs/heads/develop' && format('unit-tests-develop-{0}', github.sha) || format('unit-tests-{0}', github.ref) }}
+      # Only allow a single run of this workflow on each branch, automatically cancelling older runs.
+      group: ${{ format('unit-tests-{0}', github.ref) }}
```
pixlwave, line 18: Are you sure we want to do this. This will skip merges to `develop`.

stefanceriu: Yep, I think that's fine, we're more likely to just go fix it instead of reverting the commit.

pixlwave (approved):

PR #2605, opened 2024-03-25
Update the SDK.

pixlwave:
Includes https://github.com/matrix-org/matrix-rust-sdk/pull/3273 which fixes a bug where sometimes new messages in a room wouldn't be shown to the user.


stefanceriu (approved):
🚀

PR #2607, opened 2024-03-26
Fixes #2518, fixes #2590 - Implement public room search list paginati…

### LINKED ISSUE #2518: [Task] Public room search view
The should be accessed by a button in the search screen when the search is highlighted but is empty.
In the meta issue you can find the designs, however we only want to implement the search, no categorisation, sorting or additional filtering is required as of right now

### LINKED ISSUE #2590: [Task] Join a public room from the room search directory
If a room is joinable it should show a Join button (even if the whole cell should be tappable).
This will show a loader, the search view will be dismissed, and then the user will land on the room that they have joined.

stefanceriu:
…on and room joining

### DIFF
--- a/ElementX/Sources/Screens/RoomDirectorySearchScreen/RoomDirectorySearchScreenViewModel.swift
+++ b/ElementX/Sources/Screens/RoomDirectorySearchScreen/RoomDirectorySearchScreenViewModel.swift
@@ -20,7 +20,8 @@ import SwiftUI
 typealias RoomDirectorySearchScreenViewModelType = StateStoreViewModel<RoomDirectorySearchScreenViewState, RoomDirectorySearchScreenViewAction>
 
 class RoomDirectorySearchScreenViewModel: RoomDirectorySearchScreenViewModelType, RoomDirectorySearchScreenViewModelProtocol {
-    private let roomDirectorySearch: RoomDirectorySearchProxyProtocol
+    private let clientProxy: ClientProxyProtocol
+    private let roomDirectorySearchProxy: RoomDirectorySearchProxyProtocol
     private let userIndicatorController: UserIndicatorControllerProtocol
     
     private let actionsSubject: PassthroughSubject<RoomDirectorySearchScreenViewModelAction, Never> = .init()
@@ -28,16 +29,20 @@ class RoomDirectorySearchScreenViewModel: RoomDirectorySearchScreenViewModelType
         actionsSubject.eraseToAnyPublisher()
     }
     
-    init(roomDirectorySearch: RoomDirectorySearchProxyProtocol,
+    @CancellableTask
+    private var dataLoadingTask: Task<Void, Never>?
+    
+    init(clientProxy: ClientProxyProtocol,
          userIndicatorController: UserIndicatorControllerProtocol,
          imageProvider: ImageProviderProtocol) {
-        self.roomDirectorySearch = roomDirectorySearch
+        self.clientProxy = clientProxy
+        roomDirectorySearchProxy = clientProxy.roomDirectorySearchProxy()
         self.userIndicatorController = userIndicatorController
         super.init(initialViewState: RoomDirectorySearchScreenViewState(), imageProvider: imageProvider)
         
-        roomDirectorySearch.resultsPublisher
+        roomDirectorySearchProxy.resultsPublisher
             .receive(on: DispatchQueue.main)
-            .weakAssign(to: \.state.searchResults, on: self)
+            .weakAssign(to: \.state.rooms, on: self)
             .store(in: &cancellables)
         
         context.$viewState.map(\.bindings.searchString)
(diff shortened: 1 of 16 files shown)


ElementX/Sources/Screens/RoomDirectorySearchScreen/RoomDirectorySearchScreenViewModel.swift
```diff
@@ -88,4 +114,25 @@ class RoomDirectorySearchScreenViewModel: RoomDirectorySearchScreenViewModelType
...
+    
+    private func loadNextPage() {
+        dataLoadingTask = Task {
+            state.isLoading = true
+            let _ = await roomDirectorySearchProxy.nextPage()
+            state.isLoading = false
+        }
+    }
+    
+    private static let loadingIndicatorIdentifier = "\(RoomDirectorySearchScreenViewModel.self)-Loading"
```
pixlwave, line 126: Oh wait, did this land in the end? I didn't see the point of `"\(MyViewModel.self)-Loading"` when it's not dynamic and essentially equivalent to a static string.

ElementX/Sources/Screens/RoomDirectorySearchScreen/RoomDirectorySearchScreenViewModel.swift
```diff
@@ -20,24 +20,29 @@ import SwiftUI
...
     private let userIndicatorController: UserIndicatorControllerProtocol
     
     private let actionsSubject: PassthroughSubject<RoomDirectorySearchScreenViewModelAction, Never> = .init()
     var actionsPublisher: AnyPublisher<RoomDirectorySearchScreenViewModelAction, Never> {
         actionsSubject.eraseToAnyPublisher()
     }
     
-    init(roomDirectorySearch: RoomDirectorySearchProxyProtocol,
+    @CancellableTask
+    private var dataLoadingTask: Task<Void, Never>?
```
pixlwave, line 33: What's the benefit in storing the task when we never do anything to check if it's cancelled? (Mauro also decided it wasn't necessary for loading rooms given the update comes separately through an observation).

ElementX/Sources/Screens/RoomDirectorySearchScreen/View/RoomDirectoryCell.swift
```diff
@@ -20,6 +20,7 @@ import SwiftUI
 struct RoomDirectorySearchCell: View {
     let result: RoomDirectorySearchResult
     let imageProvider: ImageProviderProtocol?
+    let joinCallback: () -> Void
```
pixlwave, line 23: Only a thought:
```suggestion
    let joinAction: () -> Void
```

ElementX/Sources/Screens/RoomDirectorySearchScreen/View/RoomDirectoryCell.swift
```diff
@@ -36,9 +37,17 @@ struct RoomDirectorySearchCell: View {
     }
 
     var body: some View {
-        ListRow(label: .avatar(title: result.name ?? result.alias ?? result.id,
-                               description: description,
-                               icon: avatar), kind: .label)
+        if result.canBeJoined {
```
pixlwave, line 40: Would it be tidier to put this in the kind? E.g. a single row with
```
kind: result.canBeJoined ? .navigationLink(join) : .label)
```

ElementX/Sources/Screens/RoomDirectorySearchScreen/View/RoomDirectorySearchScreen.swift
```diff
@@ -46,25 +60,33 @@ struct RoomDirectorySearchScreen: View {
...
-              canBeJoined: false)])),
-    userIndicatorController: UserIndicatorControllerMock(),
-    imageProvider: MockMediaProvider())
+    static let viewModel: RoomDirectorySearchScreenViewModel = {
+        let results = [RoomDirectorySearchResult(id: "test_1",
+                                                 alias: "#test_1:example.com",
+                                                 name: "Test 1",
+                                                 topic: "Test description 1",
+                                                 avatarURL: nil,
+                                                 canBeJoined: true)]
```
pixlwave, line 80: How come we dropped down to only 1 result?

pixlwave (approved):
LGTM 👍

ElementX/Sources/Screens/RoomDirectorySearchScreen/RoomDirectorySearchScreenViewModel.swift
stefanceriu: Yep, because I said something along the lines of: at least this way you can rename the file without worrying about indicator identifiers

ElementX/Sources/Screens/RoomDirectorySearchScreen/RoomDirectorySearchScreenViewModel.swift
stefanceriu: There is still a chance it gets cancelled before it starts and never gets enqueued, no?

ElementX/Sources/Screens/RoomDirectorySearchScreen/View/RoomDirectorySearchScreen.swift
stefanceriu: I think my eyes just missed it when refactoring the previous indentation hell

ElementX/Sources/Screens/RoomDirectorySearchScreen/RoomDirectorySearchScreenViewModel.swift
pixlwave: Ahhh in a refactor, got it 👍

ElementX/Sources/Screens/RoomDirectorySearchScreen/View/RoomDirectoryCell.swift
stefanceriu: Definitely, at least for now or until they diverge too much 👍

PR #2608, opened 2024-03-26
Fix various bugs in the moderation feature.

pixlwave:
Fixes the following bugs from https://github.com/element-hq/element-meta/issues/2369

- Use a consistent string in all member list search fields.
- Show your own user as an Admin when changing roles.
- Also show invited users when changing roles.
- Don't allow admin's to kick/ban other admins (or themselves).
- Show admin/mod/user sections in the edit roles screen.
- Show pending users in the edit roles screen.
- Show selection for disabled members in the edit roles screen.

Additionally fixes a bug where left members were being counted as admins/moderators.


UnitTests/Sources/RoomMembersListScreenViewModelTests.swift
stefanceriu, line 1: Very nice 🙌

stefanceriu (approved):
Looks sensible to me 👍

pixlwave:
In fixing the snapshots I ended up including the final part of the changes (Show role of user & show 'not selectable' checkbox) in this PR.

stefanceriu (approved):
lgtm 👍

PR #2620, opened 2024-04-01
Rust permalinks

stefanceriu:
This PR is the first step towards adopting RustSDK permalink builders. Requires https://github.com/matrix-org/matrix-rust-sdk/pull/3289

### DIFF
--- a/ElementX/Sources/Services/Room/RoomProxy.swift
+++ b/ElementX/Sources/Services/Room/RoomProxy.swift
@@ -528,6 +528,40 @@ class RoomProxy: RoomProxyProtocol {
     func elementCallWidgetDriver() -> ElementCallWidgetDriverProtocol {
         ElementCallWidgetDriver(room: room)
     }
+    
+    // MARK: - Permalinks
+    
+    func matrixToPermalink() async -> Result<URL, RoomProxyError> {
+        do {
+            let urlString = try await room.matrixToPermalink()
+            
+            guard let url = URL(string: urlString) else {
+                MXLog.error("Invalid permalink URL string: \(urlString)")
+                return .failure(.generic("Invalid permalink URL string: \(urlString)"))
+            }
+            
+            return .success(url)
+        } catch {
+            MXLog.error("Failed creating permalink for roomID: \(id) with error: \(error)")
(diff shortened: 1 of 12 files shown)


ElementX/Sources/Services/Room/RoomProxyProtocol.swift
```diff
@@ -19,6 +19,7 @@ import Foundation
 import MatrixRustSDK
 
 enum RoomProxyError: Error, Equatable {
+    case generic(String?)
```
pixlwave, line 22: 2 thoughts:
- Wouldn't it be better to have a nested enum here of say `.rustFailure` and `.invalidURL`.
- But also, what's the point of the associated value here anyway, seeing as it isn't used?

UnitTests/Sources/PermalinkBuilderTests.swift
```diff
@@ -57,6 +57,17 @@ class PermalinkBuilderTests: XCTestCase {
         }
     }
     
+    func testMautrixBridgePermalink() throws {
```
pixlwave, line 60: Nice 👍

ElementX/Sources/Services/Room/RoomProxy.swift
```diff
@@ -528,6 +528,40 @@ class RoomProxy: RoomProxyProtocol {
     func elementCallWidgetDriver() -> ElementCallWidgetDriverProtocol {
         ElementCallWidgetDriver(room: room)
     }
+    
+    // MARK: - Permalinks
+    
+    func matrixToPermalink() async -> Result<URL, RoomProxyError> {
```
pixlwave, line 534: Do you think we should be replacing the host in here with the permalink base URL?

stefanceriu: Talked about it offline, we will handle that through the SDK later

ElementX/Sources/Services/Room/RoomProxyProtocol.swift
stefanceriu: Talked about it offline, we'll handle this in a separate PR and clean up all existing errors.

PR #2626, opened 2024-04-02
Use the new RTE package repo.

pixlwave:
Now that https://github.com/matrix-org/matrix-rich-text-editor/pull/954 is merged, this PR uses [matrix-org/matrix-rich-text-editor-swift](https://github.com/matrix-org/matrix-rich-text-editor-swift) removing the old dependency which was a rather large download at this point.


stefanceriu (approved):
🥳

PR #2628, opened 2024-04-03
Fixes #2624, fixes #2625 - Add user suggestions in the create room flows.

### LINKED ISSUE #2624: [Task] Show recently accessed users as potential invitees when creating a new room

### LINKED ISSUE #2625: [Task] Show last visited rooms in the "start chat" screen

stefanceriu:
Requires https://github.com/matrix-org/matrix-rust-sdk/pull/3293

https://github.com/element-hq/element-x-ios/assets/637564/a705dfb5-e058-4067-aa89-9c6f3c7e58c5

### DIFF
--- a/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
@@ -1009,11 +1009,12 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
         let selectedUsersSubject: CurrentValueSubject<[UserProfileProxy], Never> = .init([])
         
         let stackCoordinator = NavigationStackCoordinator()
-        let inviteParameters = InviteUsersScreenCoordinatorParameters(selectedUsers: .init(selectedUsersSubject),
-                                                                      roomType: .room(roomProxy: roomProxy),
+        let inviteParameters = InviteUsersScreenCoordinatorParameters(clientProxy: userSession.clientProxy,
                                                                       mediaProvider: userSession.mediaProvider,
                                                                       userDiscoveryService: UserDiscoveryService(clientProxy: userSession.clientProxy),
-                                                                      userIndicatorController: userIndicatorController)
+                                                                      userIndicatorController: userIndicatorController,
+                                                                      selectedUsers: .init(selectedUsersSubject),
+                                                                      roomType: .room(roomProxy: roomProxy))
         
         let coordinator = InviteUsersScreenCoordinator(parameters: inviteParameters)
         stackCoordinator.setRootCoordinator(coordinator)
--- a/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
@@ -460,6 +460,10 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
         if case .invitesScreen = stateMachine.state, availableInvitesCount == 1 {
             dismissInvitesList(animated: true)
         }
+        
+        Task {
+            await userSession.clientProxy.trackRecentlyVisitedRoom(roomID)
+        }
     }
     
     private func tearDownRoomFlow(animated: Bool) {
--- a/ElementX/Sources/Services/Client/ClientProxy.swift
+++ b/ElementX/Sources/Services/Client/ClientProxy.swift
@@ -574,7 +574,7 @@ class ClientProxy: ClientProxyProtocol {
         RoomDirectorySearchProxy(roomDirectorySearch: client.roomDirectorySearch())
     }
     
-    // MARK: - Ignored users
+    // MARK: Ignored users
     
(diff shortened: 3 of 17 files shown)


ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -459,6 +459,10 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
         if case .invitesScreen = stateMachine.state, availableInvitesCount == 1 {
             dismissInvitesList(animated: true)
         }
+        
+        Task {
+            await userSession.clientProxy.trackRecentlyVisitedRoom(roomID)
```
pixlwave, line 464: Should this happen when `showingRoomDetails` is true? Might be fine, just double-checking.

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -577,7 +577,58 @@ class ClientProxy: ClientProxyProtocol {
         }
     }
     
-    // MARK: Private
+    // MARK: Recently visited rooms
+    
+    func trackRecentlyVisitedRoom(_ roomID: String) async -> Result<Void, ClientProxyError> {
```
pixlwave, line 582: Just a thought but should we even bother with the `Result` here? I don't think there's anything useful that could be done from the callers perspective so logging the error and returning void seems fine to me?

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -577,7 +577,58 @@ class ClientProxy: ClientProxyProtocol {
...
+        var users = [UserProfileProxy]()
+        
+        for roomID in roomIdentifiers {
+            guard let room = await roomForIdentifier(roomID),
+                  room.isDirect,
+                  let members = await room.members() else {
+                continue
+            }
+            
+            for member in members where member.isActive && member.userID != userID {
```
pixlwave, line 618: Should we add a `!users.contains { $0.userID == member.userID }` to this too? Or make `users` a `Set` (although maybe an `OrderedDictionary` would be better as you could still end up with the same user with different profiles otherwise)

ElementX/Sources/Services/Client/ClientProxyProtocol.swift
```diff
@@ -38,6 +38,7 @@ enum ClientProxyLoadingState {
 }
 
 enum ClientProxyError: Error {
+    case generic(Error?)
```
pixlwave, line 41: Ok passing the error around is nicer than strings, but maybe it should be a `ClientError`?

pixlwave (approved):
Nice! 🤩

ElementX/Sources/Services/Client/ClientProxyProtocol.swift
stefanceriu: Yeah, let me handle all of that together with the room proxy on a different PR

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -985,11 +985,12 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
...
         let stackCoordinator = NavigationStackCoordinator()
-        let inviteParameters = InviteUsersScreenCoordinatorParameters(selectedUsers: .init(selectedUsersSubject),
-                                                                      roomType: .room(roomProxy: roomProxy),
+        let inviteParameters = InviteUsersScreenCoordinatorParameters(clientProxy: userSession.clientProxy,
                                                                       mediaProvider: userSession.mediaProvider,
                                                                       userDiscoveryService: UserDiscoveryService(clientProxy: userSession.clientProxy),
-                                                                      userIndicatorController: userIndicatorController)
+                                                                      userIndicatorController: userIndicatorController,
+                                                                      selectedUsers: .init(selectedUsersSubject),
+                                                                      roomType: .room(roomProxy: roomProxy))
```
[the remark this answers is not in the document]
stefanceriu: Not me, obviously 😁 I like to order them by overall project importance, global vs local, public vs private etc.

ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu: The user is just holding it wrong if that happens but seriously, yes, I'll clean it up 👍

ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu: I'm treating the ClientProxy more as an API, which shouldn't be too opinionated. The result isn't particularly interesting in the current use case but it might be in the future so I would prefer to keep this generic.

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
stefanceriu: I don't think anybody thought about that but I'm inclined to keep it like it is just because it's simpler.

PR #2642, opened 2024-04-03
Attempt to fix the test on CI

stefanceriu:


PreviewTests/__Snapshots__/PreviewTests/test_roomDetailsScreen-iPhone-15-en-GB.Simple-Room.png
pixlwave, line 1: Should we be concerned that the iPad has a Share button in this snapshot but the iPhone does not?

pixlwave (approved):
LGTM 👍

PR #2646, opened 2024-04-04
Cleanup proxy errors

stefanceriu:
- proxy errors have been getting repetitive and not particularly useful
- differentiate between sdk and client errors and keep what provides value
- add error logs everywhere a failure occurs

### DIFF
--- a/ElementX/Sources/Services/Room/RoomProxyProtocol.swift
+++ b/ElementX/Sources/Services/Room/RoomProxyProtocol.swift
@@ -18,26 +18,11 @@ import Combine
 import Foundation
 import MatrixRustSDK
 
-enum RoomProxyError: Error, Equatable {
-    case generic(String?)
-    case failedRedactingEvent
-    case failedReportingContent
-    case failedRetrievingMember
-    case failedLeavingRoom
-    case failedAcceptingInvite
-    case failedRejectingInvite
-    case failedInvitingUser
-    case failedSettingRoomName
-    case failedSettingRoomTopic
-    case failedRemovingAvatar
-    case failedUploadingAvatar
-    case failedCheckingPermission
-    case failedSettingPermission
-    case failedFlaggingAsUnread
-    case failedMarkingAsRead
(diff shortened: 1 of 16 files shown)


ElementX/Sources/Services/Room/RoomProxyProtocol.swift
```diff
@@ -19,25 +19,21 @@ import Foundation
...
-    case failedFlaggingAsFavourite
-    case failedModeration
+    case sdkError(Error)
+    
+    case invalidURL
+    case invalidMedia
+    
+    static func == (lhs: RoomProxyError, rhs: RoomProxyError) -> Bool {
+        switch (lhs, rhs) {
+        case (.sdkError, .sdkError),
```
pixlwave, line 29: Errrr, depending on what this conformance is for, it might be worth making sure both of the embedded errors have the same `description` here?

pixlwave (approved):
Nice, this is a great tidy up 👏

ElementX/Sources/Services/Room/RoomProxyProtocol.swift
stefanceriu: Hmm.. that's a good idea. I'm entirely sure what this is about but the new power level permission checks fail if this is not equatable. I was expecting it to know better than that

`state.canEditRoomName = await roomProxy.canUser(userID: roomProxy.ownUserID, sendStateEvent: .roomName) == .success(true)` -> `Operator function '==' requires that 'RoomProxyError' conform to 'Equatable'`

pixlwave: Meh, then maybe we remove the equatable here and use 

```
await try? roomProxy.canUser(userID: roomProxy.ownUserID, sendStateEvent: .roomName).get() == true
```

instead?

stefanceriu: I honestly have no strong feelings, but what that might be able to do is help with our preview problems 🤔

PR #2654, opened 2024-04-05
Update appearance of admins when changing moderators.

pixlwave:
Another tweak for #2548. Includes an update to Compound, but is missing a change for the appearance of pending users.

roomRolesAndPermissionsFlow-2-iPhone-15-en-GB UI


stefanceriu (approved):
lgtm 👍

PR #2666, opened 2024-04-08
Allow showing invites in the room list

stefanceriu:
Expose invites directly on the room list as opposed to a completely different screen
* similar UI to the existing invites screen
* can action them directly from the room list
* they will move down in the room list based on account activity
* original screen still available
* hidden behind feature flag, **cache reset is necessary** after disabling the feature flag as the invites will be stored in the sliding sync cold cache

Screenshot 2024-04-09 at 10 14 29


ElementX/Sources/Services/UserSession/UserSessionStore.swift
```diff
@@ -135,11 +135,8 @@ class UserSessionStore: UserSessionStoreProtocol {
         let completeBuilder = builder
         
         do {
-            let client: Client = try await Task.dispatch(on: .global()) {
-                let client = try completeBuilder.build()
-                try client.restoreSession(session: credentials.restorationToken.session)
-                return client
-            }
+            let client = try await completeBuilder.build()
+            try client.restoreSession(session: credentials.restorationToken.session)
```
pixlwave, line 139: This one still needs that `Task.dispatch` as it blocks.

ElementX/Sources/Screens/HomeScreen/View/HomeScreenInviteCell.swift
```diff
@@ -0,0 +1,156 @@
...
+    
+    // MARK: - Private
+
+    private var mainContent: some View {
+        VStack(alignment: .leading, spacing: 0) {
+            HStack(alignment: .firstTextBaseline, spacing: 16) {
+                textualContent
+                    .padding(.trailing, room.badges.isDotShown ? 0 : 16)
+                
+                if room.badges.isDotShown {
```
pixlwave, line 53: Does the badge actually work here in the same way that the old one was used? Wondering if we should simply remove it instead 🤔

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
```diff
@@ -130,6 +133,18 @@ struct HomeScreenViewStateBindings {
 }
 
 struct HomeScreenRoom: Identifiable, Equatable {
+    enum RoomType {
+        case placeholder
+        case room
+        case invite
+    }
```
pixlwave, line 140: I kinda wonder if at this stage, we should be putting associated values for e.g. inviter and others that are unique for a single type in here.

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -438,4 +442,73 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
             }
         }
     }
+    
+    // MARK: Invites
+    
+    private func acceptInvite(roomID: String) {
```
pixlwave, line 448: Could we make this `async` and put the `Task` at the call site?

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -438,4 +442,73 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
...
+            case .success:
+                actionsSubject.send(.presentRoom(roomIdentifier: roomID))
+                analyticsService.trackJoinedRoom(isDM: roomProxy.isDirect, isSpace: roomProxy.isSpace, activeMemberCount: UInt(roomProxy.activeMembersCount))
+            case .failure:
+                displayError()
+            }
+        }
+    }
+    
+    private func maybeDeclineInvite(roomID: String) {
```
pixlwave, line 471: Maybe this should be called `showDeclineInviteConfirmation` or similar?

pixlwave (approved):
Exciting! 👏

ElementX/Sources/Screens/HomeScreen/View/Filters/RoomListFilterModels.swift
```diff
@@ -29,6 +29,15 @@ enum RoomListFilter: Int, CaseIterable, Identifiable {
     case people
     case rooms
     case favourites
+    case invites
+    
+    static var availableFilters: [RoomListFilter] {
+        if ServiceLocator.shared.settings.roomListInvitesEnabled {
```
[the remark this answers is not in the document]
pixlwave: Actually, clearly that won't make a difference now I've seen the whole thing. Still, it should probably be set from above via injected settings rather than using the `ServiceLocator`?

ElementX/Sources/Services/UserSession/UserSessionStore.swift
stefanceriu: Oh right, good catch 👍

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryDetails.swift
```diff
@@ -19,6 +19,10 @@ import MatrixRustSDK
 
 struct RoomSummaryDetails {
     let id: String
+    
+    let isInvite: Bool
```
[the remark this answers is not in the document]
stefanceriu: Could give it a try, not sure it would change much tbh

ElementX/Sources/Screens/HomeScreen/View/HomeScreenInviteCell.swift
stefanceriu: Based on the talk we had this morning apparently it's supposed to always be shown ¯\_(ツ)_/¯ I'll just change the cell to always do that

ElementX/Sources/Screens/HomeScreen/View/HomeScreenInviteCell.swift
```diff
@@ -0,0 +1,156 @@
...
+        }
+        return attributedString
+    }
+    
+    private var badge: some View {
+        Circle()
+            .scaledFrame(size: 12)
+            .foregroundColor(.compound.iconAccentTertiary)
+    }
+}
```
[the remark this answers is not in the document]
stefanceriu: Oups, I promise I wanted to but then things broke when I copy pasted and forgot all about it 😁

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
stefanceriu: it's probably more flexible like we have it now..

ElementX/Sources/Screens/HomeScreen/View/Filters/RoomListFilterModels.swift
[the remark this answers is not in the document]
pixlwave: Oh dear, totally missed that one 🤦‍♂️

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
stefanceriu: I do agree with you that this might become problematic in the future, but I think I would just go with different model types directly e.g. `.room(RoomDetails)` and  `.invite(InviteDetails)`. Let's see how this feature plays out and then we can do a bit of refactoring

pixlwave: Makes sense to me 👍

PR #2667, opened 2024-04-08
QR Code Login Initial view state

Velin92:
First state of the qr code login view.


ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
```diff
@@ -103,6 +102,21 @@ class AuthenticationFlowCoordinator: FlowCoordinatorProtocol {
         navigationRootCoordinator.setRootCoordinator(navigationStackCoordinator)
     }
     
+    private func startQrCodeLogin() {
```
pixlwave, line 105: Only for consistency.
```suggestion
    private func startQRCodeLogin() {
```

ElementX/Sources/Other/SwiftUI/Views/SFNumberedListView.swift
```diff
@@ -0,0 +1,81 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import SFSafeSymbols
+import SwiftUI
+
+/// The view can only display a max 9 items as of right now
+struct SFNumberedListView: View {
```
pixlwave, line 21: Nice! Could switch it to use `systemImage: String` for the max supported value, although hopefully we'll never have a screen with instructions that are 9 steps long 🙈

ElementX/Sources/Screens/QRCodeLoginScreen/QRCodeLoginScreenViewModel.swift
```diff
@@ -0,0 +1,61 @@
...
+        switch viewAction {
+        case .cancel:
+            actionsSubject.send(.cancel)
+        case .startScan:
+            Task { await startScanIfPossible() }
+        }
+    }
+    
+    private func startScanIfPossible() async {
+        let status = AVCaptureDevice.authorizationStatus(for: .video)
```
pixlwave, line 48: Should this method be using the injected controller?

ElementX/Sources/Services/QRCode/QRCodeLoginController.swift
```diff
@@ -0,0 +1,37 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import AVFoundation
+
+final class QRCodeLoginController: QRCodeLoginControllerProtocol {
```
pixlwave, line 19: Maybe we should call this the `QRCodeLoginService` assuming it will provide more functionality going forwards? That would be more inline with the rest of the project.

pixlwave (approved):
LGTM 👏

ElementX/Sources/Other/SwiftUI/Views/SFNumberedListView.swift
Velin92: I think for now is okay to leave to display 0

PR #2669, opened 2024-04-09
Fix child presentation of the same room.

pixlwave:
Make sure we don't stack the same room directly underneath itself (but allow it with intermediate rooms)


stefanceriu (approved):
Noice! 👌

PR #2676, opened 2024-04-10
Attempt to improve unit test reliability:

stefanceriu:
- synchronize mock properties on the main queue
- sleep for longer when processing routes
- finish encoding before deallocating the keyed archiver
- wait for NotificationCenter notifications to be delivered
- change were we process remote notification permissions


pixlwave (approved):
Easy when you know how! 👏👏👏

PR #2677, opened 2024-04-10
Update strings and add productionAppName value for the new placeholder.

pixlwave:
Syncs the strings and adds a common place for the name `Element` to be configured.


Velin92 (approved):

stefanceriu (approved):
🚀

PR #2678, opened 2024-04-10
QR Code error views

Velin92:
error state views for the QR Code Login screen.

Review after merging: https://github.com/element-hq/element-x-ios/pull/2677


ElementX/Sources/Services/Background/ApplicationProtocol.swift
```diff
@@ -34,4 +36,11 @@ extension UIApplication: ApplicationProtocol {
     func open(_ url: URL) {
         open(url, options: [:], completionHandler: nil)
     }
+    
+    func openAppSettings() {
```
pixlwave, line 40: Smart! Should we adopt this everywhere we use `openSettingsURLString`?

pixlwave (approved):
👌

ElementX/Sources/Services/Background/ApplicationProtocol.swift
Velin92: oh yeah definitely! will do that!

Issue #2634, opened 2024-04-03
[Task] Click on user permalink


bmarty:
Please see and discuss the details in the meta issue.

pixlwave:
Completed by #2687 and #2690

PR #2687, opened 2024-04-11
Add a UserProfileScreen for profiles of non-members.

pixlwave:
The PR is the first part of #2634. It makes the following changes:
- Adds a `UserProfileScreen` which is a cut down version of the `RoomMemberDetailsScreen`.
- Replaces the member details screen in the stack with the user profile screen when the user can't be found in the room.

Still required to close the task:
- Handle external links with a different route and present the user profile screen directly in the user session flow.

### DIFF
--- a/ElementX/Sources/Other/SwiftUI/Views/AvatarHeaderView.swift
+++ b/ElementX/Sources/Other/SwiftUI/Views/AvatarHeaderView.swift
@@ -58,6 +58,22 @@ struct AvatarHeaderView<Footer: View>: View {
         self.onAvatarTap = onAvatarTap
         self.footer = footer
     }
+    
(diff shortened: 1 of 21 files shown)


ElementX/Sources/Other/SwiftUI/Views/AvatarHeaderView.swift
```diff
@@ -58,6 +58,22 @@ struct AvatarHeaderView<Footer: View>: View {
         self.onAvatarTap = onAvatarTap
         self.footer = footer
     }
+    
+    init(user: UserProfileProxy,
```
stefanceriu, line 62: 👌

stefanceriu (approved):
Very nice, 🚀  it!

PR #2689, opened 2024-04-12
Permalinks additions

stefanceriu:
This PR adopts the newly added SDK APIs for detecting Matrix Entities and building user permalinks:
- it refactors the AttributedString builder and how it detects plain test matrix entities
- it removes the base permalink URL from throughout the application
- it also removes all entity regexes apart from User which is still used (should be moved to the SDK side at some point)


ElementX.xcodeproj/project.pbxproj
```diff
@@ -535,11 +534,9 @@
 		7F61F9ACD5EC9E845EF3EFBF /* BugReportServiceTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = EFFD3200F9960D4996159F10 /* BugReportServiceTests.swift */; };
 		7F7EA51A9A43125A8CB6AC90 /* NotificationSettingsScreenViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 46D560DDA3B20C82766ACFAD /* NotificationSettingsScreenViewModel.swift */; };
 		7F941B063C94E1718DFC2CF3 /* RoomChangeRolesScreenRow.swift in Sources */ = {isa = PBXBuildFile; fileRef = 23E6EB7960BC9D0F7396B3BD /* RoomChangeRolesScreenRow.swift */; };
-		7FED77802940EA7DF4D0D3A2 /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = 36DA824791172B9821EACBED /* PrivacyInfo.xcprivacy */; };
```
pixlwave, line 538: `brew update && brew upgrade` pls 🥺

ElementX/Sources/Other/HTMLParsing/ElementXAttributeScope.swift
```diff
@@ -36,19 +36,29 @@ enum RoomAliasAttribute: AttributedStringKey {
...
-    let roomID: String
+enum EventOnRoomIDAttribute: AttributedStringKey {
+    typealias Value = EventOnRoomIDAttributeValue
+    static var name = "MXEventOnRoomIDAttribute"
+}
+
+struct EventOnRoomAliasAttributeValue: Hashable {
+    let alias: String
     let eventID: String
 }
```
pixlwave, line 52: If this is defined inside of `EventOnRoomAliasAttribute` as `struct Value`, can you then remove the `typealias`? If so, same comment for the room ID one too.

pixlwave (approved):
LGTM 👍

ElementX/Sources/Other/HTMLParsing/ElementXAttributeScope.swift
stefanceriu: Yep yep, seems to work 👍

PR #2691, opened 2024-04-12
Stop relying on the NotificationCenter for removing delivered push notifications

stefanceriu:
- this will improve the architecture but more importantly will remove some very flakey unit tests
- didn't bother writing a test for the invites one as we're removing that screen soon


UnitTests/Sources/UserSessionFlowCoordinatorTests.swift
```diff
@@ -81,6 +85,8 @@ class UserSessionFlowCoordinatorTests: XCTestCase {
         try await process(route: .roomList, expectedState: .roomList(selectedRoomID: nil))
         XCTAssertNil(detailNavigationStack?.rootCoordinator)
         XCTAssertNil(detailCoordinator)
+        
+        XCTAssertEqual(notificationManager.removeDeliveredMessageNotificationsForReceivedInvocations, ["1", "1", "2"])
```
pixlwave, line 89: 👌

pixlwave (approved):
LGTM!

PR #2694, opened 2024-04-15
Replace deprecated danger-swift-with-swiftlint docker image

stefanceriu:


.github/workflows/danger.yml
```diff
@@ -5,17 +5,18 @@ on: [pull_request]
 jobs:
   build:
     name: Danger
-    runs-on: ubuntu-latest
+    runs-on: macos-14
 
     steps:
-      - uses: actions/checkout@v3
+      - name: Checkout
+        uses: actions/checkout@v1
```
pixlwave, line 12: I doubt we want to downgrade this?
```suggestion
        uses: actions/checkout@v3
```

pixlwave (approved):
👏

.github/workflows/danger.yml
stefanceriu: Yup, sorry, copy pasted the wrong thing

PR #2696, opened 2024-04-15
Bring back the original native `MessageComposerTextField`, use it whe…

stefanceriu:
…n the Rich Text Editor one is disabled (advanced settings)

- this basically reverts parts of https://github.com/element-hq/element-x-ios/pull/1703


ElementX/Sources/Screens/ComposerToolbar/View/MessageComposerTextField.swift
```diff
@@ -0,0 +1,246 @@
...
+        textView.textColor = .compound.textPrimary
+        textView.isEditable = true
+        textView.font = font
+        textView.isSelectable = true
+        textView.isUserInteractionEnabled = true
+        textView.backgroundColor = UIColor.clear
+        textView.returnKeyType = .default
+        textView.textContainer.lineFragmentPadding = 0.0
+        textView.textContainerInset = .zero
+        textView.keyboardType = .default
```
Velin92, line 74: for the init of this I suggest checking the RTE or the messagetextview, if in the future we want to support also pills in it. Since they require TextKit v1 which requires a slightly different initialisation

stefanceriu: Nice tip, thanks! 👍

PR #2698, opened 2024-04-16
Expose paths for focusing replied-to timeline items by tapping on an in-reply-to message bubble

stefanceriu:


ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
```diff
@@ -220,16 +220,22 @@ struct TimelineItemBubbledStylerView<Content: View>: View {
...
                     // the infinite width from increasing the overall width of the view.
-                    TimelineReplyView(placement: .timeline, timelineItemReplyDetails: replyDetails)
-                        .fixedSize(horizontal: false, vertical: true)
-                        .padding(4.0)
-                        .frame(maxWidth: .infinity, alignment: .leading)
-                        .background(Color.compound.bgCanvasDefault)
-                        .cornerRadius(8)
-                        .layoutPriority(TimelineBubbleLayout.Priority.visibleQuote)
+                    
+                    Button {
```
pixlwave, line 228: Did you check if this is exposed in voiceover? I have a feeling it'll be merged with the rest of the message so we might need to add an explicit accessibility action (or however we exposed the context menu on timeline items).

pixlwave (approved):
Very nice, thank you 🙏

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
```diff
@@ -220,16 +220,22 @@ struct TimelineItemBubbledStylerView<Content: View>: View {
...
+                        context.send(viewAction: .focusOnItem(itemID: timelineItem.id))
```
pixlwave, line 229: Waiiit a second. Shouldn't this focus on the replyDetails event ID 😉

Which I then think means we can't do a focusOnItem as we don't have enough information (the stable ID) to create a `TimelineItemIdentifier`.

pixlwave (changes requested):
Just had a thought whilst working on the permalink side of things and came back to notice the below:

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemPlainStylerView.swift
```diff
@@ -52,7 +52,12 @@ struct TimelineItemPlainStylerView<Content: View>: View {
                         Rectangle()
                             .foregroundColor(.compound.iconTertiary)
                             .frame(width: 4.0)
-                        TimelineReplyView(placement: .timeline, timelineItemReplyDetails: replyDetails)
+                        
+                        Button {
+                            context.send(viewAction: .focusOnItem(itemID: timelineItem.id))
```
pixlwave, line 57: (Same issue here too)

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
stefanceriu: Handled, it was overriding everything so we switched it to a tap gesture

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemBubbledStylerView.swift
stefanceriu: Of course 🤦, I took care of it 👍

ElementX/Sources/Screens/RoomScreen/RoomScreenInteractionHandler.swift
```diff
@@ -230,8 +230,12 @@ class RoomScreenInteractionHandler {
                 }
             }
         case .reply:
+            guard let eventId = eventTimelineItem.id.eventID else {
```
pixlwave, line 233: > [Acronyms and initialisms](https://en.wikipedia.org/wiki/Acronym) that commonly appear as all upper case in American English should be uniformly up- or down-cased according to case conventions:

😉😉

```suggestion
            guard let eventID = eventTimelineItem.id.eventID else {
```

pixlwave (approved):
LGTM 👌

PR #2709, opened 2024-04-18
Membership aware room routes

stefanceriu:
Handle room permalinks based on their membership state and show the according room joining screens:
- modify the room flow coordinator to only take a roomID and be able to decide its own room flows (especially for unjoined rooms)
- add support for the new room preview API
- have RoomDirectorySearch go through the same paths

- randomly decided to remove the invites button when room list invites are enabled in this PR too, you're welcome 😂  

Requires https://github.com/matrix-org/matrix-rust-sdk/pull/3330


UnitTests/Sources/RoomFlowCoordinatorTests.swift
```diff
@@ -196,7 +280,14 @@ class RoomFlowCoordinatorTests: XCTestCase {
         navigationStackCoordinator = NavigationStackCoordinator()
         navigationSplitCoordinator.setDetailCoordinator(navigationStackCoordinator)
         
-        roomFlowCoordinator = await RoomFlowCoordinator(roomProxy: RoomProxyMock(with: .init(id: "1")),
+        let roomId = switch roomType {
```
pixlwave, line 283:
```suggestion
        let roomID = switch roomType {
```

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -51,6 +51,11 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
     private let userIndicatorController: UserIndicatorControllerProtocol
     private let orientationManager: OrientationManagerProtocol
     
+    private var roomProxy: RoomProxyProtocol!
+    
+    // periphery:ignore - used to avoid deallocation
```
pixlwave, line 56: Do you need the ignore on this given you read it to check if it is `nil`?

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -51,6 +51,11 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
     private let userIndicatorController: UserIndicatorControllerProtocol
     private let orientationManager: OrientationManagerProtocol
     
+    private var roomProxy: RoomProxyProtocol!
```
pixlwave, line 54: I think my only question about the changes here, is whether this should be an optional and then we go back to the following at the top of each method that uses it.
```
guard let roomProxy else { fatalError("Missing room proxy") }
```

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -107,49 +107,98 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
...
+            Task {
+                guard let roomProxy = await userSession.clientProxy.roomForIdentifier(roomID) else {
+                    return
+                }
+                
+                guard roomProxy.membership == .joined else {
+                    fatalError("Requesting room details for an unjoined room")
+                }
+                
+                await storeAndSubscribeToRoomProxy(roomProxy)
```
pixlwave, line 133: - If think if the room proxy already exists we should bother with this.
- It would be nice to add a comment here just to remind us that we need it when starting the flow with room details otherwise it'll be nil.

ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenViewModel.swift
```diff
@@ -47,21 +55,93 @@ class JoinRoomScreenViewModel: JoinRoomScreenViewModelType, JoinRoomScreenViewMo
...
+        switch await clientProxy.roomPreviewForIdentifier(roomID) {
+        case .success(let roomDetails):
+            state.roomDetails = roomDetails
+        case .failure:
+            userIndicatorController.submitIndicator(UserIndicator(title: L10n.errorUnknown))
+        }
+    }
+    
     private func joinRoom() async {
-        state.isJoining = true
```
pixlwave, line 59: I have to admit, given this screen isn't particularly interactive, I imagined using this to disable the buttons (and maybe put a loading spinner in the join/accept button and then not using an indicator here.

ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenViewModel.swift
```diff
@@ -47,21 +55,93 @@ class JoinRoomScreenViewModel: JoinRoomScreenViewModelType, JoinRoomScreenViewMo
...
     }
     
+    func stop() {
+        hideLoadingIndicator()
+    }
+    
     // MARK: - Private
     
+    private func loadRoomDetails() async {
+        showLoadingIndicator()
```
pixlwave, line 71: Might be nicer to show a toast for this one rather than a modal, wdyt?

ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenViewModel.swift
```diff
@@ -47,21 +55,93 @@ class JoinRoomScreenViewModel: JoinRoomScreenViewModelType, JoinRoomScreenViewMo
...
+                                         primaryButton: .init(title: L10n.actionCancel, role: .cancel, action: nil),
+                                         secondaryButton: .init(title: L10n.actionDecline, role: .destructive, action: { Task { await self.declineInvite() } }))
+    }
+    
+    private func declineInvite() async {
+        defer {
+            userIndicatorController.retractIndicatorWithId(roomID)
+        }
+        
+        userIndicatorController.submitIndicator(UserIndicator(id: roomID, type: .modal, title: L10n.commonLoading, persistent: true))
```
pixlwave, line 120: Should probably say "Declining" rather than loading.

pixlwave (approved):
Wooop, I love how this turned out. Thank you for tackling this one 🙏

ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenModels.swift
```diff
@@ -21,34 +21,43 @@ enum JoinRoomScreenViewModelAction {
...
+        
+        if roomDetails.isPublic {
+            return .join
+        }
+        
+        if roomDetails.canKnock {
+            return .knock
+        }
+        
+        return .unknown
```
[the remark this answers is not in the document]
stefanceriu: Guess we need to talk that over with the product team. Let's tackle it separately from this PR

PreviewTests/__Snapshots__/PreviewTests/test_timelineItemPlainStylerView-iPad-en-GB.Replies.png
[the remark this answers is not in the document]
stefanceriu: Eh.. we didn't have iPads when we built that :))

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
stefanceriu: Good point 👍

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -474,17 +528,69 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
...
+                    Task { [weak self] in
+                        guard let self else { return }
+                        
+                        if let roomProxy = await userSession.clientProxy.roomForIdentifier(roomID) {
+                            await storeAndSubscribeToRoomProxy(roomProxy)
+                            stateMachine.tryEvent(.presentRoom, userInfo: EventUserInfo(animated: animated))
+                            
+                            analytics.trackJoinedRoom(isDM: roomProxy.isDirect, isSpace: roomProxy.isSpace, activeMemberCount: UInt(roomProxy.activeMembersCount))
+                        } else {
+                            stateMachine.tryEvent(.dismissFlow, userInfo: EventUserInfo(animated: animated))
```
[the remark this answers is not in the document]
stefanceriu: If you fail to load room (even though you just joined it i.e. 30 second timeout) this will dismiss the whole flow (the join room screen in this case) and take you back to wherever you were, because there's really nothing else we can do.

ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenViewModel.swift
stefanceriu: Copied these from the existing invites screen, not the right time to tackle that.

ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenViewModel.swift
stefanceriu: I find it a bit weird to have a full screen with disable buttons and a small toast at the top. I would much rather we have a chat with the design team and figure this out holistically

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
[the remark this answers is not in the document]
stefanceriu: Okay, the problem was the dismiss flow wasn't doing anything in the child, we talked about and fixed it offline

pixlwave (approved):
🚢 it!

PR #2712, opened 2024-04-18
Guard against rooms prematurely unsubscribing from sliding sync:

stefanceriu:
- with the advent of permalinks the same room can be now pushed multiple times onto the stack which means multiple calls to `subscribeForUpdates`
- the reverse doesn't hold and one call to `unsubscribeFromUpdates` will unsubscribe all the rooms at the same time
- we guard against that by tracking the number of subscriptions and calling the underlying `roomListItem.unsubscribe()` when the count goes back to zero


pixlwave (approved):
LGTM 👌

PR #2716, opened 2024-04-19
Introduce an AppMediator which will:

### LINKED ISSUE #2703: EX sends read receipts whenever the app is open on a timeline, whether or not the user is present or active.
### Steps to reproduce

1. Launch EXI on macOS, open a room, and background the window
2. Notice that whenever a message is sent into that room, it sends a RR
3. This feels spooky in the extreme; EW only sends RRs if the user is actively using the app (i.e. has interacted with the UI within the last N minutes).  So instead it feels like someone might have hijacked your account, given you know it's not you reading the msgs.  Or it makes other people think you're reading your msgs when you're not.

The same behaviour may be true on EXA (even if the app is backgrounded on Android), according to @kongo09 

#### What did you expect?

Only send RRs if the user has interacted with the UI of the app (i.e. touched the screen; moved a mouse; pressed the keyboard) in the last N minutes.  Also, to match EW, one should wait for 500ms (READ_RECEIPT_INTERVAL_MS) of user activity before actually sending an RR, just in case the user is just flicking through and hasn't had time to read.

See
https://github.com/matrix-org/matrix-react-sdk/blob/a0c8575113ac74119cd7841e956cc13463ff9244/src/UserActivity.ts#L63-L77 and
https://github.com/matrix-org/matrix-react-sdk/blob/193237f20457fd161c15068eb2825b2587379140/src/components/structures/TimelinePanel.tsx#L1026 for how EW does it.

#### What happened instead?

Spooky invisible people appear to be reading your msgs who aren't there.

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

574

### Homeserver

_No response_

### Will you send logs?

No

stefanceriu:
- combine multiple services to handle app wide requests
- remove direct UIApplication access from everywhere
- fix #2703 by combining UIApplication.applicationState with the WindowManager's main window appearance
- leave the door open for more app control in the future (especiall for the mac)

Notes:
- This will probably be followed up by 
   - a refactoring of the BackgroundTaskService. 
   - a fix for https://github.com/element-hq/element-x-ios/issues/2707 (idleTimer access through the AppMediator)
- Another interestion option might be to put the WindowManager+OrientationManager in the AppMediator so there's less things to pass around


pixlwave (approved):
LGTM 👍

PR #2718, opened 2024-04-19
Fixes #2707 - Prevent the app from locking while recording a voice message

### LINKED ISSUE #2707: Disable automatic screen locking while recording a voice message
Even when I'm recording a voice message the screen will still lock automatically, especially for users who's screens lock after just 30 seconds this is very annoying.

I would like if the screen would stay unlocked while recording a voice message similar to the way it is done in Signal and WhatsApp.

I could not think of a good alternative

stefanceriu:


Velin92 (approved):
LGTM

PR #2720, opened 2024-04-19
Update strings with API breaks.

pixlwave:


stefanceriu (approved):

stefanceriu:
lgtm but the qr login screen test is failing. We also need to re-run the translations pr when this gets merged.

PR #2723, opened 2024-04-22
Cleanup following the AppMediator introduction

stefanceriu:
- stop using multiple background task, the appCoordinator sync one is enough for the whole app
- move the AppMeditor to the MainActor
- expose the WindowManager through the AppMediator

Discussion:
`UIApplication.beginBackgroundTask` is basically a glorified way of asking the system to not directly suspend the app when entering the background. We do that for a lot of network bound requests but, more importantly, we do it to allow syncing to still happen in the background for as long as the system permits it. I believe the sync one overarches the others making them redundant.

On another related note, with the changes above, I believe the BackgroundTaskService no longer brings any value, should we drop it in favor of vanilla UIApplication?


ElementX/Sources/Application/AppMediator.swift
```diff
@@ -17,7 +17,7 @@
 import UIKit
 
 class AppMediator: AppMediatorProtocol {
-    private let windowManager: WindowManagerProtocol
+    let windowManager: WindowManagerProtocol
```
pixlwave, line 20: Once again I'm going to say "please no!" to this change 🥺

Nothing should be able to call `switchToMain` other than the AppCoordinator. Maybe we should remove `switchToMain` and `switchToAlternate` into their own protocol and remove them from the `WindowManagerProtocol` seeing this is repeatedly happening?

pixlwave (changes requested):

ElementX/Sources/Application/AppMediator.swift
stefanceriu: Talked about if offline, we went with hiding the tricky bits behind a new `SecureWindowManagerProtocol`

pixlwave (approved):
Very nice, much neater. Definitely agree, we can get rid of that service now and use the mediator directly in the app coordinator 👍

ElementX/Sources/Application/Windowing/WindowManagerProtocol.swift
```diff
@@ -16,17 +16,28 @@
...
+    /// Shows the alternate window, hiding the main and overlay combo.
+    func switchToAlternate()
+}
+
 /// A window manager that supports switching between a main app window with an overlay and
 /// an alternate window to switch contexts whilst also preserving the main view hierarchy.
 /// Heavily inspired by https://www.fivestars.blog/articles/swiftui-windows/
 protocol WindowManagerProtocol: AnyObject, OrientationManagerProtocol {
-    var delegate: WindowManagerDelegate? { get set }
+    var delegate: SecureWindowManagerDelegate? { get set }
```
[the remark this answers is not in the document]
stefanceriu: Nop, it shouldn't, good point! 👍

pixlwave (approved):
👏

PR #2728, opened 2024-04-22
Handle alias room permalinks

stefanceriu:
Use the newly exposed RustSDK Client `resolveRoomAlias` method to transform room aliases into permalinks and percolate them through the rets of the application.

### DIFF
--- a/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
@@ -178,37 +178,8 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
     // MARK: - FlowCoordinatorProtocol
     
     func handleAppRoute(_ appRoute: AppRoute, animated: Bool) {
-        clearPresentedSheets(animated: animated) { [weak self] in
-            guard let self else { return }
-            
-            switch appRoute {
-            case .room(let roomID):
-                stateMachine.processEvent(.selectRoom(roomID: roomID, showingRoomDetails: false), userInfo: .init(animated: animated))
-            case .childRoom(let roomID):
-                if let roomFlowCoordinator {
-                    roomFlowCoordinator.handleAppRoute(appRoute, animated: animated)
-                } else {
-                    stateMachine.processEvent(.selectRoom(roomID: roomID, showingRoomDetails: false), userInfo: .init(animated: animated))
-                }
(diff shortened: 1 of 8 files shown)


ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -218,25 +189,76 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
...
+    
     func attemptStartingOnboarding() {
         if onboardingFlowCoordinator.shouldStart {
             clearRoute(animated: false)
             onboardingFlowCoordinator.start()
         }
     }
     
-    private func clearPresentedSheets(animated: Bool, completion: @escaping () -> Void) {
+    private func clearPresentedSheets(animated: Bool) async {
```
pixlwave, line 253: Nice I tried making this async before but ended up with all the `handleAppRoutes` also being async which was dumb. This approach looks good to me 👍

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -178,37 +178,8 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
...
-            case .genericCallLink(let url):
-                navigationSplitCoordinator.setSheetCoordinator(GenericCallLinkCoordinator(parameters: .init(url: url)), animated: animated)
-            case .oidcCallback:
-                break
-            case .settings, .chatBackupSettings:
-                settingsFlowCoordinator.handleAppRoute(appRoute, animated: animated)
-            }
+        Task {
+            await asyncHandleAppRoute(appRoute, animated: animated)
         }
```
pixlwave, line 183: I wonder if we should do something to make sure only one of these runs at once. Doesn't seem a leap to believe someone might tap a link, app opens on a poor connection, they wait a bit, go back to the link and tap again.

pixlwave (approved):
LGTM

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
stefanceriu: We do, we block the whole screen 😅

pixlwave: That doesn't help with external links 😜

PR #2731, opened 2024-04-23
Fix broken iPhone room push animation after disabling root fading for…

stefanceriu:
… the iPad


ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -17,6 +17,7 @@
 import Combine
 import Foundation
 import SwiftState
+import UIKit
```
pixlwave, line 20:
```suggestion
import SwiftState
import SwiftUI
```

pixlwave (approved):
Ah yes, good point 🙈

PR #2734, opened 2024-04-23
Gracefuly handle selecting the same room multiple times

stefanceriu:
- add support for deeplinking/navigating into the same room multiple times
    - fix bugs around the view <-> view models going out of sync
    - unwind the stack if the room is already presented
- revert attempted composer fix (navigation was the real cause)


Hywan:
<3

Hywan:
Should fix https://github.com/element-hq/element-x-ios-rageshakes/issues/1743

PR #2743, opened 2024-04-25
Add Event Routes to the flows

pixlwave:
This is the first PR to support event permalinks. It only adds the routes, which are neither called nor do anything when handled.

### DIFF
--- a/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
@@ -39,6 +39,16 @@ enum RoomFlowCoordinatorAction: Equatable {
     }
 }
 
+/// A value that represents where the room flow will be started.
+enum RoomFlowCoordinatorEntryPoint: Hashable {
+    /// The flow will start by showing the room directly.
+    case room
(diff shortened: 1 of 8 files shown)


ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -1330,6 +1372,7 @@ private extension RoomFlowCoordinator {
         case dismissJoinRoomScreen
         
         case presentRoom
+        case presentRoomFocusedOnEvent(eventID: String)
```
stefanceriu, line 1375: Talked about this offline, let's merge the 2

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -1384,7 +1427,8 @@ private extension RoomFlowCoordinator {
         case dismissRolesAndPermissionsScreen
         
         // Child room flow events
-        case startChildFlow(roomID: String)
+        enum ChildRoomFocus: Hashable { case eventID(String) }
```
stefanceriu, line 1430: ^ to be deleted

stefanceriu (approved):
Looks great, 🚀 it!

PR #2749, opened 2024-04-26
Fix some snapshots.

pixlwave:
A couple of snapshots were broken after the last translations update and I missed them 🙈


ElementX/Sources/Screens/QRCodeLoginScreen/View/QRCodeLoginScreen.swift
pixlwave, line 1: Noticed both of these on the double length pseudo-language snapshots.

Velin92 (approved):

PR #2751, opened 2024-04-26
Plain composer suggestions and pills

stefanceriu:
- allow user suggestions on the plain composer
- render pills within it for supported permalinks
- intentional mention support
- various other small tweaks


ElementX/Sources/Other/HTMLParsing/ElementXAttributeScope.swift
```diff
@@ -26,6 +26,11 @@ enum UserIDAttribute: AttributedStringKey {
     static var name = "MXUserIDAttribute"
 }
 
+enum UserDisplayNameAttribute: AttributedStringKey {
```
pixlwave, line 29: Might be worth adding a doc as to why we have both UserID and UserDisplayName attributes.

ElementX/Sources/Screens/ComposerToolbar/CompletionSuggestionServiceProtocol.swift
pixlwave, line 1: Is this file emptied without being deleted? Not sure if GH is just tricking me here.

pixlwave (approved):
Wooop 👏

pixlwave:
That drop in coverage seems very wrong for the size of this patch. Lets merge and see what happens once we have UI/integration test runs too.

PR #2767, opened 2024-04-29
QR Code Login Flow

### LINKED ISSUE #2622: [Task] Sign in with QR code

Velin92:
fixes #2622


ElementX/Sources/Services/QRCode/QRCodeLoginService.swift
```diff
@@ -0,0 +1,102 @@
...
+        qrLoginProgressSubject.eraseToAnyPublisher()
+    }
+    
+    init(oidcConfiguration: OidcConfiguration,
+         encryptionKeyProvider: EncryptionKeyProviderProtocol,
+         userSessionStore: UserSessionStoreProtocol) {
+        self.oidcConfiguration = oidcConfiguration
+        self.userSessionStore = userSessionStore
+        passphrase = encryptionKeyProvider.generateKey().base64EncodedString()
+    }
```
Velin92, line 40: @pixlwave is this the correct way to generate the passphrase? Maybe some changes should be done in place of it according if is a debug build or not?

pixlwave: Don't worry, we [removed that](https://github.com/element-hq/element-x-ios/pull/2692) a while ago now :)

ElementX/Sources/Screens/QRCodeLoginScreen/View/QRCodeScannerView.swift
```diff
@@ -73,15 +108,26 @@ final class QRScannerController: UIViewController {
...
+    override func viewWillLayoutSubviews() {
+        super.viewWillLayoutSubviews()
+        videoPreviewLayer?.frame = view.layer.bounds
+    }
+    
+    func startScan() {
         // Start video capture.
         DispatchQueue.global(qos: .userInitiated).async {
             self.captureSession.startRunning()
+            MXLog.info("QRCodeScannerView: catpure session started")
```
hughns, line 122:
```suggestion
            MXLog.info("QRCodeScannerView: capture session started")
```

ElementX/Sources/Screens/QRCodeLoginScreen/View/QRCodeScannerView.swift
```diff
@@ -73,15 +108,26 @@ final class QRScannerController: UIViewController {
...
     }
     
-    override func viewWillLayoutSubviews() {
-        super.viewWillLayoutSubviews()
-        videoPreviewLayer?.frame = view.layer.bounds
+    func stopScan() {
+        // Stop video capture.
+        DispatchQueue.global(qos: .userInitiated).async {
+            self.captureSession.stopRunning()
+            MXLog.info("QRCodeScannerView: catpure session stopped")
```
hughns, line 130:
```suggestion
            MXLog.info("QRCodeScannerView: capture session stopped")
```

ElementX/Sources/Other/Extensions/AVMetadataMachineReadableCodeObject.swift
```diff
@@ -0,0 +1,154 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+// Helpers to remove ECI headers from QR Code raw data
+// https://gist.github.com/PetrusM/267e2ee8c1d8b5dca17eac085afa7d7c
+import AVKit
+import Foundation
+
+extension AVMetadataMachineReadableCodeObject {
```
stefanceriu, line 22: This all feels very magical, I think some simple unit tests are in order, if only just to check that it doesn't break between OS versions

stefanceriu, line 22: Also, what happened to handling this on the Rust side?

ElementX/Sources/Screens/QRCodeLoginScreen/View/QRCodeLoginScreen.swift
```diff
@@ -273,37 +273,98 @@ struct QRCodeLoginScreen: View {
...
+        case .linkingNotSupported:
+            L10n.screenQrCodeLoginErrorLinkingNotSuportedTitle
+        case .deviceNotSupported:
+            L10n.screenQrCodeLoginErrorSlidingSyncNotSupportedTitle(InfoPlistReader.main.bundleDisplayName)
         case .unknown:
-            VStack(spacing: 16) {
-                HeroImage(icon: \.error, style: .critical)
+            L10n.commonSomethingWentWrong
+        default:
+            // Unused
```
stefanceriu, line 297: Let's change it to a `fatalError` then?

ElementX/Sources/Screens/QRCodeLoginScreen/View/QRCodeScannerView.swift
```diff
@@ -19,22 +19,57 @@ import SwiftUI
 import UIKit
 
 struct QRCodeScannerView: UIViewControllerRepresentable {
+    @Binding var result: Data?
```
stefanceriu, line 22: Having this a binding is a bit weird, wouldn't a callback/viewAction be easier to manage?

ElementX/Sources/Screens/QRCodeLoginScreen/View/QRCodeScannerView.swift
```diff
@@ -19,22 +19,57 @@ import SwiftUI
...
+            guard metadataObjects.count > 0,
+                  let metadataObj = metadataObjects[0] as? AVMetadataMachineReadableCodeObject,
+                  metadataObj.type == AVMetadataObject.ObjectType.qr,
+                  let data = metadataObj.binaryValue else {
+                MXLog.info("QRCodeScannerView: invalid qr scan")
+                return
+            }
+            
+            scanResult = data
+            MXLog.info("QRCodeScannerView: scanned \(data)")
```
stefanceriu, line 61: Is printing the raw data here really useful? I would remove it or at least change the log level to `debug`

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -807,11 +807,15 @@ class ClientProxy: ClientProxyProtocol {
     // MARK: - Encryption
     
     func ed25519Base64() async -> String? {
-        await client.encryption().ed25519Key()
+        // TODO: Restore this when the PR is updated
```
stefanceriu, line 810: And when will that be?

ElementX/Sources/Services/QRCode/QRCodeLoginService.swift
```diff
@@ -0,0 +1,136 @@
...
+    func loginWithQRCode(data: Data) async -> Result<UserSessionProtocol, QRCodeLoginServiceError> {
+        let qrData: QrCodeData
+        do {
+            qrData = try QrCodeData.fromBytes(bytes: data)
+        } catch {
+            MXLog.error("QRCode decode error: \(error)")
+            return .failure(.invalidQRCode)
+        }
+        
+        let listener = QrLoginProgressListenerProxy { [weak self] progress in
```
stefanceriu, line 50: With this `listener` an optional (not sure it needs to be) should we also nil it after the `login` finishes?

ElementX/Sources/Services/QRCode/QRCodeLoginService.swift
```diff
@@ -0,0 +1,136 @@
...
+        } catch let error as HumanQrLoginError {
+            MXLog.error("QRCode login error: \(error)")
+            return .failure(error.serviceError)
+        } catch {
+            MXLog.error("QRCode login unknown error: \(error)")
+            return .failure(.unknown)
+        }
+    }
+    
+    func requestAuthorizationIfNeeded() async -> Bool {
```
stefanceriu, line 74: I don't think this doesn't belong in the `QRCodeLoginService`, I would move it directly into the view model. Later we might decide to put it somewhere shared with the rest of the app

stefanceriu (approved):
Looks good, great to finally see it over the line! Very well done! 👏

ElementX/Sources/Services/Client/ClientProxy.swift
Velin92: oh wait I guess I can now

ElementX/Sources/Services/QRCode/QRCodeLoginService.swift
Velin92: it's to allow mocking and testing

ElementX/Sources/Other/Extensions/AVMetadataMachineReadableCodeObject.swift
Velin92: seems that this is not easily possible...

Velin92: The only thing is that I have no idea how to test this, because I would need to have an AVCaptureSession create this object already for me with some known errorCorrectedPayload that then I can check has been decoded properly...

stefanceriu: well.. it seems like it wasn't easily possible on iOS either 🙈  Hope android doesn't have to do the same

Velin92: no luckily on Android there is a library that allows binary decoding and even mixed decoding of QR Codes very easily, while on iOS I couldn't find such a thing. Also this only works with pure binary data qr codes, if they have mixed encoding of alphanumeric and binary data this will fail. Luckily EW uses only binary data

stefanceriu: I'm thinking the most important bits to test there are `removeQrProtocolData` and `takeBatch` and for that you can hardcode some real Data that thing generates for you on one of your testing QR codes. What do you think?

ElementX/Sources/Services/QRCode/QRCodeLoginService.swift
Velin92: You know what? Let's move this into the AppMediator

ElementX/Sources/Services/QRCode/QRCodeLoginService.swift
Velin92: actually we can remove this now, is not necessary anymore

ElementX/Sources/Services/QRCode/QRCodeLoginService.swift
stefanceriu: Works for me, good idea 👍

ElementX/Sources/Screens/QRCodeLoginScreen/View/QRCodeScannerView.swift
Velin92: I kinda like the simplicity of the binding, because it allows to easily chain it in the state and create a simple subscription to its state changes, and remove duplicates without too much `if`s

stefanceriu: I see, cool cool 👍

ElementX/Sources/Other/Extensions/AVMetadataMachineReadableCodeObject.swift
Velin92: Tests have been added!

PR #2769, opened 2024-04-30
Account URL is now fetched async

Velin92:
Requires an SDK update

### DIFF
--- a/ElementX/Sources/FlowCoordinators/SettingsFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/SettingsFlowCoordinator.swift
@@ -113,7 +113,7 @@ class SettingsFlowCoordinator: FlowCoordinatorProtocol {
                     presentSecureBackupScreen(animated: true)
                 case .userDetails:
                     presentUserDetailsEditScreen()
-                case .accountProfile(let url), .accountSessions(let url):
+                case let .manageAccount(url):
                     presentAccountManagementURL(url)
                 case .analytics:
                     presentAnalyticsScreen()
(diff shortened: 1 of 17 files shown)


ElementX/Sources/FlowCoordinators/SettingsFlowCoordinator.swift
```diff
@@ -244,26 +246,22 @@ class SettingsFlowCoordinator: FlowCoordinatorProtocol {
...
         // Note to anyone in the future if you come back here to make this open in Safari instead of a WAS.
         // As of iOS 16, there is an issue on the simulator with accessing the cookie but it works on a device. 🤷‍♂️
-        accountSettingsPresenter = OIDCAccountSettingsPresenter(accountURL: url, presentationAnchor: parameters.windowManager.mainWindow)
-        accountSettingsPresenter?.start()
+        accountSettingsPresenter = OIDCAccountSettingsPresenter(presentationAnchor: parameters.windowManager.mainWindow,
+                                                                client: parameters.userSession.clientProxy,
+                                                                accountAction: action)
+        Task {
+            await accountSettingsPresenter?.start()
+        }
```
Velin92, line 265: maybe we should put a spinner or a loader here? @pixlwave

Velin92: Or maybe we should revert this, make the function async and just wrap them in a Task since they are still quite fast to load?

pixlwave (approved):
Veeeery nice clean up 👌

PR #2774, opened 2024-05-02
 Support for posthog super properties

BillCarsonFr:
This PR is adding support for the [SuperProperities](https://github.com/matrix-org/matrix-analytics-events/blob/main/schemas/SuperProperties.json) (added to all events)

I also had to modify a bit the `PostHogAnalyticsClient` to make it more testable (can use a factory to pass the actual `PHGPostHog`).
Also made `PHGPostHog` as a protocol to have automock enabled.
Then added a test

### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Services/Analytics/AnalyticsClientProtocol.swift
+++ b/ElementX/Sources/Services/Analytics/AnalyticsClientProtocol.swift
@@ -22,7 +22,7 @@ protocol AnalyticsClientProtocol {
     var isRunning: Bool { get }
     
     /// Starts the analytics client reporting data.
-    func start(analyticsConfiguration: AnalyticsConfiguration, posthogFactory: PostHogFactory?)
+    func start(analyticsConfiguration: AnalyticsConfiguration)
        
     /// Reset all stored properties and any event queues on the client. Note that
     /// the client will remain active, but in a fresh unidentified state.
--- a/ElementX/Sources/Services/Analytics/AnalyticsClientProtocol.swift
+++ b/ElementX/Sources/Services/Analytics/AnalyticsClientProtocol.swift
@@ -22,7 +22,7 @@ protocol AnalyticsClientProtocol {
     var isRunning: Bool { get }
     
     /// Starts the analytics client reporting data.
-    func start(analyticsConfiguration: AnalyticsConfiguration)
+    func start(analyticsConfiguration: AnalyticsConfiguration, posthogFactory: PostHogFactory?)
        
     /// Reset all stored properties and any event queues on the client. Note that
     /// the client will remain active, but in a fresh unidentified state.
--- a/ElementX/Sources/Services/Analytics/PostHogAnalyticsClient.swift
+++ b/ElementX/Sources/Services/Analytics/PostHogAnalyticsClient.swift
@@ -20,6 +20,14 @@ import PostHog
 
 /// An analytics client that reports events to a PostHog server.
 class PostHogAnalyticsClient: AnalyticsClientProtocol {
+    private var posthogFactory: PostHogFactory = DefaultPostHogFactory()
+    
+    init(posthogFactory: PostHogFactory? = nil) {
+        if let factory = posthogFactory {
+            self.posthogFactory = factory
+        }
+    }
+    
     /// The PHGPostHog object used to report events.
     private var postHog: PHGPostHogProtocol?
     
@@ -33,16 +41,12 @@ class PostHogAnalyticsClient: AnalyticsClientProtocol {
     
     var isRunning: Bool { postHog?.enabled ?? false }
     
-    func start(analyticsConfiguration: AnalyticsConfiguration, posthogFactory: PostHogFactory?) {
+    func start(analyticsConfiguration: AnalyticsConfiguration) {
         // Only start if analytics have been configured in BuildSettings
         guard let configuration = PHGPostHogConfiguration.standard(analyticsConfiguration: analyticsConfiguration) else { return }
         
         if postHog == nil {
-            if let factory = posthogFactory {
-                postHog = factory.createPostHog(config: configuration)
-            } else {
-                postHog = PHGPostHog(configuration: configuration)
-            }
+            postHog = posthogFactory.createPostHog(config: configuration)
         }
         // Add super property cryptoSDK to the captured events, to allow easy
         // filtering of events across different client by using same filter.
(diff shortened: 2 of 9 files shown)


ElementX/Sources/Services/Analytics/PostHogAnalyticsClient.swift
```diff
@@ -89,4 +115,19 @@ class PostHogAnalyticsClient: AnalyticsClientProtocol {
         pendingUserProperties = nil
         return properties
     }
+    
+    /// Attach super properties to events.
+    /// If the property is already set on the event, the already set value will be kept.
+    private func attachSuperProperties(to properties: [String: Any]) -> [String: Any] {
+        guard isRunning, let superProperties = superProperties else { return properties }
```
pixlwave, line 122:
```suggestion
        guard isRunning, let superProperties else { return properties }
```

ElementX/Sources/Services/Analytics/AnalyticsClientProtocol.swift
```diff
@@ -22,7 +22,7 @@ protocol AnalyticsClientProtocol {
     var isRunning: Bool { get }
     
     /// Starts the analytics client reporting data.
-    func start(analyticsConfiguration: AnalyticsConfiguration)
+    func start(analyticsConfiguration: AnalyticsConfiguration, posthogFactory: PostHogFactory?)
```
pixlwave, line 25: Instead of passing the factory in `start`, could we pass it in the initialiser of `PostHogAnalyticsClient` and store it as a property until it is needed? `AnalyticsClientProtocol` is intended to be agnostic to the backend so forks can drop their own solution into `AnalyticsService` without needing API changes.

ElementX/Sources/Services/Analytics/PostHogAnalyticsClient.swift
```diff
@@ -15,26 +15,38 @@
...
         // Only start if analytics have been configured in BuildSettings
         guard let configuration = PHGPostHogConfiguration.standard(analyticsConfiguration: analyticsConfiguration) else { return }
         
         if postHog == nil {
-            postHog = PHGPostHog(configuration: configuration)
+            if let factory = posthogFactory {
+                postHog = factory.createPostHog(config: configuration)
+            } else {
+                postHog = PHGPostHog(configuration: configuration)
+            }
```
pixlwave, line 45: We could then tidy this up by making the factory non-optional and supplying a factory for the real `PHGPostHog` too.

ElementX/Sources/Services/Analytics/PostHogAnalyticsClient.swift
```diff
@@ -73,6 +85,20 @@ class PostHogAnalyticsClient: AnalyticsClientProtocol {
...
+            superProperties = updatedProperties
+            return
+        }
+        
+        superProperties = AnalyticsEvent.SuperProperties(appPlatform: updatedProperties.appPlatform ??
+            currentProperties.appPlatform,
+            cryptoSDK: updatedProperties.cryptoSDK ??
+                currentProperties.cryptoSDK,
+            cryptoSDKVersion: updatedProperties.cryptoSDKVersion ??
+                currentProperties.cryptoSDKVersion)
```
pixlwave, line 99: The preview looks dumb with the wrapping, but would be nice to keep both sides of the `??` on the same line like we do for `UserProperties`.
```suggestion
        superProperties = AnalyticsEvent.SuperProperties(appPlatform: updatedProperties.appPlatform ?? currentProperties.appPlatform,
                                                         cryptoSDK: updatedProperties.cryptoSDK ?? currentProperties.cryptoSDK,
                                                         cryptoSDKVersion: updatedProperties.cryptoSDKVersion ?? currentProperties.cryptoSDKVersion)
```

ElementX/Sources/Services/Analytics/PostHogAnalyticsClient.swift
```diff
@@ -15,26 +15,38 @@
 //
 
 import AnalyticsEvents
+import MatrixRustSDK
```
pixlwave, line 18: Is this used for anything? I don't think it is, but if I'm missing something then we should ideally map any Rust types outside of here to maintain a separation of concerns.

pixlwave (changes requested):
Very nice, thanks for the tests! Just a few small thoughts inline 👍

ElementX/Sources/Services/Analytics/PostHogAnalyticsClient.swift
BillCarsonFr: It is used for `sdkGitSha`. 
We could remove it by properly calling `updateUserProperties` with that value from somewere else. What would you recommand?

ElementX/Sources/Services/Analytics/PostHogAnalyticsClient.swift
BillCarsonFr: can't make swiftformat happy with any other formatting

ElementX/Sources/Services/Analytics/PostHogAnalyticsClient.swift
BillCarsonFr: done

ElementX/Sources/Services/Analytics/AnalyticsClientProtocol.swift
BillCarsonFr: done

ElementX/Sources/Services/Analytics/PostHogAnalyticsClient.swift
BillCarsonFr: done

ElementX/Sources/Services/Analytics/PostHogAnalyticsClient.swift
pixlwave: Right yes I missed that 🤦‍♂️. We could pass it in to the init externally when we configure the service in [setupServiceLocator](https://github.com/element-hq/element-x-ios/blob/e511ac3cb2e55f9a3b7d8e4183fececfb7fa651a/ElementX/Sources/Application/AppCoordinator.swift#L316-L318). That would align quite nicely with the bug report service too. WDYT?

BillCarsonFr: yes put it there , thx 🙏

PR #2778, opened 2024-05-03
Fix for extra newline in emotes

### LINKED ISSUE #2503: Emotes combined with rich text has spurious linebreak after the displayname
### Steps to reproduce

1. Send an emote with some rich text content:

Screenshot 2024-02-27 at 09 44 16

2. Observe that after `* Matthew` there's a linebreak, due presumably to the styling of the paragraph tag wrapping the rich text acting as a block style element rather than inline.

#### What did you expect?

The rich contents should be inline styled rather than block styled.

Alternatively, one could hoist the `* Matthew` text up inside the block styled paragraph somehow (e.g. prepending it to the first non-whitespace text content), or the enclosing paragraph tag could be stripped out entirely (which is how I did it on EW, although it looks to have since regressed).

#### What happened instead?

Ugly glitchy linebreaks.

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

529

### Homeserver

_No response_

### Will you send logs?

No

Velin92:
fixes #2503


pixlwave (approved):
LGTM, maybe we should add a test case for this one?

PR #2780, opened 2024-05-04
Allow emotes in the menu to scroll if their size overflows the container

### LINKED ISSUE #2406: Message react tool tray fails to render emoji's when Button Shapes Accessibility option is enabled
### Steps to reproduce

I am going blind and have accessibility options enabled on my phone. Specifically:
- Button Shapes
- On/Off Labels
- Reduce Transparency
- Increase Contrast
- Differentiate without color
- And my text scale is turned up all of the way but before "Larger Accessibility Sizes" is enabled
From any channel or dm chat. When i initiate a message react for a message or picture i hope to see & use the emoji's that i react to a message photo. 

### Outcome

The emoji widget menu fails to load and shows blank empty gray boxes where emoji's should load. 
Once i disable Button shapes the emoji's load but are clipping into one another

### Your phone model

iPhone 11 Pro Max

### Operating system version

17.3

### Application version

1.5.4

### Homeserver

catboyindustries.co Synapse 1.99.0

### Will you send logs?

Yes

Velin92:
fixes #2406 

This allows to keep the existing behaviour where the emojis take all the available space if possible.
In case their size overflows the existing space (because the font has been increased or button shapes are showed) they will instead scroll horizontally.

However the preview tests for some reason do not show the correct behaviour on iPad (even if the normal SwiftUI previews do) so here are some screenshot of the behaviour on iPad normally and on iPhone with the button shapes enabled:

Simulator Screenshot - iPad (10th generation) - 2024-05-04 at 02 25 49

https://github.com/element-hq/element-x-ios/assets/34335419/fac9ef14-0c42-4f84-8323-94e5bb23e97d


ElementX/Sources/Screens/RoomScreen/View/TimelineItemMenu.swift
```diff
@@ -212,28 +214,32 @@ struct TimelineItemMenu: View {
...
-                    .padding(10)
+                .accessibilityLabel(L10n.actionReact)
             }
-            .accessibilityLabel(L10n.actionReact)
+            .padding(.horizontal)
+            .frame(minWidth: reactionsFrame.width, maxWidth: .infinity, alignment: .center)
         }
-        .padding(.horizontal)
-        .frame(maxWidth: .infinity, alignment: .center)
+        .scrollIndicators(.hidden)
```
pixlwave, line 241: Could add a `.scrollBounceBehavior(.basedOnSize)` here so that it isn't scrollable when everything fits.

pixlwave (approved):
Ah nice, that's a decent solution to the problem 👌

PR #2788, opened 2024-05-06
Bump posthog SDK version to 3.2.0

BillCarsonFr:
Bump posthog version to 3.2.0.
We were using a quite old posthog sdk. Updating to the new one mainly because it now support [sessions](https://posthog.com/docs/data/sessions). Allow us to have a per device view instead of only unique users that is merging events from all the devices of the same user.
Element X Android is already using [posthog 3.x](https://github.com/element-hq/element-x-android/blob/86838e7277d7a094496b9d231f06aad99f1e445e/gradle/libs.versions.toml#L183)

### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Services/Analytics/PostHogAnalyticsClient.swift
+++ b/ElementX/Sources/Services/Analytics/PostHogAnalyticsClient.swift
@@ -38,11 +38,11 @@ class PostHogAnalyticsClient: AnalyticsClientProtocol {
     /// Not persisted for now, should be set on start.
     private var superProperties: AnalyticsEvent.SuperProperties?
     
-    var isRunning: Bool { postHog?.enabled ?? false }
+    var isRunning: Bool { postHog != nil }
     
     func start(analyticsConfiguration: AnalyticsConfiguration) {
         // Only start if analytics have been configured in BuildSettings
-        guard let configuration = PHGPostHogConfiguration.standard(analyticsConfiguration: analyticsConfiguration) else { return }
+        guard let configuration = PostHogConfig.standard(analyticsConfiguration: analyticsConfiguration) else { return }
         
         if postHog == nil {
             postHog = posthogFactory.createPostHog(config: configuration)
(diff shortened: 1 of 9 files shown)


ElementX/Sources/Services/Analytics/PostHogAnalyticsClient.swift
```diff
@@ -59,8 +59,8 @@ class PostHogAnalyticsClient: AnalyticsClientProtocol {
     }
     
     func stop() {
-        postHog?.disable()
-        
+        postHog?.optOut()
+        postHog = nil
```
Velin92, line 63: there is a comment below saying that setting it to nil does not release it, maybe we should remove that comment if its the case

ElementX/Sources/Services/Analytics/PostHogAnalyticsClient.swift
```diff
@@ -38,19 +38,19 @@ class PostHogAnalyticsClient: AnalyticsClientProtocol {
...
+    var isRunning: Bool { postHog != nil }
     
     func start(analyticsConfiguration: AnalyticsConfiguration) {
         // Only start if analytics have been configured in BuildSettings
-        guard let configuration = PHGPostHogConfiguration.standard(analyticsConfiguration: analyticsConfiguration) else { return }
+        guard let configuration = PostHogConfig.standard(analyticsConfiguration: analyticsConfiguration) else { return }
         
         if postHog == nil {
             postHog = posthogFactory.createPostHog(config: configuration)
         }
```
Velin92, line 49: Since now we can rely on the fact that postHog is going to be nil when starting it, maybe we should avoid doing this check and always instatiate it.

I would even go as far as doing 
```
if postHog != nil {
	MXLog.failure("Post hog should always be nil when it's being started")
}
postHog = posthogFactory.createPostHog(config: configuration)
```

ElementX/Sources/Services/Analytics/PostHogAnalyticsClient.swift
BillCarsonFr: I removed the comment.

ElementX/Sources/Services/Analytics/PostHogAnalyticsClient.swift
BillCarsonFr: Updated

Velin92 (approved):

PR #2791, opened 2024-05-06
Updated settings screen

### LINKED ISSUE #2789: [Task] Cleanup: settings screen
Implement design changes. Please see and discuss the details in the meta issue.

Velin92:
fixes #2789

### DIFF
--- a/ElementX/Sources/Screens/Settings/SettingsScreen/View/SettingsScreen.swift
+++ b/ElementX/Sources/Screens/Settings/SettingsScreen/View/SettingsScreen.swift
@@ -100,7 +100,6 @@ struct SettingsScreen: View {
         }
     }
     
-    @ViewBuilder
     private var manageAccountSection: some View {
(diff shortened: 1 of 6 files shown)


ElementX/Sources/Screens/Settings/SettingsScreen/View/SettingsScreen.swift
```diff
@@ -91,7 +100,7 @@ struct SettingsScreen: View {
         }
     }
     
-    private var mainSection: some View {
+    private var manageAccountSection: some View {
```
pixlwave, line 103: This section needs to be conditionally shown if there isn't an account URLs or blocked users. Otherwise there'll be a bigger gap for the empty section. Maybe tidier to put the if directly in the body though?

pixlwave (approved):
Looks good although think we need another `if` in here somewhere.

PR #2793, opened 2024-05-07
Posthog: use correct API for user properties

BillCarsonFr:
Follow up on https://github.com/element-hq/element-x-ios/pull/2788

Posthog SDK has now an API to pass user properties, so no need to manually add them as `$set` in the event properties.
Updated the tests

### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Services/Analytics/PostHogAnalyticsClient.swift
+++ b/ElementX/Sources/Services/Analytics/PostHogAnalyticsClient.swift
@@ -69,12 +69,13 @@ class PostHogAnalyticsClient: AnalyticsClientProtocol {
     
     func capture(_ event: AnalyticsEventProtocol) {
         guard isRunning else { return }
-        postHog?.capture(event.eventName, properties: attachUserProperties(to: attachSuperProperties(to: event.properties)))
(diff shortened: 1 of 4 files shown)


ElementX/Sources/Services/Analytics/PostHogAnalyticsClient.swift
```diff
@@ -69,12 +69,13 @@ class PostHogAnalyticsClient: AnalyticsClientProtocol {
     
     func capture(_ event: AnalyticsEventProtocol) {
         guard isRunning else { return }
-        postHog?.capture(event.eventName, properties: attachUserProperties(to: attachSuperProperties(to: event.properties)))
+        postHog?.capture(event.eventName, properties: attachSuperProperties(to: event.properties), userProperties: pendingUserProperties?.properties.compactMapValues { $0 })
```
Velin92, line 72: is this supposed to be the name of the API? `attachSuperProperties` instead of `attachUserProperties`?

Velin92 (approved):

ElementX/Sources/Services/Analytics/PostHogAnalyticsClient.swift
BillCarsonFr: No we still attach super properties to the existing event properies.
The new thing is now that `capture()` has a dedicated param for user properties

PR #2815, opened 2024-05-08
Update the SDK

pixlwave:
OTLP support seems to have been removed. I've kept those changes in a separate commit so they're easy to revert if this was unexpected or we ever need it again.


Velin92 (approved):

Velin92 (approved):
Are we sure we want to merge this? maybe let's check first why OTLP has been removed

pixlwave:
We discussed internally and found out why it was removed.

PR #2821, opened 2024-05-09
Enable room list invites by default and remove old code paths

stefanceriu:


Velin92 (approved):
LGTM, but is worth also doing a nice periphery run to double check if there is anything else worth removing

stefanceriu:
> is worth also doing a nice periphery run

Good shout, all clear! 👍

PR #2822, opened 2024-05-09
UX Update for user profile screen and room member details

### LINKED ISSUE #2816: [Task] UX cleanup: user profile screen
Add a 'create/join DM' CTA and a 'call' one. Please see and discuss the details in the meta issue.

Velin92:
This PR also changed the behaviour of the form button style all across the app, and introduced a way to start calls directly when checking a user profile (if a DM already exists)

fixes #2816


ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -176,6 +176,14 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
         }
     }
     
+    private func handleCallRoute(roomID: String) async {
```
pixlwave, line 179: Maybe this should be renamed now it doesn't use a route and moved below the rest of the route handling. Something like `displayCallScreen` perhaps?

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -718,6 +720,14 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
...
     }
     
+    private func presentCall(roomID: String) async {
+        guard let roomProxy = await userSession.clientProxy.roomForIdentifier(roomID) else {
+            return
+        }
+        
+        presentCallScreen(roomProxy: roomProxy)
+    }
+
```
pixlwave, line 730: Might be nice to put this just above `presentCallScreen` and maybe give it the same name?

ElementX/Sources/Screens/RoomMemberDetailsScreen/RoomMemberDetailsScreenViewModel.swift
```diff
@@ -178,7 +186,12 @@ class RoomMemberDetailsScreenViewModel: RoomMemberDetailsScreenViewModelType, Ro
             if isNewRoom {
                 analytics.trackCreatedRoom(isDM: true)
             }
-            actionsSubject.send(.openDirectChat(roomID: roomID))
+            
+            if shouldStartCall {
+                actionsSubject.send(.startCall(roomID: roomID))
```
pixlwave, line 191: Not sure it makes sense to route the call through the `openDirectChat` method seeing as it involves `createDirectRoomIfNeeded`. We could store the discovered room ID above and in the view do 
```
if let dmRoomID {
    Button("Call") { send(viewAction: .startCall(dmRoomID)) }
}
```

with no extra SDK calls needed.

ElementX/Sources/Screens/UserProfileScreen/UserProfileScreenViewModel.swift
```diff
@@ -112,7 +120,11 @@ class UserProfileScreenViewModel: UserProfileScreenViewModelType, UserProfileScr
             if isNewRoom {
                 analytics.trackCreatedRoom(isDM: true)
             }
-            actionsSubject.send(.openDirectChat(roomID: roomID))
+            if shouldStartCall {
+                actionsSubject.send(.startCall(roomID: roomID))
```
pixlwave, line 124: Same here too :)

PreviewTests/__Snapshots__/PreviewTests/test_userProfileScreen-iPhone-15-pseudo.Other-User.png
pixlwave, line 1: Should we handle this case differently so the height matches and the icons are top aligned or something?

pixlwave (approved):

PreviewTests/__Snapshots__/PreviewTests/test_userProfileScreen-iPhone-15-pseudo.Other-User.png
Velin92: I think this is a very edge case, so maybe we shouldn't worry too much on its behaviour for now, in case someone complains we will decide accordingly on how we should handle the rendering.

PR #2823, opened 2024-05-09
Add via parameters to routes and give them to the SDK

### LINKED ISSUE #2825: Permalink error handling - #netbsd:nil.im won’t open
### Steps to reproduce

1. Click on a link to https://matrix.to/#/%23netbsd:nil.im
2. Loding spinner appears and then disappears
3. nothing happens

#### What did you expect?

Some sort of error message to be shown

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

595

### Homeserver

_No response_

### Will you send logs?

Yes

pixlwave:
Depends on https://github.com/matrix-org/matrix-rust-sdk/pull/3393

- Routes with room IDs now include via parameters.
- Parsing permalinks into a route now includes said via parameters.
- Resolved aliases add the via parameters to the resulting routes.
- The via parameters are passed to the JoinRoomScreen.
- Joining a room by ID now uses the via parameters.

I've reversed the event routes so the `eventID` comes first as this makes more sense.

Note: This doesn't yet pass them to the room preview API (as I haven't figured out how that one works).

Also fixes #2825 with a failure toast.

### DIFF
--- a/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
@@ -21,14 +21,11 @@ import UserNotifications
 
 // swiftlint:disable file_length
 enum RoomFlowCoordinatorAction: Equatable {
-    case presentRoom(roomID: String)
     case presentCallScreen(roomProxy: RoomProxyProtocol)
     case finished
     
     static func == (lhs: RoomFlowCoordinatorAction, rhs: RoomFlowCoordinatorAction) -> Bool {
         switch (lhs, rhs) {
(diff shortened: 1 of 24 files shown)


ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -21,14 +21,11 @@ import UserNotifications
 
 // swiftlint:disable file_length
 enum RoomFlowCoordinatorAction: Equatable {
-    case presentRoom(roomID: String)
```
pixlwave, line 24: Turns out this was unused but not detected by periphery because of it being a bit cyclic in nature with child flows.

Velin92 (approved):

PR #2826, opened 2024-05-10
Prevent links from showing up in color within code blocks, simplify f…

stefanceriu:
…oreground color stripping


Velin92 (approved):

ElementX/Sources/Screens/RoomScreen/View/Timeline/FormattedBodyText.swift
```diff
@@ -231,6 +242,8 @@ struct FormattedBodyText_Previews: PreviewProvider, TestablePreview {
             }
             .padding()
         }
+        .previewLayout(.sizeThatFits)
```
pixlwave, line 245: 👌

PR #2827, opened 2024-05-10
Fixes #2805 - Fixes animation artifacts when focusing on in-reply-to …

### LINKED ISSUE #2805: tapping on replies to messages that are in top half of screen causes weird animation
### Steps to reproduce

1. Open room with a message and reply on the initial screen and the original message is in the top half
2. tap on the replied message permalink in the reply
3. observe weird animation

#### What did you expect?
smooth animation or no animation at all

#### What happened instead?

https://github.com/element-hq/element-x-ios/assets/29716248/a5203148-18d3-48e3-82e7-dda8cd27b0d6

To me, that animation looks like a combination of the timeline being redrawn at the new event and a try to fake smooth scrolling instead of actually scrolling to the visible event (which seems reasonable for unloaded or off-screen events, but not for visible ones).

### Your phone model

iPhone 13 mini / iPad Pro

### Operating system version

iOS/iPadOS 17.4.1

### Application version

1.6.4 (104)

### Homeserver

Synapse 1.106.0 / Proxy 0.99.16

### Will you send logs?

No

stefanceriu:
…messages

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/View/Timeline/TimelineTableViewController.swift
+++ b/ElementX/Sources/Screens/RoomScreen/View/Timeline/TimelineTableViewController.swift
@@ -340,10 +340,13 @@ class TimelineTableViewController: UIViewController {
     
     /// Scrolls to the item with the corresponding event ID if loaded in the timeline.
     private func scrollToItem(eventID: String, animated: Bool) {
-        if let kvPair = timelineItemsDictionary.first(where: { $0.value.identifier.eventID == eventID }),
-           let indexPath = dataSource?.indexPath(for: kvPair.key) {
-            tableView.scrollToRow(at: indexPath, at: .middle, animated: animated)
-            coordinator.send(viewAction: .scrolledToFocussedItem)
+        DispatchQueue.main.async { [weak self] in // Fixes #2805
+            guard let self else { return }
+            if let kvPair = timelineItemsDictionary.first(where: { $0.value.identifier.eventID == eventID }),
+               let indexPath = dataSource?.indexPath(for: kvPair.key) {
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Screens/RoomScreen/View/Timeline/TimelineTableViewController.swift
```diff
@@ -340,10 +340,13 @@ class TimelineTableViewController: UIViewController {
...
-        if let kvPair = timelineItemsDictionary.first(where: { $0.value.identifier.eventID == eventID }),
-           let indexPath = dataSource?.indexPath(for: kvPair.key) {
-            tableView.scrollToRow(at: indexPath, at: .middle, animated: animated)
-            coordinator.send(viewAction: .scrolledToFocussedItem)
+        DispatchQueue.main.async { [weak self] in // Fixes #2805
+            guard let self else { return }
+            if let kvPair = timelineItemsDictionary.first(where: { $0.value.identifier.eventID == eventID }),
+               let indexPath = dataSource?.indexPath(for: kvPair.key) {
+                tableView.scrollToRow(at: indexPath, at: .middle, animated: animated)
+                coordinator.send(viewAction: .scrolledToFocussedItem)
```
pixlwave, line 348: We can remove the task around `scrolledToFocussedItem` in the view model's process view action now too.

pixlwave (approved):
Nice find 🙌

ElementX/Sources/Screens/RoomScreen/View/Timeline/TimelineTableViewController.swift
stefanceriu: Nice 👍

PR #2831, opened 2024-05-10
UI update for room details

Velin92:
New CTAs have been, badges alongside badges representing the room type and a way to open the call directly in the details


pixlwave (approved):
Nice! 🤩

PR #2835, opened 2024-05-13
Update the `build-sdk` tool

pixlwave:
- Support the new `--target` option on Rust allowing multiple targets to be built.
- Add new `--device` and `--simulator` shortcuts so we don't need to remember triples.

Requires https://github.com/matrix-org/matrix-rust-sdk/pull/3408


stefanceriu (approved):
Very very nice! 👏

PR #2837, opened 2024-05-15
Switch to element-hq Rust components.

pixlwave:
This PR changes the Rust components dependency to use [element-hq/matrix-rust-components-swift](https://github.com/element-hq/matrix-rust-components-swift) (only supports iOS and the simulator).

Additionally it adopts the shared command line tools for running commands in our Tools script.


stefanceriu (approved):
YEY! 👏 👏

PR #2838, opened 2024-05-16
Add device keys to rageshakes

BillCarsonFr:
Fixes https://github.com/element-hq/element-x-ios/issues/2550

Draft because depends on https://github.com/matrix-org/matrix-rust-sdk/pull/3420

Sample RS containing the device keys https://github.com/element-hq/element-x-ios-rageshakes/issues/1807

We want to add curve25519/ed25519 device keys in rageshakes to help debugging unable to decrypt issues.
Will appear like that in the RS
```
device_keys: curve25519:GJ1N4ps7v9+1SXMZ6fNXCk/skwBrVaDbepCJ9lZ5HyY, ed25519:rurFYehF4BKjkCGrTQhIrZb1LuQlbDKXOrF42f39dh8
```

This info was not available in the bindings, so added it (see rust PR), they can be accessed via `client.encryption().ed25519Key()`

Previously the `BugReportScreenCoordinatorParameters` was just having a list of params for `userID`, `deviceID`. Couldn't just add the 2 new ones easilly because they require an async call. So now the userSessionProtocol is passed as a parameter, the Rageshake submit code can then just access what it needs (userId, deviceId, curve25519, ed25519). 
I think it make sense to pass it anyhow, because there could be more things to get from the userSession when submitting a rageshake.

### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Screens/BugReportScreen/BugReportScreenViewModel.swift
+++ b/ElementX/Sources/Screens/BugReportScreen/BugReportScreenViewModel.swift
@@ -21,8 +21,7 @@ typealias BugReportScreenViewModelType = StateStoreViewModel<BugReportScreenView
 
 class BugReportScreenViewModel: BugReportScreenViewModelType, BugReportScreenViewModelProtocol {
     private let bugReportService: BugReportServiceProtocol
-    private let userID: String?
-    private let deviceID: String?
+    private let userSession: UserSessionProtocol?
     
     private let actionsSubject: PassthroughSubject<BugReportScreenViewModelAction, Never> = .init()
     // periphery:ignore - when set to nil this is automatically cancelled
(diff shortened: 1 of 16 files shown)


ElementX/Sources/Screens/BugReportScreen/BugReportScreenViewModel.swift
```diff
@@ -86,8 +83,12 @@ class BugReportScreenViewModel: BugReportScreenViewModelType, BugReportScreenVie
...
             }
         }
-        let bugReport = BugReport(userID: userID,
-                                  deviceID: deviceID,
+        let ed25519 = await userSession?.clientProxy.ed25519Base64()
+        let curve25519 = await userSession?.clientProxy.curve25519Base64()
+        let bugReport = BugReport(userID: userSession?.userID,
+                                  deviceID: userSession?.deviceID,
+                                  ed25519: ed25519,
+                                  curve25519: curve25519,
```
stefanceriu, line 91: Seeing as both the `userID` and `deviceID` come from the `ClientProxy` I would only pass that in here and not the whole `UserSession`. I think it's also time we get rid of those properties from the `UserSession` and just use the `ClientProxy` everywhere

stefanceriu (approved):
One small comment about simplified the interfaces otherwise looks good to me 👍

ElementX/Sources/Screens/BugReportScreen/BugReportScreenViewModel.swift
BillCarsonFr: See here https://github.com/element-hq/element-x-ios/pull/2838/commits/bd53979adafa916112cf16d93698647b25e70256

pixlwave (approved):
LGTM, thanks for the changes 👌

PR #2857, opened 2024-05-17
Bump SDK

Velin92:
this requires a new SDK release.

The pagination handling has been changed slightly, this PR was made to check if the changes are valid.

### DIFF
--- a/ElementX/Sources/Services/Timeline/TimelineProxy.swift
+++ b/ElementX/Sources/Services/Timeline/TimelineProxy.swift
@@ -29,7 +29,8 @@ final class TimelineProxy: TimelineProxyProtocol {
     // periphery:ignore - retaining purpose
     private var timelineListener: RoomTimelineListener?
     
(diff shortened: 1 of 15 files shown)


ElementX/Sources/Services/Timeline/TimelineProxy.swift
```diff
@@ -536,7 +536,19 @@ final class TimelineProxy: TimelineProxyProtocol {
...
+            }
+            
+            switch status {
+            case .idle:
+                backPaginationStatusSubject.send(.idle)
+            case .paginating:
+                backPaginationStatusSubject.send(.paginating)
+            case .fetchingTargetEvent, .initial:
+                // TODO: NO IDEA
+                break
```
Velin92, line 550: @pixlwave how do you think this case should be handled (if should be handled at all)?

pixlwave: Have pushed a commit after chatting to @bnjbvr about how it works now 👍

PR #2862, opened 2024-05-18
Add support for showing the native OS incoming call screen when starting new Element Calls

stefanceriu:
Handle incoming VoIP calls, show the native OS call screen as per https://developer.apple.com/documentation/callkit/sending-end-to-end-encrypted-voip-calls
Defined through https://github.com/matrix-org/matrix-spec-proposals/pull/4075

Fixes https://github.com/element-hq/element-x-ios/issues/2876

Requires https://github.com/matrix-org/matrix-rust-sdk/pull/3434

The way this works is the following:
- the NSE receives the notification and decrypts it
- checks if it's still time relevant (max 10 seconds old) and whether it should ring
- otherwise it goes on to show it as a normal notification
- if it should ring then it discards the notification but invokes `reportNewIncomingVoIPPushPayload`
so that the main app can handle it
- the main app picks this up in `PKPushRegistry.didReceiveIncomingPushWith` and
`CXProvider.reportNewIncomingCall` to show the system UI and handle actions on it.

N.B. this flow works properly only when background processing capabilities are enabled

IMG_2432

### DIFF
--- a/ElementX/Sources/Other/HTMLParsing/AttributedStringBuilder.swift
+++ b/ElementX/Sources/Other/HTMLParsing/AttributedStringBuilder.swift
@@ -143,6 +143,7 @@ struct AttributedStringBuilder: AttributedStringBuilderProtocol {
         attributedString.removeAttribute(.foregroundColor, range: .init(location: 0, length: attributedString.length))
     }
     
+    // swiftlint:disable:next cyclomatic_complexity
     private func addLinksAndMentions(_ attributedString: NSMutableAttributedString) {
         let string = attributedString.string
         
--- a/ElementX/Sources/Services/ElementCall/ElementCallServiceProtocol.swift
+++ b/ElementX/Sources/Services/ElementCall/ElementCallServiceProtocol.swift
@@ -1,0 +1,40 @@
+//
+// Copyright 2024 New Vector Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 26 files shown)


ElementX/Sources/Other/HTMLParsing/AttributedStringBuilder.swift
```diff
@@ -143,6 +143,7 @@ struct AttributedStringBuilder: AttributedStringBuilderProtocol {
         attributedString.removeAttribute(.foregroundColor, range: .init(location: 0, length: attributedString.length))
     }
     
+    // swiftlint:disable:next cyclomatic_complexity
```
Velin92, line 146: why is this here?

ElementX/Sources/Services/ElementCall/ElementCallServiceProtocol.swift
```diff
@@ -0,0 +1,40 @@
...
+}
+
+enum ElementCallServiceNotificationKey: String {
+    case roomID
+    case roomDisplayName
+    case senderID
+    case senderDisplayName
+}
+
+let ElementCallServiceNotificationDiscardDelta = 10.0
```
Velin92, line 31: maybe let's put this into a `ElementCallServiceConstants` enum as a `static let`

Velin92 (approved):
LGTM just some suggestions

ElementX/Sources/Other/HTMLParsing/AttributedStringBuilder.swift
stefanceriu: Because we have a warning there? 🤔 

Screenshot 2024-05-22 at 10 40 07

ElementX/Sources/Services/ElementCall/ElementCallServiceProtocol.swift
stefanceriu: Why would that be better? 🤔

Velin92 (approved):

PR #2873, opened 2024-05-21
Update the SDK.

pixlwave:
Includes a fix to a crash with the day divider.


ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -238,7 +238,7 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
         let details = RoomSummaryDetails(id: roomInfo.id,
                                          isInvite: roomInfo.membership == .invited,
                                          inviter: inviterProxy,
-                                         name: roomInfo.rawName ?? roomInfo.id,
+                                         name: roomInfo.displayName ?? roomInfo.id,
```
pixlwave, line 241: `rawName` now returns `nil` for DMs so we need the computed name. The sliding sync `name` is no longer a thing.

Velin92 (approved):

PR #2874, opened 2024-05-21
Add a config for Enterprise.

pixlwave:
Copy and paste of the Nightly config without the build number on the icon.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -180,7 +180,9 @@ lane :config_enterprise do
   target_file_path = "../project.yml"
   data = YAML.load_file target_file_path
   data["settings"]["BASE_APP_GROUP_IDENTIFIER"] = "io.element.enterprise"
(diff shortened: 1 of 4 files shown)


Velin92 (approved):

fastlane/Fastfile
```diff
@@ -171,9 +176,19 @@ lane :config_nightly do |options|
   update_app_icon(caption_text: "#{release_version}(#{build_number})", modulate: "100,20,100")
 end
 
-lane :config_production do
+lane :config_enterprise do
+  target_file_path = "../project.yml"
+  data = YAML.load_file target_file_path
+  data["settings"]["BASE_APP_GROUP_IDENTIFIER"] = "io.element.enterprise"
+  data["settings"]["BASE_BUNDLE_IDENTIFIER"] = "io.element.elementx.enterprise"
```
manuroe, line 183: I think we want to use `io.element.enterprise`. There will be no `X` for that project

pixlwave: Fixed.

PR #2879, opened 2024-05-24
Tweak the onboarding flow identity confirmation screen

stefanceriu:
- hide the recovery key option if recovery wasn't set up beforehand
- add a button to present the reset encryption instructions screen

There's 2 scenarios we're interested in here:

1) The user doesn't have recovery set up, logs into a new session but it's not their last session. The only option presented should be interactive verification and if the user doesn't have access to any of the other sessions then the only way forward is to reset

2) The user doesn't have recovery set up, logs into a new session which happens to be their last session. Again, only option available is to reset their encryption

In both cases we need to show them instructions on how to do that reset.

Screenshot 2024-05-24 at 13 56 12


pmaier1:
A couple of points that come to my mind:

1. In this scenario we assume that the user has published cross-signing keys before. If they haven't, then the "Confirm it's you" screen should not be shown at all. Instead cross-signing should be set up for them.
2. If the user **has not set up recovery yet** and this is their only device (and cross-signing was set up before), they should directly be guided to reset.
3. If the user **has not set up recovery yet** and they have other devices (and cross-signing was set up before), they should be guided to verify with another device or reset (recovery key is not an option)

I think this matches your assessment having thought about it and read it again 👍 . The copy for the "reset" button should be "Lost your recovery key?"

Velin92 (approved):

PR #2892, opened 2024-05-30
Add a flag to disable hidden profiles for ignored users.

pixlwave:
We won't use the flag, it is for forks.


stefanceriu (approved):
lgtm 👍

PR #2899, opened 2024-06-04
SwiftLint fixes.

pixlwave:
Fixes warnings from `non_optional_string_data_conversion` and `static_over_final_class` rules.

Note: This leaves the Data -> String warnings as that part of the rule is going to be [reverted](https://github.com/realm/SwiftLint/pull/5601).


stefanceriu (approved):
😎

PR #2900, opened 2024-06-04
Use the room heroes when computing a DM avatar.

### LINKED ISSUE #1659: Tinted avatar can be different in the room setting screen
I saw the problem in a single DM. It may be a race condition: killing the app does not fix the problem but clearing the cache, yes.

In the room list and in the timeline:
IMG_8072

In the DM room:
IMG_8073

pixlwave:
Requires https://github.com/matrix-org/matrix-rust-sdk/pull/3503

- Add heroes from the SDK to both `RoomSummaryDetails` and `RoomProxy`.
- Add a new `RoomAvatar` enum that contains either
    - The room's ID, name and avatar URL.
    - The room's heroes for a 1:1 DM where the avatar is nil.
- Add a new `RoomAvatarImage` and use this everywhere instead of `LoadableAvatarImage` for rooms.
- Move the stacked avatar layout from `AvatarHeaderView` into the `RoomAvatarImage` in preparation for clustering up to 4 hero avatars.

Fixes #1659


ElementX/Sources/Mocks/RoomProxyMock.swift
```diff
@@ -54,6 +54,7 @@ extension RoomProxyMock {
         id = configuration.id
         name = configuration.name
         topic = configuration.topic
+        avatar = .room(id: configuration.id, name: configuration.name, avatarURL: configuration.avatarURL) // TODO: What happens for the logic in here?
```
stefanceriu, line 57: What's this about? 🤔

ElementX/Sources/Other/SwiftUI/Views/AvatarHeaderView.swift
```diff
@@ -17,37 +17,18 @@
...
-        }
-        
-        init(from user: UserProfileProxy) {
-            id = user.userID
-            name = user.displayName
-            avatarURL = user.avatarURL
-        }
+    private enum AvatarInfo {
+        case room(RoomAvatar)
+        case user(UserProfileProxy)
```
stefanceriu, line 22: 👌

stefanceriu (approved):
Very nice and great cleanup! 👏

ElementX/Sources/Mocks/RoomProxyMock.swift
pixlwave: Ah, I was unsure if it was worth copying the `RoomProxy` logic into the mock. I don't think it is, it won't prove anything on tests. I've replace it with a comment. Maybe one day we'll remove RoomProxyMock all together an use `RoomProxy` with an SDK `RoomMock`.

stefanceriu: Sounds like a plan 👌

PR #2901, opened 2024-06-05
Generate mocks for all SDK classes and introduce a simple TimelineIte…

stefanceriu:
…mFactory test to try them out


Velin92 (approved):

pixlwave (approved):
There's a *lot* going on here, but on the surface looks good to me. Love that we can finally do this 👏

PR #2903, opened 2024-06-06
Adopt Rust's new sessionPath that replaces the basePath

pixlwave:

- Hard code the OIDC `registrations.json` file into the `OIDCConfiguration` as the SDK no longer uses a base path for this.
- Generate a session path when creating an `AuthenticationService` or `QRCodeLoginService` using a UUID as sub directory inside of the base path.
- Stores this path inside the `RestorationToken` after successful login, along with a fallback mechanism that decodes tokens without this property using the same logic we had for session restoration.
- Uses this path everywhere where we previously used the logic mentioned above.

Depends on https://github.com/matrix-org/matrix-rust-sdk/pull/3514

### DIFF
--- a/ElementX/Sources/Application/AppSettings.swift
+++ b/ElementX/Sources/Application/AppSettings.swift
@@ -161,7 +161,8 @@ final class AppSettings {
                                                                      tosURI: acceptableUseURL,
                                                                      policyURI: privacyURL,
                                                                      contacts: [supportEmailAddress],
-                                                                     staticRegistrations: oidcStaticRegistrations.mapKeys { $0.absoluteString })
+                                                                     staticRegistrations: oidcStaticRegistrations.mapKeys { $0.absoluteString },
+                                                                     dynamicRegistrationsFile: .sessionsBaseDirectory.appending(path: "oidc/registrations.json"))
 
     /// A dictionary of accounts that have performed an initial sync through their proxy.
     ///
--- a/ElementX/Sources/Services/UserSession/UserSessionStore.swift
+++ b/ElementX/Sources/Services/UserSession/UserSessionStore.swift
@@ -27,21 +27,16 @@ class UserSessionStore: UserSessionStoreProtocol {
     /// All the user IDs managed by the store.
     var userIDs: [String] { keychainController.restorationTokens().map(\.userID) }
     
-    /// The base directory where all session data is stored.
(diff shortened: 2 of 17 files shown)


ElementX/Sources/Application/AppSettings.swift
```diff
@@ -161,7 +161,8 @@ final class AppSettings {
                                                                      tosURI: acceptableUseURL,
                                                                      policyURI: privacyURL,
                                                                      contacts: [supportEmailAddress],
-                                                                     staticRegistrations: oidcStaticRegistrations.mapKeys { $0.absoluteString })
+                                                                     staticRegistrations: oidcStaticRegistrations.mapKeys { $0.absoluteString },
+                                                                     dynamicRegistrationsFile: .sessionsBaseDirectory.appending(path: "oidc/registrations.json"))
```
pixlwave, line 165: This `base_path/oidc/registrations.json` matches what the SDK previously did so there's no need for a migration here.

ElementX/Sources/Services/UserSession/UserSessionStore.swift
```diff
@@ -59,20 +54,21 @@ class UserSessionStore: UserSessionStoreProtocol {
             MXLog.error("Failed restoring login with error: \(error)")
             
             // On any restoration failure reset the token and restart
-            keychainController.removeAllRestorationTokens()
```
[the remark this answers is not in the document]
stefanceriu: Upsie, good catch! 😅

Package.swift
```diff
@@ -13,7 +13,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMinor(from: "1.4.0")),
-        .package(url: "https://github.com/element-hq/swift-command-line-tools.git", revision: "a6ad90808f4f6cac615ab8496c6ff1bc5f9fa192"),
+        .package(url: "https://github.com/element-hq/swift-command-line-tools.git", revision: "e5eaab1558ef664e6cd80493f64259381670fb3a"),
```
pixlwave, line 16: Realised today that I hadn't updated this dependency to include the fix to make sure `ctrl + c` kills cargo when running `build-sdk`.

Velin92 (approved):

PR #84, opened 2022-06-17
Add GH labels to bug reports

ismailgulek:
Fixes #77

PR #2906, opened 2024-06-06
Update Compound.

pixlwave:
- Update Compound to the latest revision (the snapshots are no longer in Git LFS, see [#84](https://github.com/element-hq/compound-ios/pull/84)).
- Update Introspect and Prefire packages to match Compound (it would be nice if we could remove these from project.yml seeing as they're pulled in by Compound but I got an error about trying to add them in target.yml).
- Use the new markAsRead/Unread icons when long-pressing a room.

### DIFF
--- a/ElementX.xcodeproj/project.pbxproj
+++ b/ElementX.xcodeproj/project.pbxproj
@@ -7287,7 +7287,7 @@
 			repositoryURL = "https://github.com/BarredEwe/Prefire";
 			requirement = {
 				kind = upToNextMinorVersion;
-				minimumVersion = 2.0.4;
+				minimumVersion = 2.7.0;
 			};
 		};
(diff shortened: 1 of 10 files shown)


stefanceriu (approved):
Very nice! 👍

ElementX.xcodeproj/project.pbxproj
```diff
@@ -7287,7 +7287,7 @@
 			repositoryURL = "https://github.com/BarredEwe/Prefire";
 			requirement = {
 				kind = upToNextMinorVersion;
-				minimumVersion = 2.0.4;
+				minimumVersion = 2.7.0;
```
stefanceriu, line 7290: Seems this change breaks the preview tests -> `Switch to using iPhone15,2 for these tests.`
I'll just change the required version to `iPhone15` to avoid this in the future

PR #2912, opened 2024-06-11
Allow showing the developer options on any build type by tapping the …

stefanceriu:
…settings screen version footer 7 times (a la android)


stefanceriu:
Yeah, that's what the people want ¯\_(ツ)_/¯

PR #2915, opened 2024-06-11
Roll back prefire and swift snapshot to see if it fixes the preview tests

stefanceriu:


stefanceriu:
`2.6.0` failed. Rolled it all the way back to `2.0.4`

stefanceriu:
Still failing. I'm officially out of ideas as to why these tests started failing. I'll lower their percetual precision for now.

PR #2916, opened 2024-06-11
Reduce perceptual precision for snapshots that fail with a different …

stefanceriu:
…date format on the CI


stefanceriu:
`0.7` was enough locally but still fails on CI. Closing this in favor of more investigation based on https://github.com/element-hq/element-x-ios/pull/2917

PR #2925, opened 2024-06-14
Allow timeline messages text selection on both the Mac and on iOS (no…

stefanceriu:
…t on long press but a double tap)

- remove double tap to show the emoji picker as it clashes with the selection and was introduced before the rich timeline item menu was introduced


wrjlewis:
Just tried this on nightly and it's awesome 🤩

stefanceriu:
🥳

PR #2931, opened 2024-06-17
Bump the RustSDK to v1.0.14

stefanceriu:


ElementX/Sources/Services/Timeline/TimelineProxy.swift
```diff
@@ -78,6 +78,7 @@ final class TimelineProxy: TimelineProxyProtocol {
         MXLog.info("Paginating backwards")
         
         do {
+            let _ = try await timeline.paginateBackwards(numEvents: requestSize)
             let _ = try await timeline.paginateBackwards(numEvents: requestSize)
```
pixlwave, line 82: Errr, are you sure? 😅

pixlwave (approved):
LGTM other than that 👆

ElementX/Sources/Services/Timeline/TimelineProxy.swift
stefanceriu: Crap, got left there from my debugging session with Ivan 🤦‍♂️

PR #2937, opened 2024-06-18
Fixes #2679 - Timestamps in room list inconsistently use absolute and…

### LINKED ISSUE #2679: Timestamps in room list inconsistently use absolute and relative formatting
### Steps to reproduce

1. Have a room in the room list where the latest event (that's shown in the room list) was yesterday but less than 24 hours ago.
2. Have another room in the room list where the latest relevant event was today.
3. Look at the timestamp for these room in the room list

This is very easy to reproduce relatively early in the morning with a medium size account.

#### What did you expect?
Consistently formatted timestamps (always "x hours ago",  "yesterday", ... or always "hh:mm", "yesterday", ...)

#### What happened instead?
Room timestamps mix absolute and relative representation for "recent" events.
In the screenshot, the top three rooms show the absolute time because their latest relevant events were "today" (when I took the screenshot). The three rooms below that show a relative timestamp because their latest relevant events were yesterday but are younger than 24 hours. The remaining rooms show "yesterday", because their latest relevant events were yesterday, and they are older than 24 hours.
IMG_3857-obfuscated

Note that "yesterday" is obviously a relative timestamp as well, however, this feels correct and is also what Apple's apps (at least mail and messages) do. (They do "hh:mm", "yesterday", "day of the week", "date".)

### Additional context
I [asked this](https://matrix.to/#/!amwlFMejXQcoCxNYHH:matrix.org/$TivHao3J2dHBsfDkODvFFMnhq0s6E1JNQJZB2-g4pFg?via=matrix.org&via=beeper.com&via=element.io) in the matrix room which resulted in a brief discussion with @pixlwave and @stefanceriu. From this discussion I take that this is [known but unintentional behavior](https://matrix.to/#/!amwlFMejXQcoCxNYHH:matrix.org/$VmLTuhOGIJ0zhfUkn70EDYDOjAlxPISAJmehAJOL-bo?via=matrix.org&via=beeper.com&via=element.io) and I was asked to open an issue.

I was bored, so I looked at the code to see if something jumps out at me, and I think I found the source of this behavior. I never wrote any meaningful amount of Swift, though, let alone SwiftUI, so take this with a grain of salt. Also, I don't know if this is all already known, so please ignore this paragraph if it is.
The function [`Date.formattedMinimal`](https://github.com/element-hq/element-x-ios/blob/eda7d595181257f053f89a574707809a95e5ecec/ElementX/Sources/Other/Extensions/Date.swift#L21-L42) is used [here](https://github.com/element-hq/element-x-ios/blob/eda7d595181257f053f89a574707809a95e5ecec/ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift#L231) to format timestamps for the room list.
It delegates all formatting for dates that are "yesterday" to the Foundation framework, which seemingly was intended as a way to localize the string "yesterday". Instead, however, the `Date.RelativeFormatStyle` formats dates as "x hours ago" if they are less than 24 hours ago. In combination with the first branch in that function that forces "hh:mm" formatting for dates from "today" this produces the described behavior.

Date formatting is also inconsistent across features in the app. `Date.formattedMinimal` is also used for the [reactions](https://github.com/element-hq/element-x-ios/blob/eda7d595181257f053f89a574707809a95e5ecec/ElementX/Sources/Screens/RoomScreen/View/Supplementary/ReactionsSummaryView.swift#L132), but not for [read receipts](https://github.com/element-hq/element-x-ios/blob/eda7d595181257f053f89a574707809a95e5ecec/ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemFactory.swift#L469) (I hope I found the right locations).

### Your phone model

iPhone 13 mini / iPad Pro

### Operating system version

iOS/iPadOS 17.4.1

### Application version

1.6.2 (102)

### Homeserver

Synapse 1.104.0 / Proxy 0.99.15

### Will you send logs?

No

stefanceriu:
… relative formatting

### DIFF
--- a/UnitTests/Sources/DateTests.swift
+++ b/UnitTests/Sources/DateTests.swift
@@ -31,6 +31,9 @@ class DateTests: XCTestCase {
         let yesterday = calendar.date(byAdding: .hour, value: 1, to: startOfYesterday)!
         XCTAssertEqual(yesterday.formattedMinimal(), yesterday.formatted(Date.RelativeFormatStyle(presentation: .named, capitalizationContext: .beginningOfSentence)))
         
+        let nearYesterday = calendar.date(byAdding: DateComponents(hour: -10), to: today)!
+        XCTAssertEqual(nearYesterday.formattedMinimal(), yesterday.formatted(Date.RelativeFormatStyle(presentation: .named, capitalizationContext: .beginningOfSentence)))
+        
         let threeDaysAgo = calendar.date(byAdding: .day, value: -3, to: startOfToday)!
         XCTAssertEqual(threeDaysAgo.formattedMinimal(), threeDaysAgo.formatted(.dateTime.weekday(.wide)))
         
--- a/ElementX/Sources/Other/Extensions/Date.swift
+++ b/ElementX/Sources/Other/Extensions/Date.swift
@@ -26,7 +26,8 @@ extension Date {
             return formatted(date: .omitted, time: .shortened)
         } else if calendar.isDateInYesterday(self) {
             // Simply "Yesterday" if it was yesterday.
-            return formatted(Date.RelativeFormatStyle(presentation: .named, capitalizationContext: .beginningOfSentence))
+            guard let yesterday = Calendar.current.date(byAdding: .day, value: -1, to: Date()) else { fatalError() }
+            return yesterday.formatted(Date.RelativeFormatStyle(presentation: .named, capitalizationContext: .beginningOfSentence))
         } else if let sixDaysAgo = calendar.date(byAdding: .day, value: -6, to: calendar.startOfDay(for: .now)),
                   sixDaysAgo <= self {
             // The named day if it was in the last 6 days.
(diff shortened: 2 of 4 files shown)


UnitTests/Sources/DateTests.swift
```diff
@@ -31,6 +31,9 @@ class DateTests: XCTestCase {
         let yesterday = calendar.date(byAdding: .hour, value: 1, to: startOfYesterday)!
         XCTAssertEqual(yesterday.formattedMinimal(), yesterday.formatted(Date.RelativeFormatStyle(presentation: .named, capitalizationContext: .beginningOfSentence)))
         
+        let nearYesterday = calendar.date(byAdding: DateComponents(hour: -10), to: today)!
+        XCTAssertEqual(nearYesterday.formattedMinimal(), yesterday.formatted(Date.RelativeFormatStyle(presentation: .named, capitalizationContext: .beginningOfSentence)))
+
```
pixlwave, line 36: This feels like its going to be a fragile test if run after 10am?

pixlwave (approved):
👌

UnitTests/Sources/DateTests.swift
stefanceriu: Hmm.. why? `today` is always 9:30
(also, I happened to raise this pr after 10am 😁)

pixlwave: 🤦‍♂️ I got `today` and `startOfToday` mixed up.

PR #2941, opened 2024-06-19
Update SDK

Velin92:
this also implements the usage of a timeline independent reply and edit API

### DIFF
--- a/ElementX/Sources/Services/Timeline/TimelineProxy.swift
+++ b/ElementX/Sources/Services/Timeline/TimelineProxy.swift
@@ -126,8 +126,8 @@ final class TimelineProxy: TimelineProxyProtocol {
               intentionalMentions: IntentionalMentions) async -> Result<Void, TimelineProxyError> {
         MXLog.info("Editing timeline item: \(timelineItemID)")
         
-        guard let eventTimelineItem = await timelineProvider.itemProxies.firstEventTimelineItemUsingID(timelineItemID) else {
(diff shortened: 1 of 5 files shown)


pixlwave (approved):

ElementX/Sources/Services/Timeline/TimelineProxy.swift
```diff
@@ -126,8 +126,8 @@ final class TimelineProxy: TimelineProxyProtocol {
               intentionalMentions: IntentionalMentions) async -> Result<Void, TimelineProxyError> {
         MXLog.info("Editing timeline item: \(timelineItemID)")
         
-        guard let eventTimelineItem = await timelineProvider.itemProxies.firstEventTimelineItemUsingID(timelineItemID) else {
-            MXLog.error("Unknown timeline item: \(timelineItemID)")
+        guard let eventID = timelineItemID.eventID else {
+            MXLog.error("Missing event ID")
             return .failure(.failedEditing)
         }
```
Velin92, line 132: @stefanceriu is it possible to edit local echos ? or that case is handled by another function?

stefanceriu: It is not possible, the SDK side `MatrixRustSDK.EventTimelineItem.isEditable()` returns false for local echoes

Velin92: perfect!

PR #2944, opened 2024-06-19
Avoid creating multiple clients if the NSE is invoked more than once …

stefanceriu:
…on the same process

This came out of a discussion with @bnjbvr from which I quote: 

>it's of uttermost importance that there's at most one Client per process ... We put all our context in there, viz. locks, viz. OIDC data, so there needs to be at most one per different account

### DIFF
--- a/NSE/Sources/NotificationServiceExtension.swift
+++ b/NSE/Sources/NotificationServiceExtension.swift
@@ -51,8 +51,9 @@ class NotificationServiceExtension: UNNotificationServiceExtension {
     private var modifiedContent: UNMutableNotificationContent?
     
     // Used to create one single UserSession across process/instances/runs
+    private static let serialQueue = DispatchQueue(label: "io.element.elementx.nse")
     private static var userSession: NSEUserSession?
-
+    
     override func didReceive(_ request: UNNotificationRequest,
                              withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void) {
         guard !DataProtectionManager.isDeviceLockedAfterReboot(containerURL: URL.appGroupContainerDirectory),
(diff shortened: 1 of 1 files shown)


Velin92 (approved):

NSE/Sources/NotificationServiceExtension.swift
```diff
@@ -75,6 +78,17 @@ class NotificationServiceExtension: UNNotificationServiceExtension {
         MXLog.info("\(tag) Payload came: \(request.content.userInfo)")
 
         Task {
+            if Self.userSession == nil {
+                // This function might be run concurrently and from different processes
+                // It's imperative that we create **at most** one UserSession/Client per process
```
bnjbvr, line 83: Is it possible that two callers both see the `userSession` is nil, and race creating one user session each? (in which case having a lock for creating the user session would be sufficient to prevent that from happening)

bnjbvr (commented):
Sweet, thanks!

NSE/Sources/NotificationServiceExtension.swift
stefanceriu: Good point, I added a little something

NSE/Sources/NotificationServiceExtension.swift
```diff
@@ -49,7 +49,15 @@ private let keychainController = KeychainController(service: .sessions,
...
     private var modifiedContent: UNMutableNotificationContent?
-
+    
+    // Used to create one single UserSession across process/instances/runs
+    private static let serialQueue = DispatchQueue(label: "io.element.elementx.nse")
+    private static var innerUserSession: NSEUserSession?
+    private static var userSession: NSEUserSession? {
+        get { serialQueue.sync { innerUserSession } }
+        set { serialQueue.sync { innerUserSession = newValue } }
+    }
```
Velin92, line 59: Even if the assignment will run sync and will be safe, we can't guarantee this will happen immediately since this is till dispatched on the task, it will happen still when the queue can serve it. I would use another solution

NSE/Sources/NotificationServiceExtension.swift
```diff
@@ -75,6 +83,17 @@ class NotificationServiceExtension: UNNotificationServiceExtension {
         MXLog.info("\(tag) Payload came: \(request.content.userInfo)")
 
         Task {
+            if Self.userSession == nil {
+                // This function might be run concurrently and from different processes
+                // It's imperative that we create **at most** one UserSession/Client per process
+                do {
+                    Self.userSession = try await NSEUserSession(credentials: credentials, clientSessionDelegate: keychainController)
```
Velin92, line 90: we could use a sync Task function, something like this:

```
extension Task where Failure == Error {
    /// Performs an async task in a sync context.
    ///
    /// - Note: This function blocks the thread until the given operation is finished. The caller is responsible for managing multithreading.
    static func synchronous(priority: TaskPriority? = nil, operation: @escaping @Sendable () async throws -> Success) {
        let semaphore = DispatchSemaphore(value: 0)

        Task(priority: priority) {
            defer { semaphore.signal() }
            return try await operation()
        }

        semaphore.wait()
    }
    
    do the assignment of the session synchronously inside this block and wait for it so we will be sure that the user session is definitely assigned when will be checked later otherwise we may risk the sync block to run later than the async `run` block

PR #2954, opened 2024-06-21
Use a Client in AuthenticationService now that the Rust one has been removed.

pixlwave:
Rust's AuthenticationService has been removed in https://github.com/matrix-org/matrix-rust-sdk/pull/3594. This PR moves what was very simple logic that called into a `Client` directly into our app-side AuthenticationService.

### DIFF
--- a/ElementX/Sources/Other/Extensions/ClientBuilder.swift
+++ b/ElementX/Sources/Other/Extensions/ClientBuilder.swift
@@ -1,0 +1,42 @@
+//
+// Copyright 2024 New Vector Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+import MatrixRustSDK
+
+extension ClientBuilder {
+    /// A helper method that applies the common builder modifiers needed for the app.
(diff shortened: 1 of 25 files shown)


ElementX/Sources/Other/Extensions/ClientBuilder.swift
```diff
@@ -0,0 +1,42 @@
...
+                .autoEnableBackups(autoEnableBackups: true)
+        }
+        
+        if let httpProxy {
+            builder = builder.proxy(url: httpProxy)
+        }
+        
+        return builder
+    }
+}
```
Velin92, line 42: I like this and would like to know if it would be possible to create a similiar function also for the qr one that we use in the qr code login service.
Also maybe we should store the versions as a `static let` in the extensions, so that they can easily be changed for all the client builders usage.

ElementX/Sources/Services/QRCode/QRCodeLoginService.swift
```diff
@@ -53,14 +53,14 @@ final class QRCodeLoginService: QRCodeLoginServiceProtocol {
...
+                             sessionDelegate: userSessionStore.clientSessionDelegate)
                 .sessionPath(path: sessionDirectory.path(percentEncoded: false))
                 .passphrase(passphrase: passphrase)
-                .userAgent(userAgent: UserAgentBuilder.makeASCIIUserAgent())
-                .enableCrossProcessRefreshLock(processId: InfoPlistReader.main.bundleIdentifier,
-                                               sessionDelegate: userSessionStore.clientSessionDelegate)
-                .serverVersions(versions: ["v1.0", "v1.1", "v1.2", "v1.3", "v1.4", "v1.5"]) // FIXME: Quick and dirty fix for stopping version requests on startup https://github.com/matrix-org/matrix-rust-sdk/pull/1376
-                .buildWithQrCode(qrCodeData: qrData, oidcConfiguration: oidcConfiguration, progressListener: listener)
+                .requiresSlidingSync()
+                .buildWithQrCode(qrCodeData: qrData, oidcConfiguration: appSettings.oidcConfiguration.rustValue, progressListener: listener)
```
Velin92, line 63: my bad I now noticed this, but I see the serverVersions are missing from the base builder

Velin92 (approved):
LGTM

PR #2956, opened 2024-06-21
Fix ui tests that use the composer draft service

Velin92:


ElementX/Sources/Mocks/ComposerDraftServiceMock.swift
```diff
@@ -0,0 +1,29 @@
...
+//
+
+import Foundation
+
+struct ComposerDraftServiceMockConfiguration {
+    var draft: ComposerDraftProxy?
+}
+
+extension ComposerDraftServiceMock {
+    convenience init(config: ComposerDraftServiceMockConfiguration) {
```
stefanceriu, line 24:
```suggestion
    convenience init(_ config: ComposerDraftServiceMockConfiguration) {
```

So that it matches the other Mocks

PR #2965, opened 2024-06-24
Improve UX when answering EL-Call through the notification

Velin92:
- The navigation is now received on DispatchQueue.main since it was triggerd by background threads, and creating a possible deadlock.
- When a call is answered the 15 seconds timeout is cancelled.

This will also solve the issue that makes the call screen when the device is locked disappear after some seconds (which as of right now only works when answering the call with the push notification)


pixlwave (approved):
LGTM 👌

PR #2966, opened 2024-06-25
Replace Towncrier with GitHub releases + labels.

pixlwave:
We weren't using Towncrier very well so this PR replaces it with the use of GitHub's generated release notes along with a configuration that will categorise each entry based upon one of the following labels being applied:

Labels added here: https://github.com/element-hq/element-x-ios/labels?q=pr-

These labels can be added after the PR is merged, so long as they're there *before* the release is made.

The second commit removes Towncrier. I made a [second repo](https://github.com/pixlwave/Release-Notes-Test) to figure the flow out, but not sure how to validate the changes on EX other than wait for the next release and tidy up anything that comes out of that.


Velin92 (approved):
LGTM but what happens if a pr has multiple labels? Would it appear under multiple categories?

stefanceriu (approved):
👏

pixlwave:
> LGTM but what happens if a pr has multiple labels? Would it appear under multiple categories?

This [PR](https://github.com/pixlwave/Release-Notes-Test/pull/19) had 3 tags, and the first one (I assume from the release.yml ordering) [got chosen](https://github.com/pixlwave/Release-Notes-Test/releases/tag/v0.20.0).

PR #2967, opened 2024-06-25
Cleanup how we setup the CallKit provider and have it be used for outgoing calls as well

stefanceriu:
…going calls as well


ElementX/Sources/Services/ElementCall/ElementCallService.swift
```diff
@@ -85,6 +84,10 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
                 MXLog.error("Failed transaction with error: \(error)")
             }
         }
+        
+        self.ongoingCallID = nil
+        ongoingCallRoomID = nil
+        callProvider = nil
```
pixlwave, line 90: Would it make sense to wrap these 3 in a single type where they're all non-nil?

ElementX/Sources/Services/ElementCall/ElementCallService.swift
```diff
@@ -46,15 +46,14 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
         pushRegistry.desiredPushTypes = [.voIP]
     }
     
-    func setupCallSession(title: String) async {
-        guard ongoingCallID == nil else {
-            return
-        }
```
pixlwave, line 52: Presumably by removing this if you try join another call while one is ongoing we switch to the new one automatically?

ElementX/Sources/Services/ElementCall/ElementCallService.swift
```diff
@@ -44,47 +56,30 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
         
         pushRegistry.delegate = self
         pushRegistry.desiredPushTypes = [.voIP]
+        
+        callProvider.setDelegate(self, queue: DispatchQueue.main)
```
Velin92, line 60: i think we can just use bg queue, to execute delegate methods this should make everything faster and not block UI if not needed, the important thing is that whatever listens to changes made by the delegate is on the main queue

Velin92 (approved):
LGTM just a few comments

ElementX/Sources/Application/Navigation/AppRoutes.swift
```diff
@@ -44,8 +44,8 @@ enum AppRoute: Equatable {
     case childEventOnRoomAlias(eventID: String, alias: String)
     /// The profile of a matrix user (outside of a room).
     case userProfile(userID: String)
-    /// An Element Call running in a particular room
-    case call(roomID: String)
+    /// An Element Call running in a particular room. Nil if having to end the call
```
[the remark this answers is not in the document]
stefanceriu: I didn't want to expose a completely different method for that but I just realised that I can listen to the action directly in the user session flow coordinator.

ElementX/Sources/Services/ElementCall/ElementCallService.swift
```diff
@@ -44,47 +61,43 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
...
-            try AVAudioSession.sharedInstance().setActive(true, options: .notifyOthersOnDeactivation)
+            try await callController.request(CXTransaction(action: startCallAction))
         } catch {
-            MXLog.error("Failed setting up VoIP session with error: \(error)")
-            tearDownCallSession()
+            MXLog.error("Failed requesting start call action with error: \(error)")
         }
+        
+        try? AVAudioSession.sharedInstance().setCategory(.playAndRecord, mode: .videoChat, options: [])
+        try? AVAudioSession.sharedInstance().setActive(true, options: .notifyOthersOnDeactivation)
```
pixlwave, line 96: Could we add a comment why these are outside of the `do try`? E.g. it seems unexpected to me that we would set these if the call failed.

ElementX/Sources/Services/ElementCall/ElementCallService.swift
```diff
@@ -147,29 +148,55 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
...
         } else {
-            MXLog.error("Failed answering incoming call, missing room ID")
+            MXLog.error("Failed answering incoming call, missing incomingCallID")
         }
         
         action.fulfill()
     }
     
+    func provider(_ provider: CXProvider, perform action: CXSetMutedCallAction) {
+        // Forward this to the widget somehow
```
pixlwave, line 173: This might be helpful: https://github.com/pixlwave/Element-Call-macOS/blob/6407a23458ae95b402922e4ebfdb357b4ecef819/Sources/VOIPCoordinator.swift#L67-L73

pixlwave (approved):
Very nice 👌

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
```diff
@@ -128,15 +128,15 @@ class CallScreenViewModel: CallScreenViewModelType, CallScreenViewModelProtocol
     
     private func hangUp() async {
         let hangUpMessage = """
-        "api":"toWidget",
+        {"api":"fromWidget",
```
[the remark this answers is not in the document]
stefanceriu: Yeah, both were needed to fix the error ¯\_(ツ)_/¯

ElementX/Sources/Services/ElementCall/ElementCallService.swift
stefanceriu: Nice but the bigger problem here is how to get the call service to talk to the active widget

ElementX/Sources/Services/ElementCall/ElementCallService.swift
stefanceriu: Well failing to setup a CallKit session doesn't mean ElementCall isn't running or that the audio session shouldn't get set up, guess I might as well treat their errors separately

pixlwave: Right, yeah ok got it.

PR #2968, opened 2024-06-26
Hide timeline style selection

Velin92:
First part of https://github.com/element-hq/element-x-ios/issues/2964

This will default the timeline style to always be bubble, and will hide to the user the possibility of switching to plain mode.
This PR won't fully remove the plain text code since doing so is a delicate process, which we will do in a follow up PR.


stefanceriu (approved):
👌

PR #2969, opened 2024-06-26
Stop the timeline from requesting back pagination whilst loading existing items.

pixlwave:
This PR makes 3 changes (and is probably easiest to review commit by commit):
- Detached timelines were erroring due to attempting to subscribe to back pagination status (apparently no longer supported by the SDK).
- Make sure we await the loading of existing timeline items now that Rust doesn't provide them to us when setting up the subscription - fixes a bug where an error would be shown when navigating to an event permalink.
- Even once the above was fixed, the TimelineTableViewController could sometimes attempt to back paginate whilst loading the existing items. Now we use an initial pagination state that will prevent this from happening until the items have loaded.


Velin92 (approved):
LGTM

PR #2971, opened 2024-06-27
Allow Element Call's widget URL to be configured by the homeserver.

### LINKED ISSUE #2911: [Task] Make Element Call widget URL configurable

stefanceriu:
- use the SDK to make a GET requests and parse the response into an Element specific well-known settings struct
- use those values to setup the ElementCall baseURL
- keep the call baseURL override but separate it from the default URL and have it be nil by default

Closes #2911.

### DIFF
--- a/ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
+++ b/ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
@@ -38,9 +38,11 @@ class CallScreenViewModel: CallScreenViewModelType, CallScreenViewModelProtocol 
     ///   - callBaseURL: Which Element Call instance should be used
     ///   - clientID: Something to identify the current client on the Element Call side
     init(elementCallService: ElementCallServiceProtocol,
+         clientProxy: ClientProxyProtocol,
          roomProxy: RoomProxyProtocol,
-         callBaseURL: URL,
-         clientID: String) {
+         clientID: String,
+         elementCallBaseURL: URL,
+         elementCallBaseURLOverride: URL?) {
         self.elementCallService = elementCallService
         self.roomProxy = roomProxy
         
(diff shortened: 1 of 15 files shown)


ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/View/DeveloperOptionsScreen.swift
```diff
@@ -38,14 +38,14 @@ struct DeveloperOptionsScreen: View {
             }
                                     
             Section("Element Call") {
-                TextField(context.elementCallBaseURL.absoluteString, text: $elementCallBaseURLString)
+                TextField(context.elementCallBaseURLOverride?.absoluteString ?? "", text: $elementCallBaseURLString)
```
pixlwave, line 41: Might be nice to leave this placeholder as the real base URL?

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
```diff
@@ -90,7 +92,15 @@ class CallScreenViewModel: CallScreenViewModelType, CallScreenViewModelProtocol
...
         
         Task {
-            switch await widgetDriver.start(baseURL: callBaseURL, clientID: clientID) {
+            let baseURL = if let elementCallBaseURLOverride {
+                elementCallBaseURLOverride
+            } else if case .success(let wellKnown) = await clientProxy.getElementWellKnown(), let wellKnownCallWidgetURL = wellKnown?.call?.widgetURL {
+                wellKnownCallWidgetURL
+            } else {
+                elementCallBaseURL
+            }
```
pixlwave, line 101: Would it make sense to put this logic in ElementCallService?

ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/View/DeveloperOptionsScreen.swift
pixlwave: On a bigger level, do we need this setting any more now we can discover any call URL?

pixlwave (approved):
Nice, LGTM 📞

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
stefanceriu: Honestly it's all the same to me ¯\_(ツ)_/¯

ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/View/DeveloperOptionsScreen.swift
stefanceriu: We left in there so that the VoIP team can easily change what instance they use.

stefanceriu: > Might be nice to leave this placeholder as the real base URL

Yeah, good point 👍

PR #2972, opened 2024-06-27
Trigger a Danger CI run when labelling a PR too.

pixlwave:
Copies the same triggers as we have on the PR Build workflow.

Will test by adding the label to this PR after it fails ✅


stefanceriu (approved):
🚀

PR #2975, opened 2024-06-27
Various CallKit and ElementCall fixes

stefanceriu:
- fixes video stream problems
- hooks up muting to call kit but keeps it disabled for now as EC doesn't expose a way to forward the events yet
- infers remote hangups and cancels local ringing

### DIFF
--- a/ElementX/Sources/Services/Room/RoomProxy.swift
+++ b/ElementX/Sources/Services/Room/RoomProxy.swift
@@ -155,6 +155,17 @@ class RoomProxy: RoomProxyProtocol {
         subscribeToTypingNotifications()
     }
     
+    func subscribeToRoomInfoUpdates() {
+        guard roomInfoObservationToken == nil else {
+            return
+        }
+        
+        roomInfoObservationToken = room.subscribeToRoomInfoUpdates(listener: RoomInfoUpdateListener { [weak self] in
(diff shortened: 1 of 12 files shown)


ElementX/Sources/Services/Room/RoomProxy.swift
```diff
@@ -155,6 +155,13 @@ class RoomProxy: RoomProxyProtocol {
         subscribeToTypingNotifications()
     }
     
+    func subscribeToRoomInfoUpdates() {
+        roomInfoObservationToken = room.subscribeToRoomInfoUpdates(listener: RoomInfoUpdateListener { [weak self] in
```
pixlwave, line 159: Maybe this should guard that `roomInfoObservationToken == nil` or similar before continuing?

pixlwave (approved):
LGTM 👍

ElementX/Sources/Services/Room/RoomProxy.swift
stefanceriu: I initially figured this should really do anything if not guarded but now that you mentioned it it might produce unwanted side effects on the rust side.

On the other hand I'm wondering, should we always subscribe to this as soon as the RoomProxy is created? It's really weird to hook into that publisher and have nothing coming in because you forgot to call this 🤔

PR #2978, opened 2024-06-27
Use local room list sorting from Rust.

stefanceriu:
- adopt the new SDK side local room list sorting API
- replace the old RoomSummary enum with plain details


pixlwave (approved):
Seems decent to me. I guess my comments are in the "Tidy up" commit 😅

PR #2980, opened 2024-06-27
Remove plain style layout

Velin92:
The first part of the plain style removal PR, this only removes the plain case and all the tests and code to related to plain, but does not remove the styling abstraction yet. That will be done in a separate PR so that we can always revert it in case we would need a new style.


ElementX/Sources/Screens/RoomScreen/View/Supplementary/TimelineItemStatusView.swift
```diff
@@ -51,13 +51,7 @@ struct TimelineItemStatusView: View {
                 TimelineDeliveryStatusView(deliveryStatus: .sent)
             }
         case .sendingFailed:
-            // Bubbles handle the failure internally, but plain items don't.
```
pixlwave, line 54: Might be good to leave the comment until we've tidied up the next part.

pixlwave (approved):
👏

pixlwave:
Maybe the title could be a bit more descriptive for a changelog entry though 🤔

PR #2982, opened 2024-06-27
Remove layout styling abstraction

Velin92:


ElementX/Sources/Screens/RoomScreen/View/Style/TimelineStyler.swift
```diff
@@ -20,8 +20,6 @@ import SwiftUI
 // MARK: - TimelineStyler
 
 struct TimelineStyler<Content: View>: View {
```
Velin92, line 22: @pixlwave  Maybe we could rename this to `TimelineStylerContainer` and the `TimelineBubbleStylerView` to `TimelineStyler`

pixlwave: No strong feelings, although it's nice to type `bubble` when searching for the part of the code where message bubbles are made.

pixlwave (approved):
Thank you!

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineStyler.swift
Velin92: then let's leave it as it is

PR #2984, opened 2024-06-28
Setup unit tests for Enterprise.

pixlwave:
Adds an extra workflow that runs the unit tests with the Enterprise configuration.


Velin92 (approved):
LGTM

PR #2985, opened 2024-06-28
Setup Sentry instrumentation on top of the existing Signposter

stefanceriu:


ElementX/Sources/Services/BugReport/BugReportService.swift
```diff
@@ -22,23 +22,21 @@ import UIKit
...
-    private let sentryURL: URL
     private let applicationId: String
     private let sdkGitSHA: String
     private let maxUploadSize: Int
     private let session: URLSession
-    private var lastCrashEventId: String?
     private let progressSubject = PassthroughSubject<Double, Never>()
     private var cancellables = Set<AnyCancellable>()
     
+    var lastCrashEventId: String?
```
pixlwave, line 32: 😉
```suggestion
    var lastCrashEventID: String?
```

ElementX/Sources/Services/Analytics/AnalyticsService.swift
```diff
@@ -49,6 +49,7 @@ class AnalyticsService {
     init(client: AnalyticsClientProtocol, appSettings: AppSettings) {
         self.client = client
         self.appSettings = appSettings
+        signpost = Signposter(isDevelopmentBuild: AppSettings.isDevelopmentBuild)
```
pixlwave, line 52: Couldn't this init have been left on the variable itself?

ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
```diff
@@ -75,6 +75,7 @@ struct HomeScreen: View {
                     gradientView.alpha = 1
                 }
             }
+            .sentryTrace("HomeScreen")
```
pixlwave, line 78: Unimportant, but should we define these somewhere together so we don't accidentally use the same one twice? Or use `"\(Self.self)"` assuming that works on a view.

ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -733,6 +733,51 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
         }
     }
     
+    private func setupSentry() {
+        SentrySDK.start { options in
+            #if DEBUG
+            options.enabled = false
+            #endif
+            
+            options.dsn = ServiceLocator.shared.settings.bugReportSentryURL.absoluteString
```
pixlwave, line 742: I think we have a local property for this one.
```suggestion
            options.dsn = appSettings.bugReportSentryURL.absoluteString
```

ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -733,6 +733,51 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
...
+            } else {
+                options.tracesSampleRate = 0.5
+            }
+            
+            // This callback is only executed once during the entire run of the program to avoid
+            // multiple callbacks if there are multiple crash events to send (see method documentation)
+            options.onCrashedLastRun = { event in
+                MXLog.error("Sentry detected a crash in the previous run: \(event.eventId.sentryIdString)")
+                ServiceLocator.shared.bugReportService.lastCrashEventId = event.eventId.sentryIdString
+            }
```
pixlwave, line 777: Might be nice to have a `Sentry Started` log line here so we have an idea if a rageshake has it enabled or not.

ElementX/Sources/Services/Analytics/Signposter.swift
```diff
@@ -49,37 +65,56 @@ class Signposter {
...
     
     func beginFirstRooms() {
         firstRoomsState = signposter.beginInterval(Name.firstRooms)
+        firstRoomsSpan = appStartupSpan.startChild(operation: "\(Name.firstRooms)")
     }
     
     func endFirstRooms() {
+        defer {
+            appStartupSpan.finish()
+        }
```
pixlwave, line 109: Does this mean all the other spans that are children are stopped if they happened to still be running?

pixlwave (approved):
Very nice! I think my only question is should we perhaps hash the server name instead of collecting it?

ElementX/Sources/Services/Analytics/AnalyticsService.swift
stefanceriu: Yeah, i initialy forgot I made that a static var

ElementX/Sources/Application/AppCoordinator.swift
stefanceriu: This way I don't need to access self in the block anymore

ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
stefanceriu: Yep okay, let's go with Self then

ElementX/Sources/Services/Analytics/Signposter.swift
stefanceriu: I would hope so but no idea honestly ¯\_(ツ)_/¯

PR #2996, opened 2024-07-02
Volatile draft to restore the composer after an edit.

Velin92:
save a volatile draft when editing a message, and restore it after the edit is completed or cancelled

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/RoomScreenInteractionHandler.swift
+++ b/ElementX/Sources/Screens/RoomScreen/RoomScreenInteractionHandler.swift
@@ -276,6 +276,7 @@ class RoomScreenInteractionHandler {
             text = messageTimelineItem.body
         }
         
+        // Always update the mode first and then the text so that the composer has time to save the text draft
         actionsSubject.send(.composer(action: .setMode(mode: .edit(originalItemId: messageTimelineItem.id))))
         actionsSubject.send(.composer(action: .setText(plainText: text, htmlText: htmlText)))
     }
--- a/ElementX/Sources/Screens/RoomScreen/RoomScreenInteractionHandler.swift
+++ b/ElementX/Sources/Screens/RoomScreen/RoomScreenInteractionHandler.swift
@@ -276,8 +276,8 @@ class RoomScreenInteractionHandler {
             text = messageTimelineItem.body
         }
         
-        actionsSubject.send(.composer(action: .setText(plainText: text, htmlText: htmlText)))
         actionsSubject.send(.composer(action: .setMode(mode: .edit(originalItemId: messageTimelineItem.id))))
+        actionsSubject.send(.composer(action: .setText(plainText: text, htmlText: htmlText)))
     }
     
     // MARK: Polls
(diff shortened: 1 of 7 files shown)


ElementX/Sources/Screens/RoomScreen/RoomScreenInteractionHandler.swift
```diff
@@ -276,8 +276,8 @@ class RoomScreenInteractionHandler {
             text = messageTimelineItem.body
         }
         
-        actionsSubject.send(.composer(action: .setText(plainText: text, htmlText: htmlText)))
         actionsSubject.send(.composer(action: .setMode(mode: .edit(originalItemId: messageTimelineItem.id))))
+        actionsSubject.send(.composer(action: .setText(plainText: text, htmlText: htmlText)))
```
pixlwave, line 280: Is this order important now? If so we should probably add a comment to note this and explain the reason.

ElementX/Sources/Services/ComposerDraft/ComposerDraftService.swift
```diff
@@ -71,4 +72,16 @@ final class ComposerDraftService: ComposerDraftServiceProtocol {
...
+        volatileDraft = draft
+    }
+    
+    func loadVolatileDraft() -> ComposerDraftProxy? {
+        volatileDraft
+    }
+    
+    func clearVolatileDraft() {
+        volatileDraft = nil
+    }
```
pixlwave, line 86: Should we combine these 2 together? Presumably if we only clear the volatile draft when leaving the room like we do for the disk based draft then the following can happen:
- Write message `a` but don't send.
- Edit message `b` and send.
- Finish message `a` (now restored) and send.
- Edit message `c` and send.
- The draft of message `a` will now be restored again which doesn't make sense.

pixlwave (approved):
Nice, LGTM, only 1 question about clearing.

ElementX/Sources/Screens/RoomScreen/RoomScreenInteractionHandler.swift
Velin92: yeah, because we want to first change the mode and then change the text, so that the mode change will internally save the previous text

ElementX/Sources/Services/ComposerDraft/ComposerDraftService.swift
Velin92: you mean that loading a volatile draft should also clear it?
The thing is that the last two steps, will call another save, and since the edit of message `c` will trigger a save and was done when `a` was already sent this means that the content of the draft was empty, and when the save is called on an empty state, it actually clears (which means saving an empty)

Velin92: However I just realised another bug... when we sand a... we restore a... so we definitely need to clear when loading the volatile one lol

pixlwave: > the edit of message `c` will trigger a save

Oh right yeah I forgot that part that part

> when we sand a... we restore a...

lol that's fun.

Velin92: fixed!

pixlwave (approved):
👍

PR #3000, opened 2024-07-05
Fix for the left room event using member instead of the sender

Velin92:
Member always contains the display name which is not always available for the sender display name sometimes.

This fix however will only be seen if the current SDK version includes: https://github.com/matrix-org/matrix-rust-sdk/pull/3648

### DIFF
--- a/ElementX/Sources/Services/Timeline/TimelineItems/RoomStateEventStringBuilder.swift
+++ b/ElementX/Sources/Services/Timeline/TimelineItems/RoomStateEventStringBuilder.swift
@@ -44,7 +44,7 @@ struct RoomStateEventStringBuilder {
         case .joined:
             return memberIsYou ? L10n.stateEventRoomJoinByYou : L10n.stateEventRoomJoin(senderDisplayName)
         case .left:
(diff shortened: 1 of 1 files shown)


ElementX/Sources/Services/Timeline/TimelineItems/RoomStateEventStringBuilder.swift
```diff
@@ -44,7 +44,7 @@ struct RoomStateEventStringBuilder {
         case .joined:
             return memberIsYou ? L10n.stateEventRoomJoinByYou : L10n.stateEventRoomJoin(senderDisplayName)
         case .left:
-            return memberIsYou ? L10n.stateEventRoomLeaveByYou : L10n.stateEventRoomLeave(senderDisplayName)
+            return memberIsYou ? L10n.stateEventRoomLeaveByYou : L10n.stateEventRoomLeave(member)
```
pixlwave, line 47: Probably worth adding a comment here that we're intentionally using `member` here in order to get their previous name.

pixlwave (approved):
👍

PR #3003, opened 2024-07-05
FIX: saving draft did not save the pill markdown content

Velin92:
This fix will allow the saved draft to contain the markdown content of a pill... however when restored we will still display the markdown only since the plain text composer still has no way to directly interpret and convert a markdown into a pill before its sent.


ElementX/Sources/Screens/RoomScreen/ComposerToolbar/ComposerToolbarViewModel.swift
```diff
@@ -349,7 +349,7 @@ final class ComposerToolbarViewModel: ComposerToolbarViewModelType, ComposerTool
         }
     }
     
-    private func sendPlainComposerText() {
+    private func getPlainComposerContent() -> PlainComposerContent {
```
pixlwave, line 352: Would imagine this can just be `plainComposerContent`.

pixlwave (approved):
Nice 👌

PR #3015, opened 2024-07-08
Update RTE to 2.37.5

Velin92:
This fixes a bug when a newline is set in a blockquote with a single that has been restored through html


pixlwave (approved):
👍

PR #3018, opened 2024-07-08
Restore Mentions in plain text mode

Velin92:
Will work both for draft and edit

### DIFF
--- a/UnitTests/Sources/ComposerToolbarViewModelTests.swift
+++ b/UnitTests/Sources/ComposerToolbarViewModelTests.swift
@@ -553,6 +553,63 @@ class ComposerToolbarViewModelTests: XCTestCase {
         await fulfillment(of: [expectation1, expectation2])
         XCTAssertEqual(viewModel.context.plainComposerText, NSAttributedString(string: "Hello world"))
     }
+    
+    func testRestoreUserMentionInPlainText() async throws {
+        viewModel.context.composerFormattingEnabled = false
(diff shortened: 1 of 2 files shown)


UnitTests/Sources/ComposerToolbarViewModelTests.swift
```diff
@@ -553,6 +553,78 @@ class ComposerToolbarViewModelTests: XCTestCase {
...
+        
+        viewModel.process(viewAction: .sendMessage)
+        try await deferred.fulfill()
+    }
+    
+    func testRestoreMixedMentionsInPlainText() async throws {
+        viewModel.context.composerFormattingEnabled = false
+        let text = "Hello [User1](https://matrix.to/#/@user1:matrix.org), [User2](https://matrix.to/#/@user2:matrix.org) and @room"
+        viewModel.process(roomAction: .setText(plainText: text, htmlText: nil))
+
```
pixlwave, line 597: Is it worth adding an assertion on the composer's plain text content here to make sure it restored the pills in the correct place?

pixlwave (approved):
Nice, LGTM 👌

UnitTests/Sources/ComposerToolbarViewModelTests.swift
pixlwave: Actually I guess the assertion on the plain text would go on the send action as that is where it will replace the range of the pill with the markdown.

PR #3040, opened 2024-07-15
Fix RTE framework signature error when building DEBUG builds for a device.

pixlwave:
Follow on from #3029 there is now a bug when trying to build locally for a device:

Screenshot 2024-07-15 at 3 12 32 PM

I tried using the condition of `"$ACTION" == "archive"` but that didn't work so for now just setting it to only run for release builds.


stefanceriu (approved):
👌

PR #3043, opened 2024-07-15
Update RTE with inline prediction working

Velin92:
This new version of RTE re-enables inline predictive text and fixes a bunch of issues with it.


pixlwave (approved):
🚀

PR #3051, opened 2024-07-17
Show Encryption Authenticity warnings on messages in the timeline.

BillCarsonFr:
This PR adds warnings to the timeline items in encrypted rooms when the authenticity can't be validated. It is behind a feature flag and doesn't surface the error message, only an icon. The icons are subject to change.

| Not guaranteed | Media | Unencrypted |
| - | - | - |
| Simulator Screenshot - iPhone 15 Pro - 2024-08-05 at 10 08 01 | Simulator Screenshot - iPhone 15 Pro - 2024-08-05 at 10 08 08 | Simulator Screenshot - iPhone 15 Pro - 2024-08-05 at 10 07 49 |

<details>
<summary>Original</summary>
Adds message shields (lax mode)
Depends on SDK change https://github.com/matrix-org/matrix-rust-sdk/pull/3679

Design: https://www.figma.com/design/2aqj9A5WRvGY3tcdTrBek4/Unencrypted-message-warning-in-Timeline?node-id=910-77038&t=pF5GGduo3EsrFG2Q-0

image
image

Note for reviewer / Open questions

- I am not familiar wth swift UI so looking for feedback on how to do that design correctly
- ~~I need to add a new Icon to the Compound Icons (plain shield OFF mode - strike through)~~
- How to handle a11y?
- The preview looks sometimes different than what I see on device, not sure how to cover all cases

</details>


ElementX/Sources/Services/Timeline/TimelineItemProxy.swift
```diff
@@ -120,10 +120,46 @@ class EventTimelineItemProxy {
...
+        case .grey(let message):
+            return MessageShield(color: .GRAY, message: translatableFromRawString(message))
+        default: break
+        }
+        return nil
+    }
+    
+    // There is no i18n in the rust sdk, so we have to do it here from the raw string.
+    private static func translatableFromRawString(_ message: String) -> String {
+        return switch message {
```
pixlwave, line 148: Could the SDK maybe emit an enum here instead of a string? That'll keep this switch much safer as currently we could easily upgrade the SDK and not know about any changes needed here.

ElementX/Sources/Services/Timeline/TimelineItemContent/RoomTimelineItemProperties.swift
```diff
@@ -26,4 +26,6 @@ struct RoomTimelineItemProperties: Hashable {
     var deliveryStatus: TimelineItemDeliveryStatus?
     /// The read receipts of the item, ordered from newest to oldest
     var orderedReadReceipts: [ReadReceipt] = []
+    /// Message shield for authenticity warnings
+    var shield: MessageShield?
```
pixlwave, line 30: Could we maybe call this something like `authenticityWarning: AuthenticityWarning` or message/encryption warning/details or similar? At this level we're no where near rendering anything that looks like a shield, so it would be nicer to keep it UI agnostic.

Could even just be `authenticity` that's an enum and not an optional, but that might be one step too far.

ElementX/Sources/Services/Timeline/TimelineItemProxy.swift
```diff
@@ -120,10 +120,46 @@ class EventTimelineItemProxy {
...
+            return MessageShield.fromRustShield(shieldState: shield)
+        }
+        return nil
+    }()
+    
     lazy var readReceipts = item.readReceipts()
 }
 
+extension MessageShield {
+    static func fromRustShield(shieldState: ShieldState) -> MessageShield? {
```
pixlwave, line 135: This is really an init so may as well use once :)
```suggestion
    init(rustShieldState: ShieldState) {
```

ElementX/Sources/Services/Timeline/TimelineItemContent/MessageShield.swift
```diff
@@ -0,0 +1,27 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+
+struct MessageShield: Hashable {
+    var color: ShieldColor
```
pixlwave, line 20: And again, this could probably be something more like `var severity: Severity` where `Severity` had say `info` and `warning` states. We can inline the enum too to not need a prefix.

pixlwave (commented):
A few thoughts about the data below. Will take a look at the views separately.

ElementX/Sources/Services/Timeline/TimelineItemContent/MessageShield.swift
BillCarsonFr: > And again, this could probably be something more like `var severity: Severity` where `Severity` had say `info` and `warning` states. We can inline the enum too to not need a prefix.

Actually we try to avoid that. It's preferable that all clients display these shields the same way with same colors. That's why it's not just a severity level that you can render as you prefer.

ElementX/Sources/Services/Timeline/TimelineItemProxy.swift
BillCarsonFr: Ideally we want the SDK to give the right string and support i18n. We don't want clients to use different strings :/ 
But for now no i18n in SDK

ElementX/Sources/Services/Timeline/TimelineItemProxy.swift
BillCarsonFr: It may not create a MessageShield

pixlwave: My bad, we can use `init?` for optional initialisers in Swift, have updated the suggestion.

ElementX/Sources/Services/Timeline/TimelineItemContent/MessageShield.swift
pixlwave: Got it 👍

ElementX/Sources/Services/Timeline/TimelineItemProxy.swift
pixlwave: Hmmm, ok I see the intention but seeing as it doesn't right now, an enum would still be the right choice here?

Also worth questioning if the SDK should support i18n as different platforms might want to handle languages differently (e.g. on iOS a user can add multiple languages). Whereas it could do the same as SAS and give clients an identifier to use against a known set of strings (`SessionVerificationEmoji.description`), leaving the platform to choose how to pick the right one for the user.

Velin92: might be worth checking how we implemented localisation on iOS for SAS using SASL10n

ElementX/Sources/Screens/RoomScreen/View/Style/TimelineItemSendInfoLabel.swift
```diff
@@ -172,14 +199,14 @@ struct TimelineItemSendInfoLabel_Previews: PreviewProvider, TestablePreview {
...
-                                                      status: .authenticityUnknown,
+                                                      status: .encryptionAuthenticity(.unsignedDevice(color: .red)),
                                                       layoutType: .horizontal()))
             TimelineItemSendInfoLabel(sendInfo: .init(localizedString: "09:47 AM",
-                                                      status: .unencrypted,
+                                                      status: .encryptionAuthenticity(.notGuaranteed(color: .gray)),
                                                       layoutType: .horizontal()))
+//            TimelineItemSendInfoLabel(sendInfo: .init(localizedString: "09:47 AM",
+//                                                      status: .unencrypted,
+//                                                      layoutType: .horizontal()))
```
pixlwave, line 209: This is brought back in a follow-up branch.

stefanceriu (approved):
Very nice, lgtm! 👍

ElementX/Sources/Services/Timeline/TimelineItemContent/MessageShield.swift
bmarty: Just cross posting, without knowing this first discussion, I had the same remark here: https://github.com/element-hq/element-x-android/pull/3240#discussion_r1716474739...

PR #3052, opened 2024-07-17
Run the integration tests on GitHub again.

pixlwave:
We're going to use Sentry instead of the performance tests, so let's move these back here to get them running again.

Test run: https://github.com/element-hq/element-x-ios/actions/runs/9970988482 ✅


stefanceriu (approved):
🚀 👏

PR #3053, opened 2024-07-17
Simplify how we setup Sentry to make sure it's configured before any …

stefanceriu:
…spans get created - they don't get reported otherwise

### DIFF
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -95,10 +95,6 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
         
         navigationRootCoordinator = NavigationRootCoordinator()
         
-        Self.setupServiceLocator(appSettings: appSettings, appHooks: appHooks)
-        
-        ServiceLocator.shared.analytics.startIfEnabled()
-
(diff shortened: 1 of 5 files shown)


ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -115,6 +111,12 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
         notificationManager = NotificationManager(notificationCenter: UNUserNotificationCenter.current(),
                                                   appSettings: appSettings)
         
+        Self.setupSentry(appSettings: appSettings)
+        
+        Self.setupServiceLocator(appSettings: appSettings, appHooks: appHooks)
+        
+        ServiceLocator.shared.analytics.startIfEnabled()
+
```
pixlwave, line 119: It would be nice to get rid of the service locator's singleton at some point. E.g. looking at the UserSessionStore it references it and whilst that shouldn't be called before this is run it would still be nice to have more compile time safety.

pixlwave (approved):
Nice, LGTM 👍

ElementX/Sources/Application/AppCoordinator.swift
stefanceriu: Yeah, I agree, I didn't particularly enjoy moving these methods around.

PR #3057, opened 2024-07-17
Add a feature flag for Simplified Sliding Sync.

pixlwave:
Requires #3056 and https://github.com/matrix-org/matrix-rust-sdk/pull/3723 (plus an SDK release).

### DIFF
--- a/ElementX/Sources/Application/AppSettings.swift
+++ b/ElementX/Sources/Application/AppSettings.swift
@@ -63,6 +63,7 @@ final class AppSettings {
     static func resetSessionSpecificSettings() {
         MXLog.warning("Resetting the user session specific AppSettings.")
         store.removeObject(forKey: UserDefaultsKeys.hasRunIdentityConfirmationOnboarding.rawValue)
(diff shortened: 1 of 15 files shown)


ElementX/Sources/Application/AppSettings.swift
```diff
@@ -62,6 +63,7 @@ final class AppSettings {
     static func resetSessionSpecificSettings() {
         MXLog.warning("Resetting the user session specific AppSettings.")
         store.removeObject(forKey: UserDefaultsKeys.hasRunIdentityConfirmationOnboarding.rawValue)
+        store.removeObject(forKey: UserDefaultsKeys.simplifiedSlidingSyncEnabled.rawValue)
```
stefanceriu, line 66: Forgot we do this 🙈

stefanceriu (approved):
lgtm 👍

ElementX/Sources/Application/AppSettings.swift
pixlwave: Yeah, I only thought about it after realising `.requires_sliding_sync` doesn't work with `.simplified_sliding_sync`.

PR #3062, opened 2024-07-18
Fixes #2840 - Use both the room list room and the room preview details to populate the join room screen

### LINKED ISSUE #2840: Invite had no accept button
### Steps to reproduce

1. Tapped on a push notif, invite from the CVE bot
2. Saw this:

IMG_6601

#### What did you expect?

Accept/Reject style buttons

#### What happened instead?

No buttons. I went back to the roomlist; invite wasn't there (perhaps SS hadn't synced?); waited 

### Your phone model

iPhone 12 Pro Max

### Operating system version

_No response_

### Application version

597

### Homeserver

_No response_

### Will you send logs?

Yes

stefanceriu:
- the room summary API is indeed enabled on matrix.org and working fine for most rooms
- it is not however capable of giving us data about non-joined + private rooms
- the SDK addresses that by first trying to use known rooms before resorting to the preview endpoint
- that fails if it's a brand new room that the client doesn't know about yet i.e. a sync hasn't ran, which is exactly what's happening here
- the ClientProxy instead does wait for the room list to go into the first loaded before returning the room

### DIFF
--- a/ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenModels.swift
+++ b/ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenModels.swift
@@ -29,11 +29,19 @@ enum JoinRoomScreenInteractionMode {
     case knock
 }
 
+struct JoinRoomScreenRoomDetails {
+    let name: String?
+    let topic: String?
+    let canonicalAlias: String?
+    let avatar: RoomAvatar
+    let memberCount: UInt
+}
+
 struct JoinRoomScreenViewState: BindableState {
     // Maybe use room summary details or similar here??
     let roomID: String
     
-    var roomDetails: RoomPreviewDetails?
+    var roomDetails: JoinRoomScreenRoomDetails?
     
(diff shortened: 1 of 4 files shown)


ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenModels.swift
```diff
@@ -29,11 +29,19 @@ enum JoinRoomScreenInteractionMode {
     case knock
 }
 
+struct JoinRoomScreenRoomDetails {
+    let name: String?
+    let topic: String?
+    let canonicalAlias: String?
+    let avatarURL: URL?
```
pixlwave, line 36: Can we use `RoomAvatar` here and construct one from the preview if we receive one. Then we should 🤞 still see the user's avatar when it's an invite to a DM.

pixlwave (approved):
Nice! 😎

pixlwave:
Also, sorry for the conflicts.

PR #3063, opened 2024-07-18
Pinning items Feature Flag

Velin92:
this also includes swiftlint error on line_legth removal alongside increasing the warning value from 1000 to 2000
Simulator Screenshot - iPhone 15 - 2024-07-18 at 15 27 25


stefanceriu (approved):
lgtm but are you sure you want to put `Pinning items Feature Flag` as a feature in the changelog?

pixlwave:
> in the changelog?

Also (drive-by comment here) would be nice to use `pr-wip` as the category for things most people won't see.

PR #3064, opened 2024-07-18
Update RTE to 2.37.7

Velin92:
No difference in terms of features, but this new version builds the XCFramework correctly


stefanceriu (approved):

pixlwave (commented):
We can tidy this up too: https://github.com/element-hq/element-x-ios/blob/7fbdc2c20fa31784f1cb48d0142a6d40eaf321ca/ElementX/SupportingFiles/target.yml#L177-L189

PR #3072, opened 2024-07-19
Pinned Items Banner UI

Velin92:
Implemented the UI for the banner, without any logic for now.

Simulator Screenshot - iPhone 15 - 2024-07-19 at 19 07 38

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/View/PinnedItemsBanner/PinnedItemsBannerView.swift
+++ b/ElementX/Sources/Screens/RoomScreen/View/PinnedItemsBanner/PinnedItemsBannerView.swift
@@ -1,0 +1,101 @@
+//
+// Copyright 2024 New Vector Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import SwiftUI
+
+import Compound
+
+struct PinnedItemsBannerView: View {
+    let pinIndex: Int
+    let pinsCount: Int
+    let pinContent: AttributedString
+    
+    let onMainButtonTap: () -> Void
+    let onViewAllButtonTap: () -> Void
+    
+    private var bannerIndicatorDescription: AttributedString {
+        let index = pinIndex + 1
+        let boldPlaceholder = "{bold}"
+        var finalString = AttributedString(L10n.Screen.Room.pinnedBannerIndicatorDescription(boldPlaceholder))
+        var boldString = AttributedString(L10n.Screen.Room.pinnedBannerIndicator(index, pinsCount))
+        boldString.bold()
+        finalString.replace(boldPlaceholder, with: boldString)
+        return finalString
+    }
+    
+    var body: some View {
+        HStack(spacing: 0) {
+            mainButton
+            viewAllButton
+        }
+        .padding(.vertical, 16)
+        .padding(.leading, 16)
+        .background(Color.compound.bgCanvasDefault)
+        .shadow(color: Color(red: 0.11, green: 0.11, blue: 0.13).opacity(0.1), radius: 12, x: 0, y: 4)
+    }
+    
+    private var mainButton: some View {
+        Button(action: onMainButtonTap,
+               label: {
+                   HStack(spacing: 0) {
+                       HStack(spacing: 10) {
+                           PinnedItemsIndicatorView(pinIndex: pinIndex, pinsCount: pinsCount)
+                               .accessibilityHidden(true)
+                           CompoundIcon(\.pinSolid, size: .small, relativeTo: .compound.bodyMD)
+                               .foregroundColor(Color.compound.iconSecondaryAlpha)
+                               .accessibilityHidden(true)
+                           content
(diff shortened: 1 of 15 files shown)


ElementX/Sources/Screens/RoomScreen/View/PinnedItemsBanner/PinnedItemsBannerView.swift
```diff
@@ -0,0 +1,101 @@
...
+    
+    private var bannerIndicatorDescription: AttributedString {
+        let index = pinIndex + 1
+        let boldPlaceholder = "{bold}"
+        var finalString = AttributedString(L10n.Screen.Room.pinnedBannerIndicatorDescription(boldPlaceholder))
+        var boldString = AttributedString(L10n.Screen.Room.pinnedBannerIndicator(index, pinsCount))
+        boldString.bold()
+        finalString.replace(boldPlaceholder, with: boldString)
+        return finalString
+    }
```
pixlwave, line 37: Do you think we should make this a stored property in the view state and build it in the view model? I imagine this will rarely need to be done more than once for a room, but as it is, this could likely be run many times on the same screen.

ElementX/Sources/Screens/RoomScreen/View/PinnedItemsBanner/PinnedItemsBannerView.swift
```diff
@@ -0,0 +1,101 @@
...
+                       HStack(spacing: 10) {
+                           PinnedItemsIndicatorView(pinIndex: pinIndex, pinsCount: pinsCount)
+                               .accessibilityHidden(true)
+                           CompoundIcon(\.pinSolid, size: .small, relativeTo: .compound.bodyMD)
+                               .foregroundColor(Color.compound.iconSecondaryAlpha)
+                               .accessibilityHidden(true)
+                           content
+                       }
+                       Spacer()
+                   }
```
pixlwave, line 63: Can we replace the outer stack + spacer with a `.frame(maxWidth: .infinity, alignment: .leading)` on the inner stack?

ElementX/Sources/Screens/RoomScreen/View/PinnedItemsBanner/PinnedItemsIndicatorView.swift
```diff
@@ -0,0 +1,67 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import SwiftUI
+
+import Compound
```
pixlwave, line 19:
```suggestion
import Compound
import SwiftUI
```

ElementX/Sources/Screens/RoomScreen/View/PinnedItemsBanner/PinnedItemsBannerView.swift
```diff
@@ -0,0 +1,101 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import SwiftUI
+
+import Compound
```
pixlwave, line 19:
```suggestion
import Compound
import SwiftUI
```

ElementX/Sources/Screens/RoomScreen/View/PinnedItemsBanner/PinnedItemsBannerView.swift
```diff
@@ -0,0 +1,101 @@
...
+                       }
+                       Spacer()
+                   }
+               })
+               .accessibilityElement(children: .contain)
+    }
+    
+    private var viewAllButton: some View {
+        Button(action: onViewAllButtonTap,
+               label: {
```
pixlwave, line 70: Can we stick to trailing closures for the label like we do elsewhere for both of the buttons on this view?

ElementX/Sources/Screens/RoomScreen/View/PinnedItemsBanner/PinnedItemsIndicatorView.swift
Velin92: shouldn't the order be 
- First party frameworks in alphabetical order
- newline
- Third party in alphabetical order

ElementX/Sources/Screens/RoomScreen/View/PinnedItemsBanner/PinnedItemsBannerView.swift
Velin92: yeah I would like to move this in the viewModel, but not right now, I am still deciding how the viewmodel -> banner interaction would be (injecting the context, or just passing the value), will probably do that later when we have a clear idea of how these live items will get updated

ElementX/Sources/Screens/RoomScreen/View/PinnedItemsBanner/PinnedItemsIndicatorView.swift
pixlwave: Ummmm, we're not doing that anywhere else: https://github.com/element-hq/element-x-ios/blob/ff33c07596e7620512b22253317b96e7767e0bfe/ElementX/Sources/Application/AppCoordinator.swift#L17-L24 🤷‍♂️

ElementX/Sources/Screens/RoomScreen/View/PinnedItemsBanner/PinnedItemsBannerView.swift
pixlwave: Ah got it, make sense 👍

PR #3075, opened 2024-07-22
Fix editing items not in the timeline failing

Velin92:
Requires an SDK release.

### DIFF
--- a/ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift
+++ b/ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift
@@ -191,15 +191,38 @@ class RoomTimelineController: RoomTimelineControllerProtocol {
               html: String?,
               intentionalMentions: IntentionalMentions) async {
         MXLog.info("Edit message in \(roomID)")
+        MXLog.info("Editing timeline item: \(timelineItemID)")
         
-        switch await activeTimeline.edit(timelineItemID,
-                                         message: message,
-                                         html: html,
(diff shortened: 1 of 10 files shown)


ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift
```diff
@@ -191,15 +191,38 @@ class RoomTimelineController: RoomTimelineControllerProtocol {
...
+        let editMode: EditMode
+        if !timelineItemID.timelineID.isEmpty,
+           let timelineItem = liveTimelineProvider.itemProxies.firstEventTimelineItemUsingStableID(timelineItemID) {
+            editMode = .byEvent(timelineItem)
+        } else if let eventID = timelineItemID.eventID {
+            editMode = .byID(eventID)
+        } else {
+            MXLog.error("Unknown timeline item: \(timelineItemID)")
+            return
+        }
```
pixlwave, line 205: Seeing as we make this enum and then switch on it in the same place, should we maybe just build the content first and put the contents of the switch directly in these if branches?

pixlwave (approved):
👌

ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift
Velin92: I mean if there is an error, we just return before we build the content, which is more optimal, ofc the error is probably a particularly rare edge case, but at least it is accounted for

PR #3077, opened 2024-07-22
ElementCall unable to access media on ongoing CallKit session.

stefanceriu:


pixlwave (approved):
Thank you for the details comment 👍

PR #3080, opened 2024-07-23
Hide/Show pin banner based on scroll direction

Velin92:
https://github.com/user-attachments/assets/09e6bb70-9b18-4438-b813-b16dc03d6813


ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
```diff
@@ -165,7 +169,24 @@ struct RoomScreenViewState: BindableState {
     var canCurrentUserRedactSelf = false
     var canCurrentUserPin = false
     var isViewSourceEnabled: Bool
+    
     var isPinningEnabled = false
+    var lastScrolledDirection: ScrollDirection?
```
pixlwave, line 174:
```suggestion
    var lastScrollDirection: ScrollDirection?
```

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
```diff
@@ -49,15 +49,11 @@ struct RoomScreen: View {
...
-                    PinnedItemsBannerView(pinIndex: 1,
-                                          pinsCount: 3,
-                                          pinContent: .init(stringLiteral: "Content"),
-                                          onMainButtonTap: { },
-                                          onViewAllButtonTap: { })
+                if context.viewState.shouldShowPinBanner {
+                    pinnedItemsBanner
                 }
             }
+            .animation(.elementDefault, value: context.viewState.shouldShowPinBanner)
```
pixlwave, line 56: Worth double checking this (might want to be a ZStack or something instead), but this would make sure to isolate the animations to only the banner. (It's a bit sad that while we now have `if` expressions you can't put the modifier directly on the `if` but oh well)
```suggestion
                Group {
                    if context.viewState.shouldShowPinBanner {
                        pinnedItemsBanner
                    }
                }
                .animation(.elementDefault, value: context.viewState.shouldShowPinBanner)
            }
```

ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
```diff
@@ -165,7 +169,24 @@ struct RoomScreenViewState: BindableState {
...
+    // These are just mocked items used for testing, their types might change
+    let pinnedItems = [
+        "Hello 1",
+        "How are you 2",
+        "I am fine 3",
+        "Thank you 4"
+    ]
+    var currentPinIndex = 0
```
pixlwave, line 182: Not for now, but might be worth thinking about a dedicated struct to hold these (and the `selectedPinContent`) together in a single property. Could even put some of the methods like `selectNextItem()` directly on it for unit testing.

pixlwave (approved):
Nice!

PR #3081, opened 2024-07-24
Fix the restoration of a Simplified Sliding Sync session.

pixlwave:
This fixes the following bugs:

- Force closing the app signs you out.
- Clearing the cache in Developer Options signs you out.
- Missing verification/restoration prompts.

### DIFF
--- a/ElementX/Sources/Services/UserSession/RestorationToken.swift
+++ b/ElementX/Sources/Services/UserSession/RestorationToken.swift
@@ -47,7 +47,8 @@ extension MatrixRustSDK.Session: Codable {
                          deviceId: container.decode(String.self, forKey: .deviceId),
                          homeserverUrl: container.decode(String.self, forKey: .homeserverUrl),
                          oidcData: container.decodeIfPresent(String.self, forKey: .oidcData),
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Services/UserSession/RestorationToken.swift
```diff
@@ -47,7 +47,8 @@ extension MatrixRustSDK.Session: Codable {
                          deviceId: container.decode(String.self, forKey: .deviceId),
                          homeserverUrl: container.decode(String.self, forKey: .homeserverUrl),
                          oidcData: container.decodeIfPresent(String.self, forKey: .oidcData),
-                         slidingSyncProxy: container.decode(String.self, forKey: .slidingSyncProxy))
+                         // Note: the proxy is optional now that we support Simplified Sliding Sync.
+                         slidingSyncProxy: container.decodeIfPresent(String.self, forKey: .slidingSyncProxy))
```
pixlwave, line 51: In theory we could add a check to the app settings here, but the implementation ends up being a bit hacky and in practice we should trust the SDK to give us the correct data.

Velin92 (approved):

PR #3084, opened 2024-07-24
Pin/Unpin Logic

Velin92:
Requires an updated SDK containing the following PR: https://github.com/matrix-org/matrix-rust-sdk/pull/3750

fixes https://github.com/element-hq/element-x-ios/issues/3058

https://github.com/user-attachments/assets/b2e86389-5f16-40fc-a1c3-4b1476c2dd09

### DIFF
--- a/ElementX/Resources/Localizations/en.lproj/Localizable.strings
+++ b/ElementX/Resources/Localizations/en.lproj/Localizable.strings
@@ -70,6 +70,7 @@
 "action_ok" = "OK";
 "action_open_settings" = "Settings";
 "action_open_with" = "Open with";
+"action_pin" = "Pin";
 "action_quick_reply" = "Quick reply";
 "action_quote" = "Quote";
 "action_react" = "React";
@@ -99,11 +100,10 @@
 "action_take_photo" = "Take photo";
 "action_tap_for_options" = "Tap for options";
 "action_try_again" = "Try again";
+"action_unpin" = "Unpin";
 "action_view_source" = "View source";
 "action_yes" = "Yes";
 "action.load_more" = "Load more";
-"action.pin" = "Pin";
-"action.unpin" = "Unpin";
 "common_about" = "About";
 "common_acceptable_use_policy" = "Acceptable use policy";
 "common_advanced_settings" = "Advanced settings";
@@ -314,9 +314,9 @@
 "screen_advanced_settings_element_call_base_url_description" = "Set a custom base URL for Element Call.";
 "screen_advanced_settings_element_call_base_url_validation_error" = "Invalid URL, please make sure you include the protocol (http/https) and the correct address.";
 "screen_room_mentions_at_room_subtitle" = "Notify the whole room";
-"screen.room.pinned_banner_indicator" = "%1$@ of %2$@";
-"screen.room.pinned_banner_indicator_description" = "%1$@ Pinned messages";
-"screen.room.pinned_banner_view_all_button_title" = "View All";
+"screen_room_pinned_banner_indicator" = "%1$@ of %2$@";
+"screen_room_pinned_banner_indicator_description" = "%1$@ Pinned messages";
+"screen_room_pinned_banner_view_all_button_title" = "View All";
 "screen_account_provider_change" = "Change account provider";
 "screen_account_provider_form_hint" = "Homeserver address";
 "screen_account_provider_form_notice" = "Enter a search term or a domain address.";
--- a/ElementX/Resources/Localizations/en.lproj/Localizable.strings
+++ b/ElementX/Resources/Localizations/en.lproj/Localizable.strings
@@ -103,6 +103,7 @@
 "action_yes" = "Yes";
 "action.load_more" = "Load more";
 "action.pin" = "Pin";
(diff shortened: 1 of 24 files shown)


ElementX/Sources/Screens/RoomScreen/View/ItemMenu/TimelineItemMenuActionProvider.swift
```diff
@@ -21,6 +21,7 @@ struct TimelineItemMenuActionProvider {
     let canCurrentUserRedactSelf: Bool
     let canCurrentUserRedactOthers: Bool
     let canCurrentUserPin: Bool
+    let pinnedEvents: Set<String>
```
pixlwave, line 24: Might be slightly clearer if this was either `pinnedEventIDs` or `: Set<TimelineItemIdentifier>`, not sure it matters which :)

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
```diff
@@ -109,10 +110,10 @@ struct RoomScreen: View {
     }
     
     private var pinnedItemsBanner: some View {
-        PinnedItemsBannerView(pinIndex: context.viewState.currentPinIndex,
-                              pinsCount: context.viewState.pinnedItems.count,
-                              pinContent: context.viewState.selectedPinContent,
-                              onMainButtonTap: { context.send(viewAction: .nextPin) },
+        PinnedItemsBannerView(pinIndex: context.viewState.pinnedEventsState.selectedPinIndex,
+                              pinsCount: context.viewState.pinnedEventsState.pinnedEvents.count,
+                              pinContent: context.viewState.pinnedEventsState.selectedPinContent,
```
pixlwave, line 115: Would it make sense to pass this the pinnedEventsState as a single parameter now?

ElementX/Sources/Services/Room/RoomProxyProtocol.swift
```diff
@@ -38,6 +38,7 @@ protocol RoomProxyProtocol {
     var isSpace: Bool { get }
     var isEncrypted: Bool { get }
     var isFavourite: Bool { get async }
+    var pinnedEvents: [String] { get async }
```
pixlwave, line 41: Same here I guess.

ElementX/Sources/Services/Timeline/TimelineProxy.swift
```diff
@@ -202,6 +202,24 @@ final class TimelineProxy: TimelineProxyProtocol {
         }
     }
     
+    func pin(eventID: String) async -> Result<Bool, TimelineProxyError> {
+        do {
+            return try await .success(timeline.pinEvent(eventIdStr: eventID))
```
pixlwave, line 207: Lol at the `Str` suffix, we should send the Rust folks a few more pennies so they can fill in the missing characters 🤭

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -404,12 +407,17 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
         roomProxy
             .actionsPublisher
             .filter { $0 == .roomInfoUpdate }
-            .throttle(for: .seconds(1), scheduler: DispatchQueue.main, latest: true)
+            .receive(on: DispatchQueue.main)
```
pixlwave, line 410: What happened to the throttle, I would imagine this was needed for a specific reason no?

ElementX/Resources/Localizations/en.lproj/Localizable.strings
```diff
@@ -103,6 +103,7 @@
 "action_yes" = "Yes";
 "action.load_more" = "Load more";
 "action.pin" = "Pin";
+"action.unpin" = "Unpin";
```
pixlwave, line 106: Think these are missing the nested option on Localazy btw:
Screenshot 2024-07-24 at 6 13 36 PM

ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
```diff
@@ -304,3 +294,40 @@ enum ScrollDirection: Equatable {
...
+                selectedPin = pinnedEvents.first
+            } else if pinnedEvents.isEmpty {
+                selectedPin = nil
+            } else if let selectedPin, !pinnedEvents.contains(selectedPin) {
+                self.selectedPin = pinnedEvents.first
+            }
+        }
+    }
+    
+    var selectedPin: String?
```
pixlwave, line 312: And the same here, `selectedPinEventID` would be clearer :)

pixlwave (approved):
🤩

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
Velin92: Yeah so regarding the throttle... it makes the pinning banner updates slower... maybe what I could is remove it and do an if check on every value, and only change them if they are not equal to the new value, WDYT?

PR #3099, opened 2024-07-29
Pinned items timeline implementation for the banner

Velin92:
fixes #3072 

Requires https://github.com/matrix-org/matrix-rust-sdk/pull/3773 to be merged

https://github.com/user-attachments/assets/7f296271-c4ed-4501-95d2-7d4401dda13d


ElementX/Sources/Services/Room/RoomSummary/RoomEventStringBuilder.swift
```diff
@@ -21,6 +21,7 @@ struct RoomEventStringBuilder {
     let stateEventStringBuilder: RoomStateEventStringBuilder
     let messageEventStringBuilder: RoomMessageEventStringBuilder
     let shouldDisambiguateDisplayNames: Bool
+    let shouldPrefixSenderName: Bool
```
stefanceriu, line 24: Conversely, can we apply the same Enum based solution here and reduce the api surface area? e.g 
```
enum Prefix {
    case senderName(displayName, shouldDisambiguate)
}
```

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -660,17 +658,19 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
     
     // MARK: - Timeline Item Building
     
-    private func buildPinnedEventsContent(timelineItems: [RoomTimelineItemProtocol]) {
-        var timelineItemsDictionary = OrderedDictionary<String, RoomTimelineItemType>()
+    private func buildPinnedEventsContent(timelineItems: [TimelineItemProxy]) {
```
stefanceriu, line 661: Guess this answers the question above about them being items not events. I think we should rename this whole thing on our side to be `pinnedTimelineItems`

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -49,12 +50,12 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
     private var paginateBackwardsTask: Task<Void, Never>?
     private var paginateForwardsTask: Task<Void, Never>?
     
+    // Needs to be stored to allow the provider to keep sending updates.
     private var pinnedEventsTimelineController: RoomTimelineControllerProtocol?
```
stefanceriu, line 54: Do you need a periphery ignore then?

ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
```diff
@@ -178,7 +178,7 @@ struct RoomScreenViewState: BindableState {
     var pinnedEventsState = PinnedEventsState()
     
     var shouldShowPinBanner: Bool {
```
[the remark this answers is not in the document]
Velin92: in the end they will just be attributed strings in the room screen/banner

ElementX/Sources/Services/Room/RoomSummary/RoomMessageEventStringBuilder.swift
```diff
@@ -18,10 +18,17 @@ import Foundation
 import MatrixRustSDK
 
 struct RoomMessageEventStringBuilder {
+    enum Prefix {
+        case senderName
```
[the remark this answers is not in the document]
Velin92: The problem lies within emotes, which ALWAYS require the senderDisplayName

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
Velin92: I think we should instead rename these to 
```
private func buildPinnedEventContent(timelineItems: [TimelineItemProxy]) {
        var pinnedEventContents = OrderedDictionary<String, AttributedString>()
```

Since the timeline items are only used to build the content, but we don't need to store or use them in any other way in the room screen (they will be used as timeline items later when implementing the pinned events timeline view, but not here)

ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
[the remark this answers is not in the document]
stefanceriu: Okay, fair enough 👍

ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
```diff
@@ -172,10 +172,13 @@ struct RoomScreenViewState: BindableState {
     var isPinningEnabled = false
     var lastScrollDirection: ScrollDirection?
     
+    // The `pinnedEventIDs` is used only to determine if an item can be pinned or not since it does not depend on the pinned events timeline, but on the room info update which is faster
+    var pinnedEventIDs: Set<String> = []
+    // This is used to controler the banner
     var pinnedEventsState = PinnedEventsState()
```
stefanceriu, line 178:
```suggestion
    // This is used to control the banner
    var pinnedEventsState = PinnedEventsState()
```

Although I would just delete that comment ¯\_(ツ)_/¯

ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
```diff
@@ -172,10 +172,13 @@ struct RoomScreenViewState: BindableState {
     var isPinningEnabled = false
     var lastScrollDirection: ScrollDirection?
     
+    // The `pinnedEventIDs` is used only to determine if an item can be pinned or not since it does not depend on the pinned events timeline, but on the room info update which is faster
```
stefanceriu, line 175: As in if I have permissions to pin an item or if that item has been already pinned or something else? I think we need to make it more explicit than this.

stefanceriu (approved):
Very nice, 🚀 it! 👏

PR #3100, opened 2024-07-29
Refactor TimelineItemSendInfo out of the styler.

pixlwave:
Small refactoring in preparation to add encryption shields to messages in a similar way to send failures.

Last commit extends this with some (unexposed) encryption shield states.
test_timelineItemSendInfoLabel-iPhone-15-pseudo 1
(Technically they're on the wrong side, but will discuss with @amshakal tomorrow as it would be nice to have consistency for the status icon in this view)


stefanceriu (approved):
Very very nice!

My only comment is that I would've named that new file `TimelineItemSendInfoLabel` instead of `TimelineItemSendInfoModifier` as the label is the important thing in there.

PR #3114, opened 2024-08-05
Fix a bug where moving a new line in the composer could move the caret.

### LINKED ISSUE #3104: Cursor jumps in composer
### Steps to reproduce

Start a longer text in the composer, then jump back a paragraph
3. What do you click?
Then click "enter" to create a new line

#### What did you expect?
A new line where I pressed enter

#### What happened instead?
The cursor jumps to the end of the already typed text

https://github.com/user-attachments/assets/ddecc580-fc71-405b-8547-7977ac796af0

### Your phone model

iPhone 12 mini

### Operating system version

17.5.1

### Application version

1.7.2 (118)

### Homeserver

Synapse

### Will you send logs?

No

pixlwave:
Fixes #3104.

The diffs revealed some missing attributes on the new line so I'm only restoring the selection when the plain string is unchanged.

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/ComposerToolbar/View/MessageComposerTextField.swift
+++ b/ElementX/Sources/Screens/RoomScreen/ComposerToolbar/View/MessageComposerTextField.swift
@@ -105,6 +105,9 @@ private struct UITextViewWrapper: UIViewRepresentable {
                                      .foregroundColor: UIColor(.compound.textPrimary)]
         
         if textView.attributedText != text {
+            // Remember the selection if only the attributes have changed.
+            let selection = textView.attributedText.string == text.string ? textView.selectedTextRange : nil
+            
             textView.attributedText = text
             
             // Re-apply the default font when setting text for e.g. edits.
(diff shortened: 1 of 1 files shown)


ElementX/Sources/Screens/RoomScreen/ComposerToolbar/View/MessageComposerTextField.swift
```diff
@@ -105,6 +105,9 @@ private struct UITextViewWrapper: UIViewRepresentable {
                                      .foregroundColor: UIColor(.compound.textPrimary)]
         
         if textView.attributedText != text {
+            // Remember the selection if only the attributes have changed.
```
stefanceriu, line 108: 👌

ElementX/Sources/Screens/RoomScreen/ComposerToolbar/View/MessageComposerTextField.swift
```diff
@@ -120,6 +123,10 @@ private struct UITextViewWrapper: UIViewRepresentable {
                     textView.keyboardType = .default
                     textView.reloadInputViews()
                 }
+            } else if let selection {
+                // Fixes a bug where pressing Return in the middle of two paragraphs
```
stefanceriu, line 127: I tend to find it easier to link the original ticket in comments like these instead of relying on git history

pixlwave: Oooops, for once I actually meant to but forgot to paste it in.

Velin92 (approved):

PR #3116, opened 2024-08-05
Add Encryption Authenticity explanations.

pixlwave:
- Tapping on the status icon in the SendInfo label shows an alert with a description of the status.
- Shows a message's authenticity warning in its action menu.
- Tidy up VoiceOver slightly in the room list and timeline item menu.

Closes https://github.com/element-hq/element-x-ios/issues/3109

https://github.com/user-attachments/assets/3bb51c0a-d93c-4523-9e9d-5a4dbf905422


ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -607,6 +609,24 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
...
     }
+    
+    private func handleItemSendInfoTapped(itemID: TimelineItemIdentifier) {
+        guard let timelineItem = timelineController.timelineItems.firstUsingStableID(itemID) else {
+            MXLog.warning("Couldn't find timeline item.")
+            return
+        }
+        
+        guard let eventTimelineItem = timelineItem as? EventBasedTimelineItemProtocol else {
+            MXLog.warning("Not an event.")
```
stefanceriu, line 620: Should non-events even have this action, should this be a fatal error instead?

ElementX/Sources/Screens/RoomScreen/View/ItemMenu/TimelineItemMenu.swift
```diff
@@ -169,23 +179,54 @@ struct TimelineItemMenu: View {
     }
 }
 
+private extension EncryptionAuthenticity {
```
stefanceriu, line 182: I'm not a big fan of having Colors encoded into the EncryptionAuthenticity. Wouldn't it be better if we would use the Compound approach and replace `enum Color { case red, gray }` with an `enum Importance { case primary, secondary }`  or similar?

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
pixlwave: Actually yeah true, this can definitely be a fatal.

ElementX/Sources/Screens/RoomScreen/View/ItemMenu/TimelineItemMenu.swift
stefanceriu: Okay, I now understand that this has been done this way to keep in line with [the Rust implementation](https://github.com/element-hq/element-x-ios/pull/3051#discussion_r1683017277). 

For posterity though, we belive colors to be inherently localisable and `severity` to be a way better abstraction. 

Otherwise we can find ourselves in weird situations in which colors don't mean what we expect them to e.g. `In China, red is auspicious—associated with life-generating energy (the sun, blood, and fire)—and is the color of celebrations and prosperity`

stefanceriu (approved):

ElementX/Sources/Screens/RoomScreen/View/ItemMenu/TimelineItemMenu.swift
BillCarsonFr: For context this is on purpose, and we want consistency across clients on how crypto shields/warning are represented(idealy same icons, colors, text).

stefanceriu: I understand, just pointing out that doesn't seem to be a winning approach seeing as how both icons and colors need to also be localized for the various cultures matrix/element will be used in.

Another random example:
```UX designers need to take into account cultural differences when selecting symbols and colors. For example, while the color green is associated with life and growth in many Western cultures, it symbolizes death in some Asian countries```

PR #3118, opened 2024-08-05
Pinned events banner loading state

Velin92:
- also updated the SDK
- reversed the timeline for the banner
- implemented a system to retry getting the timeline when networking is available
- Improved the string building for polls and stickers

https://github.com/user-attachments/assets/a886103c-5398-4921-9939-25b7c5a06279

### DIFF
--- a/ElementX/Resources/Localizations/en.lproj/Localizable.strings
+++ b/ElementX/Resources/Localizations/en.lproj/Localizable.strings
@@ -318,6 +318,7 @@
 "screen_room_mentions_at_room_subtitle" = "Notify the whole room";
 "screen_room_pinned_banner_indicator" = "%1$@ of %2$@";
 "screen_room_pinned_banner_indicator_description" = "%1$@ Pinned messages";
+"screen_room_pinned_banner_loading_description" = "Loading message...";
 "screen_room_pinned_banner_view_all_button_title" = "View All";
 "screen_account_provider_change" = "Change account provider";
 "screen_account_provider_form_hint" = "Homeserver address";
@@ -851,6 +852,12 @@
 "state_event_room_name_removed_by_you" = "You removed the room name";
 "state_event_room_none" = "%1$@ made no changes";
 "state_event_room_none_by_you" = "You made no changes";
+"state_event_room_pinned_events_changed" = "%1$@ changed the pinned messages";
+"state_event_room_pinned_events_changed_by_you" = "You changed the pinned messages";
+"state_event_room_pinned_events_pinned" = "%1$@ pinned a message";
+"state_event_room_pinned_events_pinned_by_you" = "You pinned a message";
+"state_event_room_pinned_events_unpinned" = "%1$@ unpinned a message";
+"state_event_room_pinned_events_unpinned_by_you" = "You unpinned a message";
 "state_event_room_reject" = "%1$@ rejected the invitation";
 "state_event_room_reject_by_you" = "You rejected the invitation";
 "state_event_room_remove" = "%1$@ removed %2$@";
--- a/ElementX/Resources/Localizations/en.lproj/Localizable.strings
+++ b/ElementX/Resources/Localizations/en.lproj/Localizable.strings
@@ -318,7 +318,7 @@
 "screen_room_mentions_at_room_subtitle" = "Notify the whole room";
 "screen_room_pinned_banner_indicator" = "%1$@ of %2$@";
 "screen_room_pinned_banner_indicator_description" = "%1$@ Pinned messages";
-"screen_room_pinned_banner_loading_description" = "Loading message...";
+"screen_room_pinned_banner_loading_description" = "Loading message…";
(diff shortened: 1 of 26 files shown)


ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -509,6 +505,48 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
...
+    private func setupPinnedEventsTimelineProviderIfNeeded() {
+        Task {
+            guard let pinnedEventsTimelineProvider = await roomProxy.pinnedEventsTimeline?.timelineProvider else {
+                return
+            }
+            
+            if self.pinnedEventsTimelineProvider == nil {
+                self.pinnedEventsTimelineProvider = pinnedEventsTimelineProvider
+            }
+        }
```
Velin92, line 540: So I have some doubts about this... since this function is called both in the init, and when the reachability sends a value of `reachable`, but what if this happens to schedule these Task both at the same time?

Since internally `await roomProxy.pinnedEventsTimeline` creates a timeline, could it try to create 2 of them at the same time?

ElementX/Resources/Localizations/en.lproj/Localizable.strings
```diff
@@ -318,6 +318,7 @@
 "screen_room_mentions_at_room_subtitle" = "Notify the whole room";
 "screen_room_pinned_banner_indicator" = "%1$@ of %2$@";
 "screen_room_pinned_banner_indicator_description" = "%1$@ Pinned messages";
+"screen_room_pinned_banner_loading_description" = "Loading message...";
```
pixlwave, line 321: This should probably use a `…` character :)

ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
```diff
@@ -323,30 +323,101 @@ struct PinnedEventsState: Equatable {
...
+            return numberOfEvents
+        }
+    }
+    
+    var selectedPinIndex: Int {
+        switch self {
+        case .loaded(let state):
+            return state.selectedPinIndex
+        case .loading(let numbersOfEvents):
+            return numbersOfEvents - 1
```
pixlwave, line 390: It's not immediately obvious to me why this is the last event, maybe there should be a comment? Also is it an issue that this could become negative?

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
pixlwave, line 1: Definitely not for this PR, but I'm wondering if there's enough logic in here that we should make a baby `PinnedItemsBannerViewModel` similar to how we have a `ComposerToolbarViewModel` for the Room Screen

ElementX/Sources/Screens/RoomScreen/View/PinnedItemsBanner/PinnedItemsBannerView.swift
```diff
@@ -18,11 +18,20 @@ import Compound
...
     let onViewAllButtonTap: () -> Void
     
+    private var displayedMessage: AttributedString {
+        switch state {
+        case .loading:
+            return AttributedString(L10n.screenRoomPinnedBannerLoadingDescription)
+        case .loaded(let state):
+            return state.selectedPinContent
+        }
+    }
```
pixlwave, line 33: Could this be a property on the state?

ElementX/Sources/Screens/RoomScreen/View/Timeline/TimelineView.swift
```diff
@@ -86,7 +86,7 @@ struct TimelineView_Previews: PreviewProvider, TestablePreview {
                                                mediaPlayerProvider: MediaPlayerProviderMock(),
                                                voiceMessageMediaManager: VoiceMessageMediaManagerMock(),
                                                userIndicatorController: ServiceLocator.shared.userIndicatorController,
-                                               appMediator: AppMediatorMock.default,
+                                               networkMonitor: ServiceLocator.shared.networkMonitor, appMediator: AppMediatorMock.default,
```
pixlwave, line 89:
```suggestion
                                               networkMonitor: ServiceLocator.shared.networkMonitor,
                                               appMediator: AppMediatorMock.default,
```

ElementX/Sources/Services/Room/RoomSummary/RoomMessageEventStringBuilder.swift
```diff
@@ -20,7 +20,7 @@ import MatrixRustSDK
 struct RoomMessageEventStringBuilder {
     enum Prefix {
         case senderName
-        case mediaType
+        case type
```
pixlwave, line 23: Not important but perhaps `messageType` would be a bit more descriptive here?

pixlwave (approved):
Nice 👏

PR #3120, opened 2024-08-06
Fixes #3050 - Sync mute state between ElementCall and CallKit

### LINKED ISSUE #3050: [ElementCall] Sync mute state between EC and CallKit
Once a call is started the user has the option to mute the call directly from the CallKit screen. 

Currently [this action is intentionally failing](https://github.com/element-hq/element-x-ios/blob/develop/ElementX/Sources/Services/ElementCall/ElementCallService.swift#L208-L209) as ElementCall doesn't expose any methods for  setting that and CallKit muting isn't hardware level so sound will still go through. 

The opposite should also happen: muting EC should forward the mute state to CallKit.

stefanceriu:
* listen to changes in the mute state and sync up the CallKit controls with ElementCall and back
* rewrite how we construct and pass custom messages between the ElementCall widget and the widget driver

https://github.com/user-attachments/assets/6c84d9c7-c5f2-4846-9db7-40422c31220b

### DIFF
--- a/ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
+++ b/ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
@@ -62,9 +62,6 @@ class CallScreenViewModel: CallScreenViewModelType, CallScreenViewModelProtocol 
                 return
             }
             
-            // TODO: intercept EC mute state changes and pass them over to CallKit
-            // elementCallService.setCallMuted(roomID: roomProxy.id, muted: muted)
-            
             Task {
                 await self.widgetDriver.sendMessage(message)
             }
@@ -76,14 +73,14 @@ class CallScreenViewModel: CallScreenViewModelType, CallScreenViewModelProtocol 
                 guard let self else { return }
                 
                 switch action {
-                case let .setCallMuted(muted, roomID):
+                case let .setAudioEnabled(enabled, roomID):
                     guard roomID == roomProxy.id else {
                         MXLog.error("Received mute request for a different room: \(roomID) != \(roomProxy.id)")
                         return
                     }
                     
                     Task {
-                        await self.setMuted(muted)
+                        await self.setAudioEnabled(enabled)
                     }
                 default:
                     break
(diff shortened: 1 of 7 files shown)


ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
```diff
@@ -20,6 +20,11 @@ import SwiftUI
 
 typealias CallScreenViewModelType = StateStoreViewModel<CallScreenViewState, CallScreenViewAction>
 
+private enum ElementCallAction: String {
+    case hangup = "im.vector.hangup"
+    case mediaState = "io.element.device_mute"
```
Velin92, line 25: since these are shared with the JSON model, maybe we could have the ones from the JSON model exposed as internal, so in case they might change we won't need to change them in 2 different places

ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
```diff
@@ -18,22 +18,40 @@ import Combine
...
+struct ElementCallWidgetMessage: Codable {
     enum Direction: String, Codable {
         case fromWidget
         case toWidget
     }
     
     enum Action: String, Codable {
         case hangup = "im.vector.hangup"
+        case mediaState = "io.element.device_mute"
+    }
```
Velin92, line 30: I think these can be extracted from the struct and be internal to the whole project so they are used only once

Velin92 (approved):
just a small suggestion, the rest LGTM

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
stefanceriu: Oh my bad, these are not actually used anymore, just forgot to delete them 👍

ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
stefanceriu: Yup yup, I just forgot to delete the others. `ElementCallWidgetMessage` is the only correct way to create a mesage now and it's internal to the project.

PR #3121, opened 2024-08-06
Finalise strings/icons for EncryptionAuthenticity.

pixlwave:
The final designs are now available, so this uses the correct icons and updates a localised string.


stefanceriu (approved):
lgtm 👍

PR #3124, opened 2024-08-06
Put a space between the version and build numbers on the Nightly icon

pixlwave:
Highly important PR - I've started using Nightly on macOS and the icon is bothering me 😅


stefanceriu (approved):
😂

PR #3141, opened 2024-08-08
Fixes #3126 - Prevent identity confirmation from blocking the user's progress after registering through OIDC

### LINKED ISSUE #3126: Device verification appears after a fresh new registration
### Steps to reproduce

1. Select a HS with a MAS accepting registration (ask if you do not know one)
2. Create an account

<details>
<summary>The device verification unexpectly appears</summary>

</details>

#### What did you expect?
Enteting within the app and seeing an empty room list.

#### What happened instead?
I had to reset the identity on a fresh new account

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

_No response_

### Homeserver

_No response_

### Will you send logs?

Yes

stefanceriu:
- happened because `hasRunIdentityConfirmationOnboarding` is false while the verification state is still `.unknown`
- verification eventually changes to `.verified` but the flow coordinator doesn't catch up on it and blocks the flow

### DIFF
--- a/ElementX/Sources/FlowCoordinators/OnboardingFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/OnboardingFlowCoordinator.swift
@@ -231,7 +231,7 @@ class OnboardingFlowCoordinator: FlowCoordinatorProtocol {
                 presentRecoveryKeyScreen()
             case .skip:
                 appSettings.hasRunIdentityConfirmationOnboarding = true
-                stateMachine.tryEvent(.next)
+                stateMachine.tryEvent(.nextSkippingIdentityConfimed)
             case .reset:
                 presentEncryptionResetScreen()
             }
@@ -246,10 +246,14 @@ class OnboardingFlowCoordinator: FlowCoordinatorProtocol {
                 .map(\.verificationState)
                 .removeDuplicates()
                 .sink { [weak self] value in
+                    guard let self else { return }
+                    
                     if value == .verified {
-                        self?.stateMachine.tryEvent(.nextSkippingIdentityConfimed)
+                        appSettings.hasRunIdentityConfirmationOnboarding = true
+                        stateMachine.tryEvent(.nextSkippingIdentityConfimed)
                     } else {
-                        verificationStateCancellable = nil
+                        // Captured by the block below, nil-ing it wouldn't work
+                        verificationStateCancellable?.cancel()
                     }
                 }
         }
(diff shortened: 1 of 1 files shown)


ElementX/Sources/FlowCoordinators/OnboardingFlowCoordinator.swift
```diff
@@ -43,6 +43,7 @@ class OnboardingFlowCoordinator: FlowCoordinatorProtocol {
     
     enum Event: EventType {
         case next
+        case nextSkippingIdentityConfimed
```
pixlwave, line 46: Can we use this on our debug skip button too? 😇

ElementX/Sources/FlowCoordinators/OnboardingFlowCoordinator.swift
```diff
@@ -223,7 +238,25 @@ class OnboardingFlowCoordinator: FlowCoordinatorProtocol {
...
+        // If the verification state is still unknown wait for it to resolve
+        // and just move on to the next steps if verified
+        var verificationStateCancellable: AnyCancellable?
+        if userSession.sessionSecurityStatePublisher.value.verificationState == .unknown {
+            verificationStateCancellable = userSession.sessionSecurityStatePublisher
+                .map(\.verificationState)
+                .removeDuplicates()
+                .sink { [weak self] value in
+                    if value == .verified {
+                        self?.stateMachine.tryEvent(.nextSkippingIdentityConfimed)
```
pixlwave, line 250: Do we need to set the app setting to true here too?

ElementX/Sources/FlowCoordinators/OnboardingFlowCoordinator.swift
```diff
@@ -223,7 +238,25 @@ class OnboardingFlowCoordinator: FlowCoordinatorProtocol {
...
+                    } else {
+                        verificationStateCancellable = nil
+                    }
+                }
+        }
+        
+        presentCoordinator(coordinator) { [verificationStateCancellable] in
+            verificationStateCancellable?.cancel()
```
pixlwave, line 258: Wait, if this is captured, then setting it to nil above won’t do anything right? As we push the next screens onto the stack.

ElementX/Sources/FlowCoordinators/OnboardingFlowCoordinator.swift
stefanceriu: Good point!

ElementX/Sources/FlowCoordinators/OnboardingFlowCoordinator.swift
stefanceriu: Same here!

PR #3154, opened 2024-08-12
Remove SS proxy migration and waitlist screens

stefanceriu:
`We must release this in EX apps only after the synapse release and global SSS availability on matrix.org.`


pixlwave (approved):
LGTM, although I thought we were waiting until we enable SSS to remove the migration screen?

PR #3168, opened 2024-08-14
Automatically try reloading failed images on network changes

### LINKED ISSUE #3110: Profiles often don’t load in timeline.
### Steps to reproduce

1. open a room for the first time in a session
2. about 30% of the time, no profiles load - just raw mxids:

IMG_7912

3. go back to roomlist and reenter room
4. profiles now appear

#### What did you expect?

reliable profiles everywhere

#### What happened instead?

racey flakiness

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

667

### Homeserver

_No response_

### Will you send logs?

No

stefanceriu:
Fixes #3110

This PR passes over the network monitor to the LoadableImage so that it can be used together with the already passed ImageLoader to automatically reload images when network conditions change.


UITests/Sources/__Snapshots__/Application/roomRolesAndPermissionsFlow-1-iPad-10th-generation-en-GB.UI.png
pixlwave, line 1: We should really fix this and make the mock media provider return the same image on both paths 🫠

ElementX/Sources/Other/SwiftUI/Views/LoadableImage.swift
pixlwave, line 1: My 1 constructive comment: Would it be worth only setting up the subscription if the loading fails?

pixlwave (approved):

Velin92 (changes requested):
Before merging this I wanted to give another possible solution which I feel may be easier and more scalable for this refactor, here is my take in our internal room:

https://matrix.to/#/!JTahvxBRtQfZViMwKT:matrix.org/$ZPQcvASHkqaIFhXNRnK0ETios37A9FVsUpFMg5gvCD8?via=matrix.org&via=element.io

I'll be on holidays until next week, so feel free to disregard this suggestion if you feel is not worth it.

stefanceriu:
> I wanted to give another possible solution which I feel may be easier and more scalable for this refactor

Thanks for that! We ended up building something similar but based on plain tasks instead of publishers in #3170

PR #3169, opened 2024-08-14
RoomScreenViewModel refactor part 2

### LINKED ISSUE #2991: Power level changes don't get reflected immediately when in a room.
### Steps to reproduce

1. Be in a room where you don't have any room admin privs
2. Try to start a video call
3. Discover you don't have permission to do so.
4. Get someone to make you a mod or similar
5. Discover you still can't start a video call
6. Realise you have to go to the roomlist and back again before the app releases that you are now a mod.

#### What did you expect?

The UI should update immediately on power level changes

#### What happened instead?

You have to reload the room to see the changes.

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

628 or so

### Homeserver

_No response_

### Will you send logs?

No

Velin92:
Also added more tests and also fixes #2991

### DIFF
--- a/ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
+++ b/ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
@@ -81,13 +81,10 @@ class TimelineViewModel: TimelineViewModelType, TimelineViewModelProtocol {
                                                                 analyticsService: analyticsService)
         
         super.init(initialViewState: TimelineViewState(roomID: roomProxy.id,
-                                                       roomTitle: roomProxy.roomTitle,
-                                                       roomAvatar: roomProxy.avatar,
                                                        isEncryptedOneToOneRoom: roomProxy.isEncryptedOneToOneRoom,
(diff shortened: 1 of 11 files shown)


ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
```diff
@@ -449,7 +423,7 @@ class TimelineViewModel: TimelineViewModelType, TimelineViewModelProtocol {
                 case .displayMediaUploadPreviewScreen(let url):
                     actionsSubject.send(.displayMediaUploadPreviewScreen(url: url))
                 case .displayRoomMemberDetails(userID: let userID):
```
pixlwave, line 425: What sends this? Should it line up naming-wise or something?

pixlwave (approved):
✨🧹✨

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
Velin92: oh good catch, this seems to be dead code since the interaction handler doesn't send this from anywhere

PR #3170, opened 2024-08-15
Automatically try reloading failed images on network changes

stefanceriu:
This PR exposes a new mechanism in the ImageProvider that can hold off image loading resolution until the connectivity is re-established. It automatically handles cancelling tasks when views get deallocated and will bail out if the second attempt fails while on a reachable connection to avoid retrying indefinitely.

It also refactors the MediaProvider tests and moves towards a generated mock MediaLoader.


ElementX/Sources/Services/Media/Provider/MediaProvider.swift
```diff
@@ -14,17 +14,21 @@
...
 //
 
+import Combine
 import Kingfisher
 import UIKit
 
 struct MediaProvider: MediaProviderProtocol {
     private let mediaLoader: MediaLoaderProtocol
     private let imageCache: Kingfisher.ImageCache
+    private let networkMonitor: NetworkMonitorProtocol!
```
pixlwave, line 24: Not exactly sure how I feel about this being implicitly unwrapped, given we pass `nil` in the NSE. It should probably be optional or maybe we should use `#if IS_MAIN_APP` to only expose the retrying method to the main app where it was set?

ElementX/Sources/Services/Media/Provider/MediaProvider.swift
```diff
@@ -71,6 +75,39 @@ struct MediaProvider: MediaProviderProtocol {
...
+            guard !Task.isCancelled else {
+                throw MediaProviderError.cancelled
+            }
+            
+            for await reachability in networkMonitor.reachabilityPublisher.values {
+                guard !Task.isCancelled else {
+                    throw MediaProviderError.cancelled
+                }
+                
+                if reachability == .reachable {
```
pixlwave, line 93: Tiny and unimportant but could make this a `guard else continue` to reduce the indentation 1 level.

pixlwave (approved):

PR #3172, opened 2024-08-15
Translations update

RiotRobot:
- Translations update


pixlwave (approved):
Not sure what the coverage drop is about these are only strings…

PR #3174, opened 2024-08-16
Navigation support for upcoming Element Call Picture in Picture mode.

pixlwave:
This PR adds the initial support for EC PiP with the following changes:
- Replace the sheet with a SwiftUI fullscreen overlay (so we can hide the web view without removing it from the view hierarchy).
- Listen for AVPictureInPictureController notifications and use these to show/hide the call overlay.
- Make sure that joining the same call that's running in PiP, dismisses the PiP instead of creates a new call.
- Dismiss the composer before starting a call (as there's no longer a sheet to take the focus away).

### DIFF
--- a/ElementX/Sources/Application/AppSettings.swift
+++ b/ElementX/Sources/Application/AppSettings.swift
@@ -252,7 +252,8 @@ final class AppSettings {
 
     // MARK: - Element Call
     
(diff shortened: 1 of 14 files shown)


stefanceriu (approved):
Nothing to comment, looks great! 👏

PR #3181, opened 2024-08-19
Remove GenericCallLinkCoordinator, merging it into CallScreen.

pixlwave:
This PR removes the `GenericCallLinkCoordinator` so we can share logic for the 2 types of call. This will allow for the use of PiP when available and so generic calls are now presented with the same overlay.


ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
pixlwave, line 1: Note to self: The navigation bar and bottom safe area now show up in light mode even though EC links are always in dark mode.

ElementX/Sources/Application/Navigation/NavigationRootCoordinator.swift
```diff
@@ -127,5 +171,14 @@ private struct NavigationRootCoordinatorView: View {
...
         }
+        .overlay {
+            Group {
+                if let coordinator = rootCoordinator.overlayModule?.coordinator {
+                    coordinator.toPresentable()
+                        .transition(.opacity)
+                }
+            }
+            .animation(.elementDefault, value: rootCoordinator.overlayModule)
+        }
```
Velin92, line 182: what about the navigation triggered by notifications? I imagine this won't dismiss the overlay but just navigate behind it right?

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
```diff
@@ -194,6 +152,54 @@ class CallScreenViewModel: CallScreenViewModelType, CallScreenViewModelProtocol
     
     // MARK: - Private
     
+    private func start() {
```
Velin92, line 155: start is a bit misleading because we usually use this signature name as an exposed function in the protocol, to be used when the coordinator is starting,  I suggest changing the name to something else like `setup()` ?

pixlwave: Oh yeah good point 👍

ElementX/Sources/Application/Navigation/NavigationRootCoordinator.swift
pixlwave: Hmmm, so it doesn't matter in this instance and it is only used when signed out (and we're not supporting PiP there even if there was somewhere to navigate to). But you're right, I need to think about this in the `UserSessionFlowCoordinator` 🤔

pixlwave: Will follow this up in another PR.

PR #3187, opened 2024-08-20
Introduce a new `RoomProxyType` and treat rooms differently based on their membership state

stefanceriu:


ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -415,19 +413,16 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
             }
             userIndicatorController.submitIndicator(UserIndicator(id: Self.leaveRoomLoadingID, type: .modal, title: L10n.commonLeavingRoom, persistent: true))
             
-            let room = await userSession.clientProxy.roomForIdentifier(roomId)
-            let result = await room?.leaveRoom()
```
pixlwave, line 419: Err, don't you need to call this on the joined room?

Also unrelated but *cough* `roomID` *cough*

ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenViewModel.swift
```diff
@@ -27,7 +27,7 @@ class JoinRoomScreenViewModel: JoinRoomScreenViewModelType, JoinRoomScreenViewMo
     private let userIndicatorController: UserIndicatorControllerProtocol
     
     private var roomPreviewDetails: RoomPreviewDetails?
-    private var roomProxy: RoomProxyProtocol?
+    private var room: RoomProxyType?
```
pixlwave, line 30: Maybe I'm missing something here, but why would this be anything other than an `InvitedRoomProxyProtocol`? Then we wouldn't need to fetch it twice in this screen.

ElementX/Sources/Services/Room/JoinedRoomProxy.swift
```diff
@@ -77,13 +77,17 @@ class RoomProxy: RoomProxyProtocol {
...
     }
         
-    private let actionsSubject = PassthroughSubject<RoomProxyAction, Never>()
-    var actionsPublisher: AnyPublisher<RoomProxyAction, Never> {
+    private let actionsSubject = PassthroughSubject<JoinedRoomProxyAction, Never>()
+    var actionsPublisher: AnyPublisher<JoinedRoomProxyAction, Never> {
         actionsSubject.eraseToAnyPublisher()
     }
     
     lazy var id: String = room.id()
```
pixlwave, line 85: Is it a dumb question to ask why this one is still lazy?

ElementX/Sources/Services/Room/InvitedRoomProxy.swift
```diff
@@ -0,0 +1,109 @@
...
+
+import Foundation
+import MatrixRustSDK
+import UIKit
+
+class InvitedRoomProxy: InvitedRoomProxyProtocol {
+    private let roomListItem: RoomListItemProtocol
+    private let room: RoomProtocol
+    
+    lazy var id: String = room.id()
```
pixlwave, line 25: I guess same question here.

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -868,19 +844,46 @@ class ClientProxy: ClientProxyProtocol {
...
+        }
+        
+        guard let roomListItem = try? roomListService.room(roomId: identifier) else {
+            MXLog.error("Failed retrieving room, invalid room list item")
+            return nil
+        }
+        
+        if roomListItem.isTimelineInitialized() == false {
+            do {
+                try await roomListItem.initTimeline(eventTypeFilter: eventFilters, internalIdPrefix: nil)
```
pixlwave, line 861: Wait so does this now work for a room invite too?

pixlwave (approved):
Very nice tidy-up, this is much safer 👏👏👏

Did it need any Rust changes in the end?

ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu: Nop, that's still completely broken 😁

ElementX/Sources/Services/Room/JoinedRoomProxy.swift
stefanceriu: No idea to be honest ¯\_(ツ)_/¯

ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenViewModel.swift
stefanceriu: Talked about it offline, the sort answer is: to pave the way for knocked rooms.

PR #3189, opened 2024-08-21
Fixes problems processing invites, build rooms differently based on their membership

stefanceriu:
Requires https://github.com/matrix-org/matrix-rust-sdk/pull/3869


pixlwave (approved):
Niiice 🤩

PR #3190, opened 2024-08-21
Use native video call picture in picture!

pixlwave:
This PR uses an `AVPictureInPictureVideoCallController` to PiP the whole EC web view instead of the previous solution where EC would request the web view to PiP the speaker's video. I've made the feature flag user configurable now as it no longer requires a specific EC deployment*.

This PR doesn't handle responses from the web view about whether PiP is possible so it is possible to get stuck in a call when an error occurs.

The second and third commits improves the handling of whether or not the we should start the PiP or dismiss the screen.

The fourth commit hides the call button when you're already joined to the call in that room.

\* It subsequently does now, however when using an unsupported deployment it will simply navigate back which isn't the case with the old implementation on `develop`. The user can enter https://pr2573--element-call.netlify.app/ as the base URL above the flag for testing purposes.

### DIFF
--- a/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
@@ -580,7 +580,7 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
     
     private var callScreenPictureInPictureController: AVPictureInPictureController?
     private func presentCallScreen(configuration: ElementCallConfiguration) {
-        guard elementCallService.ongoingCallRoomID != configuration.callID else {
+        guard elementCallService.ongoingCallRoomID != configuration.callRoomID else {
             MXLog.info("Returning to existing call.")
             callScreenPictureInPictureController?.stopPictureInPicture()
             return
--- a/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
@@ -477,6 +477,7 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
                                                     roomTimelineControllerFactory: roomTimelineControllerFactory,
                                                     navigationStackCoordinator: detailNavigationStackCoordinator,
                                                     emojiProvider: EmojiProvider(),
+                                                    ongoingCallRoomIDPublisher: elementCallService.ongoingCallRoomIDPublisher,
                                                     appMediator: appMediator,
                                                     appSettings: appSettings,
                                                     analytics: analytics,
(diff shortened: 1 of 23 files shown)


ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -580,7 +580,7 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
     
     private var callScreenPictureInPictureController: AVPictureInPictureController?
     private func presentCallScreen(configuration: ElementCallConfiguration) {
-        guard elementCallService.ongoingCallRoomID != configuration.callID else {
+        guard elementCallService.ongoingCallRoomID != configuration.callRoomID else {
```
pixlwave, line 583: unrelated refactor on a previous PR, this just seems a bit more correct given they're strings.

ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
```diff
@@ -46,52 +47,60 @@ struct CallScreen: View {
...
                 .id(context.viewState.url)
                 .ignoresSafeArea(edges: .bottom)
         }
     }
 }
 
-private struct WebView: UIViewRepresentable {
+private struct CallView: UIViewRepresentable {
+    /// The top-level view this representable displays. It wraps the web view when picture in picture isn't running.
+    class WebViewWrapper: UIView { }
```
pixlwave, line 60: Honestly not sure on this. I think it's nicer to have a subclass to help explain what the `UIViewRepsentable` is generic over (and that it's not the actual web view). But equally it's redundant seeing as this view doesn't need to know about the web view.

Velin92: Maybe you could just use a typealias

ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
```diff
@@ -120,6 +129,21 @@ private struct WebView: UIViewRepresentable {
             webView.isOpaque = false
             webView.backgroundColor = .compound.bgCanvasDefault
             webView.scrollView.backgroundColor = .compound.bgCanvasDefault
+            
+            webViewWrapper.addMatchedSubview(webView)
```
Velin92, line 133: actually why do we need the wrapper? what would prevent the WebView to be instantiated before `super.init()` and just have its delegates set to self after `super.init()`

Velin92 (approved):
LGTM, just a few suggestions/nits

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
```diff
@@ -175,8 +175,10 @@ struct RoomScreen: View {
         
         if !ProcessInfo.processInfo.isiOSAppOnMac {
             ToolbarItem(placement: .primaryAction) {
-                callButton
-                    .disabled(roomContext.viewState.canJoinCall == false)
+                if roomContext.viewState.shouldShowCallButton {
+                    callButton
+                        .disabled(roomContext.viewState.canJoinCall == false)
```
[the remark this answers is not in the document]
pixlwave: Oh yeah good spot, I wonder why this isn't caught by SwiftLint/Format 🤔

ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
pixlwave: It's because a view can only be added to a single parent, so when the PiP starts and we add it to the PiPVC, it has to be removed from the SwiftUI view hierarchy. So it can't be the base view for the UIViewRepresentable as that doesn't have a way to let you take the view away from it.

Will try the type alias.

PR #3193, opened 2024-08-22
`TimelineKind` refactor

Velin92:
`isLive` has been replaced with a `TimelineKind enum` to allow a better handling of the pinned timeline


pixlwave (approved):
Awesome, this feels much clearer about what's going on 🙌

PR #3196, opened 2024-08-23
Update the SDK

### LINKED ISSUE #3202: warning shields appear on local echos of edits
### Steps to reproduce

1. send a msg in an e2ee room
2. edit it
3. observe that the local echo’d edit sprouts a false red warning shield as if the msg was unencrypted
4. shield goes away once remote echo received

https://github.com/user-attachments/assets/1318643f-7e65-4658-a6bf-1eabb01adef6

#### What did you expect?

no unexpected transient shields.

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

688

### Homeserver

_No response_

### Will you send logs?

No

pixlwave:
This PR updates the SDK to the latest version which with it brings the following code changes:

- New send failure types that will be picked up as part of https://github.com/element-hq/element-internal/issues/614.
- A new message shield/`EncryptionAuthenticity` state for users who were previously verified but have now reset their identity.
- The dedicated event cache store (which will result in all media being deleted from the device).
    - We're storing this in `appGroupContainer/Library/Caches/…` using the new `cachePath` on the client builder and we're implementing this at the right time to not require a migration.

And the following SDK changes for free:
- Various fixes under the hood for the pinned items timeline.
- A bug fix for edits where their local echoes would show up as [Not encrypted](https://github.com/matrix-org/matrix-rust-sdk/issues/3882)
- Updates to how active calls are computed.

This is definitely one to review commit-by-commit.

Fixes #3202

### DIFF
--- a/ElementX/Sources/Services/Authentication/AuthenticationService.swift
+++ b/ElementX/Sources/Services/Authentication/AuthenticationService.swift
@@ -20,7 +20,7 @@ import MatrixRustSDK
 
 class AuthenticationService: AuthenticationServiceProtocol {
     private var client: Client?
-    private var sessionDirectory: URL
+    private var sessionDirectories: SessionDirectories
     private let passphrase: String
     
     private let userSessionStore: UserSessionStoreProtocol
@@ -31,7 +31,7 @@ class AuthenticationService: AuthenticationServiceProtocol {
     var homeserver: CurrentValuePublisher<LoginHomeserver, Never> { homeserverSubject.asCurrentValuePublisher() }
     
     init(userSessionStore: UserSessionStoreProtocol, encryptionKeyProvider: EncryptionKeyProviderProtocol, appSettings: AppSettings, appHooks: AppHooks) {
-        sessionDirectory = .sessionsBaseDirectory.appending(component: UUID().uuidString)
+        sessionDirectories = .init()
         passphrase = encryptionKeyProvider.generateKey().base64EncodedString()
         self.userSessionStore = userSessionStore
         self.appSettings = appSettings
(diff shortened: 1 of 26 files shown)


ElementX/Sources/Services/Authentication/AuthenticationService.swift
```diff
@@ -149,20 +149,24 @@ class AuthenticationService: AuthenticationServiceProtocol {
...
     }
     
     private func rotateSessionDirectory() {
-        if FileManager.default.directoryExists(at: sessionDirectory) {
-            try? FileManager.default.removeItem(at: sessionDirectory)
+        if FileManager.default.directoryExists(at: sessionDirectories.dataDirectory) {
+            try? FileManager.default.removeItem(at: sessionDirectories.dataDirectory)
+        }
+        if FileManager.default.directoryExists(at: sessionDirectories.cacheDirectory) {
+            try? FileManager.default.removeItem(at: sessionDirectories.cacheDirectory)
```
stefanceriu, line 162: SessionDirectories should maybe handle this on its own and be reused for the QRLoginService below.

ElementX/Sources/Services/UserSession/RestorationToken.swift
```diff
@@ -29,10 +30,22 @@ extension RestorationToken: Codable {
         let container = try decoder.container(keyedBy: CodingKeys.self)
         
         let session = try container.decode(MatrixRustSDK.Session.self, forKey: .session)
-        let sessionDirectory = try container.decodeIfPresent(URL.self, forKey: .sessionDirectory)
+        let dataDirectory = try container.decodeIfPresent(URL.self, forKey: .sessionDirectory)
+        let cacheDirectory = try container.decodeIfPresent(URL.self, forKey: .cacheDirectory)
+        
+        let sessionDirectories = if let dataDirectory {
```
stefanceriu, line 36: Continuing from the comment above I wonder if we can or should make the SessionDirectories create itself from a container 🤔

ElementX/Sources/Services/UserSession/UserSessionStore.swift
```diff
@@ -158,22 +158,36 @@ class UserSessionStore: UserSessionStoreProtocol {
                           appSettings: appSettings)
     }
     
-    private func deleteSessionDirectory(for credentials: KeychainCredentials) {
+    private func deleteSessionDirectories(for credentials: KeychainCredentials) {
```
stefanceriu, line 161: Same question here I guess: can we move and hide a bunch of this logic into the new `SessionDirectories`?

UnitTests/Sources/RestorationTokenTests.swift
```diff
@@ -0,0 +1,108 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import XCTest
+
+@testable import ElementX
+import MatrixRustSDK
+
+class RestorationTokenTests: XCTestCase {
```
stefanceriu, line 22: Very nice, gotta love tests! 🙌

stefanceriu (approved):
Looks great! 👏  
.. just wondering if we can leverage the new types better

pixlwave:
Discussed in a call that I'll do a follow-up PR that refactors this a bit more after we make the RC. We're happy with the serialisation of the `RestorationToken` so it won't be a breaking refactor.

PR #3204, opened 2024-08-27
Picture in Picture support for Calls (requires a supported widget deployment).

pixlwave:
Remove the feature flag on EC PiP support - this will now be driven by the web view `canEnablePip` API.


pixlwave:
Failing snapshots are unrelated to these changes.

PR #3210, opened 2024-08-29
Expand the CallScreen into the safe area and let ElementCall handle layout

stefanceriu:


stefanceriu:
This is not on the table just yet and I got tired of seeing it around. Can be restored later if need be.

PR #3211, opened 2024-08-29
Native Sliding Sync Discovery

pixlwave:
This PR adopts the latest changes in the SDK for Sliding Sync discovery. Changes are as follows:

- Restoration: This is fully handled through the Session, and we don't to use `.simplified` based on the FF any more.
- Configuration: There is a new picker in the developer options to choose the SSS type
    - `Proxy only` is the current type of discovery
    - `Automatic` will attempt native discovery and fallback to proxy discovery if native SS isn't available.
    - `Force Native` will use native SS without any discovery. This is necessary for matrix.org as support isn't advertised without a token that's on the allowlist.
- Authentication: Make a new AuthenticationClientBuilder that handles the above configurations appropriately.


ElementX/Sources/Services/Authentication/AuthenticationService.swift
```diff
@@ -64,7 +66,7 @@ class AuthenticationService: AuthenticationServiceProtocol {
         } catch ClientBuildError.WellKnownDeserializationError(let error) {
             MXLog.error("The user entered a server with an invalid well-known file: \(error)")
             return .failure(.invalidWellKnown(error))
-        } catch ClientBuildError.SlidingSyncNotAvailable {
+        } catch ClientBuildError.SlidingSyncVersion(let error) {
```
pixlwave, line 69: This error is a flattened string, I need to dig deeper to find out if we need the unflattened version.

stefanceriu (approved):

PR #3212, opened 2024-08-29
Add a logout button on the identity confirmation screen

stefanceriu:


ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -185,6 +185,17 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
...
             }
             .store(in: &cancellables)
+        
+        onboardingFlowCoordinator.actions
+            .sink { [weak self] action in
+                guard let self else { return }
+                
+                switch action {
+                case .logout:
+                    Task { await self.runLogoutFlow() }
```
pixlwave, line 195: Is this going to ask you about losing history/keys etc? Seems a bit weird if the user hasn't made it into the app yet to say that they'll lose stuff.

pixlwave (approved):
LGTM!

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
stefanceriu: Good point, yeah.

PR #3219, opened 2024-09-02
Accept current autocorrection/suggestion before sending a message

stefanceriu:


manuroe (approved):
LGTM.
I trust you about the temporary `UITextField` field ;)

stefanceriu:
Not particularly happy about it either but the only other solution I found was to resign+become firstResponder which was creating UI glitches.

PR #3232, opened 2024-09-04
Replace Prefire with a very similar but simpler and more direct approach

stefanceriu:
I think we still need to:

- [ ] Get rid of Prefire in Compound?
- [ ] Maybe replace the Snapshotting.swift extensions
- [ ] Clean up the template?


.gitattributes
pixlwave, line 1: Asking a question right at the start of the review so I haven't seen any code, but how come these moved?

ElementX/Sources/Other/Extensions/Snapshotting.swift
```diff
@@ -0,0 +1,58 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import SwiftUI
+
+public struct DelayPreferenceKey: PreferenceKey {
```
pixlwave, line 19: We should probably prepend these 3 keys with `Snapshot`?

ElementX/Sources/Other/TestablePreview.swift
```diff
@@ -16,6 +16,4 @@
 
 import SwiftUI
 
-import Prefire
-
-protocol TestablePreview: PrefireProvider { }
+protocol TestablePreview { }
```
pixlwave, line 19: Now that we don't have a separate module we can make this conform to `PreviewProvider` and only need a single conformance on the provider.

ElementX/SupportingFiles/Settings.bundle/Acknowledgements.plist
pixlwave, line 1: Can you reset package caches and re-commit this? Sounds like it'll be fixed in [Xcode 16](https://github.com/FelixHerrmann/swift-package-list/issues/117#issuecomment-2282802723).

PreviewTests/Sources/PreviewTests.swift
pixlwave, line 1: It would be really nice now if we could separate out the generated code from `assertSnapshots(matching: preview)` and everything it calls.

PreviewTests/SupportingFiles/target.yml
```diff
@@ -29,10 +29,6 @@ targets:
     type: bundle.unit-test
     platform: iOS
     
-    buildToolPlugins:
-    - plugin: PrefireTestsPlugin
-      package: Prefire
-
```
pixlwave, line 35: We should also get rid of the CI trusting plugins in the Fastfile preview tests run.

pixlwave (approved):
Very nice! 👏

ElementX/Sources/Other/TestablePreview.swift
stefanceriu: Agreed offline to keep it for now

PreviewTests/Sources/PreviewTests.swift
stefanceriu: Will do that as the next step, separating everything in a differnet repo to be used by compound as well

PR #3236, opened 2024-09-06
Fixes #3197 - Allow voice message playback in the background

### LINKED ISSUE #3197: Disable locking during voice message playback / play in background
Listening to voice messages longer than the auto locking cycle is very annoying.
One has to either tap the screen during playback or resume playback when the voice message got paused by the phone auto locking.

When playing back a voice message, the phone could be prevented from locking, similar to how it does not lock while recording one.

Alternatively or even in addition (if one wants to lock the phone manually during playback), the voice message playback could also continue in the background, while the phone is locked.

So far I've not found one, except just manually preventing the phone from locking. I did not want to increase the phone's locking time as it a) does not solve the problem for longer voice messages either and b) reduces security

**Additional context**
(None)

stefanceriu:
- inform other audio sessions on deactivation
- let CallKit handle the audioSession for ElementCall

### DIFF
--- a/ElementX/Sources/Services/ElementCall/ElementCallService.swift
+++ b/ElementX/Sources/Services/ElementCall/ElementCallService.swift
@@ -260,8 +260,6 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
     // MARK: - Private
     
     func tearDownCallSession(sendEndCallAction: Bool = true) {
-        try? AVAudioSession.sharedInstance().setActive(false)
-        
         if sendEndCallAction, let ongoingCallID {
             let transaction = CXTransaction(action: CXEndCallAction(call: ongoingCallID.callKitID))
             callController.request(transaction) { error in
--- a/ElementX/Sources/Services/Audio/Player/AudioPlayer.swift
+++ b/ElementX/Sources/Services/Audio/Player/AudioPlayer.swift
@@ -142,7 +142,7 @@ class AudioPlayer: NSObject, AudioPlayerProtocol {
     private func setupAudioSession() {
         releaseAudioSessionTask = nil
         do {
-            try audioSession.setCategory(AVAudioSession.Category.playback)
+            try audioSession.setCategory(.playback)
             try audioSession.setActive(true)
         } catch {
             MXLog.error("Could not redirect audio playback to speakers.")
@@ -166,7 +166,7 @@ class AudioPlayer: NSObject, AudioPlayerProtocol {
         releaseAudioSessionTask = nil
         if audioSession.category == .playback, !audioSession.isOtherAudioPlaying {
             MXLog.info("releasing audio session")
-            try? audioSession.setActive(false)
+            try? audioSession.setActive(false, options: .notifyOthersOnDeactivation)
         }
     }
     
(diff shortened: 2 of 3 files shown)


ElementX/Sources/Services/Audio/Player/AudioPlayerState.swift
```diff
@@ -200,6 +214,93 @@ class AudioPlayerState: ObservableObject, Identifiable {
...
+    private func setUpRemoteCommandCenter() {
+        UIApplication.shared.beginReceivingRemoteControlEvents()
+        
+        let commandCenter = MPRemoteCommandCenter.shared()
+        
+        commandCenter.playCommand.isEnabled = true
+        commandCenter.playCommand.removeTarget(nil)
+        commandCenter.playCommand.addTarget { [weak self] _ in
+            guard let audioPlayer = self?.audioPlayer else {
+                return MPRemoteCommandHandlerStatus.commandFailed
```
pixlwave, line 227: Can these all be simplified to?
```suggestion
                return .commandFailed
```

ElementX/Sources/Services/Audio/Player/AudioPlayerState.swift
```diff
@@ -200,6 +214,93 @@ class AudioPlayerState: ObservableObject, Identifiable {
...
+        
+        commandCenter.skipForwardCommand.isEnabled = true
+        commandCenter.skipForwardCommand.removeTarget(nil)
+        commandCenter.skipForwardCommand.addTarget { [weak self] event in
+            guard let audioPlayer = self?.audioPlayer, let skipEvent = event as? MPSkipIntervalCommandEvent else {
+                return MPRemoteCommandHandlerStatus.commandFailed
+            }
+            
+            Task {
+                await audioPlayer.seek(to: audioPlayer.currentTime + skipEvent.interval)
```
pixlwave, line 255: What happens if this is past the end?

ElementX/Sources/Services/Audio/Player/AudioPlayerState.swift
```diff
@@ -200,6 +214,93 @@ class AudioPlayerState: ObservableObject, Identifiable {
...
+        
+        commandCenter.skipBackwardCommand.isEnabled = true
+        commandCenter.skipBackwardCommand.removeTarget(nil)
+        commandCenter.skipBackwardCommand.addTarget { [weak self] event in
+            guard let audioPlayer = self?.audioPlayer, let skipEvent = event as? MPSkipIntervalCommandEvent else {
+                return MPRemoteCommandHandlerStatus.commandFailed
+            }
+            
+            Task {
+                await audioPlayer.seek(to: audioPlayer.currentTime - skipEvent.interval)
```
pixlwave, line 269: And equally, what happens if this is < 0?

ElementX/Sources/Services/ElementCall/ElementCallService.swift
```diff
@@ -260,8 +260,6 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
     // MARK: - Private
     
     func tearDownCallSession(sendEndCallAction: Bool = true) {
-        try? AVAudioSession.sharedInstance().setActive(false)
```
pixlwave, line 263: I'm not sure I understand this change. How come there isn't a corresponding `setActive(true)` that doesn't need tidying up too?

pixlwave (approved):
Nice!

ElementX/Sources/Services/ElementCall/ElementCallService.swift
stefanceriu: That's the problem, I removed the counterpart because CallKit already takes care of it. The deactivation should'be been removed as well but I missed it.

ElementX/Sources/Services/Audio/Player/AudioPlayerState.swift
stefanceriu: Nothing really, the AVPlayer guards against it. Same for the one above

ElementX/Sources/Services/ElementCall/ElementCallService.swift
pixlwave: Ahhh got it 👍

ElementX/Sources/Services/Audio/Player/AudioPlayerState.swift
pixlwave: Ok nice 👍

PR #3242, opened 2024-09-09
Pin Tweaks

Velin92:
Pun intended

- [x] Remove delete action from the pinned context menu
- [x] Move Pin/Unpin action below Edit
- [x] Timeline in pinned modes should NOT display the typing notifications
- [x] Remove the 1 of 1 copy when there is only 1 message in the banner
- [x] Filter reactions in the pinned timeline client side


pixlwave (approved):
Nice 😎

PR #3245, opened 2024-09-10
Fix wrong durations for uploaded media

stefanceriu:
At some point Uniffi introduced [automatic conversions](https://github.com/mozilla/uniffi-rs/blob/35140607dce7092f04d619a5d0b07f9ffe98b4ff/uniffi_bindgen/src/bindings/swift/templates/DurationHelper.swift#L19) from Rust's `Duration` to `TimeInterval`s but the original implementation for media uploading was using milliseconds. When the conversion happened there were no breaking changes and iOS kept sending millis instead of seconds, resulting in **very** wrong values.

Android seems fine as it's using [java.time.Duration](https://github.com/mozilla/uniffi-rs/blob/35140607dce7092f04d619a5d0b07f9ffe98b4ff/uniffi_bindgen/src/bindings/kotlin/templates/DurationHelper.kt#L32)

### DIFF
--- a/ElementX/Sources/Services/Media/MediaUploadingPreprocessor.swift
+++ b/ElementX/Sources/Services/Media/MediaUploadingPreprocessor.swift
@@ -77,7 +77,7 @@ private struct VideoProcessingInfo {
     let url: URL
     let height: Double
     let width: Double
-    let duration: Double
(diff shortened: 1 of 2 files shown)


pixlwave (approved):
Good spot!

PR #3247, opened 2024-09-10
crypto: Add configuration flag to enable invisible crypto

BillCarsonFr:
Fixes https://github.com/element-hq/element-x-ios/issues/2932

Adds a configuration flag (via `AppSettings` to enabled invisible crypto mode). The setting is Off by default.
Also adds a toggle to the dev tool screen for testing:
image

### Pull Request Checklist

- [ ] Pull request includes a [sign off](https://github.com/matrix-org/matrix-ios-sdk/blob/develop/CONTRIBUTING.md#sign-off).

- [ ] Voiceover enabled.


pixlwave:
Heads up that I just merged #3215 which also calls `roomKeyRecipientStrategy(strategy:)` on the ClientBuilder. I don't think it should conflict with your PR given it is called first, but feel free to move it around to make it clearer if you want :)

BillCarsonFr:
> Heads up that I just merged #3215 which also calls `roomKeyRecipientStrategy(strategy:)` on the ClientBuilder. I don't think it should conflict with your PR given it is called earlier, but feel free to move it around to make it clearer if you want :)

Ok, I'll rebase. I will move your call to `roomKeyRecipientStrategy` around the same place to set it as default unless the lab flag is set

BillCarsonFr:
> Ok, I'll rebase. I will move your call to `roomKeyRecipientStrategy` around the same place to set it as default unless the lab flag is set

@pixlwave  https://github.com/element-hq/element-x-ios/pull/3247/commits/a1e57398973d0316d7cbdb93f6df99d881e6aa5f

pixlwave:
Nice, thanks 👍

ElementX/Sources/Other/Extensions/ClientBuilder.swift
```diff
@@ -14,12 +14,12 @@ extension ClientBuilder {
                             httpProxy: String? = nil,
                             slidingSync: ClientBuilderSlidingSync,
                             sessionDelegate: ClientSessionDelegate,
-                            appHooks: AppHooks) -> ClientBuilder {
+                            appHooks: AppHooks,
+                            appSettings: CommonSettingsProtocol) -> ClientBuilder {
```
pixlwave, line 18: Could we replace this with an `invisibleCryptoEnabled: Bool` parameter rather than passing the settings in directly please?

ElementX/Sources/Application/AppSettings.swift
```diff
@@ -8,6 +8,12 @@
 import Foundation
 import SwiftUI
 
+// Common settings between app and NSE
+protocol CommonSettingsProtocol {
+    var logLevel: TracingConfiguration.LogLevel { get }
+    var invisibleCryptoEnabled: Bool { get }
+}
```
pixlwave, line 15: Nice, this seems much cleaner like this to me 👍

ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/View/DeveloperOptionsScreen.swift
```diff
@@ -52,6 +52,17 @@ struct DeveloperOptionsScreen: View {
...
             
+            Section {
+                Toggle(isOn: $context.invisibleCryptoEnabled) {
+                    Text("Enabled Invisible Crypto")
+                    Text("Requires app reboot")
+                }
+            } header: {
+                Text("Trust and Decoration")
+            } footer: {
+                Text("This setting controls how end-to-end encryption (E2E) keys are shared. Enabling it will prevent the inclusion of devices that have not been explicitly verified by their owners.")
```
pixlwave, line 63: Minor and optional nit: I would expect an extra `E` on the end of this given it comes after the word `encryption`.
```suggestion
                Text("This setting controls how end-to-end encryption (E2EE) keys are shared. Enabling it will prevent the inclusion of devices that have not been explicitly verified by their owners.")
```

pixlwave (approved):
Nice, its exciting to see invisible crypto being added.

PR #3252, opened 2024-09-10
Add a banner that offers the user to transition to native sliding sync

stefanceriu:
This PR introduces the following logic to support migrating users to native sliding sync
* if the user is logged in and using native SS then perfect, nothing to do
* if using a proxy then:
    * if native is available as advertised through /versions a banner will be shown to prompt logout
    * if the proxy is no longer available as advertised through .well-known then force logout

Screenshot 2024-09-10 at 16 59 02

### DIFF
--- a/ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
+++ b/ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
@@ -19,6 +19,7 @@ enum HomeScreenViewModelAction {
     case presentStartChatScreen
     case presentGlobalSearch
     case presentRoomDirectorySearch
+    case logoutWithoutConfirmation
     case logout
 }
 
@@ -31,6 +32,8 @@ enum HomeScreenViewAction {
     case startChat
     case confirmRecoveryKey
     case skipRecoveryKeyConfirmation
+    case confirmSlidingSyncUpgrade
+    case skipSlidingSyncUpgrade
     case updateVisibleItemRange(Range<Int>)
     case globalSearch
     case markRoomAsUnread(roomIdentifier: String)
@@ -59,10 +62,10 @@ enum HomeScreenRoomListMode: CustomStringConvertible {
     }
 }
 
-enum SecurityBannerMode {
+enum HomeScreenBannerMode {
     case none
     case dismissed
-    case recoveryKeyConfirmation
+    case shown
 }
 
 struct HomeScreenViewState: BindableState {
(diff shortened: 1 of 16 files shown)


ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
```diff
@@ -59,18 +62,20 @@ enum HomeScreenRoomListMode: CustomStringConvertible {
     }
 }
 
-enum SecurityBannerMode {
+enum HomeScreenBannerMode {
     case none
     case dismissed
-    case recoveryKeyConfirmation
+    case shown
```
Velin92, line 68: I would rename this to `showing` or `show`, instead of `shown` I was kinda confused while reviewing, because I thought `shown` meant that it **was** shown but had been dismissed by the user or by completing the flow.
Instead this seems to be the **currently** showing

stefanceriu: Fair point 👍

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -192,12 +201,15 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
     
     // perphery: ignore - used in release mode
     func presentCrashedLastRunAlert() {
-        state.bindings.alertInfo = AlertInfo(id: UUID(),
-                                             title: L10n.crashDetectionDialogContent(InfoPlistReader.main.bundleDisplayName),
-                                             primaryButton: .init(title: L10n.actionNo, action: nil),
-                                             secondaryButton: .init(title: L10n.actionYes) { [weak self] in
-                                                 self?.actionsSubject.send(.presentFeedbackScreen)
-                                             })
+        // Delay setting the alert otherwise it automatically gets dismissed. Same as the force logout one.
```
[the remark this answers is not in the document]
stefanceriu: I tested both and for logout the next run loop was fine but for the crash alert it wasn't so I just decided to keep it safe and use 0.5 in both.

Velin92 (approved):
LGTM

ElementX/Sources/Screens/HomeScreen/View/HomeScreenSlidingSyncMigrationBanner.swift
```diff
@@ -0,0 +1,68 @@
...
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+// Please see LICENSE in the repository root for full details.
+//
+
+import Combine
+import SwiftUI
+
+struct HomeScreenSlidingSyncMigrationBanner: View {
+    @ObservedObject var context: HomeScreenViewModel.Context
```
pixlwave, line 12: This doesn't need to be observed from what I can tell, `let context` should be fine here.

pixlwave (approved):
🚀

PR #3257, opened 2024-09-10
Added a pin icon

Velin92:
Note: The preview tests require a delay, and the delay function to set them in the tests seems to be partially broken, it only applies the delay on iPad but not on iPhone tests.

Simulator Screenshot - iPhone 15 - 2024-09-10 at 18 47 02

### DIFF
--- a/ElementX/Sources/Screens/Timeline/View/Style/TimelineItemBubbledStylerView.swift
+++ b/ElementX/Sources/Screens/Timeline/View/Style/TimelineItemBubbledStylerView.swift
@@ -322,24 +322,25 @@ private struct PinnedIndicatorViewModifier: ViewModifier {
     let isOutgoing: Bool
     
     func body(content: Content) -> some View {
-        HStack(alignment: .top, spacing: 8) {
-            if isOutgoing {
-                pinnedIndicator
+        if isPinned {
+            HStack(alignment: .top, spacing: 8) {
+                if isOutgoing {
+                    pinnedIndicator
+                }
+                content
+                    .layoutPriority(1)
+                if !isOutgoing {
+                    pinnedIndicator
+                }
             }
+        } else {
             content
-                .layoutPriority(1)
-            if !isOutgoing {
-                pinnedIndicator
-            }
         }
     }
     
-    @ViewBuilder
     private var pinnedIndicator: some View {
-        if isPinned {
-            CompoundIcon(\.pinSolid, size: .xSmall, relativeTo: .compound.bodyMD)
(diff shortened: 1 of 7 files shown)


ElementX/Sources/Screens/Timeline/View/Style/TimelineItemBubbledStylerView.swift
```diff
@@ -146,6 +153,7 @@ struct TimelineItemBubbledStylerView<Content: View>: View {
                     context.send(viewAction: .handleTimelineItemMenuAction(itemID: timelineItem.id, action: action))
                 }
             }
+            .pinnedIndicator(isPinned: isPinned, isOutgoing: timelineItem.isOutgoing)
```
stefanceriu, line 156: Feels very wasteful to wrap every single item in an HStack when the chances of it being pinned are slim to none. Can we rewrite this to bail out early please?

ElementX/Sources/Screens/Timeline/View/Style/TimelineItemBubbledStylerView.swift
```diff
@@ -555,4 +613,97 @@ struct TimelineItemBubbledStylerView_Previews: PreviewProvider, TestablePreview
         }
         .environmentObject(viewModel.context)
     }
+        
+    // These akwats include a reply
```
stefanceriu, line 617:
```suggestion
    // These **dont' know what it's actually supposed to be** include a reply
```

stefanceriu (changes requested):

stefanceriu:
>  The preview tests require a delay, and the delay function to set them in the tests seems to be partially broken, it only applies the delay on iPad but not on iPhone tests.

Not entirely sure what's wrong there but all the old tests were passing when the new snapshotting flows got merged. Let's tackle whatever it is separately from this PR

ElementX/Sources/Screens/Timeline/View/Style/TimelineItemBubbledStylerView.swift
Velin92: Sure

ElementX/Sources/Screens/Timeline/View/Style/TimelineItemBubbledStylerView.swift
Velin92: just some copy and paste garbage, removed

stefanceriu (approved):
Nice, ship it! 🚀

PR #3259, opened 2024-09-11
Force frame sizes for timeline items that are missing sizing info

stefanceriu:
Screenshot 2024-09-11 at 10 57 56


ElementX/Sources/Other/SwiftUI/Layout/TimelineMediaFrame.swift
```diff
@@ -11,21 +11,20 @@ extension View {
     /// Constrains the max height of a media item in the timeline, whilst preserving its aspect ratio.
     @ViewBuilder
     func timelineMediaFrame(height contentHeight: CGFloat?, aspectRatio contentAspectRatio: CGFloat?) -> some View {
-        let minMediaHeight = 100.0
-        let maxMediaHeight = 300.0
+        let minMediaSize = 100.0
+        let maxMediaSize = 300.0
```
pixlwave, line 15: A little bit of me thinks we should leave the naming of these 2 alone and add an additional `let defaultMediaSize = 100.0` for the new case. It might be adding a bit of redundancy but it's probably simpler to understand for someone new reading this code.

pixlwave (approved):
LGTM, one small thought about the naming.

ElementX/Sources/Other/SwiftUI/Layout/TimelineMediaFrame.swift
stefanceriu: Can do 👍

pixlwave (approved):
Still LGTM 😁

PR #3261, opened 2024-09-11
Make the SessionDirectories type responsible for cleaning up data.

pixlwave:
Addresses the remaining [comments](https://github.com/element-hq/element-x-ios/pull/3196#pullrequestreview-2262612591) from #3196


stefanceriu (approved):
Suuper nice cleanup! 👏

PR #3272, opened 2024-09-12
Make sure we reset the feature flag when upgrading to SSS.

pixlwave:
Currently if you tap the banner to Logout & Upgrade, if you have changed the discovery to Proxy only, you will be logged back in using SS and not SSS and will be presented with the banner again.

Same for the alert I would imagine too.


stefanceriu (approved):
Nice catch 👍

PR #3273, opened 2024-09-12
Make sure we don't reuse an old NSEUserSession after logging out and back in.

pixlwave:
There have been reports of users receiving blank notifications after upgrading to SSS. Turns out when the user logs out, the NSE isn't aware of this so attempts to re-use the same UserSession/Client from the old session.

This PR fixes it with an additional comparison at the start of the flow.


stefanceriu (approved):
🤦

Issue #3243, opened 2024-09-10
Can't swipe down to dismiss photo on iOS/iPadOS 18


Avi0n:
### Steps to reproduce
- Tap on any photo. Attempt to swipe down to dismiss the photo. See that you are unable to do so.

This bug can make you have to force close ElX:
1. Tap the room header
2. Tap the room's chat image
3. Attempt to swipe down to dismiss the photo
4. See that you are unable to do so
5. See that you must force close ElX in order to get rid of the photo as there is no other way to close the room's photo

#### What did you expect?
Photo to dismiss on swipe down
#### What happened instead?
Photo unable to be dismissed via swipe down

### Your phone model

iPhone 15 Pro Max

### Operating system version

iOS 18.0

### Application version

1.7.5 (121)

### Homeserver

matrix.org

### Will you send logs?

No

pixlwave:
Note: For now we've made sure there's a Done button to dismiss. There's a small tweak still to make that will restore swipe down to dismiss when the project is built using Xcode 16.

PR #3274, opened 2024-09-12
Fix interactive dismissal of our QLPreviewController iOS 18 (when built with Xcode 16).

pixlwave:
This PR is the second part of #3243 which (when built with Xcode 16) will allow the view to be interactively dismissed again. This is what it looks like the the source view as set up in the PR (it should be pretty similar to the current behaviour):

| Xcode 15, iOS 17 | Xcode 15, iOS 18 | Xcode 16, iOS 17 | Xcode 16, iOS 18 |
| - | - | - | - |
| Xcode 15 iOS 17 | Xcode 15 iOS 18 | Xcode 16 iOS 17 | Xcode 16 iOS 18 |


stefanceriu (approved):
🚀

amshakal:
🔥 🧨 🚒

PR #3276, opened 2024-09-12
Handle notifications properly when a call is happening.

### LINKED ISSUE #3240: PIP hangs around after you hang up a call.
### Steps to reproduce

1. Join a SPA EC call via EX
2. Hang up
3. Background the app
4. PIP of the "How was your call?" hangs around.

#### What did you expect?

PIP should only show video, not "How was your call?" text.

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

1.7.5 (121)

### Homeserver

_No response_

### Will you send logs?

No

pixlwave:
This PR makes 2 changes:
- Puts a call into picture in picture when a notification (or other room-based app route) is handled during the call so that the navigation is visible to the user - following up on an earlier [comment](https://github.com/element-hq/element-x-ios/pull/3181#discussion_r1721682250)
- Exits the PiP if it wasn't actually showing a call anymore (e.g. it had ended) when tapping the notification (or also when backgrounding the app). Closes #3240.

https://github.com/user-attachments/assets/3997a282-b305-42f9-be04-45f13008dc79

https://github.com/user-attachments/assets/d76a2e4a-ea00-44bc-ac99-169c73f1bb70

### DIFF
--- a/ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
+++ b/ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
@@ -223,6 +223,18 @@ private struct CallView: UIViewRepresentable {
             }
         }
         
+        nonisolated func pictureInPictureControllerDidStartPictureInPicture(_ pictureInPictureController: AVPictureInPictureController) {
+            Task { @MainActor in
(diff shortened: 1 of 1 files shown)


Velin92 (approved):
LGTM

PR #3279, opened 2024-09-13
Replace client side room awaiting with the SDKs new `awaitRoomRemoteEcho` method

### LINKED ISSUE #3256: New invites show entirely blank push notifs and room profiles.
### Steps to reproduce

1. Receive a DM.
2. Blank push notif is displayed
3. Tap on the blank push notif
4. App opens on entirely blank room profile
5. Go back to roomlist
6. See the invite is shown there with correct metadata.

100% repro: all invites are broken if you answer them via push, as far as I can see.

#### What did you expect?

Non-corrupt invite pushes, and non-blank room profile previews.

#### What happened instead?

image

then

image

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

702

### Homeserver

_No response_

### Will you send logs?

No

stefanceriu:
Partially fixes #3256 where the join room screen doesn't have details about an invited room because the sync loop hasn't run yet (the preview api can't help here either because previews are disabled). Also fixes edgecases in which the room summary was available but the room not fully synced when used for creating/joining rooms.

- the part about the push notification not having content seems to be related to a missing `event_id` field between SS and SSS

### DIFF
--- a/ElementX/Sources/Services/Client/ClientProxy.swift
+++ b/ElementX/Sources/Services/Client/ClientProxy.swift
@@ -350,8 +350,11 @@ class ClientProxy: ClientProxyProtocol {
                                                   invite: [userID],
                                                   avatar: nil,
                                                   powerLevelContentOverride: Self.roomCreationPowerLevelOverrides)
-            let result = try await client.createRoom(request: parameters)
-            return await waitForRoomSummary(with: .success(result), name: expectedRoomName)
+            let roomID = try await client.createRoom(request: parameters)
+            
+            await waitForRoomToSync(roomID: roomID)
+            
+            return .success(roomID)
         } catch {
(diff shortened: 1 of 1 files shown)


pixlwave (approved):
LGTM 👏

Is there another place where we do this? Maybe the whole can init timeline thing? Or would that not work?

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
pixlwave, line 1: What's going on with this change?

stefanceriu: There's a guard for the same exact thing right above it.

PR #3281, opened 2024-09-13
Add a WebRegistrationScreen (not included in the flow yet).

pixlwave:
This PR adds a screen that can perform registration using a helper URL. A follow-up PR will be made that includes the changes to the flow.

Note: This screen is based upon the login screen which also seems to include the logic in the Coordinator instead of the ViewModel. As it is temporary it didn't seem worth re-writing.

### DIFF
--- a/ElementX/Sources/Services/Client/Client.swift
+++ b/ElementX/Sources/Services/Client/Client.swift
@@ -1,0 +1,32 @@
+//
+// Copyright 2024 New Vector Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+// Please see LICENSE in the repository root for full details.
+//
+
+import Foundation
+import MatrixRustSDK
+
+extension ClientProtocol {
+    func getElementWellKnown() async -> Result<MatrixRustSDK.ElementWellKnown, ClientProxyError> {
+        do {
+            let serverName = if let userIDServerName = try? userIdServerName() {
+                "https://\(userIDServerName)"
+            } else {
+                server()
+            }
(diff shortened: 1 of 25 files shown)


ElementX/Sources/Services/Client/Client.swift
```diff
@@ -0,0 +1,32 @@
...
+import MatrixRustSDK
+
+extension ClientProtocol {
+    func getElementWellKnown() async -> Result<MatrixRustSDK.ElementWellKnown, ClientProxyError> {
+        do {
+            let serverName = if let userIDServerName = try? userIdServerName() {
+                "https://\(userIDServerName)"
+            } else {
+                server()
+            }
```
stefanceriu, line 18: Shouldn't this be the other way around, try to use `server()` first? 🤔

stefanceriu (approved):
Couple of comments but lgtm 👍

stefanceriu:
> also seems to include the logic in the Coordinator instead of the ViewModel

Super weird, I was wondering why we did that and I think we just copied it from Riot. Why we did it like that there beats me. We should leave a todo in the code to get rid of this one and rewrite the other.

ElementX/Sources/Services/Client/Client.swift
pixlwave: I went with the assumption that the overwhelming majority of the time that we query this, the user will already be logged in. So seemed better to use the user ID in this instance in case the user entered `matrix.server.com` instead of `server.com` when signing in.

PR #3288, opened 2024-09-17
quick and dirty /join command

ara4n:
In order to avoid getting crucified by the matrix community for shipping a matrix client which can't join rooms by alias (somewhat like shipping a web browser without a url bar), here's a quick go at adding a `/join` slash command as an easter-egg for powerusers.

I've probably done it wrong, but:
 * The slashcommand code is on the TimelineViewModel for now, alongside the logic which turns composer input into sending messages.
 * In order to actually open the room which you then join, it uses `openURL` - but it has to use the one inside the SwiftUI environment given that's where the local app routing happens (otherwise it goes and opens it in an external app).  So I pull an openURL handler in a closure through in the state bindings from the TimelineView to the TimelineViewModel.  I assume there may be a fancier way to get SwiftUI to handle the openURL, but this seems to work fine even if it's a bit ugly.  (How come AppMediator has a useless open() method which doesn't use the App's openURL impl?)

### Pull Request Checklist

### DIFF
--- a/ElementX/Sources/Services/Authentication/AuthenticationService.swift
+++ b/ElementX/Sources/Services/Authentication/AuthenticationService.swift
@@ -117,7 +117,7 @@ class AuthenticationService: AuthenticationServiceProtocol {
             return await userSession(for: client)
         } catch {
             MXLog.error("Failed logging in with error: \(error)")
-            // FIXME: How about we make a proper type in the FFI? 😅
+            // How about we make a proper type in the FFI? 😅
             guard let error = error as? ClientError else { return .failure(.failedLoggingIn) }
             
             switch error.code {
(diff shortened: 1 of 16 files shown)


ElementX/Sources/Services/Authentication/AuthenticationService.swift
```diff
@@ -117,7 +117,7 @@ class AuthenticationService: AuthenticationServiceProtocol {
             return await userSession(for: client)
         } catch {
             MXLog.error("Failed logging in with error: \(error)")
-            // FIXME: How about we make a proper type in the FFI? 😅
+            // FIXMOI: How about we make a proper type in the FFI? 😅
```
ara4n, line 120: to make swiftlint let me commit...

pixlwave (approved):
LGTM 👍

PR #3303, opened 2024-09-18
Update the project to use Xcode 16.

pixlwave:
Update the project to use Xcode 16 and the iOS 18 simulator.

- [x] Fastlane uses Xcode 16
- [x] Unit Tests
  - [x] Pass locally
  - [x] Pass on CI
- [x] Preview Tests
  - [x] Pass locally
  - [x] Pass on CI
- [ ] UI Tests
  - [x] Pass locally
  - [ ] Pass on CI (we're going to prioritise building with Xcode 16 over these as we need it to fix a bug with interactive dismissal of media).
- [x] Integration Tests
  - [x] Pass locally
  - [x] Pass on CI
- [x] Xcode Cloud uses Xcode 16

**Bugs**
- [x] You can't scroll the timeline by swiping on a bubble.
  - Fixed with the caveat that if you take too long to swipe to reply, you'll trigger the long-press gesture (https://forums.developer.apple.com/forums/thread/760035)


Velin92 (approved):
LGTM

PR #3308, opened 2024-09-19
Enable message pinning

Velin92:
enbling the FF for message pinning


pixlwave (approved):
LGTM, presumably we'll remove the flag in a follow-up?

Velin92:
> LGTM, presumably we'll remove the flag in a follow-up?

yup

PR #3316, opened 2024-09-23
Configure the AuthenticationService later now that we have 2 flows on the start screen.

### LINKED ISSUE #3028: Long timeout before proceeding to change account provider screen when matrix.org URL is blocked or inaccessible
### Steps to reproduce

The launch screen with Sign in manually button
2. What do you click?
Sign in manually button

#### What did you expect?
I should be able to change account provider regardless matrix.org is accessible or not. 
In Android Element X latest version, this issues does not occur.
#### What happened instead?
The loading indicator keeps going on forever when Sign in manually button is tapped while matrix.org is inaccessible.

### Your phone model

iPhone 11

### Operating system version

iOS 17.5.1

### Application version

1.16.13

### Homeserver

_No response_

### Will you send logs?

No

pixlwave:
This PR is the first of a series to re-work the authentication flow:
- We no longer configure the authentication service before proceeding from the `AuthenticationStartScreen`. Instead we do it either when selecting a server, or if this hasn't happened, when continuing to use the default homeserver. (This lines up with how the flow works on Android).
- We can now show an alert when attempting to register on a server that isn't open for registration instead of setting the subtitle on the screen which wasn't particularly obvious.
- Allow the creation of the AuthenticationService with a client builder factory so we can test it with a `ClientSDKMock`.
- Make the factory preconfigured enough to replace the hand built `MockAuthenticationService` with the real one in all of our tests.
- Add a bunch of tests on the `AuthenticationService` and `ServerConfirmationScreenViewModel`.

Fixes #3028 from the perspective of if matrix.org is blocked (although that long timeout still needs addressing if you were to tap continue to try matrix.org in this instance.

More tests to follow, but this PR has already grown larger than I would have liked. Definitely one to review commit-by-commit.


UnitTests/Sources/AuthenticationServiceTests.swift
```diff
@@ -0,0 +1,96 @@
...
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+// Please see LICENSE in the repository root for full details.
+//
+
+import XCTest
+
+@testable import ElementX
+
+class AuthenticationServiceTests: XCTestCase {
```
stefanceriu, line 12: 👏

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/ServerSelectionScreenCoordinator.swift
```diff
@@ -107,6 +106,8 @@ final class ServerSelectionScreenCoordinator: CoordinatorProtocol {
             viewModel.displayError(.invalidWellKnownAlert(error))
         case .slidingSyncNotAvailable:
             viewModel.displayError(.slidingSyncAlert)
+        case .registrationNotSupported:
+            viewModel.displayError(.registrationAlert) // TODO: [DOUG] Test me!
```
stefanceriu, line 110: 😅

UnitTests/Sources/AuthenticationServiceTests.swift
```diff
@@ -0,0 +1,92 @@
...
+            XCTAssertEqual(error, .registrationNotSupported)
+        }
+        
+        XCTAssertEqual(service.flow, .login)
+        XCTAssertEqual(service.homeserver.value, .init(address: "matrix.org", loginMode: .unknown))
+    }
+    
+    // MARK: - Helpers
+    
+    private func setupMocks(clientConfiguration: ClientSDKMock.Configuration = .init()) {
```
stefanceriu: Really nice!

ElementX/Sources/Services/Authentication/AuthenticationService.swift
```diff
@@ -197,3 +197,13 @@ class AuthenticationService: AuthenticationServiceProtocol {
         }
     }
 }
+
+// MARK: - Mocks
+
+extension AuthenticationService {
+    static var mock = AuthenticationService(userSessionStore: UserSessionStoreMock(configuration: .init()),
```
stefanceriu, line 204: 🙌

UnitTests/Sources/ServerConfirmationScreenViewModelTests.swift
```diff
@@ -11,5 +11,117 @@ import XCTest
 
 @MainActor
 class ServerConfirmationScreenViewModelTests: XCTestCase {
```
stefanceriu, line 13: image

stefanceriu (approved):
Very very nice cleanup and testing! 👏 
Talked about general api design concerns offline, feel free to tackle them in a separate PR if necessary.

pixlwave:
Merging this as the snapshot failures are emoji provider flakiness. Going to spend some time on these failures now.

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/ServerSelectionScreenCoordinator.swift
pixlwave: I have this already in a follow-up branch.

PR #3318, opened 2024-09-23
Remove message pinning FF

Velin92:
let's merge this after the release, just for safety

### DIFF
--- a/ElementX/Sources/Services/Client/ClientProxy.swift
+++ b/ElementX/Sources/Services/Client/ClientProxy.swift
@@ -816,12 +816,13 @@ class ClientProxy: ClientProxyProtocol {
         })
     }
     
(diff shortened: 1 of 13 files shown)


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -830,12 +831,6 @@ class ClientProxy: ClientProxyProtocol {
                                                    .policyRuleRoom,
                                                    .policyRuleServer,
                                                    .policyRuleUser]
-        
-        // Reminder: once the feature flag is not required anymore, change the lazy var back to a let
```
pixlwave, line 834: Nice reminder 👌

pixlwave (approved):
LGTM 😎

PR #3320, opened 2024-09-23
Fix a regression where you can't scroll the timeline on iOS 17

pixlwave:
It was caused by a fix for the same bug on iOS 18 🤦‍♂️

Seems we need to use separate paths for configuring these gestures. Might be worth exploring the use of UIGestureRecognizer within SwiftUI on iOS 18 at this point, but I'll leave that for a future PR.


pixlwave:
I'm going to yolo this one (it's not very controversial) to get it into the next Nightly.

PR #3323, opened 2024-09-24
Fix a bug where the Join Room screen was sometimes shown instead of the Room.

pixlwave:
Updates the SDK which fixes #3270 by including https://github.com/matrix-org/matrix-rust-sdk/pull/4021

PR #3324, opened 2024-09-24
Fix a bug on iOS 18 where the timeline background would use the wrong colour scheme when using the app switcher.

pixlwave:
Use the UIColors directly instead of wrapping the SwiftUI Colors.
| Before | After |
| - | - |
| before | after |

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -60,6 +60,12 @@ custom_rules:
     message: "Please use explicit spacing in HStacks."
     severity: warning
 
+  compound_color_conversion:
+    regex: "UIColor\\(\\.compound"
+    match_kinds: identifier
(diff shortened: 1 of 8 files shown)


.swiftlint.yml
```diff
@@ -60,6 +60,12 @@ custom_rules:
     message: "Please use explicit spacing in HStacks."
     severity: warning
 
+  compound_color_conversion:
+    regex: "UIColor\\(\\.compound"
+    match_kinds: identifier
+    message: "Please use the UIColor token directly to ensure the colour scheme is honoured."
+    severity: warning
+
```
stefanceriu, line 68: Suuuper nice 👌

pixlwave:
Same on the tests as #3323, they're being flakey on CI.

PR #3326, opened 2024-09-24
Upgrade Kingfisher to fix a bug that prevented GIFs from being tapped.

pixlwave:
For more info see https://github.com/onevcat/Kingfisher/issues/2295


Velin92 (approved):

pixlwave:
Another timeout failure.

PR #3330, opened 2024-09-25
Integration test improvements

stefanceriu:
* use the home screen search field to find the special integration tests room
* send 2 messages in that room following both the plain and the rich message composers
* redirect logs to `/Users/Shared`
* configure all rust logs to `tracing`
* check that the previously sent messages do not appear in the logs
* allow more time for a photo to be selected
* allow more time for pickers to load
* fix problems with the segmented control on the members screen


pixlwave (approved):
Nice 👍

IntegrationTests/Sources/UserFlowTests.swift
```diff
@@ -8,6 +8,9 @@
 import XCTest
 
 class UserFlowTests: XCTestCase {
+    private static let integrationTestsRoomName = "Element X iOS Integration Tests"
+    private static let integrationTestsMessage = "Go down in flames!"
```
bmarty, line 12: Maybe remove the `!` to exactly match what the grep is looking for?

.github/workflows/integration-tests.yml
```diff
@@ -37,6 +37,12 @@ jobs:
           INTEGRATION_TESTS_USERNAME: ${{ secrets.INTEGRATION_TESTS_USERNAME }}
           INTEGRATION_TESTS_PASSWORD: ${{ secrets.INTEGRATION_TESTS_PASSWORD }}
 
+      - name: Check logs are set to the `trace` level
+        run: (grep ' TRACE ' /Users/Shared -qR)
+
+      - name: Check logs don't contain private messages
+        run: (! grep 'Go down in flames' /Users/Shared -qR)
```
bmarty, line 44: To validate that the message is really not found, it's maybe worth trying to also check that an expected log is found? This will ensure that logs are present as expected in `/Users/Shared`

bmarty (commented):
Thanks for adding privacy check!

.github/workflows/integration-tests.yml
stefanceriu: Great minds, Doug literally suggested the same like an hour before you. I've taken care of it already, your other comment as well (for other reasons but whatever) 👍

ElementX/Sources/Other/Logging/RustTracing.swift
```diff
@@ -23,7 +29,13 @@ enum RustTracing {
         // as the app is unlikely to be running continuously.
         let maxFiles: UInt64 = 24 * 7
         
-        setupTracing(config: .init(filter: configuration.filter,
+        let filter = if ProcessInfo.isRunningIntegrationTests {
+            "trace"
+        } else {
+            configuration.filter
+        }
```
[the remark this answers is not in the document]
stefanceriu: Honestly no idea how I might pull that off. I did change the flow a bit so that we use `TracingConfiguration(logLevel: .trace, target: nil)` instead so I reckon that's good enough.

stefanceriu:
Going to merge this now. I will follow it up with a PR that disables custom tracing configurations (as we don't control 3rd party crates and don't want to take that risk) and a few other changes.

PR #3336, opened 2024-09-26
Tracing and integration test tweaks

stefanceriu:


ElementX/Sources/Other/Logging/TracingConfiguration.swift
```diff
@@ -27,32 +26,32 @@ struct TracingConfiguration {
...
+                true
+            case (.warn, .error):
+                false
+            case (.warn, _):
+                true
+            case (.info, .error), (.info, .warn):
+                false
+            case (.info, _):
+                true
+            case (.debug, .info), (.debug, .warn), (.debug, .error):
```
pixlwave, line 44: Nit: To keep it consistent with the one above.
```suggestion
            case (.debug, .error), (.debug, .warn), (.debug, .info):
```

ElementX/Sources/Other/Logging/TracingConfiguration.swift
```diff
@@ -67,7 +66,6 @@ struct TracingConfiguration {
     }
     
     static let targets: OrderedDictionary<Target, LogLevel> = [
-        .common: .info, // Never set this lower than info - 3rd-party crates may start logging user data.
```
pixlwave, line 70: Should we leave a comment here saying something like
> We don't include the common target because 3rd-party creates may end up logging user data.

ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/View/DeveloperOptionsScreen.swift
```diff
@@ -135,16 +135,8 @@ struct DeveloperOptionsScreen: View {
 private struct LogLevelConfigurationView: View {
     @Binding var logLevel: TracingConfiguration.LogLevel
     
-    @State private var customTracingConfiguration: String
-
     init(logLevel: Binding<TracingConfiguration.LogLevel>) {
```
pixlwave, line 138: We can use the synthesised init here now.

pixlwave (approved):
LGTM 👍

ElementX/Sources/Other/Logging/TracingConfiguration.swift
stefanceriu: Right, good point! 😅

ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/View/DeveloperOptionsScreen.swift
stefanceriu: Oh nice 👌

ElementX/Sources/Other/Logging/TracingConfiguration.swift
stefanceriu: Sure thing!

ElementX/Sources/Other/Logging/TracingConfiguration.swift
```diff
@@ -90,13 +88,13 @@ struct TracingConfiguration {
     /// - Parameter logLevel: the desired log level
     /// - Parameter target: the name of the target being configured
     /// - Returns: a custom tracing configuration
-    init(logLevel: LogLevel, target: String?) {
-        fileName = if let target {
-            "\(RustTracing.filePrefix)-\(target)"
+    init(logLevel: LogLevel, target: String, filePrefix: String?) {
```
pixlwave, line 91: Having been utterly confused looking at the new changes for a second, maybe we should rename `target` to `currentTarget` at this point (and in the MXLog.configure) but totally optional.

pixlwave (approved):

PR #3339, opened 2024-09-27
Log any failures when creating a call widget.

pixlwave:
I'm unable to use Element Call on my Nightly, and all I'm getting from the logs are

```
2024-09-27T09:39:50.559436Z ERROR elementx: Failed starting ElementCall Widget Driver with error: failedBuildingWidgetSettings | CallScreenViewModel.swift:163 | spans: root
```

Seems worth adding some additional logs here.


stefanceriu (approved):
👌

PR #3341, opened 2024-09-27
Allow focusing the different avatars making up a DM details cluster separately.

stefanceriu:


ElementX/Sources/Other/SwiftUI/Views/LoadableAvatarImage.swift
```diff
@@ -13,24 +13,42 @@ struct LoadableAvatarImage: View {
...
-            .background(Color.compound.bgCanvasDefault)
-            .clipShape(Circle())
+        if let onTap, let url {
+            Button {
+                onTap(url)
+            } label: {
+                avatar
+                    .frame(width: frameSize, height: frameSize)
+                    .background(Color.compound.bgCanvasDefault)
+                    .clipShape(Circle())
```
pixlwave, line 43: Probably worth extracting this out into its own property now that it is repeated.

pixlwave (approved):
Very nice 👌

PR #3356, opened 2024-10-01
Bump the RustSDK to v1.0.53: adopt latest record based timeline item APIs

stefanceriu:
* the TimelineItems and their content are now records, which in turns means a lot of changes had to happen.
* a new way to identity items was also introduced in the form of `eventOrTransactionId` which we now use within the TimelineItemIdentifier
* the `timelineID` has been renamed to `uniqueID` to match the SDK side and avoid confusion
* 
* various other tweaks and fixes


ElementX/Sources/Services/Timeline/TimelineItemProxy.swift
```diff
@@ -118,43 +132,43 @@ class EventTimelineItemProxy {
...
+    lazy var canBeRepliedTo = item.canBeRepliedTo
             
-    lazy var content = item.content()
+    lazy var content = item.content
 
-    lazy var isOwn = item.isOwn()
+    lazy var isOwn = item.isOwn
 
-    lazy var isEditable = item.isEditable()
+    lazy var isEditable = item.isEditable
```
stefanceriu, line 141: I wonder if it's worth it to rewrite these 🤔

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemFactory.swift
```diff
@@ -652,12 +638,12 @@ struct RoomTimelineItemFactory: RoomTimelineItemFactoryProtocol {
     // MARK: - Reply details
     
     func buildReply(details: InReplyToDetails) -> TimelineItemReply {
-        let isThreaded = details.event.isThreaded
-        switch details.event {
+        let isThreaded = details.event().isThreaded
```
pixlwave, line 641: Wait, did InReplyToDetails go the other way and become an object?

pixlwave (approved):
Very nice 👏

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemFactory.swift
stefanceriu: Seems so, yeah 😅

ElementX/Sources/Services/Timeline/TimelineItemProxy.swift
stefanceriu: Talked about it offline, we'll leave it alone for now

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemFactory.swift
stefanceriu: As per https://github.com/matrix-org/matrix-rust-sdk/commit/67df36f73356fe85b2d4004a6a365b77f44e2ada

`- Make "InReplyToDetails" an object instead since a record can't recursively contain itself.`

PR #3360, opened 2024-10-02
use element-hq RTE version

Velin92:
And update its acknowledgements license.


ElementX/SupportingFiles/Settings.bundle/Acknowledgements.plist
```diff
@@ -124,9 +124,9 @@
 		</dict>
 		<dict>
 			<key>File</key>
-			<string>Packages/matrix-rich-text-editor-swift</string>
+			<string>Packages/WysiwygComposer</string>
 			<key>Title</key>
-			<string>matrix-rich-text-editor-swift</string>
+			<string>WysiwygComposer</string>
```
stefanceriu, line 129: I prefer the old name to be honest

stefanceriu (approved):
Noice!

ElementX/SupportingFiles/Settings.bundle/Acknowledgements.plist
pixlwave: @Velin92 I think if you do a Reset package caches in Xcode it will revert back to `matrix-rich-text-editor-swift`.

PR #3368, opened 2024-10-04
Display a warning when a user's pinned identity changes

stefanceriu:
As per https://github.com/element-hq/element-meta/issues/2544

Using https://github.com/matrix-org/matrix-rust-sdk/pull/4022 and https://github.com/matrix-org/matrix-rust-sdk/pull/4068

https://github.com/user-attachments/assets/1d24f874-8b0d-4bf4-9f32-1c448bd7fb8e

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
+++ b/ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
@@ -7,18 +7,24 @@
 
 import Combine
 import Foundation
+import MatrixRustSDK
 import OrderedCollections
 import SwiftUI
 
 typealias RoomScreenViewModelType = StateStoreViewModel<RoomScreenViewState, RoomScreenViewAction>
 
 class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol {
+    private let clientProxy: ClientProxyProtocol
     private let roomProxy: JoinedRoomProxyProtocol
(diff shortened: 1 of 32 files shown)


ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -124,6 +139,21 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
         }
         .store(in: &cancellables)
         
+        let identityStatusChangesPublisher = roomProxy.identityStatusChangesPublisher.receive(on: DispatchQueue.main)
+        
+        Task { [weak self] in
+            for await changes in identityStatusChangesPublisher.values {
+                guard let self else { return }
```
Velin92, line 146: don't do `guard let self` if you check the previously used similiar solution this seems to create a retain cycle between each awaited published value, is better to keep the value optional so it's not strongified by mistake (I suspect this is a Swift bug acually)

Velin92 (changes requested):
LGTM, but there is one thing that needs to be changed

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
Velin92: Actually maybe we should add a comment to explain this bug in all its occurrences

Velin92 (approved):
LGTM

PR #3375, opened 2024-10-07
Update the SDK and use media `filename` and `caption` internally.

### LINKED ISSUE #3278: Images with long alt text can't be viewed fullscreen
### Steps to reproduce

When receiving images with a long alt text, trying to zoom in on the image will only show the full alt text.

Let's imagine I received 2 messages of type `m.image` that also contain a body as depicted here:
IMG_5967

The first image has a very short alt text - tapping on it opens the image in fullscreen and the alt text ("hello") is shown above:
IMG_5968

But now let's tap on the second image with a longer alt text.

#### What did you expect?
I would have expected a full screen version of the tapped image, so I could potentially zoom in on the image or simply look at it in a bigger version.

#### What happened instead?

An image with a longer alt text when tapped on, will show both an abbreviated version of the alt text at the top as well as the alt text in full instead of the image:

IMG_5969

### Your phone model

iPhone 12 mini

### Operating system version

iOS 17.6.1

### Application version

1.8.0 (122)

### Homeserver

Synapse 1.114.0

### Will you send logs?

No

pixlwave:
This PR adopts https://github.com/matrix-org/matrix-rust-sdk/pull/4081 so that we correctly compute a media file's filename/caption. It doesn't render captions in the timeline yet (it does in replies and the room list but essentially we did this anyway).

Fixes #3278


ElementX/Sources/Screens/Timeline/View/Style/TimelineItemBubbledStylerView.swift
```diff
@@ -612,96 +521,114 @@ struct TimelineItemBubbledStylerView_Previews: PreviewProvider, TestablePreview
...
-                                                      content: .init(body: "audio.ogg",
-                                                                     duration: 100,
-                                                                     waveform: EstimatedWaveform.mockWaveform,
-                                                                     source: nil,
-                                                                     contentType: nil),
-                                                      replyDetails: nil))
-            
-            FileRoomTimelineView(timelineItem: .init(id: .init(uniqueID: "", eventOrTransactionID: .eventId(eventId: "")),
-                                                     timestamp: "10:42",
+private struct MockTimelineContent: View {
```
stefanceriu, line 526: Nice 👍

stefanceriu (approved):
Awesome, ship it!

PR #3389, opened 2024-10-09
Ask the iPad to reveal the keyboard in UI Tests when it's hidden.

pixlwave:
Finally a successful run since upgrading to Xcode 16 🥳

https://github.com/element-hq/element-x-ios/actions/runs/11259522717

Thanks for all the help @manuroe!


stefanceriu (approved):
Yey! 👏

PR #3395, opened 2024-10-10
Fix identity pinning link.

pixlwave:
Based on https://github.com/element-hq/element-meta/issues/2544 the link is missing part of the fragment.

### DIFF
--- a/ElementX/Sources/Application/AppSettings.swift
+++ b/ElementX/Sources/Application/AppSettings.swift
@@ -126,7 +126,7 @@ final class AppSettings {
     /// A URL where users can go read more about the chat backup.
     let chatBackupDetailsURL: URL = "https://element.io/help#encryption5"
     /// A URL where users can go read more about identity pinning violations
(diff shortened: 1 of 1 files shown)


stefanceriu (approved):

ElementX/Sources/Application/AppSettings.swift
```diff
@@ -126,7 +126,7 @@ final class AppSettings {
     /// A URL where users can go read more about the chat backup.
     let chatBackupDetailsURL: URL = "https://element.io/help#encryption5"
     /// A URL where users can go read more about identity pinning violations
-    let identityPinningViolationDetailsURL: URL = "https://element.io/help#18"
+    let identityPinningViolationDetailsURL: URL = "https://element.io/help#encryption18"
```
bmarty, line 129: I double checked on Android we're [good](https://github.com/element-hq/element-x-android/blob/develop/appconfig/src/main/kotlin/io/element/android/appconfig/LearnMoreConfig.kt#L12).

PR #3407, opened 2024-10-14
Fix a bug opening images with a valid filename but a mimetype of `image/*` (sent by EXA).

pixlwave:
When renaming the get_media_file method in this [PR](https://github.com/matrix-org/matrix-rust-sdk/pull/4091) by chance I also looked at the switch and decided to match the inferred extension as `_` instead of `Some(_)` when the filename has an extension, which happily fixes this issue 🎉

So this PR simply updates the SDK.

### DIFF
--- a/ElementX/Sources/Services/Client/ClientProxy.swift
+++ b/ElementX/Sources/Services/Client/ClientProxy.swift
@@ -845,7 +845,7 @@ class ClientProxy: ClientProxyProtocol {
             let roomListItem = try roomListService.room(roomId: roomID)
             
             switch roomListItem.membership() {
-            case .invited:
+            case .invited, .knocked:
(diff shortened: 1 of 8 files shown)


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -845,7 +845,7 @@ class ClientProxy: ClientProxyProtocol {
             let roomListItem = try roomListService.room(roomId: roomID)
             
             switch roomListItem.membership() {
-            case .invited:
+            case .invited, .knocked:
```
pixlwave, line 848: We might want to add a `KnockedRoomProxy` for this in the future when there's a `roomListItem.knockedRoom()` but for now this seemed like a reasonable compromise as the semantics in the room list should be the same for both types.

Velin92 (approved):

PR #3408, opened 2024-10-14
Allow video uploads to be optimised to reduce bandwidth.

pixlwave:
This PR is the first part of #3398 and makes the following changes:

- Add a toggle (in Developer Options for now) to reduce the quality of media uploads
- Use this toggle to change the video quality preset from 1080p to 480p (which is 360p when video are 16:9).

### DIFF
--- a/ElementX/Sources/Application/AppSettings.swift
+++ b/ElementX/Sources/Application/AppSettings.swift
@@ -42,9 +42,9 @@ final class AppSettings {
         
         // Feature flags
         case slidingSyncDiscovery
+        case optimizeMediaUploads
         case publicSearchEnabled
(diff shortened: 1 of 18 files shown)


ElementX/Sources/Application/AppSettings.swift
```diff
@@ -42,9 +42,9 @@ final class AppSettings {
         
         // Feature flags
         case slidingSyncDiscovery
+        case optimizeMediaUploads
         case publicSearchEnabled
         case fuzzyRoomListSearchEnabled
-        case pinningEnabled
```
Velin92, line 47: whoops

PR #3418, opened 2024-10-16
Refactor the`TimelineItemIdentifier` handling

stefanceriu:
- stop relying on optional `EventOrTransactionId`s and be explicit when setting composer modes from the draft service.

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/ComposerToolbar/ComposerToolbarModels.swift
+++ b/ElementX/Sources/Screens/RoomScreen/ComposerToolbar/ComposerToolbarModels.swift
@@ -6,6 +6,7 @@
 //
 
 import Compound
+import MatrixRustSDK
 import SwiftUI
 import WysiwygComposer
 
@@ -283,9 +284,14 @@ extension FormatType {
 }
 
 enum ComposerMode: Equatable {
+    enum EditSource {
+        case timeline
+        case draftService
+    }
+    
     case `default`
-    case reply(itemID: TimelineItemIdentifier, replyDetails: TimelineItemReplyDetails, isThread: Bool)
-    case edit(originalItemId: TimelineItemIdentifier)
+    case reply(eventID: String, replyDetails: TimelineItemReplyDetails, isThread: Bool)
+    case edit(originalEventOrTransactionID: EventOrTransactionId, source: EditSource)
     case recordVoiceMessage(state: AudioRecorderState)
     case previewVoiceMessage(state: AudioPlayerState, waveform: WaveformSource, isUploading: Bool)
 
@@ -323,8 +329,8 @@ enum ComposerMode: Equatable {
     
     var replyEventID: String? {
         switch self {
-        case .reply(let itemID, _, _):
-            return itemID.eventID
+        case .reply(let eventID, _, _):
+            return eventID
         default:
             return nil
         }
--- a/ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift
+++ b/ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift
@@ -143,22 +143,13 @@ class RoomTimelineController: RoomTimelineControllerProtocol {
     
     func sendMessage(_ message: String,
                      html: String?,
-                     inReplyTo itemID: TimelineItemIdentifier?,
+                     inReplyToEventID: String?,
                      intentionalMentions: IntentionalMentions) async {
-        var inReplyTo: String?
-        if itemID == nil {
-            MXLog.info("Send message in \(roomID)")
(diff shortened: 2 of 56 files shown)


ElementX/Sources/Screens/RoomScreen/ComposerToolbar/ComposerToolbarModels.swift
```diff
@@ -283,9 +284,14 @@ extension FormatType {
 }
 
 enum ComposerMode: Equatable {
+    enum EditSource {
```
pixlwave, line 287: Could call this `EditActionSource` if you think it makes the `timeline` case clearer.

ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineControllerProtocol.swift
```diff
@@ -72,14 +73,3 @@ protocol RoomTimelineControllerProtocol {
...
-extension RoomTimelineControllerProtocol {
-    func sendMessage(_ message: String,
-                     html: String?,
-                     intentionalMentions: IntentionalMentions) async {
-        await sendMessage(message,
-                          html: html,
-                          inReplyTo: nil,
-                          intentionalMentions: intentionalMentions)
-    }
-}
```
pixlwave, line 85: Not sure what was wrong with this, seems like a reasonable thing to use when you don't want to reply?

ElementX/Sources/Services/Timeline/TimelineProxyProtocol.swift
```diff
@@ -112,14 +112,3 @@ protocol TimelineProxyProtocol {
...
-extension TimelineProxyProtocol {
-    func sendMessage(_ message: String,
-                     html: String?,
-                     intentionalMentions: IntentionalMentions) async -> Result<Void, TimelineProxyError> {
-        await sendMessage(message,
-                          html: html,
-                          inReplyTo: nil,
-                          intentionalMentions: intentionalMentions)
-    }
-}
```
pixlwave, line 125: Ditto.

ElementX/Sources/Services/Timeline/TimelineProxy.swift
```diff
@@ -631,4 +631,21 @@ extension Array where Element == TimelineItemProxy {
...
+            if case let .event(eventTimelineItem) = item {
+                switch eventTimelineItem.id {
+                case .event(_, let identifier):
+                    if identifier == eventOrTransactionID {
+                        return eventTimelineItem.item
+                    }
+                default:
+                    break
+                }
+            }
```
pixlwave, line 646: This might be slightly cleaner:
```suggestion
            if case let .event(eventTimelineItem) = item,
               case let .event(_, identifier) = eventTimelineItem.id,
               identifier == eventOrTransactionID {
                return eventTimelineItem.item
            }
```

ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineController.swift
```diff
@@ -398,9 +383,9 @@ class RoomTimelineController: RoomTimelineControllerProtocol {
                 let date = Date(timeIntervalSince1970: TimeInterval(timestamp / 1000))
                 let dateString = date.formatted(date: .complete, time: .omitted)
                 
-                return SeparatorRoomTimelineItem(id: .init(uniqueID: dateString), text: dateString)
+                return SeparatorRoomTimelineItem(id: .virtual(uniqueID: uniqueID), text: dateString)
```
pixlwave, line 386: Just want to highlight that this changed from `dateString` to `uniqueID`.

pixlwave (approved):
👏

Would be nice to have our own `EventOrTransactionID` enum so we have `ID` everywhere instead of `Id` but no need right now.

ElementX/Sources/Screens/RoomScreen/ComposerToolbar/ComposerToolbarModels.swift
stefanceriu: I think `ComposerMode.EditActionSource` is clear enough

ElementX/Sources/Services/Timeline/TimelineController/RoomTimelineControllerProtocol.swift
stefanceriu: I found them extra to be honest. On the highest level they're both called from the same place and each variant evetually trickles down into the same timeline one.

ElementX/Sources/Services/Timeline/TimelineProxy.swift
stefanceriu: Yup yup, nice, thank you!

PR #3424, opened 2024-10-16
Knock and knocked state for the join room screen

Velin92:
Some functionalities are still missing, also the SDK still sees the knocked rooms as rooms where you are invited (which is ofc not true)


ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenCoordinator.swift
```diff
@@ -34,6 +34,7 @@ final class JoinRoomScreenCoordinator: CoordinatorProtocol {
     init(parameters: JoinRoomScreenCoordinatorParameters) {
         viewModel = JoinRoomScreenViewModel(roomID: parameters.roomID,
                                             via: parameters.via,
+                                            appSettings: ServiceLocator.shared.settings,
```
pixlwave, line 37: We should really pass this in from above. It's there in the flow coordinator so only 1 extra parameter to add.

ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenModels.swift
```diff
@@ -48,23 +49,30 @@ struct JoinRoomScreenViewState: BindableState {
...
+        case .knocked: nil
         }
     }
     
     var avatar: RoomAvatar {
         roomDetails?.avatar ?? .room(id: roomID, name: title, avatarURL: nil)
     }
+    
+    var isEmptyKnockMessage: Bool {
+        bindings.knockMessage.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
```
pixlwave, line 61: Oh neat. Could add this as an extension on String tbf, something like `isEmptyAfterTrimmingWhitespace` or `isEmptyOrWhitespaceOnly`.

ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenViewModel.swift
```diff
@@ -135,7 +145,7 @@ class JoinRoomScreenViewModel: JoinRoomScreenViewModelType, JoinRoomScreenViewMo
         
         if roomPreviewDetails?.isPublic ?? false {
             state.mode = .join
-        } else if roomPreviewDetails?.canKnock ?? false, allowKnocking { // Knocking is not supported yet, the flag is purely for preview tests.
+        } else if roomPreviewDetails?.canKnock ?? false, appSettings.knockingEnabled {
```
pixlwave, line 148: This is outside of my understanding so question: Aren't we only allowing the creation of public rooms that can be knocked? If that's true, would this branch ever be reached as the `isPublic` would be hit first?

ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenViewModel.swift
```diff
@@ -171,6 +181,34 @@ class JoinRoomScreenViewModel: JoinRoomScreenViewModelType, JoinRoomScreenViewMo
...
+        } else {
+            switch await clientProxy.knockRoom(roomID,
+                                               message: state.isEmptyKnockMessage ? nil : state.bindings.knockMessage) {
+            case .success:
+                state.mode = .knocked
+            case .failure(let error):
+                MXLog.error("Failed knocking room id: \(roomID) with error: \(error)")
+                userIndicatorController.submitIndicator(.init(title: L10n.errorUnknown))
+            }
+        }
```
pixlwave, line 209: I think we could reduce the duplicates here (and for the other modes too) by doing

```
let result = if let alias {
    knockRoomAlias()
} else {
    knockRoom()
} 

switch result {
case .success: // stuff
case .failure: // stuff
}

File: ElementX/Sources/Mocks/KnockedRoomProxyMock.swift [hunk 5]
```diff
@@ -0,0 +1,29 @@
...
+    var members: [RoomMemberProxyMock] = .allMembers
+}
+
+extension KnockedRoomProxyMock {
+    @MainActor
+    convenience init(_ configuration: KnockedRoomProxyMockConfiguration) {
+        self.init()
+        id = configuration.id
+        name = configuration.name
+        avatarURL = avatarURL
```
pixlwave, line 25: SonarCloud is upset with you!
```suggestion
        avatarURL = configuration.avatarURL
```

ElementX/Sources/Screens/JoinRoomScreen/View/JoinRoomScreen.swift
```diff
@@ -66,18 +76,69 @@ struct JoinRoomScreen: View {
                         .multilineTextAlignment(.center)
                         .lineLimit(3)
                 }
+                
+                if context.viewState.mode == .knock {
+                    Spacer()
+                        .frame(height: 19)
+                    knockMessage
```
pixlwave, line 83: Minor nit:
```suggestion
                    knockMessage
                        .padding(.top, 19)
```

ElementX/Sources/Screens/JoinRoomScreen/View/JoinRoomScreen.swift
```diff
@@ -66,18 +76,69 @@ struct JoinRoomScreen: View {
...
+                    .lineLimit(4, reservesSpace: true)
+                    .font(.compound.bodyMD)
+                    .padding(.horizontal, 16)
+                    .padding(.vertical, 12)
+            }
+            .background(.compound.bgCanvasDefault)
+            .cornerRadius(8)
+            .overlay(RoundedRectangle(cornerRadius: 8)
+                .inset(by: 0.5)
+                .stroke(.compound.borderInteractivePrimary))
```
pixlwave, line 123: Technically I think that `overlay(some View)` is deprecated?
```suggestion
            .overlay {
                RoundedRectangle(cornerRadius: 8)
                    .inset(by: 0.5)
                    .stroke(.compound.borderInteractivePrimary)
            }
```

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -415,6 +415,30 @@ class ClientProxy: ClientProxyProtocol {
         }
     }
     
+    func knockRoom(_ roomID: String, message: String?) async -> Result<Void, ClientProxyError> {
```
pixlwave, line 418: I would imagine we (meaning the SDK) need the serverNames via parameters here like we have for Join, otherwise a Knock from a permalink might fail.

pixlwave (approved):
👌

ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenViewModel.swift
Velin92: we only allow the creation of public rooms that are knockable, however we will not prevent an existing E2EE to change its settings and become knockable

ElementX/Sources/Services/Client/ClientProxy.swift
Velin92: maybe we need to check with @jmartinesp about this, since I don't know.

ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenViewModel.swift
Velin92: the only problem is then is not easy to log in the error if is it was a roomID request or an alias one.

pixlwave: Fair 👍

PR #3428, opened 2024-10-18
Incoming session verification support

### LINKED ISSUE #1227: Verification Request from other devices
### Steps to reproduce

1. Be signed in and verified in Element X on Device A (Element X 1.1.7 (269) on iPhone 12 mini)
2. Be signed in and verified in Element on Device B (Element 1.5.32 Android on Fairphone 4)
3. Sign in on Element X Android on Device C (Element X 0.1.0-nightly June 30, 2023 build)
4. Tap Continue to request verification on the _Access your message history_ prompt

#### What did you expect?
Verification request from Device C should show up on both Device A and Device B

#### What happened instead?
Only Device B showed the request, Device A running Element X iOS did not.

### Your phone model

iPhone 12 mini

### Operating system version

iOS 16.5

### Application version

1.1.7 (269)

### Homeserver

matrix.org

### Will you send logs?

Yes

stefanceriu:
Introduce support for receiving and interacting with incoming device verification requests:
- adop new rust side ffi level session verification controller api
- update the sessionVerificationControllerProxy state machine and available screens
- introduce a new sessionVerificationControllerProxy actions publisher and hook it onto the UserSessionFlowCoordinator to display incoming requests
- update overall verification designs and copy, introduce more previews

Relates to element-hq/element-meta/issues/2464, fixes #1227

Powered by https://github.com/matrix-org/matrix-rust-sdk/pull/4153

https://github.com/user-attachments/assets/7d367f3c-cc2f-42a5-92c6-d4970c76af02


ElementX/Sources/Screens/Onboarding/SessionVerificationScreen/SessionVerificationScreenViewModel.swift
```diff
@@ -22,21 +23,25 @@ class SessionVerificationScreenViewModel: SessionVerificationViewModelType, Sess
...
         
-        sessionVerificationControllerProxy.callbacks
+        sessionVerificationControllerProxy.actions
             .receive(on: DispatchQueue.main)
             .sink { [weak self] callback in
                 guard let self else { return }
                 
                 switch callback {
+                case .receivedVerificationRequest:
+                    break
```
pixlwave, line 44: Could we add a comment why we don't do anything here (I assume its because we now listen to this elsewhere but would be helpful to see in line here?

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -157,7 +159,16 @@ class ClientProxy: ClientProxyProtocol {
...
-            self?.updateVerificationState(verificationState)
+            guard let self else { return }
+            
+            updateVerificationState(verificationState)
+            
+            // The session verification controller requires the user's identity which
+            // isn't available before a keys query response. Use the verification
+            // state updates as an aproximation for when that happens.
+            Task {
+                await self.buildSessionVerificationControllerProxyIfPossible(verificationState: verificationState)
```
pixlwave, line 170: Is it an issue if this gets run twice? Thinking about the await within it which might give this another chance to fire before the `guard controller == nil` would bail out on the second call.

ElementX/Sources/Services/ElementCall/ElementCallService.swift
pixlwave, line 1: Unrelated to this PR I think?

ElementX/Sources/Services/SessionVerification/SessionVerificationControllerProxy.swift
```diff
@@ -54,86 +58,136 @@ class SessionVerificationControllerProxy: SessionVerificationControllerProxyProt
...
     }
     
     deinit {
         sessionVerificationController.setDelegate(delegate: nil)
     }
     
-    let callbacks = PassthroughSubject<SessionVerificationControllerProxyCallback, Never>()
+    let actions = PassthroughSubject<SessionVerificationControllerProxyAction, Never>()
+    
+    func acknowledgeVerificationRequest(senderID: String, flowID: String) async -> Result<Void, SessionVerificationControllerProxyError> {
```
pixlwave, line 70: From what I can tell we get both the senderID and the flowID from the `SessionVerificationRequestDetails`? So could we just pass that in directly here?

ElementX/Sources/Services/SessionVerification/SessionVerificationControllerProxy.swift
```diff
@@ -54,86 +58,136 @@ class SessionVerificationControllerProxy: SessionVerificationControllerProxyProt
...
+        do {
+            try await sessionVerificationController.acknowledgeVerificationRequest(senderId: senderID, flowId: flowID)
+            return .success(())
+        } catch {
+            MXLog.error("Failed requesting session verification with error: \(error)")
+            return .failure(.failedAcknowledgingVerificationRequest)
+        }
+    }
+    
+    func acceptVerificationRequest(senderID: String, flowID: String) async -> Result<Void, SessionVerificationControllerProxyError> {
```
pixlwave, line 82: And here too?

ElementX/Sources/Screens/Onboarding/SessionVerificationScreen/View/SessionVerificationScreen.swift
```diff
@@ -5,16 +5,18 @@
 // Please see LICENSE in the repository root for full details.
 //
 
+import Compound
+import MatrixRustSDK
```
pixlwave, line 9: This seems wrong, we shouldn't be needing the Rust SDK in the view itself - we have the view model and the models to map SDK data to actual view state.

ElementX/Sources/Screens/Onboarding/SessionVerificationScreen/View/SessionVerificationScreen.swift
```diff
@@ -83,22 +59,73 @@ struct SessionVerificationScreen: View {
...
-                .tint(.compound.textSecondary)
-                .scaleEffect(2)
         default:
-            // In All other cases, we just want an empty view
             EmptyView()
         }
     }
     
+    private let outerShape = RoundedRectangle(cornerRadius: 8)
+    private func requestDetails(_ details: SessionVerificationRequestDetails) -> some View {
```
pixlwave, line 78: Can we extract this out into its own view struct rather than using a method?

ElementX/Sources/Screens/Onboarding/SessionVerificationScreen/View/SessionVerificationScreen.swift
```diff
@@ -194,29 +228,159 @@ struct SessionVerificationScreen: View {
             .padding(8.0)
         }
     }
+    
+    private var headerImageName: String {
```
pixlwave, line 232: I've got to be honest, I'm not a fan of moving this amount of logic into the view, when it could happily be used from the model.

pixlwave (approved):
Overall looks good to me, can't wait to have this working 🙌

ElementX/Sources/Services/SessionVerification/SessionVerificationControllerProxy.swift
bmarty: (In this method it seems that the 2 params are not used)

stefanceriu: Nice catch, yeah, those are left over from the first version of the API, before introducing `acknowledgeVerificationRequest`

ElementX/Sources/Screens/Onboarding/SessionVerificationScreen/View/SessionVerificationScreen.swift
stefanceriu: The thing is, the image was in the view and the titles in the view state. Switching between the 2 got old pretty fast while designing the new versions. 
Also, they're not really logic, just static view configs. I honestly have no strong feeling one way or another, it would be nice if we could find some good reason to help us chose.

ElementX/Sources/Services/ElementCall/ElementCallService.swift
stefanceriu: It's part of that `[Fix a couple of random small warnings](https://github.com/element-hq/element-x-ios/pull/3428/commits/b2280b948f6c9cb99a519a72296ae6acd2b1ae76)` commit

ElementX/Sources/Screens/Onboarding/SessionVerificationScreen/View/SessionVerificationScreen.swift
pixlwave: View config/state seem like the same things to me. I think I tend to move these kind of things from view to the models when it becomes much more than an `if else` to keep the view focussed on layout. At least if it's in the models, it becomes unit testable without having to spin up a preview for every state.

stefanceriu: Okay fair, that's a very good point 👍

ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu: It would create a new `SessionVerificationControllerProxy` on top of the same rust `SessionVerificationController` instance but it wouldn't really change or affect anything.

ElementX/Sources/Services/SessionVerification/SessionVerificationControllerProxy.swift
stefanceriu: Sure thing 👍

PR #3437, opened 2024-10-23
Adopt various Rust side API changes

stefanceriu:


Velin92 (approved):

ElementX/Sources/Services/Client/ClientProxyProtocol.swift
```diff
@@ -39,16 +39,7 @@ enum ClientProxyError: Error {
 }
 
 enum SlidingSyncConstants {
-    static let defaultTimelineLimit: UInt32 = 20
```
pixlwave, line 42: Where did this go, wasn't the idea that one day we would use a higher value here for e.g. iPad so that the screen is pre-filled better like we do on EI?

pixlwave: Oh or was that pagination size, I can't remember now.

stefanceriu: It moved to the rust side. We agreed that until we make sliding things work properly it's best to keep things in one single place.

PR #3438, opened 2024-10-23
Switch fastlane's xcodebuild formatter from xcpretty to xcbeautify

stefanceriu:


stefanceriu:
Closing this as xcbeautify was already being used by default 🤦‍♂️

PR #3439, opened 2024-10-23
Update HeroImage to match the BigIcon component from Compound.

pixlwave:
`HeroImage` is now completely in line with the [Figma](https://www.figma.com/design/ux3tYoZV9WghC7hHT9Fhk0/Compound-iOS-Components?node-id=1662-28402&m=dev) (inc renaming to `BigIcon`). It is still in the EX project for now to be less disruptive as this PR feels far reaching enough already. We can move it across in a following PR.

This is the iOS equivalent to https://github.com/element-hq/element-x-android/pull/3719


stefanceriu (approved):
👏 🚀 👏

PR #3441, opened 2024-10-23
Fix a bug where the pinned items banner could overlay the composer.

### LINKED ISSUE #3411: On small screens (iPhone SE), you can't cancel edits in rooms with pinned msgs
### Steps to reproduce

[sm says](https://matrix.to/#/!amwlFMejXQcoCxNYHH:matrix.org/$soDI_OAx8KOfBeLRU5OSbKoklggSeYAWhRcH-zHUx2M?via=matrix.org&via=element.io&via=beeper.com):
> EXI devs: on a small screen (iPhone SE), when editing a long message, the pinned messages banner overlays the X button, so you can’t cancel the edit (only accept it)

#### What did you expect?

UI should still work on small screens.

Perhaps "if editing or replying to message, hide pinned msgs" would do it?

#### What happened instead?

UI became unusable due to small screen.

### Your phone model

iPhone SE of some kind

### Operating system version

_No response_

### Application version

_No response_

### Homeserver

_No response_

### Will you send logs?

No

pixlwave:
This PR puts the overlay on just the timeline, and not the timeline + composer. Probably not a final fix given the clipping that takes place, but good enough for now. Closes #3411.

https://github.com/user-attachments/assets/b224196c-21fd-490d-b49f-bdff3a48f8e2

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
+++ b/ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
@@ -28,6 +28,14 @@ struct RoomScreen: View {
     var body: some View {
         timeline
             .background(Color.compound.bgCanvasDefault.ignoresSafeArea())
+            .overlay(alignment: .top) {
+                Group {
+                    if roomContext.viewState.shouldShowPinnedEventsBanner {
+                        pinnedItemsBanner
+                    }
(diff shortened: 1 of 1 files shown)


stefanceriu (approved):
Nice and simple 👍

PR #3442, opened 2024-10-24
Fix a bug where the room state wouldn't indicate when a call was in progress.

pixlwave:
Fixes a bug with call state by including https://github.com/matrix-org/matrix-rust-sdk/pull/4166. Additionally updates the analytics events package to handle some new UTD cases.

https://github.com/user-attachments/assets/ad12ca42-8c2c-4911-be73-70b6e9fa8221


stefanceriu (approved):
lgtm, 🚀 it!

PR #3450, opened 2024-10-25
Add alias to public room creation

Velin92:
fixes https://github.com/element-hq/element-x-ios/issues/3419
And update the SDK to fix some bugs on the availability and validity API, but should not impact the current iOS code.

https://github.com/user-attachments/assets/eb6c0b0d-af95-4a87-85b1-0cd78886e124


ElementX/Sources/Screens/CreateRoom/CreateRoomModels.swift
```diff
@@ -25,20 +25,24 @@ enum CreateRoomViewModelAction {
 }
 
 struct CreateRoomViewState: BindableState {
+    var roomName: String
+    let homeserver: String
```
pixlwave, line 29: I think it would be good to be consistent here and call this `serverName` so it's clear that it isn't a homeserver URL.

ElementX/Sources/Screens/CreateRoom/View/CreateRoomScreen.swift
```diff
@@ -150,22 +156,73 @@ struct CreateRoomScreen: View {
                                     iconAlignment: .top),
                     kind: .selection(isSelected: !context.isRoomPrivate) { context.isRoomPrivate = false })
         } header: {
-            Text(L10n.commonSecurity.uppercased())
+            Text(L10n.screenCreateRoomRoomVisibilitySectionTitle.uppercased())
```
pixlwave, line 159: Not related to your changes, but I don't think we need the `uppercased()` here, it'll be handled by the header automatically.
```suggestion
            Text(L10n.screenCreateRoomRoomVisibilitySectionTitle)
```

ElementX/Sources/Screens/CreateRoom/View/CreateRoomScreen.swift
```diff
@@ -150,22 +156,73 @@ struct CreateRoomScreen: View {
...
     
     private var roomAccessSection: some View {
         Section {
-            ListRow(label: .plain(title: L10n.screenCreateRoomAccessSectionAnyoneOptionTitle,
-                                  description: L10n.screenCreateRoomAccessSectionAnyoneOptionDescription),
-                    kind: .selection(isSelected: !context.isKnockingOnly) { context.isKnockingOnly = false })
-            ListRow(label: .plain(title: L10n.screenCreateRoomAccessSectionKnockingOptionTitle,
-                                  description: L10n.screenCreateRoomAccessSectionKnockingOptionDescription),
+            ListRow(label: .plain(title:
+                L10n.screenCreateRoomRoomAccessSectionAnyoneOptionTitle,
```
pixlwave, line 167: I think this'll trigger some other formatting fixes too
```suggestion
            ListRow(label: .plain(title: L10n.screenCreateRoomRoomAccessSectionAnyoneOptionTitle,
```

ElementX/Sources/Screens/CreateRoom/View/CreateRoomScreen.swift
```diff
@@ -150,22 +156,73 @@ struct CreateRoomScreen: View {
...
+                description: L10n.screenCreateRoomRoomAccessSectionAnyoneOptionDescription),
+            kind: .selection(isSelected: !context.isKnockingOnly) { context.isKnockingOnly = false })
+            ListRow(label: .plain(title: L10n.screenCreateRoomRoomAccessSectionKnockingOptionTitle,
+                                  description: L10n.screenCreateRoomRoomAccessSectionKnockingOptionDescription),
                     kind: .selection(isSelected: context.isKnockingOnly) { context.isKnockingOnly = true })
         } header: {
-            Text(L10n.screenCreateRoomAccessSectionHeader.uppercased())
+            Text(L10n.screenCreateRoomRoomAccessSectionHeader.uppercased())
```
pixlwave, line 174:
```suggestion
            Text(L10n.screenCreateRoomRoomAccessSectionHeader)
```

ElementX/Sources/Screens/CreateRoom/View/CreateRoomScreen.swift
```diff
@@ -150,22 +156,73 @@ struct CreateRoomScreen: View {
...
+                    Text(":\(context.viewState.homeserver)")
+                        .font(.compound.bodyLG)
+                        .foregroundStyle(.compound.textSecondary)
+                }
+                .padding(ListRowPadding.textFieldInsets)
+                .environment(\.layoutDirection, .leftToRight)
+                .errorBackground(!context.viewState.aliasErrors.isEmpty)
+            })
+        } header: {
+            Text(L10n.screenCreateRoomRoomAddressSectionTitle.uppercased())
```
pixlwave, line 209:
```suggestion
            Text(L10n.screenCreateRoomRoomAddressSectionTitle)
```

ElementX/Sources/Screens/CreateRoom/View/CreateRoomScreen.swift
```diff
@@ -150,22 +156,73 @@ struct CreateRoomScreen: View {
...
+                                  description: L10n.screenCreateRoomRoomAccessSectionKnockingOptionDescription),
                     kind: .selection(isSelected: context.isKnockingOnly) { context.isKnockingOnly = true })
         } header: {
-            Text(L10n.screenCreateRoomAccessSectionHeader.uppercased())
+            Text(L10n.screenCreateRoomRoomAccessSectionHeader.uppercased())
+                .compoundListSectionHeader()
+        }
+    }
+    
+    private var roomAddressSection: some View {
```
pixlwave, line 179: Could we call this roomAliasSection as that's what we're really setting here?

ElementX/Sources/Screens/CreateRoom/View/CreateRoomScreen.swift
```diff
@@ -150,22 +156,73 @@ struct CreateRoomScreen: View {
...
+            VStack(alignment: .leading, spacing: 12) {
+                if context.viewState.aliasErrors.contains(.alreadyExists) {
+                    Label(L10n.screenCreateRoomRoomAddressNotAvailableErrorDescription, icon: \.error, iconSize: .xSmall, relativeTo: .compound.bodySM)
+                        .foregroundStyle(.compound.textCriticalPrimary)
+                        .font(.compound.bodySM)
+                } else if context.viewState.aliasErrors.contains(.invalidSymbols) {
+                    Label(L10n.screenCreateRoomRoomAddressInvalidSymbolsErrorDescription, icon: \.error, iconSize: .xSmall, relativeTo: .compound.bodySM)
+                        .foregroundStyle(.compound.textCriticalPrimary)
+                        .font(.compound.bodySM)
+                }
```
pixlwave, line 221: This could be cleaner if we had an optional `context.viewState.aliasErrorDescription` that returns the right string and then you only need to include the label once in this view.

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -205,6 +221,10 @@ class ClientProxy: ClientProxyProtocol {
         client.homeserver()
     }
     
+    var serverName: String? {
```
pixlwave, line 224: Might be good to also call this userIDServerName to be clear that it will be `nil` when not logged in.

ElementX/Sources/Screens/CreateRoom/CreateRoomViewModel.swift
```diff
@@ -15,6 +16,8 @@ class CreateRoomViewModel: CreateRoomViewModelType, CreateRoomViewModelProtocol
     private var createRoomParameters: CreateRoomFlowParameters
     private let analytics: AnalyticsService
     private let userIndicatorController: UserIndicatorControllerProtocol
+    private var syncNameAndAddress = true
```
pixlwave, line 19:
```suggestion
    private var syncNameAndAlias = true
```

ElementX/Sources/Screens/CreateRoom/CreateRoomViewModel.swift
```diff
@@ -80,31 +91,111 @@ class CreateRoomViewModel: CreateRoomViewModelType, CreateRoomViewModelProtocol
...
+            syncNameAndAddress = false
+        case .updateName(let name):
+            // Reset the syncing if the name is fully cancelled
+            if name.isEmpty {
+                syncNameAndAddress = true
+            }
+            state.roomName = name
+            if syncNameAndAddress {
+                state.addressName = roomAliasNameFromRoomDisplayName(roomName: name)
+            }
```
pixlwave, line 107: This pattern seems weird with the custom binding calling a view action. Wouldn't it work to use a viewstate binding and then sink both `context.$viewState.map(\.bindings..addressName)` and `context.$viewState.map(\.bindings.roomName)` like we're already doing below?

ElementX/Sources/Screens/CreateRoom/CreateRoomModels.swift
```diff
@@ -25,20 +25,24 @@ enum CreateRoomViewModelAction {
...
     let isKnockingFeatureEnabled: Bool
     var selectedUsers: [UserProfileProxy]
+    var addressName: String
```
pixlwave, line 32: And with this one could we call it `aliasLocalPart`?

pixlwave (approved):
LGTM 👍

ElementX/Sources/Screens/CreateRoom/CreateRoomViewModel.swift
Velin92: sadly not, because the action has the purpose of understanding WHERE the change did take place (the name text field or the alias text field).
Since we want to disable the sync only if the alias is changed but not the server name, while the bindings would listen to any change without being able to infer what is actually changing them.
Same for resetting the sync if the name is empty

Velin92: Also... I feel the view action makes the syncing very easy to test and reproduce

pixlwave: Ok makes sense. Could we maybe move the 2 custom bindings to a centralised place then rather than being inline with the subviews that use them?

PR #3453, opened 2024-10-25
Hook reaction pickers into the system's recently used keyboard emojis

stefanceriu:
The way this works is by opening the `com.apple.EmojiPreferences` suite and then getting the `EMFDefaults.EMFRecentsKey` from it. The same goes in opposite direction when chosing a reaction from the app.

NB
* the feature is currently only enabled on development builds
* there will be no changes on the mac until we figure out what to do with the context menu

Screenshot 2024-10-25 at 16 17 35

Screenshot 2024-10-25 at 16 17 25


ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
```diff
@@ -132,6 +136,8 @@ class TimelineViewModel: TimelineViewModelType, TimelineViewModelProtocol {
         case .itemSendInfoTapped(let itemID):
             handleItemSendInfoTapped(itemID: itemID)
         case .toggleReaction(let emoji, let itemID):
+            emojiProvider.markEmojiAsFrequentlyUsed(emoji)
```
pixlwave, line 139: This doesn't take into account that `toggle` might be about to redact a reaction. Seeing as you didn't "pick" it maybe we wouldn't include the update here?

ElementX/Sources/Screens/Timeline/View/ItemMenu/TimelineItemMenuAction.swift
```diff
@@ -8,26 +8,37 @@
...
+          debugActions: [TimelineItemMenuAction],
+          emojiProvider: EmojiProviderProtocol) {
         if !isReactable, actions.isEmpty, debugActions.isEmpty {
             return nil
         }
         
         self.actions = actions
         self.debugActions = debugActions
+        
+        var frequentlyUsed = emojiProvider.frequentlyUsedSystemEmojis().prefix(5).map { TimelineItemMenuReaction(key: $0, symbol: .heart) }
```
pixlwave, line 28: I don't *think* you need this prefix when there's another one further down 🤔
```suggestion
        var frequentlyUsed = emojiProvider.frequentlyUsedSystemEmojis().map { TimelineItemMenuReaction(key: $0, symbol: .heart) }
```

ElementX/Sources/Services/Emojis/EmojiProvider.swift
```diff
@@ -7,38 +7,79 @@
...
+            partialResult + category.emojis
+        }
+        
+        let frequentlyUsedEmojis = frequentlyUsedSystemEmojis().prefix(20)
+        
+        let emojis = allEmojis.filter { frequentlyUsedEmojis.contains($0.unicode) }
+        
+        if !emojis.isEmpty {
+            emojiCategories.insert(.init(id: EmojiCategory.frequentlyUsedCategoryIdentifier, emojis: emojis), at: 0)
+        }
```
pixlwave, line 40: Does this need to be guarded by the feature flag too?

Also, what is the filter doing? Removing an emoji from its category if it's in the frequently used? That seems unexpected if true. Worth adding a comment to explain it though.

frebib:
Does this mean the information won't be sync'ed with Element Web's own tracking of frecently used emojis?

Hywan:
That's neat!

Hywan:
@frebib indeed, this information is not synced, it's only device-specific.

simonmichael:
If this is what I think it is, thank you very much! Emoji favourites save much time.

With today's 1.9.4 testflight EXI build I see two emoji pickers.

1. when I focus the message text field and press the keyboard's emoji button:
image

2. when I long press on a message:
image

And when I press on the more icon here I get

3. a full-screen vertical scroll picker:
image

I can't quite remember what it showed before this release - something less native-looking that didn't know my favourites (perhaps it was 3). 1 is much more usable. Is it intended for favourites to be accessible in case 2 as well ?

simonmichael:
Update for other users/testers after some advice from chat.

This is disabled by default. To test it, you need to enable the secret developer options screen in app settings, then enable the Timeline -> Show frequently used emojis toggle:

image

After doing this, reacting to a message (my image 2 above) shows your recent reactions instead of generic ones , and the full browser (my image 3) now has a Frequently used section at the top.

PR #3452, opened 2024-10-25
Tweak the flow for changing a recovery key.

pixlwave:
Closes #3447

I'm not sure why the "Set up" snapshot doesn't match the preview. Tried with a longer timeout but that had no effect - the preview looks like this:

Screenshot 2024-10-25 at 11 39 56 am

PR #3456, opened 2024-10-28
Translations update

RiotRobot:
- Translations update


pixlwave:
#3452 already contains snapshot changes for a bunch of these strings.

PR #3458, opened 2024-10-28
Revert "min macos support"

pixlwave:
Turns out we didn't need #3451, the issue was that App Store Connect was comparing the uploaded build (which resolves to 14.6) to the one selected for submission to the App Store (which was still resolving to 13.3). Everything should be automatic based on the iOS version.


pixlwave:
Note: This was reapplied on `develop` as the changes to ASC didn't help.

PR #3461, opened 2024-10-29
Use an https callback for OIDC once again.

### LINKED ISSUE #1936: OIDC redirect sometimes fails to be picked up.

pixlwave:
`ASWebAuthenticationSession` added a new initialiser that takes a `Callback` with support for either a scheme (what we were using) or an https callback. This means we can revert #1937 and go back to using a regular https callback. As this isn't using universal links we don't need to use the bundle id to disambiguate between App Store, Nightly and PR builds either.

Closes #1936

### DIFF
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -195,12 +195,6 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
         
         if let route = appRouteURLParser.route(from: url) {
             switch route {
-            case .oidcCallback(let url):
-                if stateMachine.state == .softLogout {
-                    softLogoutCoordinator?.handleOIDCRedirectURL(url)
-                } else {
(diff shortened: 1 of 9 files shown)


stefanceriu (approved):
Veery nice! 👏

PR #3466, opened 2024-10-30
Update verify identity button title.

pixlwave:
iOS equivalent to https://github.com/element-hq/element-x-android/pull/3751


manuroe (approved):
Thanks @pixlwave!

LGTM.
It makes me wonder: how do we manage orphan strings? `L10n.screenRoomMemberDetailsVerifyButtonTitle()` is not going to be used anymore and it is dead code.

pixlwave:
> how do we manage orphan strings?

We don't have anything more formal than every once in a while (I think we've done it once) we can run Periphery to find unused strings, Android run something similar on their side and then we compare notes and agree what can be removed.

Issue #3399, opened 2024-10-11
[Task] Adjust the "key storage out of sync" flow and copy


mxandreas:
Please see the details in the meta issue.

PR #3468, opened 2024-10-31
Update the strings for out of sync Key Storage.

pixlwave:
This PR is an initial part of #3399 to get the strings in sync with everything done for https://github.com/element-hq/element-meta/issues/2590. I've tweaked the banner a bit too to match the Figma, preparing it for the secondary button to be added.

Unfortunately it looks like updated flows will require somewhat of a re-write so this will be handled separately, likely after the next RC.

Screenshot 2024-10-31 at 10 15 49 am


manuroe (approved):
Thanks. It looks great to me.

PR #3486, opened 2024-11-05
Fix race condition when setting up session verification controller subscriptions

stefanceriu:
We were running the SessionVerificationController in a task after updating the session verification state, which would result in trying to subscribe to a nil object.


pixlwave (approved):
LGTM 🤞

PR #3487, opened 2024-11-05
Fix a race condition observing room info updates for calls.

pixlwave:
This PR is the iOS part for https://github.com/element-hq/element-meta/issues/2599. There were 2 issues at play:
- A race condition between getting the value of `hasRoomCall` and it being used in the publisher which could result in a room info update being dropped when it shouldn't be.
- An incoming call would launch the app if it had been killed and attempt to observe the room info update before a user session had been restored.

Note: I think there are still a couple of smaller issues with call ringing (possibly just matrix.org being slow as sometimes e.g. `Call notification is too old` would happen) and I wouldn't see CallKit appear at all or it would cancel almost immediately.

https://github.com/user-attachments/assets/13b7c6f2-43a0-4b15-9b8a-bc7687878140


stefanceriu (approved):
👏👏👏

PR #3488, opened 2024-11-06
Syncing fixes

stefanceriu:
* prevent background app refresh task expirations from stopping the main app's sync
* start a background task when receiving calls to receive state updates for longer

### DIFF
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -26,8 +26,6 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
 
     /// Common background task to continue long-running tasks in the background.
     private var backgroundTask: UIBackgroundTaskIdentifier?
-
-    private var isSuspended = false
     
     private var userSession: UserSessionProtocol? {
         didSet {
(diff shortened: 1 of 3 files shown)


pixlwave (approved):
LGTM 🙌

ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -150,6 +148,8 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
                 switch action {
                 case .startCall(let roomID):
                     self?.handleAppRoute(.call(roomID: roomID))
+                case .receivedIncomingCallRequest:
+                    self?.scheduleDelayedSyncStop()
```
pixlwave, line 152: Just discussed that we need to discuss the effect of this when the app is in the fg.

stefanceriu: Just checked, seems fine

```
Background Task 23 ("Called by ElementX.debug.dylib, from $s8ElementX14AppCoordinatorC23scheduleDelayedSyncStop33_E85D5AAE1EFD17BA4327A2E0B56B370DLLyyF"), was created over 30 seconds ago. In applications running in the background, this creates a risk of termination. Remember to call UIApplication.endBackgroundTask(_:) for your task in a timely manner to avoid this.
```

PR #3491, opened 2024-11-06
Start syncing when receiving a background VoIP call for the cases in which the app was suspended but not terminated

stefanceriu:


pixlwave (approved):
🤞🤞🤞

PR #3497, opened 2024-11-07
Fix incorrect analytics mapping for UTDs.

pixlwave:
As pointed out by @BillCarsonFr, https://github.com/element-hq/element-x-ios/pull/3442 messed up the mapping here.

Additionally upgrades the CI runners to macOS 15 given the failure in https://github.com/element-hq/element-x-ios/actions/runs/11727089995/job/3266732078, caused by https://github.com/actions/runner-images/issues/10703.

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   build:
     name: Danger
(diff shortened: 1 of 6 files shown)


stefanceriu (approved):
👍

PR #3498, opened 2024-11-08
Update compound.

pixlwave:
Changes the temporary tokens to be soft-deprecated to hide all of the warnings (as we haven't addressed them yet).


stefanceriu (approved):
🚀

PR #3501, opened 2024-11-11
Make stopSync more aware of background usage.

pixlwave:
We still had a bug where a successful background task was being run in the foreground. So this PR makes `stopSync` a bit more aware about whether it's being called from a backround task or not.

### DIFF
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -561,7 +561,7 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
         
         showLoadingIndicator()
         
-        stopSync()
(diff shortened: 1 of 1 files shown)


ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -930,7 +934,7 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
         backgroundTask = appMediator.beginBackgroundTask { [weak self] in
             guard let self else { return }
             
-            stopSync()
+            stopSync(isBackgroundTask: true)
```
pixlwave, line 937: I don't think this one is strictly needed, but doesn't hurt.

stefanceriu (approved):
Makes sense to me 👍

PR #3504, opened 2024-11-11
Stacked Avatars View

Velin92:
Since we need to use the stacked avatars from the timeline receipts again for the implementation of the knocking requests, I decided to create a specific view structure for it, that is also more customisable.


ElementX/Sources/Other/SwiftUI/Views/StackedAvatarsView.swift
```diff
@@ -0,0 +1,66 @@
...
+    var body: some View {
+        HStack(spacing: spacing) {
+            ForEach(0..<avatars.count, id: \.self) { index in
+                LoadableAvatarImage(url: avatars[index].url,
+                                    name: avatars[index].name,
+                                    contentID: avatars[index].contentID,
+                                    avatarSize: avatarSize,
+                                    mediaProvider: mediaProvider)
+                    .overlay {
+                        RoundedRectangle(cornerRadius: .infinity)
```
pixlwave, line 34: `Circle()` surely 😅

ElementX/Sources/Other/SwiftUI/Views/StackedAvatarsView.swift
```diff
@@ -0,0 +1,66 @@
...
+
+struct StackedAvatarInfo {
+    let url: URL?
+    let name: String?
+    let contentID: String
+}
+
+struct StackedAvatarsView: View {
+    /// Should be negative if we want to get the stacked behaviour
+    let spacing: CGFloat
```
pixlwave, line 18: You could call this `overlap` and then below do `spacing: -overlap`. Would probably be a slightly cleaner API.

ElementX/Sources/Other/SwiftUI/Views/StackedAvatarsView.swift
```diff
@@ -0,0 +1,66 @@
...
+    let lineWidth: CGFloat
+    var shouldStackFromLast = false
+    let avatars: [StackedAvatarInfo]
+    let avatarSize: AvatarSize
+    let mediaProvider: MediaProviderProtocol?
+    
+    var body: some View {
+        HStack(spacing: spacing) {
+            ForEach(0..<avatars.count, id: \.self) { index in
```
pixlwave, line 27: Probably better to make `StackedAvatarInfo` `Identifiable` and then do this so that if we ever animated them, SwiftUI would have a better idea about what was going on.

```suggestion
            ForEach(avatars) { avatar in
```

pixlwave (approved):
Nice 👌

ElementX/Sources/Other/SwiftUI/Views/StackedAvatarsView.swift
Velin92: the problem is that I need the index to apply the zStack change but, maybe I could add an .id(contentID) to each one of the elements to solve the issue?

PR #3510, opened 2024-11-12
Knock requests banner

Velin92:
This is just the UI implementation, still no data connection.


ElementX/Resources/Localizations/en.lproj/Localizable.strings
```diff
@@ -790,7 +794,7 @@
 "screen_room_timeline_add_reaction" = "Add emoji";
 "screen_room_timeline_beginning_of_room" = "This is the beginning of %1$@.";
 "screen_room_timeline_beginning_of_room_no_name" = "This is the beginning of this conversation.";
-"screen_room_timeline_legacy_call" = "Unsupported call. Ask if the caller can use to the new Element X app.";
+"screen_room_timeline_legacy_call" = "Unsupported call. Ask if the caller can use the new Element X app.";
```
pixlwave, line 797: I think there's a snapshot to update for this one.

ElementX/Sources/Other/AvatarSize.swift
```diff
@@ -47,6 +47,7 @@ enum UserAvatarSizeOnScreen {
     case suggestions
     case blockedUsers
     case knockingUsers
+    case knockingUser
```
pixlwave, line 50: Could we maybe disambiguate these a bit more. E.g.  `knockingUsersStack` or something.

ElementX/Sources/Screens/RoomScreen/View/KnockRequestsBannerView.swift
```diff
@@ -0,0 +1,165 @@
...
+            }
+            HStack(spacing: 12) {
+                Button(L10n.screenRoomSingleKnockRequestViewButtonTitle, action: {
+                    onViewAll()
+                })
+                .buttonStyle(.compound(.secondary))
+                Button(L10n.screenRoomSingleKnockRequestAcceptButtonTitle, action: {
+                    onViewAll()
+                })
+                .buttonStyle(.compound(.primary))
```
pixlwave, line 100: You can simplify this to pass the action directly
```suggestion
                Button(L10n.screenRoomSingleKnockRequestViewButtonTitle, action: onViewAll)
                    .buttonStyle(.compound(.secondary))
                Button(L10n.screenRoomSingleKnockRequestAcceptButtonTitle, action: onViewAll)
                    .buttonStyle(.compound(.primary))
```

ElementX/Sources/Screens/RoomScreen/View/KnockRequestsBannerView.swift
pixlwave, line 1: Do you think we could break this down a bit? I find the view pretty hard to follow. It would be easier if we had something like

```
struct KnockRequestsBannerView
private struct SingleKnockRequestBannerContent
private struct MultipleKnockRequestsBannerContent
```

That way, in the single knock request content you could break it up into simpler subviews like `avatar`, `header`, `message`, `buttons`.

pixlwave (commented):
Looks good, but would be nice if we could break this up a bit:

ElementX/Sources/Screens/RoomScreen/View/KnockRequestsBannerView.swift
pixlwave: Also, are both of these buttons meant to have the same action?

ElementX/Sources/Screens/RoomScreen/View/KnockRequestsBannerView.swift
Velin92: I would still keep them in the same struct, but i can maybe split each one into pieces

pixlwave (approved):
Thank you that's much easier to follow :)

PR #3512, opened 2024-11-13
Bump the RustSDK to v1.0.67

stefanceriu:


ElementX/SupportingFiles/Settings.bundle/Packages/matrix-rust-components-swift.plist
pixlwave, line 1: This shouldn't be committed.

ElementX/Sources/Services/Timeline/TimelineItemProxy.swift
```diff
@@ -179,6 +181,40 @@ struct TimelineItemDebugInfo: Identifiable, CustomStringConvertible {
     }
 }
 
+struct SendHandleProxy: Hashable {
+    enum Error: Swift.Error {
+        case sdkError(Swift.Error)
+    }
+    
+    let timelineID: TimelineItemIdentifier
```
pixlwave, line 189: For consistency I think we should call this `itemID` tbf.

PR #3513, opened 2024-11-13
Fix Rooms that user has knocked not displaying the request sent screen

Velin92:


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -933,7 +933,8 @@ class ClientProxy: ClientProxyProtocol {
             case .knocked:
                 if appSettings.knockingEnabled {
                     return try await .knocked(KnockedRoomProxy(roomListItem: roomListItem,
-                                                               room: roomListItem.invitedRoom()))
+                                                               roomPreview: roomListItem.previewRoom(via: []),
```
pixlwave, line 936: We discussed this at length and need to find out a bit more from the SDK team. Our `JoinRoomScreen` already calls `clientProxy.roomPreviewForIdentifier` which uses the same underlying `get_room_preview`, but with the `via` passed in from its parameters.

Theoretically we don't need the `via` here as the room is already in the room list, but given `invitedRoom()` is now [marked as deprecated](https://github.com/matrix-org/matrix-rust-sdk/blob/aca83fb4ed7447effd1fe5eb2f9f0cd87bb86c18/bindings/matrix-sdk-ffi/src/room_list.rs#L584) I think we're probably doing something wrong in our flow.

Velin92: Even if `invitedRoom()` is deprecated the `previewRoom` API is still not able to give us the same amount of informations and APIs to which handle a room to which you are invited.
@jmartinesp maybe this is something that has been scheduled for future improvements of such API?

Velin92: okay so @jmartinesp is working on an improvement to allow the use of `previewRoom` also for the invited room, which we can include later, the only thing left to decide is to wether include the `via` or not

Velin92: Okay in theory we don't need it because we always wait for the sync at this step, so the room is already in the list and on the homeserver

ElementX/Sources/Services/Room/KnockedRoomProxy.swift
```diff
@@ -11,26 +11,29 @@ import UIKit
...
-    private let room: RoomProtocol
+    private let roomPreview: RoomPreviewProtocol
+    let info: BaseRoomInfoProxyProtocol
     
     // A room identifier is constant and lazy stops it from being fetched
     // multiple times over FFI
-    lazy var id: String = room.id()
-    
-    var ownUserID: String { room.ownUserId() }
+    var id: String {
```
pixlwave, line 19: The comment still says this is lazy. Is it needed for handling the room list diffs and avoiding the FFI?

pixlwave (approved):
LGTM

ElementX/Sources/Services/Room/KnockedRoomProxy.swift
Velin92: oh yeah probably we still need that

PR #3515, opened 2024-11-14
UI test fixes

stefanceriu:


pixlwave (approved):
I haven't looked at all the new snapshots, but we needed to regenerate a bunch anyway as our thresholds are lower on UI tests so sounds good to me!

PR #3530, opened 2024-11-18
using `roomPreview` API for invited rooms

Velin92:
In this way we can get rid of the deprecated `invitedRoom` API


pixlwave (approved):
Nice 😎

PR #3533, opened 2024-11-18
Knock Requests List Screen

Velin92:
Thi adds the UI for the knock requests list screen and some other small improvements, to the existing UI, however this is jsut UI work, the business logic is not ready yet.


ElementX/Sources/Screens/KnockRequestsListScreen/KnockRequestsListScreenViewModel.swift
```diff
@@ -0,0 +1,75 @@
...
+    
+    override func process(viewAction: KnockRequestsListScreenViewAction) {
+        switch viewAction {
+        case .acceptAllRequests:
+            break
+        case .acceptRequest(userID: let userID):
+            break
+        case .declineRequest(userID: let userID):
+            break
+        case .ban(userID: let userID):
```
pixlwave, line 42: FWIW, we tend to drop the label everywhere else when we `case let` something
```suggestion
        case .acceptRequest(userID: let userID):
            break
        case .declineRequest(userID: let userID):
            break
        case .ban(userID: let userID):
```

ElementX/Sources/Screens/KnockRequestsListScreen/View/KnockRequestCell.swift
```diff
@@ -15,8 +15,8 @@
 import Compound
 import SwiftUI
 
-struct KnockRequestCellInfo {
-    let userID: String
+struct KnockRequestCellInfo: Identifiable {
+    let id: String
```
pixlwave, line 19: Probably worth a doc comment that this is the user ID at least now.

ElementX/Sources/Screens/KnockRequestsListScreen/View/KnockRequestCell.swift
```diff
@@ -26,67 +26,85 @@ struct KnockRequestCellInfo {
...
-                                .font(.compound.bodyMD)
+        HStack(alignment: .top, spacing: 16) {
+            LoadableAvatarImage(url: cellInfo.avatarUrl,
+                                name: cellInfo.displayName,
+                                contentID: cellInfo.id,
+                                avatarSize: .user(on: .knockingUserList),
+                                mediaProvider: mediaProvider)
+            VStack(alignment: .leading, spacing: 12) {
+                VStack(alignment: .leading, spacing: 0) {
+                    HStack(alignment: .top, spacing: 0) {
```
pixlwave, line 42: Can we break this up into a property or 2 perhaps? It's much easier to reason about the layout if you can see "this part is the header" or whatnot.

ElementX/Sources/Screens/KnockRequestsListScreen/View/KnockRequestCell.swift
```diff
@@ -26,67 +26,85 @@ struct KnockRequestCellInfo {
...
+                }
+                if let reason = cellInfo.reason {
+                    DisclosableText(text: reason)
+                }
+                actions
+            }
+            .padding(.trailing, 16)
+            .overlay(alignment: .bottom) {
+                VStack(spacing: 0) {
+                    Divider()
```
pixlwave, line 67: Technically we should be colouring this from Compound. Gaelle is reviewing these right now though as we're out of sync with Figma.

`.overlay { Color.compound._borderInteractiveSecondaryAlpha }` gets you part of the way there but it has an alpha value so probably not correct anyway.

PreviewTests/Sources/__Snapshots__/PreviewTests/test_knockRequestCell-iPad-en-GB.No-Ban.png
pixlwave, line 1: Ok the divider layout looks wrong to me here. You probably need to add the padding to that whole stack, rather than the bottom button.

ElementX/Sources/Screens/KnockRequestsListScreen/View/KnockRequestsListScreen.swift
```diff
@@ -0,0 +1,96 @@
...
+                }
+            }
+            .padding(.top, 40)
+        }
+        .background(.compound.bgCanvasDefault)
+        .overlay {
+            if context.viewState.requests.isEmpty {
+                KnockRequestsListEmptyStateView()
+            }
+        }
```
pixlwave, line 41: Personally I think it probably makes more sense to put these on the mainContent in the body than on the scroll view directly.

ElementX/Sources/Screens/KnockRequestsListScreen/View/KnockRequestCell.swift
```diff
@@ -130,9 +148,10 @@ private struct DisclosableText: View {
                 }
             } label: {
                 CompoundIcon(\.chevronDown, size: .medium, relativeTo: .compound.bodyMD)
+                    .foregroundStyle(.compound.iconTertiary)
+                    .rotationEffect(.degrees(isExpanded ? 180 : 0))
```
pixlwave, line 152: Worth saying, iOS (and everywhere we're doing disclosures - grouping state events and viewing timeline source) uses a right pointing chevron for collapsed rather than an upwards point one. Would be nice if this design could match those.

pixlwave (approved):
Nice! 🙌

ElementX/Sources/Screens/KnockRequestsListScreen/View/KnockRequestCell.swift
Velin92: Oh interesting worth mentioning this to Gaelle

PR #3534, opened 2024-11-19
Show both defaults and frequent emojis in the timeline item menu, make the list scrollable

stefanceriu:


ElementX/Sources/Screens/Timeline/View/ItemMenu/TimelineItemMenuAction.swift
```diff
@@ -25,30 +26,32 @@ struct TimelineItemMenuActions {
...
     }
 }
 
-struct TimelineItemMenuReaction {
+struct TimelineItemMenuReaction: Hashable {
     let key: String
     let symbol: SFSymbol
+    
+    // Frequently used emojis on the all use the same .heart SFSymbol.
+    // Override equatable so we can remove duplicates.
```
pixlwave, line 51: 👌

pixlwave (approved):
Nice! Might be some snapshots to update though :)

PR #3539, opened 2024-11-20
Link the MatrixRustSDK dynamically and only embed it in the main target

stefanceriu:
This will stop copying the framework to extension targets and reduce the overall app size to ~= 49.6 MB compressed, 127 MB uncompressed


pixlwave (approved):
Changes look sane to me but CI isn't happy about something 😕

PR #3540, opened 2024-11-20
Fix UI tests, update compound to roll back snapshot testing and avoid the requirement for Swift Testing.

stefanceriu:


pixlwave (approved):
See you later weird dependencies.

PR #3541, opened 2024-11-20
Handle NSItemProvider public.image data types.

stefanceriu:


ElementX/Sources/Other/Extensions/NSItemProvider.swift
```diff
@@ -6,23 +6,134 @@
...
+        
+        switch contentType.identifier {
+        case UTType.image.identifier:
+            return "png"
+        default:
+            return nil
+        }
+    }
+    
+    private func isIdentifierSupported(_ identifier: String?) -> Bool {
```
pixlwave, line 135: Would be nice to add a comment here saying something like

> Don't filter out generic `public.image` content as screenshots are in this format and we can convert it to a PNG ourselves.

ElementX/Sources/Other/Extensions/NSItemProvider.swift
```diff
@@ -6,23 +6,134 @@
...
+            }
+        }
+        
+        guard let shareData else {
+            MXLog.error("Invalid share data")
+            return nil
+        }
+        
+        do {
+            if let suggestedName = (suggestedName as? NSString),
```
pixlwave, line 83:
```suggestion
            if let suggestedName = suggestedName as? NSString,
```

ElementX/Sources/Other/Extensions/NSItemProvider.swift
```diff
@@ -6,23 +6,134 @@
...
+            return nil
+        }
+        
+        guard let pngData = uiImage.pngData() else {
+            MXLog.error("Failed extracting PNG data out of the UIImage")
+            return nil
+        }
+        
+        do {
+            if let suggestedName = (suggestedName as? NSString),
```
pixlwave, line 43:
```suggestion
            if let suggestedName = suggestedName as? NSString,
```

ElementX/Sources/Other/Extensions/NSItemProvider.swift
```diff
@@ -6,23 +6,134 @@
...
+        
+        guard let shareData else {
+            MXLog.error("Invalid share data")
+            return nil
+        }
+        
+        do {
+            if let suggestedName = (suggestedName as? NSString),
+               // Suggestions are nice but their extension is `jpeg`
+               let filename = (suggestedName.deletingPathExtension as NSString).appendingPathExtension(contentType.fileExtension) {
```
pixlwave, line 85: Double checking here as I thought the old logic was working fine outside of screenshots: Isn't this useless jpeg suggestion only for public.image?

pixlwave (approved):
Nice 👌

ElementX/Sources/Other/Extensions/NSItemProvider.swift
stefanceriu: Sure is, sorry, copy pasta 🤦‍♂️

PR #3542, opened 2024-11-20
Enable inline replies for push notifications.

stefanceriu:
They've been disabled for a while now and it's time for them to shine.


manuroe:
To double check: which process is used to send the reply?

pixlwave (approved):
Let's goooo!

PR #3544, opened 2024-11-21
Delay handling inline notification replies until the user session is established

stefanceriu:


ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -313,40 +314,27 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
             if let userSession {
                 userSession.clientProxy.roomsToAwait.insert(roomID)
             } else {
-                storedRoomsToAwait.insert(roomID)
+                storedRoomsToAwait = [roomID]
```
pixlwave, line 317: Why is this a collection of rooms, and how come it is ok to wipe out the remaining contents of it?

ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -474,6 +462,24 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
...
+        }
+        
+        if let storedAppRoute {
+            userSessionFlowCoordinator.handleAppRoute(storedAppRoute, animated: false)
+        }
+        
+        if let storedInlineReply {
+            await processInlineReply(roomID: storedInlineReply.roomID, replyText: storedInlineReply.message)
+        }
+    }
```
pixlwave, line 482: Would it make sense to set these all back to nil afterwards?

pixlwave (approved):
Good spot 👍

ElementX/Sources/Application/AppCoordinator.swift
stefanceriu: Isn't that what `@Consumable` is supposed to do? 🤔

ElementX/Sources/Application/AppCoordinator.swift
stefanceriu: I imagine it's a collection just for flexibility but I don't see how we can run multiple `notificationTapped` before consuming it 🤔

ElementX/Sources/Application/AppCoordinator.swift
pixlwave: Oh yep, I even looked at that and thought "we should replace this with movable types". 🤦‍♂️

Too early, *puts the kettle on*

PR #3546, opened 2024-11-21
Add support for sharing URLs and text.

pixlwave:

- Add URL/Text as supported types in the share extension
- Fall back to loading a URL and then a string in the share extension
- Send both of these into the composer for the user to confirm before sending.
- If sharing starts a new flow we pass the initial text in the init, otherwise it is set into the existing composer.

https://github.com/user-attachments/assets/cb956f5d-bf4e-4262-9965-c98ad7b80fc7


stefanceriu (approved):
Woohoo, I love how clean this turned out!

PR #3547, opened 2024-11-21
Add support for adding/editing/removing media captions.

pixlwave:
Does what it says on the tin. Might add some more tests to this before merging, but we don't have anything around the `TimelineItemMenuActionProvider` right now.

https://github.com/user-attachments/assets/225f4282-7f0c-4c4f-8870-1e728410578a


ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
```diff
@@ -606,11 +606,17 @@ class TimelineViewModel: TimelineViewModelType, TimelineViewModelProtocol {
                                                  html: html,
                                                  inReplyToEventID: eventID,
                                                  intentionalMentions: intentionalMentions)
-        case .edit(let originalEventOrTransactionID):
+        case .edit(let originalEventOrTransactionID, type: .default):
```
Velin92, line 609: you can remove `type` here

Velin92 (approved):
LGTM

pixlwave:
> Might add some more tests to this before merging, but we don't have anything around the TimelineItemMenuActionProvider right now.

Change of plan: Going to add captions to file items and support copying captions and then I'll UI test it all.

PR #3553, opened 2024-11-25
Move timeline item tap gestures to the items themselves instead of the bubbled styler

stefanceriu:
While not much came out of the `UIGestureRecognizerRepresentable` prototypes we can at least move the normal tap to within the cells themselves. This will for example fix images opening in full screen when tapping on the caption text.

### DIFF
--- a/ElementX/Sources/Screens/Timeline/View/Style/TimelineItemBubbledStylerView.swift
+++ b/ElementX/Sources/Screens/Timeline/View/Style/TimelineItemBubbledStylerView.swift
@@ -391,6 +391,7 @@ struct TimelineItemBubbledStylerView_Previews: PreviewProvider, TestablePreview 
             }
         }
         .environmentObject(viewModel.context)
+        .environment(\.timelineContext, viewModel.context)
     }
     
     static var replies: some View {
@@ -422,6 +423,7 @@ struct TimelineItemBubbledStylerView_Previews: PreviewProvider, TestablePreview 
                                                   groupStyle: .single))
         }
         .environmentObject(viewModel.context)
+        .environment(\.timelineContext, viewModel.context)
     }
     
     static var threads: some View {
@@ -429,6 +431,7 @@ struct TimelineItemBubbledStylerView_Previews: PreviewProvider, TestablePreview 
             MockTimelineContent(isThreaded: true)
         }
         .environmentObject(viewModel.context)
+        .environment(\.timelineContext, viewModel.context)
     }
       
     static var pinned: some View {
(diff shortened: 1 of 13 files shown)


ElementX/Sources/Screens/Timeline/View/Style/TimelineItemBubbledStylerView.swift
```diff
@@ -124,11 +124,6 @@ struct TimelineItemBubbledStylerView<Content: View>: View {
     
     var messageBubbleWithActions: some View {
         messageBubble
-            .onTapGesture {
-                context.send(viewAction: .itemTapped(itemID: timelineItem.id))
-            }
-            // We need a tap gesture before this long one so that it doesn't
-            // steal away the gestures from the scroll view
```
stefanceriu, line 131: Tested this on both iOS 18 and 17 (simulator) and it doesn't seem to be needed anymore.

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/AudioRoomTimelineView.swift
```diff
@@ -9,6 +9,7 @@ import Foundation
 import SwiftUI
 
 struct AudioRoomTimelineView: View {
+    @EnvironmentObject private var context: TimelineViewModel.Context
```
pixlwave, line 12: Should probably use the non-observed one for all of these additions, seeing as we're only using it to send actions and don't need to react to it.
```suggestion
    @Environment(\.timelineContext) private var context
```

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/AudioRoomTimelineView.swift
```diff
@@ -22,6 +23,9 @@ struct AudioRoomTimelineView: View {
                   .padding(.vertical, 12)
                   .padding(.horizontal, 6)
                   .accessibilityLabel(L10n.commonAudio)
+                  .onTapGesture {
+                      context.send(viewAction: .itemTapped(itemID: timelineItem.id))
```
pixlwave, line 27: Also, completely optional, but maybe it makes sense to rename this action to `.mediaTapped` so its clearer where to use it and what will happen when you send it.

pixlwave (approved):
Thank you 👍

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/AudioRoomTimelineView.swift
stefanceriu: Very good point! 👍

pixlwave: One day we'll switch to Observable and everything will be beautiful 😄

PR #3555, opened 2024-11-25
Knock Requests navigation flows

Velin92:
implemented navigations flow for the actions inside the knock requests

### DIFF
--- a/ElementX/Sources/Screens/KnockRequestsListScreen/KnockRequestsListScreenModels.swift
+++ b/ElementX/Sources/Screens/KnockRequestsListScreen/KnockRequestsListScreenModels.swift
@@ -11,7 +11,7 @@ enum KnockRequestsListScreenViewModelAction { }
 
 struct KnockRequestsListScreenViewState: BindableState {
     // TODO: Not sure yet how we will fetch this, this is just for testing purposes
-    var requests: [KnockRequestCellInfo] = [.init(id: "@alice:matrix.org", displayName: "Alice", avatarUrl: nil, timestamp: "Now", reason: "Hello")]
+    var requests: [KnockRequestCellInfo] = [.init(id: "@alice:matrix.org", displayName: "Alice", avatarURL: nil, timestamp: "Now", reason: "Hello")]
     // If you are in this view one of these must have been true so by default we assume all of them to be true
     var canAccept = true
     var canDecline = true
--- a/ElementX/Sources/Screens/KnockRequestsListScreen/KnockRequestsListScreenModels.swift
+++ b/ElementX/Sources/Screens/KnockRequestsListScreen/KnockRequestsListScreenModels.swift
@@ -7,10 +7,13 @@
 
 import Foundation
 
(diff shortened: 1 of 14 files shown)


ElementX/Sources/Screens/KnockRequestsListScreen/KnockRequestsListScreenModels.swift
```diff
@@ -7,10 +7,13 @@
...
 
-enum KnockRequestsListScreenViewModelAction { }
+enum KnockRequestsListScreenViewModelAction {
+    case displayBannedUsers
+}
 
 struct KnockRequestsListScreenViewState: BindableState {
-    var requests: [KnockRequestCellInfo] = []
+    // TODO: Not sure yet how we will fetch this, this is just for testing purposes
+    var requests: [KnockRequestCellInfo] = [.init(id: "@alice:matrix.org", displayName: "Alice", avatarUrl: nil, timestamp: "Now", reason: "Hello")]
```
pixlwave, line 16:
```suggestion
    var requests: [KnockRequestCellInfo] = [.init(id: "@alice:matrix.org", displayName: "Alice", avatarURL: nil, timestamp: "Now", reason: "Hello")]
```

ElementX/Sources/Screens/KnockRequestsListScreen/KnockRequestsListScreenViewModel.swift
pixlwave, line 1: Maybe a dumb question, but could you add all of these actions now in the room proxy with empty implementations until the Rust side is ready? IIRC, it would also mean you could hook up the actions on the banner too?

ElementX/Sources/Screens/RoomMemberListScreen/RoomMembersListScreenCoordinator.swift
```diff
@@ -13,6 +13,7 @@ struct RoomMembersListScreenCoordinatorParameters {
     let roomProxy: JoinedRoomProxyProtocol
     let userIndicatorController: UserIndicatorControllerProtocol
     let analytics: AnalyticsService
+    let showBanned: Bool
```
pixlwave, line 16: Tiny nit: Could we put this first as it controls how the screen will appear when presented?

Also would be nice to put the room proxy second, then the order would match the view model init too :)

pixlwave (approved):
✊✊✊

ElementX/Sources/Screens/RoomMemberListScreen/RoomMembersListScreenCoordinator.swift
Velin92: no worries this has been removed

ElementX/Sources/Screens/KnockRequestsListScreen/KnockRequestsListScreenViewModel.swift
pixlwave: Fair enough :)

PR #3559, opened 2024-11-26
Stop delaying ElementCall until the next sync loop and only notify other participants when presumed to already be up to date.

stefanceriu:
We implementing ringing we observed that sometimes notifications were being sent multiple times, causing multiple incoming calls to appear. The reason for that was that the app wasn't necessarily up to date and the call member list wes still unknown. 

The fix was to delay the starting of the call until the next sync loop but that in turn made entering an ElementCall slow. We are now reverting that and only, intentionally, sending a notification if the user presses the button and a call is not running yet. We're assuming that the app is already up to date at that point and that the call state is correctly known. 

Relates to element-hq/voip-internal/issues/259

### DIFF
--- a/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
@@ -201,7 +201,7 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
         case .userProfile(let userID):
             stateMachine.processEvent(.showUserProfileScreen(userID: userID), userInfo: .init(animated: animated))
         case .call(let roomID):
-            Task { await presentCallScreen(roomID: roomID) }
+            Task { await presentCallScreen(roomID: roomID, notifyOtherParticipants: false) }
         case .genericCallLink(let url):
             presentCallScreen(genericCallLink: url)
         case .settings, .chatBackupSettings:
@@ -579,7 +579,8 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
             
             switch action {
             case .presentCallScreen(let roomProxy):
-                presentCallScreen(roomProxy: roomProxy)
+                // Here we assume that the app is running and the call state is already up to date
+                presentCallScreen(roomProxy: roomProxy, notifyOtherParticipants: !roomProxy.infoPublisher.value.hasRoomCall)
             case .finished:
                 stateMachine.processEvent(.deselectRoom)
             }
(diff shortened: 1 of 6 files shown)


ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -579,7 +579,8 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
             
             switch action {
             case .presentCallScreen(let roomProxy):
-                presentCallScreen(roomProxy: roomProxy)
+                // Here we assume that the app is running and the call state is already up to date
+                presentCallScreen(roomProxy: roomProxy, notifyOtherParticipants: !roomProxy.infoPublisher.value.hasRoomCall)
```
pixlwave, line 583: FWIW, I think it would signal the intent more clearly if `notifyOtherParticipants` was passed up in the action, and  it's was derived from `context.viewState.hasOngoingCall` in the room screen. WDYT?

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -654,29 +655,30 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
...
+    private func presentCallScreen(roomProxy: JoinedRoomProxyProtocol, notifyOtherParticipants: Bool) {
         let colorScheme: ColorScheme = appMediator.windowManager.mainWindow.traitCollection.userInterfaceStyle == .light ? .light : .dark
         presentCallScreen(configuration: .init(roomProxy: roomProxy,
                                                clientProxy: userSession.clientProxy,
                                                clientID: InfoPlistReader.main.bundleIdentifier,
                                                elementCallBaseURL: appSettings.elementCallBaseURL,
                                                elementCallBaseURLOverride: appSettings.elementCallBaseURLOverride,
-                                               colorScheme: colorScheme))
+                                               colorScheme: colorScheme),
+                          notifyOtherParticipants: notifyOtherParticipants)
```
pixlwave, line 677: Should we add this to the configuration rather than passing it in separately?

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
```diff
@@ -132,52 +131,44 @@ class CallScreenViewModel: CallScreenViewModelType, CallScreenViewModelProtocol
     
     // MARK: - Private
     
-    private func setupCall() {
+    private func setupCall(notifyOtherParticipants: Bool) {
```
pixlwave, line 134: Ok yeah answering my own question, I would say so given we access the configuration in this method.

pixlwave (approved):
Makes sense to me 📲

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
stefanceriu: 🤦‍♂️ completely missed and forgot about that configuration 🤦‍♂️

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
stefanceriu: Might as well yeah 👍

stefanceriu: Okay no, I changed my mind. This would need to be passed through from the room, the room details and the user profile screens. I don't think it's worth it. The `UserSessionFlowCoordinator` is perfectly capable of figuring out what's right here.

PR #3560, opened 2024-11-26
Move the media caption composer (and Add Caption action) behind a feature flag for now.

pixlwave:
We need to implement media caption rendering in the Legacy apps before we can enable this by default.


stefanceriu (approved):
Makes sense 👌

PR #3561, opened 2024-11-27
Add back missing send button when media captions are disabled.

pixlwave:
| FF off | FF on |
| - | - |
| Simulator Screenshot - iPhone 16 Pro - 2024-11-27 at 11 14 17 | Simulator Screenshot - iPhone 16 Pro - 2024-11-27 at 11 14 43 |


stefanceriu (approved):
🙈

PR #3570, opened 2024-11-29
Knock a room - added a char counter for the message

Velin92:


stefanceriu (approved):
Just a comment on placement, otherwise lgtm 👍

PR #3574, opened 2024-11-29
Add a warning to the media caption composer.

pixlwave:
This PR
- Adds the warning to the media caption composer (there's a build setting for forks that don't need this).
- Removes the feature flag now the user is made aware that not everyone can read the caption.

https://github.com/user-attachments/assets/87a7e953-fca3-436d-8ee5-65af3974feb5


stefanceriu (approved):
Nice! lgtm 👍

PR #3578, opened 2024-12-03
Update SDK 1.0.77

Velin92:


stefanceriu (approved):
👌

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemFactory.swift
```diff
@@ -153,6 +153,9 @@ struct RoomTimelineItemFactory: RoomTimelineItemFactoryProtocol {
             case .sentBeforeWeJoined:
                 encryptionType = .megolmV1AesSha2(sessionID: sessionID, cause: .sentBeforeWeJoined)
                 errorLabel = L10n.commonUnableToDecryptNoAccess
+            case .historicalMessage:
+                encryptionType = .megolmV1AesSha2(sessionID: sessionID, cause: .unknown)
+                errorLabel = L10n.commonUnableToDecryptNoAccess
```
Velin92, line 158: @stefanceriu we don't have a cause for this, is it okay to use unknown and the unable to decrypt no access message?

stefanceriu: Yes, and I think you can even use `commonUnableToDecrypt` instead of `commonUnableToDecryptNoAccess`.

Velin92: okay made a new case to handle it

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemFactory.swift
```diff
@@ -153,6 +153,9 @@ struct RoomTimelineItemFactory: RoomTimelineItemFactoryProtocol {
             case .sentBeforeWeJoined:
                 encryptionType = .megolmV1AesSha2(sessionID: sessionID, cause: .sentBeforeWeJoined)
                 errorLabel = L10n.commonUnableToDecryptNoAccess
+            case .historicalMessage:
+                encryptionType = .megolmV1AesSha2(sessionID: sessionID, cause: .historicalMessage)
+                errorLabel = L10n.commonUnableToDecryptNoAccess
```
BillCarsonFr, line 158: There is a new string for that `timeline.decryption_failure.historical_event_no_key_backup`

PR #3581, opened 2024-12-03
Add new development assets and use in mocks/previews.

pixlwave:
Follow on from #3569, this PR adds some specific assets for images, videos and avatars so that we're not building against the logo which is a circle with alpha.

Lots of snapshots, I tweaked a few so they render the full height rather than being clipped.


Velin92 (approved):
LGTM, no we have some diversity in testing images in previews

PR #3585, opened 2024-12-04
Fix UI snapshots.

pixlwave:
#3581 changed the images we use in mocks but forgot to update the UI tests so this PR fixes that.

- The first 2 commits remove snapshots from devices that don't match the names in the Fastfile. The tests will now crash when using the wrong device.
- The next commit regenerates the snapshots. However, with our precision of 0.99, only 2 snapshots were deemed failures, therefore…
- The final commit regenerates the snapshots with a precision set to 1.0. I didn't commit the change to precision though, only used it to trigger the update. It basically includes all the avatars, and some changes to text colours and the composer border colour.

UI tests run: https://github.com/element-hq/element-x-ios/actions/runs/12157954025


UITests/Sources/Application.swift
```diff
@@ -118,6 +113,20 @@ extension XCUIApplication {
     }
 }
 
+private extension UIDevice {
+    var snapshotName: String {
```
stefanceriu, line 117: 👌

stefanceriu (approved):
👏

PR #3588, opened 2024-12-05
Media gallery

stefanceriu:
This PR contains the initial work done for the new Media Gallery feature which is very much still a work in progress. We've decided to merge it early under a feature flag in order to avoid later conflicts.

This screen is following a novel architecture where 2 different `TimelineViewModel`s are composed together under the `MediaEventsTimelineScreenViewModel` to provide the 2 separate sections. They're powered by a new RustSDK API that allows filtering room messages based on their type (introduced in [4365](https://github.com/matrix-org/matrix-rust-sdk/pull/4365))

We're doing this in order to reuse most of the timeline logic we already had in play for the room screen and pinned events timeline. The UI on the other hand is completely custom and only makes use of the `TimelineViewState`s and related models.

I haven't bothered with cleaning up the commits as I intend to squash all of it on merge and they still might prove useful to see the evolution.

What's not working:
* only images and videos are supported, no separators, files, voice messages etc.
* gifs don't animate (yet)
* separators (even though not rendered) are still computed per day
* there are no empty states
* media cannot be enlarged once rendered in the grid
* and most likely quite a few other things

https://github.com/user-attachments/assets/18b225f3-08a0-4aaf-8cb0-2960e56f6df4


ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -50,11 +50,6 @@ struct FocusEvent: Hashable {
     let shouldSetPin: Bool
 }
 
-private enum PinnedEventsTimelineSource: Hashable {
-    case room
-    case details(isRoot: Bool)
-}
```
Velin92, line 56: was recently talking with Doug how we should refactor this to use the previousState instead of this, nice!

ElementX/Sources/FlowCoordinators/MediaEventsTimelineFlowCoordinator.swift
```diff
@@ -0,0 +1,97 @@
...
+//
+
+import Combine
+import Foundation
+
+enum MediaEventsTimelineFlowCoordinatorAction {
+    case finished
+}
+
+class MediaEventsTimelineFlowCoordinator: FlowCoordinatorProtocol {
```
Velin92, line 15: do we need other coordinators to be included here in the future?

stefanceriu: The intial plan was to push the QuickLook screen coordinator from here but we're not so sure anymore, we might end up presenting it directly from the root screen. Either way, an extra flow coordinator can't hurt 😁

ElementX/Sources/Screens/MediaEventsTimelineScreen/MediaEventsTimelineScreenCoordinator.swift
```diff
@@ -0,0 +1,74 @@
...
+    let appMediator: AppMediatorProtocol
+    let emojiProvider: EmojiProviderProtocol
+}
+
+enum MediaEventsTimelineScreenCoordinatorAction {
+    case dismiss
+    case displayUser(userID: String)
+    case presentLocationViewer(geoURI: GeoURI, description: String?)
+    case displayMessageForwarding(forwardingItem: MessageForwardingItem)
+    case displayRoomScreenWithFocussedPin(eventID: String)
```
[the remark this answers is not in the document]
stefanceriu: Ah no, these are just left over from the pinned events coordinator, I should probably just delete them.

ElementX/Sources/Screens/MediaEventsTimelineScreen/View/MediaEventsTimelineScreen.swift
```diff
@@ -0,0 +1,163 @@
...
+                            }
+                            .clipped()
+                            .scaleEffect(.init(width: 1, height: -1))
+                    }
+                }
+                
+                // Needs to be wrapped in a LazyStack otherwise appearance calls don't trigger
+                LazyVStack(spacing: 0) {
+                    Rectangle()
+                        .frame(height: 44)
```
Velin92, line 57: what if there is no more pagination required?  Shouldn't this frame become 0? Or we always keep 44 points of black space from the top?

ElementX/Sources/Screens/MediaEventsTimelineScreen/MediaEventsTimelineScreenViewModel.swift
```diff
@@ -0,0 +1,103 @@
...
+//
+
+import Combine
+import SwiftUI
+
+typealias MediaEventsTimelineScreenViewModelType = StateStoreViewModel<MediaEventsTimelineScreenViewState, MediaEventsTimelineScreenViewAction>
+
+class MediaEventsTimelineScreenViewModel: MediaEventsTimelineScreenViewModelType, MediaEventsTimelineScreenViewModelProtocol {
+    private let imageAndVideoTimelineViewModel: TimelineViewModelProtocol
+    private let fileAndAudioTimelineViewModel: TimelineViewModelProtocol
```
Velin92, line 15: My only concern is how mock these to write Unit tests for them? Probably is just as easy as generating an auto mock for the TimelineViewModelProtocol? Is worth giving it a try to see if this is a viable path also for testing purposes.

ElementX/Sources/Screens/MediaEventsTimelineScreen/MediaEventsTimelineScreenViewModel.swift
```diff
@@ -0,0 +1,102 @@
...
+    
+    private var isTopVisible = false
```
pixlwave, line 17: Now that we're flipped, could we do the same as `TimelineTableViewController` and go with `isOldestItemVisible` here (and same for the actions)?

ElementX/Sources/Screens/MediaEventsTimelineScreen/MediaEventsTimelineScreenViewModel.swift
```diff
@@ -0,0 +1,102 @@
...
+            default:
+                false
+            }
+        }.reversed()
+        
+        state.isBackPaginating = (timelineViewState.timelineState.paginationState.backward == .paginating)
+        backpaginateIfNecessary(paginationStatus: timelineViewState.timelineState.paginationState.backward)
+    }
+    
+    private func backpaginateIfNecessary(paginationStatus: PaginationStatus) {
```
pixlwave, line 97:
```suggestion
    private func backPaginateIfNecessary(paginationStatus: PaginationStatus) {
```

ElementX/Sources/Screens/MediaEventsTimelineScreen/View/MediaEventsTimelineScreen.swift
```diff
@@ -0,0 +1,175 @@
...
+    
+    var body: some View {
+        content
+            .navigationBarTitleDisplayMode(.inline)
+            .background(.compound.bgCanvasDefault)
+            // Doesn't play well with the transformed scrollView
+            .toolbarBackground(.visible, for: .navigationBar)
+            .toolbar {
+                ToolbarItem(placement: .principal) {
+                    Picker("", selection: $context.screenMode) {
```
pixlwave, line 22: We might want a label here for accessibility. Or it might be fine with just the labels in the segments, but either way, its worth checking what VoiceOver does here.

ElementX/Sources/Screens/MediaEventsTimelineScreen/View/MediaEventsTimelineScreen.swift
```diff
@@ -0,0 +1,175 @@
...
+                            .tag(MediaEventsTimelineScreenMode.imageAndVideo)
+                        Text(L10n.screenMediaBrowserListModeFiles)
+                            .padding()
+                            .tag(MediaEventsTimelineScreenMode.fileAndAudio)
+                    }
+                    .pickerStyle(.segmented)
+                }
+            }
+            .colorScheme(.dark)
+            .toolbarColorScheme(.dark)
```
pixlwave, line 34: Side note: There's a message in the room that we don't need to stay dark for these now.

ElementX/Sources/Screens/MediaEventsTimelineScreen/View/MediaEventsTimelineScreen.swift
```diff
@@ -0,0 +1,175 @@
...
+                }
+            }
+        }
+        .scaleEffect(.init(width: 1, height: -1))
+        .onChange(of: context.screenMode) { _, _ in
+            context.send(viewAction: .changedScreenMode)
+        }
+    }
+    
+    @ViewBuilder func viewForTimelineItem(_ item: RoomTimelineItemViewState) -> some View {
```
pixlwave, line 79: Probably nice to pull this one out into its own struct tbf.

pixlwave:
Oh yeah, we should probably figure out if that ordering makes sense too, but that doesn't need to block this PR.

ElementX/Sources/Screens/MediaEventsTimelineScreen/View/MediaEventsTimelineScreen.swift
stefanceriu: Oh yeah, I'm going to create separate files for every single timeline view, similar to the normal timeline.

ElementX/Sources/Screens/MediaEventsTimelineScreen/View/MediaEventsTimelineScreen.swift
stefanceriu: They get read as `Media, selected, tab` and `Files, unselected, tab`, I think that's okay.

Velin92 (approved):

PR #3603, opened 2024-12-11
Fix tap knockable room and custom keyboard did show scrolling

Velin92:
Fixes a bug where the knockable rooms where not recognised as joinable/tappable

Also adds an API to allow the customisation of the scrolling when the keyboard appears


pixlwave (approved):
👌

PR #3604, opened 2024-12-11
Configure the media preview screen based on the event and presentation.

pixlwave:
This PR uses a timeline item action provider to determine which actions will be included for a given item. It also expands the timeline kinds to include the 2 different presentation types we will have for a media timeline (it is still hard-coded for the media browser flow for now).

The actions still aren't hooked up to anything.


Velin92 (approved):
LGTM

PR #3606, opened 2024-12-11
Design Tweaks

Velin92:
Changed the order of the room details actions to reflect the new designs (only for non DMs):  https://www.figma.com/design/jvZ7TIyd7V9Vn8sIDB9vMd/Compound-iOS-Templates?node-id=1-21740&t=MoQwmxfuwp8OhLdz-0
- removed the bloom bg from the join room screen
- now DMs do not necessarily need to be encrypted rooms
- changed bg of the knock banner


ElementX/Sources/Screens/JoinRoomScreen/View/JoinRoomScreen.swift
```diff
@@ -30,7 +30,6 @@ struct JoinRoomScreen: View {
             buttons
         }
         .alert(item: $context.alertInfo)
-        .background()
```
pixlwave, line 33: I think you still need this one here for the `bgCanvasDefault` background style to be applied (it'll change colour slightly in dark mode)

ElementX/Sources/Screens/RoomDetailsScreen/View/RoomDetailsScreen.swift
```diff
@@ -222,7 +227,6 @@ struct RoomDetailsScreen: View {
                         })
             }
         }
-        .disabled(context.viewState.notificationSettingsState.isLoading)
```
Velin92, line 225: @pixlwave what is the reason why this was getting disabled while the notification settings were loading?

pixlwave (approved):
Nice, I would still love to see the people row moved to the top of the list - it's basically always the thing I'm looking for when I come in here (and maybe in the future the media row), but this is already an improvement 👏

PR #3607, opened 2024-12-11
Hook up the actions in the media details sheet.

pixlwave:
Probably best described by a screen recording:

https://github.com/user-attachments/assets/57c717f4-ef41-4e56-ba7c-4eb16d4b993e


Velin92 (commented):
LGTM

PR #3608, opened 2024-12-11
Fix a bug where you're shown the remove caption action when it isn't available.

pixlwave:
I missed this when we re-ordered the action menu.


stefanceriu (approved):
👍

PR #3610, opened 2024-12-12
Custom media gallery views for files and voice messages

stefanceriu:
This PR extracts the timeline item backgrounds from the BubbledStyler and into its own modifier so that it can be reused for the media gallery views.
It also introduces different views for the files, audios and voice messages making use of the above.

Screenshot 2024-12-12 at 15 28 52


ElementX/Sources/Screens/Timeline/View/Style/TimelineItemBubbleBackground.swift
```diff
@@ -0,0 +1,54 @@
...
+        modifier(TimelineItemBubbleBackgroundModifier(timelineItem: timelineItem,
+                                                      insets: insets,
+                                                      color: color))
+    }
+}
+
+private struct TimelineItemBubbleBackgroundModifier: ViewModifier {
+    @Environment(\.timelineGroupStyle) private var timelineGroupStyle
+    
+    let timelineItem: EventBasedTimelineItemProtocol
```
pixlwave, line 23: From the perspective of using this for previews, if we passed in the `isOutgoing` boolean instead of the whole timeline item here, we could use this modifier in places like `SwipeRightAction_Previews`, `LongPressWithFeedback_Previews` and `FormattedBodyText_Previews` and get rid of those made up ones.

PreviewTests/Sources/__Snapshots__/PreviewTests/test_mediaEventsTimelineScreen-iPad-en-GB.Files.png
pixlwave, line 1: Should we add a  scroll view to this preview so it is easier to develop with and then add a `.previewLayout` that fits the whole stack?

ElementX/Sources/Screens/MediaEventsTimelineScreen/View/TimelineViews/AudioFileEventsTimelineView.swift
```diff
@@ -0,0 +1,61 @@
...
+//
+
+import Compound
+import SwiftUI
+
+struct AudioMediaEventsTimelineView: View {
+    let timelineItem: AudioRoomTimelineItem
+    
+    var body: some View {
+        MediaFileRoomTimelineContent(timelineItemID: timelineItem.id,
```
pixlwave, line 15: Oh neat, I forgot I had extracted this part already, that's super useful 😅

pixlwave (approved):
👌

PR #3615, opened 2024-12-13
Various danger swift rule tweaks

stefanceriu:
* remove the ticket number check, it's noisy
* remove signoff now that we have cla checks
* fix description image attachments check


pixlwave (approved):
Nice!

Want to update https://github.com/element-hq/element-x-ios/blob/develop/.github/PULL_REQUEST_TEMPLATE.md in the PR while we're at it?

PR #3617, opened 2024-12-13
Security and privacy part 1

Velin92:
I have committed so far everything that ca be done without requiring another PR from the SDK that is in WIP too, so that we can split this work into smaller pieces, that are not blocked, instead of having a huge PR that is awaiting for an SDK PR to be merged in main.

This adds a new option for non DM rooms to manage the security and privacy settings, so far it's using the same FF as knocking, but we may change this later and use a new FF in case we want to ship this feature separately.


ElementX/Sources/Screens/SecurityAndPrivacyScreen/View/SecurityAndPrivacyScreen.swift
```diff
@@ -0,0 +1,87 @@
...
+        Section {
+            ListRow(label: .plain(title: L10n.screenSecurityAndPrivacyRoomAccessInviteOnlyOptionTitle,
+                                  description: L10n.screenSecurityAndPrivacyRoomAccessInviteOnlyOptionDescription),
+                    kind: .selection(isSelected: context.desiredSettings.accessType == .inviteOnly) { context.desiredSettings.accessType = .inviteOnly })
+            ListRow(label: .plain(title: L10n.screenSecurityAndPrivacyAskToJoinOptionTitle,
+                                  description: L10n.screenSecurityAndPrivacyAskToJoinOptionDescription),
+                    kind: .selection(isSelected: context.desiredSettings.accessType == .askToJoin) { context.desiredSettings.accessType = .askToJoin })
+            ListRow(label: .plain(title: L10n.screenSecurityAndPrivacyRoomAccessAnyoneOptionTitle,
+                                  description: L10n.screenSecurityAndPrivacyRoomAccessAnyoneOptionDescription),
+                    kind: .selection(isSelected: context.desiredSettings.accessType == .anyone) { context.desiredSettings.accessType = .anyone })
```
pixlwave, line 36: Part of me thinks we should update `kind: .picker`/`.inlinePicker` to take a `description` too, then we could use that here instead of having 3 individual selection rows.

PreviewTests/Sources/__Snapshots__/PreviewTests/test_roomDetailsScreen-iPhone-16-en-GB.Generic-Room.png
pixlwave, line 1: Awww, the People button is going down again! 🥺

ElementX/Sources/Screens/SecurityAndPrivacyScreen/View/SecurityAndPrivacyScreen.swift
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+    
+    @ViewBuilder
+    private var encryptionSection: some View {
+        let encryptionBinding = Binding<Bool>(get: {
+            context.desiredSettings.isEncryptionEnabled
+        }, set: { newValue in
+            context.send(viewAction: .tryUpdatingEncryption(newValue))
+        })
```
pixlwave, line 49: Would it be better to directly bind this to the designedSettings, and then use an `onChange` to handle the update so that the user feedback is more immediate? That way when the user toggles it on, the toggle will animate, and you can set it back to false (in a `withAnimation`) if they cancel.

ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenViewModel.swift
```diff
@@ -0,0 +1,61 @@
...
+                                                 secondaryButton: .init(title: L10n.actionCancel, role: .cancel, action: nil))
+            } else {
+                state.bindings.desiredSettings.isEncryptionEnabled = false
+            }
+        }
+    }
+}
+
+private extension RoomInfoProxy {
+    var roomAcessType: SecurityAndPrivacyRoomAccessType {
```
pixlwave, line 51:
```suggestion
    var roomAccessType: SecurityAndPrivacyRoomAccessType {
```

pixlwave (approved):
LGTM 👮‍♂️

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -1453,6 +1465,24 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
...
     
+    private func presentSecurityAndPrivacyScreen() {
+        let coordinator = SecurityAndPrivacyScreenCoordinator(parameters: .init(roomProxy: roomProxy))
+        
+        coordinator.actionsPublisher.sink { [weak self] action in
+            guard let self else { return }
+            
+            switch action {
+            case .done:
+                break
```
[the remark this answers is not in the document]
Velin92: I guess... uhm... yeah about this I'll probably check next week with design because I am unsure if we want the save button to pop or not the screen, but will address it in another PR

ElementX/Sources/Screens/SecurityAndPrivacyScreen/View/SecurityAndPrivacyScreen.swift
Velin92: so I checked the designs and while the alert displays the button should be off... but yeah I will double check with design if we want the button to stay off until the alert is accepted or if we want to revert if refused.

PR #3624, opened 2024-12-16
Fix recently used emojis

stefanceriu:
3 bug fixes:
* the recently used emojis stop working properly on 18.2 because of our usage of the private iOS preferences user defaults
* the media gallery was sometimes showing the empty state at wrong times
* video thumbnails were aspect fitting instead of filling and they weren't the right design


ElementX/Sources/Services/Timeline/TimelineController/MockRoomTimelineController.swift
```diff
@@ -36,16 +36,17 @@ class MockRoomTimelineController: RoomTimelineControllerProtocol {
     
     init(timelineKind: TimelineKind = .live, listenForSignals: Bool = false) {
         self.timelineKind = timelineKind
-        paginationState = PaginationState(backward: .idle, forward: .timelineEndReached)
+        
         callbacks.send(.isLive(true))
```
pixlwave, line 40: 2 things

- Didn't this use to send the pagination state in the init (it won't now because didSet isn't triggered in here)
- To match the real RoomTimelineController this should send the pagination state before the isLive.

ElementX/Sources/Application/AppSettings.swift
```diff
@@ -231,6 +232,9 @@ final class AppSettings {
     @UserPreference(key: UserDefaultsKeys.hasRunIdentityConfirmationOnboarding, defaultValue: false, storageType: .userDefaults(store))
     var hasRunIdentityConfirmationOnboarding
     
+    @UserPreference(key: UserDefaultsKeys.frequentlyUsedSystemEmojis, defaultValue: [String](), storageType: .userDefaults(store))
+    var frequentlyUsedSystemEmojis
```
pixlwave, line 236: Do we want to do anything about counting the frequency of use, so that we could use that to influence the order in the future? Otherwise this is probably `recentlyUsed` rather than `frequentlyUsed`.

pixlwave (approved):
LGTM 👍

Velin92 (approved):

PR #3628, opened 2024-12-17
Automatically retry decryptions in the active room when the app becom…es active again just in case the NSE received keys we're not aware of

stefanceriu:
What it says on the tin.

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
+++ b/ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
@@ -196,6 +196,14 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol 
             .throttle(for: .milliseconds(100), scheduler: DispatchQueue.main, latest: true)
             .weakAssign(to: \.state.unseenKnockRequests, on: self)
             .store(in: &cancellables)
+        
+        NotificationCenter.default.publisher(for: UIApplication.didBecomeActiveNotification)
+            .sink { [weak self] _ in
+                Task {
+                    await self?.roomProxy.timeline.retryDecryption()
+                }
+            }
+            .store(in: &cancellables)
     }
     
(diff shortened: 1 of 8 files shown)


ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -196,6 +196,14 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
...
             .weakAssign(to: \.state.unseenKnockRequests, on: self)
             .store(in: &cancellables)
+        
+        NotificationCenter.default.publisher(for: UIApplication.didBecomeActiveNotification)
+            .sink { [weak self] _ in
+                Task {
+                    await self?.roomProxy.timeline.retryDecryption()
+                }
+            }
+            .store(in: &cancellables)
```
Velin92, line 206: Couldn't we use a darwin notification, and have the NSE send one when it received new keys? (it it even possible to check that through the SDK?)

stefanceriu: Yeah, we don't know what the SDK is doing underneath. The real fix will actually happen later in the sdk when they start doing cross process locks properly and trigger cross process notification themselves. This is just a bandaid until that happens.

ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
```diff
@@ -158,7 +158,7 @@ class TimelineInteractionHandler {
             MXLog.info("Showing debug info for \(eventTimelineItem.id)")
             actionsSubject.send(.showDebugInfo(debugInfo))
         case .retryDecryption(let sessionID):
-            Task { await timelineController.retryDecryption(for: sessionID) }
+            Task { await timelineController.retryDecryption(sessionIDs: [sessionID]) }
```
pixlwave, line 161: Can we get rid of this action now? 😇

PR #3630, opened 2024-12-17
Implement the save action when previewing media.

pixlwave:
Images and videos are saved to the photo library, files ask the user to choose a directory to save them into.

https://github.com/user-attachments/assets/7a75a626-827f-46ec-9c44-c4feaa3ce977

### DIFF
--- a/ElementX/Sources/Screens/FilePreviewScreen/PhotoLibraryManager.swift
+++ b/ElementX/Sources/Screens/FilePreviewScreen/PhotoLibraryManager.swift
@@ -1,0 +1,37 @@
+//
+// Copyright 2024 New Vector Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+// Please see LICENSE in the repository root for full details.
+//
+
+import Photos
+
+enum PhotoLibraryError: Error {
+    case notAuthorized
+    case unknown(Error)
+}
+
+// sourcery: AutoMockable
+protocol PhotoLibraryManagerProtocol {
+    func add(_ type: PHAssetResourceType, at url: URL) async -> Result<Void, PhotoLibraryError>
+}
+
+struct PhotoLibraryManager: PhotoLibraryManagerProtocol {
+    func add(_ type: PHAssetResourceType, at url: URL) async -> Result<Void, PhotoLibraryError> {
+        do {
+            try await PHPhotoLibrary.shared().performChanges {
+                let request = PHAssetCreationRequest.forAsset()
+                let options = PHAssetResourceCreationOptions()
+                request.addResource(with: type, fileURL: url, options: options)
+            }
+            return .success(())
+        } catch {
+            if (error as NSError).code == PHPhotosError.accessUserDenied.rawValue {
+                return .failure(.notAuthorized)
+            } else {
+                return .failure(.unknown(error))
+            }
+        }
+    }
+}
--- a/ElementX/Sources/Screens/FilePreviewScreen/PhotoLibraryManager.swift
+++ b/ElementX/Sources/Screens/FilePreviewScreen/PhotoLibraryManager.swift
@@ -7,18 +7,18 @@
 
 import Photos
 
-enum PhotoLibraryError: Error {
+enum PhotoLibraryManagerError: Error {
     case notAuthorized
     case unknown(Error)
 }
 
 // sourcery: AutoMockable
 protocol PhotoLibraryManagerProtocol {
-    func add(_ type: PHAssetResourceType, at url: URL) async -> Result<Void, PhotoLibraryError>
+    func addResource(_ type: PHAssetResourceType, at url: URL) async -> Result<Void, PhotoLibraryManagerError>
 }
 
 struct PhotoLibraryManager: PhotoLibraryManagerProtocol {
-    func add(_ type: PHAssetResourceType, at url: URL) async -> Result<Void, PhotoLibraryError> {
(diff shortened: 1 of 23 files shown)


ElementX/Sources/Screens/FilePreviewScreen/PhotoLibraryManager.swift
```diff
@@ -0,0 +1,37 @@
...
+import Photos
+
+enum PhotoLibraryError: Error {
+    case notAuthorized
+    case unknown(Error)
+}
+
+// sourcery: AutoMockable
+protocol PhotoLibraryManagerProtocol {
+    func add(_ type: PHAssetResourceType, at url: URL) async -> Result<Void, PhotoLibraryError>
```
stefanceriu, line 17: Why not keep the original `addResource` name?

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewViewModel.swift
```diff
@@ -22,12 +24,14 @@ class TimelineMediaPreviewViewModel: TimelineMediaPreviewViewModelType {
     
     init(context: TimelineMediaPreviewContext,
          mediaProvider: MediaProviderProtocol,
-         userIndicatorController: UserIndicatorControllerProtocol) {
+         photoLibraryManager: PhotoLibraryManagerProtocol = PhotoLibraryManager(),
```
stefanceriu, line 27: Do we really need a default value?

ElementX/Sources/Screens/FilePreviewScreen/PhotoLibraryManager.swift
```diff
@@ -0,0 +1,37 @@
+//
+// Copyright 2024 New Vector Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+// Please see LICENSE in the repository root for full details.
+//
+
+import Photos
+
+enum PhotoLibraryError: Error {
```
stefanceriu, line 10:
```suggestion
enum PhotoLibraryManagerError: Error {
```

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewViewModel.swift
```diff
@@ -108,12 +109,43 @@ class TimelineMediaPreviewViewModel: TimelineMediaPreviewViewModelType {
...
+            showErrorIndicator()
+        }
+    }
+    
+    private func redactItem(_ item: TimelineMediaPreviewItem) {
+        timelineViewModel.context.send(viewAction: .handleTimelineItemMenuAction(itemID: item.id, action: .redact))
+        state.bindings.redactConfirmationItem = nil
+        state.bindings.mediaDetailsItem = nil
+        actionsSubject.send(.dismiss)
+        showRedactedIndicator()
```
stefanceriu, line 148: This assumes that the redaction will never fail?

ElementX/Sources/Screens/FilePreviewScreen/View/TimelineMediaPreviewScreen.swift
```diff
@@ -102,7 +107,7 @@ struct TimelineMediaPreviewView: View {
                 Spacer()
                 
                 Button { context.send(viewAction: .saveCurrentItem) } label: {
-                    CompoundIcon(\.download)
+                    CompoundIcon(\.downloadIos)
```
stefanceriu, line 110: Would be nice if these would be already platform specific when they hit compound-ios

stefanceriu (changes requested):
Very nice! 👌

ElementX/Sources/Screens/FilePreviewScreen/PhotoLibraryManager.swift
pixlwave: Ah fair point, this started of as a `MediaFileManagerProtocol` where adding to the photos library was one of a few actions, but the rest got scrapped when I realised the document picker can be configured to export a file.

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewViewModel.swift
pixlwave: I didn't see any harm when the default type is a struct so couldn't be shared between multiple places, but updated to remove it.

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewViewModel.swift
pixlwave: It does, yes 😅 (I considered that the timeline's problem).

ElementX/Sources/Screens/FilePreviewScreen/View/TimelineMediaPreviewScreen.swift
pixlwave: In this instance the change is because this is a new icon. I get what you're saying although if you look at the Figma, you'll still be told the icon is `download-ios` so it probably brings little value in the long run.

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewViewModel.swift
stefanceriu: Fair enough 👍

stefanceriu (approved):
🐑 it!

PR #3636, opened 2024-12-18
Fix the interactive dismiss gesture of media previews on iOS 18.2

pixlwave:
The root view was being ignored (well for gestures anyway) as it was clear.


ElementX/Sources/FlowCoordinators/MediaEventsTimelineFlowCoordinator.swift
pixlwave, line 1: The timing of this being called changed from after the animation had completed (which presumably was to do with the preview controller as it doesn't make sense looking at the navigation stack code), to as soon as you lift your finger on the drag (which is too soon to undo the flip).

pixlwave:
Closing this for now, it breaks interaction with the QL controller 🤦‍♂️

PR #3641, opened 2024-12-19
Retrofit `deferFulfillment` onto snapshot tests.

stefanceriu:
We noticed that some of our preview tests are flakey, some even with significat delays. We've already been down this path with unit tests for which we introduced the `deferFulfillment` and waiting for view state changes before asserting values.

This PR introduces the same thing in the preview tests in the form of a delayed deferment publisher parameter on the existing `snapshotPreferences` view extension.


ElementX/Sources/Other/Extensions/Snapshotting.swift
```diff
@@ -49,8 +67,18 @@ public extension SwiftUI.View {
...
-    func snapshotPreferences(delay: TimeInterval = .zero, precision: Float = 1.0, perceptualPrecision: Float = 0.98) -> some SwiftUI.View {
+    func snapshotPreferences(expect fulfillmentPublisher: (any Publisher<Bool, Never>)? = nil,
+                             precision: Float = 1.0,
+                             perceptualPrecision: Float = 0.98) -> some SwiftUI.View {
+        preference(key: SnapshotPrecisionPreferenceKey.self, value: precision)
+            .preference(key: SnapshotPerceptualPrecisionPreferenceKey.self, value: perceptualPrecision)
+            .preference(key: SnapshotFulfillmentPublisherPreferenceKey.self, value: FulfillmentPublisherEquatableWrapper(publisher: fulfillmentPublisher?.eraseToAnyPublisher()))
+    }
+    
+    @available(*, deprecated, message: "Use the fulfillment variant instead")
```
Velin92, line 78: I would still keep this as not deprecated, there might be some cases where the checking the view state is not enough, for example we may have a view that after a change in the view state updtest it's internal state with some delay (I think we have something similar in the read receipts)

Velin92 (approved):
Approved with a suggestion

ElementX/Sources/Other/Extensions/Snapshotting.swift
stefanceriu: Talked about this offline: seeing as I've rewritten all the tests to no use a delay let's try to get rid of it for now. We can always add it back if really necessary.

PR #3643, opened 2024-12-19
Media upload tweaks

pixlwave:
The PR makes a few tweaks to the media upload screen:

- Automatically focusses the caption composer when a hardware keyboard is attached.
- Adds key handlers to the caption composer to `send` (enter) and `cancel` (escape).
- Fixes a bug where interaction wasn't disabled until processing had completed (it was waiting for sending to start).
- Starts processing the media as soon as the screen is presented. This means that
    - the screen is unlikely to block anymore for images, audio and files (instant sending!)
    - the screen should block for less time for videos

https://github.com/user-attachments/assets/a0febf3c-8d25-4633-b34c-1266bd53c882

This is the iOS counterpart to https://github.com/element-hq/element-x-android/pull/3943


Velin92 (approved):
Nice, LGTM

PR #3646, opened 2024-12-20
Fix the overlapping scrollbars on the room list filters on macOS.

### LINKED ISSUE #2583: Horizontal Scrollbar overlays filters on macOS
### Steps to reproduce

Open ElementX on macOS

### Outcome

Screenshot 2024-03-19 at 16 28 40

### Your phone model

macBook Pro

### Operating system version

macOS

### Application version

Version 1.5.12 (99)

### Homeserver

_No response_

### Will you send logs?

No

pixlwave:
This PR makes 2 changes:

- Fix the filters scroll view content so that if isn't clipped on the leading edge and has matching padding on the trailing edge
- Add some extra padding to the content on macOS so that any scroll bars don't overlap (the padding is then removed from the view itself so no extra space is take up). Fixes #2583

| Before iPhone | After iPhone |
| - | - |
| before iphone | after iphone |

| Mouse macOS | No mouse macOS |
| - | - |
| mouse macOS | no mouse macOS |

### DIFF
--- a/ElementX/Sources/Screens/HomeScreen/View/Filters/RoomListFiltersView.swift
+++ b/ElementX/Sources/Screens/HomeScreen/View/Filters/RoomListFiltersView.swift
@@ -12,6 +12,10 @@ struct RoomListFiltersView: View {
     @Binding var state: RoomListFiltersState
     @Namespace private var namespace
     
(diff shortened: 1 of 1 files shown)


stefanceriu (approved):
Woo! 👏

PR #3662, opened 2025-01-06
Add the event cache to the Rust tracing configuration.

pixlwave:
Defaults to `info`, allows configuring it lower in the developer options screen.


bnjbvr (approved):
lgtm, thanks!

PR #3663, opened 2025-01-06
Include missing gems until Fastlane is updated for Ruby 3.4

pixlwave:
As mentioned in https://github.com/fastlane/fastlane/issues/29183#issuecomment-2567093826

Tested and Xcode Cloud seems happy now.


manuroe (approved):
LGTM. I can only say the PR matches the comment you mentionned :D

PR #3667, opened 2025-01-08
Delay snapshotting various flakey UI tests

stefanceriu:
- as per https://app.codecov.io/gh/element-hq/element-x-ios/tests/develop?historicalTrend=INTERVAL_30_DAY&parameter=FAILED_TESTS

We've been getting quite a few failures on `AppLockSetupUITests` but there's nothing obviously wrong with them. Running them repeatedly locally though eventually results in random snapshotting failures, which might very well be the real reason.

As such I went and added delays to all the UI tests that have been reported as flakey on Codecov 🤞

### DIFF
--- a/UITests/Sources/AppLockSetupUITests.swift
+++ b/UITests/Sources/AppLockSetupUITests.swift
@@ -35,7 +35,7 @@ class AppLockSetupUITests: XCTestCase {
         try await Task.sleep(for: .seconds(0.5))
         
         // Create PIN screen.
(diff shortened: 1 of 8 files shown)


pixlwave (approved):
Heh, I didn't realise/forgot that second parameter was even there 🙈

pixlwave:
Note: I don't think this is going to fix the AppLock tests though. They're not failing with bad snapshots, they're failing with weird lock-ups that cause the test to time-out:

Screenshot 2025-01-08 at 9 06 09 am

PR #3668, opened 2025-01-09
Move tracing configuration to the rust side.

stefanceriu:
Requires https://github.com/matrix-org/matrix-rust-sdk/pull/4492


ElementX/Sources/Services/BugReport/BugReportService.swift
```diff
@@ -163,8 +163,7 @@ class BugReportService: NSObject, BugReportServiceProtocol {
             MultipartFormData(key: "fallback_language", type: .text(value: Bundle.app.developmentLocalization ?? "null")),
             MultipartFormData(key: "local_time", type: .text(value: localTime)),
             MultipartFormData(key: "utc_time", type: .text(value: utcTime)),
-            MultipartFormData(key: "base_bundle_identifier", type: .text(value: InfoPlistReader.main.baseBundleIdentifier)),
-            MultipartFormData(key: "rust_tracing_filter", type: .text(value: RustTracing.currentTracingConfiguration?.filter ?? "null"))
```
pixlwave, line 167: Might want to check with Rust folks if they still want this in the RS - if so presumably we'll need a way to ask Rust for it back.

pixlwave (approved):
LGTM! Took me a second to figure out what the new `appTarget` parameter did in the setup call, but I got there in the end 😅

ElementX/Sources/Services/BugReport/BugReportService.swift
stefanceriu: I'll check and do it in a follow up PR if needed

PR #3670, opened 2025-01-10
Simplify how to we handle background task expirations.

stefanceriu:
Hopefuly fix https://sentry.tools.element.io/organizations/element/issues/4477794/?project=44&query=is%3Aunresolved&referrer=issue-stream&stream_index=2

### DIFF
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -965,30 +965,30 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
             return
         }
         
-        backgroundTask = appMediator.beginBackgroundTask { [weak self] in
-            guard let self else { return }
-            
+        backgroundTask = appMediator.beginBackgroundTask {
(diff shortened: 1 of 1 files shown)


ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -965,30 +965,30 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
...
+        backgroundTask = appMediator.beginBackgroundTask {
             MXLog.info("Background task is about to expire.")
-            stopSync(isBackgroundTask: true) { [weak self] in
-                guard let self, let backgroundTask else { return }
-                
-                MXLog.info("Ending background task.")
-                appMediator.endBackgroundTask(backgroundTask)
-                self.backgroundTask = nil
+            
+            self.stopSync(isBackgroundTask: true) {
```
pixlwave, line 971: Maybe worth adding a comment that the strong self is intentional to (try) fix a crash and that any retain cycle will be fixed in endActiveBackgroundTask.

pixlwave (approved):
LGTM! 🤞

PR #3679, opened 2025-01-15
Warn and block sending on verification violation

BillCarsonFr:
### Pull Request Checklist

Part of https://github.com/element-hq/element-meta/issues/2492 (still partial as there is no way to verify other users yet on EX, so you can only withdraw the verification requirement).
EXI counterpart of EXA https://github.com/element-hq/element-x-android/pull/4126

Adds support for identity violation detection. Will show a warning notice and block sending (notice that sending would fail anyhow in that case).
image

withdraw_exi

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
+++ b/ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
@@ -227,7 +227,6 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol 
 
                 identityVerificationViolations[change.userId] = member
             default:
-                // clear all
                 identityVerificationViolations[change.userId] = nil
                 identityPinningViolations[change.userId] = nil
             }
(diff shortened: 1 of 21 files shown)


ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -209,21 +212,31 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
...
+                    MXLog.error("Failed retrieving room member for identity status change: \(change)")
+                    continue
+                }
+
+                identityVerificationViolations[change.userId] = member
             default:
-                break
+                // clear all
+                identityVerificationViolations[change.userId] = nil
+                identityPinningViolations[change.userId] = nil
```
stefanceriu: Are you sure this is correct, does .pinned imply verification and the other way around?

stefanceriu (approved):
Super stuff Valere, thank you for helping us out with this!

I left one question and took the liberty of making a few small changes, hope that's okay.

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
BillCarsonFr: Technically, you are right. Pinning and Verification are orthogonal things, i.e you can have a pinned identity that is in verification violation. But the `IdentityStatus` implements a sort or priorisation mechanism. I.e if the identity is pinned but in verification violation, it will report the status as `.verificationViolation`.

So if the state moved to a non-violation state it is fine.

(FTR when we withdraw the verification, we also pin the identity. So the status will move from .verificationViolation to .pinned)

andybalaam: Yep, correct.

andybalaam (approved):

PR #3680, opened 2025-01-15
Bump the RustSDK to v25.01.15 and fix (most) concurrency sendability warnings in the generated mocks

stefanceriu:


BillCarsonFr (approved):
Looks sane to me

PR #3688, opened 2025-01-20
Fix flakey `RoomMemberDetailsScreen` preview test.

stefanceriu:
- was missing sometimes missing the call button because we weren't waiting for the `dmRoomID` resolution


pixlwave (approved):
Nice, good spot 👍

PR #3690, opened 2025-01-21
Possible fix for the join room screen not updating

Velin92:
Not sure if this is the best way to fix the issue given the API change to the room preview API that has been made, but once the request was sent, the state of the joined room was not updated, since the room preview was checked first but never updated.
Example:

https://github.com/user-attachments/assets/5ef8afdf-e98b-4f5b-bf82-9de989c42665

This should however fix the issue:

https://github.com/user-attachments/assets/7996607e-dc63-486d-8e80-193078e69090

### DIFF
--- a/ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenViewModel.swift
+++ b/ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenViewModel.swift
@@ -95,13 +95,6 @@ class JoinRoomScreenViewModel: JoinRoomScreenViewModelType, JoinRoomScreenViewMo
         // See if we known about the room locally and, if so, have that
         // take priority over the preview one.
         if let room = await clientProxy.roomForIdentifier(roomID) {
-            // We also need to update the room preview
-            switch await clientProxy.roomPreviewForIdentifier(roomID, via: via) {
-            case .success(let updatedPreview):
-                roomPreview = updatedPreview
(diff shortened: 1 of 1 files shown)


Velin92:
I wonder if we shouldn't switch around this code:
  ```
          switch roomPreview.info.membership {
            case .invited:
                state.mode = .invited
            case .knocked:
                state.mode = .knocked
            case .banned:
                state.mode = .banned(sender: membershipDetails?.senderRoomMember?.displayName ?? membershipDetails?.senderRoomMember?.userID,
                                     reason: membershipDetails?.ownRoomMember.membershipChangeReason)
            default:
                switch roomPreview.info.joinRule {
                case .private, .invite:
                    state.mode = .inviteRequired
                case .knock, .knockRestricted:
                    state.mode = appSettings.knockingEnabled ? .knockable : .joinable
                case .restricted:
                    state.mode = .restricted
                default:
                    state.mode = .joinable
                }
            }
        } else if let room {
            switch room {
            case .invited:
                state.mode = .invited
            case .knocked:
                state.mode = .knocked
            case .banned:
                state.mode = .banned(sender: nil, reason: nil)
            default:
                state.mode = .joinable
            }
        }
    }
```

So that the room takes priority over the room preview, however I already tried this, and what happens is that the preview won't render the knockable state if I do so, so what I did for now is just also refresh the room preview after we refresh the room, when a change has been made @stefanceriu

PR #3693, opened 2025-01-21
DM Design Tweaks

Velin92:
- Added a Profile button that brings you to the room member detail of the DM recipient.
- Removed the stacked avatars in DMs


ElementX/Sources/Screens/RoomDetailsScreen/View/RoomDetailsScreen.swift
```diff
@@ -79,9 +77,9 @@ struct RoomDetailsScreen: View {
         .accessibilityIdentifier(A11yIdentifiers.roomDetailsScreen.avatar)
     }
     
-    private func dmHeaderSection(accountOwner: RoomMemberDetails, recipient: RoomMemberDetails) -> some View {
-        AvatarHeaderView(accountOwner: accountOwner,
-                         dmRecipient: recipient,
+    private func dmHeaderSection(recipient: RoomMemberDetails) -> some View {
+        AvatarHeaderView(member: recipient,
+                         avatarSize: .room(on: .details),
```
pixlwave, line 82: I think this should be using
```suggestion
        AvatarHeaderView(room: context.viewState.details,
                         avatarSize: .room(on: .details),
```

so that we still get room related badges on this header as the `member` and `user` ones are for their profile screens. The detail's `avatar` property should be configured with heroes so it should still show the other user's avatar iirc.

pixlwave (approved):
1 thought on the implementation while we wait to hear back on the Profile button. But otherwise looks good to me.

pixlwave (approved):
Nice 👌

pixlwave (approved):

PR #3695, opened 2025-01-22
Better handling for editing alias in case of different HS

Velin92:
We only update the alias if the new alias belongs to the same homeserver or one did not exist yet, otherwise we add it in the list of the alternative ones

### DIFF
--- a/ElementX/Sources/Screens/EditRoomAddressScreen/EditRoomAddressScreenViewModel.swift
+++ b/ElementX/Sources/Screens/EditRoomAddressScreen/EditRoomAddressScreenViewModel.swift
@@ -129,13 +129,13 @@ class EditRoomAddressScreenViewModel: EditRoomAddressScreenViewModelType, EditRo
             hideLoadingIndicator()
         }
         
-        guard let canonicalAlias = String.makeCanonicalAlias(aliasLocalPart: state.bindings.desiredAliasLocalPart, serverName: state.serverName),
-              isRoomAliasFormatValid(alias: canonicalAlias) else {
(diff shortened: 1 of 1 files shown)


ElementX/Sources/Screens/EditRoomAddressScreen/EditRoomAddressScreenViewModel.swift
```diff
@@ -146,24 +146,38 @@ class EditRoomAddressScreenViewModel: EditRoomAddressScreenViewModelType, EditRo
...
+        // Allow to update the canonical alias only if the saved canonical alias matches the homeserver or if there is no canonical alias
+        if savedCanonicalAlias == nil || savedCanonicalAlias?.hasSuffix(state.serverName) == true {
+            if case .failure = await roomProxy.updateCanonicalAlias(desiredCanonicalAlias, altAliases: roomProxy.infoPublisher.value.alternativeAliases) {
+                userIndicatorController.submitIndicator(.init(title: L10n.errorUnknown))
+                return
+            }
+            // Otherwise, update the alternative aliases and keep the current canonical alia
+        } else {
+            var newAlternativeAliases = roomProxy.infoPublisher.value.alternativeAliases
+            newAlternativeAliases.append(desiredCanonicalAlias)
```
Velin92, line 176: @ganfra I see that in you code you added the alias to top of the list, does it matter if instead I append the alias at the bottom of the list?

stefanceriu (approved):
Makes sense to me 👍  
Would be useful to update the `EditRoomAddressScreenViewModelTests.testCorrectMethodsCalledOnSave` test as well now.

Velin92:
> Makes sense to me 👍 Would be useful to update the `EditRoomAddressScreenViewModelTests.testCorrectMethodsCalledOnSave` test as well now.

Just added a bunch of tests with expectation for all the possible cases!

PR #3696, opened 2025-01-22
Fix message completion trigger to work anywhere in the message

vickcoo:
### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/ComposerToolbar/CompletionSuggestionService.swift
+++ b/ElementX/Sources/Screens/RoomScreen/ComposerToolbar/CompletionSuggestionService.swift
@@ -90,7 +90,7 @@ final class CompletionSuggestionService: CompletionSuggestionServiceProtocol {
         
         let components = text.components(separatedBy: .whitespaces)
         
-        guard var lastComponent = components.last,
+        guard var lastComponent = components.first(where: { $0.first == SuggestionTriggerPattern.at.rawValue }),
               let range = text.range(of: lastComponent, options: .backwards),
               lastComponent.count > 0,
               let suggestionKey = SuggestionTriggerPattern(rawValue: lastComponent.removeFirst()),


stefanceriu:
Can you try testing out this new behavior within the `CompletionSuggestionServiceTests`?

vickcoo:
I've run the testing on `CompletionSuggestionServiceTests`, and they all pass

stefanceriu:
> I've run the testing on `CompletionSuggestionServiceTests`, and they all pass

As in write a brand new one just for this special case and assert that the suggestions show up

vickcoo:
Oh! Do you mean I need to write a new unit test for this?
 I have no idea what happened cause the flow fail

*I apologize If my English is unclear.

vickcoo:
Why did the CI/CD fail, And how can I solve the problems?

stefanceriu:
> Why did the CI/CD fail

Only the enterprise tests failed which is to be expected as you don't have access to that code, it's private. 
As far as I can tell your part is good but I'll review it properly next week.

vickcoo:
> As far as I can tell your part is good but I'll review it properly next week.

I got it! thank you for your reply

pixlwave (changes requested):
Hey @vickcoo, thanks for opening your first contribution 🎉

I've just tested it out, but I think we probably need to refine it a bit more. If you type `@` into the composer but don't want to mention someone it now always shows a suggestion and if you subsequently try to `@mention` someone it uses that first `@` as the trigger so your filtering doesn't work. If we're going to allow mentioning from anywhere in the message, we really need to use the caret position as the trigger rather than iterating through all the words. See an example here:

https://github.com/user-attachments/assets/390aa703-9248-4a0f-a79e-65c9dca3a07c

vickcoo:
Hi @pixlwave! thank you for your feedback, here are a few ideas

Solution 1
The `@` symbol triggers suggestions, but the cursor must be positioned next to it, even if the message contains multiple "@" symbols.

For example, suggestions appear when the cursor is positioned at 1️⃣ or 2️⃣; otherwise, no suggestions are shown.
```
    1️⃣     2️⃣
    ↓      ↓
Hi @, I'm @
```

Solution 2
We can trigger suggestions anywhere when typing the `@`, and other `@` will be a plain string

pixlwave:
Solution 1 sounds super to me if it's possible!

ElementX/Sources/Screens/RoomScreen/ComposerToolbar/CompletionSuggestionService.swift
```diff
@@ -83,23 +83,46 @@ final class CompletionSuggestionService: CompletionSuggestionServiceProtocol {
     
     // MARK: - Private
     
-    private func detectTriggerInText(_ text: String?) -> SuggestionTrigger? {
-        guard let text else {
+    private func detectTriggerInText(text: String?, selectedRange: NSRange?) -> SuggestionTrigger? {
+        guard let text, let selectedRange else {
```
pixlwave, line 87: Even if we remove the optionals, it would be good to still guard in here for `selectedRange.length == 0` so that we don't show the suggestions if the user has selected multiple characters (i.e. the cursor has to be between letters).

ElementX/Sources/Screens/RoomScreen/ComposerToolbar/CompletionSuggestionService.swift
```diff
@@ -83,23 +83,46 @@ final class CompletionSuggestionService: CompletionSuggestionServiceProtocol {
...
-              let suggestionKey = SuggestionTriggerPattern(rawValue: lastComponent.removeFirst()),
-              // If a second character exists and is the same as the key it shouldn't trigger.
-              lastComponent.first != suggestionKey.rawValue else {
-            return nil
+        var startIndex: String.Index = text.startIndex
+        let components = text
+            .components(separatedBy: .whitespaces)
+            .filter { $0.first == SuggestionTriggerPattern.at.rawValue }
+        
+        for component in components {
```
pixlwave, line 99: I wonder if we could optimise this a bit more? Breaking up the message into words, and then searching for the matching words again in the message seems like it could get slow if there are multiple `@` chars in a message. Something along the lines of:
- Use `text.ranges(of: wordRegex)` to get the ranges of individual words.
- Use the selectedRange to find which word the cursor is in/next to.
- Use that range to check whether the first character matches the pattern.

pixlwave (commented):
Thanks for the updates. I just tried this out and it is **really** nice! 🤩

I've added a couple of comments about the new algorithm inline, but one other thing I noticed was that if you accept a suggestion in the middle, the cursor jumps to the end of the message which might be a bit annoying:

https://github.com/user-attachments/assets/91e0cf61-67dc-4ea0-b90f-0d835aa04fc5

ElementX/Sources/Screens/RoomScreen/ComposerToolbar/CompletionSuggestionService.swift
pixlwave: Oh or even simpler something like:

```
let atMentionRegex = /@\w+/
text.ranges(of: atMentionRegex).last { $0.contains(selectedRange.location) }
```

would be both simple and nicely optimised (I haven't tried this out - I'm imagining the `selectedRange` will need converting to a `String.Index` at the very least). The regex could be stored in `SuggestionTriggerPattern.at` too :)

vickcoo: Thank! The regex solution seems good to me, I even haven't considered this solution.
So, the `filter` might slow down the performance right?

I'll give it a try 🤗

pixlwave: > So, the `filter` might slow down the performance right?

Not inherently, but doing that and then following it with a `for component in components` that does a `text.range(of)` feels like we are searching through data more than we might need to. There's definitely nothing wrong with this solution, simply wondering if we can optimise it a bit 🙂

vickcoo: I got it, I just wondering know more things, that could let me put more thought into my brain to improve my code next time before review 🤩

ElementX/Sources/Screens/RoomScreen/ComposerToolbar/ComposerToolbarViewModel.swift
```diff
@@ -485,6 +485,9 @@ final class ComposerToolbarViewModel: ComposerToolbarViewModelType, ComposerTool
                 let attributedString = NSMutableAttributedString(attributedString: state.bindings.plainComposerText)
                 mentionBuilder.handleUserMention(for: attributedString, in: suggestion.range, url: url, userID: item.id, userDisplayName: item.displayName)
                 state.bindings.plainComposerText = attributedString
+                
+                let newSelectedRange = NSRange(location: state.bindings.selectedRange.location - completionSuggestionService.rawSuggestionText.count, length: 0)
+                state.bindings.selectedRange = newSelectedRange
```
vickcoo, line 490: I want to discuss this part which calculates the correct cursor location, but I don't know the data flow is right or not.

For example, the cursor at location `2`, then the text converts to pill after selected suggestion, it seems to occupy one character that causes me must to do this calculation
```
  ↓
@vic test

// This is a pill.
        ↓
(@vick) test
```

pixlwave: Hey, sorry for the slow reply, last week was rather busy. Could you add it to `MentionSuggestionItem` so that it's available in this `switch suggestion`? That feels like a nicer way to pass the value down :)

ElementX/Sources/Screens/RoomScreen/ComposerToolbar/CompletionSuggestionService.swift
```diff
@@ -8,8 +8,8 @@
 import Combine
 import Foundation
 
-private enum SuggestionTriggerPattern: Character {
-    case at = "@"
+private enum SuggestionTriggerRegex: String {
+    case at = "@\\w+"
```
pixlwave, line 12: No biggie if not, but any chance we could use Swift's `Regex` here instead of a string and an `NSRegularExpression`? That way it will be compile-time safe and you won't need to guard it.

pixlwave, line 12: No biggie if not, but any chance we could use Swift's `Regex` here instead of a string and an `NSRegularExpression`? That way it will be compile-time safe and you won't need to guard it.

```suggestion
    case at = /@\w+/
```

vickcoo: It seems not to be working because the enum doesn't support the `Regex<Substring>` type, also the `NSRegularExpression` does not support the type either. maybe need to do more work on this change, what do you think?

pixlwave: Oh I didn't spot that.

Just remove the `: String` for now and change the `case` to a `static let` - there's only trigger, we don't need the ability to iterate through them right now :)

ElementX/Sources/Screens/RoomScreen/ComposerToolbar/ComposerToolbarViewModel.swift
vickcoo: It's ok! thanks, I'll try it out.

ElementX/Sources/Screens/RoomScreen/ComposerToolbar/CompletionSuggestionService.swift
vickcoo: Ok! I'm already finished.

ElementX/Sources/Screens/RoomScreen/ComposerToolbar/CompletionSuggestionService.swift
```diff
@@ -83,23 +85,26 @@ final class CompletionSuggestionService: CompletionSuggestionServiceProtocol {
...
+        let matches = text.matches(of: SuggestionTriggerRegex.at)
+        let match = matches
+            .filter { matchResult in
+                let lowerBound = matchResult.range.lowerBound.utf16Offset(in: matchResult.base)
+                let upperBound = matchResult.range.upperBound.utf16Offset(in: matchResult.base)
+                return selectedRange.location >= lowerBound
+                    && selectedRange.location <= upperBound
+                    && selectedRange.length <= upperBound - lowerBound
+            }
+            .first
```
pixlwave, line 98: We can use [`first(where:)`](https://developer.apple.com/documentation/swift/array/first(where:)) here rather than filtering and then taking the first element.
```suggestion
            .first { matchResult in
                let lowerBound = matchResult.range.lowerBound.utf16Offset(in: matchResult.base)
                let upperBound = matchResult.range.upperBound.utf16Offset(in: matchResult.base)
                return selectedRange.location >= lowerBound
                    && selectedRange.location <= upperBound
                    && selectedRange.length <= upperBound - lowerBound
            }
```

pixlwave (approved):
This looks and works great! There's just a small compile error to fix on the unit tests and then this will be ready to merge 👏

Edit: It's coming from SwiftGen. Make sure to `swift run tools setup-project` to ensure you have all the homebrew dependencies set up and then if you build the project you should see a change to the `GeneratedMocks.swift` that also needs to be committed.

ElementX/Sources/Screens/RoomScreen/ComposerToolbar/CompletionSuggestionService.swift
vickcoo: How do you know the problem, I am confused as to why the unit tests don't pass 😆.
I haven't notice this way to find result.

vickcoo:
> Thank you very much for your contribution @vickcoo. The tests passed (but appear to fail as the coverage upload fails) so I'm going to merge this PR now 🎉

Hi @pixlwave, Thank you for your help, and a lot of patience in helping me solve the problem! 😊

pixlwave:
You're welcome! I've just been using this on todays Nightly build and it rocks 🙌

PR #3699, opened 2025-01-23
Make sure the Recovery Key option is shown on the IdentityConfirmationScreen when available.

pixlwave:
There was a race condition where 2 different states could be processed at the same time:
Even though the method is `@MainActor`, the fact it has an `await` means that if another Task is spawned when a previous one is suspended, there was no guarantee that they would resume in the same order. This PR calls the method using a `for await in` to process the states serially.


stefanceriu (approved):
🙌

PR #3707, opened 2025-01-27
Media Browser: Listen to the timeline in the preview screen

pixlwave:

- Listen to updates from the timeline in the media preview view model so that you can swipe to items that weren't known when you opened the preview.
- Add a dedicated Data Source for the preview controller that adds padding either side of the initial set of items so that we don't need to reload the data when a response comes in (see the docs inline).
- Adds a new state for pagination to tell the user that more items are being loaded.
    - Note: this doesn't actually handle the pagination yet, nor is it aware of the timeline ends.

https://github.com/user-attachments/assets/e2588350-a661-424a-8031-b9a0bc324e8d

### DIFF
--- a/ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
+++ b/ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
@@ -1,0 +1,264 @@
+//
+// Copyright 2025 New Vector Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Combine
+import Foundation
+import QuickLook
+
+/// A dedicated data source for QLPreviewController to support timeline updates. This was added to
+/// workaround the fact that calling `reloadData` on the controller **always** reloads the current
+/// item (even if hasn't changed), so any interaction (zoom, media playback, scroll position) would be
+/// lost.
+///
+/// This data source pads the initial array with 100 spaces before and after, adding any pagination into
+/// this fixed space. This removes the need to reload the data and preserves the current item's index
+/// in the data.
+class TimelineMediaPreviewDataSource: NSObject, QLPreviewControllerDataSource {
+    /// All of the items in the timeline that can be previewed.
+    private(set) var previewItems: [TimelineMediaPreviewItem]
+    let previewItemsPaginationPublisher = PassthroughSubject<Void, Never>()
+    
+    private let initialItem: EventBasedMessageTimelineItemProtocol
+    /// The index of the initial item inside of `previewItems` that is to be shown.
+    let initialItemIndex: Int
+    
+    /// The media item that is currently being previewed.
+    private(set) var currentItem: TimelineMediaPreviewItem?
+    
+    private var backwardPadding: Int
+    private var forwardPadding: Int
+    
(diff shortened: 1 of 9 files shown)


ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
```diff
@@ -0,0 +1,264 @@
...
+import QuickLook
+
+/// A dedicated data source for QLPreviewController to support timeline updates. This was added to
+/// workaround the fact that calling `reloadData` on the controller **always** reloads the current
+/// item (even if hasn't changed), so any interaction (zoom, media playback, scroll position) would be
+/// lost.
+///
+/// This data source pads the initial array with 100 spaces before and after, adding any pagination into
+/// this fixed space. This removes the need to reload the data and preserves the current item's index
+/// in the data.
```
stefanceriu, line 19: 👌

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
```diff
@@ -0,0 +1,264 @@
...
+            
+            return newItem
+        }
+        
+        var hasPaginated = false
+        if let range = newItems.map(\.id).firstRange(of: previewItems.map(\.id)) {
+            let backPaginationCount = range.lowerBound
+            let forwardPaginationCount = newItems.indices.upperBound - range.upperBound
+            
+            // Don't worry about negative padding here. Turns out that it just limits
```
stefanceriu, line 69: Interesting behavior

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewViewModel.swift
```diff
@@ -79,42 +83,48 @@ class TimelineMediaPreviewViewModel: TimelineMediaPreviewViewModelType {
...
-            switch await mediaProvider.loadFileFromSource(source, filename: previewItem.filename) {
-            case .success(let handle):
-                previewItem.fileHandle = handle
-                state.fileLoadedPublisher.send(previewItem.id)
-            case .failure(let error):
-                MXLog.error("Failed loading media: \(error)")
-                context.objectWillChange.send() // Manually trigger the SwiftUI view update.
-                previewItem.downloadError = error
+        if let previewItem {
+            currentItemIDHandler?(previewItem.id)
```
stefanceriu, line 92: Is this still used? 🤔

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
```diff
@@ -0,0 +1,264 @@
...
+            forwardPadding -= forwardPaginationCount
+            
+            if backPaginationCount > 0 || forwardPaginationCount > 0 {
+                hasPaginated = true
+            }
+        } else {
+            // Do nothing! Not ideal but if we reload the data source the current item will
+            // also be, reloaded resetting any interaction the user has made with it. If we
+            // ignore the pagination, then the next time they swipe they'll land on a different
+            // media but this is probably less jarring overall. I hate QLPreviewController!
```
stefanceriu, line 82: > I hate QLPreviewController!

😅 
Best to remember this feeling and that we probaly reached its limits.

stefanceriu (approved):

pixlwave:
> Unfortunately it seems to break if I scroll to the next/previous page while the current one is still loading. That results in an black screen that never resolves.

We both managed to reproduce this on todays Nightly (not sure what the issue is yet) so it seems unrelated to this PR. Will look into it when going back to the "correct" approach to preview presentation.

PR #3708, opened 2025-01-27
Bump the RustSDK to 25.01.27, adopt the new emoji boosting API.

stefanceriu:


ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FormattedBodyText.swift
```diff
@@ -35,9 +35,7 @@ struct FormattedBodyText: View {
         
         let string = String(attributedString.characters)
         
-        if boostEmojiSize,
-           string.containsOnlyEmoji,
-           let range = adjustedAttributedString.range(of: string) {
+        if boostFontSize, let range = adjustedAttributedString.range(of: string) {
```
pixlwave, line 38: I think we can simplify this further now by getting rid of `let string = String(attributedString.characters)` and the range and simply doing `adjustedAttributedString.font = UIFont.systemFont(ofSize: 48.0)`.

pixlwave (commented):
LGTM, but I'm rebuilding the SDK as `25.01.27` 😉

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FormattedBodyText.swift
stefanceriu: I don't think we can do that: we're searching for the original string the adjusted one so that we don't also boost the extra characters needed for LTR and timestamp rendering.

pixlwave: My bad, forgot about that 🤦‍♂️

ElementX/Sources/Other/Extensions/ClientBuilder.swift
```diff
@@ -44,7 +44,7 @@ extension ClientBuilder {
                     .roomDecryptionTrustRequirement(trustRequirement: .crossSignedOrLegacy)
             } else {
                 builder = builder
-                    .roomKeyRecipientStrategy(strategy: .deviceBasedStrategy(onlyAllowTrustedDevices: false, errorOnVerifiedUserProblem: true))
+                    .roomKeyRecipientStrategy(strategy: .errorOnVerifiedUserProblem)
```
pixlwave, line 47: Changed based on the following:

https://github.com/matrix-org/matrix-rust-sdk/pull/4581/files#diff-8d716e55b5bccd9bb93db0bab9b6ea1813a8ab16eda1a98b43e767a158c38e74R112-R115

@richvdh does this look correct?

richvdh: can't vouch for the syntax, but the general idea looks right

pixlwave: Thank you!

PR #3714, opened 2025-01-29
Detect the timeline start/end when swiping through media files.

pixlwave:

- Make `TimelineMediaPreviewItem` an enum with `.loading` and `.media` cases to remove all the optionals.
    - I started the refactor whilst doing everything below and twigging that it makes more sense but I'm afraid that means it is mixed in to the same commit.
- Add the pagination state to the media preview data source and use it to vend loading items that indicate the timeline start/end.
- Automatically jump back to the first/last item when hitting the timeline start/end and show a user indicator that there are no more items to load.
- Bug fix: Configure the initial timeline proxy pagination state on the media browser like a live timeline so it knows it has already hit the end.

Note: This PR doesn't include detection of the start/end when loading more items and none are found.

https://github.com/user-attachments/assets/069e9a80-1cf0-4e47-88c1-6aafcabdc5f7

### DIFF
--- a/ElementX/Sources/Services/Timeline/TimelineProxy.swift
+++ b/ElementX/Sources/Services/Timeline/TimelineProxy.swift
@@ -586,10 +586,13 @@ final class TimelineProxy: TimelineProxyProtocol {
                 MXLog.error("Failed to subscribe to back pagination status with error: \(error)")
             }
             forwardPaginationStatusSubject.send(.timelineEndReached)
-        case .detached, .media:
+        case .detached:
             // Detached timelines don't support observation, set the initial state ourself.
             backPaginationStatusSubject.send(.idle)
             forwardPaginationStatusSubject.send(.idle)
+        case .media(let presentation):
+            backPaginationStatusSubject.send(.idle)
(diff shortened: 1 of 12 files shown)


ElementX/Sources/Services/Timeline/TimelineProxy.swift
```diff
@@ -586,10 +586,13 @@ final class TimelineProxy: TimelineProxyProtocol {
                 MXLog.error("Failed to subscribe to back pagination status with error: \(error)")
             }
             forwardPaginationStatusSubject.send(.timelineEndReached)
-        case .detached, .media:
+        case .detached:
             // Detached timelines don't support observation, set the initial state ourself.
```
stefanceriu, line 590: Is this comment still valid?

stefanceriu (approved):
Very nice, lgtm!

ElementX/Sources/Services/Timeline/TimelineProxy.swift
pixlwave: Yep, turns out I was wrong with my comment on that Rust PR. Everything continues to work as is right now, but maybe in the future we will get an API that matches between the different types.

PR #3716, opened 2025-01-29
Attempt to fix the every flakey app lock setup test

stefanceriu:
We noticed that confirming the PIN will automatically hide the keyboard which makes the whole screen shift down while attempting to press the allow button. We also see an ` App animations complete notification not received, will attempt to continue.` warning which seems to indicate the same.


pixlwave (approved):
🤞

PR #3719, opened 2025-01-30
Revert back to UIKit for the presentation of the timeline media preview.

pixlwave:
This PR essentially undoes #3619 as embedding the QLPreviewController wasn't going so well:
- Interactive dismiss broke in iOS 18.2 (it triggered immediately instead of letting you slowly drag down).
- Attempting to add any interactivity on top of the preview controller made the preview completely un-interactive.
- The audio player didn't show the progress waveform as it attempted to attach it to it's own bottom bar which it was unable to present.
- Sometimes a preview would present without a navigation bar as the preview controller *really* wants to mess with it in all kinds of ways.

This means we lose some control and have to accept the iOS defaults, but it works *much* better and that is important as we'll be replacing the timeline previews too so that swiping works. The last commit is a bit of a workaround but it seems good enough for now.

https://github.com/user-attachments/assets/d78774ab-d4ca-40b7-9504-25e0dad16424


ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewModifier.swift
```diff
@@ -0,0 +1,188 @@
...
+extension View {
+    /// Preview a media file using a QuickLook Preview Controller. The preview is interactive with
+    /// the dismiss gesture working as expected if it was presented from UIKit.
+    func timelineMediaPreview(viewModel: Binding<TimelineMediaPreviewViewModel?>) -> some View {
+        modifier(TimelineMediaPreviewModifier(viewModel: viewModel))
+    }
+}
+
+private struct TimelineMediaPreviewModifier: ViewModifier {
+    @Binding var viewModel: TimelineMediaPreviewViewModel?
```
stefanceriu, line 21: Talked about this offline: let's try to pass the immutable Context around instead of the whole VM

stefanceriu (approved):
I mostly skimmed it because a lot of the code moved around but it lgtm, nothing jumped out 👍

PR #3720, opened 2025-01-31
Disable `continueAfterFailure` for the AppLock UI tests

stefanceriu:


UITests/Sources/AppLockUITests.swift
pixlwave, line 1: Do we need it on these?

pixlwave (approved):

PR #3722, opened 2025-01-31
Ensure multiple mandatory verification flows can be ran consecutively (e.g. following encryption resets)

stefanceriu:
* running multiple remote resets would find the onboarding flow coordinator in a finished state and prevent showing the mandatory verification again. Fixed that by transitioning from finished to initial automatically
* added onboarding flow coordinator state machine logs
* removed security state deduplication and added extra main queue dispatches

### DIFF
--- a/ElementX/Sources/FlowCoordinators/OnboardingFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/OnboardingFlowCoordinator.swift
@@ -38,7 +38,7 @@ class OnboardingFlowCoordinator: FlowCoordinatorProtocol {
     
     enum Event: EventType {
         case next
-        case nextSkippingIdentityConfimed
+        case nextSkippingIdentityConfirmed
     }
     
     private let stateMachine: StateMachine<State, Event>
@@ -79,6 +79,8 @@ class OnboardingFlowCoordinator: FlowCoordinatorProtocol {
         
         stateMachine = .init(state: .initial)
         
+        configureStateMachine()
+        
         // Verification can change as part of the onboarding flow by verifying with
         // another device, using a recovery key or by resetting one's crypto identity.
         // It can also happen that onboarding started before it had a chance to update,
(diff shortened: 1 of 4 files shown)


ElementX/Sources/Screens/Onboarding/SessionVerificationScreen/View/SessionVerificationScreen.swift
pixlwave, line 1: Ok I just looked, I definitely meant the `IdentityConfirmedScreen` here, as in when the whole thing is complete.

pixlwave (changes requested):
LGTM, but I think it hides the button on the wrong screen.

ElementX/Sources/FlowCoordinators/OnboardingFlowCoordinator.swift
```diff
@@ -79,22 +79,23 @@ class OnboardingFlowCoordinator: FlowCoordinatorProtocol {
...
-            .removeDuplicates()
+            .receive(on: DispatchQueue.main)
             .sink { [weak self] value in
                 guard let self,
                       value == .verified,
                       stateMachine.state == .identityConfirmation else { return }
                 
                 appSettings.hasRunIdentityConfirmationOnboarding = true
-                stateMachine.tryEvent(.nextSkippingIdentityConfimed)
-                self.verificationStateCancellable = nil
```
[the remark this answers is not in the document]
stefanceriu: Yep, I tested it and looks fine without that.

ElementX/Sources/Screens/Onboarding/SessionVerificationScreen/View/SessionVerificationScreen.swift
stefanceriu: Talked about it offline but the gist is that we need to do this in the session verification screens as well. Decided to only do it for the `.verified` state for now

pixlwave (approved):

PR #3731, opened 2025-02-03
Refactor Rust timeline identifiers into our own.

pixlwave:
In preparation for swiping media in the room screen, this PR replaces all the uses of Rust's `EventOrTransactionId` and `TimelineUniqueId` with our own local versions as it seems silly to keep importing the SDK into UI layers just to identify things.

Broken into 2 commits, one for each refactor so it should be easier to review.

### DIFF
--- a/ElementX/Sources/Services/Timeline/TimelineItemIdentifier.swift
+++ b/ElementX/Sources/Services/Timeline/TimelineItemIdentifier.swift
@@ -6,7 +6,7 @@
 //
 
 import Foundation
-import MatrixRustSDK
+@preconcurrency import MatrixRustSDK
 
 /// A timeline item identifier
 /// - uniqueID: Stable id across state changes of the timeline item, it uniquely identifies an item in a timeline.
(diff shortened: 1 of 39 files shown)


ElementX/Sources/Services/Timeline/TimelineItemIdentifier.swift
```diff
@@ -6,61 +6,82 @@
...
-        switch self {
-        case .event(let uniqueID, _):
-            return uniqueID
-        case .virtual(let uniqueID):
-            return uniqueID
+    struct UniqueID: Hashable {
+        let value: String
+        
+        init(_ value: String) {
+            self.value = value
```
Velin92, line 21: maybe instead of defining a struct which comes at the cost of defining also the init and this way of converting the types, and most importantly the thing i don't like the most, doing stuff like `uniqueID.value`... couldn't instead use a typealias `UniqueID` for the `String` type? the typealas can also have some dedicated extension to convert from the rust type and viceversa, but comes with the benefit that it can easily be used as also a normal String, without the need of calling a value. It would also be automatically an Hashable and Equatable.

Velin92 (changes requested):
Couldn't we use a typealias of String with some methods on top, instead of defining a struct with a single value?

ElementX/Sources/Services/Timeline/TimelineItemIdentifier.swift
pixlwave: I discussed this one with Stefan before-hand and we decided on keeping the struct so that if that `TimelineUniqueId` ever changes on the Rust side we can simply wrap it in here and not have to do a massive refactor in the codebase on an SDK update.

Velin92 (approved):

PR #3732, opened 2025-02-03
Knocking polishing part 1

Velin92:
- using joined members count for the preview
- listen to the membership state change when you are in the knocked state

### DIFF
--- a/ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenModels.swift
+++ b/ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenModels.swift
@@ -29,7 +29,7 @@ struct JoinRoomScreenRoomDetails {
     let topic: String?
     let canonicalAlias: String?
     let avatar: RoomAvatar
-    let memberCount: UInt
+    let memberCount: Int
     let inviter: RoomInviterDetails?
 }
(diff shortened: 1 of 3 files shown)


ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenModels.swift
pixlwave, line 1: I wonder if it might make more sense to keep this as `UInt` and update the `RoomInfoProxy` to do the right thing here too seeing as all the counts come out of the SDK as `UInt`.

pixlwave (approved):
👍

ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenModels.swift
Velin92: I would prefer to always use `Int` on our side since from the swift guide:
> Use UInt only when you specifically need an unsigned integer type with the same size as the platform’s native word size. If this is not the case, Int is preferred, even when the values to be stored are known to be non-negative. A consistent use of Int for integer values aids code interoperability, avoids the need to convert between different number types, and matches integer type inference, as described in Type Safety and Type Inference.

PR #3738, opened 2025-02-04
Knocking feature polishing part 2

Velin92:
- Added an error state when joining fails due to forbidden access
- Improved the presentation of invites for DMs
- Forget action for banned room previews implemented


ElementX/Sources/Screens/JoinRoomScreen/View/JoinRoomScreen.swift
```diff
@@ -296,6 +297,7 @@ struct JoinRoomScreen_Previews: PreviewProvider, TestablePreview {
         makePreview(viewModel: restrictedViewModel, previewDisplayName: "Restricted")
         makePreview(viewModel: inviteRequiredViewModel, previewDisplayName: "InviteRequired")
         makePreview(viewModel: invitedViewModel, previewDisplayName: "Invited")
+        makePreview(viewModel: dmInviteViewModel, previewDisplayName: "DM Invite")
```
stefanceriu, line 300: Can we use `Invited DM` & `invitedDM` here and in the mocks so it's consistend with the existing one?

ElementX/Sources/Services/Room/RoomProxyProtocol.swift
```diff
@@ -45,6 +45,12 @@ protocol KnockedRoomProxyProtocol: RoomProxyProtocol {
     func cancelKnock() async -> Result<Void, RoomProxyError>
 }
 
+// sourcery: AutoMockable
+protocol BannedRoomProxyProtocol: RoomProxyProtocol {
+    var info: BaseRoomInfoProxyProtocol { get }
+    func forgetRoom() async -> Result<Void, RoomProxyError>
+}
+
```
stefanceriu: This feels like a good idea. I remember adding it myself and then removing it later, same as you, but I can't remember the reason behind it. Is is that the sdk can't provide any room details for banned rooms?

Velin92: the problem is that you can't get the preview from the room list item since it filters banned rooms.

stefanceriu: Not sure we're thinking about the same thing. I'm thinking of using the `BannedRoomProxyProtocol` as an associated value on the `Client.roomForIdentifier` return value 🤔

ElementX/Sources/Screens/JoinRoomScreen/View/JoinRoomScreen.swift
```diff
@@ -161,13 +161,21 @@ struct JoinRoomScreen: View {
         case .loading:
             EmptyView()
         case .joinable:
-            joinButton
-        case .unknown, .restricted: // If unknown, do our best.
-            VStack(spacing: 24) {
-                bottomNoticeMessage(L10n.screenJoinRoomJoinRestrictedMessage)
-                
+            if context.viewState.shouldShowForbiddenError {
```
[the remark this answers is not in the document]
Velin92: Design wants us to show this forbidden view when the join fails for a `forbidden` reason. Maybe I could make an additional forbidden state, but felt was not really worth doing it, since this is more like and alert flag than a specific state of the view, but I can try if things get easier with an associated value or a dedicated state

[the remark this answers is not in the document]
stefanceriu: Alrighty fair enough 👍

ElementX/Sources/Services/Room/RoomProxyProtocol.swift
stefanceriu: Even better, yeah! 👍

stefanceriu (approved):
Very nice, 🚀 it!

PR #3739, opened 2025-02-04
Bottom Sheet to confirm DM creation

### LINKED ISSUE #3729: [Task] Confirmation alert when creating a new DM
While browsing the profile of another user, if you tap the "Message" action you should either go directly to the DM if it exist, or get a confirmation alert asking if you wish to create a new one, instead of directly opening it.
Reference [here](https://github.com/element-hq/customer-success/issues/265)

Velin92:
Replacing the alert with a nice bottom sheet.
Also adding the confirmation, when selecting a user in the start chat screen.

fixes #3729 

https://github.com/user-attachments/assets/bda88efc-4dd1-4dde-a783-5a15ab4b0c8c


ElementX/Sources/Screens/StartChatScreen/View/SendInviteConfirmationView.swift
```diff
@@ -0,0 +1,95 @@
...
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Compound
+import SwiftUI
+
+struct UserToInvite: Identifiable {
+    let avatarUrl: URL?
```
pixlwave, line 12:
```suggestion
    let avatarURL: URL?
```

ElementX/Sources/Screens/StartChatScreen/View/SendInviteConfirmationView.swift
```diff
@@ -0,0 +1,95 @@
...
+// Copyright 2025 New Vector Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Compound
+import SwiftUI
+
+struct UserToInvite: Identifiable {
```
pixlwave, line 11: Actually, wait… Can't we use the already existing `UserProfileProxy` here? It represents the exact same thing.

ElementX/Sources/Screens/StartChatScreen/View/SendInviteConfirmationView.swift
```diff
@@ -0,0 +1,95 @@
...
+                                name: userToInvite.displayName,
+                                contentID: userToInvite.id,
+                                avatarSize: .user(on: .sendInviteConfirmation),
+                                mediaProvider: mediaProvider)
+            VStack(spacing: 8) {
+                Text(L10n.screenBottomSheetCreateDmTitle)
+                    .multilineTextAlignment(.center)
+                    .font(.compound.headingMDBold)
+                    .foregroundStyle(.compound.textPrimary)
+                Text(L10n.screenBottomSheetCreateDmMessage(userToInvite.displayName ?? "", userToInvite.id))
```
pixlwave, line 56: Having a `nil` displayname is a real thing. We should either
- Create a different string that we can use in this instance.
- Grab the localpart of the displayname out of the user ID and use that in place of the displayname (like we do in `RoomScreenFooterView`).

ElementX/Sources/Screens/StartChatScreen/StartChatScreenModels.swift
```diff
@@ -37,10 +37,13 @@ struct StartChatScreenViewStateBindings {
...
     /// Information describing the currently displayed alert.
     var alertInfo: AlertInfo<StartChatScreenErrorType>?
+    
+    var selectedUserToInvite: UserToInvite?
 }
 
 enum StartChatScreenViewAction {
     case close
     case createRoom
+    case createDM(userID: String, displayName: String?)
```
pixlwave, line 47: We can also pass the `UserProfileProxy` in here instead of breaking it up.

ElementX/Sources/Screens/StartChatScreen/StartChatScreenViewModel.swift
```diff
@@ -107,12 +110,12 @@ class StartChatScreenViewModel: StartChatScreenViewModelType, StartChatScreenVie
         }
     }
         
-    private func createDirectRoom(with user: UserProfileProxy) async {
+    private func createDirectRoom(userID: String, displayName: String?) async {
```
pixlwave, line 113: And then this wouldn't need to change :)

pixlwave (changes requested):
LGTM, but can we use the `UserProfileProxy` instead please. It is due a refactor (`userID` -> `id` so it can be identifiable, drop the `Proxy` suffix as it doesn't proxy anything etc) so feel free to do that if it helps.

ElementX/Sources/Screens/RoomMemberDetailsScreen/RoomMemberDetailsScreenModels.swift
```diff
@@ -73,7 +73,7 @@ struct RoomMemberDetailsScreenViewStateBindings {
     
     var ignoreUserAlert: IgnoreUserAlertItem?
     var alertInfo: AlertInfo<RoomMemberDetailsScreenAlertType>?
-    var sheetItem: UserToInvite?
+    var sheetItem: UserProfileProxy?
```
pixlwave, line 76: Maybe this should be called `inviteConfirmationUser` or something a bit more descriptive.

ElementX/Sources/Screens/UserProfileScreen/UserProfileScreenModels.swift
```diff
@@ -36,7 +36,7 @@ struct UserProfileScreenViewState: BindableState {
 
 struct UserProfileScreenViewStateBindings {
     var alertInfo: AlertInfo<UserProfileScreenAlertType>?
-    var sheetItem: UserToInvite?
+    var sheetItem: UserProfileProxy?
```
pixlwave, line 39: Same here.

pixlwave (approved):
LGTM, thanks for the tweaks 👍

PR #3741, opened 2025-02-05
Fix some concurrency warnings, update missed licence headers.

stefanceriu:


ElementX/Sources/Services/Timeline/TimelineController/TimelineController.swift
```diff
@@ -386,7 +386,9 @@ class TimelineController: TimelineControllerProtocol {
     @objc private func contentSizeCategoryDidChange() {
         // Recompute all attributed strings on content size changes -> DynamicType support
         serialDispatchQueue.async { [activeTimelineProvider] in
-            self.updateTimelineItems(itemProxies: activeTimelineProvider.itemProxies, paginationState: activeTimelineProvider.paginationState)
+            MainActor.assumeIsolated {
+                self.updateTimelineItems(itemProxies: activeTimelineProvider.itemProxies, paginationState: activeTimelineProvider.paginationState)
+            }
```
pixlwave, line 391: Not sure I'm for this one. Instead we should try replacing the serial dispatch queue with an async sequence as the signal to update the items.

pixlwave (commented):
Overall lgtm, except for that assumeIsolated.

ElementX/Sources/Application/AppCoordinatorProtocol.swift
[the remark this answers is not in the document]
pixlwave: Oh interesting, TIL.

ElementX/Sources/Services/Timeline/TimelineController/TimelineController.swift
stefanceriu: Talked about it offline: there is indeed a simpler way! 🎉

ElementX/Sources/Other/UserIndicator/UserIndicatorController.swift
```diff
@@ -75,10 +77,11 @@ class UserIndicatorController: ObservableObject, UserIndicatorControllerProtocol
             indicatorQueue.removeAll { $0.id == id }
             return
         }
-    
-        Timer.scheduledTimer(withTimeInterval: minimumDisplayDuration, repeats: false) { [weak self] _ in
-            self?.indicatorQueue.removeAll { $0.id == id }
-            self?.displayTimes[id] = nil
+        
+        Task {
+            try await Task.sleep(for: .seconds(minimumDisplayDuration))
```
pixlwave, line 82: Ok thinking about this a bit more. If you cancel a task that spawned another task (e.g. this one), then that task will also be cancelled. So this one should probably `try?` as if anything ever cancels the task, we still want the indicator to be removed.

ElementX/Sources/Services/Timeline/TimelineController/TimelineController.swift
```diff
@@ -375,57 +375,60 @@ class TimelineController: TimelineControllerProtocol {
...
-        updateTimelineItemsCancellable = activeTimelineProvider
-            .updatePublisher
-            .receive(on: serialDispatchQueue)
-            .sink { [weak self] items, paginationState in
-                self?.updateTimelineItems(itemProxies: items, paginationState: paginationState)
+        updateTimelineItemsCancellable = Task { [weak self, activeTimelineProvider] in
+            let publisher = activeTimelineProvider.updatePublisher
+            let notification = NotificationCenter.default.publisher(for: UIContentSizeCategory.didChangeNotification)
+            
+            let whatever = publisher.merge(with: notification.map { _ in (activeTimelineProvider.itemProxies, activeTimelineProvider.paginationState) })
```
pixlwave, line 382: whatever-3983078853

Pointing it out just in case you forgot to rename these.

stefanceriu: Most certainly did 😂

ElementX/Sources/Other/UserIndicator/UserIndicatorController.swift
stefanceriu: Makes sense 👍

pixlwave (approved):

PR #3743, opened 2025-02-05
Updated Bottom Sheet message string

Velin92:
The display name and the id pair is now localised with its own string that is reusable, and we use it when the display name is available, otherwise we just replace it with the user id.


stefanceriu (approved):
Oh I'm sorry, I completely missed this pr 😞

PR #3744, opened 2025-02-05
Added the banned room proxy

Velin92:
- This PR updates the SDK to allow a banned room proxy to be generated from the room list item
- It allows to have `forget` be used specifically only for banned rooms
- It also includes an alert to retry loading the room in case it fails for a connection issue


pixlwave (approved):
👍

PR #3746, opened 2025-02-06
Fix a crash in the media browser by storing the active timeline context.

pixlwave:
That `guard let self else { fatalError() }` was being hit when the screen was dismissed so isn't going to fly 😅

This PR stores the context in the state instead, updating it whenever the user changes the screen mode.


stefanceriu (approved):
Makes sense to me 👍

PR #3748, opened 2025-02-06
Added a delayed loading when opening a room

Velin92:
While this won't even appear for joined rooms, this will help greatly for invited and knocked rooms which require the preview API to load first

### DIFF
--- a/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
@@ -229,10 +229,12 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
     private func handleRoomRoute(roomID: String, via: [String], presentationAction: PresentationAction? = nil, animated: Bool) async {
         guard roomID == self.roomID else { fatalError("Navigation route doesn't belong to this room flow.") }
         
+        showLoadingIndicator(delay: .milliseconds(250))
         guard let room = await userSession.clientProxy.roomForIdentifier(roomID) else {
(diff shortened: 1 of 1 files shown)


ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -1618,6 +1620,16 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
         
         flowCoordinator.start()
     }
+    
+    private static let loadingIndicatorID = "RoomFlowCoordinator.loadingIndicator"
```
stefanceriu, line 1624: We started doing `"\(WhateverClass.self)-Loading"` in other places to avoid problems when renaming, we should do the same here.

`private static let loadingIndicatorIdentifier = "\(RoomFlowCoordinator.self)-Loading"`

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -1618,6 +1620,16 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
...
+    
+    private func showLoadingIndicator(delay: Duration? = nil) {
+        userIndicatorController.submitIndicator(.init(id: Self.loadingIndicatorID, type: .modal(progress: .indeterminate, interactiveDismissDisabled: false, allowsInteraction: false), title: L10n.commonLoading, persistent: true), delay: delay)
```
stefanceriu, line 1627: Let's wrap this to multiple lines

stefanceriu (approved):
Couple of nits otherwise looks good to me 👍

PR #3750, opened 2025-02-06
Fix a bug where the preview controller breaks when swiping quickly.

pixlwave:
<del>This is clearly going to break at some point, but for now, it's a fix 🙈</del>

The updated solution uses recursion, it could still break but it's likely at that point so will the navbar and toolbar.

### DIFF
--- a/ElementX/Sources/Screens/FilePreviewScreen/View/TimelineMediaPreviewController.swift
+++ b/ElementX/Sources/Screens/FilePreviewScreen/View/TimelineMediaPreviewController.swift
@@ -25,7 +25,7 @@ class TimelineMediaPreviewController: QLPreviewController {
     
     private var navigationBar: UINavigationBar? { view.subviews.first?.subviews.first { $0 is UINavigationBar } as? UINavigationBar }
     private var toolbar: UIToolbar? { view.subviews.first?.subviews.last { $0 is UIToolbar } as? UIToolbar }
-    private var pageScrollView: UIScrollView? { view.subviews.first?.subviews.first?.subviews.first?.subviews.first?.subviews.first?.subviews.first as? UIScrollView }
+    private var pageScrollView: UIScrollView? { view.firstScrollView() }
     private var captionView: UIView { captionHostingController.view }
     
     override var overrideUserInterfaceStyle: UIUserInterfaceStyle {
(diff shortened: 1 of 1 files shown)


ElementX/Sources/Screens/FilePreviewScreen/View/TimelineMediaPreviewController.swift
```diff
@@ -25,6 +25,7 @@ class TimelineMediaPreviewController: QLPreviewController {
     
     private var navigationBar: UINavigationBar? { view.subviews.first?.subviews.first { $0 is UINavigationBar } as? UINavigationBar }
     private var toolbar: UIToolbar? { view.subviews.first?.subviews.last { $0 is UIToolbar } as? UIToolbar }
+    private var pageScrollView: UIScrollView? { view.subviews.first?.subviews.first?.subviews.first?.subviews.first?.subviews.first?.subviews.first as? UIScrollView }
```
stefanceriu, line 28: Should we maybe add a debug assertion here or something?

stefanceriu: On a slightly different note: wouldn't it be generally safer/more reliable long term to use generic subview recursion to find the first `UIScrollView` and return that?

ElementX/Sources/Screens/FilePreviewScreen/View/TimelineMediaPreviewController.swift
```diff
@@ -184,6 +185,13 @@ class TimelineMediaPreviewController: QLPreviewController {
     
     private func handleFileLoaded(itemID: TimelineItemIdentifier.EventOrTransactionID) {
         guard (currentPreviewItem as? TimelineMediaPreviewItem.Media)?.id == itemID else { return }
+        
+        // There's a bug where refreshCurrentPreviewItem completely breaks the QLPreviewController
+        // if it's called whilst swiping between items. So don't let that happen.
```
stefanceriu, line 190: That last sentence gives me plumber vibes 😅

stefanceriu (approved):

ElementX/Sources/Screens/FilePreviewScreen/View/TimelineMediaPreviewController.swift
pixlwave: Went with the recursion.

PR #3779, opened 2025-02-10
Knock Polishing part 4

Velin92:
- added a decline and block button
- redesigned the inviter to be at the top
- some changes in the top padding behaviour


pixlwave (approved):
LGTM 🎉

PR #3691, opened 2025-01-21
Image magick replacement + app variants

stefanceriu:
* setup an app variants folder and move the nightly and alpha build configs to it
* replace ImageMagick with a custom tool written in swift

PR #3789, opened 2025-02-13
Update the Enterprise submodule.

pixlwave:
Preparation for the pipeline.


fastlane/README.md
pixlwave, line 1: Unrelated to this change but it was missed off in #3691.

stefanceriu (approved):
nice, lgtm 👍

PR #3796, opened 2025-02-16
Stop the message composer from randomly changing the cursor position

stefanceriu:
Check if the selection range is different at the same time as reading it not before being dispatched to the next main queue loop.


pixlwave (approved):
LGTM, thanks.

stefanceriu:
cc @vickcoo

PR #3801, opened 2025-02-18
Remove support for the sliding sync proxy.

pixlwave:
The SDK no longer supports a sliding sync proxy, so this update will force logout anyone using a sliding sync proxy:

https://github.com/user-attachments/assets/48b60070-ade1-4b0e-b2ed-837a3d14b147


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -263,6 +261,11 @@ class ClientProxy: ClientProxyProtocol {
     }
 
     func startSync() {
+        guard !needsSlidingSyncMigration else {
+            MXLog.warning("Ignoring request, this client needs to be migrated to native sliding sync.")
+            return
+        }
+
```
pixlwave, line 268: Not sure if this was strictly necessary but seemed worthwhile to me.

ElementX/Sources/Other/Extensions/ClientBuilder.swift
```diff
@@ -58,12 +56,8 @@ extension ClientBuilder {
 }
 
 enum ClientBuilderSlidingSync {
```
stefanceriu, line 58: Are we sure we want to keep this around?

ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu: Makes perfect sense to me 👍

ElementX/Sources/Services/UserSession/RestorationToken.swift
```diff
@@ -38,36 +45,60 @@ extension RestorationToken: Codable {
...
         try container.encode(sessionDirectories.dataDirectory, forKey: .sessionDirectory)
         try container.encode(sessionDirectories.cacheDirectory, forKey: .cacheDirectory)
         try container.encode(passphrase, forKey: .passphrase)
         try container.encode(pusherNotificationClientIdentifier, forKey: .pusherNotificationClientIdentifier)
     }
 }
 
+/// Temporary struct to smooth the forced migration by keeping a user session.
+/// In the future we can remove this and throw a migration error when the URL
+/// is decoded to a non-nil value.
```
stefanceriu, line 70: 👌

stefanceriu (approved):
Love a good (almost) spring cleanup!

ElementX/Sources/Other/Extensions/ClientBuilder.swift
pixlwave: We discussed offline, there's a couple of cases that are both used here.

PR #3809, opened 2025-02-20
Render Room and Message Pills

### LINKED ISSUE #2633: [Task] Render permalinks in the timeline

Velin92:
Now is possible to generate pills in the timeline for 
- Any room permalink
- Any room event permalink
- Room addresses

fixes #2633


ElementX/Sources/Screens/RoomScreen/RoomScreenCoordinator.swift
```diff
@@ -86,7 +86,8 @@ final class RoomScreenCoordinator: CoordinatorProtocol {
                                               appSettings: parameters.appSettings,
                                               analyticsService: ServiceLocator.shared.analytics,
                                               emojiProvider: parameters.emojiProvider,
-                                              timelineControllerFactory: parameters.timelineControllerFactory)
+                                              timelineControllerFactory: parameters.timelineControllerFactory,
+                                              roomListPublisher: parameters.clientProxy.roomSummaryProvider?.roomListPublisher)
```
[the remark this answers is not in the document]
Velin92: I'll test this, what is the difference between the two btw? Since this could also affect the JoinRoomScreen.

I don't see alternatives that do not involve using API calls which I feel we should avoid

ElementX/Sources/Other/Pills/MentionBuilder.swift
```diff
@@ -9,13 +9,20 @@ import Foundation
 import UIKit
 
 struct MentionBuilder: MentionBuilderProtocol {
```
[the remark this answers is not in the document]
Velin92: Already reworked most of the code to do this, will try if I can squeeze some more

ElementX/Sources/Screens/RoomScreen/RoomScreenCoordinator.swift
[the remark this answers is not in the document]
pixlwave: How about we abstract the summary provider out of both of these and add a `ClientProxy.roomSummaryForIdentifier` that handles this internally itself? It seems reasonable for the timeline to know about the ClientProxy anyway to me.

Would just need some documentation to say
- `roomForID` is for the full room, building a timeline ready for use.
- `roomSummaryForID` gets the details about known rooms.
- `roomPreviewForID` attempts to fetch a preview of any room from the server.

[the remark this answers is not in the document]
stefanceriu: Oh nice, yeah, I like that. It's easy enough to adopt both here and in the join room screen but it also hides away all the nasty bits and gives us one single failure point 👍

ElementX/Sources/Services/Client/ClientProxyProtocol.swift
pixlwave, line 1: I think we should add some doc comments here too. We know what the difference between a `room` and `roomPreview` and a `roomSummary` is and when to ask for each one, but someone new to the project probably won't know.

pixlwave (commented):
Can we use the new APIs in the JoinRoomScreen now too? i.e. to hide the publisher from #3817.

ElementX/Sources/Other/Pills/PillView.swift
```diff
@@ -22,43 +23,71 @@ struct PillView: View {
...
+            Text(context.viewState.displayText)
+                .font(.compound.bodyLGSemibold)
+                .foregroundColor(textColor)
+                .lineLimit(1)
+        }
+        .padding(.leading, 4)
+        .padding(.trailing, 6)
+        .padding(.vertical, 1)
+        .background { Capsule().foregroundColor(backgroundColor)
+        }
```
pixlwave, line 45:
```suggestion
        .background { Capsule().foregroundColor(backgroundColor) }
```

orrrrr…

```suggestion
        .background(backgroundColor, in: Capsule())
```

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FormattedBodyText.swift
```diff
@@ -142,8 +142,6 @@ struct FormattedBodyText_Previews: PreviewProvider, TestablePreview {
         let htmlStrings = [
             """
             Plain text\n
-            @bob:matrix.org\n
-            #room:matrix.org\n
```
pixlwave, line 146: Were these rendering as pills?

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -515,6 +515,16 @@ class ClientProxy: ClientProxyProtocol {
...
     }
+    
+    func roomSummaryForIdentifier(_ identifier: String) -> RoomSummary? {
+        // the alternate room summary provider is not impacted by filtering
+        alternateRoomSummaryProvider?.roomListPublisher.value.first(where: { $0.id == identifier })
+    }
+    
+    func roomSummaryForAlias(_ alias: String) -> RoomSummary? {
+        // the alternate room summary provider is not impacted by filtering
+        alternateRoomSummaryProvider?.roomListPublisher.value.first(where: { $0.canonicalAlias == alias })
```
pixlwave, line 526: Do you think we should we support all aliases here?

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
```diff
@@ -826,6 +834,72 @@ class TimelineViewModel: TimelineViewModelType, TimelineViewModelProtocol {
         actionsSubject.send(.displayMessageForwarding(forwardingItem: .init(id: itemID, roomID: roomProxy.id, content: content)))
     }
     
+    // MARK: Pills
+    
+    private func pillContextUpdater(_ pillContext: PillContext) {
```
pixlwave, line 839: Nice, this worked well in the end 👏

pixlwave (approved):
Doing a complete review and looks great to me! 👏

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FormattedBodyText.swift
Velin92: yup

ElementX/Sources/Services/Client/ClientProxy.swift
Velin92: oh yeah good point but do we have them in the RoomSummary? I'll check tomorrow

Velin92:
> Can we use the new APIs in the JoinRoomScreen now too? i.e. to hide the publisher from #3817.

we can't because that function is reactive subscription, while the API is just a check/get

PR #3813, opened 2025-02-23
Show error screens in group calls

robintown:
Element Call now sends a 'close' widget action when the widget is ready to close. Usually this will be sent immediately after the 'hangup' action, but it could be sent later if the widget wants to present an error screen before closing. So by listening to the 'close' action rather than the 'hangup' action, we get to see these error screens.

For https://github.com/element-hq/element-call/issues/2955
Depends on https://github.com/element-hq/element-call/pull/3011

### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
+++ b/ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
@@ -17,6 +17,7 @@ struct ElementCallWidgetMessage: Codable {
     
     enum Action: String, Codable {
         case hangup = "im.vector.hangup"
+        case close = "io.element.close"
         case mediaState = "io.element.device_mute"
     }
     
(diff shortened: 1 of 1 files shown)


robintown:
Looks like the CI isn't happy with me working from a fork. Should I request write access from IT and make a new branch, or is there some other fix?

stefanceriu:
No, it's fine, it's expected for the enterprise tests to fail in this case and the normal unit tests passed. 

Danger is failing because you're missing a `pr-` tag but it's supposed to print out an error message, no idea what's happening there.

hughns (approved):
As requested out of band, I have followed the instructions from https://github.com/element-hq/element-x-android/pull/4297

And I can confirm that the Element Call screen continued to be shown and looked like this:

IMG_6742

Furthermore, I confirmed that the WebView was still running by using the Safari Web Inspector connected to the test device.

pixlwave (approved):
SGTM then!

PR #3819, opened 2025-02-24
Revert "Show error screens in group calls"

hughns:
Reverts element-hq/element-x-ios#3813

The original PR breaks UX on Element Call <=0.7.2 so we don't want to merge it yet.


pixlwave (approved):
Sorry my bad. I didn't realise it wasn't ready to merge.

PR #3820, opened 2025-02-24
Configure diagnostics (and MapLibre) using Pkl.

pixlwave:
Rather than using the secrets.xcconfig to inject values into the Info.plist, which was a very roundabout way to inject secrets, this PR uses our new favourite language Pkl to generate (with the help of SwiftGen) a Secrets enum directly in the app so that safe access is guaranteed at build time.
*caveat:* Safe for strings anyway, it could still crash at runtime when given a malformed URL.

Additionally this PR bumps the release version ready for tomorrow's RC.


Secrets/Secrets.swift
```diff
@@ -0,0 +1,19 @@
...
+// MARK: - YAML Files
+
+// swiftlint:disable identifier_name line_length number_separator type_body_length
+internal enum Secrets {
+  internal static let mapLibreAPIKey: String = "your_key"
+  internal static let postHogAPIKey: String = "your_key"
+  internal static let postHogHost: String = "https://get.your.own.io"
+  internal static let rageshakeServerURL: String = "https://⁦bring.your.own.and.en.joy"
+  internal static let sentryDSN: String = "https://⁦insert.your.hearts.desire.here"
+}
```
Velin92, line 18: there is something going on here with weirdi bidi unicode characters

pixlwave: lol that's what I get for copying from EX and accidentally selecting the bit before the timestamp 🙈

Velin92 (approved):

PR #3821, opened 2025-02-24
Don't set the room topic when creating a room if it is blank.

pixlwave:
Fixes the following bug:

>  If I create a room from EXI, all apps (EXA, EXI and EW) display “you deleted the room topic”.

Whilst here I tweaked the styling on the room name text field as it had the wrong colours and the font wasn't explicitly specified.


PreviewTests/Sources/__Snapshots__/PreviewTests/test_createRoom-iPhone-16-en-GB.Create-Public-Room-existing-alias.png
stefanceriu, line 1: I wonder if we should change these to `.previewLayout(.sizeThatFits)` 🤔

stefanceriu (approved):
Nice, lgtm 👍

PreviewTests/Sources/__Snapshots__/PreviewTests/test_createRoom-iPhone-16-en-GB.Create-Public-Room-existing-alias.png
pixlwave: Yeah, sadly that doesn't work when there's a scroll view.

PR #3824, opened 2025-02-25
Show DM recipient verification badges on the room details screen profile button

stefanceriu:
See updated snapshots for details.

### DIFF
--- a/ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenModels.swift
+++ b/ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenModels.swift
@@ -78,6 +78,8 @@ struct RoomDetailsScreenViewState: BindableState {
     var bindings: RoomDetailsScreenViewStateBindings
 
     var dmRecipient: RoomMemberDetails?
+    var dmRecipientVerificationState: UserIdentityVerificationState?
+    
     var accountOwner: RoomMemberDetails?
     
     var shortcuts: [RoomDetailsScreenViewShortcut] {
--- a/ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenViewModel.swift
+++ b/ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenViewModel.swift
@@ -234,11 +234,21 @@ class RoomDetailsScreenViewModel: RoomDetailsScreenViewModelType, RoomDetailsScr
             .receive(on: DispatchQueue.main)
             .sink { [weak self, ownUserID = roomProxy.ownUserID] members in
                 guard let self else { return }
+                
                 let accountOwner = members.first { $0.userID == ownUserID }
                 let dmRecipient = members.first { $0.userID != ownUserID }
+                
                 self.dmRecipient = dmRecipient
                 self.state.dmRecipient = dmRecipient.map(RoomMemberDetails.init(withProxy:))
                 self.state.accountOwner = accountOwner.map(RoomMemberDetails.init(withProxy:))
+                
+                if let dmRecipient {
+                    Task { [weak self] in
+                        if case let .success(userIdentity) = await self?.clientProxy.userIdentity(for: dmRecipient.userID) {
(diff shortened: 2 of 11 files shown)


ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenModels.swift
```diff
@@ -78,6 +78,8 @@ struct RoomDetailsScreenViewState: BindableState {
     var bindings: RoomDetailsScreenViewStateBindings
 
     var dmRecipient: RoomMemberDetails?
+    var dmRecipientVerificationState: UserIdentityVerificationState?
```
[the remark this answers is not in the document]
stefanceriu: Will do, good idea 👍

ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenViewModel.swift
```diff
@@ -234,11 +234,21 @@ class RoomDetailsScreenViewModel: RoomDetailsScreenViewModelType, RoomDetailsScr
...
+                
                 self.dmRecipient = dmRecipient
                 self.state.dmRecipient = dmRecipient.map(RoomMemberDetails.init(withProxy:))
                 self.state.accountOwner = accountOwner.map(RoomMemberDetails.init(withProxy:))
+                
+                if let dmRecipient {
+                    Task { [weak self] in
+                        if case let .success(userIdentity) = await self?.clientProxy.userIdentity(for: dmRecipient.userID) {
+                            self?.state.dmRecipientVerificationState = userIdentity?.verificationState
+                        }
```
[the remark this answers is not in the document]
stefanceriu: Bah, I just realised that this thing should update on changes coming from the `identityStatusChangesPublisher`, similar to the `hasMemberIdentityVerificationStateViolations`. I need to figure out some other way to write all of this.

ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenViewModel.swift
```diff
@@ -251,16 +255,25 @@ class RoomDetailsScreenViewModel: RoomDetailsScreenViewModelType, RoomDetailsScr
...
-        for member in roomProxy.membersPublisher.value {
-            if case let .success(identity) = await clientProxy.userIdentity(for: member.userID) {
-                if identity?.verificationState == .verificationViolation {
-                    state.hasMemberIdentityVerificationStateViolations = true
-                    return
+        if roomProxy.isDirectOneToOneRoom {
+            if var dmRecipientInfo = state.dmRecipientInfo {
+                if case let .success(userIdentity) = await clientProxy.userIdentity(for: dmRecipientInfo.member.id) {
+                    dmRecipientInfo.verificationState = userIdentity?.verificationState
+                    state.dmRecipientInfo = dmRecipientInfo
```
Velin92, line 262: way better! Maybe instead of overriding the whole `state.dmRecipientInfo = dmRecipientInfo`
We could just add to `DMRecipientInfo` a mutating func like setVerificationState, or even simpler just make it a var

Velin92 (approved):
way better just 1 more suggestion

ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenViewModel.swift
stefanceriu: I would need to make it an observable object in that case, no? Plus a similar thing is done for updating the member after ignoring and unignoring the user.

stefanceriu: Talked about it offline. We'll leave it like this because of the crashes we've seen on release builds in other places where mutating an optional state property.

stefanceriu: For posterity, the crash still happens on release builds:
`Swift/KeyPath.swift:2838: Fatal error: could not demangle keypath type from '��#`

PR #3826, opened 2025-02-25
Cache account management URL

### LINKED ISSUE #3825: [Task] Preload account management URLs
A memory cache has been implemented in the SDK. See https://github.com/matrix-org/matrix-rust-sdk/pull/4663

We need the application to pre-load the 2 accounts management URLs to populate the in-memory cache. That way, when the user open the settings, the flickering will not be observed for the two entries:

Image

Note that if the app is started without network access, the url cannot be loaded and the settings entries will not be displayed.

There is no change expected on the setting screen, that should still requires the URLs from the SDK.

Velin92:
fixes #3825


stefanceriu (approved):
Talked it over offline, looks sensible to me 👍

PR #3833, opened 2025-02-26
Fix editing messages not placing the cursor at the end of the text

### LINKED ISSUE #3830: Edits no longer position cursor at end of msg
### Steps to reproduce

1. Send a msg
2. Edit it
3. Get disoriented that the cursor is at the start of the composer; everywhere else the cursor is at the end of the msg - ie where you last where when you hit send/enter, not randomly at the beginning which is probably the least likely bit to be edited, and from which you can’t hit delete to rewrite the msg

#### What did you expect?

cursor to be where i left it when i edit a msg (or failing that, the end of the msg, which is likely where i left it)

#### What happened instead?

cursor at the beginning. much disorientation.

### Your phone model

_No response_

### Operating system version

18.3.1

### Application version

847

### Homeserver

_No response_

### Will you send logs?

No

stefanceriu:
Fixes #3830


pixlwave (approved):
👍

PR #3837, opened 2025-02-26
Project updates.

pixlwave:
The PR makes the following changes:
- Stop using the `AppIdentifierPrefix` build var as it isn't compatible with re-signing.
    - Instead we can use `DEVELOPMENT_TEAM` which works if you know who's going to re-sign before you build.
- Manage the `.entitlements` files using XcodeGen like with do with the `Info.plist`s.
- Remove the old NCE extension as its configuration hasn't been kept up to date with project changes (and the implementation was empty).


ElementX/SupportingFiles/target.yml
```diff
@@ -43,7 +43,7 @@ targets:
     platform: iOS
 
     info:
-      path: ../SupportingFiles/Info.plist
+      path: Info.plist
```
stefanceriu, line 46: What's this about?

ElementX/SupportingFiles/ElementX.entitlements
```diff
@@ -28,7 +28,7 @@
 	<true/>
 	<key>keychain-access-groups</key>
 	<array>
-		<string>$(AppIdentifierPrefix)$(BASE_BUNDLE_IDENTIFIER)</string>
+		<string>$(KEYCHAIN_ACCESS_GROUP_IDENTIFIER)</string>
```
stefanceriu, line 31: 👨‍🍳 👌

ElementX/SupportingFiles/target.yml
pixlwave: We're in `SupportingFiles` already, so this was saying to go up one directory and then come straight back into it again.

companioncube2-943456321

stefanceriu: Oh lol, how did that happen?! 😂

stefanceriu (approved):

PR #3850, opened 2025-02-27
Private rooms should have `.invited` history visibility

### LINKED ISSUE #3843: [Task] Set history visibility to "invited" when creating private room
Makes sure to override the visibility when creating the room thanks to https://github.com/matrix-org/matrix-rust-sdk/pull/4682

Velin92:
fixes #3843

We want to always have private rooms to always start their room history with the .invited override


stefanceriu (changes requested):
This looks good to me but can you expand the PR to describe why this is necessary and what used to happen before? 

Can you also link the relevant ticket and rust PR? 

Also please make sure you change the commit title from `bugfix` into something more reasonable. 

Ideally all of this should've been done before sending it off to review to help the reviewer get better context.

PR #3854, opened 2025-02-28
show "Room" for unresolved event permalinks

Velin92:
This is the same behaviour of EW, that we want to have in EX for event permalinks


pixlwave (approved):
LGTM 👌

Velin92:
> Just to double check, do we just want to do this for events? I.e. should we also do the same for the roomID case? Screenshot 2025-02-28 at 10 39 16 am

Checked with @Anderas and we want to have the same behaviour of EW, so... we apply the generic "Room" string only for event permalinks, room permalinks are already consistent with EW

pixlwave:
Makes sense 👍

PR #3856, opened 2025-02-28
Re-enable the error screens in group calls

hughns:
This PR reintroduces what was originally in https://github.com/element-hq/element-x-ios/pull/3813 but then reverted in element-hq/element-x-ios#3819:

> Element Call now sends a 'close' widget action when the widget is ready to close. Usually this will be sent immediately after the 'hangup' action, but it could be sent later if the widget wants to present an error screen before closing. So by listening to the 'close' action rather than the 'hangup' action, we get to see these error screens.
> 
> For https://github.com/element-hq/element-call/issues/2955
> Depends on https://github.com/element-hq/element-call/pull/3011

### DIFF
--- a/ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
+++ b/ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
@@ -17,6 +17,7 @@ struct ElementCallWidgetMessage: Codable {
     
     enum Action: String, Codable {
         case hangup = "im.vector.hangup"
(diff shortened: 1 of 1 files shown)


robintown:
I believe this can be reviewed and merged now that the criteria in https://github.com/element-hq/element-call/issues/2955#issuecomment-2678978558 are met (we backported the Element Call change that it depended upon to [a new release](https://github.com/element-hq/element-call/releases/tag/v0.7.2))

PR #3863, opened 2025-03-03
Use a static room summary provider to resolve room summaries through id and aliases

### LINKED ISSUE #3852: Raw room ID exposed in message permalink
### Steps to reproduce

1. Sent a permalink to a message in another room using EW:

Image

2. Viewed it on EXI:

Image

Tapping on it in EXI correctly takes you to the right room.

#### What did you expect?

Permalink message bubbles shouldn't leak raw room IDs.

#### What happened instead?

Ugly raw room ID.

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

848

### Homeserver

_No response_

### Will you send logs?

Yes

Velin92:
This fixes two issues:
- We clean the state of the `alternateRoomSummaryProvider` when we are done with it to avoid the various views to clash with each others
- We made a new `staticRoomSummaryProvider` to which we can't update its filtering through the protocol making its usage just for the sake of finding a specific in the room list, independent of the filtering or query state, also we set a max page size of `.max` over it

fixes #3852


pixlwave:
We figured it out - using the forwarding screen or global search leaves the alternate provider in a filtered state.

stefanceriu (approved):
Nice, lgtm 👍

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -846,6 +866,14 @@ class ClientProxy: ClientProxyProtocol {
...
                                                                appSettings: appSettings)
             try await alternateRoomSummaryProvider?.setRoomList(roomListService.allRooms())
+            
+            staticRoomSummaryProvider = RoomSummaryProvider(roomListService: roomListService,
+                                                            eventStringBuilder: eventStringBuilder,
+                                                            name: "StaticAllRooms",
+                                                            roomListPageSize: .max,
+                                                            notificationSettings: notificationSettings,
+                                                            appSettings: appSettings)
+            try await staticRoomSummaryProvider?.setRoomList(roomListService.allRooms())
```
stefanceriu, line 876: We need to make sure this doesn't negativelly interact with the normal room summary provider as they're both using the same underlying list now.

PR #3868, opened 2025-03-04
Room mentioning in the composer

Velin92:
This PR adds support for rendering pills in the composer, alongside the auto-suggestions for them once typing the `#` character.

The behaviour is the same as EW, as soon as you type # you will see all the rooms with a canonical alias, and by typing more text you can filter them by searching either their name or the alias, once selecting one of the options, the room address permalink will be added to the message  but rendered as a nice room pill.


ElementX/Sources/Screens/RoomScreen/ComposerToolbar/CompletionSuggestionServiceProtocol.swift
```diff
@@ -19,33 +19,62 @@ struct SuggestionTrigger: Equatable {
     let range: NSRange
 }
 
-enum SuggestionItem: Identifiable, Equatable {
-    case user(item: MentionSuggestionItem)
-    case allUsers(item: MentionSuggestionItem)
+struct SuggestionItem: Identifiable, Equatable {
```
stefanceriu, line 23: Nice 👍

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -695,7 +695,8 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
                                                                                    mediaProvider: userSession.mediaProvider)
         self.timelineController = timelineController
         
-        let completionSuggestionService = CompletionSuggestionService(roomProxy: roomProxy)
+        let completionSuggestionService = CompletionSuggestionService(roomProxy: roomProxy,
+                                                                      roomListPublisher: userSession.clientProxy.staticRoomSummaryProvider?.roomListPublisher.eraseToAnyPublisher() ?? Empty().replaceEmpty(with: []).eraseToAnyPublisher())
```
stefanceriu, line 699: I think we should fatalError if the `staticRoomSummaryProvider` doesn't exist at this point, I don't think there's any good reason for it not to, right?

ElementX/Sources/Screens/RoomScreen/ComposerToolbar/CompletionSuggestionService.swift
stefanceriu, line 1: Would've loved to see the changes to the `completionService` in a different commit, completely separate from the UI layer and together with the unit tests you added at the end. It would've been easier to review and it would've also avoided the question I had bellow about `atOrHash`

ElementX/Sources/Screens/RoomScreen/ComposerToolbar/CompletionSuggestionService.swift
```diff
@@ -77,25 +66,70 @@ final class CompletionSuggestionService: CompletionSuggestionServiceProtocol {
...
+        }
+
         guard let match else {
             return nil
         }
-        
+
         var suggestionText = String(text[match.range])
-        suggestionText.removeFirst()
+        let firstChar = suggestionText.first
```
stefanceriu: My previous suggestion would've avoided this mistake as well. Changes to service should always be accompanied by the unit tests otherwise it can easily lead to commits like `small code correction` which to me, as a reviewer, really doesn't mean anything.

ElementX/Sources/Screens/RoomScreen/ComposerToolbar/CompletionSuggestionService.swift
```diff
@@ -119,7 +119,8 @@ final class CompletionSuggestionService: CompletionSuggestionServiceProtocol {
             return nil
         }
 
-        let firstChar = String(text[match.range]).first
+        let suggestionText = String(text[match.range])
+        let firstChar = suggestionText.first
```
stefanceriu: Again here, a commit called just `fix`. What does that mean? What does it fix and why? If it fixes something then why doesn't it come with the unit test to show it? If it's not that important then why didn't you squash it into some other commit?

ElementX/Sources/Other/Pills/MentionBuilder.swift
stefanceriu, line 1: Is this a right time to add some unit tests for the mention builders?

UnitTests/Sources/CompletionSuggestionServiceTests.swift
stefanceriu, line 1: Noice! 👌

stefanceriu (approved):
This looks and works great, nicely done! 👏 I'm a bit disappointed that we don't match on room names too, I would find that useful.

Also, I would've loved to see more atomic commits with better explanations. At `+558 −174` this is rather big PR and it would've made my life as a reviewer so much easier.

ElementX/Sources/Screens/RoomScreen/ComposerToolbar/CompletionSuggestionService.swift
```diff
@@ -9,7 +9,7 @@ import Combine
 import Foundation
 
 private enum SuggestionTriggerRegex {
-    static let atOrHash = /[@#]\w+/
+    static let atOrHash = /[@#]\w*/
```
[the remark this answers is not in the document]
Velin92: the wildcard * makes the match trigger with 0 or more chars after @ or # while + with at least one, the desired behaviour is to give autocompletion as soon as we type either @ or #, so this is preferred, tests have been updated to check this.

ElementX/Sources/Other/Pills/MentionBuilder.swift
Velin92: sure, already added them for the previous PR where I made the room mentions in the timeline, but can add a small test for the added `MatrixRoomDisplayName`

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
Velin92: I wonder if we should then just make the RoomSummaryProviders unwrapped optionals instead of full optionals, so that knowing they should always be present we don't need to unwrap them

PreviewTests/Sources/__Snapshots__/PreviewTests/test_mentionSuggestionItemView-iPad-en-GB.User-no-display-name.png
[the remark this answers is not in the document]
Velin92: Yeah these were some of the issues I have experiencing when doing the rebase

ElementX/Sources/Screens/RoomScreen/ComposerToolbar/CompletionSuggestionService.swift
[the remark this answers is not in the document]
Velin92: In fact we had an existing bug where the RTE started the suggestion immediately, while the plain text composer did not, this is now fixed and tested

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
Velin92: @pixlwave what do you think about this too? May handle this in another PR

stefanceriu: That sounds like a good idea to me. Unfortunately I don't remember why I made them optional to being with. There must've been a reason but with so many changes to sliding sync over the years I don't remember

stefanceriu: Okay, @pixlwave jogged my memory: back when we had 2 sliding sync lists the `visible_rooms` one was added after the initial setup, once the `all_rooms` would finish its initial SS fetch and as such we couldn't build a summary for it.

PR #3870, opened 2025-03-05
Bump the RustSDK to v25.03.05

stefanceriu:
- adopt the new timeline configuration which now exposes read receipst tracking. For now it should be set to false for all timelines except the live one which still gets configured properly on the rust side through init_timeline. Eventually we will converge on this configurationi API for all of them.
- also adopt the new rust side cache clearing method and remove our own

### DIFF
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -757,13 +757,12 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
         stopSync(isBackgroundTask: false)
         userSessionFlowCoordinator?.stop()
         
-        let userID = userSession.clientProxy.userID
         tearDownUserSession()
     
         // Allow for everything to deallocate properly
         Task {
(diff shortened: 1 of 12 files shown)


ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -757,13 +757,12 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
...
         userSessionFlowCoordinator?.stop()
         
-        let userID = userSession.clientProxy.userID
         tearDownUserSession()
     
         // Allow for everything to deallocate properly
         Task {
             try? await Task.sleep(for: .seconds(2))
-            userSessionStore.clearCache(for: userID)
+            await userSession.clientProxy.clear_caches()
```
pixlwave, line 765: What are you, some kind of Rust developer?!
```suggestion
            await userSession.clientProxy.clearCaches()
```

pixlwave (approved):
LGTM apart from that 🦀 abomination.

pixlwave (approved):
**Much** better! 😄

PR #3873, opened 2025-03-05
Prevent various room subscription task from blocking opening rooms

stefanceriu:
- this came from having seen the opening room modal show up too often which was caused by `isEncrypted` being block_on on the rust side and sometimes making network requests
- we aggreed that the `block_on`s should go, which is handled in https://github.com/matrix-org/matrix-rust-sdk/pull/4756 and https://github.com/matrix-org/matrix-rust-sdk/pull/4757

### DIFF
--- a/ElementX/Sources/Services/Room/JoinedRoomProxy.swift
+++ b/ElementX/Sources/Services/Room/JoinedRoomProxy.swift
@@ -135,15 +135,17 @@ class JoinedRoomProxy: JoinedRoomProxyProtocol {
         
         await timeline.subscribeForUpdates()
         
-        subscribeToRoomInfoUpdates()
-        
-        if isEncrypted {
-            subscribeToIdentityStatusChanges()
+        Task {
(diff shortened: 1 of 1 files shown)


Velin92 (approved):

ElementX/Sources/Services/Room/JoinedRoomProxy.swift
```diff
@@ -135,15 +135,17 @@ class JoinedRoomProxy: JoinedRoomProxyProtocol {
...
-        if isEncrypted {
-            subscribeToIdentityStatusChanges()
+        Task {
+            subscribeToRoomInfoUpdates()
+            
+            subscribeToTypingNotifications()
+            
+            await subscribeToKnockRequests()
+            
+            if isEncrypted { // This is actually blocking on the rust side and might make network requests
```
Hywan, line 145: This may not be blocking anymore soon.

PR #3874, opened 2025-03-06
FF event cache true by default and updated the SDK

Velin92:
Set the event cache FF to true, so that it's always enabled, and updated the SDK


stefanceriu (approved):
Nice, lgtm 👍

PR #3875, opened 2025-03-06
ClientProxy refactor

Velin92:
Refactored the client proxy to:
- throw an error during the init
- to have its init directly instantiate the sync service and the providers

### DIFF
--- a/ElementX/Sources/Services/Client/ClientProxy.swift
+++ b/ElementX/Sources/Services/Client/ClientProxy.swift
@@ -20,13 +20,13 @@ class ClientProxy: ClientProxyProtocol {
     private let mediaLoader: MediaLoaderProtocol
     private let clientQueue: DispatchQueue
     
-    private var roomListService: RoomListService?
+    private var roomListService: RoomListService
     // periphery: ignore - only for retain
     private var roomListStateUpdateTaskHandle: TaskHandle?
     // periphery: ignore - only for retain
     private var roomListStateLoadingStateUpdateTaskHandle: TaskHandle?
 
-    private var syncService: SyncService?
+    private var syncService: SyncService
     // periphery: ignore - only for retain
     private var syncServiceStateUpdateTaskHandle: TaskHandle?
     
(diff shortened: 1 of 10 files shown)


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -1205,3 +1123,57 @@ private class SendQueueRoomErrorListenerProxy: SendQueueRoomErrorListener {
         onErrorClosure(roomId, error)
     }
 }
+
+private struct ConfiguredAppService {
```
pixlwave, line 1127: Maybe this would be better called `AppServices` or `ClientProxyServices`. The fact that they're configured in the init probably doesn't need to be called out. (Not important, just thinking out loud as I see this).

ElementX/Sources/Services/UserSession/UserSessionStore.swift
```diff
@@ -138,17 +138,26 @@ class UserSessionStore: UserSessionStoreProtocol {
             
             MXLog.info("Set up session for user \(credentials.userID) at: \(credentials.restorationToken.sessionDirectories)")
             
-            return await .success(setupProxyForClient(client, needsSlidingSyncMigration: credentials.restorationToken.needsSlidingSyncMigration))
+            return try await .success(setupProxyForClient(client, needsSlidingSyncMigration: credentials.restorationToken.needsSlidingSyncMigration))
+        } catch UserSessionStoreError.failedSettingUpClientProxy(let error) {
+            // If this has failed, there is likely something wrong with the creationg of the sync service
+            // There is nothing we can do, but at the same time we don't want the user the get logged out
+            // So it's better to crash here and let the app restart
```
pixlwave, line 145: Thank you!

Also
```suggestion
            // If this has failed, there is likely something wrong with the creation of the sync service
            // There is nothing we can do, but at the same time we don't want the user to the get logged out
            // So it's better to crash here and let the app restart
```

pixlwave (approved):
Nice, LGTM 👍

PR #3879, opened 2025-03-07
Check if the git is shallow before fetching

Velin92:
Title says it all.
This should fix the current CI issue.


stefanceriu (approved):
Nice, lgtm 👍 

Let's trigger a new nightly with it and see what happens 🚀

PR #3880, opened 2025-03-07
CI: git fetch unshallow only in post-clone

Velin92:


ci_scripts/ci_post_xcodebuild.sh
```diff
@@ -3,6 +3,8 @@
 source ci_common.sh
 
 setup_xcode_cloud_environment
+# Xcode Cloud shallow clones the repo. We need to deepen it to fetch tags, commit history and be able to rebase main on develop at the end of releases.
+git fetch --unshallow --quiet
```
stefanceriu, line 7: Let's put this into a separate `unshallow_repository()` function within `ci_common` and just call it from here.

Velin92: done

stefanceriu (approved):
🚀

PR #3888, opened 2025-03-10
Update sentry to 8.35

Velin92:
Since 8.30+ has a bug that might make reported crashes as "unhandled" 8.35 is now considered the stable one:
https://github.com/getsentry/sentry-cocoa/releases/tag/8.30.1


pixlwave (approved):
LGTM, although if we're not going all the way up to date right now, we should probably at least take 8.31.1 as that has a crash fix in it.

Velin92:
> LGTM, although if we're not going all the way up to date right now, we should probably at least take 8.31.1 as that has a crash fix in it.

Well Sentry is suggesting to skip any non 8.35+ version since they all contain a bug which make crash reports useless, I feel we should at least go to 8.35 to keep the most stable implementation without too many changes.

pixlwave:
Ooops that was a typo. I meant 8.35.1 fixes a crash.

Issue #3893, opened 2025-03-12
Share extension flashes up a spurious timeline when it loads, causing a slight heart attack


ara4n:
### Steps to reproduce

1. Want to share a photo with someone
2. Hit the share button in the Photos app
3. Select Element X from the share list. (EX is already running in the bg)
4. EX pops up... showing you the timeline of whatever room you were last in, for a fraction of a second.
5. If this is a room which it would be Really Bad To Share Into, your heart jumps into your mouth a bit, because naively it looks like EX might be about to hurl your image into that room without further prompting.
6. Then, however, the actual list of rooms to share into gets displayed.

#### What did you expect?

Either jump straight to showing the room-selection-list, or first show a blank view rather than the last loaded timeline, in order to prevent panicking the user that the app is in the process of sharing their photo into the wrong room.

#### What happened instead?

An unnecessary adrenalin rush.

### Application version

862

### Will you send logs?

No

PR #3902, opened 2025-03-14
Fix sharing from in-app QuickLook to itself

stefanceriu:
Delay the presentation of the share extension room selection screen until the currently presented room is correctly torn down

If QuickLook is presented (e.g. a timeline media is being viewed in full screen) then its animated dismissal will interfere with presenting other screens and break the state machine. Unfortunately I wasn't able to find any reasonable way to have the QuickLook dismiss itself without an animation

On the other hand, we can fix both this issue and improve #3893 by just delaying the presentation of the room selection list, which is what this patch does.

### DIFF
--- a/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
@@ -307,9 +307,15 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
             case (.userProfileScreen, .dismissedUserProfileScreen, .roomList):
                 break
                 
-            case (.roomList, .showShareExtensionRoomList, .shareExtensionRoomList(let sharePayload)):
-                clearRoute(animated: animated)
-                presentRoomSelectionScreen(sharePayload: sharePayload, animated: animated)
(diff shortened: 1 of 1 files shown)


ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -307,9 +307,15 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
...
                 break
                 
-            case (.roomList, .showShareExtensionRoomList, .shareExtensionRoomList(let sharePayload)):
-                clearRoute(animated: animated)
-                presentRoomSelectionScreen(sharePayload: sharePayload, animated: animated)
+            case (.roomList(let selectedRoomID), .showShareExtensionRoomList, .shareExtensionRoomList(let sharePayload)):
+                Task {
+                    if selectedRoomID != nil {
+                        self.clearRoute(animated: animated)
+                        try? await Task.sleep(for: .seconds(1.5))
```
Velin92, line 314: 1.5 sounds like a lot, maybe the task should be cancellable?
Or could we reduce this a bit?

stefanceriu: It's intentionally large so that the user sees the room get popped before showing the share room selection screen and to address #3893

PR #3905, opened 2025-03-17
Removed images in pills in favour of text decorations

Velin92:
Pills have been redesigned, to use text decorations instead of avatars


pixlwave (approved):
Sad to see these go, but PR lgtm 👍

PR #3907, opened 2025-03-17
Added a test to check if URLs with RTL are not marked as phishing

Velin92:


stefanceriu (approved):
👌

UnitTests/Sources/AttributedStringBuilderTests.swift
```diff
@@ -745,6 +745,22 @@ class AttributedStringBuilderTests: XCTestCase {
         XCTAssertEqual(link.absoluteString, "https://matrix.org")
     }
     
+    func testValidLinkWithRTLOverride() {
+        let htmlString = "<a href=\"https://matrix.org\">\u{202E}https://matrix.org</a>"
```
bmarty, line 749: @Velin92 it's not the text which needs to be tested, but the URL. So I think you should test something like:

`let htmlString = "<a href=\"https://\u{202E}matrix.org\">https://matrix.org</a>"`

PR #3908, opened 2025-03-17
Pin all 3rd party github actions to their full length commit SHA

stefanceriu:
- this is the recommended security practice as per https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions


pixlwave (approved):
LGTM!

Curious to see if Renovate can still stick to codecov@v3 when using commits and what it does with the comments you've added.

stefanceriu:
Guess we have our answer https://github.com/element-hq/element-x-ios/pull/3915

pixlwave:
Woah, nice!

PR #3910, opened 2025-03-18
Add overrides for OIDC configuration.

pixlwave:
This PR adds more overrides to `AppSettings.swift`.

It also fixes our SwiftLint configuration as we were only linting the main app and UI/Unit tests.


ElementX/Sources/Application/AppSettings.swift
```diff
@@ -81,10 +81,38 @@ final class AppSettings {
...
     
-    func override(defaultHomeserverAddress: String? = nil) {
+    func override(defaultHomeserverAddress: String? = nil,
+                  oidcRedirectURL: URL? = nil,
+                  websiteURL: URL? = nil,
+                  logoURL: URL? = nil,
+                  copyrightURL: URL? = nil,
+                  acceptableUseURL: URL? = nil,
+                  privacyURL: URL? = nil,
+                  supportEmailAddress: String? = nil) {
```
Velin92, line 91: Couldn't the default values be the constructor so that we don't have to do this `if let` cascade?

Velin92 (approved):
Approved just a small suggestion because I don't like all those ifs

.swiftlint.yml
```diff
@@ -12,13 +12,12 @@ opt_in_rules:
   - trailing_closure
 
 included:
-  - ElementX
-  - UnitTests
-  - UITests
+  - "*/Sources/*"
```
stefanceriu, line 15: 🙌

.swiftlint.yml
```diff
@@ -12,13 +12,12 @@ opt_in_rules:
...
   - Tools/Scripts/Templates
 excluded:
-  - IntegrationTests
-  - ElementX/Sources/Generated
+  - "*/Sources/Generated/*"
+  - IntegrationTests # prints to stdout for performance tests.
```
stefanceriu, line 19: Actually we don't need that anymore? 🤔

ElementX/Sources/Application/AppSettings.swift
stefanceriu: I think having the default values defined inline with the properties is fine, but Mauro's got a point: maybe we should make them all mandatory and have the user gather all the data beforehand.

.swiftlint.yml
pixlwave: Ok, then I'll tidy it in a follow-up PR 👍

ElementX/Sources/Application/AppSettings.swift
pixlwave: Done in https://github.com/element-hq/element-x-ios/pull/3910/commits/d1e92cdd6be0646bec4a8b2d8fc237fb141a0892

ElementX.xcodeproj/xcshareddata/xcschemes/PreviewTests.xcscheme
pixlwave, line 1: I think this thing got committed in a previous PR by mistake. This is from a fresh `xcodegen` run.

stefanceriu, line 1: Committed by mistake?

pixlwave: https://github.com/element-hq/element-x-ios/pull/3910#discussion_r2002930287

PR #3911, opened 2025-03-18
Fix: completion service matches now any character

Velin92:
Aside from whitespaces
Also added a test for it


pixlwave (approved):
LGTM, thanks! 🙌

PR #3916, opened 2025-03-19
Updated the SDK to 25.03.24

### LINKED ISSUE #3909: Error: Missing Rust Targets to Build the SDK
I am encountering the following error when trying to build the SDK for Element X iOS:

"Error: Rust is missing the necessary targets to build the SDK.
Run the following command to install them:

rustup target add aarch64-apple-ios x86_64-apple-ios aarch64-apple-darwin x86_64-apple-darwin aarch64-apple-ios-sim"

I’ve followed the instructions provided to install the necessary Rust targets using rustup target add, but I keep getting the same error. I’ve tried updating Rust, resetting packages, and ensuring all dependencies are resolved, but the issue persists.

Steps I’ve taken:
	1.	Installed the necessary Rust targets:
"rustup target add aarch64-apple-ios x86_64-apple-ios aarch64-apple-darwin x86_64-apple-darwin aarch64-apple-ios-sim"

	2.	Ran swift run tools build-sdk to try building the SDK.
	3.	Verified that Rust is up-to-date.

Despite all this, the issue continues, and I am unable to successfully build the SDK.

Is there any additional configuration required, or is this an issue with the project setup?

Velin92:
Updates the SDK to 25.3.24 
- This includes some refactors of the OIDC flow
- A refactor for the element call widget driver to support newer EC versions
- This also updates the build sdk tool to check the list of targets in a more precise way

fixes #3909

### DIFF
--- a/ElementX/Sources/Application/AppSettings.swift
+++ b/ElementX/Sources/Application/AppSettings.swift
@@ -267,7 +267,12 @@ final class AppSettings {
 
     // MARK: - Element Call
     
-    let elementCallBaseURL: URL = "https://call.element.io"
+    let elementCallBaseURL: URL = "https://call.element.io/room"
+    
+    // These are publicly availble on https://call.element.io so we don't neeed to treat them as secrets
+    let elementCallPosthogAPIHost = "https://posthog-element-call.element.io"
+    let elementCallPosthogAPIKey = "phc_rXGHx9vDmyEvyRxPziYtdVIv0ahEv8A9uLWFcCi1WcU"
+    let elementCallPosthogSentryDSN = "https://3bd2f95ba5554d4497da7153b552ffb5@sentry.tools.element.io/41"
     
     @UserPreference(key: UserDefaultsKeys.elementCallBaseURLOverride, defaultValue: nil, storageType: .userDefaults(store))
     var elementCallBaseURLOverride: URL?
--- a/ElementX/Sources/Screens/CallScreen/CallScreenCoordinator.swift
+++ b/ElementX/Sources/Screens/CallScreen/CallScreenCoordinator.swift
@@ -43,7 +43,9 @@ final class CallScreenCoordinator: CoordinatorProtocol {
         viewModel = CallScreenViewModel(elementCallService: parameters.elementCallService,
                                         configuration: parameters.configuration,
                                         allowPictureInPicture: parameters.allowPictureInPicture,
-                                        appHooks: parameters.appHooks)
+                                        appHooks: parameters.appHooks,
+                                        appSettings: ServiceLocator.shared.settings,
+                                        analyticsService: ServiceLocator.shared.analytics)
     }
     
     func start() {
--- a/ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
+++ b/ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
@@ -50,6 +50,7 @@ struct ElementCallWidgetMessage: Codable {
 class ElementCallWidgetDriver: WidgetCapabilitiesProvider, ElementCallWidgetDriverProtocol {
     private let room: RoomProtocol
     private let deviceID: String
+    
     private var widgetDriver: WidgetDriverAndHandle?
     
     let widgetID = UUID().uuidString
@@ -65,14 +66,14 @@ class ElementCallWidgetDriver: WidgetCapabilitiesProvider, ElementCallWidgetDriv
         self.deviceID = deviceID
     }
     
-    func start(baseURL: URL, clientID: String, colorScheme: ColorScheme) async -> Result<URL, ElementCallWidgetDriverError> {
+    func start(baseURL: URL, clientID: String, colorScheme: ColorScheme, analyticsConfiguration: ElementCallAnalyticsConfiguration?) async -> Result<URL, ElementCallWidgetDriverError> {
         guard let room = room as? Room else {
             return .failure(.roomInvalid)
         }
         
         let useEncryption = await (try? room.latestEncryptionState() == .encrypted) ?? false
-        
         let widgetSettings: WidgetSettings
+        
         do {
             widgetSettings = try newVirtualElementCallWidget(props: .init(elementCallUrl: baseURL.absoluteString,
                                                                           widgetId: widgetID,
@@ -84,13 +85,13 @@ class ElementCallWidgetDriver: WidgetCapabilitiesProvider, ElementCallWidgetDriv
                                                                           confineToRoom: true,
                                                                           font: nil,
                                                                           encryption: useEncryption ? .perParticipantKeys : .unencrypted,
-                                                                          intent: .startCall,
-                                                                          hideScreensharing: false, // The web view will hide it automatically
+                                                                          intent: nil,
+                                                                          hideScreensharing: false,
                                                                           posthogUserId: nil,
-                                                                          posthogApiHost: nil,
-                                                                          posthogApiKey: nil,
-                                                                          rageshakeSubmitUrl: nil,
-                                                                          sentryDsn: nil,
+                                                                          posthogApiHost: analyticsConfiguration?.posthogAPIHost,
+                                                                          posthogApiKey: analyticsConfiguration?.posthogAPIKey,
+                                                                          rageshakeSubmitUrl: analyticsConfiguration?.rageshakeSubmitURL,
+                                                                          sentryDsn: analyticsConfiguration?.sentryDSN,
                                                                           sentryEnvironment: nil))
         } catch {
             MXLog.error("Failed to build widget settings: \(error)")
--- a/ElementX.xcodeproj/project.pbxproj
+++ b/ElementX.xcodeproj/project.pbxproj
@@ -8565,7 +8565,7 @@
 			repositoryURL = "https://github.com/element-hq/matrix-rust-components-swift";
 			requirement = {
 				kind = exactVersion;
-				version = 25.03.21;
+				version = 25.03.24;
 			};
 		};
 		701C7BEF8F70F7A83E852DCC /* XCRemoteSwiftPackageReference "GZIP" */ = {
--- a/ElementX.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/ElementX.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -149,8 +149,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/element-hq/matrix-rust-components-swift",
       "state" : {
-        "revision" : "93b8303fa004060c8ba6a856e0b66a1d47514490",
-        "version" : "25.3.21"
+        "revision" : "0da89f2a2ddaf2306144f42aed642af3f3228a1c",
+        "version" : "25.3.24"
       }
     },
     {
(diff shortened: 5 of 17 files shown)


ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
```diff
@@ -81,11 +81,17 @@ class ElementCallWidgetDriver: WidgetCapabilitiesProvider, ElementCallWidgetDriv
...
                                                                           fontScale: nil,
                                                                           appPrompt: false,
-                                                                          skipLobby: true,
                                                                           confineToRoom: true,
                                                                           font: nil,
-                                                                          analyticsId: nil,
-                                                                          encryption: useEncryption ? .perParticipantKeys : .unencrypted))
+                                                                          encryption: useEncryption ? .perParticipantKeys : .unencrypted,
+                                                                          intent: nil,
+                                                                          hideScreensharing: false,
```
Velin92, line 88: not sure about this one

ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
```diff
@@ -81,11 +81,17 @@ class ElementCallWidgetDriver: WidgetCapabilitiesProvider, ElementCallWidgetDriv
                                                                           preload: nil,
                                                                           fontScale: nil,
                                                                           appPrompt: false,
-                                                                          skipLobby: true,
                                                                           confineToRoom: true,
                                                                           font: nil,
-                                                                          analyticsId: nil,
-                                                                          encryption: useEncryption ? .perParticipantKeys : .unencrypted))
+                                                                          encryption: useEncryption ? .perParticipantKeys : .unencrypted,
+                                                                          intent: nil,
```
Velin92, line 87: Also no idea what this thing is

Velin92: okay Hugh confirmed to set this to nil

ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
pixlwave: FWIW EC is disabled entirely on macOS as webRTC doesn't work properly in the iOS web view on macOS.

Velin92: I see! NP Hugh told me that is fine to keep this as false, since it should only show only if supported, however for some reason I am seeing the button on non embedded EC regardless

pixlwave: Presumably you want **hide**Screensharing set to `true` 😉

ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
```diff
@@ -73,6 +73,11 @@ class ElementCallWidgetDriver: WidgetCapabilitiesProvider, ElementCallWidgetDriv
         let useEncryption = await (try? room.latestEncryptionState() == .encrypted) ?? false
         
         let widgetSettings: WidgetSettings
+        // If the user has opted out of analytics, we should configure these values as nil
+        let widgetSecrets = ServiceLocator.shared.analytics.isEnabled ? WidgetSecrets(posthogApiHost: Secrets.postHogHost,
```
Velin92, line 77: So I could inject the analyticsService into this object, however I would also need to inject it first in the `JoinedRoomProxy`, if you guys feel having the `ServiceLocator` being used here is not an issue, I'll leave it as it is

ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
```diff
@@ -73,6 +73,11 @@ class ElementCallWidgetDriver: WidgetCapabilitiesProvider, ElementCallWidgetDriv
...
+                                                                                      posthogApiKey: Secrets.postHogAPIKey,
+                                                                                      rageshakeSubmitUrl: Secrets.rageshakeServerURL,
+                                                                                      sentryDsn: Secrets.sentryDSN) : nil
```
pixlwave, line 80: I think a Sentry DSN is per project, so we're probably going to need a specific Element Call one here.

ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
```diff
@@ -207,3 +212,10 @@ class ElementCallWidgetDriver: WidgetCapabilitiesProvider, ElementCallWidgetDriv
         }
     }
 }
+
+private struct WidgetSecrets {
+    let posthogApiHost: String
+    let posthogApiKey: String
+    let rageshakeSubmitUrl: String
+    let sentryDsn: String
```
pixlwave, line 220:
```suggestion
    let posthogAPIHost: String
    let posthogAPIKey: String
    let rageshakeSubmitURL: String
    let sentryDSN: String
```

ElementX/Sources/Application/AppSettings.swift
```diff
@@ -267,7 +267,7 @@ final class AppSettings {
 
     // MARK: - Element Call
     
-    let elementCallBaseURL: URL = "https://call.element.io"
+    let elementCallBaseURL: URL = "https://call.element.io/room"
```
pixlwave, line 270: Is there anywhere else we could append this? Because given that we expose `elementCallBaseURLOverride` in the developer settings it would require anyone setting that to know what the path of the call is which feels awkward to me.

ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
Velin92: Okay I'll check how to include it in our project with @hughns

ElementX/Sources/Application/AppSettings.swift
Velin92: Considering that is a dev option, that probably will be from the moment we embed EC mostly used to test EC dev on EX, I don't think we should worry too much about it. Also, this base URL is going to get removed pretty soon in favour of embedding.

pixlwave: Sure, but appending it [here](https://github.com/element-hq/element-x-ios/blob/8b070d9a64133e97ac5e759d23a8cc70d9d5e193/ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift#L151) wouldn't exactly be any more work.

ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
hughns: Two answers to this:

1. Discussion with @fkwp on this topic last week concluded that:
  - it might make sense to send Element Call related analytics (Posthog and Sentry) to a different destinations from the Element Messenger apps. However, this would require discussion with the platform teams (which, I suppose, is what is happening in this thread). If you prefer to add in the extra config so that you can send to different Posthog and Sentry destinations at this time then I am supportive (assuming also done on Android for consistency))  
  - in the meantime, from an Element Call point of view, it would be adequate to send them to the same places as the Messenger apps (which is what is currently done in https://github.com/element-hq/element-web/pull/29309 for EW/ED)
2. If you want to side step the topic for this PR then you could simply set all of these "secrets" to nil in this PR. This is because they won't be used until the subsequent embedding PR. Yes, this would be kicking the can down the road. 😇

fkwp: > I think a Sentry DSN is per project, so we're probably going to need a specific Element Call one here.

Let's be pragmatic and don't block this PR. If we want this behavior we can add it in futures.

jmartinesp: IMO it makes sense to create new Sentry and posthog projects for this, unless there's an easy way to categorise the bug reports/statistics so they don't end up polluting the EX ones. Same for rageshakes I guess?

ElementX/Sources/Application/AppSettings.swift
pixlwave: Ok nice 👍

ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
Velin92: getting a bit confused now, so how do we want to proceed?

pixlwave (approved):

ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
hughns: I don't have the authority to say how this should be handled in EX. I am happy to discuss with anyone who gets to make the decision.

From the sound of Doug's original feedback I think you want to use EC specific values.

As such, these are the values to use:

- posthogApiHost: `https://posthog-element-call.element.io`
- posthogApiKey: `phc_rXGHx9vDmyEvyRxPziYtdVIv0ahEv8A9uLWFcCi1WcU`
- rageshakeSubmitUrl: `https://element.io/bugreports/submit`
- sentryDsn: `https://3bd2f95ba5554d4497da7153b552ffb5@sentry.tools.element.io/41`

n.b. the rageshake URL is already the same as for EX because I believe it is already routed/categorised sensibly on the server side.

ElementX/Sources/Screens/CallScreen/CallScreenCoordinator.swift
```diff
@@ -43,7 +43,9 @@ final class CallScreenCoordinator: CoordinatorProtocol {
         viewModel = CallScreenViewModel(elementCallService: parameters.elementCallService,
                                         configuration: parameters.configuration,
                                         allowPictureInPicture: parameters.allowPictureInPicture,
-                                        appHooks: parameters.appHooks)
+                                        appHooks: parameters.appHooks,
+                                        appSettings: ServiceLocator.shared.settings,
+                                        analyticsService: ServiceLocator.shared.analytics)
```
pixlwave, line 48: We can inject these into the parameters surely :)

ElementX/Sources/Application/AppSettings.swift
```diff
@@ -267,7 +267,13 @@ final class AppSettings {
...
+    
+    let elementCallPosthogAPIHost = "https://posthog-element-call.element.io"
+    
+    let elementCallPosthogAPIKey = "phc_rXGHx9vDmyEvyRxPziYtdVIv0ahEv8A9uLWFcCi1WcU"
+    
+    let elementCallPosthogSentryDSN = "https://3bd2f95ba5554d4497da7153b552ffb5@sentry.tools.element.io/41"
```
pixlwave, line 276: Maybe group these together and put a comment above saying that unlike our analytics, these are publicly available in the source code so no need to add them as Secrets?

pixlwave (approved):

ElementX/Sources/Screens/CallScreen/CallScreenCoordinator.swift
Velin92: Usually we are okay in injecting these from the ServiceLocator in our coordinators. This is done pretty much in all our coordinators

ElementX/Sources/Application/AppSettings.swift
Velin92: they are already under the `// MARK: - Element Call` section

PR #3917, opened 2025-03-19
Stop CI from uploading Codecov results when a PR comes from a fork.

pixlwave:
Looking at #3914, unit tests failed because the codecov upload failed. Also the enterprise tests were being run (which obviously then failed).
- The condition for the latter has been updated to actually work.
    - It was checking `'element-hq/element-x-ios' == 'element-hq/element-x-ios'` regardless of whether the PR came from our repo or a fork 🙈
- The same condition has been added to the codecov steps in the former.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -60,6 +60,7 @@ jobs:
           xcresultparser -q -o cobertura -t ElementX -p $(pwd) fastlane/test_output/PreviewTests.xcresult > fastlane/test_output/preview-cobertura.xml
       
       - name: Upload coverage to Codecov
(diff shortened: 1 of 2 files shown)


.github/workflows/unit_tests.yml
```diff
@@ -60,6 +60,7 @@ jobs:
           xcresultparser -q -o cobertura -t ElementX -p $(pwd) fastlane/test_output/PreviewTests.xcresult > fastlane/test_output/preview-cobertura.xml
       
       - name: Upload coverage to Codecov
+        if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository }} # Skip in forks
```
pixlwave, line 63: Note to self: Make sure both of Codecov uploads take place before merging this PR.

stefanceriu (approved):
Noice 👌

PR #3919, opened 2025-03-19
Hide invite avatars when such flag is on

### LINKED ISSUE #3920: [Task] Advanced settings override for hide invite avatars and timeline media

Velin92:
This affects:
- Invited room preview inviter avatar
- Invited room preview room avatar
- Invited room cell inviter avatar in the room list
- Invited room cell room avatar in the room list
- Push notification for an invite

We also updated `UserPreference` to account for a way to handle remote preferences in the future
fixes #3920


ElementX/Sources/Application/AppSettings.swift
```diff
@@ -307,6 +309,9 @@ final class AppSettings {
     @UserPreference(key: UserDefaultsKeys.enableOnlySignedDeviceIsolationMode, defaultValue: false, storageType: .userDefaults(store))
     var enableOnlySignedDeviceIsolationMode
     
+    @UserPreference(key: UserDefaultsKeys.hideInviteAvatars, defaultValue: false, storageType: .userDefaults(store))
+    var hideInviteAvatars
```
Velin92, line 313: I was thinking, since in the future we will have an option from the BE that sets this policy and this is only an override, maybe we should have these as optional booleans?

Velin92: okay after the call we had today I feel would be better to convert this to an optional boolean, while default value nil does behave as false, but once it gets modified even once it becomes stored and will take precedence over the BE value. If this value has never been touched and will be nil, then the BE value will take precedence.

@pixlwave @stefanceriu WDYT?

ElementX/Sources/Other/UserPreference.swift
```diff
@@ -8,28 +8,59 @@
...
+///
+/// - Note: This wrapper allows enforcing a default value through the `forceDefault` closure.
 @propertyWrapper
 final class UserPreference<T: Codable> {
     private let key: String
     private var keyedStorage: any KeyedStorage<T>
-    private let defaultValue: T
+    private let defaultValue: () -> T
     private let subject: PassthroughSubject<T, Never> = .init()
+    private var cancellable: AnyCancellable?
```
Velin92, line 22: so this approach works well for now but... I wonder if defaultValue should actually be a publisher too, since we probably want to have the app settings send an update on the subject also when a remote value has changed in the future.
Another solution would be to have both `defaultValue` and an `overrideValue`. The first can just be T, while overrideValue can be a publisher and take the place of `forceDefault`

ElementX.xcodeproj/project.pbxproj
```diff
@@ -5500,6 +5502,13 @@
 			path = ShareExtension;
 			sourceTree = "<group>";
 		};
+		D174C6E7DCA00AAFC0169925 /* ElementCall */ = {
```
pixlwave, line 5505: What's this about? Do you have an empty ElementCall directory that xcodegen is picking up on and adding to the project?

ElementX/Sources/Other/Extensions/UNNotificationContent.swift
```diff
@@ -101,32 +101,24 @@ extension UNMutableNotificationContent {
...
+        if !forcePlaceholder, let mediaSource = icon.mediaSource {
             switch await mediaProvider?.loadThumbnailForSource(source: mediaSource, size: .init(width: 100, height: 100)) {
             case .success(let data):
                 fetchedImage = INImage(imageData: data)
             case .failure(let error):
                 MXLog.error("Couldn't add sender icon: \(error)")
-                // ...The provider failed to fetch
-                needsPlaceholder = true
             case .none:
                 break
```
pixlwave, line 115: Should we add an `if let mediaProvider` to the above condition so that this `case .none` can be removed? It took me a while to figure out why it was here when understanding the removal of `needsPlaceholder`.

ElementX/Sources/Screens/HomeScreen/View/HomeScreenInviteCell.swift
```diff
@@ -15,11 +15,16 @@ struct HomeScreenInviteCell: View {
     
     let room: HomeScreenRoom
     let context: HomeScreenViewModel.Context
+    let hideInviteAvatars: Bool
+    
+    private var avatar: RoomAvatar {
+        hideInviteAvatars ? .room(id: room.id, name: room.name, avatarURL: nil) : room.avatar
+    }
```
pixlwave, line 22: This is a technicality, but the `true` branch should really switch on `room.avatar` and return the same type. So that if it contains a room hero, the colour of the invite placeholder avatar matches the colour of the user's placeholder avatar once joined (for the case where they don't have an avatar set).

ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenModels.swift
```diff
@@ -68,9 +83,9 @@ struct JoinRoomScreenViewState: BindableState {
     
     var avatar: RoomAvatar? {
         if isDMInvite, let inviter = roomDetails?.inviter {
-            return .room(id: roomID, name: inviter.displayName, avatarURL: inviter.avatarURL)
-        } else if let avatar = roomDetails?.avatar {
-            return avatar
+            return .room(id: roomID, name: inviter.displayName, avatarURL: hideInviteAvatars ? nil : inviter.avatarURL)
```
pixlwave, line 86: And likewise here, if there aren't any heroes in the invite details, we should be returning
```
.heroes([.init(userID: inviter.id, displayName: inviter.displayName, avatarURL: hideInviteAvatars ? nil : inviter.avatarURL)])
```
so that the placeholder colour matches.

NSE/Sources/NotificationContentBuilder.swift
```diff
@@ -93,7 +93,8 @@ struct NotificationContentBuilder {
         notification = try await notification.addSenderIcon(using: mediaProvider,
                                                             senderID: notificationItem.senderID,
                                                             senderName: notificationItem.senderDisplayName ?? notificationItem.roomDisplayName,
-                                                            icon: icon(for: notificationItem))
+                                                            icon: icon(for: notificationItem),
+                                                            forcePlaceholder: settings.hideInviteAvatars)
```
pixlwave, line 97: Question: I don't see anywhere where we check if this notification is for an invite? I would have thought this should either happen here, or that the parameter should be called `hideInviteAvatars` and the `UNMutableNotificationContent` extension should be making the check.

PreviewTests/Sources/__Snapshots__/PreviewTests/homeScreenInviteCell.iPhone-16-en-GB-0.png
pixlwave, line 1: Would it be possible to name the new room "No avatars" or something similar so that it's obviously a mistake if someone regenerates this and the avatar is suddenly shown?

ElementX/Sources/Other/UserPreference.swift
```diff
@@ -8,28 +8,59 @@
...
+    let forceDefault: CurrentValuePublisher<Bool, Never>
     
-    init(key: String, defaultValue: T, keyedStorage: any KeyedStorage<T>) {
+    /// Initializes the property wrapper.
+    ///
+    /// - Parameters:
+    ///   - key: The key used to store and retrieve the value.
+    ///   - defaultValue: The default value to use if no stored value exists or if `forceDefault` is `true`.
+    ///   - keyedStorage: The storage instance where the value is saved.
+    ///   - forceDefault: A publisher that determines whether the default value should always be used. Defaults to  publish`false`. Useful in the context of MDM settings.
```
pixlwave, line 33:
```suggestion
    ///   - forceDefault: A publisher that determines whether the default value should always be used. Defaults to publish `false`. Useful in the context of remote settings.
```

ElementX/Sources/Other/UserPreference.swift
```diff
@@ -8,28 +8,59 @@
...
-            keyedStorage[key] ?? defaultValue
+            guard !forceDefault.value else {
+                return defaultValue()
+            }
+            return keyedStorage[key] ?? defaultValue()
         }
         set {
+            guard !forceDefault.value else {
+                return
+            }
```
pixlwave, line 63: Personally I think it would be reasonable to still set the intended value even if the `get` value ends up being overridden. That way if we later received a `forceDefault == false`, we would return to the expected value. WDYT?

pixlwave (approved):
Very nice 😎

ElementX/Sources/Other/UserPreference.swift
Velin92: oh that seems reasonable however I wonder if we should actually do this: https://github.com/element-hq/element-x-ios/pull/3919/files#r2006086885

NSE/Sources/NotificationContentBuilder.swift
Velin92: look a few lines before:
https://github.com/element-hq/element-x-ios/pull/3919/files#diff-104c57e49004c763b2557fd634bab9e1f69ae57b35ef66ca0f53c651e29f3734R81-R97

pixlwave: lol! 🤦‍♂️

Follow-up question then, how come this wouldn't work?
```
icon: settings.hideInviteAvatars ? .init(nil, nil) : icon(for: notifiationItem)
```

ElementX/Sources/Other/UserPreference.swift
pixlwave: Without knowing what this would look like, you could combine both ideas and have a plain `defaultValue` and a plain `forceDefault` and a publisher for `remoteValue`. That way you will always re-publish when remoteValue changes, and you can choose whether you want remoteValue to just override the default, or to also override the user.

pixlwave: Would probably call it `forceRemote` at that stage.

ElementX/Sources/Screens/HomeScreen/View/HomeScreenInviteCell.swift
Velin92: This actually also made me realise that for DMs, invite avatars are actually broken, no hero is ever returned so you end up showing a room that is named "Username" but has no avatar. I implemented a better solution, but probably it won't fix the underlying problem.

Velin92: Another way to fix this, would be use directly the inviter which is available in this context, to get their image and render if this is a direct room with a single member...

NSE/Sources/NotificationContentBuilder.swift
Velin92: the icon contains also the data to render the placeholder

PreviewTests/Sources/__Snapshots__/PreviewTests/homeScreenInviteCell.iPhone-16-en-GB-0.png
Velin92: changed the alias

NSE/Sources/NotificationContentBuilder.swift
pixlwave: Ah got it!

ElementX/Sources/Other/UserPreference.swift
Velin92: I think was able to pull off this solution with only two publishers: `forceDefault` and `defaultValue`, the `UserPreference` in the appSettings doesn't need to know that is being forced by a remote value or not, it just cares if it should use a defaultValue, that such defaultValue is up to date, and if it needs to override the internal value.

PR #3925, opened 2025-03-20
Update the SDK to 25.03.20.

pixlwave:
Handles breaking changes and also updates Enterprise ready for the next release.


stefanceriu (approved):
👌

PR #3939, opened 2025-03-25
Embed element call

Velin92:
This PR includes a new Swift package that embeds element call into the app for security reasons. It's still however possible to override the file URL, with a custom URL to test out a remote implementation of EC through the developer options.

### DIFF
--- a/ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
+++ b/ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
@@ -148,8 +148,6 @@ class CallScreenViewModel: CallScreenViewModelType, CallScreenViewModelProtocol 
                 
                 let baseURL = if let elementCallBaseURLOverride {
                     elementCallBaseURLOverride
-                } else if case .success(let wellKnown) = await clientProxy.getElementWellKnown(), let wellKnownCall = wellKnown?.call {
-                    wellKnownCall.widgetURL
                 } else {
                     elementCallBaseURL
                 }
--- a/ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
+++ b/ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
@@ -164,8 +164,7 @@ class CallScreenViewModel: CallScreenViewModelType, CallScreenViewModelProtocol 
                                                 colorScheme: colorScheme,
                                                 analyticsConfiguration: analyticsConfiguration) {
                 case .success(let url):
-                    // TODO: Set back to url, for now we are just testing if the baseURL is able to load, the widget url still defaults to aboutblank
-                    state.url = baseURL
+                    state.url = url
                 case .failure(let error):
                     MXLog.error("Failed starting ElementCall Widget Driver with error: \(error)")
                     state.bindings.alertInfo = .init(id: UUID(),
--- a/ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
+++ b/ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
@@ -164,7 +164,8 @@ class CallScreenViewModel: CallScreenViewModelType, CallScreenViewModelProtocol 
                                                 colorScheme: colorScheme,
                                                 analyticsConfiguration: analyticsConfiguration) {
                 case .success(let url):
-                    state.url = url
+                    // TODO: Set back to url, for now we are just testing if the baseURL is able to load, the widget url still defaults to aboutblank
+                    state.url = baseURL
                 case .failure(let error):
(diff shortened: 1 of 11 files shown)


ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
```diff
@@ -148,8 +148,6 @@ class CallScreenViewModel: CallScreenViewModelType, CallScreenViewModelProtocol
                 
                 let baseURL = if let elementCallBaseURLOverride {
                     elementCallBaseURLOverride
-                } else if case .success(let wellKnown) = await clientProxy.getElementWellKnown(), let wellKnownCall = wellKnown?.call {
```
hughns, line 151: Can we also remove the corresponding code in `ElementWellKnown` as we do not want to be using `ElementWellKnown.call` at all going forward.

e.g. remove https://github.com/element-hq/element-x-ios/blob/791ce67808cc9744a94534bd07375e2b6c3b1bc3/ElementX/Sources/Services/Client/ElementWellKnown.swift#L12-L21

project.yml
```diff
@@ -134,6 +134,9 @@ packages:
   Version:
     url: https://github.com/mxcl/Version
     minorVersion: 2.1.0
+  EmbeddedElementCall:
+    url: https://github.com/element-hq/element-call-swift
+    exactVersion: 0.9.0-rc.3
```
pixlwave, line 139: Could we put this in the top section alongside the rest of our 1st-party dependencies?

ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
```diff
@@ -159,8 +167,8 @@ private struct CallView: UIViewRepresentable {
         // MARK: - WKUIDelegate
         
         func webView(_ webView: WKWebView, decideMediaCapturePermissionsFor origin: WKSecurityOrigin, initiatedBy frame: WKFrameInfo, type: WKMediaCaptureType) async -> WKPermissionDecision {
-            // Don't allow permissions for domains different than what the call was started on
-            guard origin.host == url.host else {
+            // Allow if the origin is local, otherwise don't allow permissions for domains different than what the call was started on
+            guard origin.protocol == "file" || origin.host == url.host else {
```
pixlwave, line 171: It would be nice to check if the `origin.path` starts with `EmbeddedElementCall.bundle.bundleURL.path` rather than blindly allowing any file URLs to load here.

pixlwave (approved):
LGTM 👏

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
Velin92: sure! let me get rid of that

ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
Velin92: Sure, even if i think this is a bit overkill, because we already only give read permissions to out bundle, so technically speaking even if someone will try to override the URL with another file URL, they won't be able to run it at all

Velin92: also origin.path is always empty for file URLs

ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
```diff
@@ -98,6 +99,8 @@ private struct CallView: UIViewRepresentable {
             let userContentController = WKUserContentController()
             userContentController.add(WKScriptMessageHandlerWrapper(self), name: viewModelContext.viewState.messageHandler)
             
+            configuration.setValue(true, forKey: "allowUniversalAccessFromFileURLs")
+            configuration.preferences.setValue(true, forKey: "allowFileAccessFromFileURLs")
```
stefanceriu, line 103: Can we leave some comments here as to what these do please? Seeing as they're undocumented there's a good chance they will break in the future.

ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
pixlwave: Discovered while talking about this offline I this check was in
```
webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction)
```

We discussed that we will update that to add the extra check.

ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
Velin92: removed the universal one since it didn't seem to be necessary in favour of `allowFileAcessFromFileURLs` which while required, is also stricter and safer.

PR #3943, opened 2025-03-26
EC Embedding improvements

Velin92:
- [x] Always pass the rageshake URL, regardless of analytics
- [x] Use the `.startCall` intent to avoid seeing the join call button
- [x] Update to rc.4

### DIFF
--- a/ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
+++ b/ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
@@ -66,7 +66,11 @@ class ElementCallWidgetDriver: WidgetCapabilitiesProvider, ElementCallWidgetDriv
         self.deviceID = deviceID
     }
     
-    func start(baseURL: URL, clientID: String, colorScheme: ColorScheme, analyticsConfiguration: ElementCallAnalyticsConfiguration?) async -> Result<URL, ElementCallWidgetDriverError> {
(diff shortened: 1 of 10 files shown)


hughns (approved):
👍

ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
```diff
@@ -85,12 +85,12 @@ class ElementCallWidgetDriver: WidgetCapabilitiesProvider, ElementCallWidgetDriv
...
                                                                           font: nil,
                                                                           encryption: useEncryption ? .perParticipantKeys : .unencrypted,
-                                                                          intent: nil,
+                                                                          intent: .startCall,
                                                                           hideScreensharing: false,
                                                                           posthogUserId: nil,
                                                                           posthogApiHost: analyticsConfiguration?.posthogAPIHost,
                                                                           posthogApiKey: analyticsConfiguration?.posthogAPIKey,
-                                                                          rageshakeSubmitUrl: analyticsConfiguration?.rageshakeSubmitURL,
+                                                                          rageshakeSubmitUrl: Secrets.rageshakeServerURL,
```
pixlwave, line 93: Would you mind passing this in separately via the method parameters, and in the view model reading it from appSettings please? There's going to be runtime overrides of this value in the future.

pixlwave (approved):
LGTM 😎

PR #3950, opened 2025-03-27
a11y: better voice over for voice messages

### LINKED ISSUE #2568: Better VoiceOver implementation for audio messages.
### Steps to reproduce

Open a conversation and view a audio message with VoiceOver.

### Outcome

Currently, it are 3 elements. 
1. The play button
2. the position of the audio message.(Time elapsed)
3. How long the recording is in total.

This has some problems:
1. You cannot hear who send the voicemessage. This is a problem because sometimes people send audiorecording not recorded by that person itself.
2. It takes more time to read through the messages because you'll need to swipe more to go to the next message
3. You can open the message actions via the roter on each element, which makes it more convusing.
4. The time labels do not indicate clearly to VoiceOver user what they represent.

I believe that whatsapp has the best implementation for VoiceOver users. This is their format:

```{name} voice message, duration: {time}, current position: {time}```
Where
{name is ```your``` when you send that voicemessage
{name} is nothing when you are in a room with 1 other participent. (a direct message)
And when someone else sends a voicemessage in a group, it is formated like this: ```Voicemessage from {name}, duration: {time}, current position: {time}```

This is then all grouped into one element that VoiceOver users can easly hear. 
Whatsapp also has some rotor actions implemented for skipping a fer seconds or going back etc but that can be implemented when #2232  is fixed.

### Your phone model

iPhone 13 pro

### Operating system version

17.3.1

### Application version

1.5.10 (94)

### Homeserver

matrix.org

### Will you send logs?

No

Velin92:
fixes #2568

### DIFF
--- a/ElementX/Resources/Localizations/en-US.lproj/Localizable.strings
+++ b/ElementX/Resources/Localizations/en-US.lproj/Localizable.strings
@@ -1,11 +1,14 @@
 "Notification" = "Notification";
+"a11y_avatar" = "Avatar";
 "a11y_delete" = "Delete";
 "a11y_hide_password" = "Hide password";
+"a11y_join_call" = "Join call";
 "a11y_jump_to_bottom" = "Jump to bottom";
 "a11y_notifications_mentions_only" = "Mentions only";
 "a11y_notifications_muted" = "Muted";
 "a11y_page_n" = "Page %1$d";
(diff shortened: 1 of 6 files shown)


stefanceriu (approved):
Nice! 👍

PR #3961, opened 2025-03-31
Allow the services that are configured by secrets to be disabled.

pixlwave:
This is the first step as currently all our Secrets are non-optional. But this PR lays the ground-work so that Location Sharing, Bug Reports, Sentry and PostHog can all be disabled by not supplying a secret.

When the MapTiler API key is missing we:
- Return nil URLs to show the placeholder map.
- Disable tapping a map in the timeline.
- Disable location sharing in the composer.

When the Bug Report URL is missing we:
- Hide the button in the Settings and Start screens.

When both the Sentry and Analytics secrets are missing we:
- Hide the analytics prompt and settings when both analytics and sentry are disabled.


ElementX/Sources/Application/AppSettings.swift
```diff
@@ -98,7 +98,6 @@ final class AppSettings {
                   acceptableUseURL: URL,
                   privacyURL: URL,
                   supportEmailAddress: String,
-                  analyticsTermsURL: URL?,
```
stefanceriu: But why? 🙃

ElementX/Sources/Screens/HomeScreen/HomeScreenCoordinator.swift
```diff
@@ -87,7 +87,7 @@ final class HomeScreenCoordinator: CoordinatorProtocol {
     
     func start() {
         #if !DEBUG
-        if bugReportService.crashedLastRun {
+        if bugReportService.crashedLastRun, bugReportService.lastCrashEventID != nil {
```
stefanceriu, line 90: Why not use `isEnabled`? 🤔

ElementX/Sources/Application/AppSettings.swift
```diff
@@ -223,10 +223,10 @@ final class AppSettings {
         
     // MARK: - Bug report
 
-    let bugReportServiceBaseURL: URL! = URL(string: Secrets.rageshakeServerURL)
+    let bugReportServiceBaseURL: URL? = URL(string: Secrets.rageshakeServerURL)! // swiftlint:disable:this force_unwrapping
```
stefanceriu, line 226: What's the plan here, will the secrets eventually become optional themselves and we stop force unwrapping?

ElementX/Sources/Application/AppSettings.swift
pixlwave: It needs an update in the submodule which I've got a separate branch for ;)

ElementX/Sources/Application/AppSettings.swift
pixlwave: Correct, the secrets will become optional so we could change this to
```
let bugReportServiceBaseURL: URL? = Secrets.rageshakeServerURL.map { URL(string: $0)! }
```
I.e. fail early on a secret assuming it is a valid value at build time, otherwise do nothing.

ElementX/Sources/Screens/HomeScreen/HomeScreenCoordinator.swift
stefanceriu: Nice, thank you 🙏

stefanceriu (approved):

PR #3964, opened 2025-04-01
Bump the RustSDK to v25.03.31

stefanceriu:
- adopt new async throwing methods
- use the sdk side TimelineStart virtual timeline item
- remove async from `directRoomForUserID` as it's not async on the rust side

### DIFF
--- a/ElementX/Sources/Services/Client/ClientProxy.swift
+++ b/ElementX/Sources/Services/Client/ClientProxy.swift
@@ -353,15 +353,13 @@ class ClientProxy: ClientProxyProtocol {
         try? await client.accountUrl(action: action).flatMap(URL.init(string:))
     }
     
-    func directRoomForUserID(_ userID: String) async -> Result<String?, ClientProxyError> {
-        await Task.dispatch(on: clientQueue) {
-            do {
-                let roomID = try self.client.getDmRoom(userId: userID)?.id()
-                return .success(roomID)
-            } catch {
-                MXLog.error("Failed retrieving direct room for userID: \(userID) with error: \(error)")
-                return .failure(.sdkError(error))
-            }
+    func directRoomForUserID(_ userID: String) -> Result<String?, ClientProxyError> {
+        do {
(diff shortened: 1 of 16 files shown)


pixlwave:
Really insightful review there from Copilot.

ElementX/Sources/Services/Client/ClientProxy.swift
pixlwave, line 1: lol how does that work? The block_on was replaced with a sync Rust call? Or was this never blocking 🙈

ElementX/Sources/Services/Timeline/TimelineController/TimelineController.swift
```diff
@@ -480,6 +478,8 @@ class TimelineController: TimelineControllerProtocol {
                 return SeparatorRoomTimelineItem(id: .virtual(uniqueID: uniqueID), timestamp: date)
             case .readMarker:
                 return ReadMarkerRoomTimelineItem(id: .virtual(uniqueID: uniqueID))
+            case .timelineStart:
+                return isDM ? nil : TimelineStartRoomTimelineItem(name: roomDisplayName)
```
pixlwave, line 482: *super* unimportant, but should we give this thing a publisher so it can update it's own name if someone renames the room? It might be visible if you create a room with a typo, then rename it but see that it still shows your typo in the timeline.

pixlwave (approved):
What a nice tidy-up! ✨

ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu: Must've been async at some point and we just forgot to update it or something. Honestly I didn't look into it too much

ElementX/Sources/Services/Timeline/TimelineController/TimelineController.swift
stefanceriu: I mean that's a really nice idea but also such an edge case. You could also argue that the SDK should configure it with a name and then refresh it if needed?

PR #3966, opened 2025-04-01
Make the secrets optional.

pixlwave:
Following on from #3961, this PR goes ahead and makes all of the secrets optional by:
- Using Pkl's optional `read?` method to allow null environment values.
- Using our [StaticCode](https://github.com/element-hq/pkl-tools/releases/tag/staticcode%401.0.0) Pkl renderer to skip the SwiftGen YAML step so that the secrets are properly typed as `String?`.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -429,7 +429,6 @@ end
 
 lane :config_secrets do
   Dir.chdir "../Secrets" do
-    sh("pkl eval -f yaml -o secrets.yml Secrets.pkl")
-    sh("swiftgen run yaml -n inline-swift5 --param enumName=Secrets -o Secrets.swift secrets.yml")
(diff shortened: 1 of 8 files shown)


fastlane/Fastfile
```diff
@@ -429,7 +429,6 @@ end
 
 lane :config_secrets do
   Dir.chdir "../Secrets" do
-    sh("pkl eval -f yaml -o secrets.yml Secrets.pkl")
-    sh("swiftgen run yaml -n inline-swift5 --param enumName=Secrets -o Secrets.swift secrets.yml")
+    sh("pkl eval -o Secrets.swift Secrets.pkl")
```
stefanceriu, line 432: 🙌

stefanceriu (approved):
Absolutely love it when a plan comes together! 👏

PR #3967, opened 2025-04-01
Add alerts with reason for kick and ban

### LINKED ISSUE #3965: [Task] Add confirmation dialog before kicking users
### Description

As a admin in the room i want to get a “are you sure” dialog when i kick somebody from the room so that i dont do this accidentally

### Acceptance criteria
*  confirmation dialog is available
* it is possible to add a reason to the kick
* the reason is not mandatory
* the reason is shown in the room as a state event if one has been entered

Velin92:
fixes #3965

This adds an alert confirmation screen with a text field when an admin/moderator tries either kick or ban another user.
These alerts also include the possibility of including a reason.

https://github.com/user-attachments/assets/beac41fb-74a9-44e8-a051-217fb80b5d85

### DIFF
--- a/ElementX/Resources/Localizations/en.lproj/Localizable.strings
+++ b/ElementX/Resources/Localizations/en.lproj/Localizable.strings
@@ -208,6 +208,7 @@
 "common_privacy_policy" = "Privacy policy";
 "common_reaction" = "Reaction";
 "common_reactions" = "Reactions";
+"common_reason" = "Reason";
 "common_recovery_key" = "Recovery key";
 "common_refreshing" = "Refreshing…";
 "common_replying_to" = "Replying to %1$@";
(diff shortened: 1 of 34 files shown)


ElementX/Resources/Localizations/en.lproj/Localizable.strings
```diff
@@ -875,6 +876,9 @@
 "screen_room_member_list_ban_member_confirmation_title" = "Are you sure you want to ban this member?";
 "screen_room_member_list_banned_empty" = "There are no banned users in this room.";
 "screen_room_member_list_banning_user" = "Banning %1$@";
+"screen_room_member_list_kick_member_confirmation_action" = "Remove";
```
pixlwave, line 879: There's an `action.remove` or we should resolve the duplicity on this to avoid translators needing to double translate.

ElementX/Sources/Screens/RoomMemberListScreen/RoomMembersListScreenViewModel.swift
```diff
@@ -51,9 +51,31 @@ class RoomMembersListScreenViewModel: RoomMembersListScreenViewModelType, RoomMe
...
-            Task { await kickMember(member) }
+            var reason: String?
+            let binding: Binding<String> = .init(get: { reason ?? "" },
+                                                 set: { reason = $0.isEmpty ? nil : $0 })
+            state.bindings.alertInfo = .init(id: .kickConfirmation,
+                                             title: L10n.screenRoomMemberListKickMemberConfirmationTitle,
+                                             message: L10n.screenRoomMemberListKickMemberConfirmationDescription,
+                                             primaryButton: .init(title: L10n.actionCancel, role: .cancel) { },
+                                             secondaryButton: .init(title: L10n.screenRoomMemberListKickMemberConfirmationAction) { [weak self] in Task { await self?.kickMember(member, reason: reason) } },
+                                             textFields: [.init(placeholder: "",
```
pixlwave: Should the placeholder not be `Reason`?

pixlwave (approved):
👌

PR #3971, opened 2025-04-02
feat: report a room

Velin92:
Reporting room implementaiton.
As of right now the API doesn't work properly and requires a BE fix, however in the meantime the flow is fully implemented, and the feature can only be enabled through a feature flag in developer options.

### DIFF
--- a/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinatorStateMachine.swift
+++ b/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinatorStateMachine.swift
@@ -15,49 +15,53 @@ class UserSessionFlowCoordinatorStateMachine {
         /// The initial state, used before the coordinator starts
         case initial
                 
-        /// Showing the home screen. The `selectedRoomID` represents the timeline shown on the detail panel (if any)
-        case roomList(selectedRoomID: String?)
+        /// Showing the home screen. The `roomListSelectedRoomID` represents the timeline shown on the detail panel (if any)
+        case roomList(roomListSelectedRoomID: String?)
                 
         /// Showing the feedback screen.
-        case feedbackScreen(selectedRoomID: String?)
+        case feedbackScreen(roomListSelectedRoomID: String?)
         
         /// Showing the settings screen
-        case settingsScreen(selectedRoomID: String?)
+        case settingsScreen(roomListSelectedRoomID: String?)
         
         /// Showing the recovery key screen.
-        case recoveryKeyScreen(selectedRoomID: String?)
+        case recoveryKeyScreen(roomListSelectedRoomID: String?)
         
         /// Showing the encryption reset flow.
-        case encryptionResetFlow(selectedRoomID: String?)
+        case encryptionResetFlow(roomListSelectedRoomID: String?)
         
         /// Showing the start chat screen
-        case startChatScreen(selectedRoomID: String?)
+        case startChatScreen(roomListSelectedRoomID: String?)
         
         /// Showing the logout flows
-        case logoutConfirmationScreen(selectedRoomID: String?)
+        case logoutConfirmationScreen(roomListSelectedRoomID: String?)
         
         /// Showing Room Directory Search screen
-        case roomDirectorySearchScreen(selectedRoomID: String?)
+        case roomDirectorySearchScreen(roomListSelectedRoomID: String?)
         
         /// Showing the user profile screen. This screen clears the navigation.
         case userProfileScreen
         
+        /// Showing the report room screen, for the given room identrifier
+        case reportRoomScreen(roomListSelectedRoomID: String?)
+        
         case shareExtensionRoomList(sharePayload: ShareExtensionPayload)
         
         /// The selected room ID from the state if available.
-        var selectedRoomID: String? {
+        var roomListSelectedRoomID: String? {
             switch self {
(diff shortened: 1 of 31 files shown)


ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinatorStateMachine.swift
```diff
@@ -194,6 +201,11 @@ class UserSessionFlowCoordinatorStateMachine {
             case (.shareExtensionRoomList, .dismissedShareExtensionRoomList):
                 return .roomList(selectedRoomID: nil)
                 
+            case (.roomList(let selectedRoomID), .presentReportRoomScreen):
+                return .reportRoomScreen(selectedRoomID: selectedRoomID)
+            case (.reportRoomScreen(let selectedRoomID), .dismissedReportRoomScreen):
+                return .roomList(selectedRoomID: selectedRoomID)
```
pixlwave, line 207: I don't think this assumption is right. If I'm on an iPad looking at room `123` and I long press on room `456` then we should present the report room screen for room `456`. And when dismissing report room on `456` we should return back to `.roomList(selected: 123`.

ElementX/Sources/Screens/ReportRoomScreen/ReportRoomScreenViewModel.swift
```diff
@@ -0,0 +1,92 @@
...
+        switch result {
+        case .success:
+            userIndicatorController.submitIndicator(.init(title: L10n.dialogRoomReportedAndLeft, iconName: "checkmark"))
+            actionsSubject.send(.dismiss(shouldLeaveRoom: true))
+        case .failure:
+            state.bindings.alert = .init(id: .leaveRoomFailed,
+                                         title: L10n.screenReportRoomLeaveFailedAlertTitle,
+                                         message: L10n.screenReportRoomLeaveFailedAlertMessage,
+                                         primaryButton: .init(title: L10n.actionDismiss, role: .cancel) { [weak self] in self?.actionsSubject.send(.dismiss(shouldLeaveRoom: false)) },
+                                         secondaryButton: .init(title: L10n.actionRetry) { [weak self] in Task { await self?.leaveRoom(showLoading: true) } })
```
pixlwave, line 75: Entirely a personal preference here, but I'd say that retry is the primary button and cancel the secondary button. (switching them here shouldn't re-order them because of the `.cancel` role).

ElementX/Sources/Screens/ReportRoomScreen/ReportRoomScreenViewModel.swift
```diff
@@ -0,0 +1,92 @@
...
+    }
+    
+    private static let loadingIndicatorIdentifier = "\(BugReportScreenCoordinator.self)-Loading"
+    
+    private func showLoadingIndicator() {
+        userIndicatorController.submitIndicator(UserIndicator(id: Self.loadingIndicatorIdentifier,
+                                                              type: .modal(progress: .indeterminate, interactiveDismissDisabled: true, allowsInteraction: false),
+                                                              title: L10n.commonLoading,
+                                                              persistent: true),
+                                                delay: .milliseconds(250))
```
pixlwave, line 86: I don't think we want the delay here or you'll have a chance to hit the report button twice.

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinatorStateMachine.swift
Velin92: `selectedRoomID` is in fact the state of the selected room coming from the current selected room in the room list. It has nothing to do with the actual room we are reporting, we are going to get back to the `selectedRoomID` of the initial "123" room. `reportRoomScreen(selectedRoomID: )` uses that parameter only to remember the room list selection, it has nothing to do with the actual room we are reporting.

ElementX/Sources/Screens/ReportRoomScreen/ReportRoomScreenViewModel.swift
Velin92: report might be quite fast and 0.25 is the average human  fastest reaction time.

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinatorStateMachine.swift
pixlwave: Ohhh, riiiight I see that now. Hmmm, it feels ambiguous in this case given the state is for a specific room too.

Could we change it to `roomRoomScreen(roomID: String, selectedRoomID: String)` even if we don't need it. That would help avoid confusion.

ElementX/Sources/Screens/ReportRoomScreen/ReportRoomScreenViewModel.swift
pixlwave: It's definitely possible to double tap something by mistake though. Other option would be to disable and re-enable the form if you want to keep the delay.

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinatorStateMachine.swift
Velin92: let's rename `selectedRoomID` to `roomListSelectedRoomID` to avoid further confusion

PR #3985, opened 2025-04-07
Fix UI test toggle tapping after compound accessibility change.

stefanceriu:


pixlwave (approved):
Weird, but yay for the fix!

PR #3987, opened 2025-04-07
Add support URLs to the app setting overrides.

pixlwave:
This should complete all of the overrides.

Also adds caching of any pkl packages.


stefanceriu (approved):
👌

PR #3997, opened 2025-04-08
Re-write integration test login for OIDC.

pixlwave:
Now that matrix.org is on MAS we need an updated login flow here.

Integration tests run here: https://github.com/element-hq/element-x-ios/actions/runs/14339510101


stefanceriu (approved):

IntegrationTests/Sources/Common.swift
```diff
@@ -90,6 +90,9 @@ extension XCUIApplication {
         // `Failed to scroll to visible (by AX action) Button` https://stackoverflow.com/a/33534187/730924
         profileButton.tapCenter()
         
+        // Make the logout button visible
```
pixlwave, line 93: ah lol

PR #4002, opened 2025-04-09
Only use the appGroupTemporaryDirectory to access a file from the share extension.

pixlwave:
This PR switches back to the plain `URL.temporaryDirectory` for everything other than when a file is sent from the share extension in to the main app. When the main app receives the file it immediately moves it into the `temporaryDirectory` and continues as before.

Also adds a step to clear out the `appGroupTemporaryDirectory` on every upgrade to tidy up anything that may already be in here or anything that may have been shared and resulted in the app crashing (and as such, was left orphaned in the appGroupTemporaryDirectory).

### DIFF
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -125,6 +125,9 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
         
         if let previousVersion = appSettings.lastVersionLaunched.flatMap(Version.init) {
             performMigrationsIfNecessary(from: previousVersion, to: currentVersion)
+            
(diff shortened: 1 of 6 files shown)


ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -399,6 +407,31 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
         userSessionStore.reset()
     }
     
+    /// Manually cleans up any files in the app group's `tmp` directory.
+    ///
+    /// **Note:** If there is a single file we consider it to bar an active share payload and it is ignored.
```
stefanceriu, line 412:
```suggestion
    /// **Note:** If there is a single file we consider it to be an active share payload and ignore it.
```

stefanceriu (approved):
Very nice, thanks for taking care of this! 👍

PR #4004, opened 2025-04-09
change: hiding timeline media has three options

### LINKED ISSUE #4003: [Task] Show timeline media should now also have the private only option

Velin92:
fixes #4003 

Now is possible to select three different timeline media hiding/showing options:
- Always show
- Show only for private rooms
- Always hide


ElementX/Sources/Application/AppSettings.swift
```diff
@@ -353,8 +353,14 @@ final class AppSettings {
...
+    var timelineMediaVisibility
 }
 
 extension AppSettings: CommonSettingsProtocol { }
+
+enum TimelineMediaVisibility: Codable {
+    case always
+    case privateOnly
+    case never
+}
```
Velin92, line 366: Now this technically breaks the previous implementation, since now the stored value is different.
I checked with product, they don't care that much if the user needs to set this value again, so I left it as it is.
However if we want to go that extra step and support a migration, what I could do is write a custom `decode` implementation for the `TimelineMediaVisibility` enum where we also convert a found `Bool` from `true` -> `never` and `false` -> `always` and use the same `hideTimelineMedia` key for the UserPreference

Velin92: WDYT is it worth doing this?

pixlwave (approved):
👌

PR #4009, opened 2025-04-10
Decline and block design tweaks

Velin92:
Some improvement to the design and the flow of the decline and block screen, also added some tests

### DIFF
--- a/ElementX/Sources/Screens/DeclineAndBlockScreen/DeclineAndBlockScreenModels.swift
+++ b/ElementX/Sources/Screens/DeclineAndBlockScreen/DeclineAndBlockScreenModels.swift
@@ -7,7 +7,7 @@
 
 import Foundation
 
-enum DeclineAndBlockScreenViewModelAction {
+enum DeclineAndBlockScreenViewModelAction: Equatable {
     case dismiss(hasDeclined: Bool)
 }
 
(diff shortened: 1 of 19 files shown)


ElementX/Sources/Screens/DeclineAndBlockScreen/DeclineAndBlockScreenViewModel.swift
```diff
@@ -96,4 +96,8 @@ class DeclineAndBlockScreenViewModel: DeclineAndBlockScreenViewModelType, Declin
     private func showError() {
```
stefanceriu, line 96: We have 3 different cases here, don't we need to be a bit more specific with the messages we show?

stefanceriu (approved):
nice, lgtm 👍

ElementX/Sources/Screens/DeclineAndBlockScreen/DeclineAndBlockScreenModels.swift
```diff
@@ -7,7 +7,7 @@
 
 import Foundation
 
-enum DeclineAndBlockScreenViewModelAction {
+enum DeclineAndBlockScreenViewModelAction: Equatable {
```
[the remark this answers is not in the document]
Velin92: allow comparison in tests, nothing special

ElementX/Sources/Screens/DeclineAndBlockScreen/DeclineAndBlockScreenViewModel.swift
Velin92: Checked with design, they do not want to overcomplicate things, since if the decline API passes is very unlikely the other two might fail. So they are okay with a generic error

PR #4013, opened 2025-04-11
Adopt the new start chat button design.

pixlwave:

- Updates Compound to the latest version.
- Handles the `.plain` button style changes to `.tertiary` and `.textLink`.
  - Note: The 2 remaining uses of `.textLink` aren't necessarily correct for text links but they were both the original users of `.plain` and added that style to have a Compound button without padding (which is what Text Link is for) so it maps up to me.
- Updates the snapshots (the positioning of the button outline changed so there's quite a lot).
- Finally, it uses the new `toolbarIcon` size to make the Start Chat button a super button that matches the new design.

Simulator Screenshot - iPhone 16 Pro - 2025-04-11 at 16 37 05


ElementX/Sources/Screens/KnockRequestsListScreen/View/KnockRequestCell.swift
pixlwave, line 1: Ended up making a few extra changes to the spacing on this one after finding [this cell](https://www.figma.com/design/7TqjqdMBaPpm3IavKsMYu6/Ask-to-join--Knocking-?node-id=3180-33458&m=dev) in Figma to look for the Ban button size. Noticed that there is now a spacing of 16 everywhere so I updated to match, do please shout if I shouldn't have done this.

Velin92 (approved):
LGTM thanks for the update on the ban button

PR #4020, opened 2025-04-14
Stop showing the canonical alias for invites in the room list.

### LINKED ISSUE #3974: [Task] Remove the potential room alias from invitation cells
Please see and discuss the details in the linked (private) issue.

pixlwave:
Closes #3974.

### DIFF
--- a/ElementX/Sources/Screens/HomeScreen/View/HomeScreenInviteCell.swift
+++ b/ElementX/Sources/Screens/HomeScreen/View/HomeScreenInviteCell.swift
@@ -138,7 +138,7 @@ struct HomeScreenInviteCell: View {
     }
     
     private var subtitle: String? {
-        room.isDirect ? room.inviter?.id : room.canonicalAlias
+        room.isDirect ? room.inviter?.id : nil
     }
     
     @ViewBuilder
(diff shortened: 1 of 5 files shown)


ElementX/Sources/Screens/HomeScreen/View/HomeScreenInviteCell.swift
```diff
@@ -138,7 +138,7 @@ struct HomeScreenInviteCell: View {
     }
     
     private var subtitle: String? {
-        room.isDirect ? room.inviter?.id : room.canonicalAlias
+        room.isDirect ? room.inviter?.id : nil
```
stefanceriu, line 141: Shouldn't we just always show the inviter instead?

pixlwave: We do already but the presentation is different depending on if it is a DM or not:

https://github.com/element-hq/element-x-ios/blob/5afdaaec024912d282548a2a713dc780f67d5e5c/ElementX/Sources/Screens/HomeScreen/View/HomeScreenInviteCell.swift#L82-L91

stefanceriu: Ah I see 👍

stefanceriu (approved):

PR #4021, opened 2025-04-14
Update to 25.04.14

Velin92:
Aside from some small API changes/renames, this allows notifications to have their media filtered based on the user setting and the join rule of the room they belong to.


ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu, line 1: Isn't it work printing all these Errors' `details`?

pixlwave (approved):
LGTM 👍

PR #4022, opened 2025-04-14
Prefix the user's own messages with 'You' in the room list.

pixlwave:
iOS equivalent to https://github.com/element-hq/element-x-android/pull/3547
Closes https://github.com/element-hq/element-x-ios/issues/3286


Velin92 (approved):
LGTM

PR #4027, opened 2025-04-14
Show internet connection warning when uploading keys on log out.

### LINKED ISSUE #3980: [Task] Loading Spinner is missing when logout takes some time
Please see and discuss the details in the linked issue.

pixlwave:
This PR makes 2 changes:
- Report the progress of uploaded keys when signing out.
- Ask the user to check their internet connection if the upload stalls for more than 2 seconds.

iOS equivalent to https://github.com/element-hq/element-x-android/pull/4538
Closes #3980

| Before | After (Waiting) | After (Stalled) | After (Uploading) |
| - | - | - | - |
| secureBackupLogoutConfirmationScreen Ongoing-iPhone-16-en-GB | secureBackupLogoutConfirmationScreen Waiting-iPhone-16-en-GB | secureBackupLogoutConfirmationScreen Stalled-iPhone-16-en-GB | secureBackupLogoutConfirmationScreen Ongoing-iPhone-16-en-GB |


pixlwave:
Converted this back to draft, I think I misinterpreted what was done in https://github.com/element-hq/element-x-android/pull/4538

pixlwave:
Ok, I've re-worked it so that the warning is only shown when no progress has been made to match Android.

ElementX/Sources/Services/SecureBackup/SecureBackupController.swift
```diff
@@ -202,38 +206,19 @@ class SecureBackupController: SecureBackupControllerProtocol {
...
-    }
-    
-    func onUpdate(status: EnableRecoveryProgress) {
-        onUpdateClosure(status)
-    }
-}
+extension SecureBackupControllerListener: BackupStateListener where T == BackupState { }
+extension SecureBackupControllerListener: RecoveryStateListener where T == RecoveryState { }
+extension SecureBackupControllerListener: EnableRecoveryProgressListener where T == EnableRecoveryProgress { }
+extension SecureBackupControllerListener: BackupSteadyStateListener where T == BackupUploadState { }
```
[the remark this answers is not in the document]
Velin92: oh no wait disregard what I have said, I see we also have an underlying type

ElementX/Sources/Screens/SecureBackup/SecureBackupLogoutConfirmationScreen/SecureBackupLogoutConfirmationScreenViewModel.swift
```diff
@@ -17,6 +17,8 @@ class SecureBackupLogoutConfirmationScreenViewModel: SecureBackupLogoutConfirmat
     // periphery:ignore - auto cancels when reassigned
     @CancellableTask
     private var keyUploadWaitingTask: Task<Void, Never>?
+    @CancellableTask
+    private var keyUploadStalledTask: Task<Void, Error>?
```
Velin92, line 21: So this is a cancellable task, however I see we never check for its cancellation in the Task itself.

ElementX/Sources/Screens/SecureBackup/SecureBackupLogoutConfirmationScreen/SecureBackupLogoutConfirmationScreenViewModel.swift
```diff
@@ -90,4 +88,27 @@ class SecureBackupLogoutConfirmationScreenViewModel: SecureBackupLogoutConfirmat
...
+        } else {
+            state.mode = .offline
+        }
+    }
+    
+    private func monitorUploadProgress() {
+        keyUploadStalledTask = Task { [weak self] in
+            try await Task.sleep(for: .seconds(2))
+            guard let self, case .waitingToStart(hasStalled: false) = state.mode else { return }
+            state.mode = .waitingToStart(hasStalled: true)
```
Velin92, line 111: as I said this is a bit weird, why having it as cancellable if we never check its cancellation state

ElementX/Sources/Screens/SecureBackup/SecureBackupLogoutConfirmationScreen/SecureBackupLogoutConfirmationScreenViewModel.swift
pixlwave: I put a `try` on the `Task.sleep`, so it throws when cancelled and won't continue.

ElementX/Sources/Screens/SecureBackup/SecureBackupLogoutConfirmationScreen/SecureBackupLogoutConfirmationScreenViewModel.swift
pixlwave: https://github.com/element-hq/element-x-ios/pull/4027#discussion_r2044164922 :)

Velin92, line 111: So this is just a sort of timeout, to check if the upload has started or not yet? Maybe we should add some comments here to maerk this, however I admit I am not a huge fan of the `updateMode` the `monitorUploadProgress` being in a different function and handled by separate tasks from the `waitForKeyBackupUpload` i feel there are too many tasks racing on accessing the same resource at different time without any wait.

I think that the best solution would be to have the `secureBackupController` expose a publisher of the upload progress (maybe nil if it is not uploading), and have a new publisher that combines both reachability and such a publisher to determine how it is going. Or something similar

ElementX/Sources/Screens/SecureBackup/SecureBackupLogoutConfirmationScreen/SecureBackupLogoutConfirmationScreenViewModel.swift
Velin92: The Apple documentation specifies that: `Cancelling a task does not automatically cause arbitrary functions on the task to stop running or throw errors.`

https://developer.apple.com/documentation/swift/task/cancel%28%29

Cancellation is just a dev tool, it won't do anything if we don't handle it directly ourselves by checking the cancellation state

Velin92: right, but you are not cancelling the sleep task since is not being stored, by overwriting the task that calls it you are not automatically also cancelling any internal task.

Velin92: > However, a function that doesn’t specifically check for cancellation will run to completion normally, even if the task it is running on is canceled. However, that function might still end early if it calls other code that handles cancellation by throwing and that function doesn’t handle the error.

So even if you cancel the task that calls the `sleep` this doesn't make the sleep cancel. At most the sleep could cancel because of an error or because you are canceling it directly, but you would need to store it.

pixlwave: The sleep function doesn't create it's own `Task`, it's just a static method on `Task` the same way you check `Task.isCancelled` and don't do `keyUploadStalledTask.isCancelled`

pixlwave: I had this hooked up to re-run `monitorUploadProgress` whenever a progress is set before, it works just fine.

```
try wait Task.sleep(for: duration)
```
is equivalent to

```
try? wait Task.sleep(for: duration)
guard !Task.isCancelled else { throw CancellationError }
```

and

```
try? wait Task.sleep(for: duration)
try Task.checkCancellation()
```

Velin92: I see because this Task throws an error, and is not a `Never`, that's okay, but then this means we have an issue on with the existing `private var keyUploadWaitingTask: Task<Void, Never>?` since we never check for cancellation in there. But as I said in the other comment I feel we would be able through a `for await` or combined publisher to unify these two flows.

pixlwave: Alright, done in 0aece9abcc18a09f67c3e0ac1da997301cbd7696

ElementX/Sources/Screens/SecureBackup/SecureBackupLogoutConfirmationScreen/SecureBackupLogoutConfirmationScreenViewModel.swift
pixlwave: Combined the 2 states in 0aece9abcc18a09f67c3e0ac1da997301cbd7696

Velin92: Way better but I think we can actually also get rid of this task entirely:

```
context.$viewState
  .map(\.mode)
  .removeDuplicates()
  .filter( $0 == .waitingToStart(hasStalled: false))
  .flatMap(backupUploadStateSubject)
  .removeDuplicates()
  .collect(.byTime(DispatchQueue.main, .seconds(2)))
  .sink { values in
        if values.count > 1 {
            // Value changed within 2 seconds
        } else {
            // No change within 2 seconds
        }
    }
    .store(in: &cancellables)
```

Velin92: up to you I just feel combine gives a more controlled approach than the task

ElementX/Sources/Screens/SecureBackup/SecureBackupLogoutConfirmationScreen/SecureBackupLogoutConfirmationScreenViewModel.swift
```diff
@@ -65,15 +61,15 @@ class SecureBackupLogoutConfirmationScreenViewModel: SecureBackupLogoutConfirmat
...
-                var result = await secureBackupController.waitForKeyBackupUpload()
+                var result = await secureBackupController.waitForKeyBackupUpload(uploadStateSubject: backupUploadStateSubject)
                 
                 if case .failure = result {
                     // Retry the upload first, conditions might have changed.
-                    result = await secureBackupController.waitForKeyBackupUpload()
+                    result = await secureBackupController.waitForKeyBackupUpload(uploadStateSubject: backupUploadStateSubject)
                 }
                 
                 guard case .success = result else {
```
Velin92, line 75: Also I noticed that the `guard !Task.isCancelled else { return }` is actually below here, but I think it should be placed after any await and before any state change

Velin92 (commented):
LGTM, but left two more suggestions

ElementX/Sources/Screens/SecureBackup/SecureBackupLogoutConfirmationScreen/SecureBackupLogoutConfirmationScreenViewModel.swift
pixlwave: Ah yeah good shout.

Velin92 (approved):
LGTM

PR #4029, opened 2025-04-15
Xcode16.3 support

Velin92:
Fastlane has also been updated to use 16.3 on CI.

Also preview tests needed to be generated over again since (not all of them, but most of them), since 18.4 renders them slightly differently (even if they look the same)

NOTE: Will handle UI Tests and Integration Tests in other PRs


stefanceriu (approved):
Real nice, well done! 👏

PR #4030, opened 2025-04-15
Refactor SecureBackupControllerListener into SDKListener and use it everywhere.

pixlwave:
This isn't as clean as the `SecureBackupControllerListener` implementation was so we're not saving *that* much code now, but the PR is still +122 -171.


stefanceriu (approved):
This is super slick, thank you! 👏

PR #4031, opened 2025-04-15
UI tests on xcode16.3

Velin92:
Updated UI tests to use Xcode 16.3, it's important to note that to make this work we're using a preview of the newer version of snapshot testing to allow them to work with UI tests that has yet to be released.


pixlwave (approved):
LGTM, although do we know if this commit is going to hang around for the long-term if there are any forks out there that lag behind?

PR #4032, opened 2025-04-16
Introduce a `TimelineItemThreadSummary` object

stefanceriu:
… to hold details about threads starting from that particular item

This patch introduces a thread summary object that will be available on main timeline messages that are the root for a given thread. It currently provides the latest message content and sender for that thread but it will grow to provide info on the number of replies, unreads etc.

It also add a new UI component called `TimelineThreadSummaryView` that makes use of this data and is in turn used by the bubbled styler to render it in the timeline.

The rest of the PR is about refactoring on the `RoomTimelineItemFactory` so that replies and thread summaries use similar paths and builders, and adding a feature flag for threads.

### DIFF
--- a/ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/View/DeveloperOptionsScreen.swift
+++ b/ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/View/DeveloperOptionsScreen.swift
@@ -30,7 +30,13 @@ struct DeveloperOptionsScreen: View {
                     }
                 }
             }
-                        
+            
+            Section("General") {
+                Toggle(isOn: $context.threadsEnabled) {
+                    Text("Threads")
+                }
+            }
+            
             Section("Room List") {
                 Toggle(isOn: $context.publicSearchEnabled) {
                     Text("Public search")
--- a/ElementX/Sources/Screens/Timeline/View/Threads/TimelineThreadSummaryView.swift
+++ b/ElementX/Sources/Screens/Timeline/View/Threads/TimelineThreadSummaryView.swift
@@ -120,7 +120,7 @@ struct TimelineThreadSummaryView: View {
                     .foregroundColor(.compound.textPrimary)
                     .accessibilityLabel(L10n.commonInReplyTo(sender?.disambiguatedDisplayName ?? senderID))
                 
-                Text(messagePreview)
+                Text(context.viewState.buildMessagePreview(formattedBody: formattedBody, plainBody: plainBody))
                     .font(.compound.bodyXS)
                     .foregroundColor(.compound.textSecondary)
                     .tint(.compound.textLinkExternal)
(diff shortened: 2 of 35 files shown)


ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/View/DeveloperOptionsScreen.swift
```diff
@@ -30,7 +30,13 @@ struct DeveloperOptionsScreen: View {
                     }
                 }
             }
-                        
+            
+            Section("General") {
```
pixlwave, line 34: Totally unimportant: maybe this should be a `Timeline` section seeing as threads are only accessible from the timeline.

ElementX/Sources/Screens/Timeline/View/Threads/TimelineThreadSummaryView.swift
```diff
@@ -0,0 +1,335 @@
...
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Compound
+import SwiftUI
+
+struct TimelineThreadSummaryView: View {
+    let threadSummary: TimelineItemThreadSummary?
+    var onTap: (() -> Void)?
```
pixlwave, line 13: Seeing as this view is part of the timeline, can't we just use `@Environment(\.timelineContext)` here for consistency with e.g. `Image`/`Video`/`Audio`/`FileRoomTimelineView`?

ElementX/Sources/Screens/Timeline/View/Threads/TimelineThreadSummaryView.swift
```diff
@@ -0,0 +1,335 @@
...
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Compound
+import SwiftUI
+
+struct TimelineThreadSummaryView: View {
+    let threadSummary: TimelineItemThreadSummary?
```
pixlwave, line 12: I don't think this needs to be optional looking at the `if let` in the styler?

ElementX/Sources/Services/Timeline/TimelineItems/EventBasedMessageTimelineItemProtocol.swift
```diff
@@ -6,6 +6,7 @@
 //
 
 import Foundation
+import MatrixRustSDK
```
pixlwave, line 9: I presume this one isn't needed.

PreviewTests/Sources/__Snapshots__/PreviewTests/timelineItemBubbledStylerView.Thread-summary-iPhone-16-en-GB.png
pixlwave, line 1: Not sure if you have designs for this yet, but question: is the thread summary intentionally not aligned with the outer edge of the message bubble?

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemFactory.swift
```diff
@@ -628,6 +641,47 @@ struct RoomTimelineItemFactory: RoomTimelineItemFactoryProtocol {
...
+        switch threadSummary.latestEvent() {
+        case .unavailable:
+            return .notLoaded
+        case .pending:
+            return .loading
+        case .ready(let senderID, let senderProfile, let content):
+            let sender = buildTimelineItemSender(senderID: senderID, senderProfile: senderProfile)
+            
+            var latestEventContent: TimelineEventContent
+            switch content {
```
pixlwave, line 656: can probably do the whole

```
let latestEventContent: TimelineEventContent = switch content {
```
thing here

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemFactory.swift
```diff
@@ -755,45 +811,48 @@ struct RoomTimelineItemFactory: RoomTimelineItemFactoryProtocol {
...
     
-    private func timelineItemReplyDetails(sender: TimelineItemSender, eventID: String, messageType: MessageType?) -> TimelineItemReplyDetails {
-        let replyContent: EventBasedMessageTimelineItemContentType
-        
+    private func buildMessageTimelineItemContent(messageType: MessageType?, senderID: String, senderDisplayName: String?) -> EventBasedMessageTimelineItemContentType {
         switch messageType {
         case .audio(let content):
             if content.voice != nil {
-                replyContent = .voice(buildAudioTimelineItemContent(content))
+                return .voice(buildAudioTimelineItemContent(content))
```
pixlwave, line 835: I *think* you can use implicit returns here

ElementX/Sources/Screens/Timeline/View/Threads/TimelineThreadSummaryView.swift
```diff
@@ -0,0 +1,335 @@
...
+                CompoundIcon(\.threads, size: .xSmall, relativeTo: .compound.bodyXS)
+                    .foregroundColor(.compound.iconSecondary)
+                
+                LoadableAvatarImage(url: sender?.avatarURL,
+                                    name: sender?.displayName,
+                                    contentID: sender?.id,
+                                    avatarSize: .user(on: .threadSummary),
+                                    mediaProvider: context.mediaProvider)
+                
+                Text(sender?.displayName ?? senderID)
```
pixlwave, line 126: Not for now but maybe worth adding a TODO comment if we need to disambiguate this in the future.

ElementX/Sources/Screens/Timeline/View/Threads/TimelineThreadSummaryView.swift
```diff
@@ -0,0 +1,335 @@
...
+            .background(Color.compound.bgSubtlePrimary)
+            .cornerRadius(8)
+        }
+        
+        /// The string shown as the message preview.
+        ///
+        /// This converts the formatted body to a plain string to remove formatting
+        /// and render with a consistent font size. This conversion is done to avoid
+        /// showing markdown characters in the preview for messages with formatting.
+        private var messagePreview: String {
```
pixlwave, line 148: Could we maybe do this inside of `TimelineItemThreadSummary`? Also I wonder if that could have a `lastMessageBody` and `lasyMessageFormattedBody` and then you can get rid of that huge switch from the view and place it inside the data model.

pixlwave (approved):
Hurray! 🧵

ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/View/DeveloperOptionsScreen.swift
stefanceriu: > only accessible from the timeline
... for now 😅

PreviewTests/Sources/__Snapshots__/PreviewTests/timelineItemBubbledStylerView.Thread-summary-iPhone-16-en-GB.png
stefanceriu: Yeah, we don't have any designs so I just did what's easiest for now.

PR #4033, opened 2025-04-16
Add the new bloom style under a feature flag.

pixlwave:
- Refactor the existing bloom into its own view modifier.
- Adds a view modifier based on the new bloom style with a parameter to choose new or old.
    - This implementation is more of an exploration, the style itself will definitely need tweaking and a review by design.
- Adds a feature flag in the developer options to control the bloom.

https://github.com/user-attachments/assets/93ecd360-912a-42ad-8736-26c2252ac03c


stefanceriu (approved):
Super nice cleanup! 👏

PR #4036, opened 2025-04-17
Refactor the NSE so that the original notification content is preserved between all the different processing steps

stefanceriu:
- building brand new content is probably interfering with how iOS syncs up delivered and seen notifications between devices
- simplify all the various steps and make the flows easier to follow


NSE/Sources/NotificationContentBuilder.swift
```diff
@@ -151,43 +139,41 @@ struct NotificationContentBuilder {
...
                                                         mimeType: content.info?.mimetype))
         default:
             break
         }
-        
-        return notificationContent
     }
     
-    private func addMediaAttachment(notificationContent: UNMutableNotificationContent,
+    private func addMediaAttachment(notificationContent: inout UNMutableNotificationContent,
```
pixlwave, line 176: Pointing out that this one doesn't really need to be `inout`, but also makes sense if it's just here for consistency :)

NSE/Sources/NotificationServiceExtension.swift
```diff
@@ -34,20 +32,12 @@ private let settings: CommonSettingsProtocol = AppSettings()
...
 class NotificationServiceExtension: UNNotificationServiceExtension {
-    private var contentHandler: ((UNNotificationContent) -> Void)!
-    private var notificationContent: UNMutableNotificationContent!
+    private var notificationHandler: NotificationHandler?
     
     private let appHooks = AppHooks()
-            
+    
     deinit {
-        cleanUp()
```
pixlwave, line 50: I don't understand what this was for, so slightly concerned it was an important workaround for something. But who knows without a comment 🤷‍♂️

pixlwave (approved):
Ok yeah, this looks great now 😎

PR #4041, opened 2025-04-17
Remove delivered notifications for rooms that have meantime become fully read

stefanceriu:
… and have messages newer than what's delivered
- this is a best effort solution to cleanup outdated notifications when using multiple clients and reading them elsewhere


ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
pixlwave, line 1: Maybe we should split this into 2 functions and a switch now?

ElementX/Sources/Services/Room/RoomSummary/RoomSummary.swift
```diff
@@ -40,6 +40,7 @@ struct RoomSummary {
     let avatarURL: URL?
     let heroes: [UserProfileProxy]
     let lastMessage: AttributedString?
+    let lastMessageDate: Date?
     let lastMessageFormattedTimestamp: String?
```
pixlwave, line 44: You know we should really remove this now, right 😇

ElementX/Sources/Services/Notification/Manager/NotificationManager.swift
```diff
@@ -128,6 +128,27 @@ final class NotificationManager: NSObject, NotificationManagerProtocol {
...
+            .filter { $0.hasUnreadMessages == false }
+            .reduce(into: [:]) { partialResult, roomSummary in
+                partialResult[roomSummary.id] = roomSummary.lastMessageDate
+            }
+        
+        let notificationsIdentifiers = await notificationCenter
+            .deliveredNotifications()
+            .filter {
+                if let roomID = $0.request.content.roomID, let date = roomsToLastMessageDates[roomID] {
+                    $0.date <= date
```
pixlwave, line 143: Just to help a bit with readability
```suggestion
            .filter { notification in
                if let roomID = notification.request.content.roomID, let lastMessageDate = roomsToLastMessageDates[roomID] {
                    notification.date <= lastMessageDate
```
Would be even nicer if it was a `guard let` too :)

pixlwave (approved):
Seems reasonable, I wonder if we should put it behind a FF first and get some people to test it?

ElementX/Sources/Services/Room/RoomSummary/RoomSummary.swift
stefanceriu: Done in db77d3477f93efaf100762f500dbd43830c6f797

pixlwave (approved):
🤞

ElementX/Sources/Mocks/RoomSummaryProviderMock.swift
```diff
@@ -79,6 +79,7 @@ extension RoomSummary {
                     avatarURL: nil,
                     heroes: [],
                     lastMessage: AttributedString("I do not wish to take the trouble to understand mysticism"),
+                    lastMessageDate: nil,
                     lastMessageFormattedTimestamp: "14:56",
```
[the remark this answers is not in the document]
stefanceriu: You saw nothing 👻

PR #4044, opened 2025-04-18
Report room and decline & block screens tweaks

Velin92:
- Use the copies: "Leave room" and "Report room" everywhere, also in DMs
- Report reason in report room is now mandatory.
- Report reason in decline and block is now mandatory if the report action is set to on.
- Generic alert for failure of main API in both report room and decline and block has been added


pixlwave (approved):
LGTM 👍

PR #4059, opened 2025-04-23
Remove the support email address from the OIDC configuration.

pixlwave:
The contacts property has been removed from the SDK in https://github.com/matrix-org/matrix-rust-sdk/pull/4789 / https://github.com/matrix-org/matrix-rust-sdk/pull/4958


pixlwave:
Going for this one without a review, doesn't seem in any way controversial 😇

PR #4070, opened 2025-04-28
Remove the Learn More URL for the sliding sync proxy.

pixlwave:
It linked to documentation about the old Sliding Sync proxy. Now that we're production ready, Sliding Sync is part of the spec and Synapse has native support, we don't need these docs anymore.


stefanceriu (approved):
🚀

PR #4074, opened 2025-04-28
Dismiss room invite notifications when rejecting them from the home screen.

### LINKED ISSUE #4069: Discard Invitation notification when the invitation was rejected in App
My problem is that I received a spam invite, I saw it in app and discard it but the notification stayed in the notification panel

When you join a room the invite notification is dicarded. I'd like it to be the same when you dismiss the invite without joining

pixlwave:
Normally we reject invitations when starting the room flow, but in the case of tapping reject on the home screen, these notifications would be left visible. 

Fixes #4069.


stefanceriu (approved):
Nice, lgtm! 👍

PR #4075, opened 2025-04-28
Possible fix for flaky UI tests

Velin92:
- Increased the timeout to check the OIDC alert
- Decreased the number of options for polls, only in tests


stefanceriu (approved):
lgtm, hope it works 🤞

Issue #3883, opened 2025-03-10
Adopt the new Observation Framework


stefanceriu:
Observation is a new framework introduced in WWDC 2023 which overhauls how observation is done in Swift and it's especially useful for SwiftUI. It's a better, more powerful way to ensure that views only respond to model changes for fields that they're actually using. This is in contrast to our current setup where every view within a screen is recomputed when any piece of view state changes.

We could benefit from all of that in our more complex components (e.g. timeline view, home screen etc.) where small changes cause large invalidations.

The plan would be to first modify our templates and use this for any and all new screens and then retrofit the existing screens on a case by case basis. Chances are we don't need to update all of them but that wouldn't hurt.

More information at:
* https://developer.apple.com/documentation/observation
* https://developer.apple.com/videos/play/wwdc2023/10149/

PR #4077, opened 2025-04-29
Update the template screen to use the new(ish) Observation framework.

pixlwave:
Initial step for #3883 this PR introduces a `StateStoreViewModelV2` who's `Context` is `@Observable` instead of conforming to `ObservableObject` and therefore doesn't need an `@Published` on it's `viewState` property. Other than that the template screen now uses this type and an `@Bindable` on the context to allow the bindable state to work.

The second commit adds a variant of `deferFulfillment` that works with `Observable`s.

What this PR doesn't solve:
- What to do about expecting published view state in preview snapshots (shouldn’t be too hard to make a similar solution as the `deferFulfillment` implementation though).

### DIFF
--- a/ElementX/Sources/Other/Extensions/XCTestCase.swift
+++ b/ElementX/Sources/Other/Extensions/XCTestCase.swift
@@ -48,6 +48,61 @@ extension XCTestCase {
         }
     }
     
+    /// XCTest utility that assists in observing an `Observable`'s property and deferring the fulfilment and results until some condition has been met.
+    /// - Parameters:
+    ///   - observable: The object to observe.
+    ///   - property: The object's property to wait on.
+    ///   - timeout: A timeout after which we give up.
+    ///   - message: An optional custom expectation message
+    ///   - until: callback that evaluates outputs until some condition is reached
+    /// - Returns: The deferred fulfilment to be executed after some actions and that returns the result of the publisher.
+    func deferFulfillment<Object: Observable, Value>(_ observable: Object,
+                                                     property: KeyPath<Object, Value> & Sendable,
+                                                     timeout: TimeInterval = 10,
+                                                     message: String? = nil,
+                                                     until condition: @escaping (Value) -> Bool) -> DeferredFulfillment<Value> {
+        var result: Result<Value, Error>?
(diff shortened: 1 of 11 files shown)


ElementX/Sources/Other/Extensions/XCTestCase.swift
```diff
@@ -48,6 +48,61 @@ extension XCTestCase {
...
+                expectation.fulfill()
+                hasFulfilled = true
+            }
+            
+            @discardableResult
+            func startObservation() -> Value {
+                withObservationTracking {
+                    observable[keyPath: property]
+                } onChange: {
+                    guard !Task.isCancelled else { return }
```
pixlwave, line 80: Hmmm ok I don't think this bit will actually work if a recursive call has been made as that happens within the DispatchQueue 🤔

stefanceriu: Can we maybe replace it with `hasFulfilled`? 🤔

ElementX/Sources/Other/SwiftUI/ViewModel/StateStoreViewModelV2.swift
```diff
@@ -0,0 +1,89 @@
...
+/// A common ViewModel implementation for handling of `State` and `ViewAction`s. Version 2 of this state
+/// store has been re-written to use Swift's new Observation framework.
+///
+///
+/// Generic type State is constrained to the BindableState protocol in that it may contain (but doesn't have to)
+/// a specific portion of state that can be safely bound to.
+/// If we decide to add more features to our state management (like doing state processing off the main thread)
+/// we can do it in this centralised place.
+@MainActor
+class StateStoreViewModelV2<State: BindableState, ViewAction> {
```
stefanceriu, line 21: I'm surprised by how little needs to actually change 👏

stefanceriu (approved):
Really really nice! 👏

ElementX/Sources/Other/SwiftUI/ViewModel/StateStoreViewModelV2.swift
pixlwave: Ditto! 😄

PR #4079, opened 2025-04-29
Updated SDK and improved report flow

Velin92:
the report room flow is now based on the matrix version and the new one will only be used if the SDK checks if the server supports it.

This means that the feature flags regarding it have been removed.

Some other small changes are:
- test for the report flows have been improved
- some tweaks were made to the send media actions that had slight change in their signature


stefanceriu (approved):
Nice, lgtm 👍

PR #4084, opened 2025-05-01
Fix the integration tests.

pixlwave:
The issues were 2-fold:

- On a fresh simulator (reproducible locally by erasing an existing one) the Web Authentication Session is sometimes slow to pick up the webcredentials association with element.io (used for the call back). This is fixed by looping on the Continue button until the association is made.
- When the WAS does appear, the keyboard was getting in the way of tapping on the password field.
Screen Recording 2025-05-01 at 11 36 44 am 2025-05-01 11_42_53 am
This is fixed by dismissing the keyboard before moving onto the next step.

A couple of extra tweaks:
- If the homeserver isn't set in the environment, don't change the default (makes local testing quicker).
- Refactor `tapCenter()` in to `tap(_ point: UnitPoint)` so you can `tap(.center)`, `tap(.top)` etc and uses this when tapping the web view to bring focus back to the app (so that it doesn't end up tapping the password field by mistake).
- Fixes the OIDC presenter to
    - not show an error when pressing Cancel *inside* the web view (a regression introduced in #4010).
    - show an error when the system cancels the web view (Apple weirdly uses the `canceledLogin` error code for this instead of a specific one).

### DIFF
--- a/ElementX/Sources/Other/Extensions/XCUIElement.swift
+++ b/ElementX/Sources/Other/Extensions/XCUIElement.swift
@@ -5,11 +5,12 @@
 // Please see LICENSE files in the repository root for full details.
 //
 
+import SwiftUI
 import XCTest
 
 extension XCUIElement {
     func clearAndTypeText(_ text: String, app: XCUIApplication) {
-        tapCenter()
+        tap(.center)
         
         app.showKeyboardIfNeeded()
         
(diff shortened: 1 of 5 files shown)


pixlwave:
Going to yolo this one too, to stop all the ❌s showing up every 12 hours.

PR #4086, opened 2025-05-02
Run the UI & Integration tests once a day instead of twice with a 6 hour offset.

pixlwave:
Simpler, less annoying, just as good in practice.


.github/workflows/integration-tests.yml
pixlwave, line 1: I also added the same Tues-Sat constraint as the UI tests have, seeing as we're not gathering performance data anymore.

stefanceriu (approved):
🚀

PR #4087, opened 2025-05-02
Add a tool (based on Periphery) that reports any unused strings.

pixlwave:
Also updates an error string that was talking about using the well-known for Sliding Sync.

Output based on a run just now: [Unused Strings.txt](https://github.com/user-attachments/files/20015557/Unused.Strings.txt)


stefanceriu (approved):
Whoa, this is super nice, thank you! 👏

PR #4101, opened 2025-05-07
Correctly interpret application state transitions for the screen lock when running on the Mac

stefanceriu:


pixlwave (approved):
Very nice, just tried it out and works well 👌

PR #4106, opened 2025-05-09
Send full user agent header to server during OIDC authentication and when viewing Account and Device management screens

hughns:
A customised user-agent HTTP header is currently sent by the app to the homeserver.

However, this custom header isn't sent when using the embedded `ASWebAuthenticationSession` web views due to the security/sandbox model that it uses.

This PR sets an additional `X-Element-User-Agent` header on the `ASWebAuthenticationSession` so that the server gets to see the customised user-agent.

A use case for this is to allow MAS to know if the My Account screens are being accessed in-app vs via the Safari system browser.

@sandhose does this seem reasonable/sensible to you?

### Pull Request Checklist

- [ ] Voiceover enabled.


Velin92 (approved):

Velin92:
LGTM, the branch needs to be rebased and some conflicts need to be solved

PR #4107, opened 2025-05-09
Remove support for building Alpha/PR (adhoc) builds

stefanceriu:
We've haven't really been using these and they're proving to be quite a chore to keep running. 
Also, we recently needed an external demo build and we chose to go with custom build on top of testflight instead.
As such it's been decided that we can go ahead and remove them.

Left to do once this PR is approved:
- [ ] Delete actions secrets from github
- [ ] Delete provisioning profiles from ASC


Gemfile.lock
```diff
@@ -246,13 +228,12 @@ GEM
 
 PLATFORMS
   arm64-darwin-22
+  arm64-darwin-24
```
pixlwave, line 231: Heh, went looking to see what this is about: https://github.com/rubygems/rubygems/issues/7407 🙈

pixlwave (approved):
LGTM! 🚀

PR #4110, opened 2025-05-09
Always open manage member sheet in timeline

Velin92:
WIP: waiting for an icon to unban

### DIFF
--- a/ElementX/Sources/Other/SwiftUI/Views/AvatarHeaderView.swift
+++ b/ElementX/Sources/Other/SwiftUI/Views/AvatarHeaderView.swift
@@ -110,6 +110,21 @@ struct AvatarHeaderView<Footer: View>: View {
         badges = isVerified ? [.verified] : []
     }
     
+    init(sender: TimelineItemSender,
+         avatarSize: Avatars.Size,
+         mediaProvider: MediaProviderProtocol? = nil,
+         onAvatarTap: ((URL) -> Void)? = nil,
+         @ViewBuilder footer: @escaping () -> Footer) {
+        let profile = UserProfileProxy(sender: sender)
+        
+        self.init(user: profile,
+                  isVerified: false,
+                  avatarSize: avatarSize,
+                  mediaProvider: mediaProvider,
+                  onAvatarTap: onAvatarTap,
+                  footer: footer)
+    }
+    
     private var badgesStack: some View {
         HStack(spacing: 8) {
             ForEach(badges, id: \.self) { badge in
--- a/ElementX/Sources/Screens/ManageRoomMemberSheet/ManageRoomMemberSheetModels.swift
+++ b/ElementX/Sources/Screens/ManageRoomMemberSheet/ManageRoomMemberSheetModels.swift
@@ -10,9 +10,29 @@ enum ManageRoomMemberSheetViewModelAction: Equatable {
 }
 
 struct ManageRoomMemberSheetViewState: BindableState {
-    let member: RoomMemberDetails
-    let canKick: Bool
-    let canBan: Bool
+    let details: ManageRoomMemberDetails
+    
+    var canKick: Bool {
+        guard case let .memberDetails(_, canKick, _) = details else {
+            return false
+        }
+        return canKick
+    }
+    
+    var canBanAndUnban: Bool {
+        guard case let .memberDetails(_, _, canBan) = details else {
+            return false
+        }
+        return canBan
+    }
+    
+    var isMemberBanned: Bool {
+        guard case let .memberDetails(member, _, _) = details else {
+            return false
+        }
+        
+        return member.isBanned
+    }
     
     var bindings = ManageRoomMemberSheetViewStateBindings()
 }
@@ -24,10 +44,26 @@ struct ManageRoomMemberSheetViewStateBindings {
 enum ManageRoomMemberSheetViewAlertType {
     case kick
     case ban
+    case unban
 }
 
 enum ManageRoomMemberSheetViewAction {
     case kick
     case ban
+    case unban
     case displayDetails
+}
+
+enum ManageRoomMemberDetails {
+    case memberDetails(roomMember: RoomMemberDetails, canKick: Bool, canBanAndUnban: Bool)
+    case senderDetails(sender: TimelineItemSender)
+    
+    var id: String {
+        switch self {
+        case let .memberDetails(roomMember, _, _):
+            return roomMember.id
+        case let .senderDetails(sender):
+            return sender.id
+        }
+    }
 }
--- a/ElementX/Sources/Screens/ManageRoomMemberSheet/ManageRoomMemberSheetViewModel.swift
+++ b/ElementX/Sources/Screens/ManageRoomMemberSheet/ManageRoomMemberSheetViewModel.swift
@@ -22,9 +22,7 @@ class ManageRoomMemberSheetViewModel: ManageRoomMemberSheetViewModelType, Manage
         actionsSubject.eraseToAnyPublisher()
     }
     
-    init(member: RoomMemberDetails,
-         canKick: Bool,
-         canBan: Bool,
+    init(details: ManageRoomMemberDetails,
          roomProxy: JoinedRoomProxyProtocol,
          userIndicatorController: UserIndicatorControllerProtocol,
          analyticsService: AnalyticsService,
@@ -32,7 +30,7 @@ class ManageRoomMemberSheetViewModel: ManageRoomMemberSheetViewModelType, Manage
         self.userIndicatorController = userIndicatorController
         self.roomProxy = roomProxy
         self.analyticsService = analyticsService
-        super.init(initialViewState: .init(member: member, canKick: canKick, canBan: canBan), mediaProvider: mediaProvider)
+        super.init(initialViewState: .init(details: details), mediaProvider: mediaProvider)
     }
     
     override func process(viewAction: ManageRoomMemberSheetViewAction) {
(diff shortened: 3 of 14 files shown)


ElementX/Sources/Other/SwiftUI/Views/AvatarHeaderView.swift
```diff
@@ -110,6 +110,21 @@ struct AvatarHeaderView<Footer: View>: View {
         badges = isVerified ? [.verified] : []
     }
     
+    init(sender: TimelineItemSender,
```
pixlwave, line 113: Could we call this first parameter `unknownSender`, `loadingSender`, `senderWhileLoadingMember` or similar to indicate that this really shouldn't be used if possible.

pixlwave: Orrrr, Could we also provide a publisher for the room member so that the view could be automatically updated afterwards?

Ah, I think that might have to happen in the manage member sheet view model, but we should rename the parameter either way.

ElementX/Sources/Screens/ManageRoomMemberSheet/ManageRoomMemberSheetModels.swift
```diff
@@ -10,9 +10,35 @@ enum ManageRoomMemberSheetViewModelAction: Equatable {
...
+        // This is a best effort check, if we haven't fetched the member yet we assume we can peform the action
+        guard case let .memberDetails(member) = details else {
+            return false
+        }
+        
+        return !member.isActive || permissions.ownPowerLevel <= member.powerLevel
+    }
+    
+    var isMemberBanned: Bool {
+        // This is a best effort check, if we haven't fetched the member yet we assume we can peform the action
```
pixlwave, line 35:
```suggestion
        // This is a best effort check, if we haven't fetched the member yet we assume the member is not banned
```

ElementX/Sources/Screens/ManageRoomMemberSheet/ManageRoomMemberSheetModels.swift
```diff
@@ -24,10 +50,32 @@ struct ManageRoomMemberSheetViewStateBindings {
...
+    case senderDetails(sender: TimelineItemSender)
+    
+    var id: String {
+        switch self {
+        case let .memberDetails(roomMember):
+            return roomMember.id
+        case let .senderDetails(sender):
+            return sender.id
+        }
+    }
```
pixlwave, line 74: Can we use implicit returns here?
```suggestion
    var id: String {
        switch self {
        case let .memberDetails(roomMember):
           roomMember.id
        case let .senderDetails(sender):
           sender.id
        }
    }
```

ElementX/Sources/Screens/ManageRoomMemberSheet/ManageRoomMemberSheetModels.swift
```diff
@@ -10,9 +10,35 @@ enum ManageRoomMemberSheetViewModelAction: Equatable {
 }
 
 struct ManageRoomMemberSheetViewState: BindableState {
-    let member: RoomMemberDetails
-    let canKick: Bool
-    let canBan: Bool
+    let details: ManageRoomMemberDetails
```
pixlwave, line 13: Should this be called `memberDetails` to be more descriptive?

ElementX/Sources/Screens/ManageRoomMemberSheet/ManageRoomMemberSheetViewModel.swift
```diff
@@ -43,11 +42,15 @@ class ManageRoomMemberSheetViewModel: ManageRoomMemberSheetViewModelType, Manage
...
             actionsSubject.send(.dismiss(shouldShowDetails: true))
+        case .unban:
+            displayAlert(.unban)
         }
     }
     
     private func displayAlert(_ alertType: ManageRoomMemberSheetViewAlertType) {
-        let member = state.member
+        guard case let .memberDetails(member) = state.details else {
+            return
```
pixlwave, line 52: Might be useful to have a comment here saying that we don't show the buttons.

ElementX/Sources/Screens/ManageRoomMemberSheet/ManageRoomMemberSheetViewModel.swift
```diff
@@ -121,5 +144,12 @@ class ManageRoomMemberSheetViewModel: ManageRoomMemberSheetViewModelType, Manage
...
 extension ManageRoomMemberSheetViewModel: Identifiable {
-    var id: String { state.member.id }
+    var id: String {
+        switch state.details {
+        case .memberDetails(let member):
+            member.id
+        case .senderDetails(let sender):
+            sender.id
+        }
+    }
```
pixlwave, line 154: I think you can just do `state.details.id` here now?

ElementX/Sources/Screens/ManageRoomMemberSheet/View/ManageRoomMemberSheetView.swift
```diff
@@ -28,22 +37,34 @@ struct ManageRoomMemberSheetView: View {
...
+                                .disabled(context.viewState.isBanUnbanDisabled)
+                    } else {
+                        ListRow(label: .default(title: L10n.screenBottomSheetManageRoomMemberBan,
+                                                icon: \.block,
+                                                role: .destructive),
+                                kind: .button {
+                                    context.send(viewAction: .ban)
+                                })
+                                .disabled(context.viewState.isBanUnbanDisabled)
+                    }
```
pixlwave, line 67: Pleeease can we use ternaries rather than duplicating the view layout to its easier to read the view 🥺

(Ideally putting the title/icon etc in the view state as banUnbanTitle etc)
```suggestion
                    let title = context.viewState.isMemberBanned ? L10n.screenBottomSheetManageRoomMemberUnban : L10n.screenBottomSheetManageRoomMemberBan
                    let icon: KeyPath = context.viewState.isMemberBanned ? \.restart : \.block
                    let action: ViewAction = context.viewState.isMemberBanned ? .unban : .ban
                    
                    ListRow(label: .default(title: title,
                                            icon: icon,
                                            role: .destructive),
                            kind: .button {
                                context.send(viewAction: action)
                            })
                                .disabled(context.viewState.isBanUnbanDisabled)
```

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
```diff
@@ -271,39 +271,40 @@ class TimelineViewModel: TimelineViewModelType, TimelineViewModelProtocol {
...
+                                                       mediaProvider: mediaProvider)
+        } else {
+            viewModel = ManageRoomMemberSheetViewModel(details: .senderDetails(sender: sender),
+                                                       permissions: .init(canKick: state.canCurrentUserKick,
+                                                                          canBan: state.canCurrentUserBan,
+                                                                          ownPowerLevel: currentUserProxy?.powerLevel ?? 0),
+                                                       roomProxy: roomProxy,
+                                                       userIndicatorController: userIndicatorController,
+                                                       analyticsService: analyticsService,
+                                                       mediaProvider: mediaProvider)
```
pixlwave, line 293: Again we don't need the whole `if { } else { }` here for 1 parameter. Let's construct that and only have 1 view model init here :)

pixlwave (approved):
LGTM if we really want to show this sheet all of the time 😅

ElementX/Sources/Screens/ManageRoomMemberSheet/ManageRoomMemberSheetModels.swift
Velin92: well is not necessarily a member yet... it can also be a sender (which may not be a member anymore)

ElementX/Sources/Other/SwiftUI/Views/AvatarHeaderView.swift
Velin92: well we discussed about this with design, we would then need to introduce a spinner or a skeleton for this specific edge case which only happens for large rooms and only the first time they're loaded, so we didn't think is worth updating the sheet, the use case, is someone having to intervene ASAP to ban a malicious user which usually is not an admin or a moderator

pixlwave: Yeah fair. Lets just make sure it is clear that `TimelineItemSender` isn't a preferred init here for anyone who uses this in the future :)

ElementX/Sources/Screens/ManageRoomMemberSheet/ManageRoomMemberSheetModels.swift
pixlwave: What about `userDetails`? Something to indicate it is details about a person 😅

ElementX/Sources/Screens/ManageRoomMemberSheet/ManageRoomMemberSheetViewModel.swift
Velin92: actually need to refactor this too

ElementX/Sources/Other/SwiftUI/Views/AvatarHeaderView.swift
Velin92: I'll add a doc comment then

ElementX/Sources/Screens/ManageRoomMemberSheet/ManageRoomMemberSheetModels.swift
```diff
@@ -62,14 +62,23 @@ enum ManageRoomMemberSheetViewAction {
...
+        }
+    }
+    
+    var managedMember: ManagedRoomMember {
+        switch self {
+        case let .memberDetails(roomMember):
+            ManagedRoomMember(id: roomMember.id, name: roomMember.name)
+        case let .loadingMemberDetails(sender):
+            ManagedRoomMember(id: sender.id, name: sender.displayName)
         }
```
pixlwave, line 82: If this was `var name`, could we get away without the extra type, instead passing around a `ManageRoomMemberDetails` everywhere we need it?

ElementX/Sources/Other/SwiftUI/Views/AvatarHeaderView.swift
pixlwave: Would still be nice to change the parameter label to one of the options above, as it's easy to not see/read documentation.

pixlwave (approved):

ElementX/Sources/Screens/ManageRoomMemberSheet/ManageRoomMemberSheetModels.swift
Velin92: i guess that's also fine, maybe I can just have the function take two arguments

PR #4114, opened 2025-05-12
Fix room list heroes label format

stefanceriu:
- change from `Alice, Bob and Charlie` to `Alice, Bob, Charlie and 42 others` (when applicable)


ElementX/Resources/Localizations/es.lproj/Localizable.stringsdict
pixlwave, line 1: Noting that this change breaks the rule that common keys should match the string.

ElementX/Sources/Services/Room/RoomSummary/RoomSummary.swift
```diff
@@ -81,6 +84,17 @@ extension RoomSummary: CustomStringConvertible {
             return alias
         }
         
+        guard heroes.count > 0 else {
+            return ""
+        }
+        
+        if activeMembersCount > 0 {
+            var components = heroes.compactMap(\.displayName)
+            components.append(L10n.commonManyMembers(Int(activeMembersCount)))
```
pixlwave, line 93: Shouldn't this be something like
```suggestion
            let othersCount = Int(activeMembersCount) - heroes.count
            if othersCount > 0 {
                components.append(L10n.commonManyMembers(othersCount))
            }
```

pixlwave (approved):
LGTM with one small suggestion on the logic.

ElementX/Resources/Localizations/es.lproj/Localizable.stringsdict
stefanceriu: Seeing as there are others in there that don't respect this rule and I don't know what else to actually call it I'm going to go ahead with it as it is.

PR #4122, opened 2025-05-15
Observe room info updates and automatically dismiss the room if meanwhile left or banned

stefanceriu:


pixlwave (approved):
Should we add a test to `RoomFlowCoordinatorTests` that validates this behaviour?

PR #4123, opened 2025-05-15
Fix a bug where accepting a DM invite would be accepted as a regular room.

### LINKED ISSUE #3958: Starting a new DM creates a group for the invitee and a DM for the requestor, rather than a DM for both.
### Steps to reproduce

1. Will starts a new chat with John via the top right icon.
2. John accepts the invite
3. Will sees the chat as a DM in Element X
4. John sees the chat as a group in Element X

This is noticeable since Will sees John's profile photo in the room list, whereas John does not see Will's. 
It is also noticeable in the difference both have when tapping into the room options:

Image

Image

Perhaps related, the invite also looks like spam when John sees it in his room list (a friend of mine for instance cancelled legit invites of mine for this reason):

Image

#### What did you expect?
Both parties see the profile photo of the other person in their room list.
Element X treats both ends of the DM consistently.
The invite in the invitee's room list doesn't show the room id.

#### What happened instead?
The requestor sees the room as a DM 
The invitee sees the room as a group
The initial invite appears spam-like as it shows the room id 

### Your phone model

16

### Operating system version

18.3.2

### Application version

Version 25.03.7 (875)

### Homeserver

Synapse 1.126

### Will you send logs?

No

pixlwave:
By updating the SDK. Fixes #3958

### DIFF
--- a/ElementX.xcodeproj/project.pbxproj
+++ b/ElementX.xcodeproj/project.pbxproj
@@ -8659,7 +8659,7 @@
 			repositoryURL = "https://github.com/element-hq/matrix-rust-components-swift";
 			requirement = {
 				kind = exactVersion;
-				version = 25.05.09;
+				version = 25.05.15;
 			};
 		};
 		701C7BEF8F70F7A83E852DCC /* XCRemoteSwiftPackageReference "GZIP" */ = {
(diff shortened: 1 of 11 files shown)


stefanceriu (approved):
Lgtm, thanks!

PR #4125, opened 2025-05-16
Add mechanism for opening up a threaded timeline.

stefanceriu:
This patch introduces the components and tweaks necessary for opening a thread's timeline screen:
* a `RoomProxy` method that creates the rust side timeline and configures it accordingly
* a new builder method on the `TimelineControllerFactory`
* an action link from the previously introduced thread summary to the `RoomFlowCoordinator`
* a brand new `ThreadTimelineScreen` composing the `TimelineViewModel` (similar to the other timelines)
* `RoomFlowCoordinator` states and transitions for creating and pushing said screen
* tweaks for handling this new `TimelineKind`, for not showing thread reply details in the threaded timeline etc.

N.B.
* no filtering or live updates are being provided by the SDK yet
* timeline and timeline item actions are not handled properly on multiple levels
*

### DIFF
--- a/ElementX/Sources/Screens/ThreadTimelineScreen/ThreadTimelineScreenCoordinator.swift
+++ b/ElementX/Sources/Screens/ThreadTimelineScreen/ThreadTimelineScreenCoordinator.swift
@@ -8,7 +8,7 @@
 import Combine
 import SwiftUI
 
-struct PinnedEventsTimelineScreenCoordinatorParameters {
+struct ThreadTimelineScreenCoordinatorParameters {
     let roomProxy: JoinedRoomProxyProtocol
     let timelineController: TimelineControllerProtocol
     let mediaProvider: MediaProviderProtocol
@@ -20,7 +20,7 @@ struct PinnedEventsTimelineScreenCoordinatorParameters {
     let clientProxy: ClientProxyProtocol
 }
 
-enum PinnedEventsTimelineScreenCoordinatorAction {
+enum ThreadTimelineScreenCoordinatorAction {
     case dismiss
     case displayUser(userID: String)
     case presentLocationViewer(geoURI: GeoURI, description: String?)
@@ -28,22 +28,22 @@ enum PinnedEventsTimelineScreenCoordinatorAction {
     case displayRoomScreenWithFocussedPin(eventID: String)
 }
 
-final class PinnedEventsTimelineScreenCoordinator: CoordinatorProtocol {
-    private let parameters: PinnedEventsTimelineScreenCoordinatorParameters
-    private let viewModel: PinnedEventsTimelineScreenViewModelProtocol
+final class ThreadTimelineScreenCoordinator: CoordinatorProtocol {
+    private let parameters: ThreadTimelineScreenCoordinatorParameters
+    private let viewModel: ThreadTimelineScreenViewModelProtocol
     private let timelineViewModel: TimelineViewModelProtocol
     
     private var cancellables = Set<AnyCancellable>()
  
-    private let actionsSubject: PassthroughSubject<PinnedEventsTimelineScreenCoordinatorAction, Never> = .init()
-    var actions: AnyPublisher<PinnedEventsTimelineScreenCoordinatorAction, Never> {
+    private let actionsSubject: PassthroughSubject<ThreadTimelineScreenCoordinatorAction, Never> = .init()
+    var actions: AnyPublisher<ThreadTimelineScreenCoordinatorAction, Never> {
         actionsSubject.eraseToAnyPublisher()
     }
     
-    init(parameters: PinnedEventsTimelineScreenCoordinatorParameters) {
+    init(parameters: ThreadTimelineScreenCoordinatorParameters) {
         self.parameters = parameters
         
-        viewModel = PinnedEventsTimelineScreenViewModel(analyticsService: ServiceLocator.shared.analytics)
+        viewModel = ThreadTimelineScreenViewModel()
         timelineViewModel = TimelineViewModel(roomProxy: parameters.roomProxy,
                                               timelineController: parameters.timelineController,
                                               mediaProvider: parameters.mediaProvider,
@@ -59,20 +59,6 @@ final class PinnedEventsTimelineScreenCoordinator: CoordinatorProtocol {
     }
     
     func start() {
-        viewModel.actionsPublisher.sink { [weak self] action in
-            MXLog.info("Coordinator: received view model action: \(action)")
-            
-            guard let self else { return }
-            switch action {
-            case .viewInRoomTimeline(let itemID):
-                guard let eventID = itemID.eventID else { fatalError("A pinned event must have an event ID.") }
-                actionsSubject.send(.displayRoomScreenWithFocussedPin(eventID: eventID))
-            case .dismiss:
-                self.actionsSubject.send(.dismiss)
-            }
-        }
-        .store(in: &cancellables)
-        
         timelineViewModel.actions.sink { [weak self] action in
             MXLog.info("Coordinator: received timeline view model action: \(action)")
             guard let self else { return }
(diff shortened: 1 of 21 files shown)


ElementX/Sources/Screens/PinnedEventsTimelineScreen/PinnedEventsTimelineScreenCoordinator.swift
```diff
@@ -91,7 +91,7 @@ final class PinnedEventsTimelineScreenCoordinator: CoordinatorProtocol {
             // These other actions will not be handled in this view
             case .displayEmojiPicker, .displayReportContent, .displayCameraPicker, .displayMediaPicker,
                  .displayDocumentPicker, .displayLocationPicker, .displayPollForm, .displayMediaUploadPreviewScreen,
-                 .displayResolveSendFailure, .composer, .hasScrolled:
+                 .displayResolveSendFailure, .displayThread, .composer, .hasScrolled:
```
pixlwave, line 94: Do we have a configuration option for the pinned timeline to not render threads? Otherwise this should maybe send an action upwards to let the room flow coordinator display the thread?

ElementX/Sources/Screens/Timeline/View/Style/TimelineItemBubbledStylerView.swift
pixlwave, line 1: It's kind of annoying to have to make all these checks here in the view. Do you think we could do think we could filter out the `threadSummary`/`replyDetails` properties when building the timeline items?

ElementX/Sources/Services/Room/JoinedRoomProxy.swift
```diff
@@ -190,13 +190,32 @@ class JoinedRoomProxy: JoinedRoomProxyProtocol {
         }
     }
     
+    func threadTimeline(eventID: String) async -> Result<TimelineProxyProtocol, RoomProxyError> {
+        do {
+            let sdkTimeline = try await room.timelineWithConfiguration(configuration: .init(focus: .thread(rootEventId: eventID, numEvents: 20),
+                                                                                            filter: .all,
+                                                                                            internalIdPrefix: UUID().uuidString,
```
pixlwave, line 197: Same as with the media timeline, we don't need this internal ID here as it is only to control the animation when switching the underlying timeline provider on the same screen.
```suggestion
                                                                                            internalIdPrefix: nil,
```

ElementX/Sources/Services/Timeline/TimelineController/TimelineControllerFactory.swift
```diff
@@ -21,6 +21,23 @@ struct TimelineControllerFactory: TimelineControllerFactoryProtocol {
...
     
+    func buildThreadTimelineController(eventID: String,
+                                       roomProxy: JoinedRoomProxyProtocol,
+                                       timelineItemFactory: RoomTimelineItemFactoryProtocol,
+                                       mediaProvider: MediaProviderProtocol) async -> Result<TimelineControllerProtocol, TimelineFactoryControllerError> {
+        switch await roomProxy.threadTimeline(eventID: eventID) {
+        case .success(let timelineProxy):
+            return .success(TimelineController(roomProxy: roomProxy,
+                                               timelineProxy: timelineProxy,
+                                               initialFocussedEventID: nil,
```
pixlwave, line 32: Not necessary right now, but makes me think that presumably with a long enough thread we'll have to handle everything to start by showing the first item in the thread and not the last.

ElementX/Sources/Services/Timeline/TimelineProxy.swift
```diff
@@ -97,7 +97,7 @@ final class TimelineProxy: TimelineProxyProtocol {
         switch kind {
         case .live:
             return await paginateBackwardsOnLive(requestSize: requestSize)
-        case .detached, .media:
+        case .detached, .media, .thread:
```
pixlwave, line 100: Shouldn't this one have the same behaviour as `.pinned`? You can't back paginate a thread right?

ElementX/Sources/Services/Timeline/TimelineProxy.swift
```diff
@@ -596,7 +596,7 @@ final class TimelineProxy: TimelineProxyProtocol {
                 MXLog.error("Failed to subscribe to back pagination status with error: \(error)")
             }
             forwardPaginationStatusSubject.send(.timelineEndReached)
-        case .detached:
+        case .detached, .thread:
             // Detached timelines don't support observation, set the initial state ourself.
             backPaginationStatusSubject.send(.idle)
```
pixlwave, line 601: Presumably for that to work, we should also set `.timelineEndReached` for the back pagination status.

ElementX/Sources/Screens/ThreadTimelineScreen/ThreadTimelineScreenCoordinator.swift
```diff
@@ -0,0 +1,93 @@
...
+            case .displaySenderDetails(let userID):
+                actionsSubject.send(.displayUser(userID: userID))
+            case .displayMessageForwarding(let forwardingItem):
+                actionsSubject.send(.displayMessageForwarding(forwardingItem: forwardingItem))
+            case .displayLocation(_, let geoURI, let description):
+                actionsSubject.send(.presentLocationViewer(geoURI: geoURI, description: description))
+            case .viewInRoomTimeline(let eventID):
+                actionsSubject.send(.displayRoomScreenWithFocussedPin(eventID: eventID))
+            // These other actions will not be handled in this view
+            case .displayEmojiPicker, .displayReportContent, .displayCameraPicker, .displayMediaPicker,
```
pixlwave, line 76: No emoji… 🥺🥺🥺





(don't worry, I appreciate this is coming later)

ElementX/Sources/Screens/ThreadTimelineScreen/View/ThreadTimelineScreen.swift
```diff
@@ -0,0 +1,51 @@
...
+            .navigationTitle("Thread")
+            .navigationBarTitleDisplayMode(.inline)
+            .background(.compound.bgCanvasDefault)
+            .interactiveDismissDisabled()
+            .timelineMediaPreview(viewModel: $context.mediaPreviewViewModel)
+            .sheet(item: $timelineContext.manageMemberViewModel) {
+                ManageRoomMemberSheetView(context: $0.context)
+            }
+            .sheet(item: $timelineContext.debugInfo) { TimelineItemDebugView(info: $0) }
+            .sheet(item: $timelineContext.actionMenuInfo) { info in
```
pixlwave, line 26: Not for this PR, but we should probably make some kind of `TimelineView` wrapper at this point which could contain these 3 sheets to avoid the duplication across multiple screens.

pixlwave (approved):
Very nice! 🧵🧵

ElementX/Sources/Screens/PinnedEventsTimelineScreen/PinnedEventsTimelineScreenCoordinator.swift
stefanceriu: I imagine the thread summaries won't be populated on pinned timeline rust side but generally no, I don't think a pinned timeline should be able to interact with threads. 
I'll need to handle all of that once I start working on in-thread actions and what not.

ElementX/Sources/Services/Timeline/TimelineController/TimelineControllerFactory.swift
stefanceriu: Not entirely sure I'm following to be honest 🤔

ElementX/Sources/Screens/ThreadTimelineScreen/ThreadTimelineScreenCoordinator.swift
stefanceriu: Ha yeah, there's a ton of things missing. I would like to handle all of them separatelly and see if there's any refactoring I can do to simplify how we handle all these different timelines.

ElementX/Sources/Screens/ThreadTimelineScreen/View/ThreadTimelineScreen.swift
stefanceriu: Absolutely and I'm sure there's more we can do!

ElementX/Sources/Services/Timeline/TimelineProxy.swift
stefanceriu: It is supposed to be idle to begin with as threads can be backpaginated 🤔

ElementX/Sources/Screens/Timeline/View/Style/TimelineItemBubbledStylerView.swift
pixlwave: lovely 😄

ElementX/Sources/Services/Timeline/TimelineController/TimelineControllerFactory.swift
pixlwave: We discussed, seems odd to me but we always open a thread at the end so this can be ignored.

ElementX/Sources/Services/Timeline/TimelineProxy.swift
pixlwave: Ditto

PR #4128, opened 2025-05-17
Fix bubble animations overshooting in the timeline when SR/RRs are removed from an event

ara4n:
Top-align message bubbles within their VStack of `TimelineItemBubbledStylerView` by adding a Spacer, so that the bubble doesn't drift around as the row shrinks when the `TimelineItemStatusView` empties due to the message no longer having SR/RRs.  This causes the appearance of the "bubble animation overshoots its target" bug in https://github.com/element-hq/element-x-ios/issues/4127.

The addition of the Spacer empirically adds a gap between the bubbles, which we compensate for with some negative spacing on the VStack. 

Empirically this seems to work okay, although I did see some intermittent weirdness with the row at the top of the UITableView vanishing as the new row gets added at the bottom - feeling a bit like a cell reuse bug of some kind.  However, this seems very intermittent, and less distracting than the previous mangled animation (given you're not looking at the top of the screen when you hit send on a message).

### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Screens/Timeline/TimelineTableViewController.swift
+++ b/ElementX/Sources/Screens/Timeline/TimelineTableViewController.swift
@@ -292,7 +292,7 @@ class TimelineTableViewController: UIViewController {
             }
         }
         
-        // We only animate when there's a new last message, so its safe
+        // We only animate when there's a new last message, so it's safe
         // to animate from the bottom (which is the top as we're flipped).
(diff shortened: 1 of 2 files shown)


stefanceriu:
This has been hanging around for half a year now and nowhere near finished, I think it's time we close it.

PR #4132, opened 2025-05-19
Using 1.18.3 snapshot testing and allow UI tests to use the module properly

Velin92:
Snapshot testing can now be used in UI tests as long as a specific flag that automatically imports Swift Testing is removed from the UI tests project, more info here:

https://github.com/pointfreeco/swift-snapshot-testing/pull/989/files?short_path=343230a#diff-343230ae3a3d63635854d7031faf8ba78bb38d8a5fd511fe855a9490fe6fe580


stefanceriu (approved):
Great find! 👨‍🍳 👌

PR #4135, opened 2025-05-20
Integration tests: support for the bottom sheet

Velin92:
Integration tests update to support the change in the flow that has happened here: https://github.com/element-hq/element-x-ios/pull/4110


stefanceriu (approved):
lgtm, thank you!

PR #4136, opened 2025-05-20
EC: Native switch for audio outputs/inputs and earpiece

Velin92:
WIP: We need a new version of EC for this to work properly, in the meantime this can only be tested by using specific development branches using the Element Call link override in the Developer options.

This includes:
- Support for the earpiece.
- Support for switching audio output and inputs from a webview button (by opening the native `AVRoutePicker`).
- Support for the proximity monitor when the earpiece mode is enabled.

### DIFF
--- a/ElementX/Sources/Screens/CallScreen/CallScreenModels.swift
+++ b/ElementX/Sources/Screens/CallScreen/CallScreenModels.swift
@@ -16,7 +16,6 @@ enum CallScreenViewModelAction {
 }
 
 struct CallScreenViewState: BindableState {
-    let messageHandler: String
     let script: String?
     var url: URL?
     
@@ -39,8 +38,59 @@ enum CallScreenViewAction {
     case navigateBack
     case pictureInPictureWillStop
     case endCall
+    case ready
+    case outputDeviceSelected(deviceID: String)
 }
 
 enum CallScreenError: Error {
     case pictureInPictureNotAvailable
+}
+
+/// Identifies each event handler used by the CallScreen webview
+///
+/// The names of the enum need to always match the name of the handlers on the webview.
+enum CallScreenEventJSHandler: String, CaseIterable {
+    /// Widget actions's handler.
+    case widgetAction
+    /// Used to show the native AVRoutePickerView.
+    case showNativeOutputDevicePicker
+    /// Used to determine if the webview has selected the earpiece or not.
+    case onOutputDeviceSelect
+    
+    private var script: String {
+        switch self {
+        case .widgetAction:
+            """
+            window.addEventListener(
+                "message",
+                (event) => {
+                    let message = {data: event.data, origin: event.origin};
+                    if (message.data.response && message.data.api == "toWidget"
+                    || !message.data.response && message.data.api == "fromWidget") {
+                        window.webkit.messageHandlers.\(rawValue).postMessage(JSON.stringify(message.data));
+                    } else {
+                        console.log("-- skipped event handling by the client because it is send from the client itself.");
+                    }
+                },
+                false,
+            );
+            """
+        case .showNativeOutputDevicePicker:
+            """
+            window.controls.\(rawValue) = () => {
+                window.webkit.messageHandlers.\(rawValue).postMessage("");
+            };
+            """
+        case .onOutputDeviceSelect:
+            """
+            window.controls.\(rawValue) = (id) => {
+                window.webkit.messageHandlers.\(rawValue).postMessage(id);
+            };
+            """
+        }
+    }
+    
+    static var fullScript: String {
+        allCases.map(\.script).joined(separator: "\n")
+    }
 }
--- a/ElementX/Sources/Screens/CallScreen/CallScreenModels.swift
+++ b/ElementX/Sources/Screens/CallScreen/CallScreenModels.swift
@@ -25,7 +25,6 @@ struct CallScreenViewState: BindableState {
 }
 
 struct Bindings {
-    var javaScriptMessageHandler: ((Any) -> Void)?
     var javaScriptEvaluator: ((String) async throws -> Any)?
     var requestPictureInPictureHandler: (() async -> Result<Void, CallScreenError>)?
     
@@ -40,6 +39,7 @@ enum CallScreenViewAction {
     case endCall
     case mediaCapturePermissionGranted
     case outputDeviceSelected(deviceID: String)
+    case widgetAction(message: String)
 }
 
 enum CallScreenError: Error {
@@ -49,7 +49,7 @@ enum CallScreenError: Error {
 /// Identifies each event handler used by the CallScreen webview
 ///
 /// The names of the enum need to always match the name of the handlers on the webview.
-enum CallScreenJavascriptMessageName: String, CaseIterable {
+enum CallScreenJavaScriptMessageName: String, CaseIterable {
     /// Widget actions's handler.
     case widgetAction
     /// Used to show the native AVRoutePickerView.
@@ -90,7 +90,7 @@ enum CallScreenJavascriptMessageName: String, CaseIterable {
         }
     }
     
-    static func makeFullInjectionScript() -> String {
+    static var allCasesInjectionScript: String {
         allCases.map(\.postMessageScript).joined(separator: "\n")
     }
 }
--- a/ElementX/Sources/Screens/CallScreen/CallScreenModels.swift
+++ b/ElementX/Sources/Screens/CallScreen/CallScreenModels.swift
@@ -38,7 +38,7 @@ enum CallScreenViewAction {
     case navigateBack
     case pictureInPictureWillStop
     case endCall
-    case ready
+    case mediaCapturePermissionGranted
     case outputDeviceSelected(deviceID: String)
 }
 
@@ -49,7 +49,7 @@ enum CallScreenError: Error {
 /// Identifies each event handler used by the CallScreen webview
 ///
 /// The names of the enum need to always match the name of the handlers on the webview.
-enum CallScreenEventJSHandler: String, CaseIterable {
+enum CallScreenJavascriptMessageName: String, CaseIterable {
     /// Widget actions's handler.
     case widgetAction
     /// Used to show the native AVRoutePickerView.
@@ -57,7 +57,7 @@ enum CallScreenEventJSHandler: String, CaseIterable {
     /// Used to determine if the webview has selected the earpiece or not.
     case onOutputDeviceSelect
     
-    private var script: String {
+    private var postMessageScript: String {
         switch self {
         case .widgetAction:
             """
@@ -90,7 +90,7 @@ enum CallScreenEventJSHandler: String, CaseIterable {
         }
     }
     
-    static var fullScript: String {
-        allCases.map(\.script).joined(separator: "\n")
+    static func makeFullInjectionScript() -> String {
+        allCases.map(\.postMessageScript).joined(separator: "\n")
     }
 }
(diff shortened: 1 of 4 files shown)


ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
```diff
@@ -161,12 +169,35 @@ private struct CallView: UIViewRepresentable {
...
+        private func tapRoutePickerView() {
+            if let button = routePickerView.subviews.first(where: { $0 is UIButton }) as? UIButton {
+                button.sendActions(for: .touchUpInside)
+            }
         }
         
         // MARK: - WKUIDelegate
         
         func webView(_ webView: WKWebView, decideMediaCapturePermissionsFor origin: WKSecurityOrigin, initiatedBy frame: WKFrameInfo, type: WKMediaCaptureType) async -> WKPermissionDecision {
+            viewModelContext?.send(viewAction: .ready)
```
pixlwave: Shouldn't this should go after the guard so we don't send it if we denied media capture permissions? Then we can call it `mediaCapturePermissionGranted` which is easier to understand than `ready`.

```suggestion
            viewModelContext?.send(viewAction: .mediaCapturePermissionGranted)
```

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
```diff
@@ -132,10 +142,19 @@ class CallScreenViewModel: CallScreenViewModelType, CallScreenViewModelProtocol
         }
         
         elementCallService.tearDownCallSession()
+        UIDevice.current.isProximityMonitoringEnabled = false
```
pixlwave, line 142: Should we also do the same when starting picture in picture? I imagine you wouldn't the proximity sensor on if you tapped back in the call and wanted to find some information in a chat?

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
```diff
@@ -132,10 +142,19 @@ class CallScreenViewModel: CallScreenViewModelType, CallScreenViewModelProtocol
...
     }
     
     // MARK: - Private
     
+    private static let earpieceID = "earpiece-id"
```
pixlwave: I assume from the `-` that this is defined in the web app? Would be nice to add a doc comment saying it matches a hard-coded value in the web app.

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
```diff
@@ -242,23 +261,24 @@ class CallScreenViewModel: CallScreenViewModelType, CallScreenViewModelProtocol
...
-                || !message.data.response && message.data.api == "fromWidget") {
-                  window.webkit.messageHandlers.\(eventHandlerName).postMessage(JSON.stringify(message.data));
-                }else{
-                  console.log("-- skipped event handling by the client because it is send from the client itself.");
-                }
-            },
-            false,
-          );
-        """
+    private func setAvailableOutputDevices() async {
```
pixlwave: Could we call this `updateDevicesListInWebApp` or something? It is so hard to know if things are coming to/from web and/or to/from the Widget.

ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
```diff
@@ -75,6 +75,7 @@ private struct CallView: UIViewRepresentable {
         private var webView: WKWebView!
         private var pictureInPictureController: AVPictureInPictureController?
         private let pictureInPictureViewController: AVPictureInPictureVideoCallViewController
+        private var routePickerView: AVRoutePickerView!
```
pixlwave, line 78: So this view is always hidden and just exists for programmatic changes to the route? Could we document the how and the why on this if that's the case please? Because it's very confusing coming to this without context.

ElementX/Sources/Screens/CallScreen/CallScreenModels.swift
```diff
@@ -39,8 +38,59 @@ enum CallScreenViewAction {
...
 }
 
 enum CallScreenError: Error {
     case pictureInPictureNotAvailable
 }
+
+/// Identifies each event handler used by the CallScreen webview
+///
+/// The names of the enum need to always match the name of the handlers on the webview.
+enum CallScreenEventJSHandler: String, CaseIterable {
```
pixlwave: Please can we call this enum `CallScreenJavaScriptMessageName` please. It might be long, but this current name is confusing as it doesn't actually handle anything by itself.

```suggestion
enum CallScreenJavaScriptMessageName: String, CaseIterable {
```

ElementX/Sources/Screens/CallScreen/CallScreenModels.swift
```diff
@@ -39,8 +38,59 @@ enum CallScreenViewAction {
...
+    /// Widget actions's handler.
+    case widgetAction
+    /// Used to show the native AVRoutePickerView.
+    case showNativeOutputDevicePicker
+    /// Used to determine if the webview has selected the earpiece or not.
+    case onOutputDeviceSelect
+    
+    private var script: String {
```
pixlwave: And then call this `postMessageInjectionScript` so that it is clear when the script is used.
```suggestion
    private var postMessageInjectionScript: String {
```

ElementX/Sources/Screens/CallScreen/CallScreenModels.swift
```diff
@@ -39,8 +38,59 @@ enum CallScreenViewAction {
...
+            window.controls.\(rawValue) = (id) => {
+                window.webkit.messageHandlers.\(rawValue).postMessage(id);
+            };
+            """
+        }
+    }
+    
+    static var fullScript: String {
+        allCases.map(\.script).joined(separator: "\n")
+    }
```
pixlwave: And then can we make this into a function and call it something like `makePostMessageInjectionScriptForAllCases` please? Without the mention of all cases, it is hard to know at the call site that it is doing that.

pixlwave (changes requested):
I'm finding this one quite hard to follow (I've never found this screen easy to track the data-flow). I've added some initial comments that will hopefully help. Mind if you address those and then I do another review?

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
Velin92: I am checking with product and the EC team

Velin92: Okay checked we don't want to overcomplicate things, as long as earpiece mode is on we keep the proximity monitor enabled even when the user is using PiP Mode (we checked and the proximity sensor will still only work while the app is foregrounded, the app value is ignored when backgrounded). We will decide later based on user feedback

ElementX/Sources/Screens/CallScreen/CallScreenModels.swift
```diff
@@ -39,8 +38,59 @@ enum CallScreenViewAction {
...
 }
 
 enum CallScreenError: Error {
     case pictureInPictureNotAvailable
 }
+
+/// Identifies each event handler used by the CallScreen webview
+///
+/// The names of the enum need to always match the name of the handlers on the webview.
+enum CallScreenJavascriptMessageName: String, CaseIterable {
```
pixlwave: Based on the web view API this should be 👇
```suggestion
enum CallScreenJavaScriptMessageName: String, CaseIterable {
```

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
```diff
@@ -52,8 +52,7 @@ class CallScreenViewModel: CallScreenViewModelType, CallScreenViewModelProtocol
             widgetDriver = roomProxy.elementCallWidgetDriver(deviceID: deviceID)
         }
         
-        super.init(initialViewState: CallScreenViewState(messageHandler: Self.eventHandlerName,
-                                                         script: Self.eventHandlerInjectionScript,
+        super.init(initialViewState: CallScreenViewState(script: CallScreenJavascriptMessageName.makeFullInjectionScript(),
```
pixlwave: Can we change `Full` to `AllCases` here? Otherwise it doesn't really make much sense from the call-site.
```suggestion
        super.init(initialViewState: CallScreenViewState(script: CallScreenJavascriptMessageName.makeAllCasesInjectionScript(),
```

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
```diff
@@ -132,10 +142,20 @@ class CallScreenViewModel: CallScreenViewModelType, CallScreenViewModelProtocol
...
     
+    // This should always match the web app value
+    private static let earpieceID = "earpiece-id"
+    
+    private func handleOutputDeviceSelected(deviceID: String) {
+        let isEarpiece = deviceID == Self.earpieceID
+        MXLog.info("[handleOutputDeviceSelected] is earpiece: \(isEarpiece)")
+        UIDevice.current.isProximityMonitoringEnabled = isEarpiece
+    }
+
```
pixlwave: Maybe this should come after `setupCall` as that one's the entry point?

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
```diff
@@ -242,23 +262,24 @@ class CallScreenViewModel: CallScreenViewModelType, CallScreenViewModelProtocol
...
+        guard let currentOutput = AVAudioSession.sharedInstance().currentRoute.outputs.first else {
+            return
+        }
+        
+        let deviceList = if currentOutput.portType == .builtInSpeaker {
+            "{id: '\(currentOutput.uid)', name: '\(currentOutput.portName)', forEarpiece: true}"
+        } else {
+            // Doesn't matter because the switch is handled through the OS
+            "{id: 'dummy', name: 'dummy'}"
+        }
```
pixlwave: I don't understand why we update's the web `availableOutputDevices` if we only ever give them 1 device? Could you add some more comments explaining what is going on here?

ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
```diff
@@ -161,7 +170,31 @@ private struct CallView: UIViewRepresentable {
...
         
         func userContentController(_ userContentController: WKUserContentController, didReceive message: WKScriptMessage) {
-            viewModelContext?.javaScriptMessageHandler?(message.body)
+            guard let handlerID = CallScreenJavascriptMessageName(rawValue: message.name) else {
+                return
+            }
+            
+            switch handlerID {
+            case .widgetAction:
+                viewModelContext?.javaScriptMessageHandler?(message.body)
```
pixlwave: Do you think we should rename this `javaScriptMessageHandler` to `javaScriptWidgetActionHandler` now that we have multiple message handling paths?

pixlwave (approved):
Thank you for those changes, that definitely makes it easier to follow 🙏

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
pixlwave: Makes sense. I also now wonder if my assumption here was wrong anyway :)

ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
Velin92: Honestly I would refactor this a bit, it doesn't need to be a closure, it could just be a view action that we send that contains the body and handles it, not sure in fact why we're using a closure here.

PR #4137, opened 2025-05-20
Show an account provider picker on the server confirmation screen when required.

pixlwave:
Will only be shown if the app is configured with multiple account providers and restricted to not allow any other provider to be selected. Can be reviewed commit-by-commit.

https://github.com/user-attachments/assets/f410f6f0-dd9e-4a0e-a5aa-685bd2ef609e

### DIFF
--- a/ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
@@ -31,8 +31,6 @@ class AuthenticationFlowCoordinator: FlowCoordinatorProtocol {
         
         /// The initial screen shown when you first launch the app.
         case startScreen
-        /// The initial screen with the selection of account provider having been restricted.
-        case restrictedStartScreen
         
         /// The screen used for the whole QR Code flow.
         case qrCodeLoginScreen
@@ -55,7 +53,7 @@ class AuthenticationFlowCoordinator: FlowCoordinatorProtocol {
     
     enum Event: EventType {
         /// The flow is being started.
-        case start(allowOtherAccountProviders: Bool)
+        case start
         
         /// Modify the flow using the provisioning parameters in the `userInfo`.
         case applyProvisioningParameters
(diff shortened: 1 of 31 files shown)


ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
pixlwave, line 1: Decided to simplify this as discussed in https://github.com/element-hq/element-x-ios/pull/4131#discussion_r2097092485

Ah crap, that wasn't what the comment suggested 🙈

ElementX/Sources/Screens/AuthenticationStartScreen/AuthenticationStartScreenViewModel.swift
pixlwave, line 1: Follow-up to https://github.com/element-hq/element-x-ios/pull/4131#discussion_r2097124672

ElementX/Sources/Screens/Authentication/ServerConfirmationScreen/ServerConfirmationScreenViewModel.swift
```diff
@@ -87,6 +100,32 @@ class ServerConfirmationScreenViewModel: ServerConfirmationScreenViewModelType,
...
+        let homeserver = authenticationService.homeserver.value
+        guard homeserver.loginMode == .unknown || homeserver.address != accountProvider else {
+            await fetchLoginURLIfNeededAndContinue()
+            return
+        }
+        
+        // Note: We don't show the spinner until now as it isn't needed if the service is already
+        // configured and we're about to use password based login
+        startLoading()
+        defer { stopLoading() }
```
Velin92, line 118: I guess this code is reused across the 2 functions and could be extracted into a separate function

Velin92 (approved):
LGTM

ElementX/Sources/Screens/Authentication/ServerConfirmationScreen/ServerConfirmationScreenViewModel.swift
pixlwave: It's ever so slightly different. The first one checks whether the flow matches, the second checks whether the server is the same.

PR #4140, opened 2025-05-21
Adopt latest timeline API changes

stefanceriu:
The room list items no longer hold any timeline instances and they can now be created through the normal `room.timelineWithConfiguration` mechanism.
This also means that we can move the UTD hook away from the sync service and into the client itself but setting it is now fallible as it can only be set once.

This patch also moves the pinned timeline inline with the other timeline creation methods and has it report errors just like the others.


pixlwave (approved):
LGTM! 😎

PR #4142, opened 2025-05-22
Handle media previews and invite avatars through the account data

### LINKED ISSUE #3923: [Task] Set hide timeline media and invite avatars flag through the homeserver

Velin92:
Implemented APIs that update our media preview visibility and invite avatars visibility settings through account data.

The local settings are now discarded and a migration is implemented to check wether server configurations exist or not. In the latter case we simply push our current local configuration on the server to perform the migration, so that the user will still keep their setting.

This PR alo updates compound, so that is now possible to have loaders in inline pickers and toggles.

fixes #3923

### DIFF
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -30,11 +30,12 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
     private var userSession: UserSessionProtocol? {
         didSet {
             userSessionObserver?.cancel()
-            if userSession != nil {
+            if let userSession {
                 configureElementCallService()
                 configureNotificationManager()
                 observeUserSessionChanges()
                 startSync()
+                performSettingsToAccountDataMigration(userSession: userSession)
             }
         }
     }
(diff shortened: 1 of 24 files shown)


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -132,6 +135,16 @@ class ClientProxy: ClientProxyProtocol {
         verificationStateSubject.asCurrentValuePublisher()
     }
     
+    private let timelineMediaVisibilitySubject = CurrentValueSubject<TimelineMediaVisibility, Never>(.never)
+    var timelineMediaVisibilityPublisher: CurrentValuePublisher<TimelineMediaVisibility, Never> {
+        timelineMediaVisibilitySubject.asCurrentValuePublisher()
+    }
+    
+    private let hideInviteAvatarsSubject = CurrentValueSubject<Bool, Never>(true)
```
Velin92, line 143: The only case where this may return the initial value, is if the store wasn't initialised properly, since the first initial value from the SDK will always return either the stored value or the default value which is actually `on` (which means do not hide anything).

Not sure if I should bother too much with this considering that if the store doesn't work is an exceptional edge case... but I defaulted the app to hide everything in these cases.

The same happens for the `timelineMediaVisibility` and also for the NSE implementations.
WDYT? Should we keep this exceptional failure by handling it negatively, or should we handle it with the default values of the MSC? (where everything is displayed)

Velin92: We might be changing the MSC... so that the default matches the `off` case

ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -399,17 +400,49 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
             Tracing.deleteLogFiles()
             MXLog.info("Migrating to v1.6.7, log files have been wiped")
         }
+    }
+    
+    private func performSettingsToAccountDataMigration(userSession: UserSessionProtocol) {
```
stefanceriu, line 405: How long are we going to keep this around for? Let's leave some sort of warning in place to get rid of it in 3 months or whatever

ElementX/Sources/Application/AppSettings.swift
```diff
@@ -61,7 +58,6 @@ final class AppSettings {
         case developerOptionsEnabled
         
         // Doug's tweaks 🔧
-        case hideUnreadMessagesBadge
```
stefanceriu, line 64: Mistake?

ElementX/Sources/Services/Client/ClientProxyProtocol.swift
```diff
@@ -63,6 +63,13 @@ enum SessionVerificationState {
     case unverified
 }
 
+// The `Codable` conformance is just for the purpose of migration
```
stefanceriu, line 66: Is this still relevant?

ElementX/Sources/Screens/Settings/AvancedOptionsScreen/AdvancedSettingsScreenViewModel.swift
```diff
@@ -12,19 +12,89 @@ typealias AdvancedSettingsScreenViewModelType = StateStoreViewModelV2<AdvancedSe
 
 class AdvancedSettingsScreenViewModel: AdvancedSettingsScreenViewModelType, AdvancedSettingsScreenViewModelProtocol {
     private let analytics: AnalyticsService
+    private let clientProxy: ClientProxyProtocol
+    private let userIndicatorController: UserIndicatorControllerProtocol
     
-    init(advancedSettings: AdvancedSettingsProtocol, analytics: AnalyticsService) {
+    private var timelineMediaVisibilityTask: Task<Void, Never>?
+    private var hideInviteAvatarsTask: Task<Void, Never>?
```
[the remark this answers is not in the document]
Velin92: checked with design and we would prefer to not have a save button and a general blocking spinner, we would like the values to be updated ASAP in the UI.

ElementX/Sources/Services/Client/ClientProxyProtocol.swift
```diff
@@ -185,6 +196,8 @@ protocol ClientProxyProtocol: AnyObject, MediaLoaderProtocol {
     func isAliasAvailable(_ alias: String) async -> Result<Bool, ClientProxyError>
     
     @discardableResult func clearCaches() async -> Result<Void, ClientProxyError>
+    
+    func fetchMediaPreviewConfig() async -> Result<MediaPreviewConfig?, ClientProxyError>
```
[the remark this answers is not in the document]
Velin92: this is only used for the migration, can make a MediaPreviewConfigProxy if you prefer however.

ElementX/Sources/Application/AppSettings.swift
Velin92: oh there it is, will just move it back here then

stefanceriu (approved):
Nice, lgtm 👍

PR #4144, opened 2025-05-23
Various timeline code improvements

stefanceriu:
This patch introduces a new SwiftUI TimelineView layer (wrapping the representable) to hold all the timeline specific bindings that are currently duplicated throughout the screens a timeline is used. It also renames the `TimelineProvider` to `TimelineItemProvider` and adds more documentation.

### DIFF
--- a/ElementX/Sources/Other/NetworkMonitor/NetworkMonitor.swift
+++ b/ElementX/Sources/Other/NetworkMonitor/NetworkMonitor.swift
@@ -19,7 +19,7 @@ class NetworkMonitor: NetworkMonitorProtocol {
     }
     
     init() {
-        queue = DispatchQueue(label: "io.element.elementx.networkmonitor", qos: .background)
+        queue = DispatchQueue(label: "io.element.elementx.network_monitor", qos: .background)
         pathMonitor = NWPathMonitor()
         reachabilitySubject = CurrentValueSubject<NetworkMonitorReachability, Never>(.reachable)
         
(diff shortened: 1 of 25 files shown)


ElementX/Sources/Services/Timeline/TimelineController/TimelineControllerProtocol.swift
pixlwave, line 1: 👌

ElementX/Sources/Services/Timeline/TimelineProxyProtocol.swift
```diff
@@ -37,9 +37,11 @@ enum TimelineProxyError: Error {
     case failedPaginatingEndReached
 }
 
-// sourcery: AutoMockable
+/// ElementX proxies generally wrap the counterpart RustSDK objects while providing platform specific
+/// interfaces. In this case it composes methods for interacting a room's timeline with the
+/// ``TimelineItemProviderProtocol`` which offers a reactive interface to timeline changes.
```
pixlwave, line 42:
```suggestion
/// Element X proxies generally wrap the counterpart RustSDK objects while providing platform specific
/// interfaces. In this case it composes methods for interacting with a room's timeline to be used alongside
/// the ``TimelineItemProviderProtocol`` which offers a reactive interface to timeline changes.
```

pixlwave (approved):
Very nice refactor, thank you for the tidy-up and for adding the docs!

ElementX/Sources/Other/NetworkMonitor/NetworkMonitor.swift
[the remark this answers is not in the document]
stefanceriu: At least they're consistent now. We can rename all of them later.

PR #4147, opened 2025-05-26
Add a developer option that hides notification alerts when a sound wouldn't be played.

pixlwave:
Just trying a thing using the developer options…

Enabling this makes `isNoisy` behave like iMessage's Hide Alerts option, where the badge count is increased, but the notification alert isn't delivered.


Velin92 (approved):
LGTM

PR #4155, opened 2025-05-28
Run the 'Prevent blocked' check whenever a PR branch is updated

robintown:
Because we're now [requiring](https://github.com/element-hq/it-internal/issues/1896) the 'Prevent blocked' check to pass before merging a PR, GitHub Actions apparently expects it to be associated with the latest Git ref of the PR's branch whenever the branch is updated. Therefore we need to re-run the workflow on the 'synchronize' event.

### DIFF
--- a/.github/workflows/blocked.yml
+++ b/.github/workflows/blocked.yml
@@ -1,7 +1,7 @@
 name: Prevent blocked
 on:
   pull_request_target:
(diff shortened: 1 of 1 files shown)


pixlwave (approved):
Thanks Robin! Not sure why CI is failing but doesn't exactly matter in this instance.

PR #4157, opened 2025-05-29
Setup threaded timeline actions

stefanceriu:
These patches setup action forwarding for threaded timelines and make various changes to the room flow coordinator state machine to support them:
* extract the state machine state setup in an extension to trim down the size of the room flow coordinator
* remove simple transition handling as the state machine already enforces correct transitions
* re-arrange and better separate transitions based on their area


ElementX/Sources/FlowCoordinators/RoomFlowCoordinatorStateMachine.swift
```diff
@@ -0,0 +1,377 @@
...
+            case (.thread, .presentPollForm):
+                return .pollForm(previousState: fromState)
+                
+            case (.thread, .presentResolveSendFailure):
+                return .resolveSendFailure(previousState: fromState)
+                
+            // Room + Thread
+                
+            case (.mediaUploadPicker(_, let previousState), .dismissMediaUploadPicker):
+                return previousState
```
pixlwave, line 248: I'm mixed about whether this nicely sorted solution is better than grouping by the "new state" and keep the room/thread/dismiss in 1 block like this:

```
case (.room, …)
    return …
case (.thread, …)
    return …
case (…, .dismiss)
    return previousState
```

stefanceriu: Let's see how this plays out, the thought of reordering these again makes me think of very bad things 😅

PR #4163, opened 2025-06-02
Add a placeholder association for localhost in developer mode.

pixlwave:
Just a nice quality-of-life tweak so we don't need to check the [docs](https://developer.apple.com/documentation/BundleResources/Entitlements/com.apple.developer.associated-domains) for the correct syntax for alternate modes.

Note: I didn't try this with localhost specifically (it just seemed the most obvious placeholder).


stefanceriu (approved):
👌

PR #4167, opened 2025-06-03
Automatically open a PR to bump the calver

pixlwave:

- Updates our tools to use Swift 6.1 (because Regex literals needed Swift 6 apparently).
    - Makes the GenerateSDKMocks an Async command as some sendability warnings made a refactor necessary.
- Tidy up the tools package layout with some subdirectories 📁
- Adds a new tool `swift run tools bump-calendar-version`
    - This tool assumes a build will be release in 6 days time (i.e. Tues RC -> Mon App Store) and will bump the year/month if needed, resetting the patch back to `0` when that happens.
    - It doesn't touch the patch as that is bumped by the release script.
- Runs this tool on CI every Tuesday, creating a PR if there are changes.
    - If this doesn't work we can change to the 1st of the month, but we can ignore the PRs so it seems like it should be fine.

### DIFF
--- a/.github/workflows/automatic-calendar-version.yml
+++ b/.github/workflows/automatic-calendar-version.yml
@@ -1,0 +1,42 @@
+name: Automatic Calendar Version 
+on:
+  schedule:
+    # At 03:00 UTC every Tuesday in preparation for an RC.
+    # The tool assumes the release is published in 6-days (the following Monday).
+    # Note: Most of these runs will be no-op until the release month changes.
+    - cron: '0 3 * * 2'
+  workflow_dispatch:
+
+# Bumps the year and month, resetting the patch.
+# Patch bumps are handled by the release script.
+jobs:
+  automatic-calendar-version:
+    runs-on: macos-15
+    # Skip in forks
+    if: github.repository == 'element-hq/element-x-ios'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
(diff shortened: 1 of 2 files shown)


Tools/Sources/Commands/BumpCalendarVersion.swift
```diff
@@ -0,0 +1,54 @@
...
+        }
+        
+        try updatedYAMLString.write(to: yamlURL, atomically: true, encoding: .utf8)
+    }
+    
+    /// Returns the new version string if a change is necessary.
+    ///
+    /// **Note:** This tool does *not* handle patch bumps, those are done automatically in the release script.
+    private func generateNewVersion(from currentVersion: String) throws -> String? {
+        let releaseDate = Date.now.addingTimeInterval(6 * 24 * 60 * 60) // Always assume we're building the RC.
```
Velin92, line 41: my only concern is that this might be a problem in the case we are releasing a hotfix version, and we are at the end of the month (considering that approval might take 1 day at most), maybe an option that disables this might come in handy, however not sure how we would use that from XcodeCloud, I guess we are able to provide arguments to an Xcode Cloud run right?

pixlwave: I don't intend for us to ever run this on Xcode Cloud. I purely made it to have those GitHub PRs available for us to manually merge when we're ready.

PR #4168, opened 2025-06-03
Threaded timeline composer

stefanceriu:
This PR copies over parts of the RoomScreen to setup a composer for threaded timelines. It also sets up sending within threads on the TimelineProxy level.

Known problems:
* drafts cannot be handled in threads so they're disabled for now
* we do not have enough information to render replies to other threaded items. That will be handled separately
* similarly `replyParams` have not been setup for sending media
* `replyParams` aren't even available for locations and polls
* other things I'm probably missing now. All will be hanled in due time 😎

### DIFF
--- a/ElementX.xcodeproj/project.pbxproj
+++ b/ElementX.xcodeproj/project.pbxproj
@@ -610,7 +610,6 @@
 		7691233E3572A9173FD96CB3 /* SecureBackupKeyBackupScreenViewModelTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2E88534A39781D76487D59DF /* SecureBackupKeyBackupScreenViewModelTests.swift */; };
 		76C874243A8C440D6CF7B344 /* ProcessInfo.swift in Sources */ = {isa = PBXBuildFile; fileRef = 077B01C13BBA2996272C5FB5 /* ProcessInfo.swift */; };
 		7708976CEE6AFB5CFAEFBA68 /* PillTextAttachment.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9CF1EE0AA78470C674554262 /* PillTextAttachment.swift */; };
-		7756C4E90CABE6F14F7920A0 /* BugReportUITests.swift in Sources */ = {isa = PBXBuildFile; fileRef = C6FEA87EA3752203065ECE27 /* BugReportUITests.swift */; };
 		77574A519A4E484880053EAD /* IdentityConfirmationScreenModels.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7FDF541AE914059942B575B4 /* IdentityConfirmationScreenModels.swift */; };
 		77693820498ABF3508814D49 /* AppLockServiceTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD97F9661ABF08CE002054A2 /* AppLockServiceTests.swift */; };
 		77920AFA8091AC6B9F190C90 /* Signposter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 752A0EB49BF5BCEA37EDF7A3 /* Signposter.swift */; };
@@ -825,6 +824,7 @@
 		9E838A62918E47BC72D6640D /* UserIndicatorPresenter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6AB54B4F94686CCF0289B72F /* UserIndicatorPresenter.swift */; };
 		9EBDC79CAC9B63A0D626E333 /* LegalInformationScreenCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5EB2CAA266B921D128C35710 /* LegalInformationScreenCoordinator.swift */; };
 		9EE71509E6E7519A2B2388B3 /* KnockRequestsListScreenModels.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8BD9C9A31D9AB3B6D8128E69 /* KnockRequestsListScreenModels.swift */; };
+		9EF72884B74BC39B01640098 /* BugReportTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 27DF257F5D968E5DD719583C /* BugReportTests.swift */; };
 		9F11B9F347F9E2D236799FB3 /* ElementCallServiceConstants.swift in Sources */ = {isa = PBXBuildFile; fileRef = 406C90AF8C3E98DF5D4E5430 /* ElementCallServiceConstants.swift */; };
 		9F11E743EA01482E78A438B0 /* GlobalSearchScreenCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 22DB19219E6CC4D002E15D48 /* GlobalSearchScreenCell.swift */; };
 		9F8BEA86540D8980BDD7C176 /* AuthenticationStartScreenModels.swift in Sources */ = {isa = PBXBuildFile; fileRef = FE3AB4D399D39FBD78119011 /* AuthenticationStartScreenModels.swift */; };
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
pixlwave, line 1: While we're here, could we maybe call `roomContext` `context` to match all the other screens (and the thread/pinned screens specifically where only the timeline context is prefixed.

ElementX/Sources/Screens/ThreadTimelineScreen/View/ThreadTimelineScreen.swift
```diff
@@ -9,15 +9,95 @@ import Compound
 import SwiftUI
 
 struct ThreadTimelineScreen: View {
-    @ObservedObject var context: ThreadTimelineScreenViewModel.Context
-    @ObservedObject var timelineContext: TimelineViewModel.Context
+    @ObservedObject private var context: ThreadTimelineScreenViewModelType.Context
+    @ObservedObject private var timelineContext: TimelineViewModelType.Context
+    @ObservedObject private var composerToolbarContext: ComposerToolbarViewModelType.Context
+    @State private var dragOver = false
```
pixlwave: Oh nice to see this also 👌

ElementX/Sources/Screens/ThreadTimelineScreen/View/ThreadTimelineScreen.swift
```diff
@@ -9,15 +9,95 @@ import Compound
...
+            }
+            .onDrop(of: ["public.item", "public.file-url"], isTargeted: $dragOver) { providers -> Bool in
+                guard let provider = providers.first,
+                      provider.isSupportedForPasteOrDrop else {
+                    return false
+                }
+                
+                timelineContext.send(viewAction: .handlePasteOrDrop(provider: provider))
+                return true
+            }
```
pixlwave: Maybe we should extract this into a shared `.timelineFileDrop(context: TimelineViewModel.Context)` modifier?

ElementX/Sources/Screens/ThreadTimelineScreen/View/ThreadTimelineScreen.swift
```diff
@@ -9,15 +9,95 @@ import Compound
...
+                        // Intentionally using system primary colour to get white/black.
+                        .shadow(color: .primary.opacity(0.33), radius: 2.0)
+                }
+                .padding()
+        }
+        .opacity(isAtBottomAndLive ? 0.0 : 1.0)
+        .accessibilityHidden(isAtBottomAndLive)
+        .animation(.elementDefault, value: isAtBottomAndLive)
+        .accessibilityIdentifier(A11yIdentifiers.roomScreen.scrollToBottom)
+    }
```
pixlwave: Should we extract this into a `TimelineScrollToBottomButton` view?

ElementX/Sources/Screens/ThreadTimelineScreen/View/ThreadTimelineScreen.swift
```diff
@@ -9,15 +9,95 @@ import Compound
...
+    @ViewBuilder
+    private var composer: some View {
+        if context.viewState.canSendMessage {
+            composerToolbar
+        } else {
+            Text(L10n.screenRoomTimelineNoPermissionToPost)
+                .font(.compound.bodyLG)
+                .foregroundStyle(.compound.textDisabled)
+                .multilineTextAlignment(.center)
+                .padding(.vertical, 10) // Matches the MessageComposerStyleModifier
```
pixlwave: And this into a `ComposerDisabledView`?

ElementX/Sources/Screens/ThreadTimelineScreen/View/ThreadTimelineScreen.swift
```diff
@@ -9,15 +9,95 @@ import Compound
...
+            .safeAreaInset(edge: .bottom, spacing: 0) {
+                composer
+                    .padding(.bottom, composerToolbarContext.composerFormattingEnabled ? 8 : 12)
+                    .background {
+                        if composerToolbarContext.composerFormattingEnabled {
+                            RoundedRectangle(cornerRadius: 20)
+                                .stroke(Color.compound.borderInteractiveSecondary, lineWidth: 0.5)
+                                .ignoresSafeArea()
+                        }
+                    }
```
pixlwave: It feels like this should either be in the `ComposerToolbar` or should be a `.composerToolbarFormattingStyle(_ isEnabled: Bool)` modifier.

ElementX/Sources/Services/Timeline/TimelineProxy.swift
```diff
@@ -389,8 +389,15 @@ final class TimelineProxy: TimelineProxyProtocol {
         return .success(())
     }
     
+    /// Send a message within a room. If `inReplyToEventID` is specified then it will be sent as a reply
+    /// to that particular message. If the `threadRootEventID` is also specified then it will be sent
+    /// as a reply to the given `inReplyToEventID` within the thread rooted in `threadRootEventID`
+    ///
+    /// Internally `enforceThread` is set to true whenever `threadRootEventID` is specified
+    /// and `replyWithinThread` when `inReplyToEventID` is.
```
pixlwave, line 397: I still don't understand these 2 parameters. Why couldn't this logic be moved to Rust so that users of the SDK just get the right behaviour for free? I don't see what other behaviour someone would want.

ElementX/Sources/Screens/ThreadTimelineScreen/ThreadTimelineScreenCoordinator.swift
```diff
@@ -96,20 +120,36 @@ final class ThreadTimelineScreenCoordinator: CoordinatorProtocol {
...
     
     func stop() {
         viewModel.stop()
     }
         
     func toPresentable() -> AnyView {
-        AnyView(ThreadTimelineScreen(context: viewModel.context, timelineContext: timelineViewModel.context))
+        let composerToolbar = ComposerToolbar(context: composerViewModel.context,
+                                              wysiwygViewModel: wysiwygViewModel,
+                                              keyCommands: composerViewModel.keyCommands)
```
pixlwave: Could we not just put both of these into the `ComposerViewModel`'s view state and then we wouldn't need to store the `wysiwygViewModel` in this VM (nor the `RoomScreen`'s).

pixlwave (approved):
Woooo! 🧵✍️

ElementX/Sources/Screens/ThreadTimelineScreen/View/ThreadTimelineScreen.swift
stefanceriu: I just moved it to the recently introduced `TimelineView` ¯\_(ツ)_/¯

ElementX/Sources/Screens/ThreadTimelineScreen/View/ThreadTimelineScreen.swift
stefanceriu: Good point, moved them to the ComposerToolbar

ElementX/Sources/Services/Timeline/TimelineProxy.swift
stefanceriu: I'll have a proper look at all this does when I introduce threaded sending for locations and polls.

ElementX/Sources/Screens/ThreadTimelineScreen/ThreadTimelineScreenCoordinator.swift
stefanceriu: Great idea, thank you!

PR #4169, opened 2025-06-03
Tombstoned and upgraded rooms implementation

### LINKED ISSUE #4129: Badly mangled timeline animations for the "This is the beginning of..." text in new rooms
### Steps to reproduce

1. Create a room
2. Send a message
3. Observe a horrible mangled animation for the "This is the beginning of..." text at the top of the room
4. Go to jail, do not pass go, do not collect $200.

https://github.com/user-attachments/assets/6694cf6f-5d2b-48fc-9c0f-2bf4b10475d3

#### What did you expect?

No horribly broken animations

### Your phone model

iPhone 12 Pro Max

### Operating system version

iOS18.5

### Application version

921

### Homeserver

_No response_

### Will you send logs?

No

### LINKED ISSUE #4166: [Task] Support for upgraded rooms

Velin92:
fixes #4166 
fixes #4129

https://github.com/user-attachments/assets/7ca8c0e2-4f00-4255-a7e8-1994030ec2d1


ElementX/Sources/Services/Room/RoomProxyProtocol.swift
```diff
@@ -76,6 +76,8 @@ protocol JoinedRoomProxyProtocol: RoomProxyProtocol {
     
     var timeline: TimelineProxyProtocol { get }
     
+    var predecessorRoom: PredecessorRoom? { get }
```
stefanceriu, line 79: Let's expose the successor here too for completeness.

ElementX/Sources/Other/SwiftUI/Views/Highlight.swift
```diff
@@ -0,0 +1,35 @@
...
+// Copyright 2025 New Vector Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Compound
+import SwiftUI
+
+struct Highlight: ViewModifier {
```
stefanceriu: `Highlight` sounds a tad too generic to me. How about we call it `HorizontalHighlightGradient` or something along those lines?

ElementX/Sources/Other/SwiftUI/Views/PlaceholderAvatarImage.swift
```diff
@@ -12,6 +12,7 @@ struct PlaceholderAvatarImage: View {
     @Environment(\.redactionReasons) private var redactionReasons
 
     private let textForImage: String
+    /// When `nil` the tombstoned colors will be used
```
stefanceriu: This comment is a bit unexpected, there's no tombstone concept in this context. I would just remove all 3  comments and just let it be how this component works, especially as they're plain old compound colors.

ElementX/Sources/Other/SwiftUI/Views/PlaceholderAvatarImage.swift
```diff
@@ -12,6 +12,7 @@ struct PlaceholderAvatarImage: View {
...
     private let contentID: String?
```
stefanceriu: More importantly though, why doesn't a tombstoned room have a contentID, what's missing?

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
```diff
@@ -197,6 +197,8 @@ struct HomeScreenRoom: Identifiable, Equatable {
         
     let canonicalAlias: String?
     
+    let isTombstoned: Bool
```
stefanceriu, line 200: Let's move this next to the other `isFoo` fields.

ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomCell.swift
```diff
@@ -135,7 +135,11 @@ struct HomeScreenRoomCell: View {
     
     @ViewBuilder
     private var lastMessage: some View {
-        if let lastMessage = room.lastMessage {
+        // If the room is tombstoned, show a specific message, regardless of any last message.
+        if room.isTombstoned {
+            Text(L10n.screenRoomlistTombstonedRoomDescription)
+                .lastMessageFormatting()
+        } else if let lastMessage = room.lastMessage {
```
stefanceriu: Can we handle this on the model level?

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -326,6 +330,7 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
     }
     
     private func handleRoomInfoUpdate(_ roomInfo: RoomInfoProxy) async {
+        state.hasSuccessor = roomInfo.successor != nil
```
stefanceriu, line 333: Leave a new line here please

ElementX/Sources/Screens/RoomScreen/View/RoomScreenFooterView.swift
stefanceriu, line 1: Noice! 👌

ElementX/Sources/Screens/Timeline/TimelineModels.swift
```diff
@@ -77,6 +78,8 @@ enum TimelineViewAction {
     
     case hasScrolled(direction: ScrollDirection)
     case setOpenURLAction(OpenURLAction)
+    
+    case seePredecessorTapped
```
stefanceriu: Let's use `display` like all the other action, perhaps call it `displayPredecessorRoom`.

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
```diff
@@ -211,6 +212,9 @@ class TimelineViewModel: TimelineViewModelType, TimelineViewModelProtocol {
             actionsSubject.send(.hasScrolled(direction: direction))
         case .setOpenURLAction(let action):
             state.openURL = action
+        case .seePredecessorTapped:
+            guard let predecessorID = roomProxy.predecessorRoom?.roomId else { return }
```
stefanceriu: I would fatal error this, no reason it should ever reach this point without a predecessorRoom

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/TimelineStartRoomTimelineView.swift
```diff
@@ -5,20 +5,50 @@
...
+                    .font(.compound.bodySM)
+                    .foregroundColor(.compound.textSecondary)
+                    .padding(.bottom, 24)
+                    .padding(.top, context?.viewState.hasPredecessor == true ? 0 : 24)
+                    .frame(maxWidth: .infinity)
+            }
+        }
+    }
+    
+    private var upgrade: some View {
```
stefanceriu: Maybe call it an `upgradeDialogue` so it matches the other one (tombstonedDialogue)?

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/TimelineStartRoomTimelineView.swift
```diff
@@ -5,20 +5,50 @@
...
     var body: some View {
-        Text(title)
-            .font(.compound.bodySM)
-            .foregroundColor(.compound.textSecondary)
-            .padding(.vertical, 24)
+        VStack(spacing: 14) {
+            if context?.viewState.hasPredecessor == true {
+                upgrade
+            }
+            if context?.viewState.isDirectOneToOneRoom != true {
```
stefanceriu, line 21: Why this new condition? Would've been great if this was a different commit explaining the reason.

ElementX/Sources/Services/Timeline/TimelineItems/Items/Virtual/TimelineStartRoomTimelineItem.swift
```diff
@@ -8,6 +8,6 @@
 import Foundation
 
 struct TimelineStartRoomTimelineItem: DecorationTimelineItemProtocol, Equatable {
-    let id: TimelineItemIdentifier = .virtual(uniqueID: .init(UUID().uuidString))
+    let id: TimelineItemIdentifier = .virtual(uniqueID: .init("TimelineStart"))
```
stefanceriu: Same here, why this change, what's the reason for it. Different commit with a message would've been nice, especially for future us.

stefanceriu (changes requested):
Nice, well done! 👏

ElementX/Sources/Other/SwiftUI/Views/PlaceholderAvatarImage.swift
pixlwave, line 1: Errr, hold on. Those fallback colours are already in use somewhere iirc (possibly room creation). Otherwise `contentID` wouldn't be optional. It's also something I've wanted to make non-optional and make the caller do an

```
if let contentID = whatever.id {
    AvatarImage…
} else {
    FallbackAvatarImage
}
```

I don't like how we use the placeholder image for these cases where it isn't a placeholder.

ElementX/Sources/Other/SwiftUI/Views/Highlight.swift
```diff
@@ -0,0 +1,35 @@
...
+    let borderColor: Color
+    let primaryColor: Color
+    let secondaryColor: Color
+    
+    func body(content: Content) -> some View {
+        ZStack(alignment: .top) {
+            VStack(spacing: 0) {
+                borderColor
+                    .frame(height: 1)
+                LinearGradient(colors: [primaryColor, secondaryColor],
```
pixlwave: Can't we just do this and remove the secondary colour entirely?

```suggestion
                LinearGradient(colors: [primaryColor, .clear],
```

ElementX/Sources/Other/SwiftUI/Views/PlaceholderAvatarImage.swift
Velin92: it has a contentID, but since we want the avatar to be always a static grey background avatar with a "!" we don't need the content ID (since is used to computed one of the colors we used to determine the background and the text color)

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/TimelineStartRoomTimelineView.swift
Velin92: Yeah I'll leave a comment, but the reaosn is that we don't add the timeline start item for DMs, now instead since also DMs might be tombstoned, we want to add it in the case we have a tombstoned DM, we just need to make sure the "This is the beginning of X" text doesn't appear for that case.

ElementX/Sources/Services/Timeline/TimelineItems/Items/Virtual/TimelineStartRoomTimelineItem.swift
Velin92: The reason is this: https://github.com/element-hq/element-x-ios/pull/4130
Actually most of our virtual items already use a static string for their id, so this was more of the exception that di not align with the others.

ElementX/Sources/Other/SwiftUI/Views/Highlight.swift
Velin92: no because this might depend on the context... for example in the case we are replacing the composer, we want the view to have colored background so that timeline items don't get seen behind the replaced part.

pixlwave: Riiiight 👍

ElementX/Sources/Other/SwiftUI/Views/PlaceholderAvatarImage.swift
pixlwave: TL;DR of what I just said in a DM:

`OverridableAvatarImage` used the `.white` and `.compound.iconPrimary` colours that were removed when used to pick avatar before the room was created IIRC. Checking now this doesn't happen any more so I guess we changed that behaviour.

Based on that, it would be nicer to make `contentID` non-optional and instead create a dedicated `TombstonePlaceholderAvatarImage` that gets used in the `case: .tombstoned`

ElementX/Sources/Other/SwiftUI/Views/PlaceholderAvatarImage.swift
Velin92: reverted, i made a dedicated image for the tombstone rooms

ElementX/Sources/Other/SwiftUI/Views/PlaceholderAvatarImage.swift
Velin92: reverted, i made a dedicated image for the tombstone rooms

PR #4175, opened 2025-06-06
Add a RoomScreenHook.

pixlwave:
Allows updating the initial `RoomScreenViewState` before the screen is displayed.

Also adds `AppHooks.configure(with:)` to inform hooks about the UserSession if needed.

### DIFF
--- a/ElementX/Sources/AppHooks/AppHooks.swift
+++ b/ElementX/Sources/AppHooks/AppHooks.swift
@@ -9,6 +9,10 @@ import Foundation
 
 class AppHooks: AppHooksProtocol {
     #if IS_MAIN_APP
+    func configure(with userSession: UserSessionProtocol?) async {
+        await roomScreenHook.configure(with: userSession)
+    }
+    
     private(set) var appSettingsHook: AppSettingsHookProtocol = DefaultAppSettingsHook()
     func registerAppSettingsHook(_ hook: AppSettingsHookProtocol) {
         appSettingsHook = hook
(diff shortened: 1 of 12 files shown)


ElementX/Sources/AppHooks/AppHooks.swift
```diff
@@ -38,6 +47,10 @@ class AppHooks: AppHooksProtocol {
 
 protocol AppHooksProtocol {
     func setUp()
+    
+    #if IS_MAIN_APP
+    func configure(with userSession: UserSessionProtocol?) async
```
stefanceriu, line 52: Why is it that `setUp` has a default protocol implementation while this one doesn't? 🤔

ElementX/Sources/AppHooks/Hooks/RoomScreenHook.swift
```diff
@@ -0,0 +1,18 @@
...
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Foundation
+
+protocol RoomScreenHookProtocol {
+    func configure(with userSession: UserSessionProtocol?) async
+    func update(_ viewState: RoomScreenViewState) -> RoomScreenViewState
```
stefanceriu, line 12: Was the `inout` parameter a no go?

pixlwave: It worked fine, but didn't seem necessary once I started caching the value at launch (removing the need for an async call) and could set an initial state immediately in the init.

ElementX/Sources/AppHooks/AppHooks.swift
pixlwave: Ah, this isn't actually necessary, removed in #4220.

PR #4176, opened 2025-06-06
Fix subsequent media upload dialogue presentations 

stefanceriu:
Have the media upload preview state transition to .room or .thread after being dismissed instead of back to the media upload picker.


pixlwave (approved):
LGTM!

PR #4177, opened 2025-06-09
Send .`fullyRead` marker when leaving the room

### LINKED ISSUE #4173: Own read marker (new) does not get updated when reading the room #2281
### Steps to reproduce

1. Open a room with unread messages
2. Observe the "New" marker being at the correct place
3. Scroll and read through the room as long as you like
4. Close the room and open it again
5. Observe the "New" marker being at the same place as before instead of vanishing/updating to the latest message
6. (Bonus: Open the room with Element-Android or Element-Web and notice the "New" marker being updated in Element-X)

#### What did you expect?
"New" marker to be updated when opening the room

#### What happened instead?
Nothing. The "New" marker stays at the same place unless one opens the room with another client. This makes it very hard to keep up with a semi high-traffic room using Element-X

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

Nightlly 25.06.1(935)

### Homeserver

_No response_

### Will you send logs?

No

Velin92:
fixes #4173


stefanceriu:
I honestly believe this is the wrong solution, we shouldn't rely on exiting a room for it to be marked properly. What if they background the app and it just dies in the background, your state will still be wrong. So sure, this works sometimes but not enough. 

Do you know why reaching the bottom of the timeline doesn't do the right thing?

Am I right in believing that the unread marker should behave similarly to the read receipts, marking the top of the currently read content until the bottom is reached and then become unset? We don't have an api for that, do we?

PR #4182, opened 2025-06-10
Various thread tweaks

stefanceriu:
* bump the sdk and fix the various breaking changes
* configured timelines based on the threads feature flag
* update the thread summary design
* start showing replies to threaded items within threaded timelines
* add a "reply in thread" timeline item action to start or continue a thread


ElementX/Sources/Screens/Timeline/View/Style/TimelineItemBubbledStylerView.swift
```diff
@@ -212,10 +213,6 @@ struct TimelineItemBubbledStylerView<Content: View>: View {
         }
     }
     
-    private var shouldShowReplyDetails: Bool {
-        !timelineItem.properties.isThreaded || (timelineItem.properties.isThreaded && !context.viewState.timelineKind.isThread)
-    }
-
```
Velin92, line 218: why has this been removed? since the threads might not be enabled, wouldn't we want to still use this?

stefanceriu: I introduced this on an earlier version of threaded timelines to hide the reply bubbles, as all in-thread messages would have one with the old implementation. Now the SDK correctly handles threads and only populates replies within threaded timelines only for items that actually are replies to other items in that thread. The old behavior is left unchanged and it based on how the sdk timeline is configured on creation.

Velin92 (approved):
LGTM

PR #4183, opened 2025-06-10
Support for sending media in threads

stefanceriu:
This patch add the `threadRootEventID` as a parameter to media sending functions and forwards the timeline thread root between various application states to the place where its needed.
The 2 missing items are Locations and Polls which currently don't have any support on the Rust side for being sent within threads. That will come in a separate PR.


ElementX/Sources/Screens/RoomScreen/ComposerToolbar/ComposerToolbarModels.swift
```diff
@@ -80,6 +80,8 @@ struct ComposerToolbarViewState: BindableState {
     
     var keyCommands: [WysiwygKeyCommand] = []
     
+    var isWithinThread: Bool
```
Velin92: I would simply rename this to `isInThread`

stefanceriu: Fair 👍

PR #4191, opened 2025-06-12
Fixes #4180 - Prevent room header autolayout crashes on iOS 26.

### LINKED ISSUE #4180: App Crashes On iOS26
### Steps to reproduce

If you try to Open an Chat The app Just Crashes

#### What did you expect?
No Crash
#### What happened instead?
An Crash

### Your phone model

16 Pro Max

### Operating system version

iOS 26

### Application version

25.06.2 (172)

### Homeserver

Synapse

### Will you send logs?

Yes

stefanceriu:


ElementX/Sources/Other/SwiftUI/Views/RoomHeaderView.swift
```diff
@@ -17,6 +17,17 @@ struct RoomHeaderView: View {
     let mediaProvider: MediaProviderProtocol?
     
     var body: some View {
+        if #available(iOS 26, *) {
+            // https://github.com/element-hq/element-x-ios/issues/4180
+            // Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'NSLayoutConstraint constant is not finite!
+            content
+        } else {
+            content
```
Velin92, line 25: add the comment back at the top here:         `// Take up as much space as possible, with a leading alignment for use in the principal toolbar position`

Velin92 (approved):

ElementX/Sources/Other/SwiftUI/Views/RoomHeaderView.swift
stefanceriu: Oupsie, thanks!

PR #4194, opened 2025-06-12
Add state to the accessibility label of RTE formatting buttons

### LINKED ISSUE #4193: [Task] Localize the format button state in the rich text editor

Velin92:
fixes #4193

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/ComposerToolbar/ComposerToolbarModels.swift
+++ b/ElementX/Sources/Screens/RoomScreen/ComposerToolbar/ComposerToolbarModels.swift
@@ -239,31 +239,46 @@ extension FormatItem {
     }
 
     var accessibilityLabel: String {
-        switch type {
+        let localizedAction = switch type {
         case .bold:
-            return L10n.richTextEditorFormatBold
+            L10n.richTextEditorFormatBold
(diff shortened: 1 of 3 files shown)


ElementX/Sources/Screens/RoomScreen/ComposerToolbar/ComposerToolbarModels.swift
```diff
@@ -239,31 +239,46 @@ extension FormatItem {
...
+extension ActionState {
+    /// Returns a localized string that describes the action state.
+    var localizedDescription: String {
+        switch self {
+        case .disabled:
+            L10n.richTextEditorFormatStateDisabled
+        case .enabled:
+            L10n.richTextEditorFormatStateOff
+        case .reversed:
+            L10n.richTextEditorFormatStateOn
```
bmarty, line 281: @Velin92 why not declaring that those ui elements act like toggleButton https://developer.apple.com/documentation/uikit/uiaccessibilitytraits/togglebutton ?

There is also `notenabled` https://developer.apple.com/documentation/uikit/uiaccessibilitytraits/notenabled that could help for the `.disabled` case.

(This is what I have done on the equivalent Android API)

PR #4195, opened 2025-06-12
Make BugReportService `reportUrl` optional

NicolasBuquet:
The `BugReportService` can optionally return a `reportUrl` value. The Matrix Specification specifies it as optional.

In current ElementX code, the `reportUrl` is considered mandatory.

This PR allows `reportUrl` to be optional.

### DIFF
--- a/ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
@@ -16,7 +16,7 @@ protocol AuthenticationFlowCoordinatorDelegate: AnyObject {
 
 class AuthenticationFlowCoordinator: FlowCoordinatorProtocol {
     private let authenticationService: AuthenticationServiceProtocol
-    private let bugReportService: BugReportServiceProtocol
+    private let bugReportService: BugReportServiceProtocol?
     private let navigationRootCoordinator: NavigationRootCoordinator
     private let navigationStackCoordinator: NavigationStackCoordinator
     private let appMediator: AppMediatorProtocol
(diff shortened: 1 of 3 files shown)


stefanceriu:
The BugReport service already takes an optional `baseURL` and correctly reports `isEnabled` based on it. What's the benefit of making the whole thing optional and duplicating that behavior?

Velin92 (approved):

NicolasBuquet:
> The BugReport service already takes an optional `baseURL` and correctly reports `isEnabled` based on it. What's the benefit of making the whole thing optional and duplicating that behavior?

Sorry, it's not necessary in fact to have `BugReportService` optional in this PR.

On Tchap, we had to make it optional because : 
> // Tchap: set `bugReportService` optional because it can be nil before user is logged.

> // Tchap: verify `bugReportService` is instantiated. It can be nil in Tchap on Authentication screen while user is not logged.

The BugReportService baseUrl is dependant on the server on which the user is logged.

So I had to make it optional, and instantiate it in `AppCoordinator::userSession::didSet`.

I disabled its instantiation in `AppCoordinator::setupServiceLocator` and create an `AppCoordinator::setupServiceLocatorBugReport` method.

I had to do this because `baseUrl` is a `private let` property.

stefanceriu:
Asked for an extra review from @pixlwave as it ties in to things he's working on.

NicolasBuquet:
@stefanceriu 

I removed the modsifications about `BugReportService` optional that was not necessary to handle optional `reportUrl`. (done -> https://github.com/element-hq/element-x-ios/pull/4210)

I will open another PR about optional `BugReportService`.

pixlwave (approved):
Thanks @NicolasBuquet, looks reasonable to me. I'm going to merge this into a branch to make a small refactor before it lands on `develop`.

Note: For anyone looking for related docs, rageshakes aren't part of the spec, but the rageshake server contains the API docs directly: https://github.com/matrix-org/rageshake/blob/main/docs/api.md

PR #4210, opened 2025-06-17
Make BugReportService optional

NicolasBuquet:
Make `BugReportService` optional (because it can be undefined before user is logged in).

On Tchap, the HomeServer is known only after user is logged in (because the HomeServer is dependant on the department of the user).

The `Report a problem` button on the Start Page can't report a bug if the `BugReportService` is not defined.

Making BugReportService optional (and then nil) on the Start Page make the `Report a problem` button automatically disappear. 👍 

So, it better to:
- make `BugReportService` optional
- not instantiate it with other servioces in `setupServiceLocator`
- instantiate it when the user is logged (in `userSession::didSet`)
- check if `BugReportService` is instantiated at the beginning of `startBugReportFlow`

### DIFF
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -34,6 +34,8 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
                 configureElementCallService()
                 configureNotificationManager()
                 observeUserSessionChanges()
+                // Set up Bug Report service after user is logged because baseUrl can be dependant on user's HomeServer that is known only after logged in.
+                setupServiceLocatorBugReport(appSettings: appSettings)
                 startSync()
                 performSettingsToAccountDataMigration(userSession: userSession)
                 Task { await appHooks.configure(with: userSession) }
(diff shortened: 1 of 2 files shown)


pixlwave:
Hi @NicolasBuquet, thanks for this PR. Unfortunately we can't accept it as Element X is different to Tchap and allows the user to send rageshakes before logging in (so that we can gather logs from people who are having difficulties during the authentication flow). We [refactored](https://github.com/element-hq/element-x-ios/pull/3961/commits/6bfa8671cc2a6acc2a6590a1707d9fe2385f679c) the service already so that we could set a `nil` `baseURL` and use the `isEnabled` property in order to not make it optional.

I would suggest adding a `BugReportServiceProtocol.updateBaseURL(_:)` method and using that instead for your use case. If you create your initial service with a `nil` URL this will hide the flows until you call your `updateBaseURL` method at which point the service will then be enabled and the flows will be shown.

pixlwave:
> I would suggest adding a BugReportServiceProtocol.updateBaseURL(_:) method and using that instead for your use case.

@NicolasBuquet When I wrote this I had already thought about it as we had been planning to do it for ourselves but the idea ended up being canned. Well turns out we actually decided to do it in the end! So if you take a look at #4267 you should hopefully see everything you need to support runtime rageshake URL changes.

NicolasBuquet:
@pixlwave Thank you Doug. I did it on my side in the meantime 😆 
I did have a look at your version that is a bit more than just update the BugReportService `baseURL`.
I will surely come back to your update in the next rebases! 👍 
Thank you for the update.

PR #4215, opened 2025-06-17
Improved reactions a11y

### LINKED ISSUE #4214: [Task] Improve reactions a11y

Velin92:
fixes #4214

### DIFF
--- a/ElementX/Sources/Screens/EmojiPickerScreen/View/EmojiPickerScreen.swift
+++ b/ElementX/Sources/Screens/EmojiPickerScreen/View/EmojiPickerScreen.swift
@@ -36,6 +36,7 @@ struct EmojiPickerScreen: View {
                                         .background(Circle()
                                             .foregroundColor(emojiBackgroundColor(for: emoji.value)))
                                 }
+                                .accessibilityLabel(accessibilityLabel(for: emoji.value))
                             }
                         } header: {
                             EmojiPickerScreenHeaderView(title: category.name)
@@ -59,6 +60,14 @@ struct EmojiPickerScreen: View {
         .presentationDragIndicator(isSearching ? .hidden : .visible)
         .onChange(of: searchString) {
             context.send(viewAction: .search(searchString: searchString))
+        }
+    }
+    
+    private func accessibilityLabel(for emoji: String) -> String {
+        if selectedEmojis.contains(emoji) {
+            return L10n.a11yRemoveReaction(emoji)
+        } else {
+            return L10n.a11yAddReaction(emoji)
         }
     }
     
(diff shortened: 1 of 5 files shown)


ElementX/Sources/Screens/EmojiPickerScreen/View/EmojiPickerScreen.swift
```diff
@@ -36,6 +36,7 @@ struct EmojiPickerScreen: View {
                                         .background(Circle()
                                             .foregroundColor(emojiBackgroundColor(for: emoji.value)))
                                 }
+                                .accessibilityLabel(accessibilityLabel(for: emoji.value))
```
stefanceriu, line 39: Shouldn't this be using the underlying emoji `label`? I see value is set to `unicode` now e.g.

```
EmojiItem
  - label : "red heart"
  - unicode : "❤️"
  ▿ keywords : 4 elements
    - 0 : "emotion"
    - 1 : "heart"
    - 2 : "love"
    - 3 : "red"
  ▿ shortcodes : 1 element
    - 0 : "heart"
```

ElementX/Sources/Screens/Timeline/View/Supplementary/TimelineReactionsView.swift
```diff
@@ -138,11 +138,24 @@ struct TimelineCollapseButtonLabel: View {
...
+    private var accessibilityLabel: String {
+        if reaction.isHighlighted {
+            return reaction.count > 1 ? L10n.tr("Localizable", "screen_room_timeline_reaction_including_you_a11y", reaction.count - 1, reaction.displayKey) : L10n.screenRoomTimelineReactionYouA11y(reaction.displayKey)
+        }
+        return L10n.tr("Localizable", "screen_room_timeline_reaction_a11y", reaction.count, reaction.displayKey)
+    }
+    
+    private var toggleReactionAccessibilityActionName: String {
+        reaction.isHighlighted ? L10n.a11yRemoveReaction(reaction.displayKey) : L10n.a11yAddReaction(reaction.displayKey)
+    }
```
stefanceriu, line 157: Same question regarding the emoji label here.

ElementX/Sources/Screens/EmojiPickerScreen/View/EmojiPickerScreen.swift
Velin92: I am not sure in the other places of the code like the menu we use the string representing the emoji directly. I guess the voice over reads them the same way? But I have a feeling that the voice over will localize only the emoji Unicode directly actually...

ElementX/Sources/Screens/Timeline/View/Supplementary/TimelineReactionsView.swift
Velin92: yeah double checked we need to use the string directly that represents the emoji like ❤️, and the voice over will localize it

ElementX/Sources/Screens/EmojiPickerScreen/View/EmojiPickerScreen.swift
stefanceriu: You're right, the labels aren't localized so we don't have a choice 👍

stefanceriu (approved):

PR #4220, opened 2025-06-18
Update the SDK.

pixlwave:
There are basically no changes between the old and new version. The old one was built from a branch and the new one is built from `main` when that branch was merged.


ElementX/Sources/AppHooks/AppHooks.swift
pixlwave, line 1: Addresses https://github.com/element-hq/element-x-ios/pull/4175#discussion_r2137110542

This shouldn't have been added to the protocol.

PR #4224, opened 2025-06-18
Run Xcode select on the Translations and CalVer workflows.

pixlwave:
Seems the Translations PR workflow was failing since bumping the Swift tools version to 6.1. This should fix that 🤞


stefanceriu (approved):
👌

Velin92 (approved):

PR #4229, opened 2025-06-19
Fix MediaTimeline screen header iOS 26 crash, update remaining version predicates.

stefanceriu:


ElementX/Sources/Other/Extensions/PlatformViewVersionPredicate.swift
pixlwave, line 1: These should probably all be `.v17...` if they're staying in here to match the projects current targets.

pixlwave (approved):

ElementX/Sources/Other/Extensions/PlatformViewVersionPredicate.swift
stefanceriu: I'll okay, I have another PR incoming to move them to compound

PR #4235, opened 2025-06-20
Refactor how we deal with user permissions.

stefanceriu:
This patch introduces a new RoomPowerLevelsProxy and its associated mock and adopts newer Rust APIs to make working with user power levels and permissions easier.

### DIFF
--- a/ElementX/Sources/Mocks/RoomPowerLevelsProxyMock.swift
+++ b/ElementX/Sources/Mocks/RoomPowerLevelsProxyMock.swift
@@ -7,24 +7,35 @@
 
 import MatrixRustSDK
 
+struct RoomPowerLevelsProxyMockConfiguration {
+    var canUserSendMessage = true
+    var canUserSendState = false
+    var canUserInvite = true
+    var canUserRedactOther = false
+    var canUserRedactOwn = true
+    var canUserKick = false
+    var canUserBan = false
+    var canUserTriggerRoomNotification = false
+    var canUserPin = true
+    var canUserJoinCall = true
+}
+
 extension RoomPowerLevelsProxyMock {
-    struct Configuration { }
-    
-    convenience init(configuration: Configuration) {
+    convenience init(configuration: RoomPowerLevelsProxyMockConfiguration) {
         self.init()
         
         underlyingValues = RoomPowerLevelsValues.mock
         
-        canUserUserIDSendMessageReturnValue = .success(true)
-        canUserUserIDSendStateEventReturnValue = .success(true)
-        canUserInviteUserIDReturnValue = .success(true)
-        canUserRedactOtherUserIDReturnValue = .success(true)
-        canUserRedactOwnUserIDReturnValue = .success(true)
-        canUserKickUserIDReturnValue = .success(true)
-        canUserBanUserIDReturnValue = .success(true)
-        canUserTriggerRoomNotificationUserIDReturnValue = .success(true)
-        canUserPinOrUnpinUserIDReturnValue = .success(true)
-        canUserJoinCallUserIDReturnValue = .success(true)
+        canUserUserIDSendMessageReturnValue = .success(configuration.canUserSendMessage)
+        canUserUserIDSendStateEventReturnValue = .success(configuration.canUserSendState)
+        canUserInviteUserIDReturnValue = .success(configuration.canUserInvite)
+        canUserRedactOtherUserIDReturnValue = .success(configuration.canUserRedactOther)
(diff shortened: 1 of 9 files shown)


ElementX/Sources/Mocks/RoomPowerLevelsProxyMock.swift
```diff
@@ -0,0 +1,44 @@
...
+                              invite: 0,
+                              kick: 50,
+                              redact: 50,
+                              eventsDefault: 0,
+                              stateDefault: 50,
+                              usersDefault: 0,
+                              roomName: 50,
+                              roomAvatar: 50,
+                              roomTopic: 50)
+    }
```
Velin92: I think the defaults of the mocks should simulate either a normal user... or whatever were the default values we used in the JoinedRoomProxyMock returns.

Maybe we could also avoid creating this mock customisation entirely and rely only on the JoinedRoomProxyMock to handle permission mocking?

ElementX/Sources/Screens/KnockRequestsListScreen/KnockRequestsListScreenViewModel.swift
```diff
@@ -224,9 +224,10 @@ class KnockRequestsListScreenViewModel: KnockRequestsListScreenViewModelType, Kn
     }
     
     private func updatePermissions() async {
-        state.canAccept = await (try? roomProxy.canUserInvite(userID: roomProxy.ownUserID).get()) == true
-        state.canDecline = await (try? roomProxy.canUserKick(userID: roomProxy.ownUserID).get()) == true
-        state.canBan = await (try? roomProxy.canUserBan(userID: roomProxy.ownUserID).get()) == true
+        let powerLevels = try? await roomProxy.powerLevels().get()
+        state.canAccept = (try? powerLevels?.canUserInvite(userID: roomProxy.ownUserID).get()) == true
+        state.canDecline = (try? powerLevels?.canUserKick(userID: roomProxy.ownUserID).get()) == true
+        state.canBan = (try? powerLevels?.canUserBan(userID: roomProxy.ownUserID).get()) == true
```
Velin92, line 230: nice

ElementX/Sources/Screens/KnockRequestsListScreen/KnockRequestsListScreenViewModel.swift
pixlwave, line 1: So silly question. Can we now include this `powerLevels` struct in Rust's `roomInfo` so that we don't even need to await a get on it?

pixlwave: That way we could just use the `RoomInfoProxy` from the publisher in all these screens like we do for everything else.

ElementX/Sources/Mocks/RoomPowerLevelsProxyMock.swift
Velin92: Actually now that I have saw the whole PR, I think... we should get rid of these https://github.com/element-hq/element-x-ios/pull/4235/files#diff-7534da76896ef211f10438481d8616f0a0da4d6edf943e12b556816a9ec7601cR39-R43 and just provide the configuration for mocking the `PowerLevelProxyMock` in the init. If we don't provide it we will use the currently default values we have.

The problem is that we have these values, but then we have some tests that mock directly the `PowerLevelProxy` and other that mock only the value in the `RoomProxy` and I think this is going to get very confusing, also I don't like having this values `canUserDoX` be directly in the room proxy when the real function they are mocking is in the PowerLevelProxy.

Velin92 (changes requested):
the PR looks good from an implementation point of view, but the mocking part is a bit confusing, I think we should narrow permission mocking only to the `RoomPowerLevelsProxyMock` instrad of having this hybrid approach between the `RoomProxy` and the former.

ElementX/Sources/Mocks/RoomPowerLevelsProxyMock.swift
stefanceriu: The link doesn't work on my side but I imagine you mean the configurations from the JoinRoomProxy mock. It makes sense, let me see what I can do 👍

ElementX/Sources/Screens/KnockRequestsListScreen/KnockRequestsListScreenViewModel.swift
stefanceriu: Oh yeah, that's a really good idea. Let me see if I can push it through Rust side

Velin92 (approved):
LGTM

PR #4237, opened 2025-06-20
EC Timeout if it doesn't respond after 30 seconds

Velin92:
This is more for a future improvement, where we will handle the back button in the webview directly.

We check if the webview has sucessfully joined and if it doesn't happen within 30 seconds, we show an alert to allow the user to dismiss the whole view.

### DIFF
--- a/ElementX/Sources/Other/Extensions/String.swift
+++ b/ElementX/Sources/Other/Extensions/String.swift
@@ -137,3 +137,14 @@ extension String {
         return link
     }
 }
+
+extension String {
+    var asJsonDictionary: [String: Any]? {
+        get throws {
+            guard let data = data(using: .utf8) else {
+                return nil
+            }
+            return try JSONSerialization.jsonObject(with: data, options: []) as? [String: Any]
+        }
+    }
+}
--- a/ElementX/Sources/Other/Extensions/String.swift
+++ b/ElementX/Sources/Other/Extensions/String.swift
@@ -137,14 +137,3 @@ extension String {
         return link
     }
 }
-
-extension String {
-    var asJsonDictionary: [String: Any]? {
-        get throws {
-            guard let data = data(using: .utf8) else {
-                return nil
-            }
-            return try JSONSerialization.jsonObject(with: data, options: []) as? [String: Any]
-        }
-    }
-}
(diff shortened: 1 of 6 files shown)


ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
```diff
@@ -109,6 +112,14 @@ class CallScreenViewModel: CallScreenViewModelType, CallScreenViewModelProtocol
             .store(in: &cancellables)
         
         setupCall()
+        
+        timeoutTask = Task { [weak self] in
+            try? await Task.sleep(for: .seconds(30))
+            guard !Task.isCancelled, let self else { return }
+            MXLog.error("[CallscreenViewModel] failed to join Element Call: Timeout")
```
pixlwave: The file is already added to the log line by Rust.
```suggestion
            MXLog.error("Failed to join Element Call: Timeout")
```

ElementX/Sources/Other/Extensions/String.swift
```diff
@@ -137,3 +137,14 @@ extension String {
         return link
     }
 }
+
+extension String {
+    var asJsonDictionary: [String: Any]? {
```
pixlwave:
```suggestion
    var asJSONDictionary: [String: Any]? {
```

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
```diff
@@ -143,6 +154,16 @@ class CallScreenViewModel: CallScreenViewModelType, CallScreenViewModelProtocol
     }
     
     // MARK: - Private
+
+    private func handleWidgetAction(message: String) async {
+        if let jsonDictionary = try? message.asJsonDictionary,
+           jsonDictionary["api"] as? String == "fromWidget",
+           jsonDictionary["action"] as? String == "io.element.join" {
```
pixlwave: It would be nice to define this as a Codable struct instead of using `JSONSerialization` and a dictionary if we could?

Velin92: uhm I guess we can, in the end even if the JSON is massive I guess we can make a Codable struct out of only these two values, and discard the others.

pixlwave (approved):
Nice 👌

PR #4243, opened 2025-06-23
Fix a crash when attempting to send a bug report with excessively large logs.

pixlwave:
When a log file balloons in size (e.g. due to a tight loop) the app is terminated when sending a bug report due to memory usage. This is because we compare the file size against the upload limit *after* compressing it, but in order to compress it we load the whole file into memory.

We could fix that, but then the problematic log file would be silently dropped and in reality it is much more valuable to have the user contact us to say there's an issue with their logs so we can actually fix whatever is generating the excessively large log files.

This PR prevents the user from sending us logs if there's a file >1GB in size (the crashes were happening a around 3GB, but that's clearly way too much).

Screenshot 2025-06-23 at 5 25 52 pm


stefanceriu (approved):
Very nice, thank you! 👍

PR #4247, opened 2025-06-25
Adopt new `canOwnUser*` power level methods instead of the throwing ones

stefanceriu:
* Switch to using the power levels for retrieving a user's role.
* Bump the SDK to v25.06.25; adopt the new optional RoomInfo `joinRule` and `isPublic`.
* Make the room power levels optional as they can be missing depending on what state events are requested

### DIFF
--- a/ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenViewModel.swift
+++ b/ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenViewModel.swift
@@ -232,12 +232,9 @@ class RoomDetailsScreenViewModel: RoomDetailsScreenViewModelType, RoomDetailsScr
             state.canKickUsers = powerLevels.canOwnUserKick()
             state.canBanUsers = powerLevels.canOwnUserBan()
             state.canJoinCall = powerLevels.canOwnUserJoinCall()
+            state.canEditRolesOrPermissions = powerLevels.suggestedRole(forUser: roomProxy.ownUserID) == .administrator
         } else {
             fatalError("Missing room power levels")
-        }
-        
-        Task {
-            state.canEditRolesOrPermissions = await (try? roomProxy.suggestedRole(for: roomProxy.ownUserID).get()) == .administrator
         }
     }
     
--- a/ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenViewModel.swift
+++ b/ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenViewModel.swift
@@ -125,7 +125,7 @@ class RoomDetailsScreenViewModel: RoomDetailsScreenViewModelType, RoomDetailsScr
             }
             state.bindings.leaveRoomAlertItem = LeaveRoomAlertItem(roomID: roomProxy.id,
                                                                    isDM: roomProxy.isDirectOneToOneRoom,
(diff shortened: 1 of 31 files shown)


stefanceriu:
> LGTM, although I'm a bit confused as to why 1 version of this method throws and the other doesn't.

It's because the rust side needs to parse a user id string into it's own representation and that's fallible.

ElementX/Sources/Screens/RoomDetailsScreen/View/RoomDetailsScreen.swift
pixlwave, line 1: I don't have enough context, but should we maybe test for both of the conditions required for a valid snapshot instead of just one of them?

ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
pixlwave, line 1: I think we need to update the call room screen to not show a navigation bar to go alongside this change. I have no idea how it marries up with an SDK update without an EC update though 🤷‍♂️

ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenViewModel.swift
```diff
@@ -125,7 +125,7 @@ class RoomDetailsScreenViewModel: RoomDetailsScreenViewModelType, RoomDetailsScr
             }
             state.bindings.leaveRoomAlertItem = LeaveRoomAlertItem(roomID: roomProxy.id,
                                                                    isDM: roomProxy.isDirectOneToOneRoom,
-                                                                   state: roomProxy.infoPublisher.value.isPublic ? .public : .private)
+                                                                   state: roomProxy.infoPublisher.value.isPrivate ? .private : .public)
```
pixlwave: Can I suggest we make this `isPrivate` optional too so that we handle the optional case explicitly otherwise things are going to go horribly wrong down the line.

ElementX/Sources/Services/Room/RoomPowerLevelsProxy.swift
```diff
@@ -10,14 +10,27 @@ import MatrixRustSDK
...
     var values: RoomPowerLevelsValues {
         powerLevels.values()
     }
     
+    var userPowerLevels: [String: Int64] {
+        powerLevels.userPowerLevels()
+    }
+    
+    func suggestedRole(forUser userID: String) -> RoomMemberRole {
+        let powerLevel = powerLevels.userPowerLevels()[userID] ?? 0
```
pixlwave: This shouldn't `?? 0`, but instead should use `usersDefault` from the values.

pixlwave (approved):
LGTM, just a couple of small points 👇

ElementX/Sources/Services/Room/RoomPowerLevelsProxy.swift
stefanceriu: Oh very nice, thanks!

ElementX/Sources/Screens/RoomDetailsScreen/View/RoomDetailsScreen.swift
stefanceriu: We could but I see it's working fine with just one. We can tweak down the line depending on how it goes, I don't expect us changing these too much under normal circumstances.

PR #4250, opened 2025-06-25
EC: handle back navigation from the webview

Velin92:
WIP


stefanceriu (approved):
Nice, lgtm 👍  Have you updated the UI test snapshots for EC?

Velin92:
> Nice, lgtm 👍 Have you updated the UI test snapshots for EC?

good point let me update those

PR #4253, opened 2025-06-26
Add View Avatar a11y label

Velin92:
Add a view avatar accessibility when both the action and the avatar URL are present


pixlwave (approved):
LGTM, although I wonder if it's worth handling the case for hero avatars now so that if there was 4 heroes in a stack, the label is "View bob's avatar".

PR #4254, opened 2025-06-26
Use the close formatting option a11y label

Velin92:
Seems like we already had an accessibility label for it but was never used.


stefanceriu (approved):
👍

PR #4258, opened 2025-06-26
voice over focuses title when it changes in session verification view

Velin92:
The reason behind this PR is that when the view changes its title, or it disappears, it automatically shifts its focus to the buttons, so we should always refocus the title in voice over when it changes


pixlwave (approved):
LGTM 👌

PR #4260, opened 2025-06-26
Make the BuildSDK tool an AsyncParsableCommand and avoid help showing up after the command is run

stefanceriu:


pixlwave (approved):
Nice, thanks 😎

PR #4261, opened 2025-06-26
Be more lenient with the power levels

stefanceriu:
… as they can still be missing at the time the various screens are created e.g. after accepting an invite.

The correct solution is to subscribe to update and update the UI accordingly when receiving them, which this PR does along with converging on similar approaches in all the screens.

### DIFF
--- a/ElementX/Sources/Screens/KnockRequestsListScreen/KnockRequestsListScreenViewModel.swift
+++ b/ElementX/Sources/Screens/KnockRequestsListScreen/KnockRequestsListScreenViewModel.swift
@@ -26,7 +26,7 @@ class KnockRequestsListScreenViewModel: KnockRequestsListScreenViewModelType, Kn
         self.userIndicatorController = userIndicatorController
         super.init(initialViewState: KnockRequestsListScreenViewState(), mediaProvider: mediaProvider)
         
-        updateRoomInfo(roomInfo: roomProxy.infoPublisher.value)
+        updateRoomInfo(roomProxy.infoPublisher.value)
         
         setupSubscriptions()
     }
@@ -185,7 +185,7 @@ class KnockRequestsListScreenViewModel: KnockRequestsListScreenViewModelType, Kn
         roomProxy.infoPublisher
             .receive(on: DispatchQueue.main)
             .sink { [weak self] roomInfo in
-                self?.updateRoomInfo(roomInfo: roomInfo)
+                self?.updateRoomInfo(roomInfo)
             }
             .store(in: &cancellables)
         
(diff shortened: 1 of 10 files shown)


ElementX/Sources/Screens/KnockRequestsListScreen/KnockRequestsListScreenViewModel.swift
```diff
@@ -26,7 +26,14 @@ class KnockRequestsListScreenViewModel: KnockRequestsListScreenViewModelType, Kn
         self.userIndicatorController = userIndicatorController
         super.init(initialViewState: KnockRequestsListScreenViewState(), mediaProvider: mediaProvider)
         
-        updateRoomInfo(roomInfo: roomProxy.infoPublisher.value)
+        roomProxy.infoPublisher
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] roomInfo in
+                self?.updateRoomInfo(roomInfo)
+            }
+            .store(in: &cancellables)
```
pixlwave: This one is already done in `setupSubscriptions`

ElementX/Sources/Screens/RoomRolesAndPermissionsScreen/RoomRolesAndPermissionsScreenViewModel.swift
pixlwave, line 1: I think we need to do something different here. The whole point of this screen is it is showing you the current power levels, so if this doesn't happen you could potentially ruin the permissions.

Can we make them optional in the view state and using an `.isWaiting` in the `ListRow` or something?

pixlwave (approved):
LGTM!

ElementX/Sources/Screens/KnockRequestsListScreen/KnockRequestsListScreenViewModel.swift
stefanceriu: Oups, overzealous

ElementX/Sources/Screens/RoomRolesAndPermissionsScreen/RoomRolesAndPermissionsScreenViewModel.swift
stefanceriu: It turns out somebody already took care of that 😛 

Screenshot 2025-06-26 at 17 18 02

pixlwave: Ha! 😎

pixlwave: Ohhhh, I guess they were async before weren't they.

PR #4262, opened 2025-06-26
Replace the Report a Problem button with the app's version on the start screen.

pixlwave:
Tapping it 7-times will secretly reveal the bug report screen 🤫

https://github.com/user-attachments/assets/17ade48c-2023-4d92-8bdd-536036117bda


ElementX/Resources/Localizations/en.lproj/Localizable.strings
```diff
@@ -794,6 +794,7 @@
 "screen_notification_settings_system_notifications_action_required_content_link" = "system settings";
 "screen_notification_settings_system_notifications_turned_off" = "System notifications turned off";
 "screen_notification_settings_title" = "Notifications";
+"screen_onboarding_app_version" = "Version %1$@";
```
pixlwave, line 797: @bmarty I added `screen_onboarding_app_version` for this so we can keep it simple: `Version 25.06.5`

bmarty: ah just seen this, I have just added screen_onboarding_version with the same value :). Will remove mine!

stefanceriu:
> Tapping it 7-times will secretly reveal the bug report screen 🤫

Oh noes, wouldn't a single tap be enough?

bmarty:
> > Tapping it 7-times will secretly reveal the bug report screen 🤫
> 
> Oh noes, wouldn't a single tap be enough?

Maybe a single tap to reveal the previous "Report a problem" button would be the best solution?

manuroe:
> This looks like it's going to create more friction for people that can't login but the PR lgtm 👍

Friction is on the matrix server admin side to experiment and validate the authentication flow. "Normal", non techy users should not need this report button here.

stefanceriu:
> Friction is on the matrix server admin side to experiment and validate the authentication flow. "Normal", non techy users should not need this report button here.

Sure, but for public users that's going to fall on us to help with logs and guide them through reporting problems.

pixlwave:
> Sure, but for public users that's going to fall on us to help with logs and guide them through reporting problems.

Honestly I think it's fine. Right now, if someone sends a bug report without already being in direct contact with us, it's literally impossible for us to reply to them and point out their issue. If you look at all the rageshakes from users without a Matrix ID, the vast majority are non-actionable and simply creating noise.

pixlwave:
Also worth noting: This was probably more useful when people were spinning up a sliding sync proxy, configuring their well-known for the proxy (and MAS) etc. Nowadays EX can login to a fresh Synapse via password and ESS makes spinning up a Synapse + MAS pretty damn easy too.

PR #4279, opened 2025-07-02
Update Element Call to the actual release of 0.13.0.

pixlwave:
As requested by @toger5 this PR updates Element Call to the necessary version for the next release.


toger5:
Looks very simple. We just need to make sure we test it!

pixlwave:
Calls appear to be working well on a local build for me.

PR #4280, opened 2025-07-02
Downgrade Element Call back to 0.12.2 for now.

pixlwave:
This reverts "EC: handle back navigation from the webview" 37cbf733d49aa774c43c24c0e42290ba8f0260c6.

0.13.0 introduced a regression with bluetooth headsets.

### DIFF
--- a/ElementX.xcodeproj/project.pbxproj
+++ b/ElementX.xcodeproj/project.pbxproj
@@ -8771,7 +8771,7 @@
 			repositoryURL = "https://github.com/element-hq/element-call-swift";
 			requirement = {
 				kind = exactVersion;
-				version = 0.13.0;
+				version = 0.12.2;
 			};
 		};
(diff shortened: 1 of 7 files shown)


toger5 (changes requested):
We also need to set header: .none instead of .appBar

toger5:
Maybe check what pr inteoduced header:appbar and revert this as well

pixlwave:
> Maybe check what pr inteoduced header:appbar and revert this as well

Looks like it was just part of an SDK upgrade, nothing else in there is relevant so I'll just set it to `.none` 👍

pixlwave:
> We also need to set header: .none instead of .appBar

Done in https://github.com/element-hq/element-x-ios/pull/4280/commits/d6963d6515456bcb5a1b529851dcdc356404ad14

pixlwave:
We're going to fix the regression.

PR #4288, opened 2025-07-04
Add support for threaded read receipts

stefanceriu:
- add support for the new client wide threads feature flag
- send all receipts through the active timeline and remove unecessary making of a room as read (already handled)

### DIFF
--- a/ElementX/Sources/Services/Timeline/TimelineController/TimelineController.swift
+++ b/ElementX/Sources/Services/Timeline/TimelineController/TimelineController.swift
@@ -124,16 +124,11 @@ class TimelineController: TimelineControllerProtocol {
     func sendReadReceipt(for itemID: TimelineItemIdentifier) async {
         let receiptType: MatrixRustSDK.ReceiptType = appSettings.sharePresence ? .read : .readPrivate
         
-        // Mark the whole room as read if it's the last timeline item
-        if timelineItems.last?.id == itemID {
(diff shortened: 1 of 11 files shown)


ElementX/Sources/Services/Timeline/TimelineController/TimelineController.swift
```diff
@@ -124,16 +124,11 @@ class TimelineController: TimelineControllerProtocol {
     func sendReadReceipt(for itemID: TimelineItemIdentifier) async {
         let receiptType: MatrixRustSDK.ReceiptType = appSettings.sharePresence ? .read : .readPrivate
         
-        // Mark the whole room as read if it's the last timeline item
-        if timelineItems.last?.id == itemID {
-            _ = await roomProxy.markAsRead(receiptType: receiptType)
```
pixlwave, line 129: I think we still need to use this somehow as I think it handles hidden events too. Without that, rooms could become stuck as unread.

pixlwave (approved):
LGTM, just 1 little thought below.

ElementX/Sources/Services/Timeline/TimelineController/TimelineController.swift
stefanceriu: Talked about if offline. Seem to work well without it so we're going to merge it as is.

PR #4304, opened 2025-07-07
Accessibility label for editing the avatar

### LINKED ISSUE #4303: [Task] Add edit avatar a11y label

Velin92:
fixes #4303

### DIFF
--- a/ElementX/Sources/Screens/RoomDetailsEditScreen/View/RoomDetailsEditScreen.swift
+++ b/ElementX/Sources/Screens/RoomDetailsEditScreen/View/RoomDetailsEditScreen.swift
@@ -60,6 +60,7 @@ struct RoomDetailsEditScreen: View {
                                    contentID: context.viewState.roomID,
                                    avatarSize: .user(on: .memberDetails),
                                    mediaProvider: context.mediaProvider)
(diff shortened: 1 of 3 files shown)


ElementX/Sources/Screens/RoomDetailsEditScreen/View/RoomDetailsEditScreen.swift
```diff
@@ -60,6 +60,7 @@ struct RoomDetailsEditScreen: View {
                                    contentID: context.viewState.roomID,
                                    avatarSize: .user(on: .memberDetails),
                                    mediaProvider: context.mediaProvider)
+                .accessibilityLabel(context.viewState.canEditAvatar ? L10n.a11yEditAvatar : L10n.a11yAvatar)
```
pixlwave: As we disable the button when you can't edit the avatar, I _think_ this will read out "Avatar, button, disabled"? If it does, do you think it would make more sense for this to say "Edit avatar, button, disabled"?

pixlwave: Only mentioning this in case it sounds to the user like the entire avatar is disabled.

PR #4306, opened 2025-07-08
Aadded sender name when voiceover focuses a poll title

### LINKED ISSUE #4305: [Task] Add sender name to voice over when focusing a poll title

Velin92:
fixes #4305

### DIFF
--- a/ElementX/Sources/Screens/Timeline/View/Polls/PollView.swift
+++ b/ElementX/Sources/Screens/Timeline/View/Polls/PollView.swift
@@ -42,6 +42,7 @@ struct PollView: View {
     
     let poll: Poll
     let state: PollViewState
+    let sender: TimelineItemSender
     let actionHandler: (PollViewAction) -> Void
     
     var body: some View {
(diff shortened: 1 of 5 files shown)


ElementX/Sources/Screens/Timeline/View/Polls/PollView.swift
```diff
@@ -42,14 +42,27 @@ struct PollView: View {
...
+                            if poll.createdByAccountOwner {
+                                Text(L10n.commonYou)
+                            } else {
+                                Text(sender.disambiguatedDisplayName ?? sender.id)
+                            }
+                            label
+                        }
+                } else {
+                    questionView
+                }
```
pixlwave: This is untested, but I *think* you could do this instead:
```suggestion
                questionView
                    .accessibilityRepresentation {
                        if poll.createdByAccountOwner {
                            Text(L10n.commonYou)
                        } else {
                            Text(sender.disambiguatedDisplayName ?? sender.id)
                        }
                        questionView
                    }
```

pixlwave: Also, it might be nice to extract the sender string into something like

```
var sender: String { poll.createdByAccountOwner ? L10n.commonYou : sender.disambiguatedDisplayName ?? sender.id }
```

That would help keep the view a bit neater :)

PR #4312, opened 2025-07-10
FF for enabling share pos and defaults to `false`

Velin92:
Also updated the SDK


pixlwave:
Maybe the PR title should say that it disables the share_pos for everybody too?

PR #4322, opened 2025-07-14
Automatic Accessibility Audits on previews part 1

Velin92:
This PR is the first step in generating automatic audit accessibility tests (using the `performAccessibilityAudit` function provided by the UI Tests framework).

## What does the PR do?

Automatically runs for each `TestablePreview` we have in app a `performAccessibilityAudit` which are a series of different type of checks that validate if the app is meeting Apple standards for accessibility, they include stuff like dynamic types, labelling, hit area and more... 
- https://developer.apple.com/documentation/accessibility/performing-accessibility-audits-for-your-app
- https://developer.apple.com/documentation/XCTest/XCUIApplication/performAccessibilityAudit(for:_:)

It should be noted that sometimes this test creates some false positives regarding specific test cases, for example if we made a button with a fixed size, it will be flagged as not supporting dynamic types.

## What does the PR contain?

- Sets up an accessibility tests project
- The .stencil and the .yml required to generate a dictionary that associates each identifier to a preview type, which is used by the app to instantiate the correct preview type given the identifier
- The setup for the accessibility tests coordinator, which will use the UI signalling framework to iterate through all the previews while testing (the reason is that each preview type is composed by multiple previews each one with their own name, and to iterate on them we would need to import everything in the accessibility tests which are UI tests and that is not possible)
- A simple example test case that has been committed that will be used as basis to generate all the possible tests

## What is missing?

- A way to generate tests for each preview type
- A way to run these tests on CI and gather meaningful info on the failures
- Maybe also a way to filter out what views need the tests, and which specific type of audits are required for it (maybe we don't want to run dynamic size testing on all the views for example but only on some of them)


AccessibilityTests/SupportingFiles/target.yml
```diff
@@ -0,0 +1,64 @@
...
+    - package: KeychainAccess
+    - package: Kingfisher
+    - package: KZFileWatchers
+    - package: PostHog
+    - package: SwiftState
+    - package: GZIP
+    - package: Sentry
+    - package: SnapshotTesting
+    - package: Collections
+    - package: SwiftOGG
```
pixlwave: Did you check if we can remove any of these?

ElementX/Sources/AccessibilityTests/AccessibilityTestsAppCoordinator.swift
```diff
@@ -0,0 +1,152 @@
...
+// Copyright 2025 New Vector Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Combine
+import MatrixRustSDK
+import SwiftUI
+import UIKit
```
pixlwave: Should be fine to just import SwiftUI here.

```suggestion
import SwiftUI
```

ElementX/Sources/AccessibilityTests/AccessibilityTestsAppCoordinator.swift
```diff
@@ -0,0 +1,152 @@
...
+        
+        let analyticsClient = AnalyticsClientMock()
+        analyticsClient.isRunning = false
+        ServiceLocator.shared.register(analytics: AnalyticsService(client: analyticsClient,
+                                                                   appSettings: ServiceLocator.shared.settings))
+        
+        guard let name = ProcessInfo.accessibilityViewID,
+              let previewType = TestablePreviewsDictionary.dictionary[name]
+        else { fatalError("Unable to launch with unknown screen.")
+        }
```
pixlwave: Probably should either be on 1 line or 3 :)

```suggestion
        else { fatalError("Unable to launch with unknown screen.") }
```

ElementX/Sources/AccessibilityTests/AccessibilityTestsAppCoordinator.swift
```diff
@@ -0,0 +1,152 @@
...
+                        await previewsWrapper.updateCurrentIndex()
+                        do {
+                            guard !previewsWrapper.isDone else {
+                                try client.send(.noMorePreviews)
+                                return
+                            }
+                            
+                            try client.send(.nextPreviewReady(name: previewsWrapper.previewName))
+                        } catch {
+                            fatalError("failed sending signal: \(signal)")
```
pixlwave:
```suggestion
                            fatalError("Failed sending signal: \(signal)")
```

ElementX/Sources/AccessibilityTests/AccessibilityTestsAppCoordinator.swift
```diff
@@ -0,0 +1,152 @@
...
+            wrapper.currentPreview.content
+                .id(wrapper.previewName)
+        }
+    }
+}
+
+final class PreviewsWrapper: ObservableObject {
+    private let name: String
+    private let previews: [_Preview]
+    @Published var currentIndex = -1
```
pixlwave: Can we try this with observation instead?

```suggestion
@Observable final class PreviewsWrapper {
    private let name: String
    private let previews: [_Preview]
    private(set) var currentIndex = -1
```

ElementX/Sources/AccessibilityTests/AccessibilityTestsAppCoordinator.swift
```diff
@@ -0,0 +1,152 @@
...
+            }
+            .store(in: &cancellables)
+        } catch {
+            fatalError("Unable to start client signalling")
+        }
+    }
+}
+
+struct PreviewsWrapperView: View {
+    @ObservedObject var wrapper: PreviewsWrapper
```
pixlwave: And then also

```suggestion
    let wrapper: PreviewsWrapper
```

ElementX/Sources/Other/Extensions/TestablePreviewExtension.swift
pixlwave, line 1: I think this is an old generated file and instead we only need TestablePreviewDictionary.swift?

ElementX/Sources/Other/TestablePreview/TestablePreviewsDictionary.swift
```diff
@@ -0,0 +1,207 @@
...
+// swiftlint:disable all
+// swiftformat:disable all
+
+import SwiftUI
+
+
+enum TestablePreviewsDictionary {
+    static let dictionary: [String: any PreviewProvider.Type] = {
+        var previews: [String: any PreviewProvider.Type] = [:]
+        previews["AdvancedSettingsScreen_Previews"] = AdvancedSettingsScreen_Previews.self
```
pixlwave: Mega mega minor, but seeing as it's generated, would it make more sense to generate it as a pre-filled dictionary as this is currently mutating it to add each one.

```
static let dictionary: [String: any PreviewProvider.Type] = [
    "AdvancedSettingsScreen_Previews": AdvancedSettingsScreen_Previews.self
    …
]
```

ElementX/Sources/UITests/UITestsSignalling.swift
```diff
@@ -29,6 +29,11 @@ enum UITestsSignal: Codable, Equatable {
     
     /// Posts a notification.
     case notification(name: Notification.Name)
+    
+    /// Asks for the next preview
+    case nextPreview
+    case nextPreviewReady(name: String)
+    case noMorePreviews
```
pixlwave: I realised this would be a bit of a refactor, but could we do the same here as we've done with the timeline above and have

```suggestion
    case accessibilityAudit(AccessibilityAudit)
    enum AccessibilityAudit: Codable, Equatable {
        /// Ask the app for the next preview.
        case nextPreview
        /// Tell the test runner about a loaded preview.
        case nextPreviewReady(name: String)
        /// Tell the test runner that there are no more previews.
        case noMorePreviews
    }
```

PR #4325, opened 2025-07-16
Accessibiliy Tests part 2

Velin92:
## What does this PR include

- A generated test for each testable preview
- Some fixes to not make tests crash
- CI setup that archives the results
- a way for specific accessibility elements to be reported as a failure for a specific test subset or error
- Filtering some other false positives like matrix entities being marked as non readable, or partially supported dynamic types for which we are fine with
- A way to handle that nasty location sharing popup which broke some of the tests

## What is missing?

- a way for the previews do decide if they should be included in the accessibility tests
- Contrast and text clipping tests are avoided, since the requirements for Apple seem to be way stricter than what is considered fine by our design standard, maybe would be worth looking into them at a later point?

### DIFF
--- a/AccessibilityTests/Sources/AccessibilityTests.swift
+++ b/AccessibilityTests/Sources/AccessibilityTests.swift
@@ -17,18 +17,18 @@ final class AccessibilityTests: XCTestCase {
         await client.waitForApp()
         defer { try? client.stop() }
         
+        try client.send(.accessibilityAudit(.nextPreview))
         // To handle system interrupts
-        let allowButtonPredicate = NSPredicate(format: "label == 'Allow Once' || label == 'Allow While Using App'")
-        _ = addUIInterruptionMonitor(withDescription: "Allow to access your location?") { alert -> Bool in
-            let alwaysAllowButton = alert.buttons.matching(allowButtonPredicate).element.firstMatch
+        _ = addUIInterruptionMonitor(withDescription: "Location access alert handler") { alert in
+            let alwaysAllowButton = alert.buttons["Allow While Using App"]
             if alwaysAllowButton.exists {
                 alwaysAllowButton.tap()
                 return true
             }
             return false
         }
+        app.tap()
         
-        try client.send(.accessibilityAudit(.nextPreview))
         forLoop: for await signal in client.signals.values {
             switch signal {
             case .accessibilityAudit(let auditSignal):
--- a/AccessibilityTests/Sources/AccessibilityTests.swift
+++ b/AccessibilityTests/Sources/AccessibilityTests.swift
@@ -34,33 +34,8 @@ final class AccessibilityTests: XCTestCase {
             case .accessibilityAudit(let auditSignal):
                 switch auditSignal {
                 case .nextPreviewReady(let name):
-                    // Alows us to log the name of the preview that is being tested
-                    XCTContext.runActivity(named: name) { _ in
-                        do {
-                            // We have removed `textClipping` and `contrast` for now
-                            try app.performAccessibilityAudit(for: [.dynamicType, .elementDetection, .hitRegion, .sufficientElementDescription, .trait]) { issue in
-                                // Removew false positives for null elements
-                                guard let element = issue.element else {
-                                    return true
-                                }
-                                
-                                // Filter out false positives for specific cases
-                                if AccessibilityTests.ignoredA11yIdentifiers[element.identifier]?.contains(issue.auditType) == true {
-                                    return true
-                                }
-                                
-                                // We are fine with elements that only partially support dynamic types
-                                guard issue.compactDescription != "Dynamic Type font sizes are partially unsupported" else {
-                                    return true
-                                }
-                                
-                                return false
-                            }
-                        } catch {
-                            XCTFail("Failed to perform the accessibility audit: \(error)")
-                        }
-                    }
-                    try? client.send(.accessibilityAudit(.nextPreview))
+                    performAccessibilityAuditForPreview(named: name)
+                    try client.send(.accessibilityAudit(.nextPreview))
                 case .noMorePreviews:
                     break forLoop
                 default:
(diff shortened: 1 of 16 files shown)


AccessibilityTests/Sources/AccessibilityTests.swift
```diff
@@ -0,0 +1,78 @@
...
+                                guard issue.compactDescription != "Dynamic Type font sizes are partially unsupported" else {
+                                    return true
+                                }
+                                
+                                return false
+                            }
+                        } catch {
+                            XCTFail("Failed to perform the accessibility audit: \(error)")
+                        }
+                    }
```
pixlwave: Maybe we should extract this into its own function to help with readability?

AccessibilityTests/Sources/GeneratedTests.swift
```diff
@@ -0,0 +1,579 @@
+// Generated using Sourcery 2.2.7 — https://github.com/krzysztofzablocki/Sourcery
+// DO NOT EDIT
+
+// swiftlint:disable all
+// swiftformat:disable all
+
+
+extension AccessibilityTests {
+    func test_AdvancedSettingsScreen_Previews() async throws {
```
pixlwave: Not a biggy, but it would be nice if we could generate these as `testAdvancedSettingsScreen` so they match the preview tests.

pixlwave: And maybe with a blank line in-between too.

Tools/Sourcery/AccessibilityTests.yml
```diff
@@ -0,0 +1,7 @@
+sources:
+  include:
+    - ../../ElementX
+templates:
+  - AccessibilityTests.stencil
+output:
+  ../../AccessibilityTests/Sources/GeneratedTests.swift
```
pixlwave: And here, to match the preview tests could we generate the file as `GeneratedAccessibilityTests.swift` (so you can search for AccessibilityTests.swift and find both files).

fastlane/Fastfile
```diff
@@ -85,6 +85,27 @@ lane :ui_tests do |options|
...
+    name: "iPhone-18.5",
+    type: "com.apple.CoreSimulator.SimDeviceType.iPhone-16",
+    runtime: "com.apple.CoreSimulator.SimRuntime.iOS-18-4"
+  )
+
+  reset_simulator = ENV.key?('CI')
+  
+  run_tests(
+    scheme: "AccessibilityTests",
+    device: "iPhone-18.5",
```
pixlwave: Rather than using the UI test runner devices should we remove the `create_simulator_if_necessary` and simply do this like we do for unit tests?

```suggestion
    device: "iPhone 16 (18.5)",
```

I don't think it matters so much given we don't need pixel perfect snapshots out of it.

ElementX/Sources/AccessibilityTests/AccessibilityTestsAppCoordinator.swift
```diff
@@ -143,7 +143,9 @@ struct PreviewsWrapperView: View {
         
         switch fulfillmentSource {
         case .publisher(let publisher):
-            _ = await publisher.values.first { $0 == true }
+            _ = await publisher
+                .timeout(.seconds(1), scheduler: DispatchQueue.main)
+                .values.first { $0 == true }
```
pixlwave: Oh neat 👌

pixlwave (commented):
Love it!

AccessibilityTests/Sources/AccessibilityTests.swift
```diff
@@ -59,12 +60,17 @@ final class AccessibilityTests: XCTestCase {
                     guard let element = issue.element else {
                         return true
                     }
-                    
+
```
pixlwave:
```suggestion
                    
```

AccessibilityTests/Sources/AccessibilityTests.swift
```diff
@@ -78,7 +84,12 @@ final class AccessibilityTests: XCTestCase {
         }
     }
     
+    private static func isMatrixString(_ string: String) -> Bool {
```
pixlwave: Minor, but this seems more descriptive to me.
```suggestion
    private static func isMatrixIdentifier(_ string: String) -> Bool {
```

AccessibilityTests/Sources/AccessibilityTests.swift
```diff
@@ -78,7 +84,12 @@ final class AccessibilityTests: XCTestCase {
...
+        MatrixEntityRegex.isMatrixRoomAlias(string) || MatrixEntityRegex.isMatrixUserIdentifier(string)
+    }
+    
     private static let partiallyUnsupportedDynamicTypeMessage = "Dynamic Type font sizes are partially unsupported"
+    private static let notHumanReadableMessage = ""
```
pixlwave: I think we should be setting this and using it above?

```suggestion
    private static let notHumanReadableMessage = "Label not human-readable"
```

ElementX/Sources/AccessibilityTests/AccessibilityTestsAppCoordinator.swift
```diff
@@ -51,8 +52,10 @@ class AccessibilityTestsAppCoordinator: AppCoordinatorProtocol {
             fatalError("Unable to launch with unknown screen.")
         }
         previewsWrapper = .init(name: name, previews: previewType._allPreviews)
-        
+
```
pixlwave:
```suggestion
        
```

ElementX/Sources/AccessibilityTests/AccessibilityTestsAppCoordinator.swift
```diff
@@ -51,8 +52,10 @@ class AccessibilityTestsAppCoordinator: AppCoordinatorProtocol {
             fatalError("Unable to launch with unknown screen.")
         }
         previewsWrapper = .init(name: name, previews: previewType._allPreviews)
-        
+                
         setupSignalling()
+        // Used to get rid of the request check for the rest of the tests on CI
+        CLLocationManager().requestWhenInUseAuthorization()
```
pixlwave:
```suggestion
        // Used to perform the request check before the tests run on CI
        CLLocationManager().requestWhenInUseAuthorization()
```

ElementX/Sources/Other/Pills/PillUtilities.swift
```diff
@@ -8,10 +8,8 @@
 import Foundation
 
 enum PillUtilities {
+    // Make sure to change also the matrix entity regex one
```
pixlwave: Think we can get rid of this now too.
```suggestion
```

pixlwave (approved):
LGTM 👌

PR #4326, opened 2025-07-16
Show an alert when entering an account provider that requires Element Pro.

pixlwave:
https://github.com/user-attachments/assets/0affbd42-d1e6-4b73-8ae7-15461b861514


ElementX/Sources/AppHooks/Hooks/ElementWellKnownHook.swift
```diff
@@ -0,0 +1,50 @@
...
+
+private struct ElementWellKnown: Decodable {
+    var version: Int
+    var enforceElementPro: Bool?
+    
+    enum CodingKeys: String, CodingKey {
+        case version
+        case enforceElementPro = "enforce_element_pro"
+    }
+}
```
pixlwave, line 49: Note to self: Make sure this is the final format before merging.

Velin92 (approved):

ElementX/Sources/Services/Authentication/AuthenticationService.swift
```diff
@@ -52,6 +52,7 @@ class AuthenticationService: AuthenticationServiceProtocol {
             var homeserver = LoginHomeserver(address: homeserverAddress, loginMode: .unknown)
             
             let client = try await makeClient(homeserverAddress: homeserverAddress)
+            try await appHooks.elementWellKnownHook.validate(using: client).get()
```
bmarty, line 55: @pixlwave just to confirm, Element Pro application will not perform any check regarding the enforcement with this code, right?

bmarty: And now I see this [diff](https://github.com/element-hq/element-ios-enterprise/compare/233f49aa2063ef7113007e7ea6a90225be903e5e...245f823e92296bace1dcef31f358025b4dd7a4a9). So I guess it's fine!

On Android, I'll just check that the build is not an Enterprise build so I do not need to update the enterprise module at all.

PR #4337, opened 2025-07-21
Refactor Rageshake URL overrides and Target configuration.

pixlwave:
This PR moves the rageshakeURL configuration from the BugReportService to AppSettings (#4267 should have done it like this now that I've seen what using the API is like 🙈) and refactors Target to return <del>a handle</del> the configuration when configuring.

Note that these changes are unused but are in preparation for another PR.


ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -65,7 +65,6 @@ final class AppSettings {
     }
     
     private static var suiteName: String = InfoPlistReader.main.appGroupIdentifier
-    private static var remoteSuiteName = "\(InfoPlistReader.main.appGroupIdentifier).remote"
```
pixlwave, line 68: This was unused

ElementX/Sources/Application/Settings/ConfigurableSetting.swift
pixlwave, line 1: There's probably a better name for this. Maybe `RemotelyConfiguredSetting` or something with `setRemoteValue` and `clearRemoteValue` or similar?

stefanceriu: Seeing we have a `UserPreference` perhaps a `ExternallyConfiguredPreference`?

pixlwave: Going to handle this in the next PR as I'll rename a few other things too.

PR #4340, opened 2025-07-21
A11y test detected improvements

Velin92:
For the complete list of fixes check: 
https://github.com/element-hq/element-x-ios/issues/4336

Also this adds some exceptions to the test and improved the dynamic type testing to avoid false positives.
We also handle location sharing by using the springboard since its faster and more consistent than the interrupt monitor

Is better to review this PR commit by commit

### DIFF
--- a/AccessibilityTests/Sources/AccessibilityTests.swift
+++ b/AccessibilityTests/Sources/AccessibilityTests.swift
@@ -91,9 +91,9 @@ final class AccessibilityTests: XCTestCase {
     private static let notHumanReadableMessage = "Label not human-readable"
     
     /// Use this array to filter add specific filters to ignore specific issues for certain elements
-    private static let ignoredA11yIdentifiers: [String: [FilterType]] = [A11yIdentifiers.authenticationStartScreen.appVersion: [.auditType(.hitRegion)],
-                                                                         A11yIdentifiers.pollView.summary: [.auditType(.hitRegion)],
-                                                                         A11yIdentifiers.serverConfirmationScreen.serverPicker: [.compactDescription(notHumanReadableMessage)]]
+    private static let ignoredA11yIdentifiers: [String: [FilterType]] = [
+        A11yIdentifiers.serverConfirmationScreen.serverPicker: [.compactDescription(notHumanReadableMessage)]
+    ]
 }
 
 private enum FilterType {
--- a/ElementX/Sources/Other/AccessibilityIdentifiers.swift
+++ b/ElementX/Sources/Other/AccessibilityIdentifiers.swift
@@ -42,7 +42,6 @@ enum A11yIdentifiers {
     static let pollFormScreen = PollFormScreen()
     static let roomPollsHistoryScreen = RoomPollsHistoryScreen()
     static let manageRoomMemberSheet = ManageRoomMemberSheet()
-    static let pollView = PollView()
     
     struct AlertInfo {
         let primaryButton = "alert_info-primary_button"
@@ -297,9 +296,5 @@ enum A11yIdentifiers {
     
     struct ManageRoomMemberSheet {
         let viewProfile = "manage_room_member_sheet-view_profile"
-    }
-    
-    struct PollView {
-        let summary = "poll_view-summary"
     }
 }
--- a/ElementX/Sources/Screens/HomeScreen/View/HomeScreenRecoveryKeyConfirmationBanner.swift
+++ b/ElementX/Sources/Screens/HomeScreen/View/HomeScreenRecoveryKeyConfirmationBanner.swift
@@ -68,9 +68,6 @@ struct HomeScreenRecoveryKeyConfirmationBanner: View {
                         Image(systemName: "xmark")
                             .foregroundColor(.compound.iconSecondary)
                             .frame(width: 12, height: 12)
-                            // Increase hit area
-                            .padding(.leading, 5)
-                            .padding(.bottom, 2)
                     }
                 }
             }
--- a/ElementX/Sources/Screens/MediaUploadPreviewScreen/View/MediaUploadPreviewScreen.swift
+++ b/ElementX/Sources/Screens/MediaUploadPreviewScreen/View/MediaUploadPreviewScreen.swift
@@ -82,9 +82,6 @@ struct MediaUploadPreviewScreen: View {
             context.isPresentingMediaCaptionWarning = true
         } label: {
             CompoundIcon(\.infoSolid, size: .xSmall, relativeTo: .compound.bodyLG)
-                // Increase hit area
-                .padding(.leading, 5)
-                .padding(.vertical, 2)
         }
         .tint(.compound.iconCriticalPrimary)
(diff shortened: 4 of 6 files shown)


AccessibilityTests/Sources/AccessibilityTests.swift
```diff
@@ -92,7 +91,9 @@ final class AccessibilityTests: XCTestCase {
     private static let notHumanReadableMessage = "Label not human-readable"
     
     /// Use this array to filter add specific filters to ignore specific issues for certain elements
-    private static let ignoredA11yIdentifiers: [String: [FilterType]] = [A11yIdentifiers.authenticationStartScreen.appVersion: [.auditType(.hitRegion)]]
+    private static let ignoredA11yIdentifiers: [String: [FilterType]] = [A11yIdentifiers.authenticationStartScreen.appVersion: [.auditType(.hitRegion)],
+                                                                         A11yIdentifiers.pollView.summary: [.auditType(.hitRegion)],
+                                                                         A11yIdentifiers.serverConfirmationScreen.serverPicker: [.compactDescription(notHumanReadableMessage)]]
```
pixlwave: Super minor nit just for readability could we do it like this?

```suggestion
    private static let ignoredA11yIdentifiers: [String: [FilterType]] = [
        A11yIdentifiers.authenticationStartScreen.appVersion: [.auditType(.hitRegion)],
        A11yIdentifiers.pollView.summary: [.auditType(.hitRegion)],
        A11yIdentifiers.serverConfirmationScreen.serverPicker: [.compactDescription(notHumanReadableMessage)]
    ]
```

ElementX/Sources/Other/AccessibilityIdentifiers.swift
```diff
@@ -42,6 +42,7 @@ enum A11yIdentifiers {
     static let pollFormScreen = PollFormScreen()
     static let roomPollsHistoryScreen = RoomPollsHistoryScreen()
     static let manageRoomMemberSheet = ManageRoomMemberSheet()
+    static let pollView = PollView()
```
pixlwave: Should we maybe put this before/after `pollFormScreen`?

ElementX/Sources/Other/AccessibilityIdentifiers.swift
```diff
@@ -297,4 +298,8 @@ enum A11yIdentifiers {
     struct ManageRoomMemberSheet {
         let viewProfile = "manage_room_member_sheet-view_profile"
     }
+    
+    struct PollView {
+        let summary = "poll_view-summary"
+    }
```
pixlwave: Same comment but with `PollFormScreen`

ElementX/Sources/Other/SwiftUI/Views/TombstonedAvatarImage.swift
```diff
@@ -25,6 +25,7 @@ struct TombstonedAvatarImage: View {
                     .font(.system(size: geometry.size.width * 0.5625, weight: .semibold))
                     .minimumScaleFactor(0.001)
                     .frame(alignment: .center)
+                    .accessibilityLabel(L10n.a11yTombstonedRoom)
```
pixlwave, line 28: Do we call these "tombstoned" rooms publicly or would "Upgraded room" be more approachable here?

ElementX/Sources/Screens/HomeScreen/View/HomeScreenRecoveryKeyConfirmationBanner.swift
```diff
@@ -68,6 +68,9 @@ struct HomeScreenRecoveryKeyConfirmationBanner: View {
                         Image(systemName: "xmark")
                             .foregroundColor(.compound.iconSecondary)
                             .frame(width: 12, height: 12)
+                            // Increase hit area
+                            .padding(.leading, 5)
+                            .padding(.bottom, 2)
                     }
```
pixlwave: Should we not also then

```suggestion
                    }
                    .padding(.leading, -5)
                    .padding(.bottom, -2)
```
here so that the layout still matches the designs?

ElementX/Sources/Screens/MediaUploadPreviewScreen/View/MediaUploadPreviewScreen.swift
```diff
@@ -82,6 +82,9 @@ struct MediaUploadPreviewScreen: View {
             context.isPresentingMediaCaptionWarning = true
         } label: {
             CompoundIcon(\.infoSolid, size: .xSmall, relativeTo: .compound.bodyLG)
+                // Increase hit area
+                .padding(.leading, 5)
+                .padding(.vertical, 2)
         }
```
pixlwave: Same here

ElementX/Sources/Screens/RoomChangeRolesScreen/View/RoomChangeRolesScreenSelectedItem.swift
```diff
@@ -39,7 +44,12 @@ struct RoomChangeRolesScreenSelectedItem: View {
                             .scaledFrame(size: 20)
                             .symbolRenderingMode(.palette)
                             .foregroundStyle(Color.compound.iconOnSolidPrimary, Color.compound.iconPrimary)
+                            // Increase hit area
+                            .padding(.leading, 5)
+                            .padding(.bottom, 5)
                     }
```
pixlwave, line 47: Same here (with the right respective values)

pixlwave (approved):
LGTM, mainly just a few comments about making sure that the increased button size doesn't change the layout compared to the designs.

ElementX/Sources/Other/SwiftUI/Views/TombstonedAvatarImage.swift
Velin92: couldn't `Upgraded room` be ambiguous between the new room and the old room?

pixlwave: Probably, I would imagine this is a question for design (and maybe tombstoned room is ok, it just that it sounds technical and unobvious to me at least).

ElementX/Sources/Screens/HomeScreen/View/HomeScreenRecoveryKeyConfirmationBanner.swift
Velin92: the preview tests did not fail, also I am not sure if reducing the padding externally of leading and bottom would be the same.

pixlwave: It would probably only be noticeable when the screen/localisation would put the button close to the label.

Velin92: Okay I found out that the hit region issues are not being flagged anymore, probably they were caused by the dynamic type scaling not working prior to the fix, because now if I run the hit region tests I don't get any issue!

PR #4342, opened 2025-07-22
Make the remote settings hook usable within the app extensions.

pixlwave:

- Renames `ConfigurableSetting` to `RemotePreference`.
- Renames `ElementWellKnownHook` to `RemoteSettingsHook` and adds a few more methods to the hook.
- Adds the `RemoteSettingsHook` to the NSE.
- Adds everything necessary to use Hooks from the ShareExtension.

Please also check the diff in the enterprise submodule as part of this review (although the changes are minor).


ShareExtension/SupportingFiles/target.yml
```diff
@@ -50,9 +51,7 @@ targets:
           NSExtensionPointIdentifier: com.apple.share-services
           NSExtensionPrincipalClass: $(PRODUCT_MODULE_NAME).ShareExtensionViewController
           NSExtensionAttributes:
-            IntentsSupported: [
-              INSendMessageIntent,
-            ]
+            IntentsSupported: [INSendMessageIntent]
```
pixlwave, line 54: Unrelated but it was breaking my editor's syntax highlighting.

PR #4344, opened 2025-07-22
Adopt new thread sending api

stefanceriu:
This PR handles the breaking API changes coming from https://github.com/matrix-org/matrix-rust-sdk/pull/5427

It moves away from passing the thread root event id throughout the application and instead uses the `TimelineController` (and its inner active timeline proxy) to automatically direct messages to the right place.


ElementX/Sources/Services/Timeline/TimelineController/MockTimelineController.swift
```diff
@@ -107,6 +113,10 @@ class MockTimelineController: TimelineControllerProtocol {
                      html: String?,
                      inReplyToEventID: String?,
                      intentionalMentions: IntentionalMentions) async { }
+    
+    func sendVoiceMessage(url: URL, audioInfo: AudioInfo, waveform: [UInt16]) async -> Result<Void, TimelineProxyError> {
+        await timelineProxy?.sendVoiceMessage(url: url, audioInfo: audioInfo, waveform: waveform, requestHandle: { _ in }) ?? .success(())
+    }
```
pixlwave: I might be missing something (I'm on commit #2 right now), but shouldn't we just move the sendVoiceMessage mock implementation from the proxy mock into this one directly? Then you wouldn't need the extra dependency?

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -280,9 +281,13 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
             // Room
                 
             case (_, .presentRoom(let presentationAction), .room):
+                guard let timelineController = (context.userInfo as? EventUserInfo)?.timelineController else {
+                    fatalError()
```
pixlwave: Should we put a message in all of these for Sentry?

ElementX/Sources/Services/Timeline/TimelineController/TimelineController.swift
```diff
@@ -301,10 +301,65 @@ class TimelineController: TimelineControllerProtocol {
         }
     }
     
-    func sendVoiceMessage(url: URL, audioInfo: AudioInfo, waveform: [UInt16]) async -> Result<Void, TimelineProxyError> {
+    func sendAudio(url: URL,
+                   audioInfo: MatrixRustSDK.AudioInfo,
+                   caption: String?,
+                   requestHandle: @MainActor (any MatrixRustSDK.SendAttachmentJoinHandleProtocol) -> Void) async -> Result<Void, TimelineControllerError> {
```
pixlwave: Should we remove the `any MatrixRustSDK.` on these?

ElementX/Sources/Services/Timeline/TimelineController/MockTimelineController.swift
```diff
@@ -155,14 +155,96 @@ class MockTimelineController: TimelineControllerProtocol {
...
+                                                 requestHandle: requestHandle).mapError(TimelineControllerError.timelineProxyError)
         }
         
         return .success(())
     }
     
+    func sendFile(url: URL,
+                  fileInfo: MatrixRustSDK.FileInfo,
+                  caption: String?,
+                  requestHandle: @MainActor (any MatrixRustSDK.SendAttachmentJoinHandleProtocol) -> Void) async -> Result<Void, TimelineControllerError> {
```
pixlwave: And if so, on these too?

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -1120,13 +1135,7 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
...
     }
     
-    private func presentPollsHistory() {
-        Task {
-            await asyncPresentRoomPollsHistory()
-        }
-    }
-    
-    private func asyncPresentRoomPollsHistory() async {
+    private func presentRoomPollsHistory(animated: Bool) async {
```
pixlwave, line 1147: 🙌

ElementX/Sources/Screens/DeclineAndBlockScreen/DeclineAndBlockScreenViewModel.swift
```diff
@@ -58,7 +58,7 @@ class DeclineAndBlockScreenViewModel: DeclineAndBlockScreenViewModelType, Declin
         case .success:
             var shouldShowFailure = false
             if state.bindings.shouldReport {
-                shouldShowFailure = await clientProxy.reportRoomForIdentifier(roomID, reason: state.bindings.reportReason.isEmpty ? nil : state.bindings.reportReason).isFailure
+                shouldShowFailure = await clientProxy.reportRoomForIdentifier(roomID, reason: state.bindings.reportReason.isEmpty ? "" : state.bindings.reportReason).isFailure
```
pixlwave: I think we can remove the `isEmpty` check here now 😅

ElementX/Sources/Screens/ReportRoomScreen/ReportRoomScreenViewModel.swift
```diff
@@ -38,7 +38,7 @@ class ReportRoomScreenViewModel: ReportRoomScreenViewModelType, ReportRoomScreen
         
     private func report() async {
         showLoadingIndicator()
-        let result = await roomProxy.reportRoom(reason: state.bindings.reason.isEmpty ? nil : state.bindings.reason)
+        let result = await roomProxy.reportRoom(reason: state.bindings.reason.isEmpty ? "" : state.bindings.reason)
```
pixlwave: Ditto

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -1231,6 +1231,8 @@ private extension MediaPreviewConfig {
             .privateOnly
         case .off:
             .never
+        case .none:
+            .always
```
pixlwave, line 1235: Should this be `.always` or `.private`. Part of me wonders if we already have a default value defined somewhere else and that it should actually just be optional?

ElementX/Sources/Mocks/SDK/ClientSDKMock.swift
```diff
@@ -56,7 +56,8 @@ extension ClientSDKMock {
         getUrlUrlClosure = { url in
             guard url.contains(".well-known/element/element.json") else { throw MockError.generic }
             if let elementWellKnown = configuration.elementWellKnown {
-                return elementWellKnown
+                guard let data = elementWellKnown.data(using: .utf8) else { fatalError() }
+                return data
```
pixlwave: Might as well flatten this one an extra layer and do

```
if let elementWellKnownData = configuration.elementWellKnown?.data(using: .utf8) {
    return elementWellKnownData
} else
…
```

pixlwave (approved):
Very nice, this API is waaaaay better 👏

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
pixlwave, line 1: Oh wait, one more thought. Weren't we going to remove the stored timelineController from here too?

ElementX/Sources/Services/Client/ClientProxy.swift
pixlwave: Yeah running this branch just now I sent an image an a DM and it was hidden which seems wrong.

ElementX/Sources/Screens/DeclineAndBlockScreen/DeclineAndBlockScreenViewModel.swift
stefanceriu: Oupsie 😅

ElementX/Sources/Services/Timeline/TimelineController/MockTimelineController.swift
stefanceriu: It's mainly because the timelineProxy is more configurable as it's generated.

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
stefanceriu: We unfortunately need it when processing room re-selections and setting up presentation actions.

ElementX/Sources/Services/Client/ClientProxy.swift
stefanceriu: Talked about it offline, turns out this is correct.

PR #4345, opened 2025-07-22
Add a TracingHook.

pixlwave:
Follow on from #4342.


ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -124,6 +124,13 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
         notificationManager.delegate = self
         notificationManager.start()
         
+        appSettings.bugReportRageshakeURL.publisher
```
stefanceriu: Reading this again, it would'be been cool if we could've kept the interfaces between all the app settings the same i.e. `appSettings.$bugReportRageshakeURL`. 

We could make the remote preference a property wrapper projecting a publisher as well but not entirely sure it's worth it.

NSE/Sources/NotificationServiceExtension.swift
```diff
@@ -52,6 +55,15 @@ class NotificationServiceExtension: UNNotificationServiceExtension {
...
         }
+        
+        super.init()
+        
+        settings.bugReportRageshakeURL.publisher
+            .sink { [weak self] _ in
+                guard let self, let targetConfiguration = Self.targetConfiguration else { return }
+                appHooks.targetHook.update(targetConfiguration, with: settings)
+            }
+            .store(in: &cancellables)
```
stefanceriu: Not a big fan of having to do this dance eveywhere so here's a thought: what if we pass the `bugReportRageshakeURL` publisher into the `Target.configure` and let `Target.Configuration` handle observations and the app hook itself? 🤔

pixlwave: Updated in https://github.com/element-hq/element-x-ios/pull/4345/commits/8d4f4bd5936daecb24278ed17dd136a06052c681

ElementX/Sources/Application/AppCoordinator.swift
pixlwave: Discussed and we weren't really sure how to achieve this.

NSE/Sources/NotificationServiceExtension.swift
pixlwave: Also it's much cleaner now, thank you! 🙌

PR #4352, opened 2025-07-23
Update the default logs directory and allow collection from elsewhere.

pixlwave:
- Change our logs directory to `{GroupContainer}/Library/Logs` for a more correct file organisation (in comparison to throwing them in the root of the group container 🙈).
- Add a migration for existing logs to move them to the new location.
- Bump the version number to trigger the migration.
- Add `Tracing.logsDirectoryOverride` which to allow the logs to be collected from a different location (setting this doesn't cause the logs to be written to that location).


ElementX/Sources/Other/Extensions/URL.swift
```diff
@@ -26,7 +26,7 @@ extension URL {
 
     /// The base directory where all session data is stored.
     static var sessionsBaseDirectory: URL {
-        let applicationSupportSessionsURL = applicationSupportBaseDirectory.appendingPathComponent("Sessions", isDirectory: true)
+        let applicationSupportSessionsURL = applicationSupportBaseDirectory.appending(component: "Sessions", directoryHint: .isDirectory)
```
stefanceriu, line 36: Oh nice one 👍

ElementX/Sources/Other/Extensions/URL.swift
```diff
@@ -23,6 +23,13 @@ extension URL {
         
         return url
     }
+    
+    static var appGroupLogsDirectory: URL {
+        appGroupContainerDirectory
+            .appending(component: "Library", directoryHint: .isDirectory)
+            .appending(component: "Logs", directoryHint: .isDirectory)
```
stefanceriu, line 30: 👌

ElementX/Sources/Other/Logging/Tracing.swift
```diff
@@ -52,13 +54,17 @@ enum Tracing {
...
+    
+    /// Collect all of the logs in the given directory, sorting them chronologically.
+    private static func logFiles(in directory: URL) -> [URL] {
         var logFiles = [(url: URL, modificationDate: Date)]()
         
         let fileManager = FileManager.default
-        let enumerator = fileManager.enumerator(at: logsDirectory, includingPropertiesForKeys: [.contentModificationDateKey])
+        let enumerator = fileManager.enumerator(at: directory,
+                                                includingPropertiesForKeys: [.contentModificationDateKey],
+                                                options: .skipsSubdirectoryDescendants)
```
stefanceriu, line 71: Did subdirectories cause any issues?

stefanceriu: Ah wait, I see, they do for the `legacyLogsDirectory`. I'm slightly worried that blindly considering any file in there to be a log might cause issues. For example my simulators also have a `.com.apple.mobile_container_manager.metadata.plist` which might or might not be problematic if we move it.

Can we maybe just filter on `.log` files?

ElementX/Sources/Other/Logging/Tracing.swift
```diff
@@ -78,6 +84,30 @@ enum Tracing {
...
+                let newFileURL = logsDirectory.appending(component: oldFileURL.lastPathComponent)
+                try fileManager.moveItem(at: oldFileURL, to: newFileURL)
+                MXLog.info("Moved \(newFileURL.lastPathComponent)")
+            } catch {
+                MXLog.error("Failed to move \(oldFileURL.lastPathComponent): \(error.localizedDescription)")
+                
+                let nsError = error as NSError
+                if nsError.domain == NSCocoaErrorDomain, nsError.code == NSFileWriteFileExistsError {
+                    // By now there will already be some logs in the new directory, so there is likely to be
+                    // one log file that cannot be removed. As this is a one-off operation lets just delete it.
```
stefanceriu, line 107: So we will lose the most recent logs? Can we maybe just rename the file somehow instead?

ElementX/Sources/Other/Logging/Tracing.swift
```diff
@@ -109,9 +109,13 @@ enum Tracing {
...
     /// Delete all log files.
-    static func deleteLogFiles() {
+    static func deleteLogFiles(in directory: URL) {
         let fileManager = FileManager.default
-        for logFileURL in logFiles {
+        
+        // We don't simply delete logsDirectory as once upon a time the logs
+        // we written to the very top-level of the app group container and
+        // there's a migration in place for old users of the app.
+        for logFileURL in logFiles(in: directory) {
```
stefanceriu, line 122: Oh this one is especially problematic now in the context of the previous comment and not filtering in only .logs files.

ElementX/Sources/Other/Logging/Tracing.swift
```diff
@@ -17,10 +17,14 @@ enum Tracing {
         if ProcessInfo.isRunningIntegrationTests {
             "/Users/Shared"
         } else {
-            .appGroupLogsDirectory
+            logsDirectoryOverride ?? .appGroupLogsDirectory
```
stefanceriu, line 20: Should this be `logsDirectoryOverride ?? legacyLogsDirectory `

ElementX/Sources/Other/Logging/Tracing.swift
pixlwave: It's fine, we filter on the prefix and so we're only moving `console.` files.

ElementX/Sources/Other/Logging/Tracing.swift
pixlwave: FWIW, this method is only called in
a) Tests
b) An migration on version 1.6.7

I added the comment in case someone came and changed the code whilst that migration was still in place.

ElementX/Sources/Other/Logging/Tracing.swift
stefanceriu: Oh I see, it happens lower on line 79, the diff threw me off.

ElementX/Sources/Other/Logging/Tracing.swift
stefanceriu: Yeah, the comment is irellevant now that you pointed out filtering is in place.

ElementX/Sources/Other/Extensions/URL.swift
pixlwave: There's still a bunch of these in the app, but whilst I was here :D

stefanceriu (approved):

ElementX/Sources/Other/Logging/Tracing.swift
pixlwave: We discussed it and will leave it as is.

PR #4358, opened 2025-07-28
Multi file uploads

stefanceriu:
This series of patches adds support for selecting and sending multiple files at the same time by:
* exposing MediaPickerScreen selection modes
* passing over multiple selected URLs from the photo library and documents picker all the way down to the media upload preview screen
* allow processing multiple files at the same time on the MediaUploadingPreprocessor

Left to do:

- [ ] feature flag
- [ ] `n out of m` label on the preview screen


ElementX/Sources/Screens/MediaUploadPreviewScreen/View/MediaUploadPreviewScreen.swift
```diff
@@ -16,15 +17,23 @@ struct MediaUploadPreviewScreen: View {
...
+    @State private var currentIndex = 0
     @FocusState private var isComposerFocussed
     
     private var title: String { ProcessInfo.processInfo.isiOSAppOnMac ? context.viewState.title ?? "" : "" }
     private var colorSchemeOverride: ColorScheme { ProcessInfo.processInfo.isiOSAppOnMac ? colorScheme : .dark }
     
     var body: some View {
         mainContent
-            .id(context.viewState.url)
+            .id(context.viewState.mediaURLs)
```
pixlwave, line 28: Woah an ID can be an array of IDs, I guess that makes sense when you think about it, but definitely a TIL.

ElementX/Sources/Screens/RoomDetailsEditScreen/RoomDetailsEditScreenCoordinator.swift
```diff
@@ -63,17 +64,22 @@ final class RoomDetailsEditScreenCoordinator: CoordinatorProtocol {
...
+                                                                  userIndicatorController: parameters.userIndicatorController,
                                                                   orientationManager: parameters.orientationManager) { [weak self] action in
             guard let self else { return }
             switch action {
             case .cancel:
                 parameters.navigationStackCoordinator?.setSheetCoordinator(nil)
-            case .selectMediaAtURL(let url):
+            case .selectedMediaAtURLs(let urls):
+                guard let url = urls.first else {
+                    fatalError("Received more than one URL")
```
pixlwave: Technically this would only fatal when we received 0 URLs 😜

ElementX/Sources/Screens/Settings/UserDetailsEditScreen/UserDetailsEditScreenCoordinator.swift
```diff
@@ -52,15 +53,22 @@ final class UserDetailsEditScreenCoordinator: CoordinatorProtocol {
...
+                                                                  userIndicatorController: parameters.userIndicatorController,
+                                                                  orientationManager: parameters.orientationManager) { [weak self] action in
             guard let self else { return }
             switch action {
             case .cancel:
                 parameters.navigationStackCoordinator?.setSheetCoordinator(nil)
-            case .selectMediaAtURL(let url):
+            case .selectedMediaAtURLs(let urls):
+                guard let url = urls.first else {
+                    fatalError("Received more than one URL")
```
pixlwave: Same.

ElementX/Sources/Screens/StartChatScreen/StartChatScreenCoordinator.swift
```diff
@@ -139,15 +140,22 @@ final class StartChatScreenCoordinator: CoordinatorProtocol {
...
+                                                                  userIndicatorController: parameters.userIndicatorController,
+                                                                  orientationManager: parameters.orientationManager) { [weak self] action in
             guard let self else { return }
             switch action {
             case .cancel:
                 parameters.navigationStackCoordinator?.setSheetCoordinator(nil)
-            case .selectMediaAtURL(let url):
+            case .selectedMediaAtURLs(let urls):
+                guard let url = urls.first else {
+                    fatalError("Received more than one URL")
```
pixlwave: And here

ElementX/Sources/Screens/MediaPickerScreen/MediaPickerScreenCoordinator.swift
```diff
@@ -7,29 +7,40 @@
...
-    init(userIndicatorController: UserIndicatorControllerProtocol,
-         source: MediaPickerScreenSource,
+    init(mode: MediaPickerScreenMode,
+         appSettings: AppSettings,
+         userIndicatorController: UserIndicatorControllerProtocol,
          orientationManager: OrientationManagerProtocol,
          callback: @escaping (MediaPickerScreenCoordinatorAction) -> Void) {
         self.userIndicatorController = userIndicatorController
-        self.source = source
+        self.mode = appSettings.multipleAttachmentUploadEnabled ? mode : .init(source: mode.source, selectionType: .single)
```
pixlwave: This is really smart. It took me a while to figure out that you've hard coded `.single` and `.multiple` where they make sense and then use the feature flag here to mutate it, but I love it. Maybe it would be good to have a comment to this effect here?

ElementX/Sources/Services/Media/MediaUploadingPreprocessor.swift
```diff
@@ -94,6 +94,19 @@ struct MediaUploadingPreprocessor {
...
+    /// from images and retrieve associated media information
+    /// - Parameter urls: the file URL
+    /// - Returns: a collection of results containing specific type of `MediaInfo` depending on the file type
+    /// and its associated details or any resulting error
+    func processMedia(at urls: [URL], maxUploadSize: UInt) async -> [Result<MediaInfo, MediaUploadingPreprocessorError>] {
+        var results = [Result<MediaInfo, MediaUploadingPreprocessorError>]()
+        for url in urls {
+            await results.append(processMedia(at: url, maxUploadSize: maxUploadSize))
+        }
+        return results
```
pixlwave: Should we use a TaskGroup here so that multiple files could be handled in parallel?

That would also give us the ability to return a `Result<[MediaInfo], MediaUploadingPreprocessorError>` too which is probably easier to reason about the failure (especially when we get to implementing galleries) and maybe even cancel any remaining conversion tasks when one fails.

ShareExtension/Sources/ShareExtensionViewController.swift
```diff
@@ -63,23 +63,32 @@ class ShareExtensionViewController: UIViewController {
...
-        } else if let url = await itemProvider.loadTransferable(type: URL.self) {
-            return .text(roomID: roomID, text: url.absoluteString)
-        } else if let string = await itemProvider.loadString() {
-            return .text(roomID: roomID, text: string)
-        } else {
-            MXLog.error("Failed loading NSItemProvider data: \(itemProvider)")
-            return nil
+        var mediaFiles = [ShareExtensionMediaFile]()
+        for itemProvider in itemProviders {
+            if let fileURL = await itemProvider.storeData(withinAppGroupContainer: true) {
```
pixlwave, line 76: Kind of wondering the same about a TaskGroup here.

ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
```diff
@@ -277,21 +277,27 @@ class TimelineInteractionHandler {
...
                 self.userIndicatorController.retractIndicatorWithId(loadingIndicatorIdentifier)
             }
             
-            guard let fileURL = await provider.storeData() else {
-                MXLog.error("Failed storing NSItemProvider data \(provider)")
-                self.actionsSubject.send(.displayErrorToast(L10n.screenRoomErrorFailedProcessingMedia))
-                return
+            var mediaURLs = [URL]()
+            for provider in providers {
+                if let fileURL = await provider.storeData() {
```
pixlwave, line 290: And here

ElementX/Sources/Screens/MediaPickerScreen/PhotoLibraryPicker.swift
```diff
@@ -56,38 +66,65 @@ struct PhotoLibraryPicker: UIViewControllerRepresentable {
...
-                    Task { @MainActor in
-                        self?.photoLibraryPicker.callback(.error(.failedLoadingFileRepresentation(error)))
+            Task {
+                var selectedURLs = [URL]()
+                for result in results {
+                    if let url = await processResult(result) {
+                        selectedURLs.append(url)
                     }
-                    return
                 }
```
pixlwave: And here

ElementX/Sources/Screens/MediaUploadPreviewScreen/MediaUploadPreviewScreenViewModel.swift
```diff
@@ -62,26 +62,39 @@ class MediaUploadPreviewScreenViewModel: MediaUploadPreviewScreenViewModelType,
...
                 
-                switch await processingTask.value {
-                case .success(let mediaInfo):
-                    switch await sendAttachment(mediaInfo: mediaInfo,
-                                                caption: caption) {
-                    case .success:
-                        actionsSubject.send(.dismiss)
+                var shouldDismissOnCompletion = true
+                for result in await processingTask.value {
+                    var shouldStopEnumerating = false
```
pixlwave: And then if we did that, we could switch on the result and either fail or send all the files so wouldn't need these 2 `should…` vars.

pixlwave (approved):
🤩

Sorry it looks like a lot of comments but I think more than half are "and here too", so all part of the same thing.

ElementX/Sources/Screens/MediaUploadPreviewScreen/View/MediaUploadPreviewScreen.swift
stefanceriu: Yeah, it's just a hashable at the end of it.

beposec:
Great work @stefanceriu! What is the reason for limiting upload to 5 files? This seems to be a very low limit.

From a user perspective I would not like to have the app dictate me how many photos I am allowed to share.

ElementX/Sources/Screens/RoomDetailsEditScreen/RoomDetailsEditScreenCoordinator.swift
stefanceriu: Not entirely sure what happened here tbh 🤔

stefanceriu:
> Great work @stefanceriu! What is the reason for limiting upload to 5 files? This seems to be a very low limit.

Thanks! The fear was that we don't support [inline media galleries](https://github.com/Johennes/matrix-spec-proposals/blob/johannes/msgtype-galleries/proposals/4274-inline-media-galleries.md) at the moment and we don't want one's timeline to be spammed. Honestly though, I can be convinced otherwise.

beposec:
> we don't want one's timeline to be spammed.

That's s fair reason. I understand that a limit is necessary and that it's really hard to find a good number for the limit. 

I was just thinking about some of my personal use cases where I would like to send more than 5 items at once. For my very personal opinion I would like to have the limit something like 10-15.

ShareExtension/Sources/ShareExtensionViewController.swift
stefanceriu: This one is tricky because it's returning different types depending on what those `itemProvider` actions return.

ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
stefanceriu: Here we want to treat each error separately and only inform instead of failing the group and showing a single error.

ElementX/Sources/Services/Media/MediaUploadingPreprocessor.swift
stefanceriu: That does indeed look way way nicer! 🙏

ElementX/Sources/Screens/MediaUploadPreviewScreen/MediaUploadPreviewScreenViewModel.swift
stefanceriu: Yep yep, this is definitely cleaner 👍

stefanceriu:
> I would like to have the limit something like 10-15.

I think we can meet in the middle and go with 10 for now. These flows will stay under a feature flag for now so we can tweak the values as we go along.

beposec:
> I think we can meet in the middle and go with 10 for now.

Perfect! I am so happy to see this feature! Thank you for your work ☺️

stefanceriu:
> Perfect! I am so happy to see this feature! Thank you for your work ☺️

🙏  Same here, it's been too long in the making!

PR #4366, opened 2025-07-29
SDK Update + PowerLevels API update

Velin92:
This PR bumps the SDK and updates the way PowerLevels are now handled, being an enum instead of a simple integer.
The aim of the PR for now is to build with the new changes and keep compatibility with previous room versions. So is not aimed to fully support v12 yet.

Fixes https://github.com/element-hq/element-x-ios/issues/4343


ElementX/Sources/Screens/ManageRoomMemberSheet/View/ManageRoomMemberSheetView.swift
```diff
@@ -6,6 +6,7 @@
 //
 
 import Compound
+import MatrixRustSDK
```
pixlwave: It seems like a code smell when we end up needing to import Rust directly at the view level. Should we maybe move this import down to above the preview so it's at least clear it isn't a dependency on the UI layer?

ElementX/Sources/Services/Room/JoinedRoomProxy.swift
```diff
@@ -815,3 +815,14 @@ private final class RoomKnockRequestsListener: KnockRequestsListener {
...
+extension PowerLevel {
+    var equivalentValue: Int64 {
+        switch self {
+        case .value(let value):
+            return value
+        case .infinite:
+            return 150
+        }
+    }
+}
```
pixlwave: I think this is used in multiple places, so we should either have a `PowerLevel.swift` extension, or personally I think it might be simpler to have a Swift `PowerLevel` and map it to/from Rust like we're already doing for roles.

Velin92:
> LGTM, although I see a few `// TODO: make a creator role` comments, did you want to handle that in this PR or is that for a follow-up?

follow up

ElementX/Sources/Services/Room/JoinedRoomProxy.swift
Velin92: like a PowerLevelProxy?

pixlwave: I don't think it particularly needs to be a Proxy, just our own type with an `init(rustValue:)` and `var rustValue` to manually map to/from. See e.g. `RoomMemberDetails.Role`

ElementX/Sources/Services/Room/RoomInfoProxyProtocol.swift
```diff
@@ -26,7 +26,7 @@ protocol BaseRoomInfoProxyProtocol {
 // sourcery: AutoMockable
 protocol RoomInfoProxyProtocol: BaseRoomInfoProxyProtocol {
     var id: String { get }
-    var creator: String? { get }
+    var creators: [String]? { get }
```
pixlwave: How come this is an optional array? Does `nil` vs `[]` represent different things? If so, maybe we should add a doc comment here.

ElementX/Sources/Services/Room/RoomPermissions.swift
```diff
@@ -100,7 +105,20 @@ extension RoomMemberDetails.Role {
         }
     }
     
-    var rustPowerLevel: Int64 {
-        suggestedPowerLevelForRole(role: rustRole)
+    /// To be used when setting the power level of a user to get the suggested equivalent power level value for that specific role
+    /// NOTE: Do not use for comparison, use the true power level instead.
```
pixlwave, line 109: 👌

ElementX/Sources/Services/Room/RoomPermissions.swift
```diff
@@ -100,7 +105,20 @@ extension RoomMemberDetails.Role {
...
+    var powerLevelValue: Int64 {
+        do {
+            switch try suggestedPowerLevelForRole(role: rustRole) {
+            case .infinite:
+                // Would be better if the SDK would return this, maybe a `suggestedPowerLevelValueForRole` function would solve the problem
+                return 150
```
pixlwave, line 115: Do we use this? I'm wondering if making this return an `Int64?` would force us to handle the infinite level on a case-by-case basis 🤔

ElementX/Sources/Services/Room/PowerLevelProxy.swift
pixlwave, line 1: I slightly question the benefit of this proxying the rust value vs a simple enum that matches the Rust one, but not super important.

ElementX/Sources/Services/Room/RoomPermissions.swift
Velin92: as of right now, no because there is no owner role, but when the creator role will be introduced and will be set I guess I can make it return 150 directly from the role.

PR #4368, opened 2025-07-30
Add a TabView to the root of the UserSession flow and refactor out a new Chats flow.

pixlwave:
With this PR, the `TabView` becomes part of the hierarchy, but is hard-coded as hidden until we're ready to show it (it will be reactive based on whether or not the user is a member of any spaces).

The commits basically explain themselves but just as a summary:
- Add a NavigationTabCoordinator to manage a set of tabs.
- Rename UserSessionFlowCoordinator to ChatsFlowCoordinator
- Make a basic UserSessionFlowCoordinator that embeds the ChatsFlowCoordinator into a tab.
- Move onboarding, verification and logout handling from the Chats flow into the UserSession flow.

Neither the implementation of the `NavigationTabCoordinator` nor the `UserSessionFlowCoordinator` are complete (the former needs badges, tab visibility and programmatic tab switching, the latter needs a state machine and improved route handling), but this PR was large enough that it seemed worthwhile opening now.

**Note:** This PR sure ain't a large as +13,901, the diff is just being dumb. Definitely review the commits individually I guess 😅

| iPhone | iPad |
| - | - |
| Simulator Screenshot - iPhone 16 Pro - 2025-07-30 at 16 02 04 | Simulator Screenshot - iPad Pro 11-inch (M4) - 2025-07-30 at 16 00 02 |

### DIFF
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -46,7 +46,7 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
     private let appLockFlowCoordinator: AppLockFlowCoordinator
     // periphery:ignore - used to avoid deallocation
     private var appLockSetupFlowCoordinator: AppLockSetupFlowCoordinator?
-    private var userSessionFlowCoordinator: ChatsFlowCoordinator?
+    private var userSessionFlowCoordinator: UserSessionFlowCoordinator?
     private var softLogoutCoordinator: SoftLogoutScreenCoordinator?
     private var appDelegateObserver: AnyCancellable?
     private var userSessionObserver: AnyCancellable?
@@ -640,18 +640,18 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
             fatalError("User session not setup")
         }
         
-        let userSessionFlowCoordinator = ChatsFlowCoordinator(userSession: userSession,
-                                                              navigationRootCoordinator: navigationRootCoordinator,
-                                                              appLockService: appLockFlowCoordinator.appLockService,
-                                                              bugReportService: ServiceLocator.shared.bugReportService,
-                                                              elementCallService: elementCallService,
-                                                              timelineControllerFactory: TimelineControllerFactory(),
-                                                              appMediator: appMediator,
-                                                              appSettings: appSettings,
-                                                              appHooks: appHooks,
-                                                              analytics: ServiceLocator.shared.analytics,
-                                                              notificationManager: notificationManager,
-                                                              isNewLogin: isNewLogin)
+        let userSessionFlowCoordinator = UserSessionFlowCoordinator(userSession: userSession,
+                                                                    navigationRootCoordinator: navigationRootCoordinator,
+                                                                    appLockService: appLockFlowCoordinator.appLockService,
+                                                                    bugReportService: ServiceLocator.shared.bugReportService,
+                                                                    elementCallService: elementCallService,
+                                                                    timelineControllerFactory: TimelineControllerFactory(),
+                                                                    appMediator: appMediator,
+                                                                    appSettings: appSettings,
+                                                                    appHooks: appHooks,
+                                                                    analytics: ServiceLocator.shared.analytics,
+                                                                    notificationManager: notificationManager,
+                                                                    isNewLogin: isNewLogin)
         
         userSessionFlowCoordinator.actionsPublisher
             .sink { [weak self] action in
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -46,7 +46,7 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
     private let appLockFlowCoordinator: AppLockFlowCoordinator
     // periphery:ignore - used to avoid deallocation
     private var appLockSetupFlowCoordinator: AppLockSetupFlowCoordinator?
-    private var userSessionFlowCoordinator: UserSessionFlowCoordinator?
+    private var userSessionFlowCoordinator: ChatsFlowCoordinator?
     private var softLogoutCoordinator: SoftLogoutScreenCoordinator?
     private var appDelegateObserver: AnyCancellable?
     private var userSessionObserver: AnyCancellable?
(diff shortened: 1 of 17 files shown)


ElementX/Sources/Application/Navigation/NavigationTabCoordinator.swift
```diff
@@ -0,0 +1,103 @@
...
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Compound
+import SwiftUI
+
+/// Class responsible for displaying an arbitrary number of coordinators within the tab bar.
+@Observable class NavigationTabCoordinator: CoordinatorProtocol, CustomStringConvertible {
```
stefanceriu, line 12: Might be worth adding some top level unit tests here similar to the other Navigation*Coordinator tests

ElementX/Sources/Application/Navigation/NavigationTabCoordinator.swift
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+    
+    /// The current set of coordinators displayed by the tabs.
+    var tabCoordinators: [any CoordinatorProtocol] {
+        tabModules.compactMap(\.module.coordinator)
+    }
+    
+    /// Updates the displayed tabs with the provided array.
+    func setTabs(_ tabs: [Tab], animated: Bool = true) {
```
stefanceriu, line 55: 👌

ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -46,7 +46,7 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
     private let appLockFlowCoordinator: AppLockFlowCoordinator
     // periphery:ignore - used to avoid deallocation
     private var appLockSetupFlowCoordinator: AppLockSetupFlowCoordinator?
-    private var userSessionFlowCoordinator: UserSessionFlowCoordinator?
+    private var userSessionFlowCoordinator: ChatsFlowCoordinator?
```
stefanceriu: Noooooo 😱 Just kidding I know it made a comback but honestly, I don't think ChatsFlowCoordinator is a good enough name for something that controls this much.

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -0,0 +1,106 @@
...
+    
+    func handleAppRoute(_ appRoute: AppRoute, animated: Bool) {
+        chatsFlowCoordinator.handleAppRoute(appRoute, animated: animated)
+    }
+    
+    func clearRoute(animated: Bool) {
+        chatsFlowCoordinator.clearRoute(animated: animated)
+    }
+    
+    #warning("Should this be a publisher instead??")
```
stefanceriu: Do we need it to be a publisher? If not then no.

project.yml
```diff
@@ -72,7 +72,7 @@ packages:
     # path: ../matrix-rust-sdk
   Compound:
     url: https://github.com/element-hq/compound-ios
-    revision: 089904dc1aeb3bacce3b7bb757b7637274e46008
+    revision: afc59afb1b1e4f4960e2f2a15e52d4e2e33fc889
```
stefanceriu, line 75: What's this about? 🙃

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -63,44 +69,231 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
...
+        // chats flow coordinator.
+        #warning("This should switch tabs to make sure the route is visible.")
+        chatsFlowCoordinator.handleAppRoute(appRoute, animated: animated)
+    }
+    
+    func clearRoute(animated: Bool) {
+        chatsFlowCoordinator.clearRoute(animated: animated)
+    }
+    
+    #warning("This should use a publisher, combining it with the active tab.")
```
stefanceriu, line 114: Why does it need the active tab?

ElementX/Sources/FlowCoordinators/ChatsFlowCoordinator.swift
```diff
@@ -37,8 +38,6 @@ class ChatsFlowCoordinator: FlowCoordinatorProtocol {
     
     private let settingsFlowCoordinator: SettingsFlowCoordinator
     
-    private let onboardingFlowCoordinator: OnboardingFlowCoordinator
```
stefanceriu: I wonder if we should move the element call presentation support one level higher too 🤔

ElementX/Sources/Application/Navigation/NavigationTabCoordinator.swift
pixlwave: Ah yeah, I was going to start that yesterday but forgot 👍

ElementX/Sources/Application/AppCoordinator.swift
pixlwave: The tab is called Chats, I couldn't think how else you would name it. Part of me thinks we should be pulling the settings flow up 1 level too and just have the button send an action. It feels like eventually it should be purely about the room list.

project.yml
pixlwave: It pulls in an update for compound icons where `workspace` was renamed to `space`. Seemed worthwhile to do now.

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
pixlwave: Outdated, replying on the most recent version.

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
pixlwave: Because if you're looking at the spaces tab, you aren't looking at the room shown by the chatsFlowCoordinator. And this bool is used to determine whether or not we show in-app notifications.

ElementX/Sources/Application/AppCoordinator.swift
stefanceriu: > It feels like eventually it should be purely about the room list.

Hmm yeah, that might be a very sensible way to go about it.

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
stefanceriu: Oh I see. I would just expect it to return false then ¯\_(ツ)_/¯

ElementX/Sources/FlowCoordinators/ChatsFlowCoordinator.swift
pixlwave: Yeah probably. I guess we'll get to that when we push a room in the spaces tab, at which point we'll need a shared presentation layer.

stefanceriu (approved):
Exciting! 👏

PR #4369, opened 2025-07-30
Restore permissions to creator and display them as owners in the list

### LINKED ISSUE #4338: [Task] Room Creators have same power as Admins

### LINKED ISSUE #4348: [Task] Display owners of v12 rooms in the member list

Velin92:
fixes #4338 
fixes #4348 

This PR adds 2 new roles:
- Creator, which is one of the original creators of a room, the app will still report them as a Owner, but they can NEVER be demoted, and have infinite power level.
- Owner, they are administrators with PL 150, so they can upgrade a room, for the rest they are like admins, and can be demoted or promoted by a Creator


ElementX/Sources/Mocks/RoomMemberProxyMock.swift
```diff
@@ -120,6 +120,14 @@ extension RoomMemberProxyMock {
                                         role: .administrator))
     }
     
+    static var mockCreator: RoomMemberProxyMock {
+        RoomMemberProxyMock(with: .init(userID: "@creator:matrix.org",
+                                        displayName: "God",
```
pixlwave, line 125: 🙇‍♂️

ElementX/Sources/Services/Room/RoomPowerLevelsProxy.swift
```diff
@@ -28,7 +30,15 @@ struct RoomPowerLevelsProxy: RoomPowerLevelsProxyProtocol {
     
     func suggestedRole(forUser userID: String) -> RoomMemberRole {
         do {
+            guard !creators.contains(userID) else {
+                return .creator
+            }
```
Velin92: the only concern I have for this is that, we should probably check first if we are on room v12 before doing so maybe?

ElementX/Sources/Services/Room/RoomMember/RoomMemberDetails.swift
```diff
@@ -19,7 +19,7 @@ struct RoomMemberDetails: Identifiable, Hashable {
     var isBanned: Bool
     var isActive: Bool
         
-    enum Role { case administrator, moderator, user }
+    enum Role { case creator, administrator, moderator, user }
```
pixlwave: Are we eventually going to add `owner` in here too?

pixlwave: Ah lol I added this before we talked about it 🤦‍♂️

PR #4370, opened 2025-07-30
Delegate the sending of call notifications to Element Call

robintown:
As of Element Call version 0.14.0, the widget is now capable of sending call notifications itself if we just request this with the `sendNotificationType` URL parameter. This will more reliably trigger a notification on the receiver side than what we were doing before with `Room::send_call_notification_if_needed`, since it always waits until the `m.call.member` state event has been sent before sending the notification event.

~~This change requires matrix-rust-sdk to be updated to https://github.com/matrix-org/matrix-rust-sdk/commit/feadfde1b54f2caf9d1dbbbd0208781b657b0c3f or later. I don't have permission to write to https://github.com/element-hq/matrix-rust-components-swift, so I'll need your help to get this done.~~ Done, thanks!

This is the iOS implementation of https://github.com/element-hq/voip-internal/issues/381
See also https://github.com/element-hq/customer-success/issues/476

Fixes https://github.com/element-hq/element-x-ios/issues/4356

### Pull Request Checklist

- [ ] Voiceover enabled.


robintown:
So… I've updated the Swift sources here based on what looks plausible, but as I don't have access to a Mac or an iPhone, I need help finishing the job. @fkwp Maybe you can start out here by trying to build it with a bleeding edge local copy of matrix-rust-sdk? And then testing that notifications are still working.

ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
```diff
@@ -75,30 +75,32 @@ class ElementCallWidgetDriver: WidgetCapabilitiesProvider, ElementCallWidgetDriv
...
+                                                                                hideScreensharing: false,
+                                                                                posthogUserId: nil,
+                                                                                posthogApiHost: analyticsConfiguration?.posthogAPIHost,
+                                                                                posthogApiKey: analyticsConfiguration?.posthogAPIKey,
+                                                                                rageshakeSubmitUrl: rageshakeURL,
+                                                                                sentryDsn: analyticsConfiguration?.sentryDSN,
+                                                                                sentryEnvironment: nil,
+                                                                                controlledMediaDevices: !ProcessInfo.processInfo.isiOSAppOnMac),
+                                                                                sendNotificationType: direct ? .ring : .notification)
```
fkwp:
```suggestion
                                                                                controlledMediaDevices: !ProcessInfo.processInfo.isiOSAppOnMac,
                                                                                sendNotificationType: direct ? .ring : .notification))
```

ElementX/Sources/Screens/CallScreen/CallScreenViewModel.swift
fkwp, line 1:
```
@@ -50,7 +50,7 @@ class CallScreenViewModel: CallScreenViewModelType, CallScreenViewModelProtocol
         switch configuration.kind {
         case .genericCallLink(let url):
             widgetDriver = GenericCallLinkWidgetDriver(url: url)
-        case .roomCall(let roomProxy, let clientProxy, _, _, _, _, _):
+        case .roomCall(let roomProxy, let clientProxy, _, _, _, _):
             guard let deviceID = clientProxy.deviceID else { fatalError("Missing device ID for the call.") }
             widgetDriver = roomProxy.elementCallWidgetDriver(deviceID: deviceID)
         }
```

ElementX/Sources/Services/ElementCall/ElementCallConfiguration.swift
fkwp, line 1:
```
@@ -72,7 +72,7 @@ struct ElementCallConfiguration {
         switch kind {
         case .genericCallLink(let url):
             url.absoluteString
-        case .roomCall(let roomProxy, _, _, _, _, _, _):
+        case .roomCall(let roomProxy, _, _, _, _, _):
             roomProxy.id
         }
     }
```

ElementX/Sources/FlowCoordinators/ChatsFlowCoordinator.swift
```diff
@@ -531,8 +531,7 @@ class ChatsFlowCoordinator: FlowCoordinatorProtocol {
             
             switch action {
             case .presentCallScreen(let roomProxy):
-                // Here we assume that the app is running and the call state is already up to date
```
pixlwave, line 534: I might be missing something but I think this comment might still be a valid assumption to note.

ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
```diff
@@ -75,30 +75,32 @@ class ElementCallWidgetDriver: WidgetCapabilitiesProvider, ElementCallWidgetDriv
             return .failure(.roomInvalid)
         }
         
-        let useEncryption = await (try? room.latestEncryptionState() == .encrypted) ?? false
+        async let useEncryption = (try? room.latestEncryptionState() == .encrypted) ?? false
+        async let direct = room.isDirect()
```
pixlwave: Super minor nit, but generally we use the `is` prefix on bools, so `isDirect`, `isDirectRoom` or similar would be more idiomatic here.

ElementX/Sources/FlowCoordinators/ChatsFlowCoordinator.swift
robintown: My understanding is that this was just about the value of `roomProxy.infoPublisher.value.hasRoomCall`, which we expected to be up to date, as this assumption wasn't highlighted on any of the other call sites

pixlwave: Ah right, yes sorry I see what you mean now. Ok makes sense.

PR #4372, opened 2025-07-31
Last Owner should edit admins, and not leave when is last

### LINKED ISSUE #4349: [Task] Edit Room Admins

### LINKED ISSUE #4350: [Task] Assign owner on leave

Velin92:
This ensure full support for v12 rooms creators and owners role.

fixes #4350 
fixes #4349


ElementX/Sources/Services/Room/RoomMember/RoomMemberProxyProtocol.swift
pixlwave, line 1: Seeing this twice (along with `isAdminOrHigher`), it is definitely time to refactor `RoomMemberDetails.Role` into a `ElementX.RoomMemberRole` and do the mapping here in the proxy so we never use the Rust one in the app. Then we can use a single implementation for all of these checks.

ElementX/Sources/Screens/RoomDetailsScreen/RoomDetailsScreenViewModel.swift
```diff
@@ -177,6 +171,34 @@ class RoomDetailsScreenViewModel: RoomDetailsScreenViewModelType, RoomDetailsScr
     
     // MARK: - Private
     
+    private func processTapToLeave() {
```
pixlwave, line 174: Also, seeing this duplicate code I *think* we could move it to `RoomProxyProtocol` too? Some kind of `func leaveConfirmation()` that returns an enum that we can switch in both locations?

Maybe it would need a parameter deciding whether or not it should call `updateMembers`, but otherwise it looks the same to me.

ElementX/Sources/Screens/RoomChangeRolesScreen/RoomChangeRolesScreenModels.swift
```diff
@@ -107,3 +121,49 @@ enum RoomChangeRolesScreenViewAction {
...
+    
+    init(from role: RoomRolesAndPermissionsScreenRole) {
+        switch role {
+        case .administrators(let ownUserRole):
+            self = .administrator(ownUserRole: ownUserRole)
+        case .moderators:
+            self = .moderator
+        }
+    }
+}
```
pixlwave: This whole thing feels like it overcomplicates the screen to me. We could keep the `mode` as the role, add a new parameter of ownUserRole (or PL) and have a simple method on the models `canDemoteMember(_:)`. I say this because `promotingRole` and `maxDemotableRole` aren't very easy to understand what they do/are for.

ElementX/Sources/Screens/RoomChangeRolesScreen/RoomChangeRolesScreenModels.swift
```diff
@@ -14,15 +14,17 @@ enum RoomChangeRolesScreenViewModelAction {
...
+    var administrators: [RoomMemberDetails] = []
     /// All of the room's members who are currently moderators.
-    var moderators: [RoomMemberDetails]
+    var moderators: [RoomMemberDetails] = []
     /// All of the room's members who are currently neither an admin or moderator.
-    var users: [RoomMemberDetails]
+    var users: [RoomMemberDetails] = []
     
-    var bindings: RoomChangeRolesScreenViewStateBindings
+    var bindings = RoomChangeRolesScreenViewStateBindings()
```
pixlwave, line 29: 👌

ElementX/Sources/Screens/RoomChangeRolesScreen/RoomChangeRolesScreenViewModel.swift
```diff
@@ -64,7 +58,7 @@ class RoomChangeRolesScreenViewModel: RoomChangeRolesScreenViewModelType, RoomCh
         case .demoteMember(let member):
             demoteMember(member)
         case .save:
-            if state.mode.isAdminOrHigher, !state.membersToPromote.isEmpty {
+            if state.mode.promotingRole.isAdminOrHigher, !state.membersToPromote.isEmpty {
                 showPromotionWarning()
```
pixlwave: We probably need to re-review this now. Because if you're a creator we never need to show the warning and if you're an Owner, we don't need to show it when promoting to Admin.

ElementX/Sources/Services/Room/RoomMember/RoomMemberDetails.swift
```diff
@@ -19,17 +19,17 @@ struct RoomMemberDetails: Identifiable, Hashable {
...
         /// Default role PL 0...49
         case user
+        /// Able to perform room moderation actions PL 50...99
+        case moderator
+        /// Able to edit room settings and perform any action aside from room upgrading PL 100...149
+        case administrator
+        /// Same power of an admin, but they can also upgrade the room, PL 150 onwards
+        case owner
+        /// Creator of the room, PL infinite
+        case creator
```
pixlwave: I thought we decided it was bad to compare Roles and we should only ever compare PowerLevels?

ElementX/Sources/Screens/RoomChangeRolesScreen/View/RoomChangeRolesScreenSelectedItem.swift
```diff
@@ -10,9 +10,22 @@ import SwiftUI
...
+                    dismissAction()
+                }
+        } else {
+            mainContent
+        }
+    }
+    
+    // MARK: - Private
+    
+    private var mainContent: some View {
```
pixlwave: Instead of doing this we could use the `.accessibilityActions` modifier in the original place:

```
        .accessibilityActions {
            if let dismissAction {
                Button(L10n.actionDismiss) { dismissAction() }
            }
        }
```

pixlwave (changes requested):
I haven't looked at the whole thing yet, just the first 2 commits :)

ElementX/Sources/Screens/RoomChangeRolesScreen/RoomChangeRolesScreenViewModel.swift
pixlwave: The obvious check here is `state.mode.powerLevel >= ownUser.powerLevel`

ElementX/Sources/FlowCoordinators/RoomRolesAndPermissionsFlowCoordinator.swift
pixlwave, line 1: At this stage it is probably neater to revert this file and update it with the new role type :)

ElementX/Sources/Mocks/RoomMemberProxyMock.swift
```diff
@@ -181,4 +182,32 @@ extension Array where Element == RoomMemberProxyMock {
...
+        .mockBob,
+        .mockCharlie,
+        .mockDan,
+        .mockInvited,
+        .mockIgnored,
+        .mockAdmin,
+        .mockModerator,
+        .mockOwner,
+        .mockCreator
+    ]
```
pixlwave: Could we maybe put this one after `AsAdmin` and document why we have 2 please?

ElementX/Sources/Screens/RoomChangeRolesScreen/RoomChangeRolesScreenViewModel.swift
```diff
@@ -20,22 +20,17 @@ class RoomChangeRolesScreenViewModel: RoomChangeRolesScreenViewModelType, RoomCh
...
     }
 
-    init(mode: RoomMemberDetails.Role,
+    init(mode: RoomRole,
          roomProxy: JoinedRoomProxyProtocol,
          mediaProvider: MediaProviderProtocol,
          userIndicatorController: UserIndicatorControllerProtocol,
          analytics: AnalyticsService) {
-        guard mode != .user else { fatalError("Invalid screen configuration: \(mode)") }
-
```
pixlwave, line 29: This 1 could probably be put back now if you wanted, along with `|| .creator` as well perhaps?

ElementX/Sources/Screens/RoomChangeRolesScreen/RoomChangeRolesScreenViewModel.swift
```diff
@@ -64,11 +59,16 @@ class RoomChangeRolesScreenViewModel: RoomChangeRolesScreenViewModelType, RoomCh
...
             demoteMember(member)
         case .save:
-            if state.mode.isAdminOrHigher, !state.membersToPromote.isEmpty {
-                showPromotionWarning()
-            } else {
-                Task { await save() }
+            if !state.membersToPromote.isEmpty {
+                if state.mode == .administrator {
+                    showPromotionWarning()
+                    return
```
pixlwave: Hold on, I don't think this covers it yet. We don't need to show the warning if `state.mode.powerLevel < ownUserPowerLevel`. As a creator you can promote someone to admin and demote them back to member again no problem.

ElementX/Sources/Screens/RoomChangeRolesScreen/RoomChangeRolesScreenViewModel.swift
```diff
@@ -77,17 +77,23 @@ class RoomChangeRolesScreenViewModel: RoomChangeRolesScreenViewModelType, RoomCh
...
     private func updateMembers(_ members: [RoomMemberProxyProtocol]) {
+        var owners = [RoomMemberDetails]()
         var administrators = [RoomMemberDetails]()
         var moderators = [RoomMemberDetails]()
         var users = [RoomMemberDetails]()
         
         for member in members.sorted() {
+            if member.userID == roomProxy.ownUserID {
+                state.ownRole = member.role
+            }
```
pixlwave, line 90: 👌

ElementX/Sources/Services/Room/RoomRole.swift
```diff
@@ -0,0 +1,140 @@
...
+extension RoomMemberRole {
+    var isAdminOrHigher: Bool {
+        switch self {
+        case .administrator, .creator:
+            return true
+        case .moderator, .user:
+            return false
+        }
+    }
+}
```
pixlwave: Am I right in thinking we could remove this extension now? 🤔

ElementX/Sources/Screens/RoomChangeRolesScreen/View/RoomChangeRolesScreenSelectedItem.swift
pixlwave: Did you try this out? It would be much neater if it works.

ElementX/Sources/Screens/RoomChangeRolesScreen/RoomChangeRolesScreenModels.swift
```diff
@@ -78,7 +90,31 @@ struct RoomChangeRolesScreenViewState: BindableState {
...
+                return .owner
+            default:
+                return .moderator
+            }
+        case .moderator:
+            return .moderator
+        default:
+            return .user
+        }
     }
```
pixlwave, line 123: I'm still convinced we can combine these into a single function `canDemoteMember(_:) -> Bool` rather than needing to understand what `maxDemotableRole` means.

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -196,7 +198,43 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
...
+                                                                    mediaProvider: userSession.mediaProvider,
+                                                                    userIndicatorController: userIndicatorController,
+                                                                    analytics: analytics)
+        let stackCoordinator = NavigationStackCoordinator()
+        let coordinator = RoomChangeRolesScreenCoordinator(parameters: parameters)
+        coordinator.actionsPublisher.sink { [weak self] action in
+            guard let self else { return }
+            switch action {
+            case .complete:
+                // When discarding changes is finalised, either use an event or remove this action.
```
pixlwave: Is there something still to do here?

ElementX/Sources/FlowCoordinators/RoomRolesAndPermissionsFlowCoordinator.swift
Velin92: I think this one is fine, you are still only setting admins when you are a owner, the only change is that I made so that we check the current user role when setting the content of the row

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
Velin92: I think that's just some copied and pasted code

ElementX/Sources/Screens/RoomChangeRolesScreen/RoomChangeRolesScreenModels.swift
Velin92: I could make such a function, but would behave exactly like making the maxdemotable role and comparing the user role with it... so I would just reuse this code

pixlwave: Yeah I realise that internally to the function, but I was hoping to make it clearer at the point of use what's going on :)

ElementX/Sources/FlowCoordinators/RoomRolesAndPermissionsFlowCoordinator.swift
pixlwave: I just meant to keep the the logic out of the state machine handler, and keep it in the present function :)

PR #4380, opened 2025-08-04
Initial setup for the SpaceListScreen.

pixlwave:
Very incomplete, but making a PR now because there's lots more flow coordinator work that needs to happen around this so want to keep the reviews manageable.

https://github.com/user-attachments/assets/c08b51dd-9d51-4ab3-a73b-e64694d5b2cf


ElementX/Sources/Application/Navigation/NavigationTabCoordinator.swift
```diff
@@ -223,7 +223,7 @@ private struct NavigationTabCoordinatorView<Tag: Hashable>: View {
                     .toolbar(module.details.barVisibility, for: .tabBar)
             }
         }
-        .introspect(.tabView, on: .iOS(.v18), customize: configureAppearance)
+        .introspect(.tabView, on: .supportedVersions, customize: configureAppearance)
```
stefanceriu: 👌

stefanceriu (approved):
Love it! ❤️

PR #4390, opened 2025-08-06
Move the Settings flow from the Chats flow up one level to the UserSession flow.

pixlwave:
Also adds a basic state machine (not tracking onboarding/verification yet). Builds on top of #4383.

This allows the Settings button work on the Spaces tab:

https://github.com/user-attachments/assets/c624888a-e7e7-432c-86cd-0ec4cf87790d

### DIFF
--- a/UnitTests/Sources/ChatsFlowCoordinatorTests.swift
+++ b/UnitTests/Sources/ChatsFlowCoordinatorTests.swift
@@ -17,6 +17,7 @@ class ChatsFlowCoordinatorTests: XCTestCase {
     var chatsFlowCoordinator: ChatsFlowCoordinator!
     var splitCoordinator: NavigationSplitCoordinator!
     var notificationManager: NotificationManagerMock!
+    let stateMachineFactory = PublishedStateMachineFactory()
     
     var cancellables = Set<AnyCancellable>()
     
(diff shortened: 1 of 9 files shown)


ElementX/Sources/FlowCoordinators/SettingsFlowCoordinator.swift
```diff
@@ -62,15 +60,7 @@ class SettingsFlowCoordinator: FlowCoordinatorProtocol {
         case .settings:
             presentSettingsScreen(animated: animated)
         case .chatBackupSettings:
-            if navigationStackCoordinator == nil {
-                presentSettingsScreen(animated: animated)
-            }
```
pixlwave, line 67: This check is no longer needed as the flow is now always being presented with the settings screen route.

ElementX/Sources/FlowCoordinators/SettingsFlowCoordinator.swift
```diff
@@ -136,15 +119,6 @@ class SettingsFlowCoordinator: FlowCoordinatorProtocol {
...
         navigationStackCoordinator.setRootCoordinator(settingsScreenCoordinator, animated: animated)
-        
-        parameters.navigationSplitCoordinator.setSheetCoordinator(navigationStackCoordinator) { [weak self] in
-            guard let self else { return }
-            
-            navigationStackCoordinator = nil
-            actionsSubject.send(.dismissedSettings)
-        }
-        
-        actionsSubject.send(.presentedSettings)
```
pixlwave, line 147: Got rid of all of this stuff so that the flow can't present itself anymore. It made sense when we first built it, but doesn't really fit with how we do flow coordinators any more.

UnitTests/Sources/ChatsFlowCoordinatorTests.swift
pixlwave, line 1: I'm going to bring these back in `UserSessionFlowCoordinatorTests`, but that can wait for the next PR as I want to try something out with state machines…

Velin92 (approved):

PR #4393, opened 2025-08-07
Fix a bug when a compact split view doesn't react to a change of root coordinator in the sidebar.

pixlwave:
We haven't seen this before because we set the HomeScreen as the split's sidebar root immediately, however with Spaces we need to await the service, so the `presetSpaceList` function is called after the tab bar is presented.

<del>This is just the fix but I'm going to take a look at adopting `@Observable` here in a follow-up PR that can be easily reverted if necessary.</del> Using `@Observable` makes It Just Work™️ so I've adopted that in all the navigation coordinators (I've already been using it in `NavigationTabCoordinator` as a bit of a test). Will rebase merge this so it's easy to revert if necessary.

To be doubly sure…
UI tests run here: https://github.com/element-hq/element-x-ios/actions/runs/16808959108
Integration tests run here: https://github.com/element-hq/element-x-ios/actions/runs/16808961942

### DIFF
--- a/ElementX/Sources/Application/Navigation/NavigationCoordinators.swift
+++ b/ElementX/Sources/Application/Navigation/NavigationCoordinators.swift
@@ -760,6 +760,5 @@ private struct NavigationStackCoordinatorView: View {
             module.coordinator?.toPresentable()
                 .id(module.id)
         }
-        .animation(.elementDefault, value: navigationStackCoordinator.rootModule)
     }
 }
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Application/Navigation/NavigationCoordinators.swift
pixlwave, line 1: ⚠️ Note to self: Rebase merge this rather than squashing.

pixlwave:
The UI test failure is related to a change in #4372, so going to merge this.

PR #4413, opened 2025-08-14
Add a workaround to include some via parameters for room v12 tombstone links.

pixlwave:
Without them, following the link fails as previously the server could use the server name from the room ID (which in and of itself seems somewhat error prone).

iOS equivalent to https://github.com/element-hq/element-x-android/pull/5155

### DIFF
--- a/ElementX/Sources/Services/Room/RoomProxyProtocol.swift
+++ b/ElementX/Sources/Services/Room/RoomProxyProtocol.swift
@@ -200,4 +200,13 @@ extension JoinedRoomProxyProtocol {
         await updateMembers()
         return membersPublisher.value
     }
+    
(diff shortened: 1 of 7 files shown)


ElementX/Sources/Services/Room/RoomProxyProtocol.swift
```diff
@@ -200,4 +200,13 @@ extension JoinedRoomProxyProtocol {
         await updateMembers()
         return membersPublisher.value
     }
+    
+    // This is a horrible workaround for not having any server names available when using tombstone links with v12 room IDs.
```
stefanceriu, line 204: 🙈

stefanceriu (approved):
lgtm 👍

PR #4415, opened 2025-08-15
Fix some app route navigation bugs.

pixlwave:
A couple of bugs have been notice with the latest release:
- Opening a notification would update the stack with an animated pop.
- Opening a permalink from the timeline wouldn't push the room as a child.

The regression was introduced by #4390 which naively made the user session flow coordinator clear the existing route before presenting the new one (whereas it should only by clearing any sheets). I need to revisit this as it seems too error prone

### DIFF
--- a/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
@@ -157,20 +157,32 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
              .roomDetails, .roomMemberDetails, .userProfile,
              .event, .eventOnRoomAlias, .childEvent, .childEventOnRoomAlias,
              .call, .genericCallLink, .share, .transferOwnership:
-            clearRoute(animated: animated) // Make sure the presented route is visible.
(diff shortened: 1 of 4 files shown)


ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -157,20 +157,32 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
              .roomDetails, .roomMemberDetails, .userProfile,
              .event, .eventOnRoomAlias, .childEvent, .childEventOnRoomAlias,
              .call, .genericCallLink, .share, .transferOwnership:
-            clearRoute(animated: animated) // Make sure the presented route is visible.
+            clearPresentedSheets(animated: animated) // Make sure the presented route is visible.
```
pixlwave, line 160: This is the actual fix and aligns with what `handleAppRoute` was doing when Settings was a part of ChatsFlowCoordinator:
https://github.com/element-hq/element-x-ios/blob/f1f6cace7751c0707a2e2a40cbe7be2ff8f3c899/ElementX/Sources/FlowCoordinators/ChatsFlowCoordinator.swift#L116-L122

pixlwave:
Going to yolo this one as it is basically reverting a mistake.

PR #4424, opened 2025-08-19
Fix document picker tint colors, remove multi-selection feature flag

stefanceriu:


ElementX/Sources/Screens/MediaPickerScreen/DocumentPicker.swift
```diff
@@ -40,7 +40,9 @@ struct DocumentPicker: UIViewControllerRepresentable {
         return documentPicker
     }
     
-    func updateUIViewController(_ uiViewController: UIDocumentPickerViewController, context: Context) { }
+    func updateUIViewController(_ uiViewController: UIDocumentPickerViewController, context: Context) {
+        uiViewController.view.tintColor = .compound.textActionAccent
```
pixlwave, line 46: Let's maybe add a comment here to say that we're doing this because the checkmark is hard to see in dark mode with our default `textActionPrimary` colour?

Also, if we wanted to only go green in dark mode where we need it, it would be pretty simple:
```swift
UIColor { $0.isLight ? .compound.textActionPrimary : .compound.textActionAccent }
```

ElementX/Sources/Screens/MediaPickerScreen/PhotoLibraryPicker.swift
```diff
@@ -48,7 +48,9 @@ struct PhotoLibraryPicker: UIViewControllerRepresentable {
         return pickerViewController
     }
     
-    func updateUIViewController(_ uiViewController: PHPickerViewController, context: Context) { }
+    func updateUIViewController(_ uiViewController: PHPickerViewController, context: Context) {
+        uiViewController.view.tintColor = .compound.textActionAccent
```
pixlwave, line 54: and here.

PR #4429, opened 2025-08-21
Add a feature flag for spaces.

pixlwave:
This PR adds a feature flag for spaces. I wasn't planning on adding this yet, but it will be useful to make some UI tests of the flow so here we are.

I also took the opportunity whilst I was here to get rid of the conditional compilation in `AppSettings`, mainly by moving the various Configuration types into the `Application/Settings` directory.

### DIFF
--- a/ElementX/Sources/Application/Settings/AppSettings.swift
+++ b/ElementX/Sources/Application/Settings/AppSettings.swift
@@ -59,6 +59,7 @@ final class AppSettings {
         case enableKeyShareOnInvite
         case knockingEnabled
         case threadsEnabled
+        case spacesEnabled
         case developerOptionsEnabled
         
         // Doug's tweaks 🔧
@@ -372,6 +373,9 @@ final class AppSettings {
     
     @UserPreference(key: UserDefaultsKeys.threadsEnabled, defaultValue: false, storageType: .userDefaults(store))
     var threadsEnabled
+    
+    @UserPreference(key: UserDefaultsKeys.spacesEnabled, defaultValue: false, storageType: .userDefaults(store))
+    var spacesEnabled
     
     @UserPreference(key: UserDefaultsKeys.developerOptionsEnabled, defaultValue: isDevelopmentBuild, storageType: .userDefaults(store))
     var developerOptionsEnabled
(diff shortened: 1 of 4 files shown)


ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -307,8 +318,10 @@ final class AppSettings {
 
     // MARK: - Element Call
     
+    #if IS_MAIN_APP
     // swiftlint:disable:next force_unwrapping
     let elementCallBaseURL: URL = EmbeddedElementCall.appURL!
+    #endif
```
pixlwave, line 324: Could also get rid of this by linking the NSE/ShareExtension against `EmbeddedElementCall` but that seemed somewhat ott to me. Also, TIL that `#if canImport(EmbeddedElementCall )` wouldn't work here as technically the extensions can import it, they just can't use/link to it 🙃

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -266,6 +269,11 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
                 }
             }
             .store(in: &cancellables)
+        
+        appSettings.$spacesEnabled
+            .map { $0 ? .automatic : .hidden }
+            .weakAssign(to: \.chatsTabDetails.barVisibility, on: self)
+            .store(in: &cancellables)
```
pixlwave, line 276: Note there's a follow-up addition to this in #4430 that will keep the tab bar hidden until the user actually has spaces from the SDK.

ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -80,8 +80,6 @@ final class AppSettings {
         return apps.contains(InfoPlistReader.main.baseBundleIdentifier)
         #endif
     }()
-    
-    #if IS_MAIN_APP
```
stefanceriu, line 84: Wooo 🥳

stefanceriu (approved):
Love it!

PR #4430, opened 2025-08-21
Space flow improvements.

pixlwave:
This PR iterates on the spaces tab flow with the following changes:

- Use a state machine when selecting a space.
- Add a `SpaceFlowCoordinator` for recursively presenting subspaces.
    - This means that the `RoomFlowCoordinator` will be able to present a space when encountering a permalink.
- Get the `spaceRoomListProxy` in the view model where a space was tapped so we can indicate failures.
- Fix the total room count in SpaceListScreen.
- Add a UI test to cover the flow.

Builds on #4429.


UITests/Sources/__Snapshots__/Application/userSessionScreen.testSpaceExploration-iPad-18-5-en-GB-6.png
pixlwave, line 1: A few notes:
- The tab bar is only clipped because we hide the status bar for tests
- Both sides of the split are still WIP and don't match the final designs (e.g. bloom size is wrong and the double title is wrong)

UITests/Sources/__Snapshots__/Application/userSessionScreen.testSpaceExploration-iPhone-18-5-en-GB-5.png
pixlwave, line 1: Also, the tab bar should be hidden on iPhone[^1] at this stage, but annoyingly there is going to be more to hook that up than I had planned.

[^1]: well, any compact horizontal size class

UITests/Sources/__Snapshots__/Application/userSessionScreen.testSpaceExploration-iPhone-18-5-en-GB-6.png
pixlwave, line 1: Also, the tab bar should be hidden on iPhone[^1] at this stage, but annoyingly there is going to be more to hook that up than I had planned.

[^1]: well, any compact horizontal size class

ElementX/Sources/FlowCoordinators/SpaceExplorerFlowCoordinator.swift
```diff
@@ -77,30 +87,81 @@ class SpaceExplorerFlowCoordinator: FlowCoordinatorProtocol {
     // MARK: - Private
     
     private func configureStateMachine() {
-        stateMachine.addRoutes(event: .start, transitions: [.initial => .spaceList]) { [weak self] _ in
+        stateMachine.addRoutes(event: .start, transitions: [.initial => .spaceList(selectedSpaceID: nil)]) { [weak self] _ in
             self?.presentSpaceList()
         }
         
+        stateMachine.addRouteMapping { event, fromState, userInfo in
+            guard event == .selectSpace, case .spaceList = fromState else { return nil }
```
stefanceriu, line 95: Don't we want to fatal error on this as well?

ElementX/Sources/FlowCoordinators/SpaceExplorerFlowCoordinator.swift
```diff
@@ -77,30 +87,81 @@ class SpaceExplorerFlowCoordinator: FlowCoordinatorProtocol {
...
+                }
             }
-        }
-        .store(in: &cancellables)
+            .store(in: &cancellables)
         
         sidebarNavigationStackCoordinator.setRootCoordinator(coordinator)
     }
+    
+    private var spaceFlowCoordinator: SpaceFlowCoordinator?
```
stefanceriu: Let's move this one to the top so it's clear?

ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
```diff
@@ -0,0 +1,177 @@
...
+        /// A child flow is in progress.
+        case presentingChild(childSpaceID: String, previousState: State)
+    }
+    
+    enum Event: EventType {
+        /// The flow is being started.
+        case start
+        
+        case startChildFlow
+        case tearDownChildFlow
```
stefanceriu: Would look better with start/stop or setUp/tearDown

ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
```diff
@@ -0,0 +1,177 @@
...
+    
+    // MARK: - Private
+    
+    private func configureStateMachine() {
+        stateMachine.addRoutes(event: .start, transitions: [.initial => .space]) { [weak self] _ in
+            self?.presentSpace()
+        }
+        
+        stateMachine.addRouteMapping { event, fromState, userInfo in
+            guard event == .startChildFlow, case .space = fromState else { return nil }
```
stefanceriu, line 109: Same question here, maybe I'm missing something.

ElementX/Sources/Screens/Spaces/SpaceListScreen/SpaceListScreenModels.swift
```diff
@@ -18,13 +18,17 @@ struct SpaceListScreenViewState: BindableState {
...
     
     var bindings: SpaceListScreenViewStateBindings
     
     var subtitle: String {
         L10n.screenSpaceListDetails(L10n.commonSpaces(joinedSpaces.count), L10n.commonRooms(joinedRoomsCount))
     }
+    
+    var joinedRoomsCount: Int {
+        joinedSpaces.map(\.childrenCount).reduce(0, +)
+    }
```
stefanceriu, line 31: Heh, nice 👍

stefanceriu (approved):
Great stuff, this is coming together super nicely! 👏

ElementX/Sources/FlowCoordinators/SpaceExplorerFlowCoordinator.swift
pixlwave: Ooops yeah, I forgot to move it after getting it working.

ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
pixlwave: Naming is hard 😆

I've added doc comments to clarify that `stopChildFlow` is for after it has been dismissed.

ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
pixlwave: Yeah so this style of configuration adds the mapping and the handler for it together. So returning `nil` here (and above) doesn't mean it is unexpected, but means that this mapping doesn't cover that particular event/state combination.

pixlwave:
SonarCloud is basically update that both view models select spaces and have failure indicators 😐

PR #4432, opened 2025-08-22
Make the room list filters smaller.

amshakal:
I changed the padding and text size of the room filters.


pixlwave (approved):
LGTM, I've pushed the snapshots and everything is now green 👌

PR #4436, opened 2025-08-24
Preserve the user chosen order when uploading multiple media files

### LINKED ISSUE #4435: Multiselecting images sends them in a random order.
### Steps to reproduce

1. Want to send a series of images in a given order
2. Select them in that order
3. Hit send
4. Discover they are sent in an entirely random order.

#### What did you expect?

Preserve the ordering. Ideally, the order would be shown with little numbers as you select them, so you can keep track of what you're sending in which order.

#### What happened instead?

Randomness

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

998

### Homeserver

_No response_

### Will you send logs?

Yes

stefanceriu:
The nice thing about task groups is that they return results as they finish processing, but that also means the order is not preserved. The generic solution here is to map back results into their original indexes before returning the final list.

This PR also enables `PHPickerConfiguration.selection = .ordered` which will show numbers on selected items based on whatever order they were chosen in.

Fixes #4435.

### DIFF
--- a/ElementX/Sources/Screens/MediaPickerScreen/PhotoLibraryPicker.swift
+++ b/ElementX/Sources/Screens/MediaPickerScreen/PhotoLibraryPicker.swift
@@ -34,7 +34,7 @@ struct PhotoLibraryPicker: UIViewControllerRepresentable {
 
     func makeUIViewController(context: Context) -> PHPickerViewController {
         var configuration = PHPickerConfiguration(photoLibrary: .shared())
-        
+        configuration.selection = .ordered
         configuration.selectionLimit = switch selectionType {
         case .single:
             1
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Screens/MediaPickerScreen/PhotoLibraryPicker.swift
```diff
@@ -34,7 +34,7 @@ struct PhotoLibraryPicker: UIViewControllerRepresentable {
 
     func makeUIViewController(context: Context) -> PHPickerViewController {
         var configuration = PHPickerConfiguration(photoLibrary: .shared())
-        
+        configuration.selection = .ordered
```
pixlwave, line 37: Neat 👌

PR #4448, opened 2025-08-27
Introduce flow parameters to simplify dependencies for child flows.

pixlwave:
This PR makes a common set of parameters that can be used for easy dependency injection into our flows.

I'm going to carry on and do some more tidying around our use of ServiceLocator within the flows (not going to touch the tests) but thought this was enough for a first PR.

### DIFF
--- a/ElementX/Sources/Application/FlowCoordinatorProtocol.swift
+++ b/ElementX/Sources/Application/FlowCoordinatorProtocol.swift
@@ -14,3 +14,25 @@ protocol FlowCoordinatorProtocol {
     func handleAppRoute(_ appRoute: AppRoute, animated: Bool)
     func clearRoute(animated: Bool)
 }
+
+/// Core parameters that are shared across the main flows for easy dependency injection.
+///
+/// Please do **not** pass this type directly to screen coordinators/view models.
+@MainActor
+struct CommonFlowParameters {
+    let userSession: UserSessionProtocol
+    let bugReportService: BugReportServiceProtocol
+    let elementCallService: ElementCallServiceProtocol
+    let timelineControllerFactory: TimelineControllerFactoryProtocol
+    let emojiProvider: EmojiProviderProtocol
+    let appMediator: AppMediatorProtocol
+    let appSettings: AppSettings
+    let appHooks: AppHooks
+    let analytics: AnalyticsService
+    let userIndicatorController: UserIndicatorControllerProtocol
+    let notificationManager: NotificationManagerProtocol
+    let stateMachineFactory: StateMachineFactoryProtocol
+    
+    var windowManager: WindowManagerProtocol { appMediator.windowManager }
+    var ongoingCallRoomIDPublisher: CurrentValuePublisher<String?, Never> { elementCallService.ongoingCallRoomIDPublisher }
+}
--- a/ElementX/Sources/FlowCoordinators/OnboardingFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/OnboardingFlowCoordinator.swift
@@ -57,23 +57,18 @@ class OnboardingFlowCoordinator: FlowCoordinatorProtocol {
     
     private var verificationStateCancellable: AnyCancellable?
     
-    init(userSession: UserSessionProtocol,
+    init(isNewLogin: Bool,
          appLockService: AppLockServiceProtocol,
-         analyticsService: AnalyticsService,
-         appSettings: AppSettings,
-         notificationManager: NotificationManagerProtocol,
          navigationStackCoordinator: NavigationStackCoordinator,
-         userIndicatorController: UserIndicatorControllerProtocol,
-         windowManager: WindowManagerProtocol,
-         isNewLogin: Bool) {
-        self.userSession = userSession
-        self.appLockService = appLockService
-        self.analyticsService = analyticsService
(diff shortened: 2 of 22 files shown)


ElementX/Sources/Application/FlowCoordinatorProtocol.swift
pixlwave, line 1: Totally open to a better name for these, this was just a working name.

ElementX/Sources/Application/FlowCoordinatorProtocol.swift
```diff
@@ -14,3 +14,25 @@ protocol FlowCoordinatorProtocol {
...
     func clearRoute(animated: Bool)
 }
+
+/// Core parameters that are shared across the main flows for easy dependency injection.
+///
+/// Please do **not** pass this type directly to screen coordinators/view models.
+@MainActor
+struct CommonFlowParameters {
+    let userSession: UserSessionProtocol
+    let bugReportService: BugReportServiceProtocol
```
pixlwave, line 24: Also, maybe this one shouldn't be in here. It is only really used in the UserSessionFlow and the SettingsFlow (it is also in the ChatsFlow but I think we could move the crash detection up one level).

ElementX/Sources/Application/FlowCoordinatorProtocol.swift
stefanceriu: I like it, I think it works well!

ElementX/Sources/Application/FlowCoordinatorProtocol.swift
stefanceriu: I don't think it's a big problems, we can tweak that later.

ElementX/Sources/FlowCoordinators/OnboardingFlowCoordinator.swift
```diff
@@ -57,23 +57,18 @@ class OnboardingFlowCoordinator: FlowCoordinatorProtocol {
...
-        self.windowManager = windowManager
+         flowParameters: CommonFlowParameters) {
         self.isNewLogin = isNewLogin
+        userSession = flowParameters.userSession
+        self.appLockService = appLockService
+        analyticsService = flowParameters.analytics
+        appSettings = flowParameters.appSettings
+        notificationManager = flowParameters.notificationManager
+        userIndicatorController = flowParameters.userIndicatorController
+        windowManager = flowParameters.windowManager
```
stefanceriu, line 71: Should we just store the `flowParameters` instead of assigning to locals now?

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -56,17 +56,19 @@ struct FocusEvent: Hashable {
...
+    
+    private var userSession: UserSessionProtocol { flowParameters.userSession }
+    private var timelineControllerFactory: TimelineControllerFactoryProtocol { flowParameters.timelineControllerFactory }
+    private var emojiProvider: EmojiProviderProtocol { flowParameters.emojiProvider }
+    private var ongoingCallRoomIDPublisher: CurrentValuePublisher<String?, Never> { flowParameters.ongoingCallRoomIDPublisher }
+    private var appMediator: AppMediatorProtocol { flowParameters.appMediator }
+    private var appSettings: AppSettings { flowParameters.appSettings }
+    private var appHooks: AppHooks { flowParameters.appHooks }
+    private var analytics: AnalyticsService { flowParameters.analytics }
+    private var userIndicatorController: UserIndicatorControllerProtocol { flowParameters.userIndicatorController }
```
stefanceriu: Oh yeah, like this

ElementX/Sources/FlowCoordinators/SettingsFlowCoordinator.swift
```diff
@@ -73,9 +65,9 @@ class SettingsFlowCoordinator: FlowCoordinatorProtocol {
     // MARK: - Private
     
     private func presentSettingsScreen(animated: Bool) {
-        let settingsScreenCoordinator = SettingsScreenCoordinator(parameters: .init(userSession: parameters.userSession,
-                                                                                    appSettings: parameters.appSettings,
-                                                                                    isBugReportServiceEnabled: parameters.bugReportService.isEnabled))
+        let settingsScreenCoordinator = SettingsScreenCoordinator(parameters: .init(userSession: flowParameters.userSession,
+                                                                                    appSettings: flowParameters.appSettings,
+                                                                                    isBugReportServiceEnabled: flowParameters.bugReportService.isEnabled))
```
stefanceriu, line 70: I think we should aggree on a single way of doing this. `flowParameters.whatever` would be my personal preference fwiw.

stefanceriu (commented):
Very nice, thank you for doing this!

ElementX/Sources/FlowCoordinators/SettingsFlowCoordinator.swift
pixlwave: Gone with that in https://github.com/element-hq/element-x-ios/pull/4448/commits/4b54b0de116c36307f3226ee7cb7f52a23ad415e however I've kept a computed property for the `userSession` seeing as that one is frequently used directly.

stefanceriu (approved):
🐑 ity 🐑 🚢

PR #4449, opened 2025-08-27
SwiftSoup + AttributedStrings = ❤️ 

stefanceriu:
This PR introduces a novel way of dealing with HTML messages and how we build attributed strings for rendering purposes. 

Instead of having `DTCoreText` parse the HTML and generate the AttributedString, which we would then postprocess to remove various artifacts or hook into to customize, we now do the following:
* Create an `AttributedStringBuilderV2` and transform the original `AttributedStringBuilder` into a somwhat class cluster kind of object so we're not forced to change any of the users
* Add a feature flag to switch between them at runtime
* Adopt `SwiftSoup` and have it build a proper HTML DOM
* Walk the hierarchy and build the AttributedString while having full control over each and every tag and attribute
* Postprocess the string *only* to detect phishing attempts and add matrix specific identifier links

This PR also:
* introduces specific `HTMLFixtures`
* overhauls `FormattedBodyText` previews
* removes the crash simulation option from the developer tools menu
* renames `detectPermalinks` to `addMatrixEntityPermalinkAttributesTo`


ElementX/Sources/Other/HTMLParsing/AttributedStringBuilder.swift
```diff
@@ -388,35 +33,34 @@ extension NSAttributedString.Key {
...
-        switch type {
-        case .userID(let identifier):
-            return try? URL(string: matrixToUserPermalink(userId: identifier))
-        case .link(let urlString):
-            return URL(string: urlString)
-        default:
-            return nil
+    static func invalidateCaches() {
+        AttributedStringBuilderV1.invalidateCaches()
+    }
```
pixlwave: Do we need to do this for V2 too?

ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/View/DeveloperOptionsScreen.swift
```diff
@@ -116,13 +120,6 @@ struct DeveloperOptionsScreen: View {
                         .frame(maxWidth: .infinity)
                         .alignmentGuide(.listRowSeparatorLeading) { _ in 0 } // Fix separator alignment
                 }
-                
-                Button {
-                    fatalError("This crash is a test.")
-                } label: {
-                    Text("💥")
-                        .frame(maxWidth: .infinity)
-                }
```
pixlwave, line 128: So long old friend 🫡

PreviewTests/Sources/__Snapshots__/PreviewTests/formattedBodyText.iPad-en-GB-0.png
pixlwave, line 1: Should we re-introduce the code block with the matrix.org links in it so we can verify that the links are stripped out?

pixlwave: And maybe we should have this one back too to make sure linespacing is consistent?

Text before blockquote

> bold italic

Text after blockquote

PreviewTests/Sources/__Snapshots__/PreviewTests/formattedBodyText.iPhone-16-en-GB-1.png
pixlwave, line 1: Can we fix H1 & H2 to match H3 like we currently do?

ElementX/Sources/Other/HTMLParsing/HTMLFixtures.swift
pixlwave, line 1: Very nice tidy-up 👌

ElementX/Sources/Other/HTMLParsing/AttributedStringBuilderV2.swift
```diff
@@ -111,6 +84,105 @@ struct AttributedStringBuilderV2: AttributedStringBuilderProtocol {
...
+                
+            case "sup":
+                content = attributedString(from: childElement, preserveFormatting: preserveFormatting)
+                content.addAttribute(.baselineOffset, value: 6, range: NSRange(location: 0, length: content.length))
+                content.setFontPreservingSymbolicTraits(UIFont.systemFont(ofSize: UIFont.systemFontSize * 0.7))
+                
+            case "sub":
+                content = attributedString(from: childElement, preserveFormatting: preserveFormatting)
+                content.addAttribute(.baselineOffset, value: -4, range: NSRange(location: 0, length: content.length))
+                content.setFontPreservingSymbolicTraits(UIFont.systemFont(ofSize: UIFont.systemFontSize * 0.7))
```
pixlwave: Should we use [NSSuperscriptAttributeName](https://duckduckgo.com/?q=NSSuperscriptAttributeName&t=osx&ia=web) here instead perhaps? I think you use negative values to subscript.

ElementX/Sources/Other/HTMLParsing/AttributedStringBuilderV1.swift
```diff
@@ -104,7 +104,7 @@ struct AttributedStringBuilderV1: AttributedStringBuilderProtocol {
         removeDTCoreTextArtifacts(mutableAttributedString)
         
         let result = try? AttributedString(mutableAttributedString, including: \.elementX)
-        Self.cacheValue(result, forKey: htmlString, cacheKey: cacheKey)
+        Self.cacheValue(result, forKey: originalHTMLString, cacheKey: cacheKey)
```
pixlwave: 🙈

UnitTests/Sources/AttributedStringBuilderV2Tests.swift
pixlwave, line 1: I'm not really a fan of these if I'm honest as it isn't super clear which ones are adjusted as a WIP… Could we maybe share the code from V1 tests (with a configuration) and use early returns for the ones we know don't work yet. That would then give us a list of tests that need fixing before we release this feature. e.g.

- `testAllUsersIsIgnoredInCode()` changed from 1 run to 3, which to me, implies that `@room` isn't being ignored here…
- `testURLs`/`HyperlinksAreIgnoredInCode()` have both been removed now.

ElementX/Sources/Other/HTMLParsing/AttributedStringBuilder.swift
stefanceriu: We sure do 👍

PreviewTests/Sources/__Snapshots__/PreviewTests/formattedBodyText.iPad-en-GB-0.png
stefanceriu: > maybe we should have this one back too

Yep yep, good point

> Should we re-introduce the code block with the matrix.org links in it so we can verify that the links are stripped out?

We already have some links in the code block fixtures, I'll make sure they work properly.

stefanceriu: These are both sorted now 👍

PreviewTests/Sources/__Snapshots__/PreviewTests/formattedBodyText.iPhone-16-en-GB-1.png
stefanceriu: Done 👍

UnitTests/Sources/AttributedStringBuilderV2Tests.swift
stefanceriu: Did this in 83426a533185a9246e4541524b0e531c316256d7 and looks quite good. Also surprisingly Xcode has special menus for running tests from subclasses.

ElementX/Sources/Other/HTMLParsing/AttributedStringBuilderV2.swift
stefanceriu: I can't find a direct replacement for `NSSuperscriptAttributeName` in swift, apparently `baselineOffset` is the way to go.

PreviewTests/Sources/__Snapshots__/PreviewTests/formattedBodyText.iPhone-16-en-GB-0.png
pixlwave, line 1: Is it expect that "Text after blockquote" has an extra space before the `T`?

pixlwave (approved):
👏

PreviewTests/Sources/__Snapshots__/PreviewTests/formattedBodyText.iPhone-16-en-GB-0.png
stefanceriu: Seems to be the same on both builders so I guess so? 🤔

PR #4457, opened 2025-09-01
A11y audit improvements and fixes

Velin92:
- Hidden some unnecessary avatars from voice over
- Ignored element descriptions in a rendering test
- Better rendering of the accessibility representation for the files list 
- Added the legacy room id to the regex, so that we can also filter false positives of non human readable labels for them since we still use them in tests sometimes

### DIFF
--- a/AccessibilityTests/Sources/AccessibilityTests.swift
+++ b/AccessibilityTests/Sources/AccessibilityTests.swift
@@ -16,7 +16,7 @@ final class AccessibilityTests: XCTestCase {
         app = Application.launch(viewID: name)
         await client.waitForApp()
         defer { try? client.stop() }
-                
+        
         try client.send(.accessibilityAudit(.nextPreview))
         
         // To handle location sharing popup in CI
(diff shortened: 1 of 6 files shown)


AccessibilityTests/Sources/AccessibilityTests.swift
```diff
@@ -84,16 +89,23 @@ final class AccessibilityTests: XCTestCase {
...
     
     /// Use this array to filter add specific filters to ignore specific issues for certain elements
     private static let ignoredA11yIdentifiers: [String: [FilterType]] = [
         A11yIdentifiers.serverConfirmationScreen.serverPicker: [.compactDescription(notHumanReadableMessage)]
     ]
+    
+    /// Use this array to filter add specific filters to ignore specific tests
+    private static let ignoredA11yTest: [String: [FilterType]] = [
+        // It's an image rendering test doesn't need to have descriptions
+        "RoomAvatarImage_Previews-0": [.auditType(.sufficientElementDescription)]
```
Velin92, line 107: sadly we can't access their names directly so we need to keep in mind that if we are ignoring a whole test specific issue and we rename the view we need to rename it manually here

PR #4459, opened 2025-09-02
Move call presentation from the chats flow into the user session flow.

pixlwave:
In preparation for pushing a Room flow in the Spaces tab.


ElementX/Sources/Application/Navigation/NavigationRootCoordinator.swift
pixlwave, line 1: Unrelated, but in moving the other overlay I realised this one also exists and didn't get the VoiceOver fix applied in #4265.

Velin92 (approved):

PR #4462, opened 2025-09-02
Fix some retain cycles that could keep the Client(Proxy) alive when clearing the cache.

pixlwave:
This was causing the `UserSession`/`Client` to stay alive after clearing the cache (if the `RoomDetailsScreen` had been pushed).

Found a few more triggered by pushing the room screen (the openURL action) and focussing the composer (the `weak` `Client` workaround).


ElementX/Sources/Screens/RoomMemberDetailsScreen/RoomMemberDetailsScreenViewModel.swift
pixlwave, line 1: Came across this one too when looking at all the uses of this publisher.

Velin92 (approved):

Issue #4465, opened 2025-09-04
iOS holds onto the Context of the last screen that had keyboard focus.


pixlwave:
Making a note here off of the back of #4462. @Velin92 and I did some more debugging as the [final commit](https://github.com/element-hq/element-x-ios/commit/927aea4bfe59acd963451e84744f741df14d6dff) in that PR was very much a workaround. We eventually discovered that we always have 1 instance of the most recent `StateStoreViewModel.Context` that had a text field focused, due to a chain of `UICompatibilityInputViewController`/`TUIKeyboardBackendController`/`UIKBAutofillController`/`AFUITargetDetectionController` seemingly keeping said text field alive.

For the room screen this results in both the `ComposerToolbarViewModel.Context` and the `TimelineViewModel.Context` being kept alive after it has been popped (along with a `WysiwygComposerViewModel`). As soon as you focus another text field (e.g. the composer on a different room screen or the element call URL in developer options) these are correctly released. It looks something like this:

<img width="2555" height="284" alt="Image" src="https://github.com/user-attachments/assets/925c9e0b-e174-4652-aa8b-6a3ee0b834db" />

The annoying part is that when one of these `Context`s contains the optional [MediaProvider](https://github.com/element-hq/element-x-ios/blob/53eb1c1718bea964bf40c9920ab537b71001bf4f/ElementX/Sources/Other/SwiftUI/ViewModel/StateStoreViewModel.swift#L66-L68), then this is also retained, keeping a (now weak) reference to the SDK's `Client`. I think for now I'm going to revert the workaround above and try making that MediaProvider `unowned` and see what happens 🤞.

Velin92:
@pixlwave we fixed this right? I guess we can close it.

PR #4466, opened 2025-09-04
Address the real lifetime issue of the SDK's `Client` by making `Context.mediaProvider` unowned.

### LINKED ISSUE #4447: Excessive disk storage on iOS
### Steps to reproduce

I noticed ElementX using 30GB of storage space. After deleting the data and app, it has risen back to 15GB after just a few days. 

#### What did you expect?
Element X uses a reasonable amount of storage

#### What happened instead?
Element X uses tens of GBs of data
Clearing the cache does not free the space

### Your phone model

iPhone 16

### Operating system version

iOS 18.5

### Application version

25.08.6 (998)

### Homeserver

Synapse 1.137

### Will you send logs?

No

pixlwave:
This PR reverts the last commit from #4462 and instead addresses the issue by recognising that #4465 is an issue and that our `Context`s should not own SDK objects. I'm going to say that between the obvious retain cycle in #4462 and this more correct fix for releasing `MediaLoader`, we should be good to close #4447.


ElementX/Sources/Other/SwiftUI/ViewModel/StateStoreViewModelV2.swift
```diff
@@ -67,7 +67,12 @@ class StateStoreViewModelV2<State: BindableState, ViewAction> {
     
         /// An optional image loading service so that views can manage themselves
         /// Intentionally non-generic so that it doesn't grow uncontrollably
-        let mediaProvider: MediaProviderProtocol?
+        ///
+        /// We noticed that the keyboard appears to hold onto the Context of the last screen that had text input focus:
+        /// https://github.com/element-hq/element-x-ios/issues/4465
+        /// Therefore this is `unowned` so that the underlying `MatrixRustSDK.Client` isn't retained when e.g. clearing
+        /// the cache, otherwise we have the potential for 2 `Client`s to be alive at the same time causing havoc.
+        unowned let mediaProvider: MediaProviderProtocol?
```
Velin92: if we are already using an optional here there is no issue in using weak instead of unowned

pixlwave: Turns out that `unowned` was completely incompatible with our preview tests anyway 😅

pixlwave: urgh, it wasn't to do with unowned, but seems more like nothing retains it in the tests now that it isn't owned by the context 😒

PR #4467, opened 2025-09-04
A11y timeout announcement for verification request, hardcoded to 1 minute

Velin92:
Name says it all.
The timeout is decided by the clients, and the least amount of time for a timeout experienced is between EW and EX which is precisely 60 seconds, between other client instances the time might actually be more.


Velin92:
> LGTM, it would be nice if we could read the actual value from the SDK and pass it as placeholder but no biggy right now.
> 
> > The timeout is decided by the clients, and the least amount of time for a timeout experienced is between EW and EX which is precisely 60 seconds, between other client instances the time might actually be more.
> 
> Is this the same for all 4 flows (incoming & outgoing for own devices & other users)?

Yes. A client can send the `m.key.verification.cancel` at any time they want if they want to cancel it, so this means that any client can timeout whenever they want with different timeout times... Which makes getting the value from the SDK not a valid fix, since the other client can have a different timeout time. (e.g EW)

PR #4472, opened 2025-09-04
Use the SDK's offline detection everywhere (except for restarting the sync loop).

pixlwave:
This should make EXI behave the same as EXA when matrix.org went down (for an already built session). Instead of showing the syncing indicator on and off (as the network is reachable), we now show the offline indicator permanently until the `/sync` loop resumes.

The first commit switches to using it for the offline indicator, the second for all of the screens that use network reachability for retries etc.

Before and after when blocking the HS with a firewall:

https://github.com/user-attachments/assets/5edc408f-3d95-4f21-910d-b3eed01ab1eb

https://github.com/user-attachments/assets/79882f67-a8c9-4031-b75c-d2b302f0907b


Hywan:
Bonus: it will finally fix my issue when my VPN shut down, and Element X doesn't enter offline mode \o/.

Velin92:
Isn't the network monitor used anymore? The homeserver is already able to distinguish between an unreachable state from the client (offline) and an unreachable state from the server?

pixlwave:
> Isn't the network monitor used anymore? The homeserver is already able to distinguish between an unreachable state from the client (offline) and an unreachable state from the server?

At the point you commented it was still used to restart the sync immediately when changing networks (compared to waiting for the SDK's timeout to retry).

I've just pushed a new commit that additionally uses it to make the network indicator distinguish between offline and the server being unreachable.

PR #4440, opened 2025-08-25
Improve global proxy detection, building and logging.

stefanceriu:
- set schemes when building the proxy URLs
- detect the socks and PAC variants next to http and https
- log the various codepaths

PR #4481, opened 2025-09-08
Fix proxies by reverting the change that introduced schemes.

pixlwave:
This partially reverts #4440, keeping the logging but returning the same value as before (with a leniency for not having a port).


stefanceriu (approved):
👌

PR #4482, opened 2025-09-08
Add a migration to expire sync sessions so that m.space.* state is up to date.

pixlwave:
The SDK added a required state for `m.space.child` and `m.space.parent`, however Synapse doesn't seem to notice this and so we only get some of the data slowly over multiple syncs and never get the rest. This PR adds a migration to expire the sync sessions so that the SDK's `SpaceService` has everything it needs to compute the space graph and only show us real top-level spaces.

### DIFF
--- a/ElementX/Sources/Mocks/Generated/GeneratedMocks.swift
+++ b/ElementX/Sources/Mocks/Generated/GeneratedMocks.swift
@@ -2387,6 +2387,41 @@ class ClientProxyMock: ClientProxyProtocol, @unchecked Sendable {
         stopSyncCompletionCallsCount += 1
         stopSyncCompletionClosure?(completion)
     }
+    //MARK: - expireSyncSessions
+
+    var expireSyncSessionsUnderlyingCallsCount = 0
+    var expireSyncSessionsCallsCount: Int {
+        get {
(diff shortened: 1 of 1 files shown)


ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -29,6 +29,7 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
     /// Common background task to continue long-running tasks in the background.
     private var backgroundTask: UIBackgroundTaskIdentifier?
     
+    private var userSessionMigrationsOldVersion: Version?
```
Velin92, line 32: I would just call this `oldVersion`, the function `performUserSessionMigration` already explains its purpose for migrating the user session

Velin92 (approved):

ElementX/Sources/Application/AppCoordinator.swift
pixlwave: Do you not think `oldVersion` would be too generic seeing as this is set to `nil` after `performUserSessionMigrations` has run? I wouldn't want that to be missed and someone to expect it was valid to use later in the flow.

PR #4495, opened 2025-09-10
Update the Enterprise submodule.

pixlwave:
Small fix for a bug introduced in #4342, already reviewed and approved.


pixlwave:
I'm going to ignore those test failures given they're on the AttributedStringBuilder.

PR #4496, opened 2025-09-10
Tweak attributed string formatting

stefanceriu:
* add new lines after headers
* preserve formatting of parentless text nodes


ElementX/Sources/Other/HTMLParsing/AttributedStringBuilderV2.swift
```diff
@@ -86,14 +86,22 @@ struct AttributedStringBuilderV2: AttributedStringBuilderProtocol {
     // MARK: - Private
     
     // swiftlint:disable:next function_body_length
-    func attributedString(from element: Element, preserveFormatting: Bool,
+    func attributedString(body: Element,
+                          element: Element,
+                          preserveFormatting: Bool,
```
pixlwave: I wonder if this should have documentation to explain the different between `body` and `element` or if we should use names that match SwiftSoup's API

PR #4501, opened 2025-09-10
Add support for joining rooms from a space.

### LINKED ISSUE #4484: [Task] Join a Room from a Space.

- [x] Make the JoinRoomScreen aware of the restricted join rule (by removing the banner if the user is a member of the specified room).
- [x] Present the JoinRoomScreen from the SpaceScreen.

pixlwave:
- Read the `restricted` join rule's `AllowRules` to hide the space membership banner on the `JoinRoomScreen`.
    - Basically has no effect as Synapse doesn't include the `allow_room_ids` anyway.
- Enable tapping an unjoined room so that it shows the `JoinRoomScreen`.
- Hook up the Join button for both rooms and spaces to join immediately.
    - Need to make tapping an unjoined space show the right thing in a later PR.
- Small refactor of a bunch of `spaceRoom` → `spaceRoomProxy`.

Closes #4484

### DIFF
--- a/ElementX/Sources/Screens/JoinRoomScreen/View/JoinRoomScreen.swift
+++ b/ElementX/Sources/Screens/JoinRoomScreen/View/JoinRoomScreen.swift
@@ -306,7 +306,8 @@ struct JoinRoomScreen: View {
 struct JoinRoomScreen_Previews: PreviewProvider, TestablePreview {
     static let unknownViewModel = makeViewModel(mode: .unknown)
     static let joinableViewModel = makeViewModel(mode: .joinable)
-    static let restrictedViewModel = makeViewModel(mode: .restricted)
+    static let restrictedViewModel = makeViewModel(mode: .restricted, canJoinRoom: false)
+    static let restrictedJoinableViewModel = makeViewModel(mode: .restricted)
     static let inviteRequiredViewModel = makeViewModel(mode: .inviteRequired)
     static let invitedViewModel = makeViewModel(mode: .invited(isDM: false))
     static let invitedDMViewModel = makeViewModel(mode: .invited(isDM: true))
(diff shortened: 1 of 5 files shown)


ElementX/Sources/Screens/Spaces/Common/SpaceRoomCell.swift
```diff
@@ -116,6 +117,12 @@ struct SpaceRoomCell: View {
             Button(L10n.actionJoin) { action(.join(spaceRoomProxy)) }
                 .font(.compound.bodyLG)
                 .foregroundStyle(.compound.textActionAccent)
+                .opacity(isJoining ? 0 : 1)
+                .overlay {
+                    if isJoining {
+                        ProgressView()
+                    }
+                }
```
Velin92, line 125: I think the progress view doesn't scale while the button could, so maybe worth also disabling the button in case in a larger text it ends tappable by mistake

ElementX/Sources/Screens/Spaces/SpaceScreen/SpaceScreenViewModel.swift
```diff
@@ -89,6 +95,25 @@ class SpaceScreenViewModel: SpaceScreenViewModelType, SpaceScreenViewModelProtoc
...
+        state.joiningRoomIDs.insert(spaceRoomProxy.id)
+        defer { state.joiningRoomIDs.remove(spaceRoomProxy.id) }
+        
+        guard case .success = await clientProxy.joinRoom(spaceRoomProxy.id, via: []) else {
+            showFailureIndicator()
+            return
+        }
+        
+        // If multiple join operations are running, then only show the last one.
+        guard state.joiningRoomIDs == [spaceRoomProxy.id] else { return }
```
Velin92, line 108: not super sure about this, is it the last one, or the only one given this check?

pixlwave: It would be the last one to finish.

If we had the operations `["space1", "space2", "space3"]` and they finished in the order `space3`, `space1`, `space2`, then we will push `space2` as that would be the last one to finish.

ElementX/Sources/Screens/Spaces/Common/SpaceRoomCell.swift
pixlwave: When a button has `opacity(0)` it is no longer tappable so this will be fine – I just double checked it to be 100% sure.

PR #4499, opened 2025-09-10
Element Call: Send rtc.decline event when incoming call is declined

BillCarsonFr:
### Pull Request Checklist

Fixes #4493

Depends on github.com/matrix-org/matrix-rust-sdk/pull/5614

- [ ] I read the [contributing guide](https://github.com/element-hq/element-ios/blob/develop/CONTRIBUTING.md).
- [ ] Pull request contains a [changelog label](https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md#changelog).

**UI changes have been tested with:**
- [ ] iPhone and iPad simulators in portrait and landscape orientations.
- [ ] Dark mode enabled and disabled.
- [ ] Various sizes of dynamic type.
- [ ] Voiceover enabled.

PR #4505, opened 2025-09-12
Listen to call decline to stop ringing when declined from other device

### LINKED ISSUE #4493: [Task] Element-Call Send RTC decline when the Incoming Call notification decline button is called
In addition to stop ringing and dismissing the notification, also send the new rtc.decline https://github.com/matrix-org/matrix-rust-sdk/pull/5614

BillCarsonFr:
### Pull Request Checklist

Fixes Fixes #4493
Based on #4499 

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Services/ElementCall/ElementCallService.swift
+++ b/ElementX/Sources/Services/ElementCall/ElementCallService.swift
@@ -358,13 +358,12 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
         
         MXLog.info("Observe decline events for notification \(rtcNotificationID)")
         
-        let ownUserID = clientProxy.userID
         let listener: CallDeclineListener = SDKListener { [weak self] senderID in
             guard let self else { return }
             
             MXLog.debug("Call declined event received from \(senderID)")
             
-            if senderID == ownUserID {
+            if senderID == roomProxy.ownUserID {
                 // Stop ringing!
                 MXLog.debug("Call declined elsewhere")
                 reportEndedCall(incomingCallID: incomingCallID, reason: .declinedElsewhere)


ElementX/Sources/Services/Room/RoomProxyProtocol.swift
```diff
@@ -64,6 +64,13 @@ enum KnockRequestsState {
     case loaded([KnockRequestProxyProtocol])
 }
 
+struct RtcDeclinedEvent {
```
pixlwave: Following Swift's API design guidelines this should be:
```suggestion
struct RTCDeclinedEvent {
```

ElementX/Sources/Services/Room/RoomProxyProtocol.swift
```diff
@@ -75,7 +82,7 @@ protocol JoinedRoomProxyProtocol: RoomProxyProtocol {
     var identityStatusChangesPublisher: CurrentValuePublisher<[IdentityStatusChange], Never> { get }
     
     var knockRequestsStatePublisher: CurrentValuePublisher<KnockRequestsState, Never> { get }
-    
+
```
pixlwave: Can we revert this?

ElementX/Sources/Services/Room/JoinedRoomProxy.swift
```diff
@@ -70,7 +70,7 @@ class JoinedRoomProxy: JoinedRoomProxyProtocol {
     var knockRequestsStatePublisher: CurrentValuePublisher<KnockRequestsState, Never> {
         knockRequestsStateSubject.asCurrentValuePublisher()
     }
-    
+
```
pixlwave: And this.

ElementX/Sources/Services/ElementCall/ElementCallService.swift
```diff
@@ -354,4 +360,43 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
...
+            MXLog.warning("Failed to fetch a joined room for the incoming call.")
+            return
+        }
+        let ownUserId = clientProxy.userID
+        
+        MXLog.info("Observe decline events for notification \(rtcNotificationID)")
+        
+        declineListenerCancellable = roomProxy
+            .callDeclineEventPublisher(notificationId: rtcNotificationID)
+            .sink { ev in
```
pixlwave: Unlike Rust, we don't have a fear of running out of characters, let's call this `event` 😄

pixlwave: Ah, also this looks like it is going to introduce a retain cycle, we need a `[weak self]` in here.

pixlwave (commented):
Thanks for the PR, looks like a neat feature 🙌

I chatted with Stefan about the new `DeclineCallbackPublisher` and we concluded that as there is only a single usage, the additional complexity probably isn't worth it. We're perfectly happy to follow the same pattern for these as we do for e.g. knocking, typing etc by subscribing and exposing a single publisher:

https://github.com/element-hq/element-x-ios/blob/df47858e7d324f06884453aed886173fd2dc7c5a/ElementX/Sources/Services/Room/JoinedRoomProxy.swift#L20-L21

https://github.com/element-hq/element-x-ios/blob/df47858e7d324f06884453aed886173fd2dc7c5a/ElementX/Sources/Services/Room/JoinedRoomProxy.swift#L59-L62

https://github.com/element-hq/element-x-ios/blob/df47858e7d324f06884453aed886173fd2dc7c5a/ElementX/Sources/Services/Room/JoinedRoomProxy.swift#L123

https://github.com/element-hq/element-x-ios/blob/df47858e7d324f06884453aed886173fd2dc7c5a/ElementX/Sources/Services/Room/JoinedRoomProxy.swift#L719-L735

ElementX/Sources/Services/ElementCall/ElementCallService.swift
```diff
@@ -341,17 +344,45 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
...
             }
+        
+        guard let rtcNotificationID = incomingCallID.rtcNotificationID else {
+            MXLog.warning("Decline: No RTC notification ID found for the incoming call.")
+            return
+        }
+        
+        MXLog.info("Observe decline events for notification \(rtcNotificationID)")
+        
+        let ownUserID = clientProxy.userID
```
pixlwave: Could we do like we did above and use `roomProxy.ownUserID` directly in the listener's closure? That way each subscription has an independent reference to the room proxy.

BillCarsonFr: oh right, good idea. I missed this, if only xcode knew how to auto-complete that
image

pixlwave: One day we'll get reliable auto-completion… One day.

PR #4506, opened 2025-09-12
Fix traling new lines appearing in attributed string when sending text separated by 2 new lines

stefanceriu:
As defined in this test which we don't want https://github.com/ruma/ruma/blob/c3dc6de3e03b2ca131eab889a9d310ef160b95ac/crates/ruma-events/src/room/message.rs#L962


stefanceriu:
Turns out we have some 😅  `Number of failures | 3  `

PR #4508, opened 2025-09-12
Show a modal dialog while inviting people to rooms

stefanceriu:


ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -1279,9 +1279,19 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
     }
     
     private func inviteUsers(_ users: [String], in room: JoinedRoomProxyProtocol) {
```
pixlwave, line 1281: Urgh, another screen that makes the flow coordinator do the work of the view model 😅

stefanceriu: Yeh, I had a bunch of fun tracking that upwords 10 levels. Couldn't be bothered to redo it though especially as would've had to pass the app settings in.

PR #4516, opened 2025-09-16
RoomAvatar in thread timeline

### LINKED ISSUE #4515: [Task] Add the room avatar in the thread screen nav bar

Velin92:
also added a fix for the erroneous reply in thread action while in a thread, and also discovered that we can lead align the principal action again on iOS 26 (so it was a beta issue)

fixes #4515 

Simulator Screenshot - iPhone 17 Pro - 2025-09-16 at 17 16 56

### DIFF
--- a/ElementX/Sources/Screens/ThreadTimelineScreen/View/ThreadTimelineScreen.swift
+++ b/ElementX/Sources/Screens/ThreadTimelineScreen/View/ThreadTimelineScreen.swift
@@ -67,9 +67,6 @@ struct ThreadTimelineScreen: View {
                 // Using a button stops it from getting truncated in the navigation bar
                 .contentShape(.rect)
         }
-        if #available(iOS 26, *) {
-            ToolbarSpacer()
-        }
     }
     
     private var isAtBottomAndLive: Bool {
--- a/ElementX/Sources/Other/SwiftUI/Views/RoomHeaderView.swift
+++ b/ElementX/Sources/Other/SwiftUI/Views/RoomHeaderView.swift
@@ -69,13 +69,18 @@ struct RoomHeaderView_Previews: PreviewProvider, TestablePreview {
             makeHeader(avatarURL: .mockMXCAvatar, verificationState: .notVerified)
             makeHeader(avatarURL: .mockMXCAvatar, verificationState: .verified)
             makeHeader(avatarURL: .mockMXCAvatar, verificationState: .verificationViolation)
+            makeHeader(avatarURL: .mockMXCAvatar,
+                       roomSubtitle: "Subtitle",
+                       verificationState: .verified)
         }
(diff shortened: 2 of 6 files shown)


ElementX/Sources/Screens/ThreadTimelineScreen/View/ThreadTimelineScreen.swift
```diff
@@ -53,6 +54,24 @@ struct ThreadTimelineScreen: View {
...
+                           roomSubtitle: context.viewState.roomTitle,
+                           roomAvatar: context.viewState.roomAvatar,
+                           dmRecipientVerificationState: context.viewState.dmRecipientVerificationState,
+                           mediaProvider: context.mediaProvider)
+                // Using a button stops it from getting truncated in the navigation bar
+                .contentShape(.rect)
+        }
+        if #available(iOS 26, *) {
+            ToolbarSpacer()
+        }
```
pixlwave: By the sounds of things this didn't work for you, so we can probably remove it now?

ElementX/Sources/Screens/ThreadTimelineScreen/ThreadTimelineScreenViewModel.swift
```diff
@@ -12,16 +12,20 @@ typealias ThreadTimelineScreenViewModelType = StateStoreViewModel<ThreadTimeline
...
     private let actionsSubject: PassthroughSubject<ThreadTimelineScreenViewModelAction, Never> = .init()
     var actionsPublisher: AnyPublisher<ThreadTimelineScreenViewModelAction, Never> {
         actionsSubject.eraseToAnyPublisher()
     }
     
-    init(roomProxy: JoinedRoomProxyProtocol) {
+    init(roomProxy: JoinedRoomProxyProtocol,
+         userSession: UserSessionProtocol) {
         self.roomProxy = roomProxy
+        self.userSession = userSession
```
pixlwave, line 25: FWIW, when I did the refactor I only stored the `clientProxy` as the media provider isn't really needed in the view model after init.

pixlwave (approved):
LGTM!

ElementX/Sources/Screens/ThreadTimelineScreen/ThreadTimelineScreenViewModel.swift
Velin92: I need it now because I have the RoomHeaderView in the screen

pixlwave: Sure, I was just noting that you access it via `context.mediaProvider` which is set in this init, so if you wanted to you could just store `userSession.clientProxy` here and everything would still work.

PR #4520, opened 2025-09-18
Implement link previews for text messages using Apple's LinkPreview framework

stefanceriu:
This PR adds link preview bubbles underneath text messages containing links through Apple's own LinkPreview framework. 

In order to do so it looks through link attributes within the text (that are not also matrix attributes i.e. matrix.to links) and makes `LPMetadataProvider` requests to fetch metadata and render it accordingly, underneath the mesage in question.

Link preview metadata is only stored in memory within the `LinkMetadataProvider` which exists throughout the application's lifetime. This is helpful in reusing the data and not having to fetch it again over the network. 
NSSecureCoding is available so that migth be a future optimisation we can do.

Link previews will follow the user's timeline media visibility settings and will not trigger in public rooms if configured as such.

The feature is currently behind a flag as it's esentially making network requests so the device's IP address and other agent metadata can be leaked to malicious hosts. Whether Apple is obfuscating these requests or not is unclear as the documentation doesn't explicitely say so. Reports point to iMessage doing it but can't be sure in our usecase.

Screenshot 2025-09-18 at 16 46 39

Screenshot 2025-09-18 at 16 46 47


ElementX/Sources/Application/FlowCoordinatorProtocol.swift
pixlwave, line 1: I love it when a plan comes together :D

ElementX/Sources/Other/SwiftUI/Views/LinkPreviewView.swift
```diff
@@ -0,0 +1,60 @@
...
+    
+    func sizeThatFits(_ proposal: ProposedViewSize, uiView: LPLinkView, context: Context) -> CGSize? {
+        let width = proposal.width ?? uiView.intrinsicContentSize.width
+        let bestFit = uiView.sizeThatFits(CGSize(width: width, height: .greatestFiniteMagnitude))
+        
+        return CGSize(width: width, height: bestFit.height)
+    }
+}
+
+import Combine
```
pixlwave: Think this one can go now

ElementX/Sources/Other/SwiftUI/Views/LinkPreviewView.swift
```diff
@@ -0,0 +1,60 @@
...
+
+struct LinkPreviewView_Previews: PreviewProvider {
```
pixlwave, line 39: Nice preview 😎

ElementX/Sources/Services/LinkMetadata/LinkMetadataProvider.swift
```diff
@@ -0,0 +1,27 @@
...
+            return .success(item)
+        }
+        
+        do {
+            let metadata = try await LPMetadataProvider().startFetchingMetadata(for: url)
+            let item = LinkMetadataProviderItem(url: url, metadata: metadata)
+            metadataItems[url] = item
+            return .success(item)
+        } catch {
+            return .failure(error)
```
pixlwave, line 24: I guess we don't log the error because it might leak room contents, but maybe we should log that it failed? Or will that cause a bunch of unnecessary noise?

ElementX/Sources/Services/Timeline/TimelineItems/Items/Messages/TextRoomTimelineItem.swift
```diff
@@ -28,4 +29,28 @@ struct TextRoomTimelineItem: TextBasedRoomTimelineItem, Equatable {
...
+                  run.elementX.roomAlias == nil,
+                  run.elementX.roomID == nil,
+                  run.elementX.userID == nil else {
+                return nil
+            }
+            
+            return run.link
+        }
+        
+        return Array(OrderedSet(links))
```
pixlwave: I think you can just return `links.uniqued()` here unless I'm missing something?

ElementX/Sources/UITests/UITestsAppCoordinator.swift
pixlwave, line 1: Booooring!

PreviewTests/Sources/__Snapshots__/PreviewTests/textRoomTimelineView.Bubble-RTL-iPad-en-GB.png
pixlwave, line 1: Is it worth reverting all of these given they're not previewing the link in the end?

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/TextRoomTimelineView.swift
```diff
@@ -8,19 +8,61 @@
 import Foundation
 import SwiftUI
 
+import OrderedCollections
+
 struct TextRoomTimelineView: View, TextBasedRoomTimelineViewProtocol {
+    static let maxLinkPreviewsToRender = 2
```
pixlwave, line 14: FWIW, I think the old app does 3.

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/TextRoomTimelineView.swift
```diff
@@ -8,19 +8,61 @@
...
+                    for url in timelineItem.links.prefix(Self.maxLinkPreviewsToRender) {
+                        if case let .success(metadata) = await context?.viewState.linkMetadataProvider?.fetchMetadataFor(url: url) {
+                            linkMetadata[url] = metadata
+                        }
+                    }
+                }
+            }
+    }
+    
+    private var content: some View {
```
pixlwave: It would be nicer here to move the contents of the task into `func  fetchLinkPreviews` and splitting the body into a content would be necessary as it's only a `.task(fetchLinkPreviews)` as the last modifier.

Additionally, let's be parallel and use a task group for the loading?

ElementX/Sources/Services/LinkMetadata/LinkMetadataProvider.swift
stefanceriu: I figured it's a best effort kind of thing and if they fail we still show a simple version of the preview. You're also right about the noise.

ElementX/Sources/Services/Timeline/TimelineItems/Items/Messages/TextRoomTimelineItem.swift
stefanceriu: I was wondering why I haven't heard of this before but it's from the Algorithms package. TIL, good stuff 👍 

Otherwise, I still need to cast it to an array before returning so overall not a terribly great improvement.

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/TextRoomTimelineView.swift
stefanceriu: Some of them felt super tall to me so I figured 2 is enough to start with. I think we can grow from there if people take them well.

ElementX/Sources/UITests/UITestsAppCoordinator.swift
stefanceriu: 🥱

PR #4521, opened 2025-09-18
MSC4075 Use expirationTS to define the call ringing window

BillCarsonFr:
### Pull Request Checklist

As per MSC4075

> Notify events that are older than lifetime are ignored (using the sender_ts timestamp).
Otherwise a client syncing for the first time would ring for outdated call events. In general ringing only makes sense in "real time". Any client which is not able to receive the event in the lifetime period should not ring to prohibit annoying/misleading/irrelevant/outdated rings.

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Services/ElementCall/ElementCallService.swift
+++ b/ElementX/Sources/Services/ElementCall/ElementCallService.swift
@@ -163,6 +163,19 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
         let callID = CallID(callKitID: UUID(), roomID: roomID, rtcNotificationID: rtcNotificationID)
         incomingCallID = callID
         
+        guard let expirationTimestamp = payload.dictionaryPayload[ElementCallServiceNotificationKey.expirationTimestampMillis.rawValue] as? UInt64 else {
+            MXLog.error("Something went wrong, missing expiration timestamp for incoming voip call: \(payload)")
+            return
+        }
+        let nowTimestampMillis = UInt64(Date().timeIntervalSince1970 * 1000)
+        
+        guard nowTimestampMillis > expirationTimestamp else {
+            MXLog.warning("Call expired for room \(roomID), ignoring incoming push")
+            return
+        }
+        
+        let ringDurationMillis = min(expirationTimestamp - nowTimestampMillis, 90000)
+        
         let roomDisplayName = payload.dictionaryPayload[ElementCallServiceNotificationKey.roomDisplayName.rawValue] as? String
         
         let update = CXCallUpdate()
@@ -182,7 +195,7 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
         }
         
         endUnansweredCallTask = Task { [weak self] in
-            try? await Task.sleep(for: .seconds(90))
+            try? await Task.sleep(for: .milliseconds(ringDurationMillis))
             
             guard let self, !Task.isCancelled else {
                 return
--- a/NSE/Sources/NotificationHandler.swift
+++ b/NSE/Sources/NotificationHandler.swift
@@ -124,10 +124,11 @@ class NotificationHandler {
                 }
                 
                 return .processedShouldDiscard
-            case .rtcNotification(let notificationType, _):
+            case .rtcNotification(let notificationType, let expirationTimestamp):
                 return await handleCallNotification(notificationType: notificationType,
                                                     rtcNotifyEventID: event.eventId(),
                                                     timestamp: event.timestamp(),
+                                                    expirationTimestamp: expirationTimestamp,
                                                     roomID: itemProxy.roomID,
                                                     roomDisplayName: itemProxy.roomDisplayName)
             case .callAnswer,
@@ -156,6 +157,7 @@ class NotificationHandler {
     private func handleCallNotification(notificationType: RtcNotificationType,
                                         rtcNotifyEventID: String,
                                         timestamp: Timestamp,
+                                        expirationTimestamp: Timestamp,
                                         roomID: String,
                                         roomDisplayName: String) async -> NotificationProcessingResult {
         // Handle incoming VoIP calls, show the native OS call screen
@@ -209,7 +211,8 @@ class NotificationHandler {
         
         let payload = [ElementCallServiceNotificationKey.roomID.rawValue: roomID,
                        ElementCallServiceNotificationKey.roomDisplayName.rawValue: roomDisplayName,
-                       ElementCallServiceNotificationKey.rtcNotifyEventID.rawValue: rtcNotifyEventID]
+                       ElementCallServiceNotificationKey.expirationTimestampMillis.rawValue: expirationTimestamp,
+                       ElementCallServiceNotificationKey.rtcNotifyEventID.rawValue: rtcNotifyEventID] as [String: Any]
         
         do {
             try await CXProvider.reportNewIncomingVoIPPushPayload(payload)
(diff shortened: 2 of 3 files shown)


UnitTests/Sources/ElementCallServiceTests.swift
```diff
@@ -0,0 +1,131 @@
...
+    func testIncomingCall() async {
+        setupService()
+        
+        XCTAssertFalse(callProvider.reportNewIncomingCallWithUpdateCompletionCalled)
+        
+        let expectation = XCTestExpectation(description: "Call accepted")
+        
+        let pkPushPayloadMock = PKPushPayloadMock().addSeconds(currentDate, lifetime: 30)
+        
+        service.pushRegistry(pushRegistry, didReceiveIncomingPushWith: pkPushPayloadMock, for: .voIP) {
```
stefanceriu, line 32: I would've moved everything in `ElementCallService.pushRegistry(didReceiveIncomingPushWith)` to a separate method taking just they payload's dictionary. We don't actually need anything else and there's no reason to involve CallKit in the unit tests.

ElementX/Sources/Services/ElementCall/ElementCallService.swift
```diff
@@ -71,15 +65,34 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
     
     private var declineListenerHandle: TaskHandle?
     
-    override init() {
+    init(callProvider: CXProviderProtocol? = nil, timeClock: Time? = nil) {
         pushRegistry = PKPushRegistry(queue: nil)
         
+        self.timeClock = timeClock ?? Time(clock: ContinuousClock(), nowDate: Date.init)
```
stefanceriu, line 71: Why not make it the default value in the constructor parameter directly?

project.yml
```diff
@@ -101,6 +101,9 @@ packages:
   AsyncAlgorithms:
     url: https://github.com/apple/swift-async-algorithms
     minorVersion: 1.0.0
+  Clocks:
+    url: https://github.com/pointfreeco/swift-clocks
+    from: 1.0.6
```
stefanceriu, line 106: I'm not a big fan of non "native" solutions for root problems and more so of code from `pointfreeco`. We should figure out a solution that doesn't bring in a third party component for something as core as this, even if it's our own component.

In this particular case though I don't think we need any of that:
* let's have that new version of that `pushRegistry(didReceiveIncomingPushWith)` method expose a cancellation callback
* pass in a significant but small timeout in the dictionary payload
* an `expectation` that we `.fulfill()` in the callback together with `waitForExpectations`

NSE/Sources/NotificationHandler.swift
```diff
@@ -209,7 +211,8 @@ class NotificationHandler {
         
         let payload = [ElementCallServiceNotificationKey.roomID.rawValue: roomID,
                        ElementCallServiceNotificationKey.roomDisplayName.rawValue: roomDisplayName,
-                       ElementCallServiceNotificationKey.rtcNotifyEventID.rawValue: rtcNotifyEventID]
+                       ElementCallServiceNotificationKey.expirationTimestampMillis.rawValue: expirationTimestamp,
```
pixlwave, line 214: Could we convert this to a `Date` here…

ElementX/Sources/Services/ElementCall/ElementCallService.swift
```diff
@@ -163,6 +176,19 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
...
+            return
+        }
+        let nowTimestampMillis = UInt64(timeClock.nowDate().timeIntervalSince1970 * 1000)
+        
+        guard nowTimestampMillis < expirationTimestamp else {
+            MXLog.warning("Call expired for room \(roomID), ignoring incoming push")
+            return
+        }
+        
+        let ringDurationMillis = min(expirationTimestamp - nowTimestampMillis, 90000)
```
pixlwave, line 190: And then handle all of this using `Date` comparisons rather than `millis` so everything is typed properly.

UnitTests/Sources/ElementCallServiceTests.swift
pixlwave: > We don't actually need anything else and there's no reason to involve CallKit in the unit tests.

Given the intention to test the timeouts, I think we need to involve CallKit here to assert the timeout properly?

ElementX/Sources/Services/ElementCall/ElementCallService.swift
```diff
@@ -13,6 +13,12 @@ import MatrixRustSDK
 import PushKit
 import UIKit
 
+// Keep this class testable
+struct Time {
+    var clock: any Clock<Duration>
+    var nowDate: () -> Date
```
pixlwave, line 19: I think we can call this `now`, so it aligns with `Date.now`, `Clock.continuous.now` etc.

ElementX/Sources/Services/ElementCall/ElementCallService.swift
```diff
@@ -22,20 +28,8 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
...
-            configuration.iconTemplateImageData = callKitIcon.pngData()
-        }
-        
-        // https://stackoverflow.com/a/46077628/730924
-        configuration.supportedHandleTypes = [.generic]
-        
-        return CXProvider(configuration: configuration)
-    }()
+    private let callProvider: CXProviderProtocol
+    private let timeClock: Time
```
pixlwave, line 32: And maybe align the names of these. so `time: Time` or maybe `timeProvider: TimeProvider`.

ElementX/Sources/Services/ElementCall/ElementCallService.swift
pixlwave: It would be nice if the struct properties have default values too. Then the constructor could be `timeClock: Time = Time()`

project.yml
pixlwave: > I'm not a big fan of non "native" solutions for root problems and more so of code from pointfreeco. We should figure out a solution that doesn't bring in a third party component for something as core as this, even if it's our own component.

Tbh, I think it would be nice to have this kind of testable clock infrastructure. I'd be happy if we implemented it ourselves but let's leave that as an exercise for us though, once this is merged?

UnitTests/Sources/ElementCallServiceTests.swift
```diff
@@ -0,0 +1,131 @@
...
+        dict[ElementCallServiceNotificationKey.roomDisplayName.rawValue] = "welcome"
+        dict[ElementCallServiceNotificationKey.rtcNotifyEventID.rawValue] = "$000"
+        dict[ElementCallServiceNotificationKey.expirationTimestampMillis.rawValue] = 10
+    }
+    
+    override var dictionaryPayload: [AnyHashable: Any] {
+        dict
+    }
+    
+    func addSeconds(_ from: Date, lifetime: Int) -> Self {
```
pixlwave, line 127: Not important as it's a test, but it would help to follow the code if this was called something like `withExpiration` seeing as it doesn't add the seconds to the current timeout.

Also, we should use `TimeInterval` here (or `Duration`) rather than `Int` for clarity.

pixlwave (commented):
Chatted with Stefan about this a bit more. I think we're happy over all with introducing the clocks library for now and we can implement something ourselves later to avoid the dependency.

Just a few comments below 👇

BillCarsonFr:
closed in favor of https://github.com/element-hq/element-x-ios/pull/4652 due to a lot of changes on develop

PR #4522, opened 2025-09-18
Update the SDK and Element Call.

pixlwave:
- The SDK includes the `is_direct` change for space rooms.
- Element Call is an RC for the Nightly.
- Also fixes a bug where there was no way to navigate out of a call link after hanging up.


pixlwave:
I'm going to yolo merge based on the tests (and that it's minimal changes) ready for the Nightly.

PR #4523, opened 2025-09-18
Adjust project and CI workflows to work on Xcode 26

stefanceriu:
tl;dr; 
* any workflow running a simulator needs an xlarge runner otherwise it times out
* alert buttons appear as duplicated
* ASWebAuthenticationSession runs its own XCUITest "app"
* Preview snapshots are **a mess** but will improve with https://github.com/element-hq/element-x-ios/pull/4498

UI tests run here https://github.com/element-hq/element-x-ios/actions/runs/17918959873

Integration tests run https://github.com/element-hq/element-x-ios/actions/runs/17921836084


IntegrationTests/Sources/UserFlowTests.swift
```diff
@@ -263,7 +263,7 @@ class UserFlowTests: XCTestCase {
     /// When there are multiple buttons with the same label in the hierarchy, all the buttons we created
     /// should have an identifier set, and so this method will ignore those picking the one with only a label.
     private func tapOnBackButton(_ label: String = "Back") {
-        let button = app.buttons.matching(NSPredicate(format: "label == %@ && identifier == ''", label)).firstMatch
+        let button = app.buttons.matching(NSPredicate(format: "label == %@ && identifier == 'BackButton'", label)).firstMatch
```
pixlwave, line 266: Probably not important, but if `BackButton` is now the identifier, we could likely remove the `label` parameter entirely here.

pixlwave (commented):
Very nice, I haven't particularly looked at the snapshots as I don't think they're really valid without Mauro's changes anyway. Maybe we should merge

IntegrationTests/Sources/UserFlowTests.swift
stefanceriu: I admit this is a bit overzealous but I couldn't be bothered

PR #4527, opened 2025-09-19
Update Element Call to RC3

pixlwave:
Fixes a decline bug.

### DIFF
--- a/ElementX.xcodeproj/project.pbxproj
+++ b/ElementX.xcodeproj/project.pbxproj
@@ -9358,7 +9358,7 @@
 			repositoryURL = "https://github.com/element-hq/element-call-swift";
 			requirement = {
 				kind = exactVersion;
-				version = "0.16.0-rc.2";
+				version = "0.16.0-rc.3";
 			};
 		};
 		821C67C9A7F8CC3FD41B28B4 /* XCRemoteSwiftPackageReference "emojibase-bindings" */ = {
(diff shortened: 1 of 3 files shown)


ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -89,10 +89,12 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
         self.roomList = roomList
         
         do {
-            listUpdatesSubscriptionResult = roomList.entriesWithDynamicAdapters(pageSize: UInt32(roomListPageSize), listener: SDKListener { [weak self] updates in
-                guard let self else { return }
-                diffsPublisher.send(updates)
-            })
+            listUpdatesSubscriptionResult = roomList.entriesWithDynamicAdaptersWith(pageSize: UInt32(roomListPageSize),
+                                                                                    enableLatestEventSorter: appSettings.lowPriorityFilterEnabled,
```
marwing, line 93: This appears to be reading the wrong setting?
```suggestion
                                                                                    enableLatestEventSorter: appSettings.latestEventSorterEnabled,
```

stefanceriu: Oupsie, thanks for that! 🙇‍♂️ Luckily they're both disabled by default.

PR #4545, opened 2025-09-26
Show space invites in the room list.

pixlwave:
This only allows them to be shown, accepting the invites isn't going to show the `SpaceScreen` yet, it'll simply be treated as a room and show it will show the space's timeline.


ElementX/Sources/Screens/HomeScreen/View/HomeScreenInviteCell.swift
```diff
@@ -223,7 +235,10 @@ private extension HomeScreenRoom {
         return .init(summary: summary, hideUnreadMessagesBadge: false)
     }
     
-    static func roomInvite(alias: String? = nil, avatarURL: URL? = nil) -> HomeScreenRoom {
+    static func roomInvite(name: String = "Awesome Room",
```
stefanceriu: Should we maybe name this `inviteRoom` instead? 🤔

ElementX/Sources/Services/Room/RoomSummary/RoomEventStringBuilder.swift
```diff
@@ -55,6 +55,8 @@ struct RoomEventStringBuilder {
                 default: L10n.commonWaitingForDecryptionKey
                 }
                 return prefix(errorMessage, with: displayName, isOutgoing: isOutgoing)
+            case .other(eventType: let eventType):
+                return nil // We shouldn't receive these without asking for custom event types.
```
stefanceriu, line 59: 👍

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -118,19 +118,26 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
     }
     
     func setFilter(_ filter: RoomSummaryProviderFilter) {
+        let baseFilter: [RoomListEntriesDynamicFilterKind] = if appSettings.spacesEnabled {
+            [.any(filters: [.all(filters: [.nonSpace, .nonLeft]),
+                            .all(filters: [.space, .invite])]),
```
stefanceriu, line 123: Cool 😎

ElementX/Sources/Screens/HomeScreen/View/HomeScreenInviteCell.swift
pixlwave: Can do, although when I see that I read it as an action, like join room.

stefanceriu: What if we just just drop `room` altogether? 😁

pixlwave: wfm!

pixlwave:
Those snapshot failures are fixed in #4553.

PR #4554, opened 2025-09-29
Accept space invites from a home screen cell.

pixlwave:
This PR supports pushing a space from the room list when tapping the Accept button on a space invite.

Depends on #4545 

https://github.com/user-attachments/assets/7bf364eb-eaf0-4b1a-8d61-19c78275c03c


UnitTests/Sources/HomeScreenViewModelTests.swift
```diff
@@ -390,3 +428,42 @@ private extension [HomeScreenRoom] {
         }
     }
 }
+
+extension HomeScreenViewModelAction: @retroactive Equatable {
```
stefanceriu, line 432: 🥱

stefanceriu: Would it have worked if we just implemented equatable on the space proxies?

ElementX/Sources/Screens/HomeScreen/HomeScreenCoordinator.swift
```diff
@@ -67,6 +68,10 @@ final class HomeScreenCoordinator: CoordinatorProtocol {
                     actionsSubject.send(.presentRoomDetails(roomIdentifier: roomIdentifier))
                 case .presentReportRoom(let roomIdentifier):
                     actionsSubject.send(.presentReportRoom(roomIdentifier: roomIdentifier))
+                case .presentDeclineAndBlock(let userID, let roomID):
+                    actionsSubject.send(.presentDeclineAndBlock(userID: userID, roomID: roomID))
```
stefanceriu, line 72: 🧹

ElementX/Sources/FlowCoordinators/ChatsFlowCoordinator.swift
```diff
@@ -178,12 +180,15 @@ class ChatsFlowCoordinator: FlowCoordinatorProtocol {
...
+            let userInfo = context.userInfo as? ChatsFlowCoordinatorStateMachine.EventUserInfo
+            let animated = userInfo?.animated ?? true
+            
             switch (context.fromState, context.event, context.toState) {
             case (.initial, .start, .roomList):
                 presentHomeScreen()
-            case(.roomList(let roomListSelectedRoomID), .selectRoom(let roomID, let via, let entryPoint), .roomList):
-                if roomListSelectedRoomID == roomID,
+            case(.roomList(let detailState), .selectRoom(let roomID, let via, let entryPoint), .roomList):
+                if case .room(roomID) = detailState,
```
stefanceriu, line 191: Would a `switch` maybe tidy this up better? (almost wrote `match` 🤦‍♂️)

ElementX/Sources/FlowCoordinators/ChatsFlowCoordinator.swift
```diff
@@ -491,6 +507,43 @@ class ChatsFlowCoordinator: FlowCoordinatorProtocol {
...
+        
+        if navigationSplitCoordinator.detailCoordinator !== detailNavigationStackCoordinator {
+            navigationSplitCoordinator.setDetailCoordinator(detailNavigationStackCoordinator, animated: animated)
+        }
+        
+        coordinator.start()
+    }
+    
+    private func dismissSpaceFlow(animated: Bool) {
+        // Based on dismissRoomFlow, past me was very insistent that this must happen after the flow has tidied the stack 😅.
```
stefanceriu, line 542: I say go all in and sign yourself on both 😀 e.g. `(Doug 30.09.2025)`

ElementX/Sources/FlowCoordinators/ChatsFlowCoordinator.swift
pixlwave: Ah, that sounded nice but I don't think it really helps much as the `else` also takes the `.room` case when those other conditions aren't met.

UnitTests/Sources/HomeScreenViewModelTests.swift
pixlwave: Yes but as we pass them round as the protocol, you can't add it retroactively and
a) I'd rather not put that conformance in the app as it isn't technically correct just checking the `id`
b) Our mocks don't have a clue how to handle it
c) I have a feeling it adds a requirement to write `any SpaceRoomListProxyProtocol` everywhere.

stefanceriu: I keep forgetting about stupid protocol conformances. It makes perfect sense to me, thanks for the explanation 👍

PR #4555, opened 2025-09-29
Update acknowledgments with resolved names.

pixlwave:
We were hitting https://github.com/FelixHerrmann/swift-package-list/issues/117 however it seems since updating to Xcode 26, resetting the package caches no longer wipes out the resolved names.


pixlwave:
CI is basically broken and this doesn't touch the app anyway, so going to merge.

PR #4565, opened 2025-10-01
Permalink from/to threads

### LINKED ISSUE #4546: [Task] Threads: Permalink to a threaded message

Velin92:
fixes #4546 

https://github.com/user-attachments/assets/44fbebf2-594e-4cb1-92c3-1592e90c558b

This PR allows to open permalinks from/to threads, specifically...
- If the permalink is opened in a thread, and belongs to the same thread, will just focus directly on the item itself: Room -> Thread -> focus on event
- If the permalink is opened in a thread, and belongs to another thread, it will open a new thread and will just focus directly on the item itself (without updating the room original focus, to keep a consistent navigation): Room -> Thread  1 -> Thread  2-> focus on event
- if the permalink is opened in a thread, and belongs to the the same room but is not a threaded message, it will open the same room again and focus the event: Room 1 -> Thread 1 -> Room 1 -> focus on event
- if the permalink is opened anywhere and belongs to another room, it will open the new room, focus the thread root and then open the thread and focus on the event: Any Room/Thread 1 -> Room 2 -> focus on root event -> Thread 2 -> focus on event

Also this PR allows us to check if the event id exists or can be fetched or not, which allows us to throw errors in case of an invalid event or if there are issues in getting it, while still navigating to the appropriate room.

### DIFF
--- a/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
@@ -68,6 +68,7 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
     private var roomProxy: JoinedRoomProxyProtocol!
     
     private var roomScreenCoordinator: RoomScreenCoordinator?
+    private var childThreads: [ThreadTimelineScreenCoordinator] = []
     private weak var joinRoomScreenCoordinator: JoinRoomScreenCoordinator?
     
     // periphery:ignore - used to avoid deallocation
@@ -163,7 +164,28 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
             } else if roomID != roomProxy.id {
                 stateMachine.tryEvent(.startChildFlow(roomID: roomID, via: via, entryPoint: .eventID(eventID)), userInfo: EventUserInfo(animated: animated))
             } else {
-                roomScreenCoordinator?.focusOnEvent(.init(eventID: eventID, shouldSetPin: false))
+                showLoadingIndicator(delay: .seconds(0.5))
+                Task {
+                    defer { hideLoadingIndicator() }
+                    switch await roomProxy.loadOrFetchEvent(for: eventID) {
+                    case .success(let event):
+                        if flowParameters.appSettings.threadsEnabled, let threadRootEventID = event.threadRootEventId() {
+                            if case .thread(threadRootEventID: threadRootEventID, _) = stateMachine.state, let threadCoordinator = childThreads.last {
+                                threadCoordinator.focusOnEvent(eventID: eventID)
+                            } else {
+                                stateMachine.tryEvent(.presentThread(threadRootEventID: threadRootEventID, focusEventID: eventID))
+                            }
+                        } else if !childThreads.isEmpty {
+                            // If we are showing a child thread, and we are navigating to a non threaded event
+                            // of the same room, we want to push the room on top of the thread
+                            stateMachine.tryEvent(.startChildFlow(roomID: roomID, via: via, entryPoint: .eventID(eventID)), userInfo: EventUserInfo(animated: animated))
+                        } else {
+                            roomScreenCoordinator?.focusOnEvent(.init(eventID: eventID, shouldSetPin: false))
+                        }
+                    case .failure:
+                        showError()
+                    }
+                }
             }
         case .share(let payload):
             guard let roomID = payload.roomID, roomID == self.roomID else {
@@ -232,12 +254,27 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
         switch room {
         case .joined(let roomProxy):
             await storeAndSubscribeToRoomProxy(roomProxy)
+            if case let .eventFocus(focusEvent) = presentationAction {
+                switch await roomProxy.loadOrFetchEvent(for: focusEvent.eventID) {
+                case .success(let event):
+                    guard flowParameters.appSettings.threadsEnabled, let threadRootEventID = event.threadRootEventId() else {
+                        break
+                    }
+                    stateMachine.tryEvent(.presentRoom(presentationAction: nil), userInfo: EventUserInfo(animated: animated))
+                    stateMachine.tryEvent(.presentThread(threadRootEventID: threadRootEventID, focusEventID: focusEvent.eventID), userInfo: EventUserInfo(animated: false))
+                    return
+                case .failure:
+                    showError()
+                    stateMachine.tryEvent(.presentRoom(presentationAction: nil), userInfo: EventUserInfo(animated: animated))
+                    return
+                }
+            }
             stateMachine.tryEvent(.presentRoom(presentationAction: presentationAction), userInfo: EventUserInfo(animated: animated))
         default:
             stateMachine.tryEvent(.presentJoinRoomScreen(via: via), userInfo: EventUserInfo(animated: animated))
         }
     }
-
+    
     func clearRoute(animated: Bool) {
         guard stateMachine.state != .initial else {
             return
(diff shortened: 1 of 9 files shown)


ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -65,6 +65,7 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
     private var roomProxy: JoinedRoomProxyProtocol!
     
     private var roomScreenCoordinator: RoomScreenCoordinator?
+    private var childThreads: [ThreadTimelineScreenCoordinator] = []
```
pixlwave: This should probably be a bit more descriptive and match the others like `childThreadScreenCoordinators`.

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -1560,4 +1605,8 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
     private func hideLoadingIndicator() {
         flowParameters.userIndicatorController.retractIndicatorWithId(Self.loadingIndicatorID)
     }
+    
+    private func showError() {
```
pixlwave:
```suggestion
    private func showErrorIndicator() {
```

ElementX/Sources/Mocks/ClientProxyMock.swift
```diff
@@ -96,11 +96,15 @@ extension ClientProxyMock {
...
             if let room = self?.roomSummaryProvider.roomListPublisher.value.first(where: { $0.id == identifier }) {
-                await .joined(JoinedRoomProxyMock(.init(id: room.id, name: room.name)))
+                let roomProxy = await JoinedRoomProxyMock(.init(id: room.id, name: room.name))
+                roomProxy.loadOrFetchEventForReturnValue = .success(TimelineEventSDKMock())
+                return .joined(roomProxy)
             } else if let spaceRoomProxy = configuration.joinedSpaceRooms.first(where: { $0.id == identifier }) {
-                await .joined(JoinedRoomProxyMock(.init(id: spaceRoomProxy.id, name: spaceRoomProxy.name)))
+                let roomProxy = await JoinedRoomProxyMock(.init(id: spaceRoomProxy.id, name: spaceRoomProxy.name))
+                roomProxy.loadOrFetchEventForReturnValue = .success(TimelineEventSDKMock())
+                return .joined(roomProxy)
```
pixlwave, line 105: Rather than doing this here, should we modify the `JoinedRoomProxyMock`'s convenience init to always return that success, with a parameter to customise it if necessary?

ElementX/Sources/Services/Room/RoomProxyProtocol.swift
```diff
@@ -97,7 +97,7 @@ protocol JoinedRoomProxyProtocol: RoomProxyProtocol {
     
     func threadTimeline(eventID: String) async -> Result<TimelineProxyProtocol, RoomProxyError>
     
-    func getThreadRootEventID(for eventID: String) async -> Result<String?, RoomProxyError>
+    func loadOrFetchEvent(for eventID: String) async -> Result<TimelineEvent, RoomProxyError>
```
pixlwave: Now I've seen what this returns, I think it would be nice to call this `loadOrFetchEventDetails` as you don't actually get the event's contents. WDYT?

UnitTests/Sources/RoomFlowCoordinatorTests.swift
pixlwave, line 1: Very nice 👌

UnitTests/Sources/RoomFlowCoordinatorTests.swift
```diff
@@ -292,7 +377,7 @@ class RoomFlowCoordinatorTests: XCTestCase {
         
         try await fulfillment.fulfill()
     }
-    
+
```
pixlwave: Probably don't need this one :)

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -160,7 +161,32 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
...
+                                threadCoordinator.focusOnEvent(eventID: eventID)
+                            } else {
+                                // If we are showing the room timeline, we want to focus the thread root
+                                if childThreads.isEmpty {
+                                    roomScreenCoordinator?.focusOnEvent(.init(eventID: threadRootEventID, shouldSetPin: false))
+                                }
+                                stateMachine.tryEvent(.presentThread(threadRootEventID: threadRootEventID, focusEventID: eventID))
+                            }
+                        } else if !childThreads.isEmpty {
+                            // If we are showing a child thread, and we are navigating to a non threaded event
```
stefanceriu:
```suggestion
                            // If we are showing a child thread and we are navigating to a non threaded event
```

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -229,12 +255,27 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
...
+                    stateMachine.tryEvent(.presentRoom(presentationAction: .eventFocus(.init(eventID: threadRootEventID, shouldSetPin: false))), userInfo: EventUserInfo(animated: animated))
+                    stateMachine.tryEvent(.presentThread(threadRootEventID: threadRootEventID, focusEventID: focusEvent.eventID), userInfo: EventUserInfo(animated: false))
+                    return
+                case .failure:
+                    showError()
+                    stateMachine.tryEvent(.presentRoom(presentationAction: nil), userInfo: EventUserInfo(animated: animated))
+                    return
+                }
+            }
             stateMachine.tryEvent(.presentRoom(presentationAction: presentationAction), userInfo: EventUserInfo(animated: animated))
```
stefanceriu: You sure we need another `.presentRoom` here? 🤔

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -160,7 +161,32 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
...
+                            } else {
+                                // If we are showing the room timeline, we want to focus the thread root
+                                if childThreads.isEmpty {
+                                    roomScreenCoordinator?.focusOnEvent(.init(eventID: threadRootEventID, shouldSetPin: false))
+                                }
+                                stateMachine.tryEvent(.presentThread(threadRootEventID: threadRootEventID, focusEventID: eventID))
+                            }
+                        } else if !childThreads.isEmpty {
+                            // If we are showing a child thread, and we are navigating to a non threaded event
+                            // of the same room, we want to push the room on top of the thread
```
stefanceriu: Wasn't expecting that tbh

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
Velin92: yeah this is generic case for non thread handling, however I think the code here doesn't make it clear I guess I'll rewrite this so that is all included in the switch flow

ElementX/Sources/Mocks/ClientProxyMock.swift
Velin92: I thought about it, and since this function is called by also all the other tests for all the other routes, I guess having here the generic implementation to always succeed with a non threaded event is fine, I feel the thread case is just a specific test case that we can override in the mock directly per test basis. For example in my test, I need to do that multiple time to test the navigation stack, and I needed to mutate the response thrice, which would not work with the configuration since it's only done once.

PR #4569, opened 2025-10-02
A couple of small tweaks.

pixlwave:
- Makes suggestions visible again on iPad (and Mac)
    - They were hidden behind the sidebar.
- Fixes a regression introduced in #4554 where the room list would keep the highlight after popping a room.
- Uses our Backports.swift file for more modifiers.
- Fixes the silly header bug on macOS by not leading aligning it.
- Silences some warnings in the project (there's some more in Compound which I haven't done yet as they might involve snapshot updates).


ElementX/Sources/Screens/RoomScreen/ComposerToolbar/View/CompletionSuggestionView.swift
```diff
@@ -99,7 +99,6 @@ private struct BackgroundView<Content: View>: View {
             .clipShape(RoundedCornerShape(radius: shadowRadius, corners: [.topLeft, .topRight]))
             .shadow(color: .black.opacity(0.20), radius: 20.0, x: 0.0, y: 3.0)
             .mask(Rectangle().padding(.init(top: -(shadowRadius * 2), leading: 0.0, bottom: 0.0, trailing: 0.0)))
-            .edgesIgnoringSafeArea(.all)
```
pixlwave, line 102: I can't find any layout where this is actually needed, but happy to tweak it a bit more if I missed something.

ElementX/Sources/Other/SwiftUI/Backports.swift
stefanceriu, line 1: Noice!

stefanceriu (approved):
🚀

Velin92 (approved):

PR #4573, opened 2025-10-03
Labs screen

### LINKED ISSUE #4561: [Task] Introduce a Labs section in EX

Velin92:
fixes #4561 

This PR adds the labs screen in the settings.

Simulator Screenshot - iPhone 17 Pro - 2025-10-03 at 18 18 31

### DIFF
--- a/ElementX/Sources/Screens/Settings/SettingsScreen/View/SettingsScreen.swift
+++ b/ElementX/Sources/Screens/Settings/SettingsScreen/View/SettingsScreen.swift
@@ -137,6 +137,19 @@ struct SettingsScreen: View {
     
     private var generalSection: some View {
         Section {
+            ListRow(label: .default(title: L10n.commonAdvancedSettings,
+                                    icon: \.settings),
+                    kind: .navigationLink {
+                        context.send(viewAction: .advancedSettings)
+                    })
+                    .accessibilityIdentifier(A11yIdentifiers.settingsScreen.advancedSettings)
+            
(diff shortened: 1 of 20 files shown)


Velin92:
@amshakal in Figma i noticed that the icon has a bg subtle primary color as a background, however the type is default solid, which uses bg subtle secondary as a background color. So for now I am using the default (non solid) type here since otherwise the background would be the same as the background of the view itself.

We do not have in compound a type that uses subtle primary as background for these big icons, should we make a new one then?

amshakal:
I changed the designs based on your feedback :)

ElementX/Sources/Screens/Settings/SettingsScreen/View/SettingsScreen.swift
```diff
@@ -137,6 +137,19 @@ struct SettingsScreen: View {
...
+                    kind: .navigationLink {
+                        context.send(viewAction: .advancedSettings)
+                    })
+                    .accessibilityIdentifier(A11yIdentifiers.settingsScreen.advancedSettings)
+            
+            ListRow(label: .default(title: L10n.screenAdvancedSettingsLabs,
+                                    icon: \.labs),
+                    kind: .navigationLink {
+                        context.send(viewAction: .labs)
+                    })
```
pixlwave, line 151: I think based on the [meta issue](https://github.com/element-hq/element-meta/issues/2982) we want this button to automatically hide if there aren't any settings in `LabsOptionsProtocol`.

PR #4581, opened 2025-10-06
Clear cache on changing the threads FF

Velin92:
In this way the user doesn't have the responsibility of killing and restarting the app to have the changes take effect.


ElementX/Sources/Generated/Strings.swift
```diff
@@ -1824,6 +1828,8 @@ internal enum L10n {
   internal static var screenKnockRequestsListTitle: String { return L10n.tr("Localizable", "screen_knock_requests_list_title") }
   /// Enable thread replies
```
stefanceriu, line 1829: Just noticed this: why `thread replies` when absolutely everybody understands what `threads` mean? Why complicate things?

Threads replies is so confusing, are we enabling threads or just replying in threads or replies I receive in threads. Will threads be enabled without replies if I disable this etc etc, so many unnecessary questions.

Nothing to do with this PR, just saying.

stefanceriu (approved):
lgtm 👍

PR #4583, opened 2025-10-06
New divider color for iOS 26

Velin92:
Also updated a LOT of preview tests


pixlwave (approved):
LGTM, although we're going to need to update Compound snapshots and probably UI test snapshots too.

PR #4601, opened 2025-10-08
Thread tweaks

### LINKED ISSUE #4598: Cannot load a permalink to a threaded message far in the thread
If you try to open a permalink to an old message in a thread, the timeline will display the loading spinner forever if the user does nothing. If the user back paginates up to the message, the spinner will disappear and the message will be highlighted as expected. Check the video in the associated rageshake.

"Old message in a thread" means there are around 50 more recent than the linked message in my test. In this test, the root message was "old" in the same way.

The problem occurs only after a fresh restart of the app. If the user already back paginated to his threaded message, the permalink will work well.

### LINKED ISSUE #4600: Thread permalink: Weird behavior on the main timeline after closing the thread
Let's have a thread with one reply on the last message of a room. If you open a permalink to it from another room with EXI, then close the thread, the main timeline will have 2 weird behaviors:
- A "Failed loading messages" banner
- Sometimes, the main timeline bounces forever. It does it only once in the video in the rageshake

Image

Velin92:
fixes #4600 
fixes #4598 

fixed an issue where threads would not do the initial focus and another issue where the timeline did throw an error by mistake


ElementX/Sources/Services/Timeline/TimelineProxy.swift
```diff
@@ -133,12 +133,10 @@ final class TimelineProxy: TimelineProxyProtocol {
         case .backwards: backPaginationStatusSubject
         case .forwards: forwardPaginationStatusSubject
         }
-        
+
```
pixlwave: Probably can be discarded :)

PR #4602, opened 2025-10-08
Add UI tests for accepting space invites.

pixlwave:
Tests the Chats/Room flow coordinator handling of space invites.


UITests/Sources/__Snapshots__/Application/userSessionScreen.testAcceptSpaceInvite-iPad-26-0-en-GB-4.png
stefanceriu, line 1: Man are those buttons pretty! 🤩 /s

ElementX/Sources/Mocks/ClientProxyMock.swift
```diff
@@ -67,6 +72,11 @@ extension ClientProxyMock {
         createDirectRoomWithExpectedRoomNameReturnValue = .failure(.sdkError(ClientProxyMockError.generic))
         createRoomNameTopicIsRoomPrivateIsKnockingOnlyUserIDsAvatarURLAliasLocalPartReturnValue = .failure(.sdkError(ClientProxyMockError.generic))
         canJoinRoomWithReturnValue = true
+        joinRoomViaClosure = { roomID, _ in
+            configuration.overrides.joinedRoomIDs.insert(roomID)
```
stefanceriu, line 76: I wonder if we should do this at the `RoomSummaryProviderMock` level so that its `roomListPublisher` returns correct values as well. I do appreciate that dumping a new full RoomSummary in there is harder but perhaps better in the long run?

ElementX/Sources/Mocks/ClientProxyMock.swift
```diff
@@ -67,6 +72,11 @@ extension ClientProxyMock {
...
+            return .success(())
+        }
+        joinRoomAliasReturnValue = .success(())
```
stefanceriu, line 79: It would also help with making this work ^

UITests/Sources/__Snapshots__/Application/userSessionScreen.testAcceptSpaceInvite-iPad-26-0-en-GB-4.png
pixlwave: I'll make the layout vertical when thy don't fit (in a separate PR).

ElementX/Sources/Mocks/ClientProxyMock.swift
pixlwave: Ok yeah makes sense to me, again for a follow-up.

PR #4604, opened 2025-10-08
Fix the bloom on the space tab for real.

pixlwave:
I ended up committing the wrong parameter value when refactoring #4585 🤦‍♂️

This PR actually fixes the bloom on the space tab (well on iPhone at least, on iPad it still isn't the best).


ElementX/Sources/Screens/Spaces/SpaceListScreen/View/SpaceListScreen.swift
pixlwave, line 1: This one is the fix.

PR #4607, opened 2025-10-09
Make the `SpaceRoomListProxy` publish its `SpaceRoomProxy` updates.

### LINKED ISSUE #4593: Space headers are not reactive
### Steps to reproduce

1. Open a space
2. Change the name, topic or avatar of the space from another device or user

#### What did you expect?
The space header to be updated with the latest data, like on EXA.

#### What happened instead?
Old data was displayed until the space is reopened.

Note there is the same reactive issue on space members info displayed on this screen.,

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

Nightly 1039 (2025-10-08)

### Homeserver

_No response_

### Will you send logs?

No

pixlwave:
Also removes an unused `parent` parameter that was missed in #4584.

Fixes #4593.


ElementX/Sources/Services/Spaces/SpaceRoomListProxyProtocol.swift
```diff
@@ -14,8 +14,9 @@ enum SpaceRoomListProxyError: Error {
 
 // sourcery: AutoMockable
 protocol SpaceRoomListProxyProtocol {
-    var spaceRoomProxy: SpaceRoomProxyProtocol { get }
+    var id: String { get }
```
stefanceriu, line 17: Hope this won't end up confusing people.

pixlwave: In what way?

stefanceriu: By conflating the room list ID (which isn't really a thing) with a the parent space's room ID. Not a big deal though.

PR #4610, opened 2025-10-10
Prevent the system from automatically hiding the sidebar when backgrounding the app

stefanceriu:


ElementX/Sources/Application/Navigation/NavigationCoordinators.swift
```diff
@@ -329,6 +329,7 @@ private struct NavigationSplitCoordinatorView: View {
     @State private var columnVisibility = NavigationSplitViewVisibility.all
     
     @Environment(\.horizontalSizeClass) private var horizontalSizeClass
+    @Environment(\.scenePhase) var scenePhase
```
pixlwave: Should probably be `private` but not exactly important.

PR #4611, opened 2025-10-10
Only run Compound tests when files are changed in Compound.

pixlwave:
This will reduce the number of runners we use.

### DIFF
--- a/compound-ios/Sources/Compound/List/ListRow.swift
+++ b/compound-ios/Sources/Compound/List/ListRow.swift
@@ -27,19 +27,19 @@ public struct ListRow<Icon: View, DetailsIcon: View, CustomContent: View, Select
     let label: ListRowLabel<Icon>
     let details: ListRowDetails<DetailsIcon>?
     
-    public enum Kind<CustomContent: View, SelectionValue: Hashable> {
+    public enum Kind<CustomView: View, Selection: Hashable> {
         case label
         case button(action: () -> Void)
         case navigationLink(action: () -> Void)
-        case picker(selection: Binding<SelectionValue>, items: [(title: String, tag: SelectionValue)])
+        case picker(selection: Binding<Selection>, items: [(title: String, tag: Selection)])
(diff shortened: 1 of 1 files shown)


pixlwave:
Ok so the run on the first commit didn't trigger the Compound CI.

Screenshot 2025-10-10 at 12 09 46 pm

Now I'll fix those lint errors and make sure that it runs.

pixlwave:
And now I've pushed some changes (fixing some warnings we were seeing), the Compound workflow has been triggered 🎉

Screenshot 2025-10-10 at 12 29 33 pm

.github/workflows/compound-ios.yml
```diff
@@ -6,17 +6,20 @@ name: Compound
...
 
 jobs:
   tests:
 
     runs-on: macos-15
 
     concurrency:
-      # Only allow a single run of this workflow on each branch, automatically cancelling older runs.
-      group: ${{ format('compound-tests-{0}', github.ref) }}
+      # When running on develop, use the sha to allow all runs of this workflow to run concurrently.
```
stefanceriu, line 20: Why bother though?

pixlwave: I wasn't sure if this could happen:
- Merge a PR that contains compound changes and trigger a run on `develop`
- Merge another PR that doesn't contain compound changes and cancel the existing run on `develop`.

I'm guessing not but it didn't seem worth even testing it when this would fix the issue if it did. Plus, seeing as it is probably very unlikely we would ever have 2 compound PRs back-to-back I doubt we would ever notice the extra run.

stefanceriu: I would've left it alone and just check if we _ever_ get any compound cancellation emails, which we really shouldn't.

pixlwave: Oh I forgot about emails… Ok then follow-up PR incoming :D

Issue #3164, opened 2024-08-13
Dark mode app icon


ara4n:
**Is your feature request related to a problem? Please describe.**

Apparently iOS18 supports dark mode icons.  It autogens these when it can, but EX's icon is too complex for it.  Therefore we should probably provide our own.

**Describe the solution you'd like**

Credit to `@maddie:spyhoodle.me`:

![image](https://github.com/user-attachments/assets/9ed22720-98eb-4653-ad06-cb47a4195440)

On the left you see an autogenned dark-mode classic Element icon. But on the right, there is no dark-mode icon.

**Additional context**

https://matrix.to/#/!amwlFMejXQcoCxNYHH:matrix.org/$Aj8mBeo2RFbOqkEkwiXb5t2LOK0DUrrThxhcRX3L3p8?via=matrix.org&via=element.io&via=beeper.com

jjj333-p:
Just adding context, ios 18 also allows coloring them to a theme, and based on how it treats that, I suspect its the gradient. i think if the background gradient was flattened it would be able to generate one on the fly.

![image](https://github.com/user-attachments/assets/afefd029-113c-4b0f-83ff-6ef7074c5631)

ProbablyRohan:
I would love if we were able to change the Element X icon to the one that the old Element app has as well, similar to how you can change the icon in the GitHub app. This would require a bit of extra work though.

s3ase:
Still no word on this one?

AgathaSorceress:
In iOS 26, it should be relatively trivial to make the icon adapt to any theme with Icon Composer:

<img width="424" height="303" alt="Image" src="https://github.com/user-attachments/assets/fc794ccb-7955-4f63-b147-2a39ad6b27d6" />

<img width="424" height="327" alt="Image" src="https://github.com/user-attachments/assets/63cd1412-069a-4003-9271-4b8495bdb3f8" />

<img width="424" height="329" alt="Image" src="https://github.com/user-attachments/assets/f71474a1-99f9-4be8-968d-630c263738b5" />

6c67b9:
https://github.com/element-hq/element-x-ios/pull/5343

DidiDidi129:
I'm happy to work with the element team for the logo, I have experience in using the correct software to do this

amshakal:
@JordanConnor can you look at this when you are back next week?

DidiDidi129:
[...]

PR #4612, opened 2025-10-10
Use the new Icon Composer .icon format.

pixlwave:
Also updates the Nightly icon to use glass layers.

Screenshot 2025-10-10 at 4 56 30 pm


pixlwave:
Linking this to #3164.

DidiDidi129:
Looks good, can't wait for the other logos to be updated!

PR #4615, opened 2025-10-11
Convert the timeline's long press gesture recogniser to UIKit and prevent scroll view conflicts

stefanceriu:


ElementX/Sources/Screens/Timeline/View/Style/LongPressWithFeedback.swift
```diff
@@ -101,3 +122,34 @@ struct LongPressWithFeedback_Previews: PreviewProvider, TestablePreview {
         }
     }
 }
+
+// Fixes the issue on iOS 18 where LongPress conflicts with the scroll view
+// https://github.com/feedback-assistant/reports/issues/542#issuecomment-2581322968
+private struct LongPressGesture: UIGestureRecognizerRepresentable {
```
pixlwave: I think this needs a custom name given it matches the existing SwiftUI one:
https://developer.apple.com/documentation/swiftui/longpressgesture

ElementX/Sources/Screens/Timeline/View/Style/LongPressWithFeedback.swift
pixlwave, line 1: There seems to be a delay on the animation starting when I test this on a device 😕

pixlwave: Ok after discussion and further investigation, it's the same with the current implementation.

PR #4621, opened 2025-10-13
FF that allows notifications to also focus events

### LINKED ISSUE #4620: [Task] Tapping a notification should also focus the event (and open thread if threaded event)

Velin92:
fixes #4620 

Only enabled on TF by default as of right now

### DIFF
--- a/ElementX/Sources/Application/Navigation/AppRoutes.swift
+++ b/ElementX/Sources/Application/Navigation/AppRoutes.swift
@@ -29,7 +29,7 @@ enum AppRoute: Hashable {
     /// The profile of a member within the current room.
     case roomMemberDetails(userID: String)
     /// An event within a room, shown as the root of the stack (popping any child rooms).
-    case event(eventID: String, threadState: ThreadState = .unknown, roomID: String, via: [String])
+    case event(eventID: String, threadRoot: ThreadRoot = .unknown, roomID: String, via: [String])
     /// The same as ``event`` but using a room alias.
     case eventOnRoomAlias(eventID: String, alias: String)
     /// An event within a room, either within the last child on the stack or pushing a new child if needed.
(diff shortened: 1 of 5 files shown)


ElementX/Sources/FlowCoordinators/ChatsFlowCoordinator.swift
```diff
@@ -488,8 +488,8 @@ class ChatsFlowCoordinator: FlowCoordinatorProtocol {
         switch entryPoint {
         case .room:
             coordinator.handleAppRoute(.room(roomID: roomID, via: via), animated: animated)
-        case .eventID(let eventID):
-            coordinator.handleAppRoute(.event(eventID: eventID, roomID: roomID, via: via), animated: animated)
+        case .eventID(let eventID, let threadSate):
```
pixlwave:
```suggestion
        case .eventID(let eventID, let threadState):
```

ElementX/Sources/Application/Navigation/AppRoutes.swift
```diff
@@ -60,6 +60,15 @@ enum AppRoute: Hashable {
...
 }
 
+/// Only used for handling notifications tap where we already know the thread state of an event
+/// Use it to skip while handling the `event` route to avoid an unnecessary fetch
+/// In theory when notifications will be able to update the event cache, we won't need this anymore
+enum ThreadState: Hashable {
+    case unknown
+    case unthreaded
+    case threaded(threadRootEventID: String)
+}
```
pixlwave, line 70: It might be useful to add/move documentation to each case for this enum. I got it after followed the flow through the code, but to begin with I wasn't 100% sure why the enum was necessary.

pixlwave: Actually thinking more about this, what do you think about naming this like so:

```
enum ThreadRoot {
    case unknown
    case none
    case eventID(String)
}
```

The `State` suffix doesn't seem to be very descriptive to me, and within the route it might be better aligned with the other parameters: `.event(eventID, threadRoot, roomID, via)`

PR #4626, opened 2025-10-14
Handle the long press gesture states better and avoid multiple action invocations

stefanceriu:


ElementX/Sources/Screens/Timeline/View/Style/LongPressWithFeedback.swift
```diff
@@ -49,8 +51,12 @@ struct LongPressWithFeedback: ViewModifier {
             // The wait time needs to be at least 0.5 seconds or the long press gesture will take precedence over long pressing links.
             try? await Task.sleep(for: .seconds(0.5))
             
-            if Task.isCancelled { return }
+            if Task.isCancelled {
+                MXLog.dev("Task got cancelled")
+                return
+            }
             
+            MXLog.dev("Doing the action")
```
pixlwave: I think you need to remove these given `.dev` is only available in `DEBUG`.

ElementX/Sources/Screens/Timeline/View/Style/LongPressWithFeedback.swift
```diff
@@ -19,10 +19,12 @@ struct LongPressWithFeedback: ViewModifier {
             mainContent(content: content)
                 .gesture(LongPressGestureRepresentable { gesture in
                     switch gesture.state {
+                    case .began:
+                        handleLongPress(isPressing: true)
                     case .ended, .cancelled, .failed:
                         handleLongPress(isPressing: false)
                     default:
```
pixlwave: Is it worth putting the actual remaining case here now for clarity? – I'm assuming it isn't a non-exhaustive enum here though.

stefanceriu: Unfortunately it is but apparently I can do a

```
case .possible, .changed:
    break
@unknown default:
    break
```

pixlwave: Silly non-exhaustive enums

PR #4629, opened 2025-10-16
Display members of a space

### LINKED ISSUE #4632: [Task] View and manage members in a space

Velin92:
fixes #4632

This PR allows spaces to present a their members.
Most of the PR is about extracting the existing logic of navigating through members into its own RoomMembersFlowCoordinator so that it can be reused multiple times (since it's used in the standard, room in the space, and in the future will also be used in the Space settings flow).

As of right now the flow coordinator implementation is only used in the space, but a follow up PR will soon be opened when I'll also replace the existing code present in the `RoomFlowCoordinator` to use this flow coordinator too.

The PR also includes a small refactor of the invite view screen that was weirdly having some of its logic handled outside to coordinators.

Things left to do in this PR:

- [x] Await for final unified copies for ban and remove actions so that can more easily be shared between spaces and rooms.
- [x] Clearing the route?
- [x] Handling routes if necessary?
- [x] How to handle the `continueWithSpace` action if we are already inside a space.

### DIFF
--- a/ElementX/Sources/Screens/Spaces/SpaceScreen/SpaceScreenViewModel.swift
+++ b/ElementX/Sources/Screens/Spaces/SpaceScreen/SpaceScreenViewModel.swift
@@ -69,6 +69,7 @@ class SpaceScreenViewModel: SpaceScreenViewModelType, SpaceScreenViewModelProtoc
         
         Task {
             if case let .joined(roomProxy) = await userSession.clientProxy.roomForIdentifier(spaceRoomListProxy.id) {
+                // Required to listen for membership updates in the members flow
                 await roomProxy.subscribeForUpdates()
                 state.roomProxy = roomProxy
                 if case let .success(permalinkURL) = await roomProxy.matrixToPermalink() {


ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
```diff
@@ -52,6 +53,8 @@ class SpaceFlowCoordinator: FlowCoordinatorProtocol {
         case presentingChild(childSpaceID: String, previousState: State)
         /// A room flow is in progress
         case roomFlow(previousState: State)
+        /// A members flow in progress
```
pixlwave:
```suggestion
        /// A members flow is in progress
```

ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
```diff
@@ -338,4 +365,34 @@ class SpaceFlowCoordinator: FlowCoordinatorProtocol {
...
         selectedSpaceRoomSubject.send(roomID)
     }
+    
+    private func startMembersFlow() async {
+        guard case let .space(spaceRoomListProxy) = entryPoint,
+              case let .joined(roomProxy) = await flowParameters.userSession.clientProxy.roomForIdentifier(spaceRoomListProxy.id) else {
+            fatalError("Attempting to show members of a non joined space")
+        }
+        
+        // Required to listen for membership updates
```
pixlwave: 👍

ElementX/Sources/FlowCoordinators/SpaceSettingsFlowCoordinator.swift
pixlwave, line 1: I don't think this one is used right? Maybe we could remove it from this PR for simplicity?

ElementX/Sources/Screens/RoomMemberListScreen/RoomMembersListScreenModels.swift
pixlwave, line 1: Looks like this should be reverted 🙂

ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
```diff
@@ -338,4 +365,34 @@ class SpaceFlowCoordinator: FlowCoordinatorProtocol {
         coordinator.handleAppRoute(.room(roomID: roomID, via: []), animated: true)
         selectedSpaceRoomSubject.send(roomID)
     }
+    
+    private func startMembersFlow() async {
+        guard case let .space(spaceRoomListProxy) = entryPoint,
+              case let .joined(roomProxy) = await flowParameters.userSession.clientProxy.roomForIdentifier(spaceRoomListProxy.id) else {
```
pixlwave: It would be much nicer for the SpaceScreen to include the `JoinedRoomProxy` in its `displayMembers` action. That's what is happening when you present a subspace and means that if you fail to get the proxy, you can show a nice error on the screen itself.

ElementX/Sources/Screens/Spaces/SpaceScreen/SpaceScreenModels.swift
```diff
@@ -25,6 +26,7 @@ struct SpaceScreenViewState: BindableState {
     var joiningRoomIDs: Set<String> = []
     
     var isSpaceManagementEnabled = false
+    var isSpaceJoined = false
```
pixlwave: Given this screen should only be shown for joined spaces, could we rename this to `roomProxyFailure` or something like that. Or even make it `var roomProxy: JoinedRoomProxyProtocol?`.

ElementX/Sources/Screens/Spaces/SpaceScreen/View/SpaceScreen.swift
```diff
@@ -60,8 +60,13 @@ struct SpaceScreen: View {
...
         ToolbarItem(placement: .primaryAction) {
             Menu {
-                if let permalink = context.viewState.permalink {
-                    Section {
+                Section {
+                    if context.viewState.isSpaceJoined {
+                        Button { context.send(viewAction: .displayMembers) } label: {
+                            Label(L10n.screenSpaceMenuActionMembers, icon: \.user)
+                        }
+                    }
```
pixlwave: Then this thing could become

```suggestion
                    if let roomProxy = context.viewState.roomProxy {
                        Button { context.send(viewAction: .displayMembers(roomProxy)) } label: {
                            Label(L10n.screenSpaceMenuActionMembers, icon: \.user)
                        }
                    }
```

ElementX/Sources/FlowCoordinators/RoomMembersFlowCoordinator.swift
```diff
@@ -0,0 +1,386 @@
...
+    case presentCallScreen(roomProxy: JoinedRoomProxyProtocol)
+    case verifyUser(userID: String)
+}
+
+enum RoomMembersFlowCoordinatorEntryPoint: Equatable {
+    /// To be used in a room when a member name is tapped
+    case roomMember(userID: String)
+    /// To be used in the context of room details, space details etc.
+    case roomMembersList
+}
```
pixlwave, line 23: Love it! 🙌

ElementX/Sources/FlowCoordinators/RoomMembersFlowCoordinator.swift
```diff
@@ -0,0 +1,386 @@
...
+    
+    func handleAppRoute(_ appRoute: AppRoute, animated: Bool) {
+        fatalError("Unavailable")
+    }
+    
+    func clearRoute(animated: Bool) {
+        // TODO: Implement
+    }
+    
+    private func configureStateMachine() {
```
pixlwave, line 111: I would love for use to re-write this using the same `Machine.addRoutes(event:transitions:handler:)` and `Machine.addRouteMapping(_:handler:)` that we use in the `SpaceFlowCoordinator`. It's so much cleaner to have the route and the handler in the same place.

ElementX/Sources/FlowCoordinators/RoomMembersFlowCoordinator.swift
```diff
@@ -0,0 +1,386 @@
...
+            case .toggleUser(let user):
+                var selectedUsers = selectedUsersSubject.value
+                
+                if let index = selectedUsers.firstIndex(where: { $0.userID == user.userID }) {
+                    selectedUsers.remove(at: index)
+                } else {
+                    selectedUsers.append(user)
+                }
+                
+                selectedUsersSubject.send(selectedUsers)
```
pixlwave: Would you mind moving these cases into the view model itself where they belong? Then we can remove `inviteUsers` `showLoadingIndicator` and `hideLoadingIndicator` from this flow.

ElementX/Sources/FlowCoordinators/RoomMembersFlowCoordinator.swift
```diff
@@ -0,0 +1,386 @@
...
+        case presentRoomMembersList
+
+        case presentRoomMemberDetails(userID: String)
+        case dismissRoomMemberDetails
+        
+        case presentInviteUsersScreen
+        case dismissInviteUsersScreen
+        
+        case presentUserProfile(userID: String)
+        case dismissUserProfile
```
pixlwave:
```suggestion
        case presentRoomMemberDetails(userID: String)
        case dismissedRoomMemberDetails
        
        case presentInviteUsersScreen
        case dismissedInviteUsersScreen
        
        case presentUserProfile(userID: String)
        case dismissedUserProfile
```

pixlwave (commented):
Amazing, I love how these flow coordinators are coming together 🙌

ElementX/Sources/FlowCoordinators/RoomMembersFlowCoordinator.swift
```diff
@@ -0,0 +1,386 @@
...
+                
+                switch action {
+                case .presentCallScreen(let roomProxy):
+                    actionsSubject.send(.presentCallScreen(roomProxy: roomProxy))
+                case .verifyUser(let userID):
+                    actionsSubject.send(.verifyUser(userID: userID))
+                case .continueWithSpaceFlow(let spaceRoomListProxy):
+                    // TODO: Implement
+                    // Not sure wha to do here since this can be presented also internally by the space flow
+//                    stateMachine.tryEvent(.startChildFlow, userInfo: SpaceFlowCoordinatorEntryPoint.space(spaceRoomListProxy))
```
Velin92: Ok this is not possible since only DMs can be opened up from this flow, so no need to handle it.

ElementX/Sources/FlowCoordinators/RoomMembersFlowCoordinator.swift
```diff
@@ -0,0 +1,386 @@
...
+        }
+    }
+    
+    func handleAppRoute(_ appRoute: AppRoute, animated: Bool) {
+        fatalError("Unavailable")
+    }
+    
+    func clearRoute(animated: Bool) {
+        // TODO: Implement
+    }
```
Velin92, line 109: Switch state machine and clear the route accordingly (check the SpaceFlowCoordinator)

ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
```diff
@@ -203,11 +203,15 @@ class SpaceFlowCoordinator: FlowCoordinatorProtocol {
...
+                return nil
+            }
             return .membersFlow
-        } handler: { [weak self] _ in
-            guard let self else { return }
-            Task { await self.startMembersFlow() }
+        } handler: { [weak self] context in
+            guard let self, let roomProxy = context.userInfo as? JoinedRoomProxyProtocol else {
+                fatalError("The room proxy must always be provided")
+            }
```
pixlwave, line 213: Nice, thank you 🙌

ElementX/Sources/Screens/Spaces/SpaceScreen/SpaceScreenViewModel.swift
```diff
@@ -69,7 +69,8 @@ class SpaceScreenViewModel: SpaceScreenViewModelType, SpaceScreenViewModelProtoc
         
         Task {
             if case let .joined(roomProxy) = await userSession.clientProxy.roomForIdentifier(spaceRoomListProxy.id) {
-                state.isSpaceJoined = true
+                await roomProxy.subscribeForUpdates()
```
pixlwave, line 73: Can we add back the comment here as it's useful to surface now rather than having to go hunt for it:

```suggestion
                // Required to listen for membership updates in the members flow
                await roomProxy.subscribeForUpdates()
```

ElementX/Sources/FlowCoordinators/RoomMembersFlowCoordinator.swift
```diff
@@ -98,7 +98,14 @@ final class RoomMembersFlowCoordinator: FlowCoordinatorProtocol {
...
     func clearRoute(animated: Bool) {
-        // TODO: Implement
+        if stateMachine.state == .inviteUsersScreen {
+            navigationStackCoordinator.setSheetCoordinator(nil, animated: animated)
+        } else if let roomFlowCoordinator {
+            roomFlowCoordinator.clearRoute(animated: animated)
+        }
+        // We don't support dismissing a sub flow by itself, only the entire chain.
+        // The presenter flow will take care of dismissing it
+        actionsSubject.send(.finished)
```
pixlwave, line 108: Ideally this would switch on the state and handle clearing the flow for all cases, then you wouldn't need that comment :)

Velin92: but the thing is that we can't.... because we can't pop to a specific coordinator, since this flow is never in the root, it's pretty much the same behaviour of a Room flow coordinator when is a child

Issue #2839, opened 2024-05-16
Implement displaying MSC2010 spoiler text 


ara4n:
Your use case
What would you like to do?
Display spoilers correctly in the timeline

Why would you like to do it?
See https://github.com/element-hq/element-meta/issues/2424

Have you considered any alternatives?
No response

Additional context
No response

Are you willing to provide a PR?
No

PR #4637, opened 2025-10-20
add attribute for spoilers

catgirlinspace:
Hi, this is a start to #2839. I've never done iOS development before (outside of playing around with the templates in Xcode years ago) but I tried to follow what was already there. Currently the string has the spoilered parts get marked with the attribute but I have no idea how to actually hide them (I'm guessing I need a pill thing?). 

Currently:
Element Web
CleanShot 2025-10-20 at 16 56 27@2x

Current Element X (spoilered text is shown as nothing)
CleanShot 2025-10-20 at 16 57 39@2x

This pull request currently (spoilered text is shown plain)
CleanShot 2025-10-20 at 16 57 55@2x

Leaving this as a draft for now while I figure out how to make it actually appear hidden. Also guessing maybe stuff for notifications would need changed to hide spoilered text? Any advice would be appreciated! 

### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Other/HTMLParsing/AttributedStringBuilder.swift
+++ b/ElementX/Sources/Other/HTMLParsing/AttributedStringBuilder.swift
@@ -32,6 +32,7 @@ extension NSAttributedString.Key {
     static let MatrixEventOnRoomAlias: NSAttributedString.Key = .init(rawValue: EventOnRoomAliasAttribute.name)
     static let MatrixAllUsersMention: NSAttributedString.Key = .init(rawValue: AllUsersMentionAttribute.name)
     static let CodeBlock: NSAttributedString.Key = .init(rawValue: CodeBlockAttribute.name)
+    static let MatrixSpoiler: NSAttributedString.Key = .init(rawValue: SpoilerAttribute.name)
 }
 
 struct AttributedStringBuilder: AttributedStringBuilderProtocol {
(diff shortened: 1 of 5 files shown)


stefanceriu:
Nice start, thank you 🙏 

Unfortunately this is indeed the easy part. 

As you need the spoiler view to be interactive (so you can tap to show) your only option is a custom NSTextAttachment and their NSTextAttachmentViewProvider, which, as you noticed, we also use for mention pills. 
I would however suggest you build the spoiler completely separate from all of that i.e. introduce a new type, a new provider and view.

I might also be a good idea to hold the spoiler text within the attribute so you can build the right size view. 
Also, keep in mind that text can wrap and your spoiler should too. [TextRenderer](https://developer.apple.com/documentation/swiftui/textrenderer) might prove useful.

lavender-sys:
Hi, sorry for the late reply--have been pretty busy. 

Thank you for those pointers! I'll try implementing something soon-ish (whenever I have free time, have been pretty busy with uni) but I'll admit I don't have much confidence in myself with my lack of iOS experience :(

stefanceriu:
This PR has been hanging around for a while now and it seems to be bit rotting Seeing as it still needs work and in the interest of cleaning up the list of opened PRs I'll close it. Do feel free to bring this up again if you find the time to work on it.

PR #4638, opened 2025-10-21
Only offer to verify if a cross-signed device is available

uhoreg:
Fixes https://github.com/element-hq/element-x-ios/issues/4190

Confirming your identity by verifying against another device only works if the other device that you're verifying against has been cross-signed.  Currently, when you try to confirm your identity, it will prompt you to verify against another device if you have *any* other device (with device keys) available.  This PR uses a new function in the Rust SDK (introduced in https://github.com/matrix-org/matrix-rust-sdk/pull/5699) to detect whether there are other devices that have been cross-signed that the user can verify against.

(The equivalent Element X Android PR was https://github.com/element-hq/element-x-android/pull/5433)

### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Mocks/Generated/GeneratedMocks.swift
+++ b/ElementX/Sources/Mocks/Generated/GeneratedMocks.swift
@@ -2287,6 +2287,11 @@ class ClientProxyMock: ClientProxyProtocol, @unchecked Sendable {
             return isOnlyDeviceLeftReturnValue
         }
     }
+
+    func hasDevicesToVerifyAgainst() async -> Result<Bool, ClientProxyError> {
+        return (await isOnlyDeviceLeft()).map { !$0 }
+    }
+
     //MARK: - startSync
 
     var startSyncUnderlyingCallsCount = 0
--- a/ElementX/Sources/Services/Client/ClientProxyProtocol.swift
+++ b/ElementX/Sources/Services/Client/ClientProxyProtocol.swift
@@ -136,6 +136,8 @@ protocol ClientProxyProtocol: AnyObject {
     
     func isOnlyDeviceLeft() async -> Result<Bool, ClientProxyError>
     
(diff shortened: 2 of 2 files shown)


uhoreg:
"Unit Tests" failure seems to be swiftformat complaining about a file that I didn't change, and not actually running the unit tests.

pixlwave:
> "Unit Tests" failure seems to be swiftformat complaining about a file that I didn't change, and not actually running the unit tests.

This should be fixed with a rebase on `develop` 👍

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -341,6 +341,16 @@ class ClientProxy: ClientProxyProtocol {
         }
     }
 
+    func hasDevicesToVerifyAgainst() async -> Result<Bool, ClientProxyError> {
+        do {
+            let result = try await client.encryption().hasDevicesToVerifyAgainst()
+            return .success(result)
+        } catch {
+            MXLog.error("Failed checking isLastDevice with error: \(error)")
```
pixlwave:
```suggestion
            MXLog.error("Failed checking hasDevicesToVerifyAgainst with error: \(error)")
```

uhoreg:
Unit Tests are now running, kind of.  It looks like some tests run, and pass.  But then it says `[13:54:38]: ▸ Testing started` in the logs, and then sits there for several hours, with nothing else showing in the logs, until the task finally gets killed.

stefanceriu:
> sits there for several hours

Don't worry about that, we've been having issues with CI runners ever since moving to os26. You need to cancel, rerun and then hope for the best (which is failing horribly in this situation).

pixlwave:
Thanks @uhoreg. I'm going to close this as I've included it in #4710 which also updates the UX to include a loading state at the same time.

PR #4640, opened 2025-10-21
Update copyright holding and dates

Velin92:
Name says it all

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,9 +46,9 @@ Please read our [forking guide](docs/FORKING.md).
 ## Copyright & License
 
 Copyright (c) 2025 Element Creations Ltd.
-Copyright (c) 2022 - 2025 New Vector Ltd
+Copyright (c) 2022 - 2025 New Vector Ltd.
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -45,6 +45,7 @@ Please read our [forking guide](docs/FORKING.md).
 
 ## Copyright & License
 
+Copyright (c) 2025 Element Creations Ltd.
 Copyright (c) 2022 - 2025 New Vector Ltd
 
 This software is dual licensed by New Vector Ltd (Element). It can be used either:
```
pixlwave: Presumably this one should be changed to Element Creations Ltd too now.

Tools/Scripts/Templates/SimpleScreenExample/ElementX/TemplateScreenCoordinator.swift
pixlwave, line 1: I always wonder about the headers on these templates. Do you think we should drop the 2022-2025 New Vector as whenever we create a new screen it immediately says 2022 which feels mega weird?

PR #4642, opened 2025-10-21
Update SDK to 25.10.21

Velin92:
This also changes the way the waveform is generated

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -11,7 +11,7 @@ on:
 jobs:
   tests:
     name: Tests
-    runs-on: macos-26
+    runs-on: macos-15
 
     concurrency:
       # Only allow a single run of this workflow on each branch, automatically cancelling older runs.


ElementX/Sources/Services/VoiceMessage/VoiceMessageRecorder.swift
```diff
@@ -127,17 +127,16 @@ class VoiceMessageRecorder: VoiceMessageRecorderProtocol {
...
         guard let url = audioRecorder.audioFileURL else {
             return .failure(.missingRecordingFile)
         }
         // build the waveform
-        var waveformData: [UInt16] = []
+        var waveformData: [Float] = []
         let analyzer = WaveformAnalyzer()
         do {
             let samples = try await analyzer.samples(fromAudioAt: url, count: 100)
-            // linearly normalized to [0, 1] (1 -> -50 dB)
```
pixlwave, line 139: Considering we're doing 0 to 1 and not -1 to 1, maybe let's keep this comment in now too?

pixlwave (approved):
LGTM 👍

PR #4644, opened 2025-10-21
Threaded notifications

### LINKED ISSUE #4645: [Task] Threaded push notifications

Velin92:
fixes #4645 

This PR improved the push notifications behaviour for threaded messages:
- Will group messages that belong to the same thread differently.
- Will show in DMs an extra line to represent that it's a threaded message.
- Will show in group chats that the message is from a thread (Copies to be confirmed)
- When a threaded notification is tapped it will bring you directly to the thread in the room.

IMG_0284

### DIFF
--- a/ElementX/Sources/Screens/Timeline/TimelineTableViewController.swift
+++ b/ElementX/Sources/Screens/Timeline/TimelineTableViewController.swift
@@ -395,6 +395,9 @@ class TimelineTableViewController: UIViewController {
             guard let self else { return }
             if let kvPair = timelineItemsDictionary.first(where: { $0.value.identifier.eventID == eventID }),
                let indexPath = dataSource?.indexPath(for: kvPair.key) {
+                // Scrolling to the middle created a small bump in the timeline
+                // Using top, which is bottom in the reversed timeline helps with rendering
+                // in full long messages and images
                 tableView.scrollToRow(at: indexPath, at: .top, animated: animated)
                 coordinator.send(viewAction: .scrolledToFocussedItem)
                 // Ensure VoiceOver focus happens after the scroll animation (if any)


ElementX/Sources/Services/Notification/NotificationConstants.swift
```diff
@@ -15,6 +15,7 @@ enum NotificationConstants {
         static let unreadCount = "unread_count"
         static let pusherNotificationClientIdentifier = "pusher_notification_client_identifier"
         static let receiverIdentifier = "receiver_id"
+        static let threadRootEventID = "thread_root_event_id"
```
stefanceriu: Let's be consistent with everything else that's around here and call this a `threadRootEventIdentifier`. I would also move it next to the existing `eventIdentifier`.

ElementX/Sources/Screens/Timeline/TimelineTableViewController.swift
```diff
@@ -395,7 +395,7 @@ class TimelineTableViewController: UIViewController {
             guard let self else { return }
             if let kvPair = timelineItemsDictionary.first(where: { $0.value.identifier.eventID == eventID }),
                let indexPath = dataSource?.indexPath(for: kvPair.key) {
-                tableView.scrollToRow(at: indexPath, at: .middle, animated: animated)
+                tableView.scrollToRow(at: indexPath, at: .top, animated: animated)
```
stefanceriu, line 401: Why did this change to `.top`? Will affect normal in-room event focusing and is that wanted?

ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -359,6 +359,12 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
...
             }
+        } else if appSettings.threadsEnabled, let threadRootEventID = content.threadRootEventID {
+            handleAppRoute(.thread(roomID: roomID, threadRootEventID: threadRootEventID, focusEventID: content.eventID))
+            return
+        } else if let eventID = content.eventID {
+            handleAppRoute(.event(eventID: eventID, roomID: roomID, via: []))
+            return
         }
         
         handleAppRoute(.room(roomID: roomID, via: []))
```
stefanceriu: This flow is a bit hard to read, would it maybe be easier to 

```diff
diff --git a/ElementX/Sources/Application/AppCoordinator.swift b/ElementX/Sources/Application/AppCoordinator.swift
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -360,13 +360,15 @@
                 storedRoomsToAwait = [roomID]
             }
-        } else if appSettings.threadsEnabled, let threadRootEventID = content.threadRootEventID {
-            handleAppRoute(.thread(roomID: roomID, threadRootEventID: threadRootEventID, focusEventID: content.eventID))
-            return
-        } else if let eventID = content.eventID {
-            handleAppRoute(.event(eventID: eventID, roomID: roomID, via: []))
-            return
+            
+            handleAppRoute(.room(roomID: roomID, via: []))
+        } else {
+            if appSettings.threadsEnabled, let threadRootEventID = content.threadRootEventID {
+                handleAppRoute(.thread(roomID: roomID, threadRootEventID: threadRootEventID, focusEventID: content.eventID))
+            } else if let eventID = content.eventID {
+                handleAppRoute(.event(eventID: eventID, roomID: roomID, via: []))
+            } else {
+                handleAppRoute(.room(roomID: roomID, via: []))
+            }
         }
-        
-        handleAppRoute(.room(roomID: roomID, via: []))
     }
```

NSE/Sources/NSEUserSession.swift
```diff
@@ -49,6 +50,7 @@ final class NSEUserSession {
          appSettings: CommonSettingsProtocol) async throws {
         sessionDirectories = credentials.restorationToken.sessionDirectories
         userID = credentials.userID
+        settings = appSettings
```
stefanceriu: We should keep with the name and call this `appSettings` too

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -162,41 +173,10 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
...
-                    case .failure:
-                        showErrorIndicator()
-                    }
-                }
-            }
+            handleChildEventRoute(eventID: eventID, roomID: roomID, via: via, animated: animated)
         case .share(let payload):
             guard let roomID = payload.roomID, roomID == self.roomID else {
-                fatalError("Navigation route doesn't belong to this room flow.")
+                fatalError("Navigation route doesn't belong to this rovom flow.")
```
stefanceriu: Typo

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -226,6 +206,41 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
         }
     }
     
+    private func handleChildEventRoute(eventID: String, roomID: String, via: [String], animated: Bool) {
+        if case .presentingChild = stateMachine.state, let childRoomFlowCoordinator {
+            childRoomFlowCoordinator.handleAppRoute(.childEvent(eventID: eventID, roomID: roomID, via: via), animated: animated)
```
stefanceriu, line 211: Nice cleanup but why did this change from

`childRoomFlowCoordinator.handleAppRoute(appRoute, animated: animated)`

to

`childRoomFlowCoordinator.handleAppRoute(.childEvent(eventID: eventID, roomID: roomID, via: via), animated: animated)`?

ElementX/Sources/Application/Navigation/AppRoutes.swift
```diff
@@ -51,6 +51,8 @@ enum AppRoute: Hashable {
     case share(ShareExtensionPayload)
     /// The change roles screen of a room with the transfer ownership setting
     case transferOwnership(roomID: String)
+    /// A thread within a room, only to be used to handle tap on notification for threaded events.
+    case thread(roomID: String, threadRootEventID: String, focusEventID: String?)
```
stefanceriu, line 55: I imagine at some point this should work with deeplinking as well, have you give any though to how that's going to fit with our existing `event`/`eventOnRoomAlias`/`childEvent`/`childEventOnRoomAlias`?

Would maybe a better solution be to put a `threadRootID` on those? instead of introducing a new `.thread` variant?

ElementX/Sources/Screens/Timeline/TimelineTableViewController.swift
Velin92: essentially I found out a bug where after focussing, there is a slight bump, this seems to be caused because once the rendering of the item is done the middle changes and results in the scrollTo being slightly off making the item not be displayed fully, but cropped in the middle (doesn't happen always but happens a lot when we get to the last item).

Using top (which is bottom since it's all reversed) allows to always have the full message rendered in the timeline, which is also better for longer messages and images, and prevents this bump.
So I think is just better in terms of UX alongside fixing this issue

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
Velin92: oh might be a typo, will double check

ElementX/Sources/Application/Navigation/AppRoutes.swift
Velin92: since we can't be sure if a thread already exists when using a permalink, those have their own implementation where they check from the event cache (or fetch remotely if not available) if there is a thread root, and then handle it at the presentation level, not on the permalink.

Velin92: Here is an example:
https://github.com/element-hq/element-x-ios/blob/bdad5f705226c64995f3d5bb2a895a2cae614186/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift#L287-L306

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
Velin92: Oh okay I double checked, the reason is that we aren't passing the AppRoute directly because at this point when the function is called we have already switch cased it, but just its content since we already know that is a `childEvent` case, so here we are just passing it again with all its own parameters

ElementX/Sources/Application/Navigation/AppRoutes.swift
Velin92: Also the idea of the thread variant is that it can also just open a thread, without also having it focus anything (in fact the focus event ID is optional).

ElementX/Sources/Screens/Timeline/TimelineTableViewController.swift
stefanceriu: Oh I see, sounds great, but I think you should leave a comment in there.

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
stefanceriu: Oh I see, smart 👍

ElementX/Sources/Application/Navigation/AppRoutes.swift
stefanceriu: Riight, right right, that makes perfect sense. Thank you for the explainer!

stefanceriu (approved):
Looks great, super cool addition! 👏

PR #4652, opened 2025-10-23
Use the call expiration timestamp to define the ringing window

BillCarsonFr:
### Pull Request Checklist

Re-edition of https://github.com/element-hq/element-x-ios/pull/4521 rebased and cherry-picked

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/SupportingFiles/Settings.bundle/Acknowledgements.plist
+++ b/ElementX/SupportingFiles/Settings.bundle/Acknowledgements.plist
@@ -180,9 +180,25 @@
 		</dict>
 		<dict>
 			<key>File</key>
+			<string>Packages/swift-clocks</string>
+			<key>Title</key>
+			<string>swift-clocks</string>
+			<key>Type</key>
+			<string>PSChildPaneSpecifier</string>
+		</dict>
+		<dict>
+			<key>File</key>
 			<string>Packages/swift-collections</string>
 			<key>Title</key>
 			<string>swift-collections</string>
+			<key>Type</key>
+			<string>PSChildPaneSpecifier</string>
+		</dict>
+		<dict>
+			<key>File</key>
+			<string>Packages/swift-concurrency-extras</string>
+			<key>Title</key>
+			<string>swift-concurrency-extras</string>
 			<key>Type</key>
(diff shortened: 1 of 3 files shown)


ElementX/Sources/Services/ElementCall/ElementCallService.swift
```diff
@@ -72,15 +66,34 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
     
     private var declineListenerHandle: TaskHandle?
     
-    override init() {
+    init(callProvider: CXProviderProtocol? = nil, timeProvider: TimeProvider? = nil) {
```
Velin92, line 69: Where is this new init being called? I don't see its usage anywhere outside of the test

ElementX/Sources/Services/ElementCall/ElementCallService.swift
```diff
@@ -72,15 +66,34 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
...
+            
+            if let callKitIcon = UIImage(named: "images/app-logo") {
+                configuration.iconTemplateImageData = callKitIcon.pngData()
+            }
+            
+            // https://stackoverflow.com/a/46077628/730924
+            configuration.supportedHandleTypes = [.generic]
+            
+            self.callProvider = CXProvider(configuration: configuration)
+        }
```
Velin92, line 89: Okay I see...
However I am not a huge fan of this approach, I would rather have the init() have non optional values but have always concrete values passed inside the init for the protocols.

ElementX/Sources/Services/ElementCall/ElementCallService.swift
```diff
@@ -14,6 +14,12 @@ import MatrixRustSDK
 import PushKit
 import UIKit
 
+// Keep this class testable
+struct TimeProvider {
+    var clock: any Clock<Duration>
+    var now: () -> Date
+}
```
Velin92, line 21: Not a fan, I think we had a unified class/protocol that can be mocked these could just be simple functions of such class that could be mocked.

PR #4663, opened 2025-10-27
Respect the order of joined spaces as defined by Element Web.

pixlwave:
… by updating the SDK 😄

### DIFF
--- a/ElementX/Sources/Services/Authentication/AuthenticationService.swift
+++ b/ElementX/Sources/Services/Authentication/AuthenticationService.swift
@@ -178,9 +178,8 @@ class AuthenticationService: AuthenticationServiceProtocol {
         
         do {
             let client = try await makeClient(homeserverAddress: scannedServerName)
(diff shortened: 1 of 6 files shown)


ElementX/Sources/Services/Authentication/AuthenticationService.swift
```diff
@@ -249,7 +248,7 @@ private extension HumanQrLoginError {
             .qrCodeError(.deviceNotSupported)
         case .OtherDeviceNotSignedIn:
             .qrCodeError(.deviceNotSignedIn)
-        case .Unknown, .OidcMetadataInvalid:
+        case .Unknown, .OidcMetadataInvalid, .CheckCodeAlreadySent, .CheckCodeCannotBeSent:
             .qrCodeError(.unknown)
```
pixlwave, line 252: @poljar: Should we be showing a specific error state for these new cases?

PR #4665, opened 2025-10-27
Refactored room flow coordinator to use the members flow coordinator

Velin92:
Since we have made a room members flow coordinator to handle members list and member details flows in spaces, it's only naturale we want to also migrate the existing room implementation into using this new flow coordinator.

This makes the room flow coordinator code a bit smaller, and improves the room members flow to be even more reusable and handle routes.


ElementX/Sources/FlowCoordinators/RoomMembersFlowCoordinator.swift
```diff
@@ -87,14 +90,37 @@ final class RoomMembersFlowCoordinator: FlowCoordinatorProtocol {
...
+        case .childEvent(let eventID, let roomID, let via):
+            if case .roomFlow = stateMachine.state, let roomFlowCoordinator {
+                roomFlowCoordinator.handleAppRoute(appRoute, animated: animated)
+            } else {
+                stateMachine.tryEvent(.startRoomFlow(roomID: roomID, via: via, eventID: eventID), userInfo: animated)
+            }
+        case .accountProvisioningLink, .roomList, .room, .roomAlias, .childRoomAlias, .roomDetails, .event,
+             .eventOnRoomAlias, .childEventOnRoomAlias, .userProfile, .call, .genericCallLink, .settings,
+             .chatBackupSettings, .share, .transferOwnership, .thread:
+            break
```
pixlwave: For clarity can we at least put the alias based routes on their own line with a comment like we do in the room flow:

https://github.com/element-hq/element-x-ios/blob/0bc832f8a60a25b8b2c7ad2bf52adc0dc24fb0cf/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift#L188-L191

ElementX/Sources/FlowCoordinators/RoomMembersFlowCoordinator.swift
```diff
@@ -52,11 +52,12 @@ final class RoomMembersFlowCoordinator: FlowCoordinatorProtocol {
         case presentUserProfile(userID: String)
         case dismissedUserProfile
         
-        case startRoomFlow(roomID: String)
+        case startRoomFlow(roomID: String, via: [String], eventID: String?)
         case stopRoomFlow
     }
     
     private let entryPoint: RoomMembersFlowCoordinatorEntryPoint
+    private let animatedEntry: Bool
```
pixlwave: Seeing this, I think I'm much more tempted to remove the entry point entirely, add `AppRoute.roomMembersList(roomID: String)`, and then force this flow coordinator to use `handleAppRoute` to start just like we do for `RoomFlowCoordinator`.

pixlwave (approved):
LGTM, although I think we should consider the app route solution for starting the flow now.

ElementX/Sources/FlowCoordinators/RoomMembersFlowCoordinator.swift
Velin92: I would refrain from doing this, because I feel that adding more AppRoutes that are not necessary could make our code more complicated, since we will have a case we don't need to handle in every other flow coordinator.
I am more inclined to give an `animated` parameter to the `start` function.
I also had a chat with Stefan on friday about this, and the idea is to keep it simple and use the start when we don't need to handle many different entries for a flow vs need to handle a lot of routable entries.

pixlwave: > I am more inclined to give an `animated` parameter to the `start` function.

Also makes sense to me given there's a dedicated FlowCoordinator protocol 👍

pixlwave:
> Warnings
> ⚠️	
> **ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift#L150** - Function should have complexity 10 or less; currently complexity is 13 (`cyclomatic_complexity`)
> 
> Generated by 🚫 [Danger Swift](https://danger.systems/swift) against [af25813](https://github.com/element-hq/element-x-ios/commit/af258138b3f2d388dc750cda368083058ada78f9)

Oh also, forgot to mention, looks like we need a swiftlint:disable on this too now 👆

PR #4669, opened 2025-10-28
Add timeouts to our GitHub actions.

pixlwave:
6 hours is way too long and eats up the runners.

In the long run it would be nice to have some kind of hang detection on the output from Fastlane, but for now this will help free up some runners.

### DIFF
--- a/.github/workflows/accessibility_tests.yml
+++ b/.github/workflows/accessibility_tests.yml
@@ -10,6 +10,7 @@ jobs:
   tests:
     name: Tests
     runs-on: macos-15
+    timeout-minutes: 120
 
     concurrency:
       # Only allow a single run of this workflow on each branch, automatically cancelling older runs.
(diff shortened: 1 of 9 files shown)


pixlwave:
> What about `danger.yml` and `translations-pr.yml`, which also utilise `macos-15` runtimes?

Both of those workflows are fast and they don't launch the simulator which is where our sticking point is right now, so I should think they're fine without it. I don't think I've seen either timeout before.

anoadragon453 (approved):
> Both of those workflows are and they don't launch the simulator which is where our sticking point is right now, so I should think they're fine without it. I don't think I've seen either timeout before.

Just because they've never timed out doesn't mean they won't in the future, or that a 6hr timeout is sensible. But it's up to you.

Changes LGTM otherwise.

pixlwave:
> Just because they've never timed out doesn't mean they won't in the future, or that a 6hr timeout is sensible.

heh, fair point 🙃

pixlwave:
Updated those and `automatic-calendar-version.yml`.

PR #4672, opened 2025-10-29
Run CI with Xcode 26.1 RC

Velin92:
Use macOS 26 as a runner and 26.1 RC to run the tests.
Everything works fine and these tests are actually generated using 26.1.1 so this PR can be made compatible with 26.1.1 once is released on GH, with minimal effort (we just need to update the `xcode select` in the `common.sh`)

- regenerated preview tests
- regenerated UI tests

a11y still fail due to that weird nav bar issue, so they will still be using 18.5 and macos-15, this seems to be an OS issue that we have reported to Apple

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -334,5 +334,3 @@ private_lane :create_simulator_if_necessary do |options|
   
   # device_id is unused right now but is useful to check e.g. the boot status of a simulator.
 end
-
-default_platform(:ios)


fastlane/Fastfile
```diff
@@ -334,3 +334,5 @@ private_lane :create_simulator_if_necessary do |options|
   
   # device_id is unused right now but is useful to check e.g. the boot status of a simulator.
 end
+
+default_platform(:ios)
```
pixlwave: This can be reverted, I pasted it in the original commit by mistake.

pixlwave (approved):
LGTM, it's hard to look at the diff of the UI test snapshots because of the renames but that's a job for another day.

PR #4674, opened 2025-10-29
Introduce a StartChatFlowCoordinator instead of handing a navigation stack to the Screen Coordinator.

pixlwave:
- Add a StartChatFlowCoordinator.
- Move all of the navigation handled by StartChatScreenCoordinator into the new StartChatFlowCoordinator.
- Show the InviteUsersScreen *after* the room was created.
- Refactor `CreateRoomCoordinator`/`ViewModel` etc to `CreateRoomScreenCoordinator`/`ViewModel` to match what we do everywhere else.
- Add a UI test that covers the entire flow.

### DIFF
--- a/ElementX/Sources/FlowCoordinators/StartChatFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/StartChatFlowCoordinator.swift
@@ -89,17 +89,17 @@ class StartChatFlowCoordinator: FlowCoordinatorProtocol {
         case .initial:
             break
         case .startChat:
-            navigationStackCoordinator.setRootCoordinator(nil)
+            navigationStackCoordinator.setRootCoordinator(nil, animated: animated) // StartChatScreen
         case .createRoom:
-            navigationStackCoordinator.setRootCoordinator(nil)
-            navigationStackCoordinator.pop()
+            navigationStackCoordinator.pop(animated: animated) // CreateRoomScreen
+            navigationStackCoordinator.setRootCoordinator(nil, animated: animated) // StartChatScreen
         case .roomAvatarPicker:
-            navigationStackCoordinator.setSheetCoordinator(nil)
+            navigationStackCoordinator.setSheetCoordinator(nil, animated: animated) // Media Picker
             clearRoute(animated: animated) // Re-run with the state machine back in the .createRoom state.
         case .inviteUsers:
-            navigationStackCoordinator.setRootCoordinator(nil)
-            navigationStackCoordinator.pop()
-            navigationStackCoordinator.pop()
+            navigationStackCoordinator.pop(animated: animated) // InviteUsersScreen
+            navigationStackCoordinator.pop(animated: animated) // CreateRoomScreen
+            navigationStackCoordinator.setRootCoordinator(nil, animated: animated) // StartChatScreen
         }
     }
     
@@ -218,7 +218,7 @@ class StartChatFlowCoordinator: FlowCoordinatorProtocol {
     private func presentInviteUsersScreen(roomProxy: JoinedRoomProxyProtocol) {
         let inviteParameters = InviteUsersScreenCoordinatorParameters(userSession: flowParameters.userSession,
                                                                       roomProxy: roomProxy,
-                                                                      isCreatingRoom: true,
+                                                                      isSkippable: true,
                                                                       userDiscoveryService: userDiscoveryService,
                                                                       userIndicatorController: flowParameters.userIndicatorController,
                                                                       appSettings: flowParameters.appSettings)
--- a/ElementX/Sources/Screens/InviteUsersScreen/InviteUsersScreenCoordinator.swift
+++ b/ElementX/Sources/Screens/InviteUsersScreen/InviteUsersScreenCoordinator.swift
@@ -12,7 +12,7 @@ import SwiftUI
 struct InviteUsersScreenCoordinatorParameters {
     let userSession: UserSessionProtocol
     let roomProxy: JoinedRoomProxyProtocol
-    let isCreatingRoom: Bool
+    let isSkippable: Bool
     let userDiscoveryService: UserDiscoveryServiceProtocol
     let userIndicatorController: UserIndicatorControllerProtocol
     let appSettings: AppSettings
@@ -34,7 +34,7 @@ final class InviteUsersScreenCoordinator: CoordinatorProtocol {
     init(parameters: InviteUsersScreenCoordinatorParameters) {
         viewModel = InviteUsersScreenViewModel(userSession: parameters.userSession,
                                                roomProxy: parameters.roomProxy,
-                                               isCreatingRoom: parameters.isCreatingRoom,
+                                               isSkippable: parameters.isSkippable,
                                                userDiscoveryService: parameters.userDiscoveryService,
                                                userIndicatorController: parameters.userIndicatorController,
                                                appSettings: parameters.appSettings)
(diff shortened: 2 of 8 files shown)


ElementX/Sources/Screens/CreateRoomScreen/CreateRoomScreenViewModel.swift
```diff
@@ -10,83 +10,70 @@ import Combine
...
+        var topic = ""
+        var isRoomPrivate = true
+        var isKnockingOnly = false
+        var avatarImageMedia: MediaInfo?
+        var aliasLocalPart: String?
+    }
+    
     private let userSession: UserSessionProtocol
-    private var createRoomParameters: CreateRoomFlowParameters
+    private var parameters: Parameters
```
pixlwave, line 26: Honestly, I'm not sure about these now that they aren't published. I feel like we could remove these parameters entirely now and simply use the view state instead, but there was quite a bit of logic here so I was nervous I might miss something important.

But either way, this PR is big enough already.

ElementX/Sources/FlowCoordinators/StartChatFlowCoordinator.swift
```diff
@@ -0,0 +1,237 @@
...
+        case .createRoom:
+            navigationStackCoordinator.setRootCoordinator(nil)
+            navigationStackCoordinator.pop()
+        case .roomAvatarPicker:
+            navigationStackCoordinator.setSheetCoordinator(nil)
+            clearRoute(animated: animated) // Re-run with the state machine back in the .createRoom state.
+        case .inviteUsers:
+            navigationStackCoordinator.setRootCoordinator(nil)
+            navigationStackCoordinator.pop()
+            navigationStackCoordinator.pop()
```
Velin92: why 2 pops here?

ElementX/Sources/Screens/CreateRoomScreen/CreateRoomScreenViewModel.swift
Velin92: They were used to restore the state between navigating back and forth between the invite screen and the create room, (so if you changed your mind and wanted to invite an additional person, you could go back, invite them, then go back to the room creatiuon and find all the things you have set up). However since now we invite users after room creation these are no more necessary to be here.

ElementX/Sources/Screens/CreateRoomScreen/CreateRoomScreenViewModel.swift
```diff
@@ -251,17 +263,23 @@ class CreateRoomViewModel: CreateRoomViewModelType, CreateRoomViewModelProtocol
...
-                                                        isRoomPrivate: createRoomParameters.isRoomPrivate,
+        switch await userSession.clientProxy.createRoom(name: parameters.name,
+                                                        topic: parameters.topic.isBlank ? nil : parameters.topic,
+                                                        isRoomPrivate: parameters.isRoomPrivate,
                                                         // As of right now we don't want to make private rooms with the knock rule
-                                                        isKnockingOnly: createRoomParameters.isRoomPrivate ? false : createRoomParameters.isKnockingOnly,
-                                                        userIDs: state.selectedUsers.map(\.userID),
+                                                        isKnockingOnly: parameters.isRoomPrivate ? false : parameters.isKnockingOnly,
+                                                        userIDs: [], // The invite users screen is shown next so we don't need to invite anyone right now.
```
Velin92, line 271: I guess we can then remove this?

ElementX/Sources/FlowCoordinators/StartChatFlowCoordinator.swift
```diff
@@ -0,0 +1,237 @@
...
+    
+    func clearRoute(animated: Bool) {
+        switch stateMachine.state {
+        case .initial:
+            break
+        case .startChat:
+            navigationStackCoordinator.setRootCoordinator(nil)
+        case .createRoom:
+            navigationStackCoordinator.setRootCoordinator(nil)
+            navigationStackCoordinator.pop()
```
Velin92: same here... Since we are already setting the root coordinator to nil, and this flow is usually presented the modal I am not sure why we are doing all of this.

ElementX/Sources/Screens/CreateRoom/CreateRoomModels.swift
```diff
@@ -16,28 +16,24 @@ enum CreateRoomScreenErrorType: Error {
...
     case displayCameraPicker
-    case removeImage
 }
 
-struct CreateRoomViewState: BindableState {
+struct CreateRoomScreenViewState: BindableState {
     var roomName: String
     let serverName: String
     let isKnockingFeatureEnabled: Bool
-    var selectedUsers: [UserProfileProxy]
```
Velin92, line 32: nice

ElementX/Sources/Screens/CreateRoomScreen/View/CreateRoomScreen.swift
```diff
@@ -134,30 +136,7 @@ struct CreateRoomScreen: View {
...
-                        context.send(viewAction: .deselectUser(user))
-                    }
-                    .frame(width: invitedUserCellWidth)
-                }
-            }
-            .padding(.horizontal, ListRowPadding.horizontal)
-            .padding(.vertical, 22)
         }
-        .frame(width: frame.width)
     }
```
Velin92, line 140: nice x2

ElementX/Sources/Screens/InviteUsersScreen/InviteUsersScreenCoordinator.swift
```diff
@@ -11,16 +11,15 @@ import SwiftUI
 
 struct InviteUsersScreenCoordinatorParameters {
     let userSession: UserSessionProtocol
-    let selectedUsers: CurrentValuePublisher<[UserProfileProxy], Never>?
-    let roomType: InviteUsersScreenRoomType
+    let roomProxy: JoinedRoomProxyProtocol
+    let isCreatingRoom: Bool
```
Velin92: since the room at this point has already been created I would rename it to "isRoomCreationFlow" however since at the end this seems to be used only to decide if the invitation is skippable or not I would just call it: "isSkippable" if you agree on that.

ElementX/Sources/FlowCoordinators/StartChatFlowCoordinator.swift
pixlwave: The stack consists of StartChatScreen, CreateRoomScreen and InviteUsersScreen (which has the back button disabled so you can't navigate back to the other 2), but you're right, I've updated it an added comments 👍

ElementX/Sources/Screens/CreateRoomScreen/CreateRoomScreenViewModel.swift
pixlwave: Heh I actually did that to begin with, but then reverted it as I remembered that generally the proxies should be a thin layer on top of the underlying Rust type, so didn't want to be opinionated.

ElementX/Sources/FlowCoordinators/StartChatFlowCoordinator.swift
pixlwave: `clearRoute` should always be implemented IMO, so that if one day someone decided to use this e.g. in an existing modal, they could call `clearRoute` from the parent flow and know that the stack is exactly as it was when that flow handed off to this one.

ElementX/Sources/Screens/InviteUsersScreen/InviteUsersScreenCoordinator.swift
pixlwave: Ah yes nice – I didn't like the name of this but hadn't thought of that as an alternative 👍

ElementX/Sources/Screens/CreateRoomScreen/CreateRoomScreenViewModel.swift
pixlwave: Ok nice, I'll open a follow-up PR then 👍

PR #4678, opened 2025-10-30
Space Settings UI tweaks

Velin92:
- Settings should be unreachable if no permissions are available
- The edit row in the settings can be seen but not tapped if none of the basic edit infor permissions are available
- Updated the ban and kick alert descriptions when such actions are taken in a space
- Added a copiable address to the space headers, if available.

### DIFF
--- a/ElementX.xcodeproj/project.pbxproj
+++ b/ElementX.xcodeproj/project.pbxproj
@@ -540,6 +540,7 @@
 		6298AB0906DDD3525CD78C6B /* LoremSwiftum in Frameworks */ = {isa = PBXBuildFile; productRef = 1A6B622CCFDEFB92D9CF1CA5 /* LoremSwiftum */; };
 		62A7FC3A0191BC7181AA432B /* AudioRecorder.swift in Sources */ = {isa = PBXBuildFile; fileRef = 907FA4DE17DEA1A3738EFB83 /* AudioRecorder.swift */; };
 		62C5876C4254C58C2086F0DE /* HomeScreenContent.swift in Sources */ = {isa = PBXBuildFile; fileRef = A3B4B58B79A6FA250B24A1EC /* HomeScreenContent.swift */; };
+		633501761094E09DFBEBFFAD /* CopyTextButton.swift in Sources */ = {isa = PBXBuildFile; fileRef = B682FE2C44C5E163E7023B05 /* CopyTextButton.swift */; };
 		63780F9DA06573E38A471ECA /* GenericCallLinkWidgetDriver.swift in Sources */ = {isa = PBXBuildFile; fileRef = 28C202C1C7E330F124981A31 /* GenericCallLinkWidgetDriver.swift */; };
 		6386EA3C898AD1A4BC1DC8A5 /* TimelineMediaPreviewModifier.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9FD40B92FCF20165658296AD /* TimelineMediaPreviewModifier.swift */; };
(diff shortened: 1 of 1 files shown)


ElementX/Sources/Screens/Spaces/Common/SpaceHeaderView.swift
```diff
@@ -83,14 +87,32 @@ struct SpaceHeaderView: View {
...
+            } icon: {
+                CompoundIcon(\.copy, size: .small, relativeTo: .compound.bodyLG)
+                    .accessibilityHidden(true)
+            }
+            .font(.compound.bodyLG)
+            .foregroundStyle(.compound.textSecondary)
+            .labelStyle(.custom(spacing: 4, iconLayout: .trailing))
+            .accessibilityHint(L10n.actionCopy)
+        }
+    }
```
pixlwave: Can we make this its own `View` struct? I think the idea is to replace the long tap to copy which we have for user IDs with this style too.

ElementX/Sources/Screens/Spaces/SpaceSettingsScreen/View/SpaceSettingsScreen.swift
```diff
@@ -26,13 +26,19 @@ struct SpaceSettingsScreen: View {
...
+                    }
+                })
+            } else {
+                ListRow(kind: .custom {
                     editSectionContent
-                }
-            })
+                })
+            }
         }
```
pixlwave, line 43: Given this is the same label just tappable or not, could we just switch out the row's content instead of the whole row? E.g.

```
Section {
    ListRow(kind: .custom {
        if condition {
            Button { action } label: { content }
        } else {
            content
        }
    })
}
```

pixlwave (approved):
Nice, LGTM 😎

ElementX/Sources/Screens/Spaces/Common/SpaceHeaderView.swift
Velin92: internal in this struct or external to be reused?

ElementX/Sources/Screens/Spaces/SpaceSettingsScreen/View/SpaceSettingsScreen.swift
Velin92: Tried, but it doesn't work because the closure of `.custom `is not a @ViewBuilder (and can't be because it's an enum closure), since the inner types are not matching.

Velin92: Okay no worries found a way to achieve a similar result, by using another computed var with ViewBuilder

ElementX/Sources/Screens/Spaces/Common/SpaceHeaderView.swift
pixlwave: External to be re-used for e.g. [here](https://www.figma.com/design/Ni6Ii8YKtmXCKYNE90cC67/Timeline--new-?node-id=1660-22712&m=dev) and [here](https://www.figma.com/design/jvZ7TIyd7V9Vn8sIDB9vMd/Compound-iOS-Templates?node-id=1-21770&m=dev)

ElementX/Sources/Screens/Spaces/SpaceSettingsScreen/View/SpaceSettingsScreen.swift
pixlwave: <del>Ahhh ok, fair enough :)</del> Nice! 😎

PR #4684, opened 2025-10-31
Reuse `RoomDetailsScreenViewModel` for the `SpaceSettingsScreen`

Velin92:
This is another way to approach the SpaceSettingsScreen, which at the end of the day is handling the details of a room.
This is more of an example PR to raise a discussion if this approach makes sense

PROs of this approach:
- Reuse existing code, so no code duplication, and saved up a bunch of hundreds of lines.
- Creating a simple screen UI and only use what is necessary from the view model
- Same for the coordinator, make it a simple one, reuse only the necessary actions from the view model that can be handled in that UI context

CONS:
- Less specific spaces customisation if it will ever be necessary (but since room and spaces are kinda the same thing, is very likely a customisation for one may also apply to the other?)
- The existing view model does internally some extra stuff that isn't really needed in this UI context
- Pass around services that are not really used in this context


ElementX/Sources/Screens/Spaces/SpaceSettingsScreen/SpaceSettingsScreenCoordinator.swift
```diff
@@ -27,18 +31,32 @@ final class SpaceSettingsScreenCoordinator: CoordinatorProtocol {
     }
     
     init(parameters: SpaceSettingsScreenCoordinatorParameters) {
-        self.parameters = parameters
-        
-        viewModel = SpaceSettingsScreenViewModel(roomProxy: parameters.roomProxy,
-                                                 userSession: parameters.userSession)
+        viewModel = RoomDetailsScreenViewModel(roomProxy: parameters.roomProxy,
```
stefanceriu, line 34: Imho this is super unorthodox. I like promimity between files and having the SpaceSettingsScreen just use a VM from a different screen's folder is unexpected.

Personally I would've kept the RoomDetailsScreen Coordinator and VM for both, would've introduced a mode/type/style enum for room/space/dm and made the split on the UI level instead of here.

PR #4687, opened 2025-10-31
Fix a potential bug where the token couldn't be refreshed when the cached server `/versions` had expired.

pixlwave:
… by updating the SDK to include https://github.com/matrix-org/matrix-rust-sdk/pull/5822.

Also has initial updates for UniFFI 0.30.0. There's more to investigate in this area but seemed worthwhile getting these first updates done to unblock SDK updates/fixes.

### DIFF
--- a/ElementX.xcodeproj/project.pbxproj
+++ b/ElementX.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 63;
+	objectVersion = 77;
 	objects = {
 
 /* Begin PBXAggregateTarget section */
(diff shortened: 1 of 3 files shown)


pixlwave:
> Also has initial updates for UniFFI 0.30.0. There's more to investigate in this area

Ok, seems that some of the changes to UniFFI mocks are crashing the app on deinit with `EXC_BAD_ACCESS` 😐

I guess that means this is a task for next week.

pixlwave:
Linking to https://github.com/mozilla/uniffi-rs/issues/2717.

In the mean time I tried adding our own empty `deinit` to the SDK mocks (and using SDK mocks everywhere – there are a few instantiations using `init(noHandle:)` rather than our wrappers) but it doesn't help as, [as noted](https://github.com/mozilla/uniffi-rs/issues/2717) in the Swift book:

> Superclass deinitializers are inherited by their subclasses, and the superclass deinitializer is called automatically at the end of a subclass deinitializer implementation. Superclass deinitializers are always called, even if a subclass doesn’t provide its own deinitializer.

pixlwave:
Have written a [workaround](https://github.com/element-hq/matrix-rust-components-swift/commit/3f0075fb8acaf202755512dfbafc28024e5125b2) that patches the generated UniFFI sources based on https://github.com/mozilla/uniffi-rs/pull/2720 which we can remove when we next upgrade UniFFI.

pixlwave:
Oh nice, tests actually passed for once!

PR #4701, opened 2025-11-04
Fix a bug where the timeline disappeared when VoiceOver was enabled.

### LINKED ISSUE #4696: VoiceOver users can't read conversations
### Steps to reproduce

1. Turn on VoiceOver and open a private chat or room.
2. Only the header buttons (back, call) and the bottom once (attach media, keyboard) are visible.
3. The whole conversation body is not reachable, neither via swipe nor tap.

#### What did you expect?
I expect VoiceOver to read out the chat messages, both being able to use the swipe gestures to focus previous and next message, as well as directly selecting a message.

#### What happened instead?
Because of this bug, it is impossible for any blind iOS user to read a chat with Element X, which is breaking the whole app.

### Your phone model

iPhone 15 Pro

### Operating system version

iOS 26.1

### Application version

25.10.2

### Homeserver

private and not relevant here

### Will you send logs?

No

pixlwave:
The bug was introduced when adopting iOS 26, where we updated the pinned items banner which had become visible in the navigation bar whilst transitioning. This PR removes the  `Color.clear` as within a `.safeAreaInset`, colours becomes greedy and take up all the available vertical space that they can.

We now have a nice reusable view modifier which we should be able to use for this banner animation anywhere we want.

Fixes #4696

https://github.com/user-attachments/assets/dcd5d4d0-df93-494e-8b44-24c5ab62422d


ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
```diff
@@ -43,9 +43,10 @@ struct RoomScreen: View {
...
             .safeAreaInset(edge: .top) {
-                // When voice over is on the table view is not reversed
-                // and the scroll gestures are not intercepted
-                // so we render the pinned banner on top.
+                // When VoiceOver is enabled, the table view isn't reversed and the scroll gestures
+                // don't trigger meaning the banner never hides itself and so the .overlay layout
+                // above permanently obscures the top of the timeline. So whenever VoiceOver is
+                // enabled we use a safe area inset to vertically stack it above the timeline.
                 if isVoiceOverEnabled {
                     pinnedItemsBanner
```
pixlwave, line 45: Realising just now that this one in the inset needs will need a different layout when there aren't any pinned items, I've got an idea to try out tomorrow.

pixlwave: Done in https://github.com/element-hq/element-x-ios/pull/4701/commits/586e34fdadde2092128ad2e4e4de9225032fc5a6

PR #4703, opened 2025-11-05
Pop to coordinator

Velin92:
Sometimes handling the `clearRoute` could be quite confusing, this setup allows the generalise the handling at least for all flow coordinators that are just pushed on the navigationStack (for flows that are entirely in modals, the presenter flow can just set the sheet to nil, so no need to do any route clearing inside the flow itself).

With this approach we store the last coordinator in the stack before the flow can push anything on top of it, and for any state that is not presenting a modal, or another flow, we just pop to this initial coordinator.

A test has been added with a simple test case for the RoomMembersFlowCoordinator, where the route is cleared after it has been started and another route has been handled.

### DIFF
--- a/ElementX/Sources/FlowCoordinators/RoomMembersFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/RoomMembersFlowCoordinator.swift
@@ -71,7 +71,7 @@ final class RoomMembersFlowCoordinator: FlowCoordinatorProtocol {
         actionsSubject.eraseToAnyPublisher()
     }
     
-    private var roomFlowCoordinator: RoomFlowCoordinator?
+    private var childFlowCoordinator: FlowCoordinatorProtocol?
     
     init(entryPoint: RoomMembersFlowCoordinatorEntryPoint,
          roomProxy: JoinedRoomProxyProtocol,
@@ -99,20 +99,20 @@ final class RoomMembersFlowCoordinator: FlowCoordinatorProtocol {
     func handleAppRoute(_ appRoute: AppRoute, animated: Bool) {
         switch appRoute {
         case .roomMemberDetails(let userID):
-            if case .roomFlow = stateMachine.state, let roomFlowCoordinator {
-                roomFlowCoordinator.handleAppRoute(appRoute, animated: animated)
+            if case .roomFlow = stateMachine.state, let childFlowCoordinator {
+                childFlowCoordinator.handleAppRoute(appRoute, animated: animated)
             } else {
                 stateMachine.tryEvent(.presentRoomMemberDetails(userID: userID), userInfo: animated)
             }
         case .childRoom(let roomID, let via):
-            if case .roomFlow = stateMachine.state, let roomFlowCoordinator {
-                roomFlowCoordinator.handleAppRoute(appRoute, animated: animated)
+            if case .roomFlow = stateMachine.state, let childFlowCoordinator {
+                childFlowCoordinator.handleAppRoute(appRoute, animated: animated)
             } else {
                 stateMachine.tryEvent(.startRoomFlow(roomID: roomID, via: via, eventID: nil), userInfo: animated)
             }
         case .childEvent(let eventID, let roomID, let via):
-            if case .roomFlow = stateMachine.state, let roomFlowCoordinator {
-                roomFlowCoordinator.handleAppRoute(appRoute, animated: animated)
+            if case .roomFlow = stateMachine.state, let childFlowCoordinator {
+                childFlowCoordinator.handleAppRoute(appRoute, animated: animated)
             } else {
                 stateMachine.tryEvent(.startRoomFlow(roomID: roomID, via: via, eventID: eventID), userInfo: animated)
             }
@@ -126,21 +126,13 @@ final class RoomMembersFlowCoordinator: FlowCoordinatorProtocol {
     }
     
     func clearRoute(animated: Bool) {
-        switch stateMachine.state {
-        case .inviteUsersScreen:
-            navigationStackCoordinator.setSheetCoordinator(nil, animated: animated)
-            clearRoute(animated: animated)
-        case .roomFlow:
-            roomFlowCoordinator?.clearRoute(animated: animated)
-            clearRoute(animated: animated)
-        case .initial:
-            break
-        case .roomMemberDetails, .roomMembersList, .userProfile:
-            guard let initialCoordinator else {
-                return
-            }
-            navigationStackCoordinator.pop(to: initialCoordinator, animated: animated)
+        childFlowCoordinator?.clearRoute(animated: animated)
+        navigationStackCoordinator.setSheetCoordinator(nil, animated: animated)
+        
+        guard let initialCoordinator else {
+            return
         }
+        navigationStackCoordinator.pop(to: initialCoordinator, animated: animated)
     }
     
     private func configureStateMachine() {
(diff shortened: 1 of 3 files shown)


ElementX/Sources/FlowCoordinators/RoomMembersFlowCoordinator.swift
```diff
@@ -123,14 +126,21 @@ final class RoomMembersFlowCoordinator: FlowCoordinatorProtocol {
...
+        case .roomFlow:
+            roomFlowCoordinator?.clearRoute(animated: animated)
+            clearRoute(animated: animated)
+        case .initial:
+            break
+        case .roomMemberDetails, .roomMembersList, .userProfile:
+            guard let initialCoordinator else {
+                return
+            }
+            navigationStackCoordinator.pop(to: initialCoordinator, animated: animated)
```
stefanceriu: Are you intentionally being recursive here? Wouldn't it be cleaner to **always** `setSheetCoordinator(nil)` + `roomFlowCoordinator.clearRoute()` + `navigationStackCoordinator.pop(to: initialCoordinator)`? 🤔

ElementX/Sources/Application/Navigation/NavigationCoordinators.swift
```diff
@@ -574,6 +574,21 @@ private struct NavigationSplitCoordinatorView: View {
         }
     }
     
+    func pop(to coordinator: CoordinatorProtocol, animated: Bool = true) {
```
stefanceriu, line 577: This looks like a good addition to me. I like that using it is optional and left to the final user.

ElementX/Sources/FlowCoordinators/SpaceSettingsFlowCoordinator.swift
```diff
@@ -95,11 +97,9 @@ final class SpaceSettingsFlowCoordinator: FlowCoordinatorProtocol {
...
         case .initial:
             break
-        case .spaceSettings:
-            navigationStackCoordinator.pop(animated: animated)
-        case .securityAndPrivacy:
-            navigationStackCoordinator.pop(animated: animated)
-            clearRoute(animated: animated)
+        case .spaceSettings, .securityAndPrivacy:
+            guard let initialCoordinator else { return }
+            navigationStackCoordinator.pop(to: initialCoordinator, animated: animated)
```
stefanceriu: `securityAndPrivacy` used to `clearRoute`, why doesn't it do that anymore? 🤔

ElementX/Sources/Other/SwiftUI/Views/OverridableAvatarImage.swift
```diff
@@ -39,3 +39,37 @@ struct OverridableAvatarImage: View {
         }
     }
 }
+
+extension View {
```
stefanceriu: Or this? 🤔

ElementX/Sources/FlowCoordinators/SpaceSettingsFlowCoordinator.swift
```diff
@@ -63,6 +63,10 @@ final class SpaceSettingsFlowCoordinator: FlowCoordinatorProtocol {
     
     private let stateMachine: StateMachine<State, Event>
     private var cancellables = Set<AnyCancellable>()
+    private var modalNavigationStackCoordinator: NavigationStackCoordinator?
+    
+    private var membersFlowCoordinator: RoomMembersFlowCoordinator?
+    private var rolesAndPermissionsFlowCoordinator: RoomRolesAndPermissionsFlowCoordinator?
```
stefanceriu: Now sure how this relates to this PR 🤔

UnitTests/Sources/RoomMembersFlowCoordinatorTests.swift
```diff
@@ -0,0 +1,80 @@
...
+            switch action {
+            case .finished:
+                true
+            default:
+                false
+            }
+        }
+        membersFlowCoordinator.clearRoute(animated: false)
+        try await deferred.fulfill()
+        try await deferredAction.fulfill()
```
stefanceriu: Every time I read deferred fulfillments I get confused. Would be so much easier if we could call them for what they are and just await directly e.g. `var navigateMemberDetails` -> `try await navigateMemberDetails`. We should at least not call them `deferred` and `deferredAction`

UnitTests/Sources/RoomMembersFlowCoordinatorTests.swift
```diff
@@ -0,0 +1,80 @@
...
+        try await deferred.fulfill()
+        XCTAssertTrue(navigationStackCoordinator.stackCoordinators.last is RoomMemberDetailsScreenCoordinator)
+        
+        deferred = deferFulfillment(stateMachineFactory.membersFlowStatePublisher) { $0 == .roomMembersList }
+        let deferredAction = deferFulfillment(membersFlowCoordinator.actions) { action in
+            switch action {
+            case .finished:
+                true
+            default:
+                false
```
stefanceriu, line 34: Would a `action == .finished` work better here instead?

ElementX/Sources/FlowCoordinators/RoomMembersFlowCoordinator.swift
Velin92: yeah, since the first line should set the state back to the previous one, and then we try clearing again normally, which when landing on a "poppable" state we just pop to.

It's pretty much what we are already doing on the roomflows the only difference is there isn't a like a cascade of pops for some cases

ElementX/Sources/FlowCoordinators/SpaceSettingsFlowCoordinator.swift
Velin92: this state doesn't have sheet or flows presented, so it can just pop directly to the initial coordinator and everything is dismissed.

ElementX/Sources/FlowCoordinators/SpaceSettingsFlowCoordinator.swift
Velin92: yeah mb, merge conflicts stuff, solved now

UnitTests/Sources/RoomMembersFlowCoordinatorTests.swift
Velin92: they are not equatable because... one of the enums has a JoinedRoomProxyProtocol as an associated value

stefanceriu: Oh boo

ElementX/Sources/FlowCoordinators/SpaceSettingsFlowCoordinator.swift
stefanceriu: Sure but if it when there's no sheets everything else is a no-op. Isn't it simpler to just always go through the same method every time instead of branching out?

ElementX/Sources/FlowCoordinators/RoomMembersFlowCoordinator.swift
stefanceriu: Same question as here I guess https://github.com/element-hq/element-x-ios/pull/4703#discussion_r2498288316

ElementX/Sources/FlowCoordinators/SpaceSettingsFlowCoordinator.swift
Velin92: Uhm not sure what you mean.
If there is a sheet -> dismiss it -> state goes to the previous one -> clear route again
If there is a flow -> clear its route -> state goes to the previous one -> clear route again
If there is none of the above just pop to the initial.

However we could argue that clear route again == pop to the initial coordinator

stefanceriu (approved):
Noice 👍

pixlwave:
Looks great btw, will be nice to start using this pattern further 😁

PR #4704, opened 2025-11-05
Handle threaded pinned events

### LINKED ISSUE #4709: [Task] View pinned threaded event

Velin92:
fixes #4709

https://github.com/user-attachments/assets/0cce6681-ac57-4c50-8d51-015d19c517e3

### DIFF
--- a/ElementX/Sources/FlowCoordinators/PinnedEventsTimelineFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/PinnedEventsTimelineFlowCoordinator.swift
@@ -87,26 +87,13 @@ class PinnedEventsTimelineFlowCoordinator: FlowCoordinatorProtocol {
                     presentMapNavigator(geoURI: geoURI, description: description, timelineController: timelineController)
                 case .displayMessageForwarding(let forwardingItem):
                     presentMessageForwarding(with: forwardingItem)
-                case .displayRoomScreenWithFocussedPin(let eventID):
-                    Task { await self.displayRoomScreenWithFocussedPin(eventID: eventID) }
+                case .displayRoomScreenWithFocussedPin(let eventID, let threadRootEventID):
+                    actionsSubject.send(.displayRoomScreenWithFocussedPin(eventID: eventID, threadRootEventID: threadRootEventID))
                 }
             }
             .store(in: &cancellables)
         
         navigationStackCoordinator.setRootCoordinator(coordinator)
-    }
-    
-    private func displayRoomScreenWithFocussedPin(eventID: String) async {
-        switch await roomProxy.loadOrFetchEventDetails(for: eventID) {
-        case .success(let event):
-            if flowParameters.appSettings.threadsEnabled, let threadRootEventID = event.threadRootEventId() {
-                actionsSubject.send(.displayRoomScreenWithFocussedPin(eventID: eventID, threadRootEventID: threadRootEventID))
-            } else {
-                actionsSubject.send(.displayRoomScreenWithFocussedPin(eventID: eventID, threadRootEventID: nil))
-            }
-        case .failure:
-            flowParameters.userIndicatorController.submitIndicator(.init(title: L10n.errorUnknown))
-        }
     }
     
     private func presentMapNavigator(geoURI: GeoURI, description: String?, timelineController: TimelineControllerProtocol) {
--- a/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
@@ -310,7 +310,8 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
                 case .success(let event):
                     if flowParameters.appSettings.threadsEnabled, let threadRootEventID = event.threadRootEventId() {
                         stateMachine.tryEvent(.presentRoom(presentationAction: .thread(rootEventID: threadRootEventID,
-                                                                                       focusEvent: focusEvent)),
+                                                                                       focusEvent: .init(eventID: focusEvent.eventID,
+                                                                                                         shouldSetPin: focusEvent.shouldSetPin))),
                                               userInfo: EventUserInfo(animated: animated))
                     } else {
                         stateMachine.tryEvent(.presentRoom(presentationAction: presentationAction), userInfo: EventUserInfo(animated: animated))
@@ -620,7 +621,9 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
         case .share(.mediaFiles(_, let mediaFiles)):
             stateMachine.tryEvent(.presentMediaUploadPreview(mediaURLs: mediaFiles.map(\.url)),
                                   userInfo: EventUserInfo(animated: animated, timelineController: timelineController))
-        case .share(.text), .eventFocus, .thread:
+        case .thread(let rootEventID, let focusEvent):
+            stateMachine.tryEvent(.presentThread(threadRootEventID: rootEventID, focusEventID: focusEvent?.eventID))
+        case .share(.text), .eventFocus:
             break // These are both handled in the coordinator's init.
         case .none:
             break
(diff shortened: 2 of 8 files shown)


ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -607,8 +621,9 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
         case .share(.mediaFiles(_, let mediaFiles)):
             stateMachine.tryEvent(.presentMediaUploadPreview(mediaURLs: mediaFiles.map(\.url)),
                                   userInfo: EventUserInfo(animated: animated, timelineController: timelineController))
-        case .thread(let rootEventID, let focusEventID):
-            stateMachine.tryEvent(.presentThread(threadRootEventID: rootEventID, focusEventID: focusEventID))
+        case .thread(let rootEventID, let focusEvent):
+            // Not sure if necessary here
```
Velin92: @pixlwave do you remember for which case this handling was about, because I was wondering if I need to update  the pinned banner also in this case, however I tested both flows (tap on banner and view in timeline) and it never want here, so I guess this is not really required.

pixlwave: This is either what gets called when you open a notification for an event that's inside a thread, or when you tap on a link from outside the app (like in Safari/Mail/Reminders etc). I can't remember which one.

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -304,7 +309,10 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
                 switch await roomProxy.loadOrFetchEventDetails(for: focusEvent.eventID) {
                 case .success(let event):
                     if flowParameters.appSettings.threadsEnabled, let threadRootEventID = event.threadRootEventId() {
-                        stateMachine.tryEvent(.presentRoom(presentationAction: .thread(rootEventID: threadRootEventID, focusEventID: focusEvent.eventID)), userInfo: EventUserInfo(animated: animated))
+                        stateMachine.tryEvent(.presentRoom(presentationAction: .thread(rootEventID: threadRootEventID,
+                                                                                       focusEvent: .init(eventID: focusEvent.eventID,
+                                                                                                         shouldSetPin: focusEvent.shouldSetPin))),
```
pixlwave, line 314: I'm probably missing something here, why can't this just be `focusEvent: focusEvent`?

ElementX/Sources/Screens/RoomScreen/RoomScreenCoordinator.swift
```diff
@@ -184,6 +187,8 @@ final class RoomScreenCoordinator: CoordinatorProtocol {
                     actionsSubject.send(.presentRoom(roomID: roomID, via: via))
                 case .displayMessageForwarding(let forwardingItem):
                     actionsSubject.send(.presentMessageForwarding(forwardingItem: forwardingItem))
+                case .displayThread(threadRootEventID: let threadRootEventID, focussedEventID: let focussedEventID):
```
pixlwave:
```suggestion
                case .displayThread(let threadRootEventID, let focussedEventID):
```

ElementX/Sources/Screens/RoomScreen/RoomScreenCoordinator.swift
```diff
@@ -201,6 +206,10 @@ final class RoomScreenCoordinator: CoordinatorProtocol {
         Task { await timelineViewModel.focusOnEvent(eventID: eventID) }
     }
     
+    func setSelectedPin(eventID: String) {
+        roomViewModel.setSelectedPinnedEventID(eventID)
+    }
```
pixlwave, line 212: Could we maybe just add a doc comment here to say that this allows us to set the banner to an item in a thread timeline that isn't directly visible in the room timeline.

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -402,6 +398,27 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
...
+    private func handleTappedPinnedEventsBanner() {
+        analyticsService.trackInteraction(name: .PinnedMessageBannerClick)
+        if let eventID = state.pinnedEventsBannerState.selectedPinnedEventID {
+            Task {
+                switch await roomProxy.loadOrFetchEventDetails(for: eventID) {
+                case .success(let event):
+                    if appSettings.threadsEnabled,
+                       let threadRootEventID = event.threadRootEventId() {
+                        actionsSubject.send(.focusEvent(eventID: threadRootEventID))
+                        actionsSubject.send(.displayThread(threadRootEventID: threadRootEventID, focussedEventID: eventID))
```
pixlwave, line 410: Do you think this should be a single action seeing as both actions can be async and there might be some weird interactions if they happen out of order?

ElementX/Sources/FlowCoordinators/PinnedEventsTimelineFlowCoordinator.swift
```diff
@@ -88,14 +88,27 @@ class PinnedEventsTimelineFlowCoordinator: FlowCoordinatorProtocol {
...
                 }
             }
             .store(in: &cancellables)
         
         navigationStackCoordinator.setRootCoordinator(coordinator)
     }
     
+    private func displayRoomScreenWithFocussedPin(eventID: String) async {
+        switch await roomProxy.loadOrFetchEventDetails(for: eventID) {
+        case .success(let event):
```
pixlwave: Could we do this bit in the `PinnedEventsTimelineScreenViewModel` and send the correct action upwards? It shouldn't be up to the flow coordinator to handle a view action this way if we can avoid it imo.

UnitTests/Sources/RoomScreenViewModelTests.swift
```diff
@@ -160,6 +161,63 @@ class RoomScreenViewModelTests: XCTestCase {
         try await deferred.fulfill()
     }
     
+    func testPinnedEventsBannerThreadedSelection() async throws {
+        ServiceLocator.shared.settings.threadsEnabled = true
```
pixlwave, line 166: Might this cause other tests to fail based on the run order? I think we should do the same as elsewhere and use a local copy that we reset:

https://github.com/element-hq/element-x-ios/blob/8974a4584af91376dfa713616ec69f6d5bb5a194/UnitTests/Sources/RoomSummaryProviderTests.swift#L25-L32

UnitTests/Sources/RoomScreenViewModelTests.swift
```diff
@@ -160,6 +161,63 @@ class RoomScreenViewModelTests: XCTestCase {
...
+        roomProxyMock.pinnedEventsTimelineReturnValue = .success(pinnedTimelineMock)
+        let viewModel = RoomScreenViewModel(userSession: UserSessionMock(.init()),
+                                            roomProxy: roomProxyMock,
+                                            initialSelectedPinnedEventID: "test1",
+                                            ongoingCallRoomIDPublisher: .init(.init(nil)),
+                                            appSettings: ServiceLocator.shared.settings,
+                                            appHooks: AppHooks(),
+                                            analyticsService: ServiceLocator.shared.analytics,
+                                            userIndicatorController: ServiceLocator.shared.userIndicatorController)
+        self.viewModel = viewModel
```
pixlwave, line 191: I think you could add a few new lines in here so this is easier to read :)

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
Velin92: yeah you're right, okay then can leave it as it is and remove this comment

ElementX/Sources/FlowCoordinators/PinnedEventsTimelineFlowCoordinator.swift
Velin92: we would need to do this also from the timeline that's why, because the view in timeline can be either triggered by the PinnedEventsTimelineScreenViewModel (from a media) and from the TimelineViewModel.

I just did not want to replicate code in both, that's why I did it here.

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
Velin92: you're absolutely right

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
Velin92: they could even be in different order, they will trigger two completely different behaviour in the coordinators. So doesn't really mattter too much.

PR #4721, opened 2025-11-10
Translations update

RiotRobot:
- Translations update


pixlwave:
Both CI runs are stuck but I've fixed the snapshots locally.

PR #4723, opened 2025-11-10
Fix the integration tests.

pixlwave:
I forgot to handle the change in order of the create room & invite users screens in the start chat flow in #4674.

Integration tests run: https://github.com/element-hq/element-x-ios/actions/runs/19232077525

### DIFF
--- a/IntegrationTests/Sources/UserFlowTests.swift
+++ b/IntegrationTests/Sources/UserFlowTests.swift
@@ -156,9 +156,8 @@ class UserFlowTests: XCTestCase {
         
         tapOnButton(A11yIdentifiers.startChatScreen.createRoom)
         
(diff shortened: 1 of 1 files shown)


pixlwave:
Actually, maybe we can just remove the `checkRoomCreation` function entirely here, given it doesn't actually contact the server and we now have a dedicated [UI test for the entire flow](https://github.com/element-hq/element-x-ios/blob/2ccbe52bc69d17288b0bb4a91bd0330f1a93cb97/UITests/Sources/StartChatTests.swift#L25) 🤔

pixlwave:
I feel like this should work but so far I haven't managed to get a successful run one way or another. Let's yolo merge it and see what happens in the morning.

PR #4730, opened 2025-11-11
Make sure the integration tests stop when running out of WAS retries 🙈

pixlwave:
I noticed that an apparently failed run actually completed successfully even though the WAS failure was being handled.

Screenshot 2025-11-11 at 11 39 35 am

A bit more debugging and the answer became clear:

618878B5-8F39-4AA6-85C1-A2C5D033C58F

### DIFF
--- a/IntegrationTests/Sources/Common.swift
+++ b/IntegrationTests/Sources/Common.swift
@@ -8,10 +8,14 @@
 
 import XCTest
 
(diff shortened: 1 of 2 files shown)


pixlwave:
Doesn't seem worth running the integration tests for this. I'd rather see them run on `develop` and maybe even pass one day.

PR #4735, opened 2025-11-11
Permissions screen redesign

Velin92:
Part of the following task: https://github.com/element-hq/element-x-ios/issues/4682

This updates the logic of the navigation, to have all the possible groups be handled directly in a single generic "Permissions" screen. The groups will also dynamically depend if the room is a space or not.

A follow up PR will be opened up to complete #4682 once we also have the "Manage rooms" permission available, which is being worked on here: https://github.com/matrix-org/matrix-rust-sdk/pull/5857/files


ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenModels.swift
pixlwave, line 1: I think this PR includes old commits? 🤔

ElementX/Sources/Screens/RoomChangePermissionsScreen/RoomChangePermissionsScreenModels.swift
```diff
@@ -48,54 +48,77 @@ enum RoomChangePermissionsScreenViewAction {
     case cancel
 }
 
+enum RoomChangePermissionsGroup: CaseIterable {
```
pixlwave:
```suggestion
enum RoomChangePermissionsScreenGroup: CaseIterable {
```

PreviewTests/Sources/__Snapshots__/PreviewTests/roomChangePermissionsScreen.Room-iPad-pseudo.png
pixlwave, line 1: Were the pseudo-language snapshots recorded before you pulled the new strings in?

ElementX/Sources/Screens/RoomChangePermissionsScreen/View/RoomChangePermissionsScreen.swift
```diff
@@ -14,24 +14,33 @@ struct RoomChangePermissionsScreen: View {
...
+        .navigationTitle(L10n.screenRoomChangePermissionsTitle)
         .navigationBarTitleDisplayMode(.inline)
         .navigationBarBackButtonHidden(context.viewState.hasChanges)
         .toolbar { toolbar }
         .alert(item: $context.alertInfo)
     }
     
+    @ViewBuilder
+    private func section(for group: RoomChangePermissionsGroup) -> some View {
+        if let settings = $context.settings[group].unwrap() {
```
pixlwave: I don't think you need the special `unwrap` extension here, Binding already has a special optional init for doing this:

```suggestion
        if let settings = Binding($context.settings[group]) {
```

ElementX/Sources/Screens/RoomChangePermissionsScreen/View/RoomChangePermissionsScreen.swift
```diff
@@ -14,24 +14,33 @@ struct RoomChangePermissionsScreen: View {
...
-            ForEach($context.settings) { $setting in
-                Section {
-                    ListRow(label: .plain(title: setting.title),
-                            kind: .inlinePicker(selection: $setting.value, items: setting.allValues))
-                } header: {
-                    Text(setting.title)
-                        .compoundListSectionHeader()
-                }
+            ForEach(RoomChangePermissionsGroup.allCases, id: \.self) { group in
+                section(for: group)
```
pixlwave, line 18: Not important, but I don't really see the advantage of the dedicated function for this. Maybe it went through a few refactors and made more sense before?

PreviewTests/Sources/__Snapshots__/PreviewTests/roomChangePermissionsScreen.Room-iPad-pseudo.png
Velin92: weird... I regenerated the tests after the strings update so not sure how did this happen. Maybe some invalid cache build?

Velin92: Weird... no I first pulled the strings and then generated all the tests.

ElementX/Sources/Screens/RoomChangePermissionsScreen/View/RoomChangePermissionsScreen.swift
Velin92: I just didn't want to have a ForEach directly into another ForEach for readability

pixlwave: Ahhh ok yeah I missed that part, makes sense now 👍

PR #4739, opened 2025-11-12
Fix missing EmojiPicker emojis: stop having identifier conflicts between different categories

stefanceriu:
(i.e. recently used vs. the rest)


stefanceriu:
Same failing `testCallIsTimingOut`, will ignore and force merge.

PR #4740, opened 2025-11-12
Manage rooms in space permission

### LINKED ISSUE #4682: [Task] Update the Roles and Permissions flow (and screens) to work with Spaces.

Velin92:
fixes #4682 

Requires the following PR to be merged: https://github.com/matrix-org/matrix-rust-sdk/pull/5857


ElementX/Sources/Services/Room/RoomPermissions.swift
```diff
@@ -67,6 +68,8 @@ struct RoomPermissions {
     var roomAvatar: RoomRole
     /// The level required to change the room's topic.
     var roomTopic: RoomRole
+    /// The level required to add/remove childrens from a space.
+    var manageRoomsInSpace: RoomRole
```
stefanceriu: Have updated the Rust PR to call this `spaceChild` so it matches the underlying state event so let's please do the same here.

PR #4741, opened 2025-11-12
Don't assert specific devices for accessibility tests and use iOS 18 again.

pixlwave:
Making sure it's on a phone should be enough, nothing is snapshotted so this isn't device critical like UI/snapshot tests.

Accessibility tests run: https://github.com/element-hq/element-x-ios/actions/runs/19307661617


pixlwave:
Ok the run with iOS was much more successful (1 random failing test), I think it makes sense to go back there for these now and re-evaluate these tests when the iOS 26 simulator is more stable.

Issue #4681, opened 2025-10-31
[Task] Update the Security & Privacy screen to work with Spaces.


pixlwave:
Please see and discuss the details in the meta issue.

PR #4742, opened 2025-11-12
Security and privacy redesign for spaces.

Velin92:
first part of #4681 

This PR redesigns the security and privacy view with the latest updates to also accomodate the future spaces implementation.
It also tweaks some copies in other screens, so that the shared views between room and spaces are more reusable.


ElementX/Resources/Localizations/be.lproj/Localizable.strings
pixlwave, line 1: Might be nice to just commit `en` and `en-US` rather than all of them in PRs. But I think I'll revert the localazy config so it downloads both `en`s do we don't need to run the tool with `--all-languages`.

pixlwave (approved):
Nice 😎

ElementX/Resources/Localizations/be.lproj/Localizable.strings
pixlwave: https://github.com/element-hq/element-x-ios/pull/4743

PR #4743, opened 2025-11-13
Download en and en-US when running download-strings

pixlwave:
Completely my mistake, I didn't realise the effect when I made it `en` only, but given that `en-US` is used for the pseudo-language snapshots, this will make life easier for us during development.


stefanceriu (approved):
👍

PR #4756, opened 2025-11-17
Rename snapshots

Velin92:
This allows for the snapshots to be easily diffed, when an upgrade in either the device or the OS happens.

Also it updates the compound tests to run with 26.1


stefanceriu (approved):
👌

PR #4758, opened 2025-11-17
Separate `displayName` from `avatarDisplayName` when generating notification icons

### LINKED ISSUE #4745: Threaded notifications do not use the right fallback avatar
### Steps to reproduce

1. Alice and Bob are in a room called "Something" with no avatar
2. Alice pinged Bob in a thread

--> Bob receives a notification on EXI with "T" as avatar instead of "S"

#### What did you expect?
The avatar in the notification should be "S" like in "Something"

#### What happened instead?
The avatar in the notification was a "T" probably like in "Thread". The color was different too

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

Nightly 25.11.2 (1066)

### Homeserver

_No response_

### Will you send logs?

Yes

Velin92:
fixes #4745 

Since sometimes we customise the string that appears in the notification, we should separate such string from the actual name that needs to be used when rendering the placeholder avatar.

This PR separates such logic for both sender and rooms.

### DIFF
--- a/NSE/Sources/NotificationContentBuilder.swift
+++ b/NSE/Sources/NotificationContentBuilder.swift
@@ -109,9 +109,11 @@ struct NotificationContentBuilder {
         
         notificationContent.body = body
         
+        let name = notificationItem.senderDisplayName ?? notificationItem.roomDisplayName
         await addSenderIcon(notificationContent: &notificationContent,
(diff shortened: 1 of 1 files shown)


stefanceriu (approved):
lgtm 👍

PR #4762, opened 2025-11-18
Test NSE

Velin92:
This PR allows to test the notification content after it has been processed by the NSE, using UnitTests.

It contains 2 major project changes

### NSE Files references

All NSE files are also referenced by Element X (the main app). 
The reason for this is to allow generating mocks, but most importantly because due to a limitation of the Unit Tests, importing as testable the NSE is not possible, the compiler will not flag any error because it's able to find the definitions of the files... but will fail when building because it's not able to compile and link their content, since for some reason the UnitTests do not include the sources of app extensions directly, but just their signatures.
The only solution is then to have also the NSE code referenced in Element X so that the testable import resolves and compiles its sources.

### Dynamic Swift Package
Use of the swift package [Dynamic](https://github.com/mhdhejazi/Dynamic). 
It allows to access private APIs from Apple... I know that sounds crazy, but the package is only used in UnitTests and it's not supposed to be used in the production code of the main app.
But what is the reason to do so, you might ask. The reason is simple, we use the quite obscure Communication Notifications APIs for instant messaging, these do a lot of stuff behind the curtains. 
For example the title and the subtitle of the notification content are completely ignored when rendering the notification, and the values provided when generating the `INIntent` to render the icon, set the sender, and the group of the message belongs to (if available) get used in its place.
This all gets stored in a private member of the notification content called `communicationContext`, and this member is not normally accessible since it's supposed to be used by the OS to render the notification as a communication notification. However there is a ton of logic behind customising the icon, the names displayed for the sender and/or the room based on mentions, threads, DMs and so on.
Accessing the `communicationContext` values is the best way to verify that the notification is formatted in the proper way and manner.

### Other changes
- Moved some files, like the NotificationItemProxy into the NSE folder
- Renamed `addSeconderIcon` to `addCommunicationContext` since the old name was misleading, and made it seem like the function just added the icon, when in reality it also controls the rendering of the title and subtitle.

### What is missing?
Would be even nicer if we could get the icon file/image path... but I can't seem to find where it's referenced in the notification content. We could in this way check that the correct placeholder avatar is displayed, or if the correct avatar is used when available. But I am starting to think that is not at all present in the notification content itself, and that the OS does some stuff behind the curtains to handle that.

### DIFF
--- a/ElementX.xcodeproj/project.pbxproj
+++ b/ElementX.xcodeproj/project.pbxproj
@@ -24,7 +24,6 @@
 
 /* Begin PBXBuildFile section */
 		0033481EE363E4914295F188 /* LocalizationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = C070FD43DC6BF4E50217965A /* LocalizationTests.swift */; };
-		0045F04EEA78E4B0F5CF0478 /* NSEUserSessionMock.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4FBDD04CDFF421AC09D4A99D /* NSEUserSessionMock.swift */; };
 		00C3023B6DF55024D8876B76 /* ShareExtension.appex in Embed Foundation Extensions */ = {isa = PBXBuildFile; fileRef = 3D8BEEFCA07BEA43F4F4BF77 /* ShareExtension.appex */; settings = {ATTRIBUTES = (RemoveHeadersOnCopy, ); }; };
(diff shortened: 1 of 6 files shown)


Velin92:
The tests take 0.5 locally... but sometimes on CI they may take more than 1 minute causing the test to fail. 
It's probably due to the rendering of the placeholder image just taking too long on CI. Since we are not able to get the image anyway for testing, maybe it's worth just skipping that part in the code when we are running a test?

PR #4764, opened 2025-11-19
Add notification sound selection feature

### LINKED ISSUE #3878: Choose Sound notification
Would be great to be able to choose the notification sound as the alert for element x.  Currently I hear a "swish" but its rather quiet and need to turn up volume - it then makes all other alert's loud on the phone. 

A solution would be to be to allow custom sound notification under the settings such as in iMessage.  This allows the option of choosing which sound to assign to element X and to be able to prioritise element X notifications

Please see the screenshots of element X vs iMessage notification settings

Thanks for the consideration

Image
Image

blackwolf3342:
Fixes #3878

- Implemented notification sound selection screen with MVVM-Coordinator pattern
- Added sound preview functionality using AVAudioPlayer
- Organized sounds into Alert Tones and Classic categories
- Added checkmark indicator for selected sound
- Integrated with NotificationSettingsScreen
- Added UNNotificationSound support in NSE for custom notification sounds
- Added 22 notification sound files (.m4r and .caf formats)
- Translated notification sound settings to all 37 supported languages
- Updated AppSettings with notificationSoundName preference

### Pull Request Checklist

- [x] Voiceover enabled.

**Not tested**
I were unable to test notification sounds when receiving notifications due to difficulties with configuring App groups and Bundle Ids, and receiving notifications from the server. So that part require testing


ElementX/Resources/Localizations/be.lproj/Localizable.strings
pixlwave, line 1: All of these translations would get wiped out by our next import from the Localazy [project](https://localazy.com/p/element) so shouldn't be included in PRs. In the future you only need add the `en` strings which (after approval from product/design) we would add to Localazy for you.

PreviewTests/Sources/GeneratedPreviewTests.swift
pixlwave, line 1: Note that the preview tests would fail on this PR due to not having committed (to LFS) the snapshots for this screen.

pixlwave (commented):
Hey @blackwolf3342 thanks very much for this contribution however it contains what appear to be assets that are copyrighted by Apple, so we cannot accept this into our AGPL licensed codebase.

We very much welcome community PRs, however we need to have Product and Design input on the issues being fixed/implemented before accepting anything into the app, which as you can see from #3878 is not yet the case.

I've added a couple of comments on your PR to act as guidance for any future PRs you may wish to open :)

PR #4769, opened 2025-11-20
Room members list redesign

Velin92:
A small redesign of the view to be more consistent with the details screen of rooms and spaces.


ElementX/Sources/Screens/RoomMemberListScreen/View/RoomMembersListScreen.swift
```diff
@@ -59,41 +58,43 @@ struct RoomMembersListScreen: View {
...
-                ForEach(entries, id: \.member.id) { entry in
-                    RoomMembersListScreenMemberCell(listEntry: entry, context: context)
+                LazyVStack(alignment: .leading, spacing: 0) {
+                    ForEach(entries, id: \.member.id) { entry in
+                        ListRow(kind: .custom {
+                            RoomMembersListScreenMemberCell(listEntry: entry, isLast: entries.last == entry, context: context)
+                        })
+                    }
                 }
+                .clipShape(sectionShape)
```
Velin92, line 83: I wonder... do I need to provide bgCanvasDefault as a background?

ElementX/Sources/Screens/RoomMemberListScreen/View/RoomMembersListScreenMemberCell.swift
```diff
@@ -44,11 +46,20 @@ struct RoomMembersListScreenMemberCell: View {
...
+                            .font(.compound.bodyLG)
                             .foregroundStyle(.compound.textSecondary)
                     }
                 }
+                .overlay(alignment: .bottom) {
+                    if !isLast {
+                        Color.compound.bgSubtlePrimary
+                            .opacity(0.12)
+                            .frame(height: 1)
+                            .offset(x: 0, y: ListRowPadding.vertical)
```
stefanceriu: Am I mistaken or this doesn't actually show up in the snapshots? 🤔

ElementX/Sources/Screens/RoomMemberListScreen/View/RoomMembersListScreen.swift
stefanceriu: I can't hurt I imagine, better to be explicit.

ElementX/Sources/Screens/RoomMemberListScreen/View/RoomMembersListScreenMemberCell.swift
Velin92: fixed it, you're right the opacity was too much, instead I am using the same Color we use in ListRow now

PR #4771, opened 2025-11-21
Redesigned empty state for room members list

Velin92:
It also includes:
- Hiding the toggle when no banned members are available
- When a user is unbanned and there are no more banned members, the mode automatically switches to `.member`

### DIFF
--- a/ElementX/Sources/Screens/RoomMemberListScreen/View/RoomMembersListScreen.swift
+++ b/ElementX/Sources/Screens/RoomMemberListScreen/View/RoomMembersListScreen.swift
@@ -77,7 +77,7 @@ struct RoomMembersListScreen: View {
                         })
                     }
                 }
-                .background(.compound.bgCanvasDefault)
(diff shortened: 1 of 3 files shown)


ElementX/Sources/Screens/RoomMemberListScreen/View/RoomMembersListScreen.swift
```diff
@@ -81,6 +79,7 @@ struct RoomMembersListScreen: View {
                 }
                 .background(.compound.bgCanvasDefault)
```
pixlwave: Unrelated to the PR but sidenote: We should use `bgCanvasDefaultLevel1` for the Form/inset-List style otherwise dark mode will be incorrect:

https://github.com/element-hq/element-x-ios/blob/e7f55a99216a5658d0def20b64f38062e8ee7c32/compound-ios/Sources/Compound/List/ListRow.swift#L66

PR #4785, opened 2025-11-25
Update the date separators to use "Today", "Yesterday" etc for messages in the past 7 days.

pixlwave:
Luckily there aren't any affected snapshots as our mock date is a loooong time ago. For dates from the previous year the formatting remains unchanged, otherwise before and afters are like so:

| Before | After |
| - | - |
| IMG_1128 | Simulator Screenshot - iPhone 17 - 2025-11-25 at 17 51 14 |
| IMG_1129 | Simulator Screenshot - iPhone 17 - 2025-11-25 at 17 52 13 |

### DIFF
--- a/UnitTests/Sources/DeferredFulfillmentTests.swift
+++ b/UnitTests/Sources/DeferredFulfillmentTests.swift
@@ -9,6 +9,7 @@
 @testable import ElementX
 import XCTest
 
+@MainActor
 class DeferredFulfillmentTests: XCTestCase {
     private let observable = SomeObservable()
     
@@ -70,7 +71,8 @@ class DeferredFulfillmentTests: XCTestCase {
 
 // MARK: - Helpers
 
-@Observable private class SomeObservable {
+@Observable
+@MainActor private class SomeObservable {
     var counter = 0
     
     func setCounter(_ newValue: Int, delay: Duration? = nil) async throws {


ElementX/Sources/Other/Extensions/Date.swift
```diff
@@ -44,3 +65,15 @@ extension Date {
...
+    // There doesn't appear to be a way to get "Today" out of
+    // `Date.RelativeFormatStyle` so use the old way instead 😐
+    static let relative: DateFormatter = {
+        let formatter = DateFormatter()
+        formatter.doesRelativeDateFormatting = true
+        formatter.dateStyle = .long
+        formatter.timeStyle = .none
+        return formatter
+    }()
+}
```
pixlwave, line 79: I hate this, but couldn't figure out a single way to get "Today" out of `Date.FormatStyle` or `Date.RelativeFormatStyle` (and I didn't try too much with `Date.AnchoredRelativeFormatStyle` because it requires iOS 18+).

UnitTests/Sources/DeferredFulfillmentTests.swift
```diff
@@ -70,7 +71,8 @@ class DeferredFulfillmentTests: XCTestCase {
 
 // MARK: - Helpers
 
-@Observable private class SomeObservable {
+@Observable
+@MainActor private class SomeObservable {
```
Velin92, line 75: since this fixes the observed issue could we try to add back the iOS 26 implementation for the observed?

Velin92 (approved):

UnitTests/Sources/DeferredFulfillmentTests.swift
pixlwave: Yeah good point, will do it in a follow-up.

PR #4787, opened 2025-11-26
Single space members access option

Velin92:
This PR finally allows (under the space settings enabled FF) to use the Space Members access option.

However as of right now it only handles the selection when there is only 1 available space that can be chosen among either the joined parent spaces of the room, or the existing values.

This change will have some follow up PRs:
- handle the ask to join with space members access
- handle the multiple selection case
- handle the multiple selection cases for the ask to join with space members access.

### DIFF
--- a/ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenModels.swift
+++ b/ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenModels.swift
@@ -7,7 +7,6 @@
 //
 
 import Foundation
-import MatrixRustSDK
 
 enum SecurityAndPrivacyScreenViewModelAction {
     case displayEditAddressScreen
@@ -39,12 +38,9 @@ struct SecurityAndPrivacyScreenViewState: BindableState {
     var canEditHistoryVisibility = false
     var joinedParentSpaces: [SpaceRoomProxyProtocol] = []
     
+    /// The count of the intersection between the set of joined parent spaces and the set of spaces in the current access type
     var selectableSpacesCount: Int {
         Set(joinedParentSpaces.map(\.id) + currentSettings.accessType.spaceIDs).count
-    }
-    
-    private var desiredJoinedParentSpaces: [SpaceRoomProxyProtocol] {
-        joinedParentSpaces.filter { bindings.desiredSettings.accessType.spaceIDs.contains($0.id) }
     }
     
     private var hasChanges: Bool {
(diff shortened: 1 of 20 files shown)


ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenModels.swift
```diff
@@ -7,23 +7,42 @@
 //
 
 import Foundation
+import MatrixRustSDK
```
pixlwave: Is this used?

ElementX/Sources/Services/Spaces/SpaceServiceProxyProtocol.swift
```diff
@@ -19,4 +19,6 @@ protocol SpaceServiceProxyProtocol {
     
     func spaceRoomList(spaceID: String) async -> Result<SpaceRoomListProxyProtocol, SpaceServiceProxyError>
     func leaveSpace(spaceID: String) async -> Result<LeaveSpaceHandleProxy, SpaceServiceProxyError>
+    /// Returns all the parent spaces of a room that user has joined.
+    func joinedParents(roomID: String) async -> Result<[SpaceRoomProxyProtocol], SpaceServiceProxyError>
```
pixlwave: My only nit here would be that this can also give you the joined parents of a space too, which is why the SDK calls it `joinedParentsOfChild`. So maybe calling this `joinedParents(of childID: String)` would be more general going forward. :)

Otherwise we should at least clarify it in the doc comment with something like "of a room or space".

UnitTests/Sources/SecurityAndPrivacyScreenViewModelTests.swift
```diff
@@ -6,15 +6,103 @@
...
+        AppSettings.resetAllSettings()
+    }
+    
+    func testSetSingleJoinedSpaceMembersAccess() async throws {
+        let singleRoom = [SpaceRoomProxyProtocol].mockSingleRoom
+        let space = singleRoom[0]
+        setupViewModel(isSpaceSettingsEnabled: true, joinedParentSpaces: singleRoom, joinRule: .public)
+        
+        let deferred = deferFulfillment(context.$viewState) { $0.joinedParentSpaces.count == 1 }
+        try await deferred.fulfill()
```
pixlwave, line 35: I really don't like these deferred fulfillments when they're created after the thing you're trying to test. Could we maybe make `setupViewModel` async and put an expectation inside the SpaceServiceMock that we can await before returning? That way there will never be a race condition.

pixlwave (approved):
Very nice 👌

UnitTests/Sources/SecurityAndPrivacyScreenViewModelTests.swift
Velin92: Uhm I would keep as it is in this case because for these specific test cases we are awaiting the async call to complete, so is not a necessary fulfilment that would be called in all cases.

PR #4791, opened 2025-11-27
Show the send state in the room list.

pixlwave:
This PR makes 2 changes:

- Enable the newer latest event sorter.
- Decorate rooms in the room list with badges for Sending and Failed Sending states.

This is the iOS part for https://github.com/element-hq/element-meta/issues/2945 and https://github.com/element-hq/element-meta/issues/2946

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -10,7 +10,7 @@ end
 simulator_version = "26.1"
 
 before_all do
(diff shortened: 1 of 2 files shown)


fastlane/Fastfile
pixlwave, line 1: With `~>` we should always keep the patch (even if just at `0`) as it was trying to use Xcode 26.2 beta.

PR #4795, opened 2025-11-28
Make multi selection accessory leading

Velin92:
Name says it all

### DIFF
--- a/.github/workflows/compound-ios.yml
+++ b/.github/workflows/compound-ios.yml
@@ -32,7 +32,7 @@ jobs:
     - uses: nschloe/action-cached-lfs-checkout@f46300cd8952454b9f0a21a3d133d4bd5684cfc2 #v1.2.3
 
     - name: Configure Xcode 26
-      run: sudo xcode-select -s /Applications/Xcode_26.1_Release_Candidate.app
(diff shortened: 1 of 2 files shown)


compound-ios/Sources/Compound/List/ListRow.swift
```diff
@@ -218,12 +218,22 @@ private struct RowContent<Label: View, DetailsIcon: View>: View {
...
     
     var body: some View {
-        HStack(spacing: ListRowTrailingSectionSpacing.horizontal) {
-            label()
-                .frame(maxWidth: .infinity)
+        HStack(spacing: 0) {
+            // We should always have multi selection shown on the leading side
+            if let accessory, accessory.kind.isMultiSelection {
+                accessory
+                    .padding(.leading, ListRowPadding.horizontal)
```
pixlwave, line 227: Maybe worth a comment that we use `0` spacing and no trailing padding because `label` provides the necessary spacing via its leading padding. (Or something to that effect).

pixlwave (approved):
LGTM! 👌

.github/workflows/compound-ios.yml
pixlwave, line 1: Oh yeah good shout, I totally missed this.

PR #4796, opened 2025-11-28
Drop support for iOS 17

Velin92:
It also removes all the availability and unavailability checks referencing 17 and 18, and bumps the min supported version to 18.5


ElementX/Sources/Screens/Timeline/View/ItemMenu/TimelineItemMacContextMenu.swift
```diff
@@ -21,25 +21,19 @@ struct TimelineItemMacContextMenu: View {
             if let menuActions = actionProvider.makeActions() {
                 Section {
                     if !menuActions.reactions.isEmpty {
-                        if #available(iOS 17.0, *) {
```
pixlwave, line 24: Huh. I wonder why we didn't get a warning about this one, I thought Xcode was meant to point out unreachable `#available` checks 🤔

compound-ios/Sources/Compound/List/ListRow.swift
pixlwave, line 1: Given this is a 1-liner, I think we should probable replace any `.isToggle()`s with `.accessibilityAddTraits(.isToggle)` and remove this extension.

PR #4798, opened 2025-11-28
Manage multiple spaces members access

### LINKED ISSUE #4681: [Task] Update the Security & Privacy screen to work with Spaces.

Velin92:
fixes #4681 

Implemented a new view called ManageAuthorizedSpaces that is used when multiple spaces are available to be selected.
Also added some tests for the multiple selection behaviour

### DIFF
--- a/ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenViewModel.swift
+++ b/ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenViewModel.swift
@@ -245,6 +245,8 @@ class SecurityAndPrivacyScreenViewModel: SecurityAndPrivacyScreenViewModelType, 
             state.bindings.desiredSettings.accessType = .spaceUsers(spaceIDs: [joinedParent.id])
         case .singleUnknown(let id):
             state.bindings.desiredSettings.accessType = .spaceUsers(spaceIDs: [id])
+        case .empty:
+            break // Very edge case. We do nothing in this case.
         case .multiple:
             displayManageAuthorizedSpacesScreen()
         }
--- a/UnitTests/Sources/SecurityAndPrivacyScreenViewModelTests.swift
+++ b/UnitTests/Sources/SecurityAndPrivacyScreenViewModelTests.swift
@@ -39,7 +39,7 @@ class SecurityAndPrivacyScreenViewModelTests: XCTestCase {
         XCTAssertTrue(context.viewState.isSaveDisabled)
         XCTAssertTrue(context.viewState.isSpaceMembersOptionSelectable)
         guard case .singleJoined = context.viewState.spaceSelection else {
-            XCTFail("Expected spaceSelection to be .singleSpace")
+            XCTFail("Expected spaceSelection to be .singleJoined")
             return
         }
         
@@ -72,7 +72,7 @@ class SecurityAndPrivacyScreenViewModelTests: XCTestCase {
         XCTAssertNil(context.viewState.accessSectionFooter)
         XCTAssertTrue(context.viewState.isSaveDisabled)
         guard case .singleUnknown = context.viewState.spaceSelection else {
-            XCTFail("Expected spaceSelection to be .singleSpace")
+            XCTFail("Expected spaceSelection to be .singleUnknown")
             return
         }
         
@@ -84,7 +84,7 @@ class SecurityAndPrivacyScreenViewModelTests: XCTestCase {
         XCTAssertTrue(context.viewState.isSaveDisabled)
         XCTAssertEqual(context.desiredSettings.accessType, .spaceUsers(spaceIDs: [space.id]))
         guard case .singleUnknown = context.viewState.spaceSelection else {
-            XCTFail("Expected spaceSelection to be .singleSpace")
+            XCTFail("Expected spaceSelection to be .singleUnknown")
             return
         }
     }
@@ -100,7 +100,7 @@ class SecurityAndPrivacyScreenViewModelTests: XCTestCase {
         XCTAssertTrue(context.viewState.isSaveDisabled)
         XCTAssertTrue(context.viewState.isSpaceMembersOptionSelectable)
         guard case .multiple = context.viewState.spaceSelection else {
-            XCTFail("Expected spaceSelection to be .singleSpace")
+            XCTFail("Expected spaceSelection to be .multiple")
             return
         }
         
@@ -146,7 +146,7 @@ class SecurityAndPrivacyScreenViewModelTests: XCTestCase {
         XCTAssertTrue(context.viewState.isSaveDisabled)
         XCTAssertTrue(context.viewState.isSpaceMembersOptionSelectable)
         guard case .multiple = context.viewState.spaceSelection else {
-            XCTFail("Expected spaceSelection to be .singleSpace")
+            XCTFail("Expected spaceSelection to be .multiple")
             return
         }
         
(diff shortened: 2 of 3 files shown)


ElementX/Sources/FlowCoordinators/SpaceSettingsFlowCoordinator.swift
```diff
@@ -1,4 +1,3 @@
-//
```
pixlwave, line 1: I think this one is accidental.

ElementX/Sources/Screens/ManageAuthorizedSpacesScreen/ManageAuthorizedSpacesScreenModels.swift
```diff
@@ -0,0 +1,44 @@
...
+enum ManageAuthorizedSpacesScreenViewModelAction {
+    case dismiss
+}
+
+struct ManageAuthorizedSpacesScreenViewState: BindableState {
+    let authorizedSpacesSelection: AuthorizedSpacesSelection
+    var desiredSelectedIDs: Set<String>
+    
+    var hasChanges: Bool {
+        authorizedSpacesSelection.currentSelectedIDs != desiredSelectedIDs
```
pixlwave: The names are a bit funky here as `current` could be interpreted as what is shown on the screen, but I'm struggling to come up with much better. Maybe just simply

```suggestion
        authorizedSpacesSelection.initialSelectedIDs != selectedIDs
```

ElementX/Sources/Screens/ManageAuthorizedSpacesScreen/View/ManageAuthorizedSpacesScreen.swift
```diff
@@ -0,0 +1,117 @@
...
+                            context.send(viewAction: .toggle(spaceID: space.id))
+                        })
+            }
+        } header: {
+            Text(L10n.screenManageAuthorizedSpacesYourSpacesSectionTitle)
+                .compoundListSectionHeader()
+        }
+    }
+    
+    private var unkwnownSpacesSection: some View {
```
pixlwave:
```suggestion
    private var unknownSpacesSection: some View {
```

ElementX/Sources/Screens/ManageAuthorizedSpacesScreen/View/ManageAuthorizedSpacesScreen.swift
```diff
@@ -0,0 +1,117 @@
...
+    
+    var body: some View {
+        Form {
+            header
+            if !context.viewState.authorizedSpacesSelection.joinedParentSpaces.isEmpty {
+                joinedParentsSection
+            }
+            if !context.viewState.authorizedSpacesSelection.unknownSpacesIDs.isEmpty {
+                unkwnownSpacesSection
+            }
```
pixlwave, line 24: Maybe add a bit of whitespace so its easier to read :)
```suggestion
            header
            
            if !context.viewState.authorizedSpacesSelection.joinedParentSpaces.isEmpty {
                joinedParentsSection
            }
            
            if !context.viewState.authorizedSpacesSelection.unknownSpacesIDs.isEmpty {
                unkwnownSpacesSection
            }
```

ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenViewModel.swift
```diff
@@ -236,15 +235,36 @@ class SecurityAndPrivacyScreenViewModel: SecurityAndPrivacyScreenViewModelType,
     }
     
     private func handleSelectedSpaceMembersAccess() {
-        switch context.viewState.spaceSelection {
+        guard !state.bindings.desiredSettings.accessType.isSpaceUsers else {
+            return
```
pixlwave, line 240: Probably worth logging an error here.

ElementX/Sources/Mocks/SpaceRoomProxyMock.swift
```diff
@@ -100,6 +100,32 @@ extension [SpaceRoomProxyProtocol] {
         ]
     }
     
+    static var mockJoinedSpaces2: [SpaceRoomProxyMock] {
```
pixlwave, line 103: To reduce the boilerplate a bit could we remove this and simply do the following in the previews/tests?

```
.mockJoinedSpaces.prefix(3)
```

I am pretty sure that adding the aliases to those existing mocks is unlikely to change any existing snapshots (but if it does, I don't see an issue with it).

UnitTests/Sources/SecurityAndPrivacyScreenViewModelTests.swift
```diff
@@ -88,6 +89,96 @@ class SecurityAndPrivacyScreenViewModelTests: XCTestCase {
...
+        guard case .multiple = context.viewState.spaceSelection else {
+            XCTFail("Expected spaceSelection to be .singleSpace")
+            return
+        }
+        
+        var desiredSpaceIDs: PassthroughSubject<Set<String>, Never>!
+        let deferredAction = deferFulfillment(viewModel.actionsPublisher) { action in
+            switch action {
+            case .displayManageAuthorizedSpacesScreen(let authorizedSpacesSelection):
+                defer { desiredSpaceIDs = authorizedSpacesSelection.desiredSelectIDs }
```
pixlwave: You don't actually need to do this here.

UnitTests/Sources/SecurityAndPrivacyScreenViewModelTests.swift
```diff
@@ -88,6 +89,96 @@ class SecurityAndPrivacyScreenViewModelTests: XCTestCase {
...
+                return authorizedSpacesSelection.joinedParentSpaces.map(\.id) == spaces.map(\.id) &&
+                    authorizedSpacesSelection.unknownSpacesIDs.isEmpty &&
+                    authorizedSpacesSelection.currentSelectedIDs.isEmpty
+            default:
+                return false
+            }
+        }
+        context.send(viewAction: .selectedSpaceMembersAccess)
+        try await deferredAction.fulfill()
+        desiredSpaceIDs.send([spaces[0].id])
```
pixlwave: Instead you can do it here along the lines of:

```suggestion
        let action = try await deferredAction.fulfill()
        guard case .displayManageAuthorizedSpacesScreen(let authorizedSpacesSelection) = action else { fail }
        authorizedSpacesSelection.desiredSpaceIDs.send([spaces[0].id])
```

I think that would be easier to follow than adding side-effects in the fulfilment.

UnitTests/Sources/SecurityAndPrivacyScreenViewModelTests.swift
```diff
@@ -88,6 +89,96 @@ class SecurityAndPrivacyScreenViewModelTests: XCTestCase {
...
+        
+        XCTAssertTrue(context.viewState.currentSettings.accessType.isSpaceUsers)
+        XCTAssertTrue(context.viewState.isSaveDisabled)
+        XCTAssertTrue(context.viewState.isSpaceMembersOptionSelectable)
+        guard case .multiple = context.viewState.spaceSelection else {
+            XCTFail("Expected spaceSelection to be .singleSpace")
+            return
+        }
+        
+        var desiredSpaceIDs: PassthroughSubject<Set<String>, Never>!
```
pixlwave: Same comment on this one.

ElementX/Sources/Other/SwiftUI/Views/ToolbarButton.swift
```diff
@@ -0,0 +1,48 @@
...
+    var body: some View {
+        if #available(iOS 26, *) {
+            Button(action: action) {
+                role.icon
+                    .accessibilityLabel(role.title)
+            }
+        } else {
+            Button(role.title, action: action)
+        }
+    }
```
pixlwave, line 64: When enabled and we're using `\.check` the button should be [tinted](https://www.figma.com/design/jvZ7TIyd7V9Vn8sIDB9vMd/Compound-iOS-Templates?node-id=3779-88089&m=dev) too.

ElementX/Sources/Other/SwiftUI/Views/ToolbarButton.swift
```diff
@@ -0,0 +1,48 @@
...
+//
+
+import Compound
+import SwiftUI
+
+struct ToolbarButton: View {
+    enum Role {
+        case cancel
+        case done
+
```
pixlwave, line 16: Should we add a case for `save` too?

ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenViewModel.swift
Velin92: actually this is not an error, is just because if the user taps the button again we want to do nothing but we do not want to disable the button, I'll add a comment

UnitTests/Sources/SecurityAndPrivacyScreenViewModelTests.swift
Velin92: why? I need the subject to set the value (simulating the press of the done button) and then check if the save has been called with the right values.

UnitTests/Sources/SecurityAndPrivacyScreenViewModelTests.swift
Velin92: Okay I see what you mean, it makes sense

ElementX/Sources/Other/SwiftUI/Views/ToolbarButton.swift
Velin92: yeah why not, in the end it should look exactly like the done one, the only thing that changes is the label

PR #4811, opened 2025-12-03
Add a Link New Device screen (behind a feature flag).

pixlwave:

- Add a new button in the Settings screen to link a new device.
- Add a Link New Device screen that asks the user whether they're linking a phone or a computer (which determines whether we show a QR code or scan one.


ElementX/Resources/Localizations/en.lproj/Untranslated.strings
pixlwave, line 1: Added in the next commit and I'll squash on merge so did seem worthwhile editing this commit.

ElementX/Sources/Screens/Settings/SettingsScreen/SettingsScreenViewModel.swift
```diff
@@ -25,6 +25,7 @@ class SettingsScreenViewModel: SettingsScreenViewModelType, SettingsScreenViewMo
         
         super.init(initialViewState: .init(deviceID: userSession.clientProxy.deviceID,
                                            userID: userSession.clientProxy.userID,
+                                           showLinkNewDeviceButton: appSettings.newQRCodeLoginFlowsEnabled, // FIXME: Also needs server detection.
```
pixlwave: Requires https://github.com/matrix-org/matrix-rust-sdk/pull/5925

pixlwave:
Converting to draft while waiting for the PR in the SDK.

ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -399,6 +400,9 @@ final class AppSettings {
     @UserPreference(key: UserDefaultsKeys.linkPreviewsEnabled, defaultValue: false, storageType: .userDefaults(store))
     var linkPreviewsEnabled
     
+    @UserPreference(key: UserDefaultsKeys.newQRCodeLoginFlowsEnabled, defaultValue: false, storageType: .userDefaults(store))
+    var newQRCodeLoginFlowsEnabled
```
Velin92: Instead of being newQRCodeLoginFlowsEnabled shouldn't it be something like `linkDeviceWithOREnabled` ?

UnitTests/Sources/LinkNewDeviceScreenViewModelTests.swift
```diff
@@ -0,0 +1,13 @@
...
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+// Please see LICENSE files in the repository root for full details.
+//
+
+import XCTest
+
+@testable import ElementX
+
+@MainActor
+class LinkNewDeviceScreenViewModelTests: XCTestCase { }
```
Velin92: I guess you can remove this file

ElementX/Sources/Application/Settings/AppSettings.swift
pixlwave: Oh yep, I added this before really understanding everything that was needed, good point 👍
https://github.com/element-hq/element-x-ios/pull/4811/commits/ceb0cd68bdffd3f7acfd81a7ce01c01cacba64b1

UnitTests/Sources/LinkNewDeviceScreenViewModelTests.swift
pixlwave: Oh yeah good spot, the preview tests were enough in the end but I forgot to come back to remove this.

PR #4812, opened 2025-12-03
Ask to join restricted/space members access option

Velin92:
This option is only available when the knock FF is enabled and also the restricted/space members option is available, and will take place over the normal ask to join option (but only if such option was not already selected, in which case both will appear).
It essentially works the same way of the restricted/space members option, handling both a single and multiple case scenario.

Also added previews and tests for its logic.

### DIFF
--- a/ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenModels.swift
+++ b/ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenModels.swift
@@ -67,7 +67,7 @@ struct SecurityAndPrivacyScreenViewState: BindableState {
     }
     
     var isSpaceMembersOptionAvailable: Bool {
-        currentSettings.accessType.isSpaceUsers || isSpaceMembersOptionSelectable
+        currentSettings.accessType.isSpaceMembers || isSpaceMembersOptionSelectable
     }
     
     var isSpaceMembersOptionSelectable: Bool {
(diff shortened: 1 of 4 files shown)


ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenModels.swift
```diff
@@ -185,6 +210,15 @@ enum SecurityAndPrivacyRoomAccessType: Equatable {
         }
     }
     
+    var isAskToJoinWithSpaceUsers: Bool {
+        switch self {
+        case .askToJoinWithSpaceUsers:
```
stefanceriu: You should probably replace `users` with `members` in `askToJoinWithSpaceUsers` and `spaceUsers`.

ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenModels.swift
```diff
@@ -215,6 +249,7 @@ enum SecurityAndPrivacyScreenViewAction {
     case tryUpdatingEncryption(Bool)
     case editAddress
     case selectedSpaceMembersAccess
+    case selectedAskToJoinWithSpaceMembersAccess
```
stefanceriu, line 252: Like here, `SpaceMembers`

ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenViewModel.swift
```diff
@@ -248,11 +250,29 @@ class SecurityAndPrivacyScreenViewModel: SecurityAndPrivacyScreenViewModelType,
...
+            return
+        }
+        
+        switch state.spaceSelection {
+        case .singleJoined(let joinedParent):
+            state.bindings.desiredSettings.accessType = .askToJoinWithSpaceUsers(spaceIDs: [joinedParent.id])
+        case .singleUnknown(let id):
+            state.bindings.desiredSettings.accessType = .askToJoinWithSpaceUsers(spaceIDs: [id])
+        case .empty:
+            break // Very edge case. We do nothing in this case.
```
stefanceriu, line 269: Should this be a fatal error then?

Velin92: it's okay, we don't want the app to crash, it's a very edge case that we will just not handle.

PR #4832, opened 2025-12-09
Bump the RustSDK to v25.12.09

stefanceriu:


ElementX/Sources/Mocks/Generated/SDKGeneratedMocks.swift
pixlwave, line 1: Wooo `toBytes` is available, I can use this tomorrow 😎

PR #4834, opened 2025-12-10
Fix issues with the last message in the room list being cleared, going missing or being slow to update.

pixlwave:
… by updating the SDK to include https://github.com/matrix-org/matrix-rust-sdk/pull/5932 amongst others.


stefanceriu (approved):
👌

PR #4835, opened 2025-12-10
Improve canSee Sec & Privacy check.

Velin92:
In this way the screen is only displayed if at least one option is available given the context of the JoinRule or if the room is also a space.


pixlwave (approved):
👌

PR #4836, opened 2025-12-10
Remove `eraseToStream` now that `any AsyncSequence` is available to us.

pixlwave:
This PR silences some deprecation warnings, the main one being that we no longer need to use `eraseToStream` now that we've dropped iOS 17 as `any AsyncSequence` is now available to us. I decided not to also jump onto the `Observations` bandwagon just yet, as I'd like to see if this is stable on CI and not make the 2 changes at once (given we previously tried `Observations` and reverted it pretty quickly).

### DIFF
--- a/ElementX/Sources/Services/Media/MediaUploadingPreprocessor.swift
+++ b/ElementX/Sources/Services/Media/MediaUploadingPreprocessor.swift
@@ -396,7 +396,7 @@ struct MediaUploadingPreprocessor {
     /// - Parameter url: the video URL
     /// - Returns: the URL for the resulting thumbnail and its sizing info as an `ImageProcessingResult`
     private func generateThumbnailForVideoAt(_ url: URL) async throws(MediaUploadingPreprocessorError) -> ImageProcessingInfo {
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Other/Extensions/Observable.swift
```diff
@@ -7,7 +7,7 @@
 //
 
 import Foundation
-import Mutex
+import Synchronization
 
 extension Observable {
     /// Creates an async stream for the specified property on this object. We probably won't need this once SE-0475 is available:
```
Velin92, line 13: Can we try the iOS 26/ Swift 6 implementation and check if that works too?

pixlwave: > I decided not to also jump onto the Observations bandwagon just yet, as I'd like to see if this is stable on CI and not make the 2 changes at once (given we previously tried Observations and reverted it pretty quickly).

Can I do it next week given this thought process? 👆

Velin92 (approved):

PR #4838, opened 2025-12-11
Fix the integration test and improve the EncryptionSettings UI ones.

stefanceriu:
The secure backup disabling sheet doesn't show up in the snapshots. No idea what the actual problem is though, completely unreproducible locally.


ElementX/Sources/Screens/SecureBackup/SecureBackupScreen/SecureBackupScreenViewModel.swift
pixlwave, line 1: 👌

PR #4842, opened 2025-12-11
Handle the edge case of one or more non parent joined spaces present in the existing allowed list

Velin92:
This applies to both the `restricted` and `knockRestricted` join rules.
Essentially what was happening is that if the existing room access type contains a space that is not a parent of the room, but is still joined by the current user trying to edit its join rule, we displayed such space as unknown.
However since the user is actually part of that space, they should be able to visualise their display name, and it should appear in the list of the spaces they are a member of.

So we added a new API that checks the availability of a joined space room given their id, and we cross check with all the ids that were not resolved to not be joined parents, and if is at least a joined room, we add it to the selection list.
We still only allow by default to add in the allowed list spaces that are joined and parent of the room plus any other value available on the current allowed list, the only difference is that if one of these values is a joined space, we display its info to the user

This PR also updates the SDK to use the newly defined `getSpace` function.
I also renamed joinedParents to joinedSpaces, since now that list could also contain a non parent joined space.

### DIFF
--- a/ElementX/Sources/Screens/ManageAuthorizedSpacesScreen/ManageAuthorizedSpacesScreenModels.swift
+++ b/ElementX/Sources/Screens/ManageAuthorizedSpacesScreen/ManageAuthorizedSpacesScreenModels.swift
@@ -37,7 +37,7 @@ enum ManageAuthorizedSpacesScreenViewAction {
 }
 
 struct AuthorizedSpacesSelection {
-    let joinedParentSpaces: [SpaceRoomProxyProtocol]
+    let joinedSpaces: [SpaceRoomProxyProtocol]
     let unknownSpacesIDs: [String]
     let initialSelectedIDs: Set<String>
     let selectedIDs: PassthroughSubject<Set<String>, Never> = .init()
--- a/ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenViewModel.swift
+++ b/ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenViewModel.swift
@@ -47,12 +47,7 @@ class SecurityAndPrivacyScreenViewModel: SecurityAndPrivacyScreenViewModelType, 
         setupRoomDirectoryVisibility()
         setupSubscriptions()
         Task {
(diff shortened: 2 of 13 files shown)


ElementX/Sources/Screens/ManageAuthorizedSpacesScreen/ManageAuthorizedSpacesScreenModels.swift
```diff
@@ -37,7 +37,7 @@ enum ManageAuthorizedSpacesScreenViewAction {
 }
 
 struct AuthorizedSpacesSelection {
-    let joinedParentSpaces: [SpaceRoomProxyProtocol]
+    let joinedSpaces: [SpaceRoomProxyProtocol]
```
Velin92, line 40: This by default are the joined parent spaces of the room however they MAY contain also a non parent joined space... Maybe I should rename this to `selectableJoinedSpaces`, or something like that?

stefanceriu: Maybe a comment along these lines would be more appropriate. I don't think we can explain it well through a var name.

ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenViewModel.swift
```diff
@@ -291,6 +288,28 @@ class SecurityAndPrivacyScreenViewModel: SecurityAndPrivacyScreenViewModelType,
...
+        for spaceID in state.currentSettings.accessType.spaceIDs where !joinedParentSpaces.contains(where: { $0.id == spaceID }) {
+            if case let .success(.some(space)) = await clientProxy.spaceService.spaceForIdentifier(spaceID: spaceID) {
+                nonParentJoinedSpaces.append(space)
+            }
+        }
+        
+        // By default we only want to allow selection among joined parents but
+        // if there is a non parent joined space already set in the access type
+        // we also include it in the known spaces selection list.
+        state.joinedSpaces = joinedParentSpaces + nonParentJoinedSpaces
```
stefanceriu: 👍

stefanceriu (approved):
Noice, lgtm 👍

PR #4738, opened 2025-11-12
Add a banner to encrypted rooms with visible history.

kaylendog:
- Adds a dismissable alert that is displayed whenever the user opens a room with `history_visibility` != `joined`. When cleared, this is recorded in the app's data store.
- When opening a room with `history_visibility` = `joined`, this flag is cleared.

Issue: element-hq/element-meta#2875

### Pull Request Checklist

- [x] I read the [contributing guide](https://github.com/element-hq/element-ios/blob/develop/CONTRIBUTING.md).
- [x] Pull request contains a [changelog label](https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md#changelog).

**UI changes have been tested with:**
- [x] iPhone and iPad simulators in portrait and landscape orientations.
- [x] Dark mode enabled and disabled.
- [x] Various sizes of dynamic type.
- [ ] Voiceover enabled - I don't believe I can do this from the simulator.

PR #4851, opened 2025-12-12
Add a banner to encrypted rooms with visible history.

kaylendog:
Take no. 2 - conditioned the banner appearance on the feature flag, added another unit test that verifies the incorrect behaviour reported no longer happens.

### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
+++ b/ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
@@ -353,8 +353,7 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol 
             if isHistoryVisible, !isHistoryVisibleBannerAcknowledged {
                 // Whenever the user opens an encrypted room with shared/world-readable history visbility, we show them a warning banner if they have not already dismissed it.
                 state.historyVisibleDetails = .historyVisible(learnMoreURL: appSettings.historySharingDetailsURL)
-            }
-            if !isHistoryVisible, isHistoryVisibleBannerAcknowledged {
+            } else if !isHistoryVisible, isHistoryVisibleBannerAcknowledged {
                 // Whenever the user opens a room with non-shared history visibility, we clear the dismiss flag to ensure that the banner is displayed again if the history is made visible in the future.
                 appSettings.acknowledgedHistoryVisibleRooms.remove(roomInfo.id)
                 state.historyVisibleDetails = nil


pixlwave:
Any chance you could squash everything from #4738 into a single commit so that it's easy to see what was reviewed already and the changes that this PR will make on top of that? 😇

UnitTests/Sources/RoomScreenViewModelTests.swift
```diff
@@ -440,8 +440,41 @@ class RoomScreenViewModelTests: XCTestCase {
...
+        
+        let deferred = deferFulfillment(viewModel.context.$viewState) { state in
+            state.footerDetails == nil
+        }
+        try await deferred.fulfill()
+    }
 
     func testHistoryVisibleBannerDoesNotAppearIfNotEncrypted() async throws {
+        ServiceLocator.shared.settings.enableKeyShareOnInvite = true
+        ServiceLocator.shared.settings.acknowledgedHistoryVisibleRooms = Set()
```
kaylendog, line 474: I'm not convinced resetting the set every test is necessary, but after I set prod on fire I don't want to take any chances.

kaylendog:
> Any chance you could squash everything from #4738 into a single commit so that it's easy to see what was reviewed already and the changes that this PR will make on top of that? 😇

@pixlwave should be done!

UnitTests/Sources/RoomScreenViewModelTests.swift
stefanceriu: > after I set prod on fire

Just for the record, it only made it to nightly before we reverted it, and while we generally aim not to break nightly either this was more of a minor annoyance.

kaylendog:
I've had a look through the CI logs, but can't actually see any unit test that's failing - either I'm blind or that's a false negative?

pixlwave:
It was just a flakey test, searching for `::error` usually jumps you straight to them. I've re-run and it is now green.

Given there's a small conflict with #4861 (it also includes the element.io URL) and that it improves the copy, I'd like to wait for that to merge and for this to be rebased before I review these changes again if you don't mind?

UnitTests/Sources/RoomScreenViewModelTests.swift
pixlwave: 💯 – this is the entire point of Nightly builds and instability should be expected with them from time to time.

pixlwave:
FYI #4861 has been merged.

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -342,6 +345,18 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
...
         }
+        
+        if appSettings.enableKeyShareOnInvite {
+            // Whever the user opens a room with joined history visibility, we clear the dismiss flag to ensure that the banner is displayed again if the history is made visible in the future.
+            if roomInfo.historyVisibility == RoomHistoryVisibility.joined {
+                appSettings.acknowledgedHistoryVisibleRooms.remove(roomInfo.id)
+                state.footerDetails = nil
+            }
+            // Whenever the user opens an encrypted room with non-join history visbility, we show them a warning banner if they have not already dismissed it.
+            else if appSettings.enableKeyShareOnInvite, roomInfo.isEncrypted, !appSettings.acknowledgedHistoryVisibleRooms.contains(roomInfo.id) {
```
pixlwave: We probably don't need the additional `appSettings.enableKeyShareOnInvite` check given we're already inside an `if appSettings.enableKeyShareOnInvite` :)

Also, please can we format the inner `if else` as

```
if condition {
    // comment
    code
} else {
    // comment
    code
}
```

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -342,6 +345,18 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
...
+                state.footerDetails = .historyVisible(learnMoreURL: appSettings.historySharingDetailsURL)
```
pixlwave: There's a race-condition on this against the identity violations:

https://github.com/element-hq/element-x-ios/blob/581c82dd0202e87de8633fc2a5a4d03d6af7a898/ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift#L246-L254

You can see it if you watch this video frame by frame. This history banner is shown momentarily and then cleared by the identity checks:

https://github.com/user-attachments/assets/7fde910e-da1f-4306-aab4-f5cc4de6ab13

We're going to need to allow setting both types of banner simultaneously in the view state and then pick a priority in the view to decide which banner should be shown when both are set.

kaylendog: Provisional fix in https://github.com/element-hq/element-x-ios/pull/4851/commits/9929b2f86623abcfff07d5548dbc644d367a8726

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
kaylendog: Addressed in https://github.com/element-hq/element-x-ios/pull/4851/commits/9929b2f86623abcfff07d5548dbc644d367a8726

UnitTests/Sources/RoomScreenViewModelTests.swift
```diff
@@ -438,4 +438,214 @@ class RoomScreenViewModelTests: XCTestCase {
...
+                                            appHooks: AppHooks(),
+                                            analyticsService: ServiceLocator.shared.analytics,
+                                            userIndicatorController: ServiceLocator.shared.userIndicatorController)
+        
+        self.viewModel = viewModel
+        
+        let deferred = deferFulfillment(viewModel.context.$viewState) { state in
+            state.footerDetails == nil
+        }
+        try await deferred.fulfill()
```
pixlwave: Given that `footerDetails` will start with a value of `nil`, we have a slightly more helpful assertion we can use here:

```
let deferred = deferFailure(viewModel.context.$viewState, timeout: 1) { $0.footerDetails != nil }
```

which waits around for 1-second to make sure `footerDetails` never becomes non-nil. Applies to a few more tests below too :)

ElementX/Sources/Screens/RoomScreen/View/RoomScreenFooterView.swift
```diff
@@ -166,12 +201,15 @@ struct RoomScreenFooterView_Previews: PreviewProvider, TestablePreview {
...
+    static let historyVisibleDetails: RoomScreenFooterViewDetails = .historyVisible(learnMoreURL: "https://element.io")
+    
     static var previews: some View {
         RoomScreenFooterView(details: bobDetails, mediaProvider: MediaProviderMock(configuration: .init())) { _ in }
             .previewDisplayName("With displayname")
         RoomScreenFooterView(details: noNameDetails, mediaProvider: MediaProviderMock(configuration: .init())) { _ in }
             .previewDisplayName("Without displayname")
         RoomScreenFooterView(details: verificationViolationDetails, mediaProvider: MediaProviderMock(configuration: .init())) { _ in }
             .previewDisplayName("Verification Violation")
+        RoomScreenFooterView(details: historyVisibleDetails, mediaProvider: MediaProviderMock(configuration: .init())) { _ in }.previewDisplayName("History Visible")
```
pixlwave: There's something wrong with our CI, it should have failed because of missing snapshots for this (sorry, didn't notice this last time around).

To generate a snapshot, you need:
- An iPhone SE (3rd gen) simulator configured with en-US as the language.
- Git LFS installed and configured in the repo (done automatically if you used `swift run tools setup-project`)
- To select the Preview tests scheme:
    Screenshot 2025-12-17 at 5 09 36 pm
- To run the corresponding test in the Tests sidebar:
    Screenshot 2025-12-17 at 5 11 35 pm

Then you can commit the generated files (making sure they're under LFS) and CI will check them on each run.

Less important, just a formatting fix for consistency with the rest of the previews:

```suggestion
        RoomScreenFooterView(details: historyVisibleDetails, mediaProvider: MediaProviderMock(configuration: .init())) { _ in }
            .previewDisplayName("History Visible")
```

ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
```diff
@@ -80,7 +80,16 @@ struct RoomScreenViewState: BindableState {
...
+    var identityViolationDetails: RoomScreenFooterViewDetails?
+    var historyVisibileDetails: RoomScreenFooterViewDetails?
+    
+    var footerDetails: RoomScreenFooterViewDetails? {
+        if identityViolationDetails != nil {
+            return identityViolationDetails
+        } else {
+            return historyVisibileDetails
+        }
+    }
```
pixlwave, line 88: We can simplify this:

```suggestion
    var footerDetails: RoomScreenFooterViewDetails? {
        identityViolationDetails ?? historyVisibileDetails
    }
```

ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
```diff
@@ -80,7 +80,16 @@ struct RoomScreenViewState: BindableState {
             (canAcceptKnocks || canDeclineKnocks || canBan)
     }
     
-    var footerDetails: RoomScreenFooterViewDetails?
+    var identityViolationDetails: RoomScreenFooterViewDetails?
+    var historyVisibileDetails: RoomScreenFooterViewDetails?
```
pixlwave:
```suggestion
    var historyVisibleDetails: RoomScreenFooterViewDetails?
```

pixlwave (commented):
My bad, I noticed a missing snapshot too. Details below.

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
pixlwave: Nice 👌

ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
kaylendog: Addressed in https://github.com/element-hq/element-x-ios/pull/4851/commits/806067e19edb6d52623fda8a24363da24bac354e

ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
kaylendog: Addressed in https://github.com/element-hq/element-x-ios/pull/4851/commits/806067e19edb6d52623fda8a24363da24bac354e

ElementX/Sources/Screens/RoomScreen/View/RoomScreenFooterView.swift
kaylendog: Snapshots here https://github.com/element-hq/element-x-ios/pull/4851/commits/21fbb2451c541956e48af97843d6bb091a77c7a8

Formatting fix in https://github.com/element-hq/element-x-ios/pull/4851/commits/806067e19edb6d52623fda8a24363da24bac354e

UnitTests/Sources/RoomScreenViewModelTests.swift
kaylendog: Addressed in https://github.com/element-hq/element-x-ios/pull/4851/commits/afdecb98d4643f313868a2a75e5bb65cc0c19701

pixlwave (commented):
Me again! It turns out that whilst the banner is behind the feature flag now, the original bug that we were seeing first time around still needs to be fixed: The banner is shown when the history is set to "Members since invited"

e.g.

https://github.com/user-attachments/assets/e8a5ba28-9c42-41bf-afb5-eb9b1fdf12fe

kaylendog:
> Me again! It turns out that whilst the banner is behind the feature flag now, the bug that we were seeing first time around still needs to be fixed: The banner is shown when the history is set to "Members since invited"
> 
> e.g.
> Simulator.Screen.Recording.-.iPhone.17.-.2025-12-18.at.10.10.33.mov

Afaik this is currently intentional, i.e. the banner should show for any non-`joined` room - I'll ask the crypto team about it.

UnitTests/Sources/RoomScreenViewModelTests.swift
```diff
@@ -438,4 +438,208 @@ class RoomScreenViewModelTests: XCTestCase {
...
+                                        analyticsService: ServiceLocator.shared.analytics,
+                                        userIndicatorController: ServiceLocator.shared.userIndicatorController)
+        deferred = deferFulfillment(viewModel.context.$viewState) { state in
+            state.footerDetails != nil
+        }
+        try await deferred.fulfill()
+    }
 }
+
+g
```
kaylendog: ??????? how does something like this even happen

richvdh:
> The banner is shown when the history is set to "Members since invited"

This sounds like correct behaviour to me.

The point is still that messages you send today may in future be shared with people who are not yet invited-to or joined-to the room

richvdh:
> > The banner is shown when the history is set to "Members since invited"
> 
> This sounds like correct behaviour to me.
> 
> The point is still that messages you send today may in future be shared with people who are not yet invited-to or joined-to the room

Sorry, ignore this.

History may be shared with new members when `history_visibility` is set to `shared` or `world_readable`, so the banner should not be shown if it is set to `invited` or `joined`.

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -346,14 +346,18 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
...
+        let isHistoryVisibleBannerAcknowledged = appSettings.acknowledgedHistoryVisibleRooms.contains(roomInfo.id)
+
         if appSettings.enableKeyShareOnInvite, roomInfo.isEncrypted {
-            if roomInfo.historyVisibility == RoomHistoryVisibility.joined {
-                // Whever the user opens a room with joined history visibility, we clear the dismiss flag to ensure that the banner is displayed again if the history is made visible in the future.
+            if isHistoryVisible, !isHistoryVisibleBannerAcknowledged {
+                // Whenever the user opens an encrypted room with shared/world-readable history visbility, we show them a warning banner if they have not already dismissed it.
+                state.historyVisibleDetails = .historyVisible(learnMoreURL: appSettings.historySharingDetailsURL)
+            }
+            if !isHistoryVisible, isHistoryVisibleBannerAcknowledged {
```
pixlwave: I think this can be an `else if` right?

```suggestion
            } else if !isHistoryVisible, isHistoryVisibleBannerAcknowledged {
```

pixlwave (commented):
Ok last question (I promise!)

Are we also intentionally showing this banner inside every DM too?

https://github.com/user-attachments/assets/606af99f-0935-499b-a6e6-5c46f3c30c1d

richvdh:
> Are we also intentionally showing this banner inside every DM too?

DMs are subject to the same rules as non-DMs: it depends on their history visibility settings. 

Note that it's not obvious that the default history visibility settings for DMs are sensible: https://github.com/element-hq/element-meta/issues/3109

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
kaylendog: It can, but the conditions are mutually exclusive anyway - I can use `else if` for clarity?

pixlwave:
> Note that it's not obvious that the default history visibility settings for DMs are sensible: [element-hq/element-meta#3109](https://github.com/element-hq/element-meta/issues/3109)

Ah ha, understood 👍

pixlwave (approved):
Thanks @kaylendog, sorry this ended up being a long review.

PR #4861, opened 2025-12-15
Update history visible settings

stefanceriu:
New copy, new order and a pretty new footer.

Fixes https://github.com/element-hq/element-x-ios/issues/4852

### DIFF
--- a/ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenModels.swift
+++ b/ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenModels.swift
@@ -16,18 +16,21 @@ enum SecurityAndPrivacyScreenViewModelAction {
 
 struct SecurityAndPrivacyScreenViewState: BindableState {
     let serverName: String
+    
     var currentSettings: SecurityAndPrivacySettings
     var bindings: SecurityAndPrivacyScreenViewStateBindings
+    let strings: SecurityAndPrivacyScreenStrings
+    
     var canonicalAlias: String?
     var isKnockingEnabled: Bool
     var isSpaceSettingsEnabled: Bool
     var isSpace: Bool
-    var historySharingDetailsURL: URL
     
     var canEditAddress = false
     var canEditJoinRule = false
     var canEnableEncryption = false
     var canEditHistoryVisibility = false
+    
     /// The union of joined parent spaces and the joined spaces in the current access type
     var selectableJoinedSpaces: [SpaceRoomProxyProtocol] = []
     
@@ -163,13 +166,13 @@ struct SecurityAndPrivacyScreenViewState: BindableState {
         self.isKnockingEnabled = isKnockingEnabled
         self.isSpace = isSpace
         self.isSpaceSettingsEnabled = isSpaceSettingsEnabled
-        self.historySharingDetailsURL = historySharingDetailsURL
         
         let settings = SecurityAndPrivacySettings(accessType: accessType,
                                                   isEncryptionEnabled: isEncryptionEnabled,
                                                   historyVisibility: historyVisibility)
         currentSettings = settings
         bindings = SecurityAndPrivacyScreenViewStateBindings(desiredSettings: settings)
+        strings = SecurityAndPrivacyScreenStrings(historySharingDetailsURL: historySharingDetailsURL)
     }
 }
 
@@ -260,5 +263,28 @@ enum SecurityAndPrivacyHistoryVisibility: Int, Comparable {
     
     static func < (lhs: SecurityAndPrivacyHistoryVisibility, rhs: SecurityAndPrivacyHistoryVisibility) -> Bool {
         lhs.rawValue < rhs.rawValue
+    }
+}
+
+struct SecurityAndPrivacyScreenStrings {
+    let accessSectionFooterString: AttributedString
+    let historySectionFooterString: AttributedString
+    
+    init(historySharingDetailsURL: URL) {
+        let linkPlaceholder = "{link}"
+        
+        var accessFooterString = AttributedString(L10n.screenSecurityAndPrivacyRoomAccessFooter(linkPlaceholder))
+        var accessLinkString = AttributedString(L10n.screenSecurityAndPrivacyRoomAccessFooterManageSpacesAction)
+        accessLinkString.link = .init(stringLiteral: "action://manageSpace") // The link address doesn't matter
+        accessLinkString.bold()
+        accessFooterString.replace(linkPlaceholder, with: accessLinkString)
+        accessSectionFooterString = accessFooterString
+        
+        var historyFooterString = AttributedString(L10n.screenSecurityAndPrivacyRoomHistorySectionFooter(linkPlaceholder))
+        var historyLinkString = AttributedString(L10n.actionLearnMore)
+        historyLinkString.link = historySharingDetailsURL
+        historyLinkString.bold()
+        historyFooterString.replace(linkPlaceholder, with: historyLinkString)
+        historySectionFooterString = historyFooterString
     }
 }
--- a/ElementX/Sources/Screens/SecurityAndPrivacyScreen/View/SecurityAndPrivacyScreen.swift
+++ b/ElementX/Sources/Screens/SecurityAndPrivacyScreen/View/SecurityAndPrivacyScreen.swift
@@ -81,26 +81,14 @@ struct SecurityAndPrivacyScreen: View {
                 .compoundListSectionHeader()
         } footer: {
             if context.viewState.shouldShowAccessSectionFooter {
-                accessSectionFooter
+                Text(context.viewState.strings.accessSectionFooterString)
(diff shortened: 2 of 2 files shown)


ElementX/Sources/Screens/SecurityAndPrivacyScreen/View/SecurityAndPrivacyScreen.swift
```diff
@@ -133,23 +133,38 @@ struct SecurityAndPrivacyScreen: View {
...
+            historySectionFooter
         }
     }
     
+    private var historySectionFooter: some View {
+        let linkPlaceholder = "{link}"
+        var footerText = AttributedString(L10n.screenSecurityAndPrivacyRoomHistorySectionFooter(linkPlaceholder))
+        
+        var linkString = AttributedString(L10n.actionLearnMore)
+        linkString.link = "https://element.io/en/help#e2ee-history-sharing"
```
pixlwave: Not a review, but please add this to AppSettings and allow it to be overridden like the rest of the support URLs.

stefanceriu: Faaaair enough 😮‍💨

stefanceriu: 11c8edf5e50d26bb58b5d9a01467e582c2eb2b82

ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenModels.swift
```diff
@@ -58,13 +48,13 @@ struct SecurityAndPrivacyScreenViewState: BindableState {
...
+        var options = [SecurityAndPrivacyHistoryVisibility.shared]
         if !bindings.desiredSettings.isEncryptionEnabled, bindings.desiredSettings.accessType == .anyone {
-            options.append(.anyone)
+            options.append(.worldReadable)
         } else {
-            options.append(.sinceInvite)
+            options.append(.invited)
         }
-        return options
+        return options.sorted()
```
Velin92, line 60: Why sorted? This could make the order not reflect the one provided here, which may be different from designs

ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenModels.swift
```diff
@@ -254,17 +244,21 @@ enum SecurityAndPrivacyScreenViewAction {
...
+        case .invited, .shared:
+            return .shared
+        case .worldReadable:
+            return .invited
         }
     }
+    
+    static func < (lhs: SecurityAndPrivacyHistoryVisibility, rhs: SecurityAndPrivacyHistoryVisibility) -> Bool {
+        lhs.rawValue < rhs.rawValue
+    }
```
Velin92, line 266: Okay I see because are making it Int, however are you sure Comparable is required? I think by having it being an Int already should be sufficient

ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenModels.swift
stefanceriu: They're supposed to be sorted `in the order from the most to least restrictive` .. okay, you already commented 😁

ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenModels.swift
stefanceriu: Honestly I expected the same exact thing but it didn't work for some reason. Couldn't be bothered to get the bottom for it.

ElementX/Sources/Screens/SecurityAndPrivacyScreen/View/SecurityAndPrivacyScreen.swift
```diff
@@ -80,17 +80,29 @@ struct SecurityAndPrivacyScreen: View {
...
+        linkString.bold()
+        footerString.replace(linkPlaceholder, with: linkString)
+        
+        return Text(footerString)
+            .compoundListSectionFooter()
+            .environment(\.openURL, OpenURLAction { _ in
+                context.send(viewAction: .manageSpaces)
+                return .handled
+            })
+    }
```
Velin92: I would suggest having this code in the viewState statically, otherwise every time there is a change the attributed string will be recomputed, which is not optimal.

ElementX/Sources/Screens/SecurityAndPrivacyScreen/View/SecurityAndPrivacyScreen.swift
```diff
@@ -133,23 +145,38 @@ struct SecurityAndPrivacyScreen: View {
...
+        var footerText = AttributedString(L10n.screenSecurityAndPrivacyRoomHistorySectionFooter(linkPlaceholder))
+        
+        var linkString = AttributedString(L10n.actionLearnMore)
+        linkString.link = context.viewState.historySharingDetailsURL
+        linkString.bold()
+        footerText.replace(linkPlaceholder, with: linkString)
+        
+        return Text(footerText)
+            .compoundListSectionFooter()
+    }
```
Velin92: Same here

Velin92 (changes requested):
generally LGTM, but I would avoid computing Attributed Strings in the view directly if possible

ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenModels.swift
stefanceriu: Tried it a brand new enum, it just doesn't work.

ElementX/Sources/Screens/SecurityAndPrivacyScreen/View/SecurityAndPrivacyScreen.swift
stefanceriu: Unfortunately we need to access the `context.viewState.historySharingDetailsURL`

ElementX/Sources/Screens/SecurityAndPrivacyScreen/View/SecurityAndPrivacyScreen.swift
stefanceriu: And `context.send(viewAction: .manageSpaces)` here

Velin92: yeah only the string generation needs to be in the viewState

ElementX/Sources/Screens/SecurityAndPrivacyScreen/View/SecurityAndPrivacyScreen.swift
Velin92: yeah we can store that in the viewState and generate the string there, I think we already do something similar in the app.

Velin92: https://github.com/element-hq/element-x-ios/blob/93dfc84a57c04e2866f02d7bb0c765ad9fb4de83/ElementX/Sources/Screens/RoomNotificationSettingsScreen/RoomNotificationSettingsScreenModels.swift#L91-L102

stefanceriu: I find that a bit strange to be honest, ideally visual components should live inside the view not in the view state. I wonder if this is even important, how often to these views change and how heavy computing a small strings is?

On a separate note, that `URL(string: "element://openGlobalSettings")` should probably come from the AppSettings, no?

pixlwave: > On a separate note, that URL(string: "element://openGlobalSettings") should probably come from the AppSettings, no?

Please let's not go crazy here, we handle it immediately in the view. If anything we should just rename it to `action://` like the other one.

Velin92: I mean this is just an example, I don't even know if "element://openGlobalSettings" is actually a valid link at all or just a placeholder to handle the openURL action in this case, there other ones however in the app.

Regarding being a visual element, i don't agree. `AttributedString` is just a structure describing a string with attributes, is not a SwiftUI Text component. We do in fact indeed store AttributedString in models, like the RoomSummary for example. It's not too different from storing a `String` in a model.

It's definitely not too expensive replacing and generating an AttributedString every time a view is recomputed, but any change in the state recomputes a view, so it's definitely a bit of a waste of resources, and having them stored in the model statically or at init time, is just good practice.

PR #4863, opened 2025-12-16
Various codecov and workflow action tweaks

stefanceriu:
Prompted by the deprecation of the codecov/test-results-action.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -82,6 +82,8 @@ jobs:
         uses: codecov/codecov-action@671740ac38dd9b0130fbe1cec585b89eea48d3de # v5.5.2
         with:
           report_type: coverage
+          files: fastlane/test_output/integration-cobertura.xml
+          disable_search: true
           fail_ci_if_error: true
           token: ${{ secrets.CODECOV_TOKEN }}
           flags: integrationtests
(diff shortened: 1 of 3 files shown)


stefanceriu:
> [codecov/project/project-unittests](https://github.com/element-hq/element-x-ios/pull/4863/checks?check_run_id=58170935838) — 70.54% (-1.44%) compared to 678649a

> [codecov/project](https://github.com/element-hq/element-x-ios/pull/4863/checks?check_run_id=58170934115)
codecov/project — 79.28% (-0.38%) compared to 678649a

🤔

.github/workflows/integration-tests.yml
```diff
@@ -71,25 +72,32 @@ jobs:
           if-no-files-found: ignore
 
       - name: Collect coverage
+        # Skip if cancelled and in forks
+        if: ${{ !cancelled() && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository) }}
```
pixlwave: FWIW `!cancelled()` will run on failure too – which is why it was on test results (as we want to know failures), but not on coverage.

stefanceriu: So should I add a `!failure()` or what?

pixlwave: `success() && …` probably 😄

stefanceriu:
> LGTM! Do we need to un-pin renovate, I can't remember 🤔

Might magically restart itself now that we updated the version.

pixlwave:
> > LGTM! Do we need to un-pin renovate, I can't remember 🤔
> 
> Might magically restart itself now that we updated the version.

Unlikely…

https://github.com/element-hq/element-x-ios/blob/189da204192fb10c73d4a1186d5b92f5323926c0/renovate.json#L29-L32

stefanceriu:
Ah, done in b668de8908ac3b932547406c07e5d04048881b4d

PR #4864, opened 2025-12-16
Have fastlane's xcbeautify use a special github actions formatter and reporter.

stefanceriu:
> --is-ci Print test result too under quiet/quieter flag.

Not particularly sure it's useful in this case but it can't hurt. We might want to play with quiet/quieter as well

> xcbeautify features an integrated GitHub Actions renderer that harnesses [workflow commands](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions) to highlight warnings, errors, and results directly within the GitHub user interface. To utilize this function, simply run xcbeautify and add the --renderer github-actions flag during execution.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -26,7 +26,8 @@ lane :unit_tests do |options|
     ensure_devices_found: true,
     result_bundle: true,
     number_of_retries: 3,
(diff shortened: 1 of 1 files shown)


pixlwave (approved):
🤞

PR #4865, opened 2025-12-16
General tidy-up related to QR codes.

pixlwave:
This PR should be reviewed commit-by-commit to make sense. It is more preparatory work for linking a new device with a QR code (the main one being a shared view for handling the error states). The commits should be self explanatory from the message.

It's hard to figure out the best way to split all this work up, so there's more to follow 😄

### DIFF
--- a/UnitTests/Sources/QRCodeLoginScreenViewModelTests.swift
+++ b/UnitTests/Sources/QRCodeLoginScreenViewModelTests.swift
@@ -53,7 +53,7 @@ final class QRCodeLoginScreenViewModelTests: XCTestCase {
         try await deferred.fulfill()
         XCTAssertTrue(appMediatorMock.requestAuthorizationIfNeededCalled)
         
-        context.send(viewAction: .openSettings)
+        context.send(viewAction: .errorAction(.openSettings))
         await Task.yield()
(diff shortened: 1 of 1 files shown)


ElementX/Sources/Screens/QRCodeLoginScreen/QRCodeLoginScreenModels.swift
```diff
@@ -101,6 +102,8 @@ enum QRCodeLoginState: Equatable {
                     L10n.screenChangeServerErrorUnauthorizedHomeserverTitle(scannedProvider)
                 case .deviceNotSignedIn:
                     L10n.screenQrCodeLoginDeviceNotSignedInScanStateSubtitle
+                case .deviceAlreadySignedIn:
+                    L10n.screenQrCodeLoginDeviceNotSignedInScanStateSubtitle // FIXME: Update string
```
stefanceriu, line 106: Still planning on updating these?

pixlwave: Yep but I'll do these after https://github.com/element-hq/element-x-ios/pull/4866 seeing as the header's icon needs to change too.

PR #4866, opened 2025-12-16
Move BigIcon into Compound and add a new TitleAndIcon component.

pixlwave:
So far I've only updated 2 of the `TitleAndIcon` instances for now as that default padding might not make this the amazing drop-in replacement it should be.

But it will help with the QR screens I'm iterating on at the very least and I can explore further there.


pixlwave:
> Approved, however could you check with Aaron if we need to also change the colors for the default and defaultSolid case? I have seen new designs where we want the foreground icon color to be iconPrimary instead of the current one.

Ah yeah good point, will do (but in a separate PR given the number of snapshots that will be involved 🙈)

PR #4874, opened 2025-12-17
Final tweaks to the existing QR code screens to match the designs.

pixlwave:
That's enough changes for now! Back to implementing the ability to show a code or scan the code on a logged out device.


ElementX/Sources/Other/SwiftUI/Layout/FullscreenDialog.swift
```diff
@@ -21,8 +21,10 @@ struct FullscreenDialog<Content: View, BottomContent: View>: View {
     
     /// Padding applied to the top of the content automatically. Use `UIConstants` for preset values.
     var topPadding: CGFloat = UIConstants.titleTopPaddingToNavigationBar
-    /// Padding applied to the content and buttons automatically
+    /// Padding applied to the content automatically
     var horizontalPadding: CGFloat = 16
```
pixlwave, line 25: It's possible this should now default to `24`, but that doesn't seem like a good use of my time right now.

Velin92 (approved):

PR #4878, opened 2025-12-17
Unban correctly depends on kick PL + removed Change Settings permission

### LINKED ISSUE #4876: [Task] Unban action available prematurely
1. Give Bob a permission to `ban` but not `kick` in the space.
2. Let Bob ban Alice from the space.
3. Let Bob unban Alice from the space.

At step 3, the Unban action is available to Bob, however, it fails because unbanning also assumes the `kick` permission.

### LINKED ISSUE #4877: [Task] iOS: Remove "Change settings" from Permissions screen
The _Change settings_ entry was inherited from Web and it corresponds to the `state_default` in the spec, but it was not realized that it can become very confusing. Depending if explicit PLs have been assigned to, for example, `m.room.join_rules` or `m.room.canonical_alias` or anything else that affects how the UI behaves, it can become very confusing, if we do not have explicit entries for those.

As it is unlikely that it is necessary to control such permissions on mobile app, it is better to remove the _Change settings_ entry. The proper fix would be to expose specific actions (like changing access to the room).

Velin92:
fixes #4876 
fixes #4877 

Seems that to unban you just need the kick permission, this PR takes care of it + removes a setting we are not sure about implementing on mobile as of right now.

### DIFF
--- a/ElementX/Sources/Screens/ManageRoomMemberSheet/View/ManageRoomMemberSheetView.swift
+++ b/ElementX/Sources/Screens/ManageRoomMemberSheet/View/ManageRoomMemberSheetView.swift
@@ -40,25 +40,33 @@ struct ManageRoomMemberSheetView: View {
             
             Section {
                 if context.viewState.permissions.canKick {
-                    ListRow(label: .default(title: L10n.screenBottomSheetManageRoomMemberRemove,
-                                            icon: \.close,
-                                            role: .destructive),
-                            kind: .button {
-                                context.send(viewAction: .kick)
-                            })
-                            .disabled(context.viewState.isKickDisabled)
+                    // Unbanning requires only the kick permission according to the specs.
+                    if context.viewState.isMemberBanned {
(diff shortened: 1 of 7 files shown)


ElementX/Sources/Screens/ManageRoomMemberSheet/View/ManageRoomMemberSheetView.swift
```diff
@@ -40,25 +40,32 @@ struct ManageRoomMemberSheetView: View {
...
                 if context.viewState.permissions.canKick {
-                    ListRow(label: .default(title: L10n.screenBottomSheetManageRoomMemberRemove,
-                                            icon: \.close,
-                                            role: .destructive),
-                            kind: .button {
-                                context.send(viewAction: .kick)
-                            })
-                            .disabled(context.viewState.isKickDisabled)
+                    if context.viewState.isMemberBanned {
+                        ListRow(label: .default(title: L10n.screenBottomSheetManageRoomMemberUnban,
```
pixlwave, line 45: Let's add a comment here so that someone doesn't come across this in a couple of years and think it's a bug…

PR #4888, opened 2025-12-18
Run client store optimizations when upgrating versions

stefanceriu:
In order to avoid eventual slowdowns of the underlying sqlite databases we found that we need to run various optimizations tasks on them. We decided to do so every time the app version is updated.

At the same time the SDK now exposes the store file sizes so we're now rendering those within the developer options menu.

Store sizes


manuroe:
Fixes https://github.com/element-hq/element-x-ios/issues/4841.

PR #4889, opened 2025-12-18
Change permissions screen is now responsive to the current user's power level

### LINKED ISSUE #4875: [Task] iOS: Space's permissions screen allows operations/levels that are not possible
Set Bob's PL to 50.
Set `m.room.power_levels` (in room power-levels) to 50.
Set `invite` (in room power-levels) to 75.

In such a scenario Bob can access Roles & permissions > Permissions and choose any level (Admin, Moderator, member) for the "Invite" operation.

Problem 1: In reality, Bob can not change the level for the "Invite" operation. This is an intentional limitation to not allow a user to assign themselves permissions that they do not yet have (regardless of having the general permission to change power-levels).

Problem 2: If Bob's PL gets upgraded to 75, they can change the level for the "Invite" operation, but not beyond PL75 (e.g. in the current case they can't change it to Admin). This is also an intentional limitation to not allow the user to remove a permission from themselves.

This is probably not a critical issue since in most cases the `m.room.power_level` is higher than most of the other permissions such as `invite`. However, the fix also is not too complicated:

1. If the PL of the operation is higher than the current user PL, then disable changing the level for it.
2. Only show PLs in the dropdown that are lower than the PL of the current user.

Velin92:
These changes are made because currently such screen assumed that the user performing permission changes was always an Admin.
However a room may utilise some customised settings for roles and permissions, where a non admin is allowed to change roles and permissions, such user might try to change permissions in a way that the server and the specs don't allow, causing a 400 response.

Now the UI is responsive to the current user's power level and follows the specs, so:

- If the current power level is below the current permission power level, its editing will be disabled
- If the user can edit roles and permission, but they are not an admin, they will only be allowed to select among the roles that are below or equal to their own power level. (e.g. a Moderator can only select between Moderator and Member)

Also added a UnitTest and a PreviewTests for these cases.

fixes #4875

### DIFF
--- a/ElementX/Sources/Services/Room/RoomPermissions.swift
+++ b/ElementX/Sources/Services/Room/RoomPermissions.swift
@@ -10,24 +10,29 @@ import Foundation
 import MatrixRustSDK
 
 struct RoomPermissionsSetting: Identifiable {
-    var id: KeyPath<RoomPermissions, RoomRole> { keyPath }
+    static let allValues: [(title: String, tag: RoomRole)] = [(title: L10n.screenRoomChangePermissionsAdministrators, tag: .administrator),
+                                                              (title: L10n.screenRoomChangePermissionsModerators, tag: .moderator),
+                                                              (title: L10n.screenRoomChangePermissionsEveryone, tag: .user)]
+    var id: KeyPath<RoomPermissions, Int64> { keyPath }
     
     /// The title of this setting.
     let title: String
     
     /// The selected role of this setting.
-    var value: RoomRole
-    /// All of the available roles that this setting can be configured with.
-    var allValues: [(title: String, tag: RoomRole)] {
-        [
-            (title: L10n.screenRoomChangePermissionsAdministrators, tag: .administrator),
-            (title: L10n.screenRoomChangePermissionsModerators, tag: .moderator),
-            (title: L10n.screenRoomChangePermissionsEveryone, tag: .user)
-        ]
-    }
+    var value: Int64
     
+    let ownPowerLevel: RoomPowerLevel
+    
+    var roleValue: RoomRole {
+        get {
+            RoomRole(powerLevelValue: value)
+        } set {
+            value = newValue.powerLevelValue
+        }
+    }
+        
     /// The `RoomPermissions` property that this setting is for.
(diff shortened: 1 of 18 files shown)


ElementX/Sources/Services/Room/RoomPermissions.swift
```diff
@@ -45,46 +50,75 @@ struct RoomPermissionsSetting: Identifiable {
...
+    var isDisabled: Bool {
+        switch ownPowerLevel {
+        case .value(let ownValue):
+            ownValue < value
+        case .infinite:
+            false
+        }
+    }
+    
+    /// All of the available roles that this setting can be configured with.
```
pixlwave, line 64: Minor but maybe we should have this here and the original back on the `static let`?
```suggestion
    /// All of the available roles that this setting is allowed to be configured with.
```

ElementX/Sources/Services/Room/RoomPermissions.swift
```diff
@@ -45,46 +50,75 @@ struct RoomPermissionsSetting: Identifiable {
         default: fatalError("Unexpected key path: \(keyPath)")
         }
     }
+    
+    /// Can the setting be edited
+    var isDisabled: Bool {
```
pixlwave, line 55: And maybe for clarity this should be called `canEdit`/`isEditable`/whatever to indicate that the user can edit it rather than the UI state?

ElementX/Sources/Services/Room/RoomRole.swift
```diff
@@ -110,21 +110,25 @@ extension RoomRole {
     /// To be used when setting the power level of a user to get the suggested equivalent power level value for that specific role
     /// NOTE: Do not use for comparison, use the true power level instead.
```
pixlwave, line 111: Should this comment live on `powerLevelValue` or the "new" `powerLevel`? Or both?

pixlwave (approved):
LGTM 👍

ElementX/Sources/Services/Room/RoomRole.swift
Velin92: yeah still here, because powerLevelValue can't account for infinite PL cases. The other one instead is good for comparisons since it can account for it.

PR #4891, opened 2025-12-19
Add support for linking new devices in the QRCodeLoginScreen.

pixlwave:
Notes:
- Based on https://github.com/hughns/element-x-ios/tree/msc4108-v2025 (although some of this is already in the app now from previous PRs).
- There's still SDK updates required for this such as https://github.com/matrix-org/matrix-rust-sdk/pull/5957 (but everything is behind a feature flag so it is safe to merge).
- I've only tested the "Link Mobile Device" flow. I believe linking against EW needs a specific build.


Velin92 (approved):
LGTM but I would have put the QR Code scanning refactor into a separate PR

pixlwave:
Ah yeah sorry, I could've split this one up better in the end.

PR #4892, opened 2025-12-19
Do not show history visible banner when the user cannot send messages.

kaylendog:
### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
+++ b/ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
@@ -84,7 +84,13 @@ struct RoomScreenViewState: BindableState {
     var historyVisibleDetails: RoomScreenFooterViewDetails?
     
     var footerDetails: RoomScreenFooterViewDetails? {
-        identityViolationDetails ?? historyVisibleDetails
+        if let identityViolationDetails {
+            return identityViolationDetails
+        }
+        guard canSendMessage else {
+            return nil
+        }
+        return historyVisibleDetails
     }
     
     var bindings = RoomScreenViewStateBindings()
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
```diff
@@ -84,7 +84,13 @@ struct RoomScreenViewState: BindableState {
...
     
     var footerDetails: RoomScreenFooterViewDetails? {
-        identityViolationDetails ?? historyVisibleDetails
+        if let identityViolationDetails {
+            return identityViolationDetails
+        }
+        guard canSendMessage else {
+            return nil
+        }
+        return historyVisibleDetails
```
kaylendog, line 93: I think this is the most sensible way to do this?

pixlwave: Yep, looks sane to me 👍

UnitTests/Sources/RoomScreenViewModelTests.swift
```diff
@@ -553,6 +553,38 @@ class RoomScreenViewModelTests: XCTestCase {
...
         try await deferred.fulfill()
     }
+    
+    func testHistoryVisibleBannerDoesNotAppearIfCannotSendMessages() async throws {
+        ServiceLocator.shared.settings.enableKeyShareOnInvite = true
+        ServiceLocator.shared.settings.acknowledgedHistoryVisibleRooms = Set()
+        
+        let powerlevels = RoomPowerLevelsProxyMockConfiguration(
+            canUserSendMessage: false
+        )
```
pixlwave: Tiny nit, we don't format multi-line parameters this way, but our formatter only actually reformats them if there are multiple parameters…

```suggestion
        let powerlevels = RoomPowerLevelsProxyMockConfiguration(canUserSendMessage: false)
```

PR #4846, opened 2025-12-11
Add "Translate" to TimelineItemMenuActions

bo2themax:
https://github.com/user-attachments/assets/cfde29e1-2233-4efb-8fa8-9515d0b93578

> [!NOTE]
> Translated by AI

### Pull Request Checklist

- [x] I read the [contributing guide](https://github.com/element-hq/element-ios/blob/develop/CONTRIBUTING.md).
- [ ] Pull request contains a [changelog label](https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md#changelog).
  > Don't see any `changlog.d` folder in this repo

**UI changes have been tested with:**
- [x] iPhone and iPad simulators in portrait and landscape orientations.
- [x] Dark mode enabled and disabled.
- [x] Various sizes of dynamic type.
- [ ] Voiceover enabled.
  > Don't have a paid developer account to run on a physical device, but this is basically the same with the "Copy" action.

PR #4912, opened 2026-01-05
Tweaks to the Translate action.

pixlwave:
Just a small tweak (along with a rebase) on top of #4846 (which is included in the first commit).

Using a PR to make sure CI is happy.


ElementX/Sources/Screens/Timeline/View/ItemMenu/TimelineItemMenuAction.swift
```diff
@@ -145,6 +146,11 @@ enum TimelineItemMenuAction: Identifiable, Hashable {
         switch self {
         case .copy:
             Label(L10n.actionCopyText, icon: \.copy)
+        case .translate:
+            Label { Text(L10n.actionTranslate) } icon: {
+                Image(systemSymbol: .translate)
+                    .scaledFrame(size: 24)
```
bo2themax: There were some weird crashes for me on the simulator without setting resizable here, and aspectRatio was to make sure the symbol doesn't stretch weirdly

bo2themax: If it works for you and on physical devices, then I guess it doesn't matter

PR #4914, opened 2026-01-05
Update the SDK

pixlwave:

- Update the SDK and handle breaking changes.
- Refactor the space service's `joinedSpaces` to `topLevelSpaces` to match the SDK more closely (and remove any ambiguity about what the property represents).
- Move the generated SDK mocks into their own target – there is a naming conflict between [`TimelineEventContent`](https://github.com/element-hq/element-x-ios/blob/5bf96321120cb45093fe4f14e3e9e81b849a8c52/ElementX/Sources/Services/Timeline/TimelineItemContent/TimelineEventContent.swift#L11-L15) and [`MatrixRustSDK.TimelineEventContent`](https://github.com/matrix-org/matrix-rust-sdk/blob/e3867aa7df16ac8dfd8b1936a4a635ce13c8de7d/bindings/matrix-sdk-ffi/src/event.rs#L285-L288) in the generated mocks which Sourcery cannot handle in the stencil (as it lacks module information). I tried a few things in the stencil, but sadly this was the only solution I could come up with to stop Swift from incorrectly using the Element X enum in the mocks.


manuroe (approved):
Thanks for the update.
LGTM but, to be honest, I can only say that the changes look consistent 😇.

PR #4926, opened 2026-01-07
Check permissions before indirectly updating desired settings in sec & privacy

### LINKED ISSUE #4873: [Task] iOS: Space's security & privacy section is not shown correctly
It is assumed that the user needs to have `m.room.join_rules` and `m.room.canonical_alias` permission to show them the Security & privacy section.

However, there were cases found:
* when user was not shown this section, even if they had those permissions (depending on what the other permissions were, either the entire Settings was unavailable or just the Security & privacy section was unavailable).
* when the user was shown this section, even if they did not have those permissions.

The details of which permissions the user had and what the UI showed, can be find in this [table](https://docs.google.com/spreadsheets/d/1nLZkudmrmeLlbsnHpWCaHWZ1q5OmJIQahOx7bP4ckY8/edit?gid=1772120308#gid=1772120308&range=B2).

Note 1: Above, the "user needs to have m.room.join_rules" (or any other permission) means the user *effectively* having this permission - e.g. regardless if a PL is assigned to `m.room.join_rules` directly or it is inherited via the `state_default`.

Note 2: During the testing it was noted even though user was PL25 and both `m.room.join_rules` and `m.room.canonical_alias` were also PL25 but `state_default` was PL50, then the user could change room access, however the room address was not shown. Therefore it should be checked if that is a bug in the code, or there is a 3rd permission that is needed to make this section work.

Velin92:
We were not checking permissions and/or if a room is also a space before calling sec and privacy APIs. This resulted in possible forbidden calls when the user changed some of the settings (which may indirectly influence other settings).

fixes #4873

### DIFF
--- a/ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenViewModel.swift
+++ b/ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenViewModel.swift
@@ -87,6 +87,7 @@ class SecurityAndPrivacyScreenViewModel: SecurityAndPrivacyScreenViewModelType, 
     
     private func setupSubscriptions() {
         context.$viewState
+            .drop { $0.isSpace || !$0.canEditHistoryVisibility }
             .map(\.availableVisibilityOptions)
             .removeDuplicates()
             // To allow the view to update properly
@@ -97,6 +98,20 @@ class SecurityAndPrivacyScreenViewModel: SecurityAndPrivacyScreenViewModelType, 
                 let desiredHistoryVisibility = state.bindings.desiredSettings.historyVisibility
                 if !availableVisibilityOptions.contains(desiredHistoryVisibility) {
                     state.bindings.desiredSettings.historyVisibility = desiredHistoryVisibility.fallbackOption
+                }
+            }
+            .store(in: &cancellables)
+        
+        context.$viewState
+            .drop { !$0.canEditAddress }
+            .map(\.bindings.desiredSettings.accessType)
+            .removeDuplicates()
+            // To allow the view to update properly
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] accessType in
+                guard let self else { return }
+                if state.bindings.desiredSettings.isVisibileInRoomDirectory == true, !accessType.isAddressRequired {
+                    state.bindings.desiredSettings.isVisibileInRoomDirectory = false
                 }
             }
             .store(in: &cancellables)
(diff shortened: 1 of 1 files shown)


ElementX/Sources/Screens/SecurityAndPrivacyScreen/SecurityAndPrivacyScreenViewModel.swift
```diff
@@ -179,27 +179,29 @@ class SecurityAndPrivacyScreenViewModel: SecurityAndPrivacyScreenViewModelType,
...
         
-        if state.currentSettings.isEncryptionEnabled != state.bindings.desiredSettings.isEncryptionEnabled {
-            switch await roomProxy.enableEncryption() {
-            case .success:
-                state.currentSettings.isEncryptionEnabled = state.bindings.desiredSettings.isEncryptionEnabled
-            case .failure:
-                userIndicatorController.submitIndicator(.init(title: L10n.errorUnknown))
-                hasFailures = true
+        if !state.isSpace {
+            if state.canEnableEncryption, state.currentSettings.isEncryptionEnabled != state.bindings.desiredSettings.isEncryptionEnabled {
```
pixlwave: Purely out of curiosity, would it make any sense to include `!isSpace` when computing `canEnableEncryption` and `canEditHistoryVisibility`?

pixlwave: Also, I find it slightly surprising that you need both `if canChangeValue, currentValue != desiredValue` here. How do the current and desired values get out of sync if you can't set them? Asking because this would impact the state of the save button iirc.

Velin92: the main issue is the history visibility which may change its desired value if the join rule is changed:

```
 context.$viewState
            .map(\.availableVisibilityOptions)
            .removeDuplicates()
            // To allow the view to update properly
            // When the available options changes always default to `sinceSelection` if the currently selected option is not available
            .sink { [weak self] availableVisibilityOptions in
                let desiredHistoryVisibility = state.bindings.desiredSettings.historyVisibility
                if !availableVisibilityOptions.contains(desiredHistoryVisibility) {
                    state.bindings.desiredSettings.historyVisibility = desiredHistoryVisibility.fallbackOption
                }
            }
            .store(in: &cancellables)
 ```
 
 This is actually the only indirect change that is affected by another change, however if you don't have such permission you still should not call the API (even if in theory the current value of history visibility setting doesn't really make sense for that join rule conceptually Element design wise, it is allowed by the specs and it is possible to be changed by itself through APIs and dev tools)

Velin92: About merging the permission check with the `isSpace` check is totally doable too, I kept them separate because we have two different subscriptions one on the info and one on the power levels (since we need to manually remove duplicates for the latter), so it's bit of an annoyance. Tell me WDYT

pixlwave: > the main issue is the history visibility which may change its desired value if the join rule is changed:

Ah makes sense. Maybe add a comment in the VM to note that we also check the `can…` because these indirect changes are possible. That would be enough to explain it.

> I kept them separate because we have two different subscriptions

Ah ok, I think I remember that now. Then yeah don't worry about it :)

PR #4935, opened 2026-01-08
Handle OIDC cancellation and workaround missing progress when linking a new device.

pixlwave:
Can be reviewed commit-by-commit. Depends on #4934.

### DIFF
--- a/ElementX/Sources/Screens/QRCodeLoginScreen/QRCodeLoginScreenCoordinator.swift
+++ b/ElementX/Sources/Screens/QRCodeLoginScreen/QRCodeLoginScreenCoordinator.swift
@@ -16,12 +16,22 @@ struct QRCodeLoginScreenCoordinatorParameters {
     let appMediator: AppMediatorProtocol
 }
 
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Screens/Authentication/OIDCAuthenticationPresenter.swift
```diff
@@ -121,3 +117,20 @@ extension ASWebAuthenticationSession.Callback {
...
+        if nsError.domain == ASWebAuthenticationSessionErrorDomain,
+           nsError.code == ASWebAuthenticationSessionError.canceledLogin.rawValue,
+           // If there's a failure reason then the cancellation wasn't made by the user.
+           nsError.localizedFailureReason == nil {
+            return true
+        }
+        
+        return false
+    }
+}
```
Velin92, line 136: would make this private/fileprivate

pixlwave: I would, but I extracted it to share the logic with `OIDCAccountSettingsPresenter.swift` 🙃

PR #4936, opened 2026-01-08
Empty spaces list screen

Velin92:
- Added an FF to create a space
- Added an empty state to the spaces list screen

Awaiting to know:
- Spacing for the empty state title
- What does the learn more button do


ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -375,6 +378,14 @@ final class AppSettings {
...
     var sharePresence
     
+    // MARK: - Spaces
+    
+    @UserPreference(key: UserDefaultsKeys.spaceSettingsEnabled, defaultValue: false, storageType: .userDefaults(store))
+    var spaceSettingsEnabled
+    
+    @UserPreference(key: UserDefaultsKeys.createSpaceEnabled, defaultValue: false, storageType: .userDefaults(store))
+    var createSpaceEnabled
+
```
pixlwave: Not that it matters, but shouldn't these still be in the Feature Flags section given that they're feature flags?

ElementX/Sources/Screens/Spaces/SpaceListScreen/View/SpaceListScreen.swift
```diff
@@ -95,27 +118,50 @@ struct SpaceListScreen: View {
...
+        
+        if context.viewState.isCreateSpaceEnabled {
+            ToolbarItem(placement: .navigationBarTrailing) {
+                Button {
+                    context.send(viewAction: .createSpace)
+                } label: {
+                    CompoundIcon(\.plus)
+                        .accessibilityHidden(true)
+                }
+                .accessibilityLabel(L10n.actionCreateSpace)
```
pixlwave, line 129: Given this one isn't using `.backportSharedBackgroundVisibility(.hidden)`, I think we should also make the size of the avatar match the Liquid Glass designs and be 40x40 when running iOS 26:
https://www.figma.com/design/jvZ7TIyd7V9Vn8sIDB9vMd/Compound-iOS-Templates?node-id=4493-110925&m=dev

The catch being that if we do that, we should also do it on the Chats tab, which in turn means we need to update the Start Chat button on the home screen to match the design here:
https://www.figma.com/design/jvZ7TIyd7V9Vn8sIDB9vMd/Compound-iOS-Templates?node-id=4493-83646&m=dev

Totally doesn't need to happen in this PR though.

compound-ios/Sources/Compound/Components/TitleAndIcon.swift
pixlwave, line 1: Nice, thanks 😄

pixlwave (approved):
LGTM! 🌌

PR #4937, opened 2026-01-08
Display a badge for messages decrypted using shared keys.

kaylendog:
### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/UnitTests/Sources/TimelineViewModelTests.swift
+++ b/UnitTests/Sources/TimelineViewModelTests.swift
@@ -527,6 +527,36 @@ class TimelineViewModelTests: XCTestCase {
         try await deferred.fulfill()
     }
     
+    // MARK: - Tap Actions
+    
+    func testTapSendInfoEncryptionAuthentictyDisplaysAlert() {
+        // Given a room with an event whose authenticity could not be verified
+        let items = [TextRoomTimelineItem(eventID: "t1", encryptionAuthenticity: .verificationViolation(color: .red))]
+        let timelineController = MockTimelineController()
+        timelineController.timelineItems = items
+        let viewModel = makeViewModel(timelineController: timelineController)
+        
+        XCTAssertNil(viewModel.state.bindings.alertInfo)
+        
+        viewModel.process(viewAction: .itemSendInfoTapped(itemID: items[0].id))
+        
+        XCTAssertEqual(viewModel.state.bindings.alertInfo?.title, "Encrypted by a previously-verified user.")
+    }
+    
+    func testTapSendInfoEncryptionForwarderDisplaysAlert() {
+        // Given a room with an event whose key was forwarded
+        let items = [TextRoomTimelineItem(eventID: "t1", keyForwarder: .test)]
+        let timelineController = MockTimelineController()
+        timelineController.timelineItems = items
+        let viewModel = makeViewModel(timelineController: timelineController)
+        
+        XCTAssertNil(viewModel.state.bindings.alertInfo)
+        
+        viewModel.process(viewAction: .itemSendInfoTapped(itemID: items[0].id))
+        
+        XCTAssertEqual(viewModel.state.bindings.alertInfo?.title, "alice (@alice:matrix.org) shared this message since you were not in the room when it was sent.")
+    }
+    
     // MARK: - Helpers
     
     private func makeViewModel(roomProxy: JoinedRoomProxyProtocol? = nil,
@@ -579,11 +609,43 @@ private extension TextRoomTimelineItem {
     }
 }
 
+private extension TextRoomTimelineItem {
+    init(eventID: String, keyForwarder: TimelineItemKeyForwarder) {
+        self.init(id: .event(uniqueID: .init(UUID().uuidString), eventOrTransactionID: .eventID(eventID)),
+                  timestamp: .mock,
+                  isOutgoing: false,
+                  isEditable: false,
+                  canBeRepliedTo: true,
+                  sender: .init(id: ""),
+                  content: .init(body: "Hello, World!"),
+                  properties: RoomTimelineItemProperties(encryptionForwarder: keyForwarder))
+    }
+}
+
+private extension TextRoomTimelineItem {
+    init(eventID: String, encryptionAuthenticity: EncryptionAuthenticity) {
+        self.init(id: .event(uniqueID: .init(UUID().uuidString), eventOrTransactionID: .eventID(eventID)),
+                  timestamp: .mock,
+                  isOutgoing: false,
+                  isEditable: false,
+                  canBeRepliedTo: true,
+                  sender: .init(id: ""),
+                  content: .init(body: "Hello, World!"),
+                  properties: RoomTimelineItemProperties(encryptionAuthenticity: encryptionAuthenticity))
+    }
+}
+
 private extension TimelineItemSender {
     init(with proxy: RoomMemberProxyMock) {
         self.init(id: proxy.userID,
                   displayName: proxy.displayName ?? "",
                   isDisplayNameAmbiguous: false,
                   avatarURL: proxy.avatarURL)
+    }
+}
+
+private extension TimelineItemKeyForwarder {
+    static var test: TimelineItemKeyForwarder {
+        TimelineItemKeyForwarder(id: "@alice:matrix.org", displayName: "alice")
     }
 }


PreviewTests/Sources/__Snapshots__/PreviewTests/timelineItemBubbledStylerView.Encryption-Forwarder-Info-iPad-en-GB.png
kaylendog, line 1: I'm a bit perplexed as to why the timestamp is merging with the text here. I can't find any specific padding differences between my implementation and the others.

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
```diff
@@ -1003,6 +1006,11 @@ class TimelineViewModel: TimelineViewModelType, TimelineViewModelProtocol {
             state.bindings.alertInfo = .init(id: type,
                                              title: message,
                                              primaryButton: .init(title: L10n.actionOk, action: nil))
+        case .encryptionForwarder(let message):
+            state.bindings.alertInfo = .init(id: type,
+                                             title: message,
+                                             primaryButton: .init(title: L10n.actionOk, action: nil),
+                                             secondaryButton: .init(title: L10n.actionLearnMore) { UIApplication.shared.open(self.appSettings.historySharingDetailsURL) })
```
Velin92: Do not use the UIApplication directly, but use indeed our `AppMediator`

ElementX/Sources/Screens/Timeline/View/Style/TimelineItemSendInfoLabel.swift
```diff
@@ -68,6 +70,7 @@ private struct TimelineItemSendInfoLabel: View {
         switch sendInfo.status {
         case .sendingFailed: L10n.commonSendingFailed
         case .encryptionAuthenticity(let authenticity): authenticity.message
+        case .encryptionForwarder(let forwarder): UntranslatedL10n.encryptionForwarderDialogContent(forwarder.displayName ?? "Unknown", forwarder.id)
```
Velin92: "Unknown" should also be using localisation, i guess we could add a `commonUnknown` untranslated key for now

ElementX/Sources/Screens/Timeline/View/Style/TimelineItemSendInfoLabel.swift
```diff
@@ -111,7 +114,7 @@ private struct TimelineItemSendInfoLabel: View {
 
 /// All the data needed to render a timeline item's send info label.
 private struct TimelineItemSendInfo {
-    enum Status { case sendingFailed, encryptionAuthenticity(EncryptionAuthenticity) }
+    enum Status { case sendingFailed, encryptionAuthenticity(EncryptionAuthenticity), encryptionForwarder(TimelineItemForwarder) }
```
Velin92: not sure why this one was on a single line, could we make it multiline instead, one for each case?

ElementX/Sources/Screens/Timeline/View/Style/TimelineItemSendInfoLabel.swift
```diff
@@ -131,21 +134,25 @@ private struct TimelineItemSendInfo {
...
-    init(timelineItem: EventBasedTimelineItemProtocol, adjustedDeliveryStatus: TimelineItemDeliveryStatus?) {
+    init(timelineItem: EventBasedTimelineItemProtocol, adjustedDeliveryStatus: TimelineItemDeliveryStatus?, enableKeyShareOnInvite: Bool) {
         itemID = timelineItem.id
         localizedString = timelineItem.localizedSendInfo
         
         status = if case .sendingFailed = adjustedDeliveryStatus {
             .sendingFailed
         } else if let authenticity = timelineItem.properties.encryptionAuthenticity {
             .encryptionAuthenticity(authenticity)
+        } else if let forwarder = timelineItem.properties.encryptionForwarder, enableKeyShareOnInvite {
```
Velin92: `if enableKeyShareOnInvite, let forwarder = ...` Better to short circuit the implicit AND on the boolean first, to avoid an unnecessary unwrap

ElementX/Sources/Services/Timeline/TimelineItemContent/RoomTimelineItemProperties.swift
```diff
@@ -26,4 +26,6 @@ struct RoomTimelineItemProperties: Hashable {
     var orderedReadReceipts: [ReadReceipt] = []
     /// Authenticity warnings for item's sent in encrypted rooms.
     var encryptionAuthenticity: EncryptionAuthenticity?
+    /// If present, information about the forwarder of the keys used to decrypt this message.
```
Velin92: `/// Information about the forwarder of the keys used to decrypt this message.`

no need to specify if present, the optional type gives it away

ElementX/Sources/Services/Timeline/TimelineItemContent/TimelineItemForwarder.swift
```diff
@@ -0,0 +1,36 @@
...
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+import MatrixRustSDK
+import SwiftUI
+
+struct TimelineItemForwarder: Identifiable, Hashable {
+    static let test = TimelineItemForwarder(id: "@alice:matrix.org", displayName: "alice")
```
Velin92: instead of adding the test here we could make it as a computed static var, in an extension of this struct where it's being used.

ElementX/Sources/Screens/Timeline/View/Style/TimelineItemSendInfoLabel.swift
Velin92: Actually I also see that the forwarder itself contains a computed `message` property that does this, so this feels redundant, we could just do `forwarder.message`

ElementX/Sources/Services/Timeline/TimelineItemContent/TimelineItemForwarder.swift
```diff
@@ -0,0 +1,36 @@
...
+            self.init(id: forwarderID,
+                      displayName: displayName)
+        default:
+            self.init(id: forwarderID,
+                      displayName: nil)
+        }
+    }
+    
+    var message: String {
+        UntranslatedL10n.encryptionForwarderDialogContent(displayName ?? "Unknown", id)
```
Velin92: Unknown should be localised (as explained above we can just use a simple `commonUnknown`)

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
```diff
@@ -107,6 +107,7 @@ class TimelineViewModel: TimelineViewModelType, TimelineViewModelProtocol {
                                                        emojiProvider: emojiProvider,
                                                        linkMetadataProvider: hideTimelineMedia ? nil : linkMetadataProvider,
                                                        mapTilerConfiguration: appSettings.mapTilerConfiguration,
+                                                       enableKeyShareOnInvite: appSettings.enableKeyShareOnInvite,
```
Velin92, line 110: I also realised, we do this on instantiating the timeline, which is fine, but we should also sink on the appSettings publisher for this value (since on iPad you can have the settings shown while a timeline is open).
This will allow to dynamically turn on and off this value by listening to changes in the app settings (check some examples with `appSettings$...weakAssign(to:, on:)`

ElementX/Sources/Services/Timeline/TimelineItems/EventBasedTimelineItemProtocol.swift
```diff
@@ -44,7 +44,7 @@ extension EventBasedTimelineItemProtocol {
     }
     
     var hasStatusIcon: Bool {
-        hasFailedToSend || properties.encryptionAuthenticity != nil
+        hasFailedToSend || properties.encryptionAuthenticity != nil || properties.encryptionForwarder != nil
```
Velin92, line 47: also... I am realising... should we display this information also when the `enableKeyShareOnInvite` is off? I imagine that we sould also hide this badge icon when the encryption sender info are present, BUT the FF is off, but correct me if I am wrong, I don't know a lot about this project

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
kaylendog: As far as I know, changing the setting isn't reflected in the underlying Rust client until a restart anyway, so I'm not sure if this is worth doing? It might cause some strange behaviour where the UI renders as if the feature is enabled when the Rust SDK doesn't think it is.

ElementX/Sources/Services/Timeline/TimelineItems/EventBasedTimelineItemProtocol.swift
kaylendog: CC @mxandreas

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
Velin92: Where is the underlying rust client getting set with the FF then? Maybe I have missed it in the review, in fact I thought this value is just always available, we just wanted to decide wether to display it or not. (Maybe was done in a previous PR?)

kaylendog: I believe here? https://github.com/element-hq/element-x-ios/blob/8f9df47a2cd416702b046e4cd77f97225d12b482/ElementX/Sources/Other/Extensions/ClientBuilder.swift#L44

Which is indeed outside of this PR.

ElementX/Sources/Services/Timeline/TimelineItemContent/TimelineItemForwarder.swift
kaylendog: I'm a bit concerned about interpolating localised strings, I'll ask Andreas what the best way forward here is.

kaylendog:
> Generally looks good, some things to sort out, also `forwarder` is not used anywhere in the app as a name convention, we usually use the `sender` noun. Could we use it here too?

I want to disambiguate between the actual sender of the message and the person who forwarded us the key with which we decrypted the message, I could perhaps use `KeyForwarder` to clarify that? Maybe `KeySender`.

ElementX/Sources/Services/Timeline/TimelineItemContent/TimelineItemForwarder.swift
kaylendog: I've gone ahead and added two translation keys for this, which isn't ideal since they're near identical. Perhaps you can think of a better way to do this? The crux of the issue is that `ProfileDetails` is not always available.

ElementX/Sources/Services/Timeline/TimelineItems/EventBasedTimelineItemProtocol.swift
kaylendog: The [Element web implementation](https://github.com/element-hq/element-web/blob/bf1c8d0d45c13e54c3b8062f9373891ae0651536/src/components/views/rooms/EventTile.tsx#L739-L745) displays the badge irrespective of whether the FF is set, perhaps we should reflect that here.

ElementX/Sources/Services/Timeline/TimelineItemContent/TimelineItemForwarder.swift
Velin92: should be okay, we sometimes interpolate localised strings for example in the case in which a piece of the string requires some additional formatting.
The other and better solution, is to have another localised string that handles the displayName == nil case, that puts Unknown in place of the displayName directly in the string itself.

ElementX/Sources/Services/Timeline/TimelineItems/EventBasedTimelineItemProtocol.swift
Velin92: Considering that from what you told me the FF is set in the rust SDK, and would require a reboot on its change, then this is fine. But we should definitely tell users in the `DeveloperOptionsScreen` for this flag that changing this FF value requires an app reboot to work properly

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
Velin92: Okay I see then I guess is fine to not make this reactive, however we should add in the description of the FF in the developer options, that changing the value this feature requires a reboot to work properly.

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
```diff
@@ -1003,6 +1006,11 @@ class TimelineViewModel: TimelineViewModelType, TimelineViewModelProtocol {
             state.bindings.alertInfo = .init(id: type,
                                              title: message,
                                              primaryButton: .init(title: L10n.actionOk, action: nil))
+        case .encryptionForwarder(let message):
+            state.bindings.alertInfo = .init(id: type,
+                                             title: message,
+                                             primaryButton: .init(title: L10n.actionOk, action: nil),
+                                             secondaryButton: .init(title: L10n.actionLearnMore) { self.appMediator.open(self.appSettings.historySharingDetailsURL) })
```
Velin92:
```
[weak self] in
appMediator.open(appSettings.historySharingDetailURL)
```

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
kaylendog: I believe this is the case!
Simulator Screenshot - iPhone 17 - 2026-01-14 at 11 40 39

Velin92:
> > Generally looks good, some things to sort out, also `forwarder` is not used anywhere in the app as a name convention, we usually use the `sender` noun. Could we use it here too?
> 
> I want to disambiguate between the actual sender of the message and the person who forwarded us the key with which we decrypted the message, I could perhaps use `KeyForwarder` to clarify that? Maybe `KeySender`.

oh yeah that would definitely make things more clear (I in fact thought it was the sender of the message just being called forwarder) `KeySender` should be fine.

kaylendog:
> oh yeah that would definitely make things more clear (I in fact thought it was the sender of the message just being called forwarder) `KeySender` should be fine.

I do think I prefer `KeyForwarder` if you don't mind? It's distinct from the routine "key sender", which would just be the author of the message, i.e. the "sender", as used in the EX codebase. In the case of history sharing, though, anyone can forward us the keys to decrypt anybody else's messages, and I think the notion of "key forwarder" makes that clear.

ElementX/Resources/Localizations/en.lproj/Untranslated.strings
```diff
@@ -14,3 +14,4 @@
 "soft_logout_clear_data_submit" = "Clear all data";
 "soft_logout_clear_data_dialog_title" = "Clear data";
 "soft_logout_clear_data_dialog_content" = "Clear all data currently stored on this device?\nSign in again to access your account data and messages.";
+"encryption_forwarder_dialog_content" = "%1$s (%2$s) shared this message since you were not in the room when it was sent.";
```
Velin92: I see L10n strings have been made in localazy and are already in the codebase, so I guess we can remove these ones from the Untranslated strings

ElementX/Sources/Screens/Timeline/View/Style/TimelineItemBubbledStylerView.swift
```diff
@@ -385,6 +392,8 @@ struct TimelineItemBubbledStylerView_Previews: PreviewProvider, TestablePreview
             .padding(.bottom, 20)
         encryptionAuthenticity
             .previewDisplayName("Encryption Indicators")
+        encryptionForwarder
+            .previewDisplayName("Encryption Forwarder Info")
```
Velin92, line 397: add `previewLayout(.sizeThatFits)` before `previewDisplayName` and regenerate the previews since internally this is just a VStack, this usually makes the previews only use the size of the preview

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
```diff
@@ -674,6 +675,8 @@ class TimelineViewModel: TimelineViewModelType, TimelineViewModelProtocol {
             actionsSubject.send(.displayResolveSendFailure(failure: failure,
                                                            sendHandle: sendHandle))
             
+        } else if let forwarderMessage = eventTimelineItem.properties.encryptionForwarder?.message, appSettings.enableKeyShareOnInvite {
+            displayAlert(.encryptionForwarder(forwarderMessage))
```
richvdh: I don't think this needs to be dependent on enableKeyShareOnInvite?

ElementX/Resources/Localizations/en.lproj/Untranslated.strings
richvdh: relatedly: it seems like Localazy has a key `crypto_event_key_forwarded_unknown_profile_dialog_content` which doesn't give space for the mxid and is therefore not useful.

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
kaylendog: I believe I added this in the case the feature is turned on and off again - I was a bit worried the items we get from the store will have the `forwarder` field set, causing the badge to display when the feature is off. It might be a little overly cautious?

richvdh: Yup, it would be fine to show the badge should be shown even if the feature is off, as long as the keys were in fact forwarded by another user,

ElementX/Resources/Localizations/en.lproj/Untranslated.strings
richvdh: > relatedly: it seems like Localazy has a key `crypto_event_key_forwarded_unknown_profile_dialog_content` which doesn't give space for the mxid and is therefore not useful.

we discussed this in chat. I was mistaken: `crypto_event_key_forwarded_unknown_profile_dialog_content` is used when we don't have the user's displayname, which is fine.

PR #4942, opened 2026-01-09
Create room redesign and refactor

Velin92:
- Redesigned the room so that it matches the create space UI (more redesigned specifically for the room to come, this will work)
- Create space UI ready (just need to hook it to its CTA, will do that in the next PR)
- Updated room creation logic to depend on the access value that has been selected
- Updated strings
- Moved to the top the icon alignment for the access selection (also done in Sec & Priv)
- Updated tests and previews

The thing that is left is to update the room avatar to have an edit badge when an image is present, but will do that in a follow up PR.


ElementX/Sources/Screens/CreateRoomScreen/CreateRoomScreenModels.swift
```diff
@@ -23,12 +23,13 @@ enum CreateRoomScreenViewModelAction {
...
 
 struct CreateRoomScreenViewState: BindableState {
+    let isSpace: Bool
     var roomName: String
     let serverName: String
     let isKnockingFeatureEnabled: Bool
     var aliasLocalPart: String
     var bindings: CreateRoomScreenViewStateBindings
-    var avatarURL: URL?
+    var mediaInfo: MediaInfo?
```
pixlwave: Maybe we should call this `avatarMediaInfo` so its immediately obvious what it is?

ElementX/Sources/Screens/CreateRoomScreen/CreateRoomScreenViewModel.swift
```diff
@@ -123,34 +110,18 @@ class CreateRoomScreenViewModel: CreateRoomScreenViewModelType, CreateRoomScreen
...
-                    old.bindings.isRoomPrivate == new.bindings.isRoomPrivate &&
-                    old.bindings.isKnockingOnly == new.bindings.isKnockingOnly &&
-                    old.aliasLocalPart == new.aliasLocalPart
-            }
-            .sink { [weak self] state in
-                guard let self else { return }
-                updateParameters(state: state)
-            }
-            .store(in: &cancellables)
-
```
pixlwave, line 153: Nice 👍

ElementX/Sources/Screens/CreateRoomScreen/View/CreateRoomScreen.swift
```diff
@@ -207,13 +199,48 @@ struct CreateRoomScreen: View {
     }
 }
 
+private extension CreateRoomAccessType {
```
pixlwave: Super minor, but might be nicer to create a `CreateRoomAccessRow: View` for this instead and then we can get ride of the `func` above and do it all in its own file.

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -474,25 +474,26 @@ class ClientProxy: ClientProxyProtocol {
...
+                    accessType: CreateRoomAccessType,
+                    isSpace: Bool,
                     userIDs: [String],
                     avatarURL: URL?,
                     aliasLocalPart: String?) async -> Result<String, ClientProxyError> {
         do {
             let parameters = CreateRoomParameters(name: name,
                                                   topic: topic,
-                                                  isEncrypted: isRoomPrivate,
+                                                  isEncrypted: isSpace ? false : accessType.isEncrypted,
```
pixlwave: Maybe a stupid question, but is this what we actually want to do? Whats wrong with creating encrypted spaces?

PreviewTests/Sources/__Snapshots__/PreviewTests/securityAndPrivacyScreen.Ask-to-join-room-iPad-en-GB.png
pixlwave, line 1: Double checking: is the Ask To Join label meant to have changed on this one?

ElementX/Sources/Services/Client/ClientProxy.swift
Velin92: actually I am not sure about this, will check with Product

PreviewTests/Sources/__Snapshots__/PreviewTests/securityAndPrivacyScreen.Ask-to-join-room-iPad-en-GB.png
Velin92: good catch! they shouldn't be linked indeed

ElementX/Sources/Services/Client/ClientProxy.swift
Velin92: okay we will keep the same behaviour as rooms for E2EE

PR #4949, opened 2026-01-12
Fix how email addresses and links with trailing closing brackets are detected

stefanceriu:


ElementX/Sources/Other/Extensions/CharacterSet.swift
```diff
@@ -30,4 +30,10 @@ extension CharacterSet {
     
     static let matrixUserIDAllowedCharacters = CharacterSet(charactersIn: "abcdefghijklmnopqrstuvwxyz0123456789._=-/@:")
     static let roomAliasAllowedCharacters = CharacterSet(charactersIn: "abcdefghijklmnopqrstuvwxyz0123456789!$&‘()*+/;=?@[]-._:#")
+    
+    static let punctuationWithoutClosingBracketCharacterSet: CharacterSet = {
```
pixlwave: We should probably align the name of this with the others in here (and the system ones) and call it `punctuationWithoutClosingBracketCharacters`

ElementX/Sources/Other/HTMLParsing/AttributedStringBuilder.swift
```diff
@@ -330,12 +330,17 @@ struct AttributedStringBuilder: AttributedStringBuilderProtocol {
...
-            guard let matchRange = Range(match.range, in: string) else {
+            guard let matchRange = Range(match.range, in: string), let url = match.url else {
                 return nil
             }
             
-            let link = String(string[matchRange]).asSanitizedLink
-            return TextParsingMatch(type: .link(urlString: link), range: match.range)
+            // If the NSDataDetector found a hyperlink then sanitise it
+            if url.absoluteString.contains("http") {
+                // Use the underlying string so it gets an `https` scheme if it didn't have any
```
pixlwave: Maybe lets be a bit more specific here…

```suggestion
            if url.scheme.contains("http") {
                // Use the underlying string so it gets an `https` scheme if it didn't have one.
```

pixlwave (approved):
Nice, thanks 👍

ElementX/Sources/Other/HTMLParsing/AttributedStringBuilder.swift
stefanceriu: Good point 👍

PR #4957, opened 2026-01-14
Create space flow

### LINKED ISSUE #4688: [Task] Allow the create room flow to be configured to create a space.

Velin92:
fixes #4688

This PR allows users to create a spaces through the create space CTA in the Space Explorer Screen.

It also contains
- A redesign of the avatar button.
- A power level override for spaces.

Also the knocking feature is disabled for spaces regardless of the knocking FF

NOTE: There seems to be a bug in Ruma that skips content overrides that match the default values, however since HSs can have their own override this may create issues where you want to force stuff like Invite to PL 0, but Ruma doesn't include it in the request since by specs the default value is already 0, and then the HSs since doesn't find any override uses their own preset which is 50. This bug will likely be fixed in a future SDK update.

### DIFF
--- a/ElementX/Sources/FlowCoordinators/ChatsTabFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/ChatsTabFlowCoordinator.swift
@@ -585,7 +585,7 @@ class ChatsTabFlowCoordinator: FlowCoordinatorProtocol {
                     case .space(let spaceRoomListProxy):
                         // This also automatically handles selecting the space.
                         stateMachine.processEvent(.selectRoom(roomID: spaceRoomListProxy.id, via: [], entryPoint: .room))
-                    case .none:
+                    case .cancelled:
                         break
                     }
                 case .showRoomDirectory:


ElementX/Sources/FlowCoordinators/SpacesTabFlowCoordinator.swift
```diff
@@ -81,16 +88,29 @@ class SpacesTabFlowCoordinator: FlowCoordinatorProtocol {
...
         
+        stateMachine.addRouteMapping { event, fromState, userInfo in
+            guard event == .selectSpace, case .spacesScreen = fromState else { return nil }
+            guard let spaceRoomListProxy = userInfo as? SpaceRoomListProxyProtocol else { fatalError("A space proxy must be provided.") }
+            return .spacesScreen(selectedSpaceID: spaceRoomListProxy.id)
+        } handler: { [weak self] context in
+            guard let self, let spaceRoomListProxy = context.userInfo as? SpaceRoomListProxyProtocol else { return }
+            startSpaceFlow(spaceRoomListProxy: spaceRoomListProxy)
+        }
+
```
pixlwave: Errrr, isn't this identical to the one below it?

ElementX/Sources/FlowCoordinators/SpacesTabFlowCoordinator.swift
```diff
@@ -167,4 +207,38 @@ class SpacesTabFlowCoordinator: FlowCoordinatorProtocol {
...
+        var spaceRoomListProxy: SpaceRoomListProxyProtocol?
+        flowCoordinator.actionsPublisher
+            .sink { [weak self] action in
+                guard let self else { return }
+                switch action {
+                case .finished(let result):
+                    switch result {
+                    case .space(let value):
+                        spaceRoomListProxy = value
+                    case .room, .none:
```
pixlwave: Optional, but it wouldn't seem unreasonable to me to to fatalError if you get a room back from this flow :)

ElementX/Sources/FlowCoordinators/StartChatFlowCoordinator.swift
```diff
@@ -11,12 +11,28 @@ import Foundation
 import SwiftState
 
 enum StartChatFlowCoordinatorAction {
-    case finished(roomID: String?)
+    case finished(Result?)
```
pixlwave: I have **no** idea why I made that roomID optional… Could we maybe separate out the action here to highlight cancellation as not finishing like we do everywhere else perhaps 😅

```suggestion
    case finished(Result)
    case cancelled
```

pixlwave: You could probably also go one step further and get rid of the nested `Result` type as I don't think it provides much value at this level:

```suggestion
    case createdRoom(roomID: String)
    case createdSpace(SpaceRoomListProxyProtocol)
    case cancelled
```

ElementX/Sources/Other/SwiftUI/Views/ToolbarButton.swift
pixlwave, line 1: Noice 😎

ElementX/Sources/Screens/CreateRoomScreen/CreateRoomScreenViewModel.swift
```diff
@@ -233,7 +237,19 @@ class CreateRoomScreenViewModel: CreateRoomScreenViewModelType, CreateRoomScreen
...
-            actionsSubject.send(.createdRoom(roomProxy))
+            
+            var spaceRoomListProxy: SpaceRoomListProxyProtocol?
+            if state.isSpace {
+                switch await userSession.clientProxy.spaceService.spaceRoomList(spaceID: roomProxy.id) {
+                case .success(let value):
+                    spaceRoomListProxy = value
+                case .failure:
+                    MXLog.error("Failed to get space room list for newly created space with id: \(roomProxy.id)")
+                    userIndicatorController.submitIndicator(.init(title: L10n.errorUnknown))
```
pixlwave, line 248: Should this `return` if there's an error rather than allowing the flow to continue?

ElementX/Sources/Screens/CreateRoomScreen/View/CreateRoomScreen.swift
```diff
@@ -101,15 +111,27 @@ struct CreateRoomScreen: View {
...
-                    .scaledFrame(size: 70, relativeTo: .title)
-                    .background(.compound.bgSubtlePrimary, in: Circle())
+                CompoundIcon(\.takePhoto, size: .medium, relativeTo: .title)
+                    .foregroundColor(.compound.iconPrimary)
+                    .scaledFrame(size: 50, relativeTo: .title)
+                    .background(.compound.bgCanvasDefault, in: Circle())
+                    .overlay(
+                        Circle()
+                            .stroke(.compound.borderInteractiveSecondary, lineWidth: 1)
+                    )
```
pixlwave: Super minor, but pretty sure that one has been deprecated:

```suggestion
                    .overlay {
                        Circle()
                            .stroke(.compound.borderInteractiveSecondary, lineWidth: 1)
                    }
```

PreviewTests/Sources/__Snapshots__/PreviewTests/createRoom.Create-Room-with-avatar-iPad-en-GB.png
pixlwave, line 1: Maybe we should add a snapshot expectation so that it waits until the image has been loaded?

ElementX/Sources/Screens/CreateRoomScreen/CreateRoomScreenViewModel.swift
Velin92: I think is fine, the space has already been created regardless and you can still invite people in it, also te likelyhood of the space graph to fail generating the list for an already created and joined space that is also empty are probably non-existent, but in case it fails, in theory it will be handled by a normal dismiss and a simple "something went wrong" error.

ElementX/Sources/FlowCoordinators/StartChatFlowCoordinator.swift
Velin92: Oh yeah I rather like this, however I should point out that you don't necessarily return a `createdRoom` from this flow, you are also going to return a room that was already existing (existing DM) or a room found through its alias.

ElementX/Sources/FlowCoordinators/SpacesTabFlowCoordinator.swift
Velin92: Considering https://github.com/element-hq/element-x-ios/pull/4957#discussion_r2694151422

ElementX/Sources/FlowCoordinators/SpacesTabFlowCoordinator.swift
Velin92: whoops my bad

PreviewTests/Sources/__Snapshots__/PreviewTests/createRoom.Create-Room-with-avatar-iPad-en-GB.png
Velin92: I was thinking the same, but how to do that? the `AsyncImage` seems to not have a way tell back when the loading is completed

pixlwave: Ah, my bad I forgot we were using that here. Never mind :)

PR #4966, opened 2026-01-16
Enable Swift 6 and default actor isolation on the Compound Inspector app.

pixlwave:
I also upgraded the project for Xcode 26, although then disabled `ENABLE_USER_SCRIPT_SANDBOXING` as Hyperion is using `rsync` (?!) to copy its assets.


stefanceriu (approved):
Woo, first blood! 🪱

PR #4972, opened 2026-01-19
Create room with space selection

### LINKED ISSUE #4685: [Task] Add parent Space selection when creating a room from the room list.

- The join rules picker should be updated when a Space is selected.

Velin92:
fixes #4685 

Review commit by commit

This PR implements creating a room within a space, both with a pre-existing selected space (even if such case is not hooked yet in the navigation due to a lack of the appropriate screen, but is fully tested) and through the selection of the space among a list of joined spaces that the user has the permission to edit (add children to).

https://github.com/user-attachments/assets/ebab5552-23fd-4660-83f2-9ee074e3c387

### DIFF
--- a/ElementX/Sources/Screens/CreateRoomScreen/View/CreateRoomScreen.swift
+++ b/ElementX/Sources/Screens/CreateRoomScreen/View/CreateRoomScreen.swift
@@ -43,7 +43,7 @@ struct CreateRoomScreen: View {
                 selectSpaceSection
             }
             roomAccessSection
-            if !context.viewState.roomAccessType.isPrivate {
+            if !context.viewState.roomAccessType.isVisibilityPrivate {
                 roomAliasSection
             }
         }
--- a/ElementX/Sources/Screens/CreateRoomScreen/CreateRoomScreenViewModel.swift
+++ b/ElementX/Sources/Screens/CreateRoomScreen/CreateRoomScreenViewModel.swift
@@ -136,7 +136,7 @@ class CreateRoomScreenViewModel: CreateRoomScreenViewModelType, CreateRoomScreen
         context.$viewState
             .dropFirst()
             .map(\.roomAccessType)
-            .filter(\.isPrivate)
+            .filter(\.isVisibilityPrivate)
             .removeDuplicates()
             .receive(on: DispatchQueue.main)
             .sink { [weak self] _ in
@@ -156,7 +156,7 @@ class CreateRoomScreenViewModel: CreateRoomScreenViewModelType, CreateRoomScreen
                     return
                 }
                 
-                guard !state.roomAccessType.isPrivate,
+                guard !state.roomAccessType.isVisibilityPrivate,
                       let canonicalAlias = String.makeCanonicalAlias(aliasLocalPart: aliasLocalPart,
                                                                      serverName: state.serverName) else {
                     // While is empty or private room we don't change or display the error
@@ -209,7 +209,7 @@ class CreateRoomScreenViewModel: CreateRoomScreenViewModelType, CreateRoomScreen
         showLoadingIndicator()
         
         // Better to double check the errors also when trying to create the room
-        if !state.roomAccessType.isPrivate {
+        if !state.roomAccessType.isVisibilityPrivate {
             guard let canonicalAlias = String.makeCanonicalAlias(aliasLocalPart: state.aliasLocalPart,
                                                                  serverName: state.serverName),
                 isRoomAliasFormatValid(alias: canonicalAlias) else {
@@ -261,7 +261,7 @@ class CreateRoomScreenViewModel: CreateRoomScreenViewModelType, CreateRoomScreen
                                                         isSpace: state.isSpace,
                                                         userIDs: [], // The invite users screen is shown next so we don't need to invite anyone right now.
                                                         avatarURL: avatarURL,
-                                                        aliasLocalPart: state.roomAccessType.isPrivate ? nil : state.aliasLocalPart) {
+                                                        aliasLocalPart: state.roomAccessType.isVisibilityPrivate ? nil : state.aliasLocalPart) {
         case .success(let roomID):
             guard case let .joined(roomProxy) = await userSession.clientProxy.roomForIdentifier(roomID) else {
                 state.bindings.alertInfo = AlertInfo(id: .failedCreatingRoom,
--- a/ElementX/Sources/Services/Client/ClientProxy.swift
+++ b/ElementX/Sources/Services/Client/ClientProxy.swift
@@ -1378,15 +1378,15 @@ private extension CreateRoomAccessType {
     }
     
     var visibility: RoomVisibility {
-        isPrivate ? .private : .public
+        isVisibilityPrivate ? .private : .public
     }
     
     var preset: RoomPreset {
-        isPrivate ? .privateChat : .publicChat
+        isVisibilityPrivate ? .privateChat : .publicChat
     }
     
     var historyVisibilityOverride: RoomHistoryVisibility? {
-        isPrivate ? .invited : nil
+        isVisibilityPrivate ? .invited : nil
     }
     
     var joinRuleOverride: JoinRule? {
--- a/ElementX/Sources/Services/Client/ClientProxyProtocol.swift
+++ b/ElementX/Sources/Services/Client/ClientProxyProtocol.swift
@@ -55,7 +55,7 @@ enum CreateRoomAccessType: Equatable {
     case askToJoin
     case `private`
     
-    var isPrivate: Bool {
+    var isVisibilityPrivate: Bool {
         switch self {
         case .private, .spaceMembers, .askToJoinWithSpaceMembers:
             true


ElementX/Sources/Screens/CreateRoomScreen/CreateRoomScreenModels.swift
```diff
@@ -59,18 +60,46 @@ struct CreateRoomScreenViewState: BindableState {
...
+            availableAccessTypes = [.public]
+            if isKnockingFeatureEnabled {
+                availableAccessTypes.append(.askToJoin)
+            }
+        }
+        availableAccessTypes.append(.private)
+        return availableAccessTypes
+    }
+    
+    var roomAccessType: CreateRoomAccessType {
```
pixlwave, line 83: I'm very confused about this, why do we have `CreateRoomScreenAccessType` and `CreateRoomAccessType` both being used in this screen?

ElementX/Sources/FlowCoordinators/StartChatFlowCoordinator.swift
```diff
@@ -193,6 +193,7 @@ class StartChatFlowCoordinator: FlowCoordinatorProtocol {
     
     private func presentCreateRoomScreen(isSpace: Bool, isRoot: Bool) {
         let createParameters = CreateRoomScreenCoordinatorParameters(isSpace: isSpace,
+                                                                     spaceSelectionMode: !isSpace && flowParameters.appSettings.createSpaceEnabled ? .list : nil,
```
pixlwave, line 196: It feels to me like we should pass `AppSettings` into the view model and let it decide (along with the existing `isSpace` parameter) what the selection mode will be as it isn't particularly clear from the outside what this means so it should be the screen's responsibility to pick the right mode.

ElementX/Sources/Screens/CreateRoomScreen/View/CreateRoomScreen.swift
```diff
@@ -392,6 +449,56 @@ struct CreateRoom_Previews: PreviewProvider, TestablePreview {
...
+                                                  spaceSelectionMode: .selected(mockedSpace),
+                                                  shouldShowCancelButton: false,
+                                                  userSession: userSession,
+                                                  analytics: ServiceLocator.shared.analytics,
+                                                  userIndicatorController: UserIndicatorControllerMock(),
+                                                  appSettings: appSettings)
+        
+        viewModel.context.selectedAccessType = .askToJoinWithSpaceMembers
+        return viewModel
+    }()
```
pixlwave, line 398: I think its about time we added a `makeViewModel` function to this screen, the repetition in these previews is started to get out of hand 😅.

ElementX/Sources/Services/Spaces/SpaceServiceProxyProtocol.swift
```diff
@@ -23,6 +23,8 @@ protocol SpaceServiceProxyProtocol {
     func leaveSpace(spaceID: String) async -> Result<LeaveSpaceHandleProxy, SpaceServiceProxyError>
     /// Returns all the parent spaces of a child that user has joined.
     func joinedParents(childID: String) async -> Result<[SpaceServiceRoomProtocol], SpaceServiceProxyError>
+    /// Returns all the parent spaces that can be edited by the user
```
pixlwave:
```suggestion
    /// Returns all the joined spaces that can be edited by the user
```

ElementX/Sources/Services/Spaces/SpaceServiceProxy.swift
```diff
@@ -80,11 +84,11 @@ class SpaceServiceProxy: SpaceServiceProxyProtocol {
         do {
             return try await .success(spaceService.removeChildFromSpace(childId: childID, spaceId: spaceID))
         } catch {
-            MXLog.error("Failed to remove child \(childID) to space \(spaceID)")
+            MXLog.error("Failed to remove child \(childID) to space \(spaceID): \(error)")
             return .failure(.sdkError(error))
         }
     }
-    
+
```
pixlwave:
```suggestion
    
```

ElementX/Sources/Screens/CreateRoomScreen/View/CreateRoomScreen.swift
```diff
@@ -215,6 +222,35 @@ struct CreateRoomScreen: View {
...
+                        kind: .navigationLink {
+                            context.showSpaceSelectionSheet = true
+                        })
+            } else {
+                ListRow(label: .default(title: L10n.screenCreateRoomSpaceSelectionNoSpaceTitle,
+                                        description: L10n.screenCreateRoomSpaceSelectionNoSpaceDescription,
+                                        icon: CompoundIcon(\.homeSolid, size: .small, relativeTo: .body)
+                                            .foregroundColor(.compound.iconPrimary)
+                                            .background(.compound.bgSubtleSecondary)
+                                            .scaledFrame(size: 32)
```
pixlwave: I think you want these 2 the other way around, otherwise you're just adding whitespace around the icon :)

```suggestion
                                            .scaledFrame(size: 32)
                                            .background(.compound.bgSubtleSecondary)
```

UnitTests/Sources/CreateRoomViewModelTests.swift
```diff
@@ -219,4 +221,131 @@ class CreateRoomScreenViewModelTests: XCTestCase {
         XCTAssertEqual(context.viewState.aliasLocalPart, "hello-matrix!")
         XCTAssertEqual(context.viewState.roomName, "Hello# Matrix!")
     }
+    
+    func testCreateRoomInASelectedSpaceFromTheList() async throws {
+        let spaces = [SpaceServiceRoomProtocol].mockJoinedSpaces2
+        setup()
+        
+        // Given a form with a blank topic.
```
pixlwave: I don't think this test has much to do with the topic, same for the other comments 😅

ElementX/Sources/FlowCoordinators/StartChatFlowCoordinator.swift
Velin92: I had the same thought at first but then I realised this would extremely overcomplicate the testing process and setup, also we will have in the future the possibility of creating a space in another space, so this will make the transition very easy, while adding this extra layer for an FF which will soon be removed is not really worth.

pixlwave: > I had the same thought at first but then I realised this would extremely overcomplicate the testing process and setup

Would it? We're already passing `isSpace` so it would just be a cased of passing the settings in.

> also we will have in the future the possibility of creating a space in another space, so this will make the transition very easy

Couldn't whatever you change at this level simply be made at the view model level though?

ElementX/Sources/Screens/CreateRoomScreen/CreateRoomScreenModels.swift
Velin92: One is just the Ui value, that does not hold any associated value (since would make the implementation not work properly when creating the ListRow of selection items for the picker). The other one gets computed by using the UI value and the selected space, and is the client proxy value used to identify and build all the values of encryption, joinrule override, PL and so on according to the selected join rule and so contains the associated value of the selected space for the space members based values.

ElementX/Sources/FlowCoordinators/StartChatFlowCoordinator.swift
Velin92: The flow should decide if you are creating a space or a room, and should also decide if are creating into a selected space or from a list of spaces, and these two values should be logically separated since they could exist in any combination.
The FF is just a temporary thing and I don't think we should build our design too much over a thing that is going to get removed soon.

What I am trying to say is that... being a space and how the room is included into a space are two external and unrelated inputs that decide the UI of the view, and the flow decides this. The setting is just an abstraction to "override" the flow behaviour for now and due to the lack of the design for creating a space in a space, but I think is more scalable to handle these values externally.

ElementX/Sources/Screens/CreateRoomScreen/CreateRoomScreenModels.swift
pixlwave: Ok, but I think you could add `isPrivate` to `CreateRoomScreenAccessType`, use the one from the bindings everywhere in this screen/vm, and then move this code directly into `CreateRoomScreenViewModel.createRoom` just before its used in `switch await userSession.clientProxy.createRoom(…, accessType:, …)`.

Then we wouldn't need the 2 different types available from the view state. WDYT?

Velin92: This would mean replicating isPrivate, which I would prefer to avoid since someone changing it might miss how its behaviours also impacts the logic of the room creation

pixlwave: Ok we discussed this. Turns out the `isPrivate` on the `CreateRoomAccessType` isn't just `self == .private`. It's still hard to reason about but that explains it.

pixlwave (approved):

ElementX/Sources/Screens/CreateRoomScreen/CreateRoomScreenModels.swift
Velin92: modified `isPrivate` to `isVisibilityPrivate`

PR #4973, opened 2026-01-19
Set a custom user agent on the Element Call web view

subnix:
This patch replaces the default User-Agent header with the common one.

The old User-Agent:
```
Mozilla/5.0 (iPhone; CPU iPhone OS 18_7 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148
```
The new one:
```
Element X/26.01.1 (Simulator (iPhone 17 Pro); iOS 26.2; Scale/3.00)
```

### Pull Request Checklist

- [ ] Voiceover enabled.
- [x] No UI changes

### DIFF
--- a/ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
+++ b/ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
@@ -123,6 +123,8 @@ private struct CallView: UIViewRepresentable {
             webView.navigationDelegate = self
             webView.isInspectable = true
             
+            webView.customUserAgent = UserAgentBuilder.makeASCIIUserAgent()
+            
             // https://stackoverflow.com/a/77963877/730924
             webView.allowsLinkPreview = true
             


stefanceriu:
> The old User-Agent:
> 
> ```
> Mozilla/5.0 (iPhone; CPU iPhone OS 18_7 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148
> ```

How did you find that? I see it as `Element X/26.01.1 (iPhone 13; iOS 26.2; Scale/3.00)` in network logs 🤔

subnix:
it's from server's access logs:
```json
{"uri":"/.well-known/matrix/client","headers":{"Origin":["file://"],"User-Agent":["Mozilla/5.0 (iPhone; CPU iPhone OS 18_7 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148"]}}
{"uri":"/livekit/jwt/sfu/get","headers":{"User-Agent":["Mozilla/5.0 (iPhone; CPU iPhone OS 18_7 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148"]}}
```

Basically, that's the reason why I made this patch. I investigated the unusual User-Agent and ended up here. The client uses the correct User-Agent for communication with a Matrix server, but not for calls. Since a separate webview is used for calls, it requires the dedicated configuration of UA.

stefanceriu:
Huh I was expecting it to inherit the user agent seeing as it's just an embedded WKWebView.

In that case, looks good to me, thank you very much for contributing! 🙏

stefanceriu (approved):

PR #4975, opened 2026-01-20
Tweak the RoomHeaderView to match Figma.

pixlwave:
Follow on from #4965, this PR makes the following changes:
- Make the font size slightly smaller.
- Use custom padding to match Figma properly (the avatar hugs the glass more closely).
- Switch back to using a tap gesture (in combination with an interactive glass effect) so the padding takes effect.
- As a consequence of the custom view, fix a tiny animation bug where the top outline was clipped during pop/push as the view was taller than it should have been.

### DIFF
--- a/ElementX/Sources/Other/SwiftUI/Views/RoomHeaderView.swift
+++ b/ElementX/Sources/Other/SwiftUI/Views/RoomHeaderView.swift
@@ -25,8 +25,8 @@ struct RoomHeaderView: View {
             // On iOS 26+ we use the toolbarRole(.editor) to leading align.
             content
                 // Not using a Button here so that we get our custom padding around the avatar. This also
-                // helps fix a bug on iOS 26.2 where the top pixel gets clipped during the push/pop animation
-                // as the resulting frame is no longer the full height of the navigation bar.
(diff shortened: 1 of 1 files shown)


ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
```diff
@@ -226,6 +226,10 @@ struct RoomScreen_Previews: PreviewProvider, TestablePreview {
     }
     
     static func makeViewModels(canSendMessage: Bool = true, hasSuccessor: Bool = false) -> ViewModels {
+        // TimelineItemBubbledStylerView sets enableKeyShareOnInvite to true which if run
+        // before these tests, ends up adding a banner to the snapshots.
+        AppSettings.resetAllSettings()
```
stefanceriu, line 231: Perhaps we should run `AppSettings.resetAllSettings()` as part of the preview test templates 🤔

pixlwave: Oh yeah interesting. Not for this PR, but that does make sense to me.

PR #4976, opened 2026-01-20
Fix the setting button's layout and animation on liquid glass

stefanceriu:
Previously it had an incorrect aspect ratio and was breaking when popping back from a room.

https://github.com/user-attachments/assets/5346111a-98ce-4a31-b5ed-4a19322b6500


ElementX/Sources/Screens/HomeScreen/View/HomeScreen.swift
pixlwave, line 1: Errr can we do the same on `SpacesScreen` too please 🥺

stefanceriu: Good point 👍

stefanceriu:
Yeah, I ran them yesterday and found no failures, don't expect any for the spaces screen either.

PR #4980, opened 2026-01-21
Normalise HTML whitespaces using SwiftSoup's own text()/TextNode.normaliseWhitespace methods

### LINKED ISSUE #4431: spaced text when multiple rows
### Steps to reproduce

Same issue as on android https://github.com/element-hq/element-x-android/issues/4695

Render is off, its spaced and its annoying. Rendering works normally on EW

#### What did you expect?

Render to work properly, not to have extra spaces inbetween

#### What happened instead?

Space, rendering not working...

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

28.08.03

### Homeserver

_No response_

### Will you send logs?

No

### LINKED ISSUE #4856: app doesn't handle whitespace in html formatted messages
### Steps to reproduce

1. Send/Find an html formatted message that has a lot of unnecessary literal spaces (the github maubot seems to like doing this)
2. view the message

One of the following messages:
```json
{
  "content": {
    "body": "**[mautrix/signal]** tulir pushed [2 commits](https://github.com/mautrix/signal/compare/3b4b6cf753df...2c78df263847) to main:\n● `c7905771` libsignal: update to v0.86.8\n● `2c78df26` changelog: update",
    "com.beeper.linkpreviews": [],
    "com.beeper.per_message_profile": {
      "avatar_url": "mxc://maunium.net/LqWtZboHroLXjDwhwxHlarNS",
      "displayname": "tulir",
      "id": "4224639"
    },
    "format": "org.matrix.custom.html",
    "formatted_body": "<strong>[<a data-mautrix-exclude-plaintext href=\"https://github.com/mautrix/signal\">mautrix/signal</a>]</strong> <a data-mautrix-exclude-plaintext href=\"https://github.com/tulir\">tulir</a>      pushed\n        <a href=\"https://github.com/mautrix/signal/compare/3b4b6cf753df...2c78df263847\">2 commits</a>\n    to\n main:<ul>         <li>\n            <a data-mautrix-exclude-plaintext href=\"https://github.com/mautrix/signal/commit/c79057718a98a1c18089a8d3dc8daef6a7f120c1\"><code>c7905771</code></a>\n            libsignal: update to v0.86.8\n        </li>\n        <li>\n            <a data-mautrix-exclude-plaintext href=\"https://github.com/mautrix/signal/commit/2c78df26384716d7c338a59d5f928395f8607089\"><code>2c78df26</code></a>\n            changelog: update\n        </li>\n </ul>",
    "msgtype": "m.notice",
    "xyz.maubot.github.webhook": {
      "delivery_ids": [
        "c4e7a7a2-d695-11f0-8770-e368efa7c801"
      ],
      "event_type": "push"
    }
  },
  "origin_server_ts": 1765459919794,
  "room_id": "!mausgp8oQVwgry4kjfKSxqNj-O9Nkez81Cl591w8jkI",
  "sender": "@github:maunium.net",
  "type": "m.room.message",
  "unsigned": {
    "membership": "join"
  },
  "event_id": "$8waw64AqZRZDnY6qaJncVItGP2j3FswZxs-ak2TXCr4",
  "user_id": "@github:maunium.net"
}
```

#### What did you expect?

I opened this as a bug, because it used to work correctly and is really annoying. I believe this to be a regression due to switch to the new attributed string builder.

The html parser should implement (a simple version of) the css whitespace algorithm and fold and trim consecutive whitespace where appropriate.
While html preserves whitespace in the dom (with exceptions that the parser lib should already implement), sending user agents (including the rust sdk and element web) seem to assume css-like whitespace handling anyway and produce html with newlines between tags that aren't necessary and should be stripped to avoid unnecessary blank lines. Then there are the extremes of automation that templates html with all sorts of whitespace everywhere (e.g., the github maubot, some twim scripts). The newlines between tags also can't be relied upon, as can be seen by the missing break before the beginning of the list (`... main:<ul>         <li>...`) in the following image.
Luckily a simplified version of the algorithm is relatively easy to implement, I have done so recently while implementing html in my own client when I noticed all the whitespace in formatted messages. A single mutable boolean that tracks whitespace while traversing the dom to build the result and inserting newlines around block elements is enough to produce reasonable results for languages with (or without) word separators. Doing both well requires detecting the languages on both sides of newlines in the source and is i18n stuff I luckily didn't have to deal with, maybe iOS has an API for that?

References that may be helpful (for full blown css, more complicated than necessary):
https://drafts.csswg.org/css-text-4/#white-space-processing
https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Text/Whitespace (a bit more approachable)

The result should be something similar to the following (screenshot from an old build with the new string builder disabled):

Image

#### What happened instead?

Image

### Your phone model

iPhone 13 mini/iPad Pro 11" 2nd-gen

### Operating system version

iOS/iPadOS 26.1

### Application version

25.12.1 (201)

### Homeserver

_No response_

### Will you send logs?

No

stefanceriu:
Fix #4856, Fix #4431


PreviewTests/Sources/__Snapshots__/PreviewTests/formattedBodyText.iPhone-en-GB-0.png
pixlwave, line 1: Looking at H2, H3, H4 etc, there appears to be an extra space before them (but not H1), is that expected?

PreviewTests/Sources/__Snapshots__/PreviewTests/messageText.With-block-quote-iPad-en-GB.png
pixlwave, line 1: I'm not really following how this preview works given there isn't any blockquote formatting, but the `<p>` appears not to create a new paragraph and there isn't even a space between the first sentence and the second one 🤔

PreviewTests/Sources/__Snapshots__/PreviewTests/formattedBodyText.iPhone-en-GB-0.png
pixlwave: This is the same if you send e.g. the following markdown

```
no formatting
**some** formatting
```

There's an extra space before **some**.

stefanceriu: I think it's because the headers are defined in fixtures with new lines after them, which get turned into spaces by SwiftSoup. Why that happens I really don't know but we should probably trust it.

PreviewTests/Sources/__Snapshots__/PreviewTests/messageText.With-block-quote-iPad-en-GB.png
stefanceriu: I got super confused by it as well, that's why I added that extar paragraph in `testSingleBlockquote`

PreviewTests/Sources/__Snapshots__/PreviewTests/formattedBodyText.iPhone-en-GB-0.png
marwing: Sorry for butting in here, but I would argue these spaces (before headings and after `<br>` tags) are not expected (just try it in any browser, e.g. by rendering such a message in EW or an otherwise blank html document) and it doesn't match my current (admittedly simplified) mental model of css whitespace processing. As a user I would certainly open another issue about that or reopen #4856 because of an incomplete fix if I hadn't noticed this PR.

stefanceriu: Turns out they're coming from newlines outside of the html body, because there's none. We couldn't find a reasonable solution that was going to fix all the cases without a massive refactoring so we're going to have to see how this actually goes in practice, I imagine it can't be worse than it is.

PreviewTests/Sources/__Snapshots__/PreviewTests/messageText.With-block-quote-iPad-en-GB.png
stefanceriu: Figured out what this was: blockquotes should have an extra new line on them but then that would need to be stripped out on component based rendering solutions. MessageText isn't one of those cases so it just renders the 2 lines together. Agreed to do nothing for now.

PR #4984, opened 2026-01-21
Create a room in a selected space

### LINKED ISSUE #4686: [Task] Add the Create Room flow to the Space screen.

- The room should be automatically added to the space.
- The join rules picker should include the relevant choices given the space parent.

Velin92:
This PR expands on what was already built in the [previous PR](https://github.com/element-hq/element-x-ios/pull/4972) and essentially adds all the navigation code to create a room inside an existing space.
It also adds a button in the empty state of the space screen, to add an additional CTA for room creation.

fixes #4686

### DIFF
--- a/ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
@@ -177,7 +177,7 @@ class SpaceFlowCoordinator: FlowCoordinatorProtocol {
             rolesAndPermissionsFlowCoordinator?.clearRoute(animated: animated)
             clearRoute(animated: animated) // Re-run with the state machine back in the .space state.
         case .createChildRoomFlow:
-            createChildRoomFlowCoordinator?.clearRoute(animated: animated)
+            navigationStackCoordinator.setSheetCoordinator(nil)
             clearRoute(animated: animated) // Re-run with the state machine back in the .space state.
         }
     }
--- a/ElementX/Sources/FlowCoordinators/StartChatFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/StartChatFlowCoordinator.swift
@@ -59,9 +59,8 @@ class StartChatFlowCoordinator: FlowCoordinatorProtocol {
     enum Event: EventType {
         /// The flow is being started.
         case start
-        
         /// The user would like to create a room.
-        case createRoom
+        case createRoom(isSpace: Bool)
         /// The user dismissed the create room screen.
         case dismissedCreateRoom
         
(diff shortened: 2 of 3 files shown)


ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
```diff
@@ -534,4 +560,41 @@ class SpaceFlowCoordinator: FlowCoordinatorProtocol {
...
+            case .showRoomDirectory:
+                fatalError("Not implemented yet")
+            }
+        }
+        .store(in: &cancellables)
+        
+        navigationStackCoordinator.setSheetCoordinator(stackCoordinator) { [weak self] in
+            guard let self else { return }
+            stateMachine.tryEvent(.stopCreateChildRoomFlow)
+            switch flowCoordinatorResult {
```
stefanceriu, line 587: Why do we need to store the `flowCoordinatorResult` instead of treating it directly in `.finished`. 

Is it that we need the state machine to process a `.stopCreateChildRoomFlow` even first? If so, I would argue that the result should be a value associated to the event and handled together with `self?.createChildRoomFlowCoordinator = nil`.

ElementX/Sources/FlowCoordinators/StartChatFlowCoordinator.swift
```diff
@@ -126,16 +127,31 @@ class StartChatFlowCoordinator: FlowCoordinatorProtocol {
...
+            }
+        } handler: { [weak self] _ in
+            guard let self else { return }
+            switch entryPoint {
+            case .startChat:
+                presentStartChatScreen()
+            case .createSpace:
+                presentCreateRoomScreen(isSpace: true, spaceSelectionMode: nil, isRoot: true)
+            case .createRoomInSpace(let space):
+                presentCreateRoomScreen(isSpace: false, spaceSelectionMode: .preSelected(space), isRoot: true)
```
stefanceriu: What's the point of having a state machine if we're not going to use its states to power the UI? If not being able to place the `SpaceServiceRoomProtocol` in an event/state is the problem, we do work around that in other parts of the app, right?

Velin92: You're right. I'll move the handling at the start() level by giving an appropriate event, like we do in other flow coordinators

ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
Velin92: Not sure about this one, I would prefer the flow coordinator when dismisses the child flow coordinator to go back to the default space state first and then try another event to make transition, the reason is that if in the future we need to implement `previousState` (which I think might happen if I rework the `start` to use the entryPoint) for any of the other events, since start chat is a transitional flow presented modally, it would never be included in `previousState` and we would probably either work our way around this or just implement the way I did right now.
But if it won't be required right now after the `start` change, if you prefer I can implement the way you suggested, they would still work exactly in the same way

stefanceriu: No no that won't be necessary, it seems you have a plan 👍

ElementX/Sources/FlowCoordinators/StartChatFlowCoordinator.swift
stefanceriu: Nice, thank you for that!

PR #4985, opened 2026-01-22
Add the empty state to SpaceScreen.

### LINKED ISSUE #4689: [Task] Add an empty state for spaces where the user has admin access.

pixlwave:

- Rename `PaginationStatus.timelineEndReached` to `.endReached` so it can be used for more than just the timeline.
- Rename `PaginationState` to `TimelinePaginationState` and `PaginationStatus` to `PaginationState` (meaning `TimelinePaginationState` is composed of a forward and a backward `PaginationState`).
- Use the `PaginationState` in the `SpaceScreen` instead of `isPaginating`
- Add the [empty state](https://www.figma.com/design/kcnHxunG1LDWXsJhaNuiHz/ER-145--Spaces-on-Element-X?node-id=544-22315&m=dev) to `SpaceScreen`.
    - Needs #4984 for the second button.

Closes #4689.

Screenshot 2026-01-22 at 10 48 06 am

### DIFF
--- a/ElementX/Sources/Screens/Spaces/SpaceScreen/View/SpaceScreen.swift
+++ b/ElementX/Sources/Screens/Spaces/SpaceScreen/View/SpaceScreen.swift
@@ -22,7 +22,11 @@ struct SpaceScreen: View {
                                     mediaProvider: context.mediaProvider)
                 }
                 
-                rooms
+                if context.viewState.shouldShowEmptyState {
+                    emptyState
+                } else {
+                    rooms
+                }
             }
         }
         .environment(\.editMode, .constant(context.viewState.editMode))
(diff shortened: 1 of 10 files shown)


ElementX/Sources/Screens/Spaces/SpaceScreen/View/SpaceScreen.swift
```diff
@@ -53,12 +57,32 @@ struct SpaceScreen: View {
...
+                         iconStyle: .defaultSolid)
+                .padding(.horizontal, 24)
+            
+            VStack(spacing: 16) {
+                Button { context.send(viewAction: .addExistingRooms) } label: {
+                    Label(L10n.actionAddExistingRooms, icon: \.plus)
+                }
+                .buttonStyle(.compound(.primary))
+                
+                // @Velin92 Create Room button goes here.
```
Velin92, line 79: shouldn't this VStack check for the availability of the `m.space.child` permissions?

pixlwave:
> or maybe we want to fill it with something like "No available rooms yet"?

I guess we can add this if someone requests it. I would imagine this is going to be a pretty uncommon occurrence to worry about right now.

PR #4997, opened 2026-01-23
Make sure the selected/available rooms are updated when adding space children fails part way through.

pixlwave:
When you attempt to add multiple rooms to a space at once, the process can fail part way through. This PR makes sure that any rooms that were successfully added won't be included when the user (likely) presses the save button to try again.

### DIFF
--- a/ElementX/Sources/Screens/Spaces/SpaceScreen/SpaceScreenModels.swift
+++ b/ElementX/Sources/Screens/Spaces/SpaceScreen/SpaceScreenModels.swift
@@ -37,6 +37,7 @@ struct SpaceScreenViewState: BindableState {
     
     var editMode: EditMode = .inactive
     var editModeSelectedIDs: Set<String> = []
+    var editModeRemovedIDs: Set<String> = []
     
     var bindings = SpaceScreenViewStateBindings()
     
@@ -48,7 +49,7 @@ struct SpaceScreenViewState: BindableState {
         if editMode == .inactive {
             rooms
         } else {
-            rooms.filter { !$0.isSpace }
+            rooms.filter { !$0.isSpace && !editModeRemovedIDs.contains($0.id) }
         }
     }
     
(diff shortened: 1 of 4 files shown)


ElementX/Sources/Screens/Spaces/SpaceAddRoomsScreen/SpaceAddRoomsScreenViewModel.swift
```diff
@@ -89,15 +92,15 @@ class SpaceAddRoomsScreenViewModel: SpaceAddRoomsScreenViewModelType, SpaceAddRo
...
-        
-        state.roomsSection = .init(type: .searchResults, rooms: rooms)
+        let rooms = roomSummaryProvider.roomListPublisher.value
+            .lazy
+            .filter { [addedRoomIDs] summary in
+                !summary.isDirect && !existingRooms.contains { $0.id == summary.id } && !addedRoomIDs.contains(summary.id)
+            }
+            .map(SpaceAddRoomsScreenRoom.init)
         
-        // rooms crap
```
pixlwave, line 100: That was my "memo" to implement this but I completely missed it when committing 🙈

stefanceriu: Let's make a swiftlint rule for crap 😁

ElementX/Sources/Screens/Spaces/SpaceAddRoomsScreen/SpaceAddRoomsScreenViewModel.swift
```diff
@@ -131,15 +134,26 @@ class SpaceAddRoomsScreenViewModel: SpaceAddRoomsScreenViewModelType, SpaceAddRo
...
+            case .failure(let error):
                 MXLog.error("Failed adding room to space: \(error)")
                 showErrorIndicator()
-                updateRooms() // Hide any rooms that were already added.
+                
+                // Hide rooms that were successfully added.
+                state.selectedRooms.removeAll { addedRoomIDs.contains($0.id) }
+                updateRooms()
+                
                 return
```
Velin92, line 151: I would avoid removing from `selectedRooms` while you are iterating on them maybe would be better to copy them first, should not be an issue since they are structs but it's still a better practice to copy the array first, modify the copied array and then overwrite it, if for any reason the become objects.

Velin92: ob wait I just spotted the return, my bad! then it's totally okay

PR #4998, opened 2026-01-23
Update SDK to 26.01.23

Velin92:
This also includes a fixes:
- You could leave a room even when the last owner of a non empty room, because also previous owners that already left the room where still counted, now only joined owners are counted.
- You can now leave the room if you are the only member, before you couldn't
- Will also work for v12 rooms where the owner is a creator or someone with PL 150 and not a simple admin (however for now there is no UI differentiation, but will be handled in a future PR)
- Renamed the code to better reflect the intent and unified the last owner check and the members count in a simple `canLeave` variable

### DIFF
--- a/ElementX/Sources/Services/Spaces/LeaveSpaceHandleProxy.swift
+++ b/ElementX/Sources/Services/Spaces/LeaveSpaceHandleProxy.swift
@@ -13,12 +13,12 @@ final class LeaveSpaceHandleProxy {
     let id: String
     var rooms: [LeaveSpaceRoomDetails]
     
-    enum Mode { case manyRooms, onlyAdminRooms, noRooms, lastSpaceAdmin }
+    enum Mode { case manyRooms, roomsNeedNewOwner, noRooms, spaceNeedsNewOwner }
     let mode: Mode
     
     private let leaveHandle: LeaveSpaceHandleProtocol
     
-    var canLeave: Bool { mode != .lastSpaceAdmin }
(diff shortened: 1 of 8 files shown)


ElementX/Sources/Services/Spaces/LeaveSpaceHandleProxy.swift
```diff
@@ -87,12 +88,14 @@ final class LeaveSpaceHandleProxy {
...
-    let isLastAdmin: Bool
+    let canLeave: Bool
+    let areCreatorsPrivileged: Bool
     var isSelected: Bool
     
-    init(spaceServiceRoom: SpaceServiceRoomProtocol, isLastAdmin: Bool, isSelected: Bool) {
+    init(spaceServiceRoom: SpaceServiceRoomProtocol, isLastOwner: Bool, areCreatorsPrivileged: Bool, isSelected: Bool) {
         self.spaceServiceRoom = spaceServiceRoom
-        self.isLastAdmin = isLastAdmin
+        canLeave = !isLastOwner || spaceServiceRoom.joinedMembersCount == 1
```
pixlwave, line 97: Should we also call this `needsNewOwnerToLeave`?

Velin92: I thought about it, but I think is fine that internally we use `canLeave`, and `needsNewOwner` is more about what the UI presents when you `!canLeave` but if you prefer I can change it

PR #5000, opened 2026-01-23
Remove the protocol for SpaceServiceRoom given its a struct type.

pixlwave:
Final part of the tidy-up start in #4952 and #4955.

### DIFF
--- a/UnitTests/Sources/CreateRoomViewModelTests.swift
+++ b/UnitTests/Sources/CreateRoomViewModelTests.swift
@@ -272,7 +272,7 @@ class CreateRoomScreenViewModelTests: XCTestCase {
     }
     
     func testCreateRoomInAnAlreadySelectedSpace() async throws {
-        let space = SpaceServiceRoomMock(.init(isSpace: true, joinRule: .private))
+        let space = SpaceServiceRoom.mock(isSpace: true, joinRule: .private)
         setup(spacesSelectionMode: .preSelected(space))
         
         context.send(viewAction: .updateRoomName("A"))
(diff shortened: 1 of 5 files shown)


ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
```diff
@@ -301,7 +301,7 @@ class SpaceFlowCoordinator: FlowCoordinatorProtocol {
             guard event == .startCreateChildRoomFlow, case .space = fromState else { return nil }
             return .createChildRoomFlow
         } handler: { [weak self] context in
-            guard let space = context.userInfo as? SpaceServiceRoomProtocol else { fatalError("The space is missing") }
+            guard let space = context.userInfo as? SpaceServiceRoom else { fatalError("The space is missing") }
```
Velin92, line 304: not sure if you want to this in this PR or in another, but since this is a struct we could avoid sending SpaceServiceRoom as a userInfo and simply include it in the event

ElementX/Sources/FlowCoordinators/StartChatFlowCoordinator.swift
```diff
@@ -25,7 +25,7 @@ enum StartChatFlowCoordinatorAction {
 enum StartChatFlowCoordinatorEntryPoint {
     case startChat
     case createSpace
-    case createRoomInSpace(SpaceServiceRoomProtocol)
+    case createRoomInSpace(SpaceServiceRoom)
```
Velin92, line 28: Now that this entrypoint uses a struct maybe we can include back the .start(EntryPoint) event

ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
pixlwave: Ohhhh, can we make it hashable now, didn't think about that. Definitely a separate PR but sounds good!

ElementX/Sources/FlowCoordinators/StartChatFlowCoordinator.swift
pixlwave: Yep, I think we can come at all of these together in a follow-up. 👍

PR #5001, opened 2026-01-23
Horizontally scrollable code blocks

stefanceriu:
This PR makes code blocks horizontally scrollable within the timeline. 
Why? Because we always wanted this but timeline gesture conflicts were always getting in the way. Recently improved the gesture handling in https://github.com/element-hq/element-x-ios/commit/91b46ac and rewrote HTML parsing in https://github.com/element-hq/element-x-ios/pull/4449 which both make this easier.

It does so by:
* changing how attributed string components are created for timeline use and introducing a code block specific one
* adding a custom view to suppor rendering said component
* removing pre-baked code text background colors
* + extra points for making the code block copyable on long press

https://github.com/user-attachments/assets/b3b43c4d-899b-43b7-ba87-9b11046b1dd4

### DIFF
--- a/ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FormattedBodyText.swift
+++ b/ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FormattedBodyText.swift
@@ -81,50 +81,53 @@ struct FormattedBodyText: View {
                 // Ignore if the string contains only the layout correction
                 if String(component.attributedString.characters) == layoutDirection.isolateLayoutUnicodeString {
                     EmptyView()
-                } else if component.isBlockquote {
-                    // The rendered blockquote with a greedy width. The custom layout prevents the
-                    // infinite width from increasing the overall width of the view.
-                    MessageText(attributedString: component.attributedString.mergingAttributes(blockquoteAttributes))
-                        .fixedSize(horizontal: false, vertical: true)
-                        .frame(maxWidth: .infinity, alignment: .leading)
-                        .padding(.leading, 12.0)
-                        .overlay(alignment: .leading) {
-                            // User an overlay here so that the rectangle's infinite height doesn't take priority
-                            Capsule()
-                                .frame(width: 2.0)
-                                .padding(.leading, 5.0)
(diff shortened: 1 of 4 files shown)


pixlwave:
> extra points for making the code block copyable on long press

So my first thought based on the video was _"hmmm, maybe that'll make long pressing the message harder if it is only a codeblock"_, but then I thought _"ah but you could long press the timestamp still"_. Then I looked again and see that there isn't a timestamp 🙈

stefanceriu:
> Then I looked again and see that there isn't a timestamp 🙈

It's there, it's just a tad lower. 
Screenshot 2026-01-26 at 12 26 44

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FormattedBodyText.swift
```diff
@@ -81,34 +81,53 @@ struct FormattedBodyText: View {
...
+                            MessageText(attributedString: component.attributedString)
+                                .padding([.horizontal, .top], 4)
+                                .padding(.bottom, 8)
+                        }
+                        .background(.compound._bgCodeBlock)
+                        .scrollIndicatorsFlash(onAppear: true)
                         .fixedSize(horizontal: false, vertical: true)
+                        .padding(.horizontal, 4)
                         .layoutPriority(TimelineBubbleLayout.Priority.regularText)
+                        .contextMenu {
```
Velin92, line 113: this is nice, but couldn't it cause the tapping of the context menu to be overridden when there is only a code block? which happens quite a lot of times.

pixlwave:
> It's there, it's just a tad lower.

Ahhhh

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FormattedBodyText.swift
Velin92: Just saw @pixlwave comment, and I realised that the attributedString also includes the whitespaces to create the space to overlay the timestamp, so I imagine if there is only a code block, the scroll view might be either covering the overlay or including the whitespaces itself

PR #5009, opened 2026-01-26
Update the crypto strings.

pixlwave:
iOS PR for https://github.com/element-hq/element-meta/issues/3081


pixlwave:
Nope, this PR purely contains key changes, the snapshots should be fine.

PR #5016, opened 2026-01-27
Display an icon in the room header for rooms with shared history

### LINKED ISSUE #4987: History Sharing: room header decoration for `shared` and E2EE rooms
Per https://github.com/element-hq/element-meta/issues/3085, we'd like to add a decoration to the header for `shared` and E2EE rooms

kaylendog:
Closes #4987, part of our push for better history sharing UX.

The icon is intended to address any potential concern from users who come from WhatsApp, Signal and other consumer E2EE messaging apps where sharing history is not possible with new members, and who may also think of this as integral part of E2EE.
 
### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Other/SwiftUI/Views/RoomHeaderView.swift
+++ b/ElementX/Sources/Other/SwiftUI/Views/RoomHeaderView.swift
@@ -15,7 +15,7 @@ struct RoomHeaderView: View {
     var roomSubtitle: String?
     let roomAvatar: RoomAvatar
     var dmRecipientVerificationState: UserIdentityVerificationState?
-    var historySharingIconState: RoomScreenHistorySharingIconState?
+    var roomHistorySharingState: RoomHistorySharingState?
     
     let mediaProvider: MediaProviderProtocol?
     
@@ -65,15 +65,9 @@ struct RoomHeaderView: View {
                     VerificationBadge(verificationState: dmRecipientVerificationState, size: .xSmall, relativeTo: .compound.bodyMDSemibold)
                 }
                 
-                if let historySharingIconState {
-                    switch historySharingIconState {
-                    case .shared:
-                        CompoundIcon(\.history, size: .xSmall, relativeTo: .compound.bodyMDSemibold)
-                            .foregroundStyle(.compound.iconInfoPrimary)
-                    case .worldReadable:
-                        CompoundIcon(\.userProfileSolid, size: .xSmall, relativeTo: .compound.bodyMDSemibold)
-                            .foregroundStyle(.compound.iconInfoPrimary)
-                    }
+                if let historySharingIcon {
+                    CompoundIcon(historySharingIcon, size: .xSmall, relativeTo: .compound.bodyMDSemibold)
+                        .foregroundStyle(.compound.iconInfoPrimary)
                 }
             }
         }
@@ -84,6 +78,14 @@ struct RoomHeaderView: View {
                         avatarSize: .room(on: .timeline),
                         mediaProvider: mediaProvider)
             .accessibilityIdentifier(A11yIdentifiers.roomScreen.avatar)
+    }
+    
+    private var historySharingIcon: KeyPath<CompoundIcons, Image>? {
+        switch roomHistorySharingState {
+        case .shared: \.history
+        case .worldReadable: \.userProfileSolid
+        case .none: nil
+        }
     }
 }
 
@@ -118,10 +120,10 @@ struct RoomHeaderView_Previews: PreviewProvider, TestablePreview {
             makeHeader(avatarURL: .mockMXCAvatar,
                        roomSubtitle: "Subtitle",
                        verificationState: .verified)
-            makeHeader(avatarURL: .mockMXCAvatar, verificationState: .notVerified, historySharingIconState: .shared)
-            makeHeader(avatarURL: .mockMXCAvatar, verificationState: .notVerified, historySharingIconState: .worldReadable)
-            makeHeader(avatarURL: .mockMXCAvatar, verificationState: .verified, historySharingIconState: .shared)
-            makeHeader(avatarURL: .mockMXCAvatar, verificationState: .verificationViolation, historySharingIconState: .worldReadable)
+            makeHeader(avatarURL: .mockMXCAvatar, verificationState: .notVerified, historySharingState: .shared)
+            makeHeader(avatarURL: .mockMXCAvatar, verificationState: .notVerified, historySharingState: .worldReadable)
+            makeHeader(avatarURL: .mockMXCAvatar, verificationState: .verified, historySharingState: .shared)
+            makeHeader(avatarURL: .mockMXCAvatar, verificationState: .verificationViolation, historySharingState: .worldReadable)
         }
         .previewLayout(.sizeThatFits)
     }
@@ -129,14 +131,14 @@ struct RoomHeaderView_Previews: PreviewProvider, TestablePreview {
     static func makeHeader(avatarURL: URL?,
                            roomSubtitle: String? = nil,
                            verificationState: UserIdentityVerificationState,
-                           historySharingIconState: RoomScreenHistorySharingIconState? = nil) -> some View {
+                           historySharingState: RoomHistorySharingState? = nil) -> some View {
         RoomHeaderView(roomName: "Some Room name",
                        roomSubtitle: roomSubtitle,
                        roomAvatar: .room(id: "1",
                                          name: "Some Room Name",
                                          avatarURL: avatarURL),
                        dmRecipientVerificationState: verificationState,
-                       historySharingIconState: historySharingIconState,
+                       roomHistorySharingState: historySharingState,
                        
                        mediaProvider: MediaProviderMock(configuration: .init())) { }
             .padding()
--- a/ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
+++ b/ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
@@ -34,29 +34,6 @@ enum RoomScreenViewAction {
     case displaySuccessorRoom
 }
 
-enum RoomScreenHistorySharingIconState: Equatable {
-    case shared
-    case worldReadable
-    
-    init?(roomInfo: RoomInfoProxyProtocol) {
-        self.init(isEncrypted: roomInfo.isEncrypted, historyVisibility: roomInfo.historyVisibility)
-    }
-    
-    init?(isEncrypted: Bool, historyVisibility: RoomHistoryVisibility) {
-        if !isEncrypted {
-            return nil
-        }
-        switch historyVisibility {
-        case .shared:
-            self = .shared
-        case .worldReadable:
-            self = .worldReadable
-        default:
-            return nil
-        }
-    }
-}
-
 struct RoomScreenViewState: BindableState {
     var roomTitle = ""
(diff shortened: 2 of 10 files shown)


ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -70,7 +70,8 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
         let viewState = RoomScreenViewState(roomTitle: roomProxy.infoPublisher.value.displayName ?? roomProxy.id,
                                             roomAvatar: roomProxy.infoPublisher.value.avatar,
                                             hasOngoingCall: roomProxy.infoPublisher.value.hasRoomCall,
-                                            hasSuccessor: roomProxy.infoPublisher.value.successor != nil)
+                                            hasSuccessor: roomProxy.infoPublisher.value.successor != nil,
+                                            isRoomHistoryShared: appSettings.enableKeyShareOnInvite && (roomProxy.infoPublisher.value.historyVisibility == RoomHistoryVisibility.shared || roomProxy.infoPublisher.value.historyVisibility == RoomHistoryVisibility.worldReadable))
```
pixlwave: We can infer the type here:

```suggestion
                                            isRoomHistoryShared: appSettings.enableKeyShareOnInvite && (roomProxy.infoPublisher.value.historyVisibility == .shared || roomProxy.infoPublisher.value.historyVisibility == .worldReadable))
```

ElementX/Sources/Screens/RoomScreen/RoomScreenViewModel.swift
```diff
@@ -342,6 +343,8 @@ class RoomScreenViewModel: RoomScreenViewModelType, RoomScreenViewModelProtocol
             state.canDeclineKnocks = powerLevels.canOwnUserKick()
             state.canBan = powerLevels.canOwnUserBan()
         }
+        
+        state.isRoomHistoryShared = appSettings.enableKeyShareOnInvite && (roomInfo.historyVisibility == RoomHistoryVisibility.shared || roomInfo.historyVisibility == RoomHistoryVisibility.worldReadable)
```
pixlwave: And here:

```suggestion
        state.isRoomHistoryShared = appSettings.enableKeyShareOnInvite && (roomInfo.historyVisibility == .shared || roomInfo.historyVisibility == .worldReadable)
```

ElementX/Sources/Screens/ThreadTimelineScreen/View/ThreadTimelineScreen.swift
```diff
@@ -65,6 +65,7 @@ struct ThreadTimelineScreen: View {
                            roomSubtitle: context.viewState.roomTitle,
                            roomAvatar: context.viewState.roomAvatar,
                            dmRecipientVerificationState: context.viewState.dmRecipientVerificationState,
+                           isRoomHistoryShared: false,
```
pixlwave: Out of interest, have we made a conscious decision to not show the badge on threads (given it is still a timeline within a room) or is this just fixing the compilation error? Might be worth a small comment to note that it's intentional if so.

UnitTests/Sources/RoomScreenViewModelTests.swift
```diff
@@ -438,4 +438,46 @@ class RoomScreenViewModelTests: XCTestCase {
...
+        }
+        try await deferredVisible.fulfill()
+    }
+}
+
+extension RoomInfoProxyMock {
+    @MainActor convenience init(configuration: JoinedRoomProxyMockConfiguration, historyVisibility: RoomHistoryVisibility) {
+        self.init(configuration)
+        self.historyVisibility = historyVisibility
+    }
```
pixlwave: We should probably follow the standard pattern here and remove this convenience init in favour of adding `historyVisibility` directly to `JoinedRoomProxyMockConfiguration` and using that in `RoomInfoProxyMock.init(_:)`.

UnitTests/Sources/RoomScreenViewModelTests.swift
```diff
@@ -438,4 +438,46 @@ class RoomScreenViewModelTests: XCTestCase {
...
+                                            initialSelectedPinnedEventID: nil,
+                                            ongoingCallRoomIDPublisher: .init(.init(nil)),
+                                            appSettings: ServiceLocator.shared.settings,
+                                            appHooks: AppHooks(),
+                                            analyticsService: ServiceLocator.shared.analytics,
+                                            userIndicatorController: ServiceLocator.shared.userIndicatorController)
+        self.viewModel = viewModel
+        
+        // the icon should not be displayed when the room history visibility is not `shared` or `worldReadable` ...
+        let deferredInvisible = deferFailure(viewModel.context.$viewState, timeout: 1) { viewState in
```
pixlwave: Entirely up to you, but just so that you know. If you want these messages to show in the results as a test failure you can do this:

```suggestion
        let deferredInvisible = deferFailure(viewModel.context.$viewState, 
                                             timeout: 1,
                                             message: "The icon should be hidden when the room history visibility is not .shared or .worldReadable") { viewState in
```

UnitTests/Sources/RoomScreenViewModelTests.swift
```diff
@@ -438,4 +438,46 @@ class RoomScreenViewModelTests: XCTestCase {
...
+            viewState.isRoomHistoryShared
+        }
+        try await deferredInvisible.fulfill()
+        
+        // ... but should be displayed when it is
+        infoSubject.send(RoomInfoProxyMock(configuration: configuration, historyVisibility: .shared))
+        let deferredVisible = deferFulfillment(viewModel.context.$viewState) { viewState in
```
pixlwave:
```suggestion
        infoSubject.send(RoomInfoProxyMock(configuration: configuration, historyVisibility: .shared))
        let deferredVisible = deferFulfillment(viewModel.context.$viewState,
                                               message: "The icon should be shown when the room history visibility is .shared") { viewState in
```

ElementX/Sources/Other/SwiftUI/Views/RoomHeaderView.swift
pixlwave, line 1: Given there isn't a Figma showing this, it seems worth double checking that we're happy to show verification badges alongside history badges rather than some algo that showed only a single badge in an order of e.g. most to least critical such as:
`.verificationViolation` otherwise `isRoomHistoryShared` otherwise `.verified`

kaylendog: CC @mxandreas

ElementX/Sources/Screens/ThreadTimelineScreen/View/ThreadTimelineScreen.swift
kaylendog: CC @mxandreas

ElementX/Sources/Other/SwiftUI/Views/RoomHeaderView.swift
kaylendog: I believe the current implementation is correct! The history sharing icon should be displayed to the right of any verification badges.

pixlwave: Nice, thanks for confirming 👍

ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
```diff
@@ -33,6 +34,29 @@ enum RoomScreenViewAction {
...
+        case .shared:
+            self = .shared
+        case .worldReadable:
+            self = .worldReadable
+        default:
+            return nil
+        }
+    }
+}
+
```
kaylendog: This feels a little messy, and I'm not sure if it's in the right place.

kaylendog:
@pixlwave changed a few things, think it's worth you having another look.

ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
pixlwave: Right, I think I would probably make the following changes:
- Rename the enum to `RoomHistorySharingState` and put it in `Sources/Services/Room`.
- Combine the contents of the two initialisers into the existing `extension RoomInfoProxyProtocol` as a single computed property.
- Possibly move the documentation from the property below onto the enum itself.

ElementX/Sources/Other/SwiftUI/Views/RoomHeaderView.swift
```diff
@@ -63,6 +64,17 @@ struct RoomHeaderView: View {
...
+                if let historySharingIconState {
+                    switch historySharingIconState {
+                    case .shared:
+                        CompoundIcon(\.history, size: .xSmall, relativeTo: .compound.bodyMDSemibold)
+                            .foregroundStyle(.compound.iconInfoPrimary)
+                    case .worldReadable:
+                        CompoundIcon(\.userProfileSolid, size: .xSmall, relativeTo: .compound.bodyMDSemibold)
+                            .foregroundStyle(.compound.iconInfoPrimary)
+                    }
+                }
```
pixlwave: At this stage you can
```suggestion
                    switch historySharingIconState {
                    case .shared:
                        CompoundIcon(\.history, size: .xSmall, relativeTo: .compound.bodyMDSemibold)
                            .foregroundStyle(.compound.iconInfoPrimary)
                    case .worldReadable:
                        CompoundIcon(\.userProfileSolid, size: .xSmall, relativeTo: .compound.bodyMDSemibold)
                            .foregroundStyle(.compound.iconInfoPrimary)
                    case .none:
                        EmptyView()
                    }
                }
```

or a solution with less repetition could be

```swift
// outside of the body
private var historySharingIcon: KeyPath<CompoundIcons, Image>? {
    switch historySharingIconState {
    case .shared: \.history
    case .worldReadable: \.userProfileSolid
    case .none: nil
    }
}

// in the body
if let historySharingIcon {
    CompoundIcon(historySharingIcon, size: .xSmall, relativeTo: .compound.bodyMDSemibold)
        .foregroundStyle(.compound.iconInfoPrimary)
}
```

UnitTests/Sources/RoomScreenViewModelTests.swift
```diff
@@ -441,12 +441,48 @@ class RoomScreenViewModelTests: XCTestCase {
...
+                                            appSettings: ServiceLocator.shared.settings,
+                                            appHooks: AppHooks(),
+                                            analyticsService: ServiceLocator.shared.analytics,
+                                            userIndicatorController: ServiceLocator.shared.userIndicatorController)
+        self.viewModel = viewModel
+        
+        let deferredInvisible = deferFailure(viewModel.context.$viewState,
+                                             timeout: 1,
+                                             message: "The icon should not be shown when the room history visibility is not .shared or .worldReadable") { viewState in
+            viewState.historySharingIconState == .shared || viewState.historySharingIconState == .worldReadable
```
pixlwave: I think these could all be simplified to `!= nil` rather than using the `||`?

ElementX/Sources/Other/SwiftUI/Views/RoomHeaderView.swift
kaylendog: Smart! I'll go with the latter.

kaylendog:
@pixlwave one last pass, if you don't mind!

PR #5017, opened 2026-01-27
Fix inline code being rendered as blocks.

pixlwave:
Fixes a regression introduced by #5001. Also includes the following changes:
- Add a non-greedy version of code blocks to the bubble layout so that they don't make the bubble greedy within the timeline.
- Disable horizontal scrolling on the code blocks when they fit within the bubble.
- Splits up the FormattedBodyText snapshots into individual cases so the diffs are easier to see.

| Before | After |
| - | - |
| Screenshot 2026-01-27 at 5 00 41 pm | Screenshot 2026-01-27 at 5 01 59 pm |

Probably easiest to review commit-by-commit but not strictly necessary.

### DIFF
--- a/ElementX/Sources/Other/HTMLParsing/AttributedStringBuilder.swift
+++ b/ElementX/Sources/Other/HTMLParsing/AttributedStringBuilder.swift
@@ -211,7 +211,7 @@ struct AttributedStringBuilder: AttributedStringBuilderProtocol {
                     // The scroll view provides the background colour for code blocks.
                 } else {
                     content.addAttribute(.InlineCode, value: true, range: NSRange(location: 0, length: content.length))
-                    // But inline code is (obviously) inline so we need the background colour here.
+                    // But inline code is (obviously) inline so it's much easier to set the background colour here.
                     content.addAttribute(.backgroundColor, value: UIColor.compound._bgCodeBlock as Any, range: NSRange(location: 0, length: content.length))
                 }
                 


ElementX/Sources/Other/HTMLParsing/AttributedStringBuilder.swift
```diff
@@ -197,15 +198,24 @@ struct AttributedStringBuilder: AttributedStringBuilderProtocol {
...
                 content.setFontPreservingSymbolicTraits(UIFont.monospacedSystemFont(ofSize: fontPointSize, weight: .regular))
                 
-                content.addAttribute(.CodeBlock, value: true, range: NSRange(location: 0, length: content.length))
+                if isCodeBlock {
+                    content.addAttribute(.CodeBlock, value: true, range: NSRange(location: 0, length: content.length))
+                    // The scroll view provides the background colour for code blocks.
+                } else {
+                    content.addAttribute(.InlineCode, value: true, range: NSRange(location: 0, length: content.length))
+                    // But inline code is (obviously) inline so we need the background colour here.
+                    content.addAttribute(.backgroundColor, value: UIColor.compound._bgCodeBlock as Any, range: NSRange(location: 0, length: content.length))
```
stefanceriu, line 215: Techincally you can do this on the UI level based on the attribute.

PreviewTests/Sources/__Snapshots__/PreviewTests/formattedBodyText.code-iPhone-pseudo.png
stefanceriu, line 1: Noice 👍

ElementX/Sources/Other/HTMLParsing/AttributedStringBuilder.swift
pixlwave: Updated the comment 😅

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FormattedBodyText.swift
```diff
@@ -147,51 +156,101 @@ struct FormattedBodyText: View {
...
+            ScrollView(.horizontal) {
+                MessageText(attributedString: attributedString)
+                    .padding([.horizontal, .top], 4)
+                    .padding(.bottom, 8)
+                    .onGeometryChange(for: CGSize.self) { $0.size } action: { maxSize = $0 }
+            }
+            .frame(maxWidth: maxSize.width)
+            .padding(.horizontal, 4)
+        }
+    }
```
pixlwave, line 178: Note: I'm going to come back to this after the RC and tidy the whole thing up a bit to make a single view for both code blocks and quote bubbles with a parameter to whether it's for layout or rendering but doesn't seem worth delaying the RC for that right now.

PR #5018, opened 2026-01-27
Transfer v12 space ownership when last owner

### LINKED ISSUE #5014: [Task] Handle leaving a v12 space when you are the last owner of a non empty space

Velin92:
fixes #5014

This PR handles the transfer ownership flow for v12 spaces when last owner:
- New leave space view state when transferring the ownership is required
- Removed the (admin) string from the rooms since the information might have been confusing in some cases
- Handled the transfer ownership flow in space and space settings
- Update toolbar buttons in change roles screen and added the cancel button ONLY when in transferring ownership state
- Refactored all the flows using the transfer ownership screen to use a state and presentation and dismissal events properly.

### DIFF
--- a/ElementX/Sources/FlowCoordinators/ChatsTabFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/ChatsTabFlowCoordinator.swift
@@ -478,6 +478,7 @@ class ChatsTabFlowCoordinator: FlowCoordinatorProtocol {
         guard case let .joined(roomProxy) = await userSession.clientProxy.roomForIdentifier(roomID) else {
             return
         }
+        // The screen requires listening to the membership state changes
         await roomProxy.subscribeForUpdates()
         
         let parameters = RoomChangeRolesScreenCoordinatorParameters(mode: .owner,
--- a/ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
@@ -55,6 +55,8 @@ class SpaceFlowCoordinator: FlowCoordinatorProtocol {
         case space
         /// The user is adding rooms to the space.
         case addingRooms
+        /// The user is transferring their ownership of the space.
+        case transferOwnership
         /// A child (space) flow is in progress.
         case presentingChild(childSpaceID: String, previousState: State)
         /// A room flow is in progress
@@ -69,11 +71,9 @@ class SpaceFlowCoordinator: FlowCoordinatorProtocol {
         case createChildRoomFlow
         
         case leftSpace
-        
-        case transferOwnership
     }
     
-    enum Event: EventType, Equatable, Hashable {
+    enum Event: EventType {
         /// The flow is being started.
         case start
         /// The flow is being started for an unjoined space.
@@ -88,6 +88,11 @@ class SpaceFlowCoordinator: FlowCoordinatorProtocol {
         case addRooms
         /// The user finished adding rooms to this space.
         case dismissedAddRooms
+        
+        /// Allow the user to transfer their ownership of the space.
+        case presentTransferOwnership
+        /// The user finished transferring their ownership of the space.
+        case dismissedTransferOwnership
         
         /// Request the presentation of a child space flow.
         ///
@@ -110,9 +115,6 @@ class SpaceFlowCoordinator: FlowCoordinatorProtocol {
         
         case startCreateChildRoomFlow
         case stopCreateChildRoomFlow
-        
-        case presentTransferOwnership
-        case dismissedTransferOwnership
     }
     
     private let stateMachine: StateMachine<State, Event>
@@ -166,6 +168,9 @@ class SpaceFlowCoordinator: FlowCoordinatorProtocol {
         case .addingRooms:
             navigationStackCoordinator.setSheetCoordinator(nil)
             clearRoute(animated: animated) // Re-run with the state machine back in the .space state.
+        case .transferOwnership:
+            navigationStackCoordinator.setSheetCoordinator(nil)
+            clearRoute(animated: animated) // Re-run with the state machine back in the .space state.
         case .presentingChild:
             childSpaceFlowCoordinator?.clearRoute(animated: animated)
             clearRoute(animated: animated) // Re-run with the state machine back in the .space state.
@@ -182,9 +187,6 @@ class SpaceFlowCoordinator: FlowCoordinatorProtocol {
             rolesAndPermissionsFlowCoordinator?.clearRoute(animated: animated)
             clearRoute(animated: animated) // Re-run with the state machine back in the .space state.
         case .createChildRoomFlow:
-            navigationStackCoordinator.setSheetCoordinator(nil)
-            clearRoute(animated: animated) // Re-run with the state machine back in the .space state.
-        case .transferOwnership:
             navigationStackCoordinator.setSheetCoordinator(nil)
             clearRoute(animated: animated) // Re-run with the state machine back in the .space state.
         }
--- a/ElementX/Sources/FlowCoordinators/SpaceSettingsFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/SpaceSettingsFlowCoordinator.swift
@@ -27,16 +27,16 @@ final class SpaceSettingsFlowCoordinator: FlowCoordinatorProtocol {
         case securityAndPrivacy
         /// The edit address screen
         case editAddress
+        /// The manage authorized spaces screen
+        case manageAuthorizedSpacesScreen
+        /// The screen that allows a user to transfer their ownership of the space.
+        case transferOwnership
         
         /// Other flows
         /// The roles and permissions screen
         case rolesAndPermissionsFlow
         /// The members flow screen
         case membersFlow
-        
-        case manageAuthorizedSpacesScreen
-
-        case transferOwnership
     }
     
     enum Event: EventType {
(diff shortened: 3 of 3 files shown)


ElementX/Sources/FlowCoordinators/SpaceSettingsFlowCoordinator.swift
```diff
@@ -326,6 +327,27 @@ final class SpaceSettingsFlowCoordinator: FlowCoordinatorProtocol {
...
+            switch action {
+            case .complete:
+                navigationStackCoordinator.setSheetCoordinator(nil)
+            }
+        }
+        .store(in: &cancellables)
+        
+        stackCoordinator.setRootCoordinator(coordinator)
+        navigationStackCoordinator.setSheetCoordinator(stackCoordinator, animated: true)
+    }
```
Velin92, line 365: Okay, so for this one and the other one in the `SpaceFlowCoordinator` I took it easy for now and just copy and pasted what we use in the `RoomFlowCoordinator,` which is just presenting this screen as a modal without even managing the state of the flow coordinator.

I did this for a quick test, but I feel we should add now a state for this in both the `RoomFlowCoordinator`, the `SpaceSettingsFlowCoordinator `and the `SpaceFlowCoordinator`.

The only doubt I am having is... how we want to handle this presentation, there 2 possibilities:
- As a special entry point for the `RoomRolesAndPermissionsFlowCoordinator`
- As a single use of the `RoomChangeRolesScreen` in each of these flow cooerdinators.

My concern for the first implementation is that modifying the whole flow coordinator and creating a flow coordinator just for one single screen modal (since the flow doesn't lead anywhere else) feels a bit of an overkill, so maybe we could just do as it right now but add a new events and a state, for each of the flow coordinators that are using this screen?

ElementX/Sources/Screens/RoomChangeRolesScreen/View/RoomChangeRolesScreen.swift
```diff
@@ -93,15 +93,15 @@ struct RoomChangeRolesScreen: View {
...
         ToolbarItem(placement: .confirmationAction) {
-            Button(L10n.actionSave) {
+            ToolbarButton(role: .save) {
                 context.send(viewAction: .save)
             }
             .disabled(!context.viewState.hasChanges)
         }
         
-        if context.viewState.hasChanges {
+        if context.viewState.mode == .owner {
```
pixlwave: This should be an `||` here – we should still replace the back button with a cancel button where there are unsaved changes.
```suggestion
        if context.viewState.hasChanges || context.viewState.mode == .owner {
```

ElementX/Sources/Screens/Spaces/LeaveSpace/View/LeaveSpaceRoomDetailsCell.swift
pixlwave, line 1: This seems weird to me. We don't let them select the row, but there's now zero explanation why that's the case?

ElementX/Sources/Screens/Spaces/SpaceScreen/SpaceScreenCoordinator.swift
```diff
@@ -78,6 +79,8 @@ final class SpaceScreenCoordinator: CoordinatorProtocol {
                 actionsSubject.send(.addExistingChildren)
             case .displayCreateChildRoomFlow(let space):
                 actionsSubject.send(.displayCreateChildRoomFlow(space: space))
+            case .presentTransferOwnership(roomProxy: let roomProxy):
```
pixlwave:
```suggestion
            case .presentTransferOwnership(let roomProxy):
```

ElementX/Sources/Screens/Spaces/SpaceScreen/SpaceScreenViewModel.swift
```diff
@@ -261,6 +261,12 @@ class SpaceScreenViewModel: SpaceScreenViewModelType, SpaceScreenViewModelProtoc
             case .didLeaveSpace:
                 state.bindings.leaveSpaceViewModel = nil
                 actionsSubject.send(.leftSpace)
+            case .presentTransferOwnership:
+                guard let roomProxy = state.roomProxy else {
+                    fatalError("The space screen should always have a room proxy")
```
pixlwave: I think its fine for the screen to not have one, maybe something like:
```suggestion
                    fatalError("There should always be a room proxy available for joined spaces.")
```

PreviewTests/Sources/__Snapshots__/PreviewTests/leaveSpaceView.Only-Admin-Rooms-iPhone-en-GB.png
pixlwave, line 1: Are the cells configured correctly for this snapshot, shouldn't the selection be hidden?

ElementX/Sources/FlowCoordinators/SpaceSettingsFlowCoordinator.swift
pixlwave: > I did this for a quick test, but I feel we should add now a state for this in both the `RoomFlowCoordinator`, the `SpaceSettingsFlowCoordinator `and the `SpaceFlowCoordinator`.

Agreed 👍



> The only doubt I am having is... how we want to handle this presentation, there 2 possibilities:
> 
> * As a special entry point for the `RoomRolesAndPermissionsFlowCoordinator`
> * As a single use of the `RoomChangeRolesScreen` in each of these flow cooerdinators.

I would agree, a single use of the screens is perfectly acceptable given it doesn't do anything else. From my perspective the flow coordinator would only be useful if this screen could go on to present more things which it can't.

pixlwave (approved):
Nice, looks good. 👌

ElementX/Sources/Screens/Spaces/LeaveSpace/View/LeaveSpaceRoomDetailsCell.swift
Velin92: The idea is that if you can't select you are either an admin or an owner but we don't want to get too technical

pixlwave: Sure, but how does the user know that if we don't say *something*?

Velin92: Screenshot 2026-01-28 at 12 56 30

I think it looks clear, if you can't select it's because you are the only admin/owner for it the subtitle explains it.

PreviewTests/Sources/__Snapshots__/PreviewTests/leaveSpaceView.Only-Admin-Rooms-iPhone-en-GB.png
Velin92: whoops was passing the entire handle `canLeave` instead of the room `canLeave` here, good catch

ElementX/Sources/Screens/Spaces/LeaveSpace/View/LeaveSpaceRoomDetailsCell.swift
pixlwave: Fair 👍

ElementX/Sources/FlowCoordinators/ChatsTabFlowCoordinator.swift
```diff
@@ -469,6 +474,34 @@ class ChatsTabFlowCoordinator: FlowCoordinatorProtocol {
         }
     }
     
+    private func presentTransferOwnershipScreen(roomID: String) async {
+        guard case let .joined(roomProxy) = await userSession.clientProxy.roomForIdentifier(roomID) else {
+            return
+        }
+        await roomProxy.subscribeForUpdates()
```
pixlwave, line 482: Should we add a comment about what the subscription is needed for?

ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
```diff
@@ -69,9 +69,11 @@ class SpaceFlowCoordinator: FlowCoordinatorProtocol {
         case createChildRoomFlow
         
         case leftSpace
+        
+        case transferOwnership
     }
     
-    enum Event: EventType {
+    enum Event: EventType, Equatable, Hashable {
```
pixlwave: Errrr, `EventType` is already `: Hashable` and `Hashable` is `: Equatable` so I don't think we need either here…

ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
```diff
@@ -69,9 +69,11 @@ class SpaceFlowCoordinator: FlowCoordinatorProtocol {
         case createChildRoomFlow
         
         case leftSpace
+        
+        case transferOwnership
```
pixlwave: Not mega important but I've been trying to keep these somewhat organised. Could we put this before the child flows so we have some kind of grouping of `initial`, `screens`, `flows`, `final` states?

ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
```diff
@@ -108,6 +110,9 @@ class SpaceFlowCoordinator: FlowCoordinatorProtocol {
         
         case startCreateChildRoomFlow
         case stopCreateChildRoomFlow
+        
+        case presentTransferOwnership
+        case dismissedTransferOwnership
```
pixlwave: And then same here, so the order of these matches the state change

ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
```diff
@@ -179,6 +184,9 @@ class SpaceFlowCoordinator: FlowCoordinatorProtocol {
         case .createChildRoomFlow:
             navigationStackCoordinator.setSheetCoordinator(nil)
             clearRoute(animated: animated) // Re-run with the state machine back in the .space state.
+        case .transferOwnership:
+            navigationStackCoordinator.setSheetCoordinator(nil)
+            clearRoute(animated: animated) // Re-run with the state machine back in the .space state.
```
pixlwave: And again here :)

ElementX/Sources/FlowCoordinators/SpaceSettingsFlowCoordinator.swift
```diff
@@ -35,6 +35,8 @@ final class SpaceSettingsFlowCoordinator: FlowCoordinatorProtocol {
         case membersFlow
         
         case manageAuthorizedSpacesScreen
+
+        case transferOwnership
```
pixlwave: Came comments on this screen too :)

PR #5020, opened 2026-01-29
Signposter and sentry metrics update

stefanceriu:
This set of commits refactor how we gather sentry metrics

As per https://github.com/element-hq/element-meta/pull/3055


ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -82,16 +88,19 @@ final class AppSettings {
...
+        case "io.element.elementx.nightly":
+            return .nightly
+        default:
+            return .release
+        }
         #endif
-    }()
-        
+    }
+
```
Velin92: extra space here

Velin92 (approved):

PR #5022, opened 2026-01-29
Fix a bug where the app wouldn't retry when the server rate-limits its requests.

pixlwave:
… by changing the retry limit to 3. We'll see how this goes in Nightlies as [other `5xx` errors](https://github.com/matrix-org/matrix-rust-sdk/blob/8f4ac73ca9b627627fa164d815430a13cb7e215f/crates/matrix-sdk/src/error.rs#L251) are also re-tried, so we may need to report these to the SDK team if they don't make sense.

Related to https://github.com/element-hq/element-meta/issues/3017 as this was seen when adding/removing children to/from a space.


manuroe:
@bmarty this is the change we want to try on EXA too.

PR #5024, opened 2026-01-30
Add a search bar to the space filters screen

stefanceriu:
because it's the right thing to do.


ElementX/Sources/Screens/ChatsSpaceFiltersScreen/ChatsSpaceFiltersScreenModels.swift
```diff
@@ -15,9 +15,22 @@ enum ChatsSpaceFiltersScreenViewModelAction {
...
+    
+    var visibleFilters: [SpaceServiceFilter] {
+        if bindings.searchQuery.isEmpty {
+            return filters
+        }
+        
+        return filters.filter { filter in
+            filter.room.name.localizedStandardContains(bindings.searchQuery) ||
+                (filter.room.canonicalAlias ?? "").localizedStandardContains(bindings.searchQuery)
+        }
```
Velin92, line 27: I know this is a quick pr to merge, but would rename this to visibleItem or visibleSpace, filter seems more like we are actuallt search among a list of filters.

PR #5026, opened 2026-01-30
Fix call service completion handler calls.

stefanceriu:
Avoid `Killing app because it never posted an incoming call to the system after receiving a PushKit VoIP push` errors.

Making this a draft as I can't tell if the async version of the API is broken or there's something else that prevents it from workign on my dev environment.


pixlwave:
> Making this a draft as I can't tell if the async version of the API is broken or there's something else that prevents it from workign on my dev environment.

I've just checked twice on my side. Using this PR I get no call ringing. Reverting back to 26.02.0 and calls are able to ring again.

pixlwave (approved):
LGTM. Might also be worthwhile to keep the async version under a different name and call that from the working delegate method in a `withCheckedContinuation` but not super important.

PR #5035, opened 2026-02-02
Fix notification delivery after powering on a device.

pixlwave:
The check for a writable app group container was happening *after* the NSE attempted to configure logs resulting in the NSE hitting [this fatalError](https://github.com/element-hq/element-x-ios/blob/35daa017bf6065dff00e3bdd9c4f0f395f3f911e/ElementX/Sources/Application/TargetConfiguration.swift#L54) when the Rust call failed. This PR performs the check right at the start of the NSE, and then uses the lack of a tracing configuration to indicate that the device is still locked and so the notification should not be processed.


stefanceriu (approved):
Perfect, thank you! 👍

PR #5036, opened 2026-02-02
Add a stale bot for X-Needs-Info issues.

pixlwave:
This should only act upon issues with the `X-Needs-Info` label. It will also automatically remove the label when there's a response.

Before merging we need to double check all existing issues and remove the label from any that do not need it any more.

### DIFF
--- a/.github/workflows/stale-issues.yml
+++ b/.github/workflows/stale-issues.yml
@@ -10,7 +10,7 @@ jobs:
     permissions:
       issues: write
     steps:
-      - uses: actions/stale@997185467fa4f803885201cee163a9f38240193d # v10.1.1
+      - uses: actions/stale@v10
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/stale-issues.yml
```diff
@@ -0,0 +1,22 @@
...
+  schedule:
+    - cron: "30 1 * * *"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - uses: actions/stale@997185467fa4f803885201cee163a9f38240193d # v10.1.1
```
stefanceriu: We don't need to bother pinning first party actions.

pixlwave: Oh yeah, totally forgot that 👍

PR #5043, opened 2026-02-03
Add history sharing badges to room details

kaylendog:
Closes https://github.com/element-hq/element-x-ios/pull/5043

### Pull Request Checklist

- [x] Voiceover enabled.

### DIFF
--- a/PreviewTests/Sources/__Snapshots__/PreviewTests/badgeLabel.iPad-en-GB-0.png
+++ b/PreviewTests/Sources/__Snapshots__/PreviewTests/badgeLabel.iPad-en-GB-0.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:6f251951b19c194c99b13a5cc258a7a3346dfc69ff054eb2d6788d7faff823ac
-size 79171
+oid sha256:2d442ec46d6e37256ef69826f71040682b91feb7d6c00c95936b62df52eed10b
+size 97026
--- a/PreviewTests/Sources/__Snapshots__/PreviewTests/badgeLabel.iPad-pseudo-0.png
+++ b/PreviewTests/Sources/__Snapshots__/PreviewTests/badgeLabel.iPad-pseudo-0.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:6f251951b19c194c99b13a5cc258a7a3346dfc69ff054eb2d6788d7faff823ac
-size 79171
+oid sha256:2d442ec46d6e37256ef69826f71040682b91feb7d6c00c95936b62df52eed10b
+size 97026
--- a/PreviewTests/Sources/__Snapshots__/PreviewTests/badgeLabel.iPhone-en-GB-0.png
+++ b/PreviewTests/Sources/__Snapshots__/PreviewTests/badgeLabel.iPhone-en-GB-0.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:52933a15fe6e8e66fb981f67f5b10622b6010e987f68e28c0db0fa077736b2b8
-size 37557
+oid sha256:8017704510d0461df8592b5e321ae5c4917ec8a9c225f917de1cdc8892ea904b
+size 58270
--- a/PreviewTests/Sources/__Snapshots__/PreviewTests/badgeLabel.iPhone-pseudo-0.png
+++ b/PreviewTests/Sources/__Snapshots__/PreviewTests/badgeLabel.iPhone-pseudo-0.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:52933a15fe6e8e66fb981f67f5b10622b6010e987f68e28c0db0fa077736b2b8
-size 37557
+oid sha256:8017704510d0461df8592b5e321ae5c4917ec8a9c225f917de1cdc8892ea904b
+size 58270


ElementX/Sources/Screens/RoomDetailsScreen/View/RoomDetailsScreen.swift
```diff
@@ -355,6 +359,15 @@ struct RoomDetailsScreen_Previews: PreviewProvider, TestablePreview {
         RoomDetailsScreen(context: dmRoomVerificationViolationViewModel.context)
             .snapshotPreferences(expect: dmRoomVerificationViolationViewModel.context.observe(\.viewState.accountOwner).map { $0 != nil })
             .previewDisplayName("DM Room Verification Violation")
```
kaylendog: I'm encountering an annoying issue here I'm not sure how to approach. The history sharing tests need `appSettings.enableKeyShareOnInvite = true` for the pills to appear, but since this is done in `makeHistorySharingViewModel` which is called before we create these views, these all render with the badges too.

pixlwave: In practice the snapshots aren't really the best way to validate whether or not the view model's feature flag check is sound – that's a job for unit tests. I would suggest enabling the flag on all the tests and configuring the room info for the existing snapshots in a way that makes most sense for each one (and if all the new cases are covered by this, then you can remove those new snapshots).

ElementX/Resources/Localizations/en.lproj/Untranslated.strings
pixlwave, line 1: Presumable you’re going to add these to Localazy?

ElementX/Sources/Mocks/JoinedRoomProxyMock.swift
pixlwave, line 1: I think you can revert this :)

ElementX/Sources/Other/SwiftUI/Views/AvatarHeaderView.swift
```diff
@@ -223,6 +243,31 @@ struct AvatarHeaderView<Footer: View>: View {
 }
 
 struct AvatarHeaderView_Previews: PreviewProvider, TestablePreview {
+    private static func makeHistorySharingPreview(state: RoomHistorySharingState) -> some View {
```
pixlwave: Super minor, but please could we put this beneath `var previews` – if it was just a `static var` it would go up here but for functions we usually put them beneath the place they are called from.

ElementX/Sources/Other/SwiftUI/Views/BadgeLabel.swift
```diff
@@ -67,6 +67,7 @@ struct BadgeLabel: View {
             .padding(.trailing, 12)
             .padding(.vertical, 4)
             .background(Capsule().fill(backgroundColor))
+            .lineLimit(1)
```
pixlwave: Is this actually necessary now with the flow layout? The issue being that setting a large accessibility font will likely result in the labels being clipped.

ElementX/Sources/Other/SwiftUI/Views/RoomHeaderView.swift
```diff
@@ -84,7 +84,7 @@ struct RoomHeaderView: View {
         switch roomHistorySharingState {
         case .shared: \.history
         case .worldReadable: \.userProfileSolid
-        case .none: nil
+        default: nil
```
pixlwave: It would be nicer on this switch to fill in the remaining cases here so that when  new one is added we get a compile error and can go ask what the correct icon would be.

ElementX/Sources/Screens/RoomDetailsScreen/View/RoomDetailsScreen.swift
```diff
@@ -7,6 +7,7 @@
 //
 
 import Compound
+import MatrixRustSDK
```
pixlwave: Is this necessary (I thought we used a Swift type now)? If it is though, let's move the import below the `// MARK: - Previews` please.

ElementX/Sources/Services/Room/RoomInfoProxyProtocol.swift
```diff
@@ -147,7 +147,7 @@ extension RoomInfoProxyProtocol {
         case .worldReadable:
             .worldReadable
         default:
```
pixlwave: For consistency with my previous comment, we should probably do the same here too, that way we'll see any new cases when making an SDK update.

pixlwave (commented):
Very nice, thank you for this and so sorry I took a long time to come back to it. Do feel free re-request a review in the GitHub UI as it helps create a notification :)

The main comment is the translations, everything else is minor.

ElementX/Sources/Other/SwiftUI/Views/BadgeLabel.swift
kaylendog: This is something that has been troubling me also - it will be an issue for languages that express the `New members don't see history` etc. sentiment in more characters than English does. Wrapping may indeed be necessary there, but it looks very strange. Perhaps a line limit of two will suffice. Will query design.

ElementX/Sources/Screens/RoomDetailsScreen/View/RoomDetailsScreen.swift
kaylendog: I don't believe there's a Swift type for `RoomHistoryVisibility` itself? I did make the enum `RoomHistorySharingState`, but since this preview uses the `JoinedRoomProxyMock` directly, I need to use the former here. Will move the import.

pixlwave: Ahhh understood 👍

PR #5051, opened 2026-02-05
Bump the RustSDK to v26.02.03

stefanceriu:
Fix breaking changes and add support for the new background task error reporting delegate and the new room invite latest event.

### DIFF
--- a/ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
+++ b/ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
@@ -294,7 +294,9 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
                 
                 if let senderID {
                     let sender = TimelineItemSender(senderID: senderID, senderProfile: profile)
-                    attributedLastMessage = AttributedString(L10n.stateEventRoomInviteYou(sender.displayName ?? sender.id))
+                    let senderDisplayName = sender.displayName ?? sender.id
+                    let invitedYouString = eventStringBuilder.stateEventStringBuilder.buildInvitedYouString(senderDisplayName)
+                    attributedLastMessage = AttributedString(invitedYouString)
                 }
             case .none:
                 break
--- a/ElementX/Sources/Services/Timeline/TimelineItems/RoomStateEventStringBuilder.swift
+++ b/ElementX/Sources/Services/Timeline/TimelineItems/RoomStateEventStringBuilder.swift
@@ -58,7 +58,7 @@ struct RoomStateEventStringBuilder {
             if senderIsYou {
                 return L10n.stateEventRoomInviteByYou(member)
             } else if memberIsYou {
-                return L10n.stateEventRoomInviteYou(senderDisplayName)
+                return buildInvitedYouString(senderDisplayName)
             } else {
                 return L10n.stateEventRoomInvite(senderDisplayName, member)
             }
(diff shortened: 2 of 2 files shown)


NSE/Sources/NSEUserSession.swift
```diff
@@ -140,4 +140,9 @@ private final class ClientDelegateWrapper: ClientDelegate {
     func didRefreshTokens() {
         MXLog.info("Delegating session updates to the ClientSessionDelegate.")
     }
+    
+    func onBackgroundTaskErrorReport(taskName: String, error: MatrixRustSDK.BackgroundTaskFailureReason) {
+        MXLog.error("Received background task error: \(error)")
+        exit(0) // Not much we can do, best restart the NSE process
```
stefanceriu: Or perhaps it's better to leave it alone as we just discovered exit will filter out the notification if you have the entitlement in the app?

stefanceriu: and even if we leave it we should probably only crash on panics?

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -289,6 +289,13 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
                 let sender = TimelineItemSender(senderID: senderID, senderProfile: profile)
                 attributedLastMessage = eventStringBuilder.buildAttributedString(for: content, sender: sender, isOutgoing: isOwn)
                 lastMessageDate = Date(timeIntervalSince1970: TimeInterval(timestamp / 1000))
+            case .remoteInvite(let timestamp, let senderID, let profile):
+                lastMessageDate = Date(timeIntervalSince1970: TimeInterval(timestamp / 1000))
+                
+                if let senderID {
+                    let sender = TimelineItemSender(senderID: senderID, senderProfile: profile)
+                    attributedLastMessage = AttributedString(L10n.stateEventRoomInviteYou(sender.displayName ?? sender.id))
```
pixlwave: I'm guessing it isn't possible to construct a membership event content to let the `eventStringBuilder` handle this for us? I wonder if we should move this to a dedicated build method just so that it is all centralised.

NSE/Sources/NSEUserSession.swift
pixlwave: Yeah good question. I would be tempted to only log and not `exit`/`fatalError` here so that we collect logs but the user still gets a "Notification" notification. If it's gone seriously wrong it's not like it will finish processing anyway, and if it is unrelated then at least they still get the intended notification.

pixlwave (approved):
Nice, thanks 👌

NSE/Sources/NSEUserSession.swift
stefanceriu: d7c7304eea35f667872bd91d4f7ed3b487e6984c

ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
stefanceriu: Presumably some info is missing otherwise @Hywan  would've done that already on the SDK side instead of a new `remoteInvite` type. Is 592232b2511f13c9b91c75c9910c65377987284a what you had in mind?

pixlwave: Yep yep, exactly 👌

Hywan: An invite room is only composed of stripped state-events, so we can't re-use `Remote`, we need to create a new variant with cherry-picked information, hence `RemoteInvite`.

stefanceriu: Bingo, thank you 🙏

PR #5054, opened 2026-02-05
Add boot detection in the NSE and use this to inform the user that there may be more notifications.

pixlwave:
This PR aims to help with the problem that APNs servers only store the most recent notification when delivery fails. This is very much a "best effort" attempt and won't work in all cases but it works like this:

- Introduce a new notification type `receivedWhileOfflineNotification`:
04 - Shorter

This notification type is *always* noisy, will be automatically discarded when opening the app and can only ever be delivered once.

- Deliver this the first time a notification is received whilst `DataProtectionManager.isDeviceLockedAfterReboot` reports `true`.

- If the phone is unlocked after reboot before a notification is received, then also deliver the new notification type for the first notification.

- If the phone has been on for more than 15-minutes before a notification is received, then don't deliver the new notification type.

### DIFF
--- a/NSE/Sources/NotificationServiceExtension.swift
+++ b/NSE/Sources/NotificationServiceExtension.swift
@@ -61,7 +61,7 @@ class NotificationServiceExtension: UNNotificationServiceExtension {
         // the target configuration will fail. We could call exit(0) here, however with the
         // notification filtering entitlement that results in the notification being discarded
         // so we need to wait for the delegate method to be called and bail out there instead.
-        if !DataProtectionManager.isDeviceLockedAfterReboot(containerURL: URL.appGroupContainerDirectory),
+        if !BootDetectionManager.isDeviceLockedAfterReboot(containerURL: URL.appGroupContainerDirectory),
            Self.targetConfiguration == nil {
             Self.targetConfiguration = Target.nse.configure(logLevel: settings.logLevel,
                                                             traceLogPacks: settings.traceLogPacks,
@@ -80,7 +80,7 @@ class NotificationServiceExtension: UNNotificationServiceExtension {
     private func handle(_ request: UNNotificationRequest, withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void) async {
         // If we skipped configuring the target it means we can't write to the app group, so we're unlikely to
         // be able to create a session (and even if we could, we would be missing the lightweightTokioRuntime).
-        // Additionally APNs servers only store the most recent notification when the device is powered off.
+        // Additionally, APNs servers only store the most recent notification when the device is powered off.
         // So lets a) skip processing the notification and b) deliver a special "offline" notification as a workaround.
         guard Self.targetConfiguration != nil else {
             // MXLog isn't configured:
@@ -180,7 +180,7 @@ class NotificationServiceExtension: UNNotificationServiceExtension {
     ///
     /// Note that this only handles the first-boot case. When the SDK is able to compute the unread count, we should start to use the NSE,
     /// remote-notifications (content-available) and background app refreshes to fetch and deliver our notifications as a more robust solution.
-    func shouldDeliverReceivedWhileOfflineNotification() -> Bool {
+    private func shouldDeliverReceivedWhileOfflineNotification() -> Bool {
         if Self.hasHandledFirstNotificationSinceBoot {
             // If we've already handled the first notification in this process there's no need to continue.
             return false
@@ -188,7 +188,7 @@ class NotificationServiceExtension: UNNotificationServiceExtension {
         
         Self.hasHandledFirstNotificationSinceBoot = true
         
-        guard let currentBootTime = systemBootTime() else {
+        guard let currentBootTime = BootDetectionManager.systemBootTime() else {
             // There's not much we can do if the boot time is unknown, so don't show the offline notification.
             return false
         }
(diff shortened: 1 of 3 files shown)


ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -291,6 +294,10 @@ final class AppSettings {
     @UserPreference(key: UserDefaultsKeys.pusherProfileTag, storageType: .userDefaults(store))
     var pusherProfileTag: String?
     
+    /// The device's last boot time as recorded by the NSE.
+    @UserPreference(key: UserDefaultsKeys.lastNotificationBootTime, storageType: .userDefaults(store))
+    var lastNotificationBootTime: TimeInterval?
```
pixlwave, line 299: Maybe this should actually be a `Date`, but I went with the timestamp given that's what comes out of the system call.

ElementX/Sources/Services/Notification/Manager/NotificationManager.swift
```diff
@@ -56,8 +56,14 @@ final class NotificationManager: NSObject, NotificationManagerProtocol {
                 self?.enableNotifications(newValue)
             }
             .store(in: &cancellables)
-    }
         
+        NotificationCenter.default.publisher(for: UIApplication.didBecomeActiveNotification)
+            .sink { [weak self] _ in
+                self?.removeReceivedWhileOfflineNotification()
```
stefanceriu, line 62: Can't you just directly do this? Presumably the app wasn't running when it was delivered 🤔

NSE/Sources/NotificationServiceExtension.swift
```diff
@@ -72,14 +78,30 @@ class NotificationServiceExtension: UNNotificationServiceExtension {
     }
     
     private func handle(_ request: UNNotificationRequest, withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void) async {
-        // If we skipped configuring the target it means we can't write to the app group, so we're unlikely
-        // to be able to create a session (and even if we could, we would be missing the lightweightTokioRuntime),
-        // so instead lets deliver the default generic notification and avoid attempting to process the notification.
+        // If we skipped configuring the target it means we can't write to the app group, so we're unlikely to
+        // be able to create a session (and even if we could, we would be missing the lightweightTokioRuntime).
+        // Additionally APNs servers only store the most recent notification when the device is powered off.
```
stefanceriu:
```suggestion
        // Additionally, APNs servers only store the most recent notification when the device is powered off.
```

NSE/Sources/NotificationServiceExtension.swift
```diff
@@ -149,7 +171,83 @@ class NotificationServiceExtension: UNNotificationServiceExtension {
...
+        //
+        // Note: We could actually solve this by having Sygnal add a timestamp to the notification payload 🤔
+        if Date.now.timeIntervalSince(Date(timeIntervalSince1970: currentBootTime)) > Self.firstNotificationThreshold {
+            return false
+        } else {
+            return true
+        }
+    }
+    
+    func systemBootTime() -> TimeInterval? {
```
stefanceriu: I know this is mostly irrelevant but let's mark this 2 new methods as private, for consistency.

stefanceriu: Might also be worth puting this systemBootTime in an utility somewhere 🤔

NSE/Sources/NotificationServiceExtension.swift
```diff
@@ -149,7 +171,83 @@ class NotificationServiceExtension: UNNotificationServiceExtension {
...
+    }
+    
+    /// Delivers a generic notification informing the user that they have one or more new messages.
+    ///
+    /// Note: it is safe to call this method multiple times as it simply replaces any existing instance of the notification
+    /// with a fresh copy, meaning it won't queue multiple copies but will still re-play the notification sound.
+    private func deliverReceivedWhileOfflineNotification(for originalRequest: UNNotificationRequest) {
+        // Don't log until the app hooks have been run:
+        // swiftlint:disable:next print_deprecation
+        print("Delivering the 'received while offline' notification.")
```
stefanceriu: How do you know if the hooks have been run or not in this context?

NSE/Sources/NotificationServiceExtension.swift
```diff
@@ -149,7 +171,83 @@ class NotificationServiceExtension: UNNotificationServiceExtension {
...
+        
+        let content = UNMutableNotificationContent()
+        content.body = L10n.notificationReceivedWhileOfflineIos
+        content.badge = originalRequest.content.unreadCount as NSNumber?
```
stefanceriu, line 233: I wonder if this is going to create more problems than it fixes 😅

stefanceriu (approved):
Nice work! 👏

NSE/Sources/NotificationServiceExtension.swift
pixlwave: Ha, well my logic was:
- We do this in [`discardNotification`](https://github.com/element-hq/element-x-ios/blob/bff328554a5038b1bdc88a23b984660500841533/NSE/Sources/NotificationHandler.swift#L86)
- Making sure the badge is as up to date as possible should help push people into the app if they miss this notification considering that if you phone has been off, you may potentially be coming back to hundreds of notifications.

NSE/Sources/NotificationServiceExtension.swift
pixlwave: Oh yeah good point, I refactored and then later undid it, but there's a nice half-way house here now.

NSE/Sources/NotificationServiceExtension.swift
pixlwave: Purely by assumption when looking at where this is called from. Comment updated 😅

ElementX/Sources/Services/Notification/Manager/NotificationManager.swift
pixlwave: We discussed this and I didn't do that because of background app refreshes.

PR #5056, opened 2026-02-05
Use our own `JoinRule` type which doesn't include the reserved `.private` case.

pixlwave:
iOS equivalent to https://github.com/element-hq/element-x-android/issues/6127 which is an implementation of [MSC4413](https://github.com/matrix-org/matrix-spec-proposals/pull/4413).

We were only using the `.private` type in mocks (first commit tidies these up) and switching on it (which was always handled as part of the same case as `.invite`). However I added the second commit which creates our own Swift equivalent of the Rust type without the `.private` case so that we won't end up using it by accident (plus it helps on my eternal quest of changing every `Id` → `ID` 😄).

### DIFF
--- a/ElementX/Sources/Services/Room/JoinRule.swift
+++ b/ElementX/Sources/Services/Room/JoinRule.swift
@@ -32,7 +32,7 @@ public enum JoinRule: Equatable, Hashable {
         case .public: .public
         case .invite: .invite
         case .knock: .knock
-        case .private: .invite // Assume .private is .invite – private shouldn't be in used (see MSC4413).
+        case .private: .invite // Assume .private is .invite – private shouldn't be used (see MSC4413).
         case .restricted(let rules): .restricted(rules: rules.map(AllowRule.init))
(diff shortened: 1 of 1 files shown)


ElementX/Sources/Services/Room/JoinRule.swift
```diff
@@ -0,0 +1,74 @@
...
+    case knockRestricted(rules: [AllowRule])
+    /// A custom join rule, up for interpretation by the consumer.
+    case custom(rawValue: String)
+    
+    init(rustValue: MatrixRustSDK.JoinRule) {
+        self = switch rustValue {
+        case .public: .public
+        case .invite: .invite
+        case .knock: .knock
+        case .private: .invite // Assume .private is .invite – private shouldn't be in used (see MSC4413).
```
stefanceriu:
```suggestion
        case .private: .invite // Assume .private is .invite – private shouldn't be in use (see MSC4413).
```

```suggestion
        case .private: .invite // Assume .private is .invite – private shouldn't be used (see MSC4413).
```

PR #5076, opened 2026-02-09
Fix PassthroughWindow logic for os26

stefanceriu:
and this crash https://sentry.tools.element.io/organizations/element/issues/11227218/?project=44&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D%20selecti&referrer=issue-stream

`ElementX/AuthenticationFlowCoordinator.swift:256: Fatal error: Unexpected transition: Context(event: Optional(ElementX.AuthenticationFlowCoordinator.Event.continueWithOIDC), fromState: ElementX.AuthenticationFlowCoordinator.State.serverSelectionScreen, toState: ElementX.AuthenticationFlowCoordinator.State.serverSelectionScreen, userInfo: Optional((ElementX.OIDCAuthorizationDataProxy(underlyingData: MatrixRustSDK.OAuthAuthorizationData), <UIWindow: 0x10be3ad50; frame = (0 0; 402 874); autoresize = W+H; tintColor = <UIDynamicCatalogColor: 0x1109944b0; name = gray1400>; gestureRecognizers = <NSArray: 0x12c861050>; layer = <UIWindowLayer: 0x11388d200>>)))`

### DIFF
--- a/ElementX/Sources/Screens/Authentication/ServerConfirmationScreen/ServerConfirmationScreenViewModel.swift
+++ b/ElementX/Sources/Screens/Authentication/ServerConfirmationScreen/ServerConfirmationScreenViewModel.swift
@@ -58,8 +58,12 @@ class ServerConfirmationScreenViewModel: ServerConfirmationScreenViewModelType, 
             Task { state.window = window }
         case .confirm:
             switch state.mode {
-            case .confirmation: Task { await confirmServer() }
-            case .picker: Task { await pickServer() }
+            case .confirmation:
+                startLoading()
+                Task { await confirmServer() }
+            case .picker:
+                startLoading()
+                Task { await pickServer() }
             }
         case .changeServer:
             actionsSubject.send(.changeServer)
@@ -69,6 +73,9 @@ class ServerConfirmationScreenViewModel: ServerConfirmationScreenViewModelType, 
     // MARK: - Private
     
     private func confirmServer() async {
+        defer { stopLoading() }
+        startLoading()
+        
         let homeserver = authenticationService.homeserver.value
         
         // If the login mode is unknown, the service hasn't been configured and we need to do it now.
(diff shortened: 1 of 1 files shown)


ElementX/Sources/Screens/Authentication/ServerConfirmationScreen/ServerConfirmationScreenViewModel.swift
```diff
@@ -58,8 +58,12 @@ class ServerConfirmationScreenViewModel: ServerConfirmationScreenViewModelType,
...
         case .confirm:
             switch state.mode {
-            case .confirmation: Task { await confirmServer() }
-            case .picker: Task { await pickServer() }
+            case .confirmation:
+                startLoading()
+                Task { await confirmServer() }
+            case .picker:
+                startLoading()
+                Task { await pickServer() }
```
pixlwave, line 66: Hmmmm, this goes against the two comments below 🤔

> // Note: We don't show the spinner until now as it isn't needed if the service is already
> // configured and we're about to use password based login

and

> // Note: We don't show the spinner until now as it isn't needed if the service is already
> // configured and we're about to use password based login

pixlwave:
> I wonder if we should do some kind of backport for `Task.immediate` so that we at least have support for it on iOS 26 🤔

Apparently a limited version exists for earlier OS versions when you're already on the MainActor:
https://forums.swift.org/t/back-deployment-of-se-0472-task-immediate/81576/7

```
@_spi(MainActorUtilities) import Swift

@MainActor func foo() {
    Task.startOnMainActor {
        await bar()
    }
}
```

stefanceriu:
> > I wonder if we should do some kind of backport for `Task.immediate` so that we at least have support for it on iOS 26 🤔
> 
> Apparently a limited version exists for earlier OS versions when you're already on the MainActor: https://forums.swift.org/t/back-deployment-of-se-0472-task-immediate/81576/7
> 
> ```
> @_spi(MainActorUtilities) import Swift
> 
> @MainActor func foo() {
>     Task.startOnMainActor {
>         await bar()
>     }
> }
> ```

We went ahead and looked into this but it ended up being too complicated to adopt. Decided to go for the simpler solution.

PR #5085, opened 2026-02-10
Enable spaces second iteration FFs

Velin92:
- Space list and filtering
- Creating spaces and rooms in spaces
- Manage spaces, add/remove children, handle settings and permissions

### DIFF
--- a/ElementX/Sources/Application/Settings/AppSettings.swift
+++ b/ElementX/Sources/Application/Settings/AppSettings.swift
@@ -75,9 +75,9 @@ final class AppSettings {
         case linkPreviewsEnabled
         case focusEventOnNotificationTap
         case linkNewDeviceEnabled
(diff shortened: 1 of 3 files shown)


pixlwave:
Seeing as we want to enable the feature in production, should we remove the flags? Otherwise anyone who has previously enabled and then disabled the flags won't see the features…

Velin92:
My idea is to keep the flag at least for now until the release date on the store of the 16th, so in the extraordinary case there is a blocking issue that might require us to even turn off one of the features, we can do that and save the release.
Is just a safety measure, we will definitely remove the flag for the next one.
As for the people having it turned off (which means they would have needed to have it turned on first and then off to store the value) I feel this is a VERY edge case, and we should not care about.

stefanceriu (approved):
lgtm, thanks! 👍

PR #5091, opened 2026-02-11
Update the active Space Filter Button in the Chats header to be Primary instead of Accent

americanrefugee:
Fixes https://github.com/element-hq/element-x-ios/issues/5090


pixlwave:
Note to self: Check the iOS 18 implementation.

PR #5098, opened 2026-02-12
Re-write the space flow state machine configuration and cover it more in UI tests.

pixlwave:
The state machine changes were done with Claude, which I combed through afterwards and made a few manual tweaks to afterwards.

### DIFF
--- a/ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
@@ -197,17 +197,28 @@ class SpaceFlowCoordinator: FlowCoordinatorProtocol {
     
     // swiftlint:disable:next function_body_length
     private func configureStateMachine() {
-        stateMachine.addRoutes(event: .start, transitions: [.initial => .space])
-        stateMachine.addRoutes(event: .startUnjoined, transitions: [.initial => .joinSpace])
-        stateMachine.addRoutes(event: .joinedSpace, transitions: [.joinSpace => .space])
-        stateMachine.addRoutes(event: .leftSpace, transitions: [.space => .leftSpace])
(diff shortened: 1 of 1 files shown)


ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
```diff
@@ -195,161 +195,144 @@ class SpaceFlowCoordinator: FlowCoordinatorProtocol {
...
-        
-        stateMachine.addRoutes(event: .addRooms, transitions: [.space => .addingRooms]) { [weak self] _ in
-            self?.presentSpaceAddRoomsScreen()
-        }
+        stateMachine.addRoutes(event: .start, transitions: [.initial => .space])
+        stateMachine.addRoutes(event: .startUnjoined, transitions: [.initial => .joinSpace])
+        stateMachine.addRoutes(event: .joinedSpace, transitions: [.joinSpace => .space])
+        stateMachine.addRoutes(event: .leftSpace, transitions: [.space => .leftSpace])
+        stateMachine.addRoutes(event: .addRooms, transitions: [.space => .addingRooms])
         stateMachine.addRoutes(event: .dismissedAddRooms, transitions: [.addingRooms => .space])
```
Velin92: Maybe is me, but I feel mixing these two routes transitions, he event transitions ones with the RouteMapping below, is a bit confusing I would rather just have all the transitions inside the route mapping, so that everything is in the same switch statement

pixlwave: Yeah, I agree with this too tbh, will update 👍

pixlwave: Done in https://github.com/element-hq/element-x-ios/pull/5098/commits/9f8255e7df415c530256d14ee47d7f1f8b36d39a

Velin92 (approved):
LGTM

PR #5099, opened 2026-02-12
Fix integration tests room header discovery

stefanceriu:
Success here https://github.com/element-hq/element-x-ios/actions/runs/21943427127/job/63381872071


pixlwave (commented):
✅ (I'm too late apparently 🙈)

PR #5104, opened 2026-02-12
Fix A11y tests

Velin92:
Review commit by commit
- First since iOS 26 and ids on top of navigation stacks do not work properly together, we decided to remove the navigation bar from the a11y tests, to do so an ElementNavigationStack has been introduced.
- Added labels to objects that were missing a description


ElementX/Sources/Other/SwiftUI/ElementNavigationStack.swift
```diff
@@ -0,0 +1,33 @@
...
+                    content
+                }
+            }
+        } else {
+            NavigationStack {
+                content
+            }
+        }
+        #else
+        NavigationStack {
```
stefanceriu, line 28: Noice 👌

PR #5105, opened 2026-02-12
Replace the use of Ruby/Fastlane for running UnitTests on CI with new tools written in Swift.

stefanceriu:
This PR replaces the Unit and Preview tests fastlane step with a version written in swift. The "workflow" sit under the CI Tools subcommand with CI.UnitTests replacing the whole fastlane lane. Can be invoked from the command line using `swift run tools ci unit-tests`

It introduces primitives for:
* dealing with linting
* generic test running
* coverage collection
* test result collection
* result zipping 

It also introduces swift-log and tool wide static logger instance.

### DIFF
--- a/Tools/Sources/Commands/CI/UnitTests.swift
+++ b/Tools/Sources/Commands/CI/UnitTests.swift
@@ -5,13 +5,13 @@ import Foundation
 struct UnitTests: AsyncParsableCommand {
     static let configuration = CommandConfiguration(commandName: "unit-tests",
                                                     abstract: "Runs the unit test CI workflow: lint, unit tests, preview tests, and result collection.")
-
+    
     @Option(help: "Device name for unit tests.")
     var device = "iPhone 17"
-
+    
     @Option(help: "iOS version for the simulator.")
     var osVersion = "26.1"
-
+    
     func run() async throws {
         try await CI.lint()
         
@@ -44,7 +44,7 @@ struct UnitTests: AsyncParsableCommand {
             failures.append("Preview tests failed: \(error)")
             logger.error("\n❌ Preview tests failed.\n")
         }
-
+        
         // Zip results (best-effort, useful for CI artifact uploads)
         await CI.zipResults(bundles: ["UnitTests.xcresult", "PreviewTests.xcresult"],
                             outputName: "UnitTests.zip")
--- a/Tools/Sources/Commands/CI/RunTests.swift
+++ b/Tools/Sources/Commands/CI/RunTests.swift
@@ -14,28 +14,28 @@ struct RunTests: AsyncParsableCommand {
                                                       swift run tools run-tests --scheme PreviewTests --create-simulator-name "iPhone SE (3rd generation)" \
                                                         --create-simulator-type com.apple.CoreSimulator.SimDeviceType.iPhone-SE-3rd-generation
                                                     """)
-
+    
     @Option(help: "The Xcode scheme to test.")
     var scheme: String
-
+    
     @Option(help: "The simulator device name to run tests on (e.g. 'iPhone 17').")
     var device = "iPhone 17"
-
+    
     @Option(help: "The iOS version to use for the simulator runtime (e.g. '26.1').")
     var osVersion = "26.1"
-
+    
     @Option(help: "Number of times to retry failed tests. Only the failing tests are re-run, not the entire suite.")
     var retries = 3
-
+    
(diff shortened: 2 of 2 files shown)


.github/workflows/compound-ios.yml
pixlwave, line 1: Let's not change this one so that we keep Compound de-coupled from EXI.

Tools/Sources/Commands/PrepareNextRelease.swift
```diff
@@ -0,0 +1,106 @@
...
+        // Regenerate the Xcode project
+        try Zsh.run(command: "xcodegen")
+
+        // Configure git
+        try setupGit()
+
+        // Commit and push
+        try Zsh.run(command: "git add \(targetFilePath) \(xcodeProjPath)")
+        try Zsh.run(command: "git commit -m 'Prepare next release'")
+        try gitPush()
```
pixlwave: It would be nice if we could use (and extend if necessary) the existing [Git wrapper](https://github.com/element-hq/swift-command-line-tools/blob/main/Sources/CommandLineTools/Release/Git.swift) from our command line tools.

stefanceriu: Yep, definitely. Same goes for `Zsh` vs `Shell`

Tools/Sources/Commands/PrepareNextRelease.swift
```diff
@@ -0,0 +1,106 @@
...
+            throw ValidationError("GITHUB_TOKEN environment variable is not set.")
+        }
+
+        // Get repo URL path without protocol prefix or .git suffix
+        guard
+            let repoURL = try Zsh.run(command:
+                "git ls-remote --get-url origin | sed 's#http://##g' | sed 's#https://##g' | sed 's#git@##g' | sed 's#.git##g'")?.trimmingCharacters(in: .whitespacesAndNewlines)
+        else {
+            throw ValidationError("Could not determine the repository URL.")
+        }
```
pixlwave: Maybe we should swiftformat all this stuff too 🫠

CONTRIBUTING.md
pixlwave, line 1: We should probably note that there's a mix here.

.gitignore
```diff
@@ -29,6 +29,8 @@ fastlane/report.xml
 fastlane/Preview.html
 fastlane/screenshots/**/*.png
 fastlane/test_output
+# Test output
```
pixlwave: Probably don't need this comment 😅

Tools/Sources/Commands/CI/CI.swift
```diff
@@ -0,0 +1,107 @@
...
+struct CI: ParsableCommand {
+    static let configuration = CommandConfiguration(abstract: "CI workflow commands that can be run both locally and in CI environments.",
+                                                    subcommands: [
+                                                        UnitTests.self,
+                                                        RunTests.self
+                                                    ])
+    
+    static let testOutputDirectory = "test_output"
+    
+    // MARK: - Logging
```
pixlwave: This one can go now.

Tools/Sources/Commands/CI/CI.swift
```diff
@@ -0,0 +1,107 @@
...
+        
+    // MARK: - Linting
+    
+    /// Runs SwiftFormat in lint mode against the current directory.
+    static func lint() async throws {
+        logger.info("\n🔍 Running SwiftFormat lint…\n")
+        
+        do {
+            _ = try await run(.name("swiftformat"), ["--lint", "."])
```
pixlwave: `_ =` should be unnecessary given the discardable result.

Tools/Sources/Commands/CI/CI.swift
pixlwave: Tbh, there's basically one of these per function, which just adds noise to me.

Tools/Sources/Commands/CI/UnitTests.swift
```diff
@@ -0,0 +1,68 @@
...
+        var failures: [String] = []
+        
+        // Run unit tests
+        do {
+            logger.info("\n🧪 Running unit tests…\n")
+            try await RunTests.parse([
+                "--scheme", "UnitTests",
+                "--device", device,
+                "--os-version", osVersion
+            ]).run()
```
pixlwave, line 27: TIL you can do this, although it seems somewhat not type-safe 😅 I thought you could also do e.g. `RunTests(scheme: "UnitTests", device: device, …).run()`.

PR #5108, opened 2026-02-13
Adopt the new tracing configuration that allows maximum size and age limits on all logs.

stefanceriu:
Set that 100Mb and a week and remove previous rageshake uploading checks as they are no longer required.

### DIFF
--- a/ElementX/Sources/Other/Logging/Tracing.swift
+++ b/ElementX/Sources/Other/Logging/Tracing.swift
@@ -10,9 +10,6 @@ import Foundation
 import MatrixRustSDK
 
 enum Tracing {
-    /// The base filename used for log files. This may be suffixed by the target
-    /// name and other log management metadata during rotation.
-    static let filePrefix = "console"
     /// The directory that stores all of the log files.
     static var logsDirectory: URL {
         if ProcessInfo.isRunningIntegrationTests {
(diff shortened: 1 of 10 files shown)


ElementX.xcodeproj/xcshareddata/xcschemes/ElementX.xcscheme
pixlwave, line 1: I think you can re-run `xcodegen`.

ElementX/Sources/Other/Logging/Tracing.swift
```diff
@@ -30,22 +27,12 @@ enum Tracing {
         .appGroupContainerDirectory
     }
     
-    static let fileExtension = "log"
+    static let fileExtension = ".log"
```
pixlwave, line 30: Do we need to tell EXA about requiring the `.`?

ElementX/Sources/Other/Logging/Tracing.swift
```diff
@@ -55,9 +42,10 @@ enum Tracing {
                      extraTargets: [currentTarget],
                      writeToStdoutOrSystem: true,
                      writeToFiles: .init(path: logsDirectory.path(percentEncoded: false),
-                                         filePrefix: fileName,
+                                         filePrefix: filePrefix,
                                          fileSuffix: fileExtension,
-                                         maxFiles: maxFiles),
+                                         maxTotalSizeBytes: 100 * 1024 * 1024, // 100Mb
+                                         maxAgeSeconds: 7 * 24 * 60 * 60), // One week
```
pixlwave, line 50: Maybe worth adding a note here that when compressed these will be below the 50MB Cloudflare limit.

UnitTests/Sources/BugReportServiceTests.swift
```diff
@@ -139,31 +135,6 @@ class BugReportServiceTests: XCTestCase {
         
         XCTAssertEqual(defaultConfigurationResponse.reportURL, initialURL.absoluteString.replacingOccurrences(of: "submit", with: "123"))
     }
-    
-    func testLogsMaxSize() {
```
pixlwave, line 143: So long old friend! 👋

ElementX/Sources/Other/Logging/Tracing.swift
stefanceriu: cc @element-hq/element-android 😅

PR #5109, opened 2026-02-13
Compound - Swift 6.2 and Main actor isolation

### LINKED ISSUE #4991: [Task] Compound - Swift 6 Migration

Velin92:
Compound now uses Swift 6.2, and Main Actor isolation.

The compound tests alongside some small places in EXI had to now be strictly made to use `@MainActor` to properly compile and run.

fixes #4991

### DIFF
--- a/compound-ios/Sources/Compound/Colors/CompoundUIColors.swift
+++ b/compound-ios/Sources/Compound/Colors/CompoundUIColors.swift
@@ -62,5 +62,3 @@ private extension UITraitCollection {
         userInterfaceStyle == .light
     }
 }
-
-extension CompoundUIColorTokens: @unchecked @retroactive Sendable { }


compound-ios/Sources/Compound/Colors/CompoundUIColors.swift
```diff
@@ -15,10 +15,12 @@ public extension UIColor {
...
 
 /// The colours used by Element as defined in Compound Design Tokens.
-/// This struct contains only the colour tokens in a more usable form.
+/// This class contains only the colour tokens in a more usable form.
+/// Since this can be used by attributed strings which may run in non isolated concurrent contexts,
+/// The object needs to be nonisolated.
 @Observable
 @dynamicMemberLookup
-public class CompoundUIColors {
+public final nonisolated class CompoundUIColors {
```
Velin92, line 23: This is required because we use this in attributed strings builder at first I thought... maybe we should do this also for the normal CompoundColors (the ones based on SwiftUI) so we don't need to add `@MainActor` in some places that use color in the main app, but this lead to a lot of changes in a lot of places, and I think, that considering the CompoundColors based on SwiftUI are all related to swiftUI stuff which is usually running on the MainActor I think it makes sense to keep this distinction where the UI one can be `nonisolated`, while the SwiftUI are isolated on the `@MainActor`

ElementX/Sources/Screens/Timeline/View/Style/TimelineItemBubbleBackground.swift
```diff
@@ -16,10 +16,10 @@ extension View {
     ///   - color: self explanatory, defaults to subtle secondary
     func bubbleBackground(isOutgoing: Bool = true,
                           insets: EdgeInsets = .init(top: 8, leading: 12, bottom: 8, trailing: 12),
-                          color: Color? = .compound.bgSubtleSecondary) -> some View {
+                          color: @autoclosure @MainActor () -> Color? = .compound.bgSubtleSecondary) -> some View {
```
Velin92, line 19: this was fun... since main actor isolation supports that everything should run on the main actor, but this means that even if you mark a function as MainActor it doesn't guarantee that the default assignment of a parameter is going to run on the main actor. I know, it's a bit stupid that it's not guaranteed..., but autoclosures explicitly fix the issue elegantly without changing the logic, we just initialise with a `@MainActor` closure

ElementX/Sources/Other/SwiftUI/Views/ToolbarButton.swift
```diff
@@ -9,6 +9,7 @@ import Compound
 import SwiftUI
 
 struct ToolbarButton: View {
+    @MainActor
```
Velin92, line 12: these `@MainActor` things needed to be added since we are accessing the CompoundColors which are MainActor isolated, check my previous comment for some extra info.

compound-ios/Package.swift
```diff
@@ -22,6 +22,9 @@ let package = Package(
                 .product(name: "CompoundDesignTokens", package: "compound-design-tokens"),
                 .product(name: "SwiftUIIntrospect", package: "SwiftUI-Introspect"),
                 .product(name: "SFSafeSymbols", package: "SFSafeSymbols")
+            ],
+            swiftSettings: [
+                .defaultIsolation(MainActor.self)
```
Velin92, line 27: We could in theory also add this to the test suite, but I see that Apple doesn't do that by default, and that XCTest would then still need to be imported as `@preocuncurrency` (I think they are all objc based stuff so they will never really support concurrency as of right now, maybe we would need to transition to SwiftTesting for that? Not sure). In the end this doesn't change anything on tests if you have it or not.

compound-ios/Tests/CompoundTests/OverrideColorTests.swift
```diff
@@ -10,20 +10,26 @@
...
+            let tokens = CompoundColorTokens()
+            XCTAssertEqual(colors.textPrimary, tokens.textPrimary)
+            
+            colors.override(\.textPrimary, with: .pink)
+            XCTAssertEqual(colors.textPrimary, .pink)
+            
+            colors.override(\.textPrimary, with: nil)
+            XCTAssertEqual(colors.textPrimary, tokens.textPrimary)
+        }
     }
```
Velin92, line 29: Yeah no idea wtf is going on here, but this is the only way to fix this... seems to be some kind of weird interaction between main actor isolation and the `@Observable` pattern for overriding the variable.

It doesn't crash on them however... it crashes at the end, when is deiniting the values. Very weird. Making it `nonisolated` fixes the issue however, as it appears below for `UIColors` but we don't really need SwiftUI colors to be nonisolated.

compound-ios/Package.swift
pixlwave: How about Swift Testing?

Velin92: I mean yeah... maybe? but that is another can of worm to open all by itself.
Also not sure how it will work with PreviewTests

Velin92: we have an answer now, and yes the bug is fixed with Swift Testing: https://github.com/element-hq/element-x-ios/pull/5109

Let's decide there if we want to use the isolated context for tests too for compound then.

compound-ios/Sources/Compound/Colors/CompoundUIColors.swift
stefanceriu: Makes sense to me, let's see how it all goes 👍

compound-ios/Tests/CompoundTests/OverrideColorTests.swift
stefanceriu: I'm honestly happy this is limited to the tests, it's good enough for me 👍

PR #5110, opened 2026-02-13
Swift Testing for Compound

Velin92:
Replaces XCTest Suite in Compound Tests with a Swift Testing Suite.

### DIFF
--- a/compound-ios/Tests/CompoundTests/PreviewTests.swift
+++ b/compound-ios/Tests/CompoundTests/PreviewTests.swift
@@ -68,15 +68,7 @@ struct PreviewTests {
         let imageRenderer = ImageRenderer(content: preferenceReadingView)
         _ = imageRenderer.uiImage
         
-        var sanitizedSuiteName = testName
-        // Handle both old XCTest style (testFoo()) and new Swift Testing style (foo())
-        if sanitizedSuiteName.hasPrefix("test") {
-            sanitizedSuiteName = String(sanitizedSuiteName.suffix(sanitizedSuiteName.count - "test".count).dropLast(2))
-            sanitizedSuiteName = sanitizedSuiteName.prefix(1).lowercased() + sanitizedSuiteName.dropFirst()
-        } else {
-            // Remove trailing () if present
-            sanitizedSuiteName = String(sanitizedSuiteName.dropLast(2))
-        }
+        let sanitizedSuiteName = String(testName.dropLast(2))
         
         for snapshotDevice in snapshotDevices {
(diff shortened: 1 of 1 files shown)


Velin92:
The only thing left is to decide if we want these tests too to be isolated by default on the `@MainActor` or if they're fine the way they are implemented right now

compound-ios/Tests/CompoundTests/FontSizeTests.swift
```diff
@@ -8,100 +8,105 @@
...
-import XCTest
+import Testing
 
 @MainActor
-final class FontSizeTests: XCTestCase {
+@Suite
+struct FontSizeTests {
     /// Test all system text styles to assert mapping between `Font` and `UIFont`.
-    func testTextStyle() {
+    func textStyle() {
```
pixlwave, line 18: Is this one missing an `@Test`?

compound-ios/Tests/CompoundTests/OverrideColorTests.swift
```diff
@@ -8,36 +8,34 @@
...
+        #expect(colors.textPrimary == tokens.textPrimary)
+        
+        colors.override(\.textPrimary, with: .pink)
+        #expect(colors.textPrimary == .pink)
+        
+        colors.override(\.textPrimary, with: nil)
+        #expect(colors.textPrimary == tokens.textPrimary)
     }
     
     /// UIColors are nonisolated, so this is fine.
```
pixlwave: I don't think this comment is relevant now.

compound-ios/Tests/CompoundTests/PreviewTests.swift
```diff
@@ -72,8 +68,15 @@ class PreviewTests: XCTestCase {
...
         
-        var sanitizedSuiteName = String(testName.suffix(testName.count - "test".count).dropLast(2))
-        sanitizedSuiteName = sanitizedSuiteName.prefix(1).lowercased() + sanitizedSuiteName.dropFirst()
+        var sanitizedSuiteName = testName
+        // Handle both old XCTest style (testFoo()) and new Swift Testing style (foo())
+        if sanitizedSuiteName.hasPrefix("test") {
+            sanitizedSuiteName = String(sanitizedSuiteName.suffix(sanitizedSuiteName.count - "test".count).dropLast(2))
+            sanitizedSuiteName = sanitizedSuiteName.prefix(1).lowercased() + sanitizedSuiteName.dropFirst()
+        } else {
+            // Remove trailing () if present
```
pixlwave: > if present

It seems to always drop them in this branch, so we probably don't need to mention that.

pixlwave (approved):
Love it! 👏

PR #5116, opened 2026-02-16
Unify session verification with Android

### LINKED ISSUE #5115: [Task] Unify session verification with Android

Velin92:
fixes #5115 
based on: https://github.com/element-hq/element-x-android/pull/5981

Essentially this
- Removes any loading screen,and just unifies every loading state in the same view used for the initial state but adds a loading state to it
- Removes the extra confirmation state before starting the SAS verification, we want everything to happen with one single tap, so we expect the other device to initiate the confirmation
- Previews have been changes to show some possible combinations of how the new loading state appears


stefanceriu (approved):
Very nice, thank you! 
I do think you're going to need to update the UI tests too, right?

PR #5119, opened 2026-02-17
Swift Testing for Unit Tests PART 1

Velin92:
PART 1
Since I have no more Copilot requests left not sure when I will do the PART 2 however this first was already very good since allowed us to:
- Define a deferFulfillment that is indepent of XCTestCase.
- Migrated most of the tests aside some bigger ones.
- Establish how Swift Testing works, how it should be used, and some testing patterns that depend on how you are setting up and tearing down the test.

Most of the code is AI generated, but a lot of manual touches were done to fix some small issues, and improve the codebase to enforce some good practices for Swift Testing.

## Key Takeaways

- Tests can now be structs, which makes them faster and more efficient, if we can we should always try to have a test to be a `struct`
- Remember that when a test changes its internals, and is a struct, you need to mark the test a mutating func, this has no impact since each test is a different instance of the struct, but keep that in mind.
- The init of a struct, is also is equivalent to the `setUp()` in `XCTestCase`
- There is only one case where we want to use a `final class` instead of a `struct`, and that's when we need to implement `deinit` which is equivalent to `tearDown` in `XCTestCase` (you could implement `deinit` to a struct but you need to conform to `~Copyable` but believe me is a rabbit hole you do not want to enter in, and performance wise would make the struct behave like a class)
- Since init and deinit are called after every tests this mens we can safely simply have (for tests that require to be initialised with a setup that requires an argument/parameter) a mutating  private setup function, and call it at the start of each test like we did before, and mark the test itself as mutating (not necessary if the test is a class).
- `#require` is very nice, it's equivalent to `XCTUnwrap` should be used in place of guard let.
- `XCTFail` is equivalent to `Issue.record`
- The most similar thing to `expectation` is `confirmation` but beware it doesn't have a `wait` with timeout mechanism, so you will need to do the waiting inside the confirmation scope. This can be a bit annoying if for example you need to wait for a mock to be called and completed, but such mock is independent of awaiting a specific publisher from the viewModel or something else, an example of this is `pollDeleted` test.
- You can name the tests as you wish. Not required but is cool.

### DIFF
--- a/UnitTests/Sources/AnalyticsTests.swift
+++ b/UnitTests/Sources/AnalyticsTests.swift
@@ -279,7 +279,7 @@ final class AnalyticsTests {
     func shouldNotReportIfNotStarted() throws {
         // Given a client with user properties set
         let client = PostHogAnalyticsClient(posthogFactory: MockPostHogFactory(mock: posthogMock))
-    
+        
         // No call to start
         
         client.screen(AnalyticsEvent.MobileScreen(durationMs: nil, screenName: .Home))
--- a/UnitTests/Sources/AppLock/AppLockSetupPINScreenViewModelTests.swift
+++ b/UnitTests/Sources/AppLock/AppLockSetupPINScreenViewModelTests.swift
@@ -20,13 +20,6 @@ final class AppLockSetupPINScreenViewModelTests {
         viewModel.context
     }
     
-    private func setup(mode: AppLockSetupPINScreenMode) {
-        AppSettings.resetAllSettings()
-        keychainController = KeychainControllerMock()
-        appLockService = AppLockService(keychainController: keychainController, appSettings: AppSettings())
-        viewModel = AppLockSetupPINScreenViewModel(initialMode: mode, isMandatory: false, appLockService: appLockService)
-    }
-    
     deinit {
         AppSettings.resetAllSettings()
     }
@@ -190,5 +183,14 @@ final class AppLockSetupPINScreenViewModelTests {
         #expect(context.viewState.isSubtitleWarning, "The subtitle should continue showing a warning.")
         #expect(context.alertInfo?.id == .forceLogout, "An alert should be shown about a force logout.")
         #expect(context.viewState.isLoggingOut, "The view should become disabled.")
+    }
+    
+    // MARK: - Helpers
+    
+    private func setup(mode: AppLockSetupPINScreenMode) {
+        AppSettings.resetAllSettings()
+        keychainController = KeychainControllerMock()
+        appLockService = AppLockService(keychainController: keychainController, appSettings: AppSettings())
+        viewModel = AppLockSetupPINScreenViewModel(initialMode: mode, isMandatory: false, appLockService: appLockService)
     }
 }
--- a/UnitTests/Sources/AuthenticationStartScreenViewModelTests.swift
+++ b/UnitTests/Sources/AuthenticationStartScreenViewModelTests.swift
@@ -13,7 +13,7 @@ import UIKit
 
 @MainActor
 @Suite
-struct AuthenticationStartScreenViewModelTests {
+final class AuthenticationStartScreenViewModelTests {
     var clientFactory: AuthenticationClientFactoryMock!
     var client: ClientSDKMock!
     var appSettings: AppSettings!
@@ -31,8 +31,12 @@ struct AuthenticationStartScreenViewModelTests {
         // ServiceLocator, the providers override that we apply will break other tests in the suite.
     }
     
+    deinit {
+        AppSettings.resetAllSettings()
+    }
+    
     @Test
-    mutating func initialState() async throws {
+    func initialState() async throws {
         // Given a view model that has no provisioning parameters.
         setupViewModel()
         #expect(authenticationService.homeserver.value.loginMode == .unknown)
@@ -59,7 +63,7 @@ struct AuthenticationStartScreenViewModelTests {
     }
     
     @Test
-    mutating func provisionedOIDCState() async throws {
+    func provisionedOIDCState() async throws {
         // Given a view model that has been provisioned with a server that supports OIDC.
         setupViewModel(provisioningParameters: .init(accountProvider: "company.com", loginHint: "user@company.com"))
         #expect(authenticationService.homeserver.value.loginMode == .unknown)
@@ -79,7 +83,7 @@ struct AuthenticationStartScreenViewModelTests {
     }
     
     @Test
-    mutating func provisionedPasswordState() async throws {
+    func provisionedPasswordState() async throws {
         // Given a view model that has been provisioned with a server that does not support OIDC.
         setupViewModel(provisioningParameters: .init(accountProvider: "company.com", loginHint: "user@company.com"), supportsOIDC: false)
         #expect(authenticationService.homeserver.value.loginMode == .unknown)
@@ -97,7 +101,7 @@ struct AuthenticationStartScreenViewModelTests {
     }
     
     @Test
-    mutating func singleProviderOIDCState() async throws {
+    func singleProviderOIDCState() async throws {
         // Given a view model that for an app that only allows the use of a single provider that supports OIDC.
         setAllowedAccountProviders(["company.com"])
         setupViewModel()
@@ -118,7 +122,7 @@ struct AuthenticationStartScreenViewModelTests {
     }
     
     @Test
-    mutating func singleProviderPasswordState() async throws {
+    func singleProviderPasswordState() async throws {
         // Given a view model that for an app that only allows the use of a single provider that does not support OIDC.
         setAllowedAccountProviders(["company.com"])
         setupViewModel(supportsOIDC: false)
@@ -138,7 +142,7 @@ struct AuthenticationStartScreenViewModelTests {
     
     // MARK: - Helpers
     
-    private mutating func setupViewModel(provisioningParameters: AccountProvisioningParameters? = nil, supportsOIDC: Bool = true) {
+    private func setupViewModel(provisioningParameters: AccountProvisioningParameters? = nil, supportsOIDC: Bool = true) {
         // Manually create a configuration as the default homeserver address setting is immutable.
         client = ClientSDKMock(configuration: .init(oidcLoginURL: supportsOIDC ? "https://account.company.com/authorize" : nil,
                                                     supportsOIDCCreatePrompt: false,
@@ -162,7 +166,7 @@ struct AuthenticationStartScreenViewModelTests {
         viewModel.context.send(viewAction: .updateWindow(UIWindow()))
     }
     
-    private mutating func setAllowedAccountProviders(_ providers: [String]) {
+    private func setAllowedAccountProviders(_ providers: [String]) {
         appSettings.override(accountProviders: providers,
                              allowOtherAccountProviders: false,
                              hideBrandChrome: false,
--- a/UnitTests/Sources/BugReportServiceTests.swift
+++ b/UnitTests/Sources/BugReportServiceTests.swift
@@ -30,12 +30,12 @@ final class BugReportServiceTests {
     deinit {
         appSettings.bugReportRageshakeURL.reset()
     }
-
+    
     @Test
     func initialStateWithMockService() {
         #expect(!bugReportService.crashedLastRun)
     }
-
+    
     @Test
     func submitBugReportWithMockService() async throws {
         let bugReport = BugReport(userID: "@mock:client.com",
@@ -85,7 +85,7 @@ final class BugReportServiceTests {
                                        sdkGitSHA: "1234",
                                        session: .mock,
                                        appHooks: AppHooks())
-
+        
         let bugReport = BugReport(userID: "@mock:client.com",
                                   deviceID: nil,
                                   ed25519: nil,
@@ -121,7 +121,7 @@ final class BugReportServiceTests {
         
         appSettings.bugReportRageshakeURL.applyRemoteValue(.url("https://bugs.server.net/submit"))
         #expect(service.isEnabled)
-
+        
         let bugReport = BugReport(userID: "@mock:client.com",
                                   deviceID: nil,
                                   ed25519: nil,
@@ -158,15 +158,15 @@ private class MockURLProtocol: URLProtocol {
             client?.urlProtocolDidFinishLoading(self)
         }
     }
-
+    
     override func stopLoading() {
         //  no-op
     }
-
+    
     override class func canonicalRequest(for request: URLRequest) -> URLRequest {
         request
     }
-
+    
     override class func canInit(with request: URLRequest) -> Bool {
         true
     }
--- a/UnitTests/Sources/DateTests.swift
+++ b/UnitTests/Sources/DateTests.swift
@@ -16,7 +16,7 @@ struct DateTests {
     var startOfToday: Date {
         Calendar.current.startOfDay(for: .now)
     }
-
+    
     var startOfYesterday: Date {
         // swiftlint: disable:next force_unwrapping
         Calendar.current.startOfDay(for: Calendar.current.date(byAdding: .day, value: -1, to: .now)!)
--- a/UnitTests/Sources/EmojiPickerScreenViewModelTests.swift
+++ b/UnitTests/Sources/EmojiPickerScreenViewModelTests.swift
@@ -35,7 +35,7 @@ struct EmojiPickerScreenViewModelTests {
                 #expect(toggledReaction == reaction)
                 return .success(())
             }
-                
+            
             context.send(viewAction: .emojiTapped(emoji: .init(id: "wave", value: reaction)))
             
             try await deferred.fulfill()
--- a/UnitTests/Sources/HomeScreenViewModelTests.swift
+++ b/UnitTests/Sources/HomeScreenViewModelTests.swift
@@ -60,7 +60,7 @@ final class HomeScreenViewModelTests {
         #expect(correctResult)
         #expect(mockRoomID == selectedRoomID)
     }
-
+    
     @Test
     func tapUserAvatar() async {
         setupViewModel()
@@ -111,7 +111,7 @@ final class HomeScreenViewModelTests {
         room.leaveRoomClosure = { .failure(.sdkError(ClientProxyMockError.generic)) }
         
         clientProxy.roomForIdentifierClosure = { _ in .joined(room) }
-
+        
         let deferred = deferFulfillment(context.$viewState) { value in
             value.bindings.alertInfo != nil
         }
@@ -119,7 +119,7 @@ final class HomeScreenViewModelTests {
         context.send(viewAction: .confirmLeaveRoom(roomIdentifier: mockRoomID))
         
         try await deferred.fulfill()
-                
+        
         #expect(context.alertInfo != nil)
     }
     
--- a/UnitTests/Sources/LocalizationTests.swift
+++ b/UnitTests/Sources/LocalizationTests.swift
@@ -15,66 +15,66 @@ final class LocalizationTests {
     deinit {
         Bundle.overrideLocalizations = nil
     }
-
+    
     /// Test ElementL10n considers app language changes
     @Test
     func appLanguage() {
         // set app language to English
         Bundle.overrideLocalizations = ["en"]
-
+        
         #expect(L10n.testLanguageIdentifier == "en")
-
+        
         // set app language to Italian
         Bundle.overrideLocalizations = ["it"]
-
+        
         #expect(L10n.testLanguageIdentifier == "it")
     }
-
+    
     /// Test fallback language for a language not supported at all
     @Test
     func fallbackOnNotSupportedLanguage() {
         //  set app language to something Element don't support at all (chose non existing identifier)
         Bundle.overrideLocalizations = ["xx"]
-
+        
         #expect(L10n.testLanguageIdentifier == "en")
     }
-
+    
     /// Test fallback language for a language supported but poorly translated
     @Test
     func fallbackOnNotTranslatedKey() {
         //  set app language to something Element supports but use a key that is not translated (we have a key that should never be translated)
         Bundle.overrideLocalizations = ["it"]
-
+        
         #expect(L10n.testLanguageIdentifier == "it")
         #expect(L10n.testUntranslatedDefaultLanguageIdentifier == "en")
     }
-
+    
     /// Test plurals that ElementL10n considers app language changes
     @Test
     func plurals() {
         //  set app language to English
         Bundle.overrideLocalizations = ["en"]
-
+        
         #expect(L10n.commonMemberCount(1) == "1 Member")
         #expect(L10n.commonMemberCount(2) == "2 Members")
-
+        
         //  set app language to Italian
         Bundle.overrideLocalizations = ["it"]
-
+        
         #expect(L10n.commonMemberCount(1) == "1 Membro")
         #expect(L10n.commonMemberCount(2) == "2 Membri")
     }
-
+    
     /// Test plurals fallback language for a language not supported at all
     @Test
     func pluralsFallbackOnNotSupportedLanguage() {
         //  set app language to something Element don't support at all ("invalid identifier")
         Bundle.overrideLocalizations = ["xx"]
-
+        
         #expect(L10n.commonMemberCount(1) == "1 Member")
         #expect(L10n.commonMemberCount(2) == "2 Members")
     }
-
+    
     /// Test untranslated strings
     @Test
     func untranslated() {
--- a/UnitTests/Sources/LoggingTests.swift
+++ b/UnitTests/Sources/LoggingTests.swift
@@ -41,7 +41,7 @@ final class LoggingTests {
         
         #expect(try String(contentsOf: logFile, encoding: .utf8).contains(infoLog))
     }
-        
+    
     @Test
     func logLevels() throws {
         try setupTest()
@@ -108,7 +108,7 @@ final class LoggingTests {
         #expect(!content.contains(lastMessage))
         #expect(!content.contains(heroName))
     }
-        
+    
     @Test
     func timelineContentIsRedacted() throws {
         try setupTest()
@@ -204,7 +204,7 @@ final class LoggingTests {
         #expect(content.contains(fileMessage.id.uniqueID.value))
         #expect(!content.contains(fileMessage.body))
     }
-        
+    
     @Test
     func rustMessageContentIsRedacted() throws {
         try setupTest()
@@ -248,7 +248,7 @@ final class LoggingTests {
         
         // Then the log file should not include the text content
         let logFile = try #require(Tracing.logFiles.first)
-
+        
         let content = try String(contentsOf: logFile, encoding: .utf8)
         #expect(content.contains(String(describing: TextMessageContent.self)))
         #expect(!content.contains(textString))
--- a/UnitTests/Sources/RoomMembersFlowCoordinatorTests.swift
+++ b/UnitTests/Sources/RoomMembersFlowCoordinatorTests.swift
@@ -15,17 +15,17 @@ struct RoomMembersFlowCoordinatorTests {
     var membersFlowCoordinator: RoomMembersFlowCoordinator!
     var navigationStackCoordinator: NavigationStackCoordinator!
     var stateMachineFactory: PublishedStateMachineFactory!
-
+    
     @Test
     mutating func clearRoute() async throws {
         try await setup(entryPoint: .roomMembersList)
         #expect(navigationStackCoordinator.stackCoordinators.last is RoomMembersListScreenCoordinator)
-
+        
         var membersFlowStateExpectation = deferFulfillment(stateMachineFactory.membersFlowStatePublisher) { $0 == .roomMemberDetails(userID: "test", previousState: .roomMembersList) }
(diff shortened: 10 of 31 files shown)


UnitTests/Sources/AppLock/AppLockScreenViewModelTests.swift
```diff
@@ -7,32 +7,34 @@
...
-import XCTest
+import Testing
 
 @MainActor
-class AppLockScreenViewModelTests: XCTestCase {
-    var appSettings: AppSettings!
-    var appLockService: AppLockService!
-    var keychainController: KeychainControllerMock!
-    var viewModel: AppLockScreenViewModelProtocol!
+@Suite(.serialized)
```
pixlwave: Do you think we should remove the `.serialized` traits on all the suites whilst we've got `parallelizable` disabled in the test plan?

UnitTests/Sources/AppLock/AppLockTimerTests.swift
```diff
@@ -7,187 +7,187 @@
...
 
 @testable import ElementX
-import XCTest
+import Foundation
+import Testing
 
-class AppLockTimerTests: XCTestCase {
-    var timer: AppLockTimer!
+@Suite
+struct AppLockTimerTests {
```
pixlwave, line 14: I think we should decide on when to use a struct vs a class or simply pick one or the other for consistency. Maybe for this initial migration we should use classes everywhere so we don't change as much at once, and not worry about mutating methods?

UnitTests/Sources/AppLock/AppLockTimerTests.swift
```diff
@@ -7,187 +7,187 @@
...
-        gracePeriod * 10
+    
+    private func gracePeriodX2(for timer: AppLockTimer) -> TimeInterval {
+        timer.gracePeriod * 2
     }
     
-    override func tearDown() {
-        timer = nil
+    private func gracePeriodX10(for timer: AppLockTimer) -> TimeInterval {
+        timer.gracePeriod * 10
```
pixlwave, line 31: Sticking with classes for the migration would also mean that changes like this aren't necessary.

UnitTests/Sources/AppLock/AppLockTimerTests.swift
Velin92: Mutating methods are fine, structs are preferred for speed (ofc talking about microseconds here) and are the default `@Suite` object when making a new test, but classes are required when you need to use a deinit (tearDown).
I have given the final touches to the code myself (not AI) to always use structs when it was allowed, and forced classes when it was the only possible solution to avoid using a `defer { }` at the end of each test

UnitTests/Sources/AppLock/AppLockTimerTests.swift
Velin92: I am pretty sure this can be improved while also keeping the struct

UnitTests/Sources/AnalyticsSettingsScreenViewModelTests.swift
```diff
@@ -7,23 +7,16 @@
...
     private var viewModel: AnalyticsSettingsScreenViewModelProtocol!
     private var context: AnalyticsSettingsScreenViewModelType.Context!
     
-    override func setUp() {
-        AppSettings.resetAllSettings()
-    }
-    
-    override func tearDown() {
-        AppSettings.resetAllSettings()
-    }
```
pixlwave, line 24: This one should be in the deinit.

UnitTests/Sources/AnalyticsTests.swift
```diff
@@ -29,93 +30,98 @@ class AnalyticsTests: XCTestCase {
         posthogMock.configureMockBehavior()
     }
     
-    override func tearDown() {
-        AppSettings.resetAllSettings()
-    }
```
pixlwave, line 34: And this one.

UnitTests/Sources/AnalyticsSettingsScreenViewModelTests.swift
Velin92: good catch!

UnitTests/Sources/BugReportServiceTests.swift
```diff
@@ -25,16 +26,14 @@ class BugReportServiceTests: XCTestCase {
         bugReportServiceMock.submitBugReportProgressListenerReturnValue = .success(SubmitBugReportResponse(reportURL: "https://www.example.com/123"))
         bugReportService = bugReportServiceMock
     }
-    
-    override func tearDown() {
-        appSettings.bugReportRageshakeURL.reset()
-    }
```
pixlwave, line 31: Another deinit here.

UnitTests/Sources/DateTests.swift
```diff
@@ -7,55 +7,61 @@
...
-    let startOfToday = Calendar.current.startOfDay(for: .now)
-    let startOfYesterday = Calendar.current.startOfDay(for: Calendar.current.date(byAdding: .day, value: -1, to: .now)!)
+    var startOfToday: Date {
+        Calendar.current.startOfDay(for: .now)
+    }
+
+    var startOfYesterday: Date {
+        // swiftlint: disable:next force_unwrapping
+        Calendar.current.startOfDay(for: Calendar.current.date(byAdding: .day, value: -1, to: .now)!)
+    }
```
pixlwave, line 23: I miss single line computed vars 😢

UnitTests/Sources/DeactivateAccountScreenViewModelTests.swift
```diff
@@ -7,75 +7,69 @@
...
     }
     
-    func validateDeactivate(erasingData shouldErase: Bool) async throws {
+    mutating func validateDeactivate(erasingData shouldErase: Bool) async throws {
         let enteredPassword = UUID().uuidString
         
-        clientProxy.deactivateAccountPasswordEraseDataClosure = { [weak self] password, eraseData in
-            guard let self else { return .failure(.sdkError(ClientProxyMockError.generic)) }
+        clientProxy.deactivateAccountPasswordEraseDataClosure = { [clientProxy] password, eraseData in
+            guard let clientProxy else { return .failure(.sdkError(ClientProxyMockError.generic)) }
```
pixlwave: Errrr, this doesn't look good. That's retain cycle on clientProxy right?

UnitTests/Sources/DeferredFulfillmentTests.swift
```diff
@@ -7,13 +7,16 @@
 //
 
 @testable import ElementX
-import XCTest
+import Observation
```
pixlwave, line 10: Do we actually need this import, the tests aren't using observation directly at all.

UnitTests/Sources/DeferredFulfillmentTests.swift
```diff
@@ -22,35 +25,38 @@ class DeferredFulfillmentTests: XCTestCase {
...
+        #expect(observable.counter == 0)
         
         // Then the test should be fulfilled once the update has taken place.
         try await deferred.fulfill()
-        XCTAssertEqual(observable.counter, newValue)
+        #expect(observable.counter == newValue)
     }
     
-    func testObservableMultipleUpdates() async throws {
+    @Test(.timeLimit(.minutes(2)))
```
pixlwave: lol this test should take at most 2-3 seconds, we don't need a custom timeout on it 😅

UnitTests/Sources/ElementCallServiceTests.swift
```diff
@@ -8,75 +8,81 @@
...
-        XCTAssertFalse(callProvider.reportNewIncomingCallWithUpdateCompletionCalled)
-        let expectation = XCTestExpectation(description: "Call accepted")
-        
-        let pushPayload = PKPushPayloadMock().updatingExpiration(currentDate, lifetime: 20)
-        
-        service.pushRegistry(pushRegistry,
-                             didReceiveIncomingPushWith: pushPayload,
-                             for: .voIP) {
-            expectation.fulfill()
+    @Test
```
pixlwave, line 55: Noting that this one used to be disabled, but maybe `confirmation` helps fix it.

UnitTests/Sources/ElementCallServiceTests.swift
```diff
@@ -87,45 +93,31 @@ class ElementCallServiceTests: XCTestCase {
...
-    func disabled_testLifetimeIsCapped() async throws {
-        setupService()
-        
-        let expectation = expectation(description: "Call has ended unanswered")
-        callProvider.reportCallWithEndedAtReasonClosure = { _, _, reason in
-            if reason == .unanswered {
-                expectation.fulfill()
-            } else {
-                XCTFail("Call should have ended as unanswered")
+    @Test
```
pixlwave, line 99: Same on this.

UnitTests/Sources/EmojiPickerScreenViewModelTests.swift
```diff
@@ -7,36 +7,50 @@
...
+                
+            context.send(viewAction: .emojiTapped(emoji: .init(id: "wave", value: reaction)))
+            
+            try await deferred.fulfill()
+            
+            // Since the reaction is called asynchronously after dismissing the picker
+            // We need to actively wait for the function to be called before fulfilling the test.
+            while !toggleReactionCalled {
+                await Task.yield()
+            }
```
pixlwave, line 47: Can't say I'm a fan of this…

Velin92: Sadly this is how confirmation works... from what I understand the old `expectation` in XCTest, kinda does the same thing, the polling every x time is just hidden in the implementation.
Confirmation doesn't it just assumes that you code will just await something that will be completed once the confirmation has been called.
However this creates some problems in cases where we want to verify that the mock has been called with a certain value but, the mock call doesn't reflect a change in the viewModel.

We have VERY few cases of this, so it's not going to happen a lot of times, but I wonder if we should either implement another way to solve this or just have a way in the mock to be listened when is called.

UnitTests/Sources/ElementCallServiceTests.swift
Velin92: yeah I tried with confirmation and seems to have improved it! but will require some more CI runs to see if is more stable.

UnitTests/Sources/EmojiProviderTests.swift
```diff
@@ -7,11 +7,13 @@
 //
 
 @testable import ElementX
-import XCTest
+import Testing
+import UIKit
```
pixlwave, line 11: Why are we suddenly importing UIKit? 😅

UnitTests/Sources/HomeScreenViewModelTests.swift
```diff
@@ -24,19 +25,14 @@ class HomeScreenViewModelTests: XCTestCase {
...
-        
+    init() {
         AppSettings.resetAllSettings()
         appSettings = AppSettings()
         ServiceLocator.shared.register(appSettings: appSettings)
     }
     
-    override func tearDown() {
-        AppSettings.resetAllSettings()
-    }
```
pixlwave, line 37: This one too

UnitTests/Sources/LocalizationTests.swift
```diff
@@ -7,72 +7,79 @@
...
+import Foundation
+import Testing
 
-class LocalizationTests: XCTestCase {
-    override func tearDown() {
-        super.tearDown()
+@Suite(.serialized)
+struct LocalizationTests {
+    init() {
         Bundle.overrideLocalizations = nil
```
pixlwave, line 16: Errr, this was in `tearDown` and has now been put in `init` 🙈. This probably explains the random language failures then 😅

UnitTests/Sources/LoggingTests.swift
```diff
@@ -7,15 +7,17 @@
...
 
-class LoggingTests: XCTestCase {
+@Suite(.serialized)
+struct LoggingTests {
     private enum Constants {
         static let genericFailure = "Test failed"
     }
     
-    override func tearDown() async throws {
+    init() async throws {
```
pixlwave: And here… 😬

UnitTests/Sources/LoggingTests.swift
```diff
@@ -24,48 +26,43 @@ class LoggingTests: XCTestCase {
...
+        let logFile = try #require(Tracing.logFiles.first)
         
-        try XCTAssertFalse(String(contentsOf: logFile, encoding: .utf8).contains(verboseLog))
+        #expect(try !String(contentsOf: logFile, encoding: .utf8).contains(verboseLog))
     }
     
     /// This is meant to test the `Target.tests.configure(…)`, but at this stage the test is somewhat pointless
-    /// as it is unlikely to have been called before `tearDown` has manually set the file prefix 😕.
-    func testTargetName() {
+    /// as it is unlikely to have been called before the init has manually set the file prefix 😕.
```
pixlwave: It even updated the docs 😂

UnitTests/Sources/NavigationTabCoordinatorTests.swift
```diff
@@ -102,90 +109,93 @@ class NavigationTabCoordinatorTests: XCTestCase {
...
-        expectation.isInverted = true
+        var callbackCalled = false
         navigationTabCoordinator.setOverlayCoordinator(overlayCoordinator) {
-            expectation.fulfill()
+            callbackCalled = true
         }
         
         navigationTabCoordinator.setOverlayPresentationMode(.minimized)
-        waitForExpectations(timeout: 1.0)
+        try await Task.sleep(for: .seconds(0.1))
```
pixlwave: Let's not change the timeout duration here :)

UnitTests/Sources/PollFormScreenViewModelTests.swift
```diff
@@ -7,168 +7,195 @@
...
-        XCTAssertEqual(context.options.count, 2)
-        XCTAssertTrue(context.options.allSatisfy(\.text.isEmpty))
-        XCTAssertTrue(context.question.isEmpty)
-        XCTAssertTrue(context.viewState.isSubmitButtonDisabled)
-        XCTAssertFalse(context.viewState.bindings.isUndisclosed)
+    @Test
+    func newPollInitialState() async throws {
+        #expect(context.options.count == 2)
+        let isEmpty = context.options.allSatisfy(\.text.isEmpty)
+        #expect(isEmpty)
```
pixlwave, line 28: I think this is simple enough not to need to have been split up into 2 statements.

UnitTests/Sources/PollFormScreenViewModelTests.swift
pixlwave, line 1: This file has a mixture of using the `viewModel` property and `let viewModel = makeViewModel(…)`. We should pick one or the other rather than mixing them.

pixlwave: Also if we keep makeViewModel, it should go at the bottom and put back the `// MARK: - Helpers` mark.

UnitTests/Sources/PollFormScreenViewModelTests.swift
```diff
@@ -7,168 +7,195 @@
...
-                                            timelineController: MockTimelineController(timelineProxy: timelineProxy),
-                                            analytics: ServiceLocator.shared.analytics,
-                                            userIndicatorController: UserIndicatorControllerMock())
+        try await confirmation { confirmation in
+            var redactReasonCalled = false
+            timelineProxy.redactReasonClosure = { eventID, _ in
+                defer {
+                    confirmation()
+                    redactReasonCalled = true
+                }
```
pixlwave, line 171: Is the `defer` actually necessary here?

UnitTests/Sources/ReportRoomScreenViewModelTests.swift
```diff
@@ -7,96 +7,102 @@
...
-        
-        let leaveExpectation = XCTestExpectation(description: "Leave success")
-        roomProxy.leaveRoomClosure = {
-            defer { leaveExpectation.fulfill() }
-            return .success(())
-        }
         
-        let deferred = deferFulfillment(viewModel.actionsPublisher) { action in
-            action == .dismiss(shouldLeaveRoom: true)
+        try await confirmation(expectedCount: 2) { confirmation in
```
pixlwave, line 59: I would suggest that this one is different to the previous implementation as there's nothing to detect that e.g. leaveRoom was called twice and reportRoom never called.

UnitTests/Sources/ReportRoomScreenViewModelTests.swift
```diff
@@ -7,96 +7,102 @@
...
-            defer { reportExpectation.fulfill() }
-            XCTAssertEqual(reasonArgument, reason)
-            return .success(())
-        }
         
-        let leaveExpectation = XCTestExpectation(description: "Leave fails")
-        roomProxy.leaveRoomClosure = {
-            defer { leaveExpectation.fulfill() }
-            return .failure(.eventNotFound)
+        try await confirmation(expectedCount: 2) { confirmation in
```
pixlwave, line 90: Same comment

UnitTests/Sources/RoomListFiltersStateTests.swift
```diff
@@ -7,121 +7,123 @@
...
-    var allCasesWithoutLowPriority = RoomListFilter.allCases.filter { $0 != .lowPriority }
-    
-    override func setUp() {
+    init() {
         AppSettings.resetAllSettings()
         appSettings = AppSettings()
         state = RoomListFiltersState(appSettings: appSettings)
     }
     
-    override func tearDown() {
```
pixlwave, line 24: This one is missing

UnitTests/Sources/URLComponentsTests.swift
```diff
@@ -7,87 +7,73 @@
...
+    @Test
+    func removeFragmentQueryItem() throws {
+        let url = try #require(URL(string: "https://test.matrix.org#random/data?first=1&second=2"))
+        var components = try #require(URLComponents(url: url, resolvingAgainstBaseURL: true))
                 
-        XCTAssertNotNil(components.fragmentQueryItems)
-        guard var fragmentQueryItems = components.fragmentQueryItems else {
-            return
-        }
+        #expect(components.fragmentQueryItems != nil)
```
pixlwave: This is probably redundant now we have the `try #require`.

UnitTests/Sources/URLComponentsTests.swift
```diff
@@ -7,87 +7,73 @@
...
+    @Test
+    func appendFragmentQueryItem() throws {
+        let url = try #require(URL(string: "https://test.matrix.org#/random/data?first=1&second=2"))
+        var components = try #require(URLComponents(url: url, resolvingAgainstBaseURL: true))
                 
-        XCTAssertNotNil(components.fragmentQueryItems)
-        guard var fragmentQueryItems = components.fragmentQueryItems else {
-            return
-        }
+        #expect(components.fragmentQueryItems != nil)
```
pixlwave: And here

UnitTests/Sources/URLComponentsTests.swift
```diff
@@ -7,87 +7,73 @@
...
-        XCTAssertNotNil(components.fragmentQueryItems)
-        guard var fragmentQueryItems = components.fragmentQueryItems else {
-            return
-        }
+    @Test
+    func changeFragmentQueryItemValue() throws {
+        let url = try #require(URL(string: "https://test.matrix.org#/random/data?first=1&second=2"))
+        var components = try #require(URLComponents(url: url, resolvingAgainstBaseURL: true))
+        
+        #expect(components.fragmentQueryItems != nil)
```
pixlwave: And here

UnitTests/Sources/UserAgentBuilderTests.swift
```diff
@@ -7,20 +7,24 @@
...
+import Testing
 
-class UserAgentBuilderTests: XCTestCase {
-    func testIsNotNil() {
-        XCTAssertNotNil(UserAgentBuilder.makeASCIIUserAgent())
+@Suite
+struct UserAgentBuilderTests {
+    @Test
+    func isNotUnknow() {
+        #expect(UserAgentBuilder.makeASCIIUserAgent() != "unknown")
```
pixlwave, line 16: Errr, unless I'm missing changes elsewhere, the meaning of this one has been changed. It should be

```suggestion
    func isNotNil() {
        #expect(UserAgentBuilder.makeASCIIUserAgent() != nil)
```

UnitTests/Sources/UserIndicatorControllerTests.swift
```diff
@@ -8,87 +8,91 @@
...
         
         indicatorController.submitIndicator(.init(id: "First", title: ""))
         indicatorController.submitIndicator(.init(id: "Second", title: ""))
         indicatorController.submitIndicator(.init(id: "Third", title: ""))
         
-        XCTAssertEqual(indicatorController.activeIndicator?.id, "Third")
+        #expect(indicatorController.activeIndicator?.id == "Third")
         
-        let fulfillment = deferFulfillment(indicatorController.$activeIndicator, message: "Waiting for last indicator to be dismissed") { indicator in
+        let fulfillment = deferFulfillment(indicatorController.$activeIndicator) { indicator in
```
pixlwave: I feel like I would prefer that we kept the message parameter on the new implementation, even if we don't use it just so that we can implement something in the future and don't have to remove these.

UnitTests/Sources/AppLock/AppLockTimerTests.swift
Velin92: done!

UnitTests/Sources/DeactivateAccountScreenViewModelTests.swift
Velin92: Claude what are you doing?!

UnitTests/Sources/DeferredFulfillmentTests.swift
Velin92: there is an `@Observable` at the bottom of file

UnitTests/Sources/EmojiProviderTests.swift
Velin92: Claud what are you doing?! Oh no wait, this is needed, because there is a `UIWindow()` in this test, before `XCTest` imported also `UIKit`, but `Testing` doesn't (XCTest also included Foundation which `Testing` doesn't, in fact there is also a lot of `Foundation` imports if you see)

UnitTests/Sources/NavigationTabCoordinatorTests.swift
Velin92: actually we can solve this with a confirmation

UnitTests/Sources/BugReportServiceTests.swift
Velin92: done

UnitTests/Sources/PollFormScreenViewModelTests.swift
Velin92: oh wait... so this is a Swift Testing bug that I fixed manually
For some reason when putting this in an #expect it believes that it's a statement that requires a try (I think is because the #expect influences the allSatisfy generic and makes it believe it requires a `try`).
If you put the try outside you get a compilation error, but if you put inside the expect you get a warning because the `try` is not required.
Fun.

UnitTests/Sources/PollFormScreenViewModelTests.swift
Velin92: It guarantees that this is run after the return. It just felt right to put in the defer

UnitTests/Sources/ReportRoomScreenViewModelTests.swift
Velin92: Yeah I see what you mean.

We should add:
```
#expect(roomProxy.reportRoomReasonCalled)
#expect(roomProxy.leaveRoomCalled)
```

After the confirmation has been awaited

UnitTests/Sources/UserAgentBuilderTests.swift
Velin92: My fix, in reality `makeASCIIUserAgent` is not optional and can never be `nil` (it returns unknown when the fetched value is not available), XCTestCase allowed for the comparison and did not complain, but the `#expect` warned about this always passing. This is the correct test.

UnitTests/Sources/UserIndicatorControllerTests.swift
Velin92: Added the message back as a part of the error!
Restored some of this, but if I have missed any please mark them.

compound-ios/Tests/CompoundTests/PreviewTests.swift
pixlwave, line 1: Interesting – I guess these Compound snapshots actually run fine in parallel then given that there's no dependencies 👏

UnitTests/Sources/AppLock/AppLockSetupPINScreenViewModelTests.swift
```diff
@@ -19,74 +20,81 @@ class AppLockSetupPINScreenViewModelTests: XCTestCase {
         viewModel.context
     }
     
-    override func setUp() {
+    private func setup(mode: AppLockSetupPINScreenMode) {
```
pixlwave: Let's put this one at the bottom under a `// MARK - Helpers` perhaps?

UnitTests/Sources/AuthenticationStartScreenViewModelTests.swift
```diff
@@ -22,22 +24,19 @@ class AuthenticationStartScreenViewModelTests: XCTestCase {
...
         AppSettings.resetAllSettings()
         appSettings = AppSettings()
         // These app settings are kept local to the tests on purpose as if they are registered in the
         // ServiceLocator, the providers override that we apply will break other tests in the suite.
     }
     
-    override func tearDown() {
-        AppSettings.resetAllSettings()
-    }
-
```
pixlwave, line 35: Ah, another deinit removed.

UnitTests/Sources/DeferredFulfillmentTests.swift
pixlwave: Doh! 🤦‍♂️

UnitTests/Sources/PollFormScreenViewModelTests.swift
pixlwave: Ha, nice.

UnitTests/Sources/RoomChangeRolesScreenViewModelTests.swift
```diff
@@ -7,149 +7,157 @@
...
-            return
-        }
+    @Test
+    mutating func toggleUserOff() throws {
+        setup(mode: .moderator)
+        let firstUser = try #require(context.viewState.users.first { !context.viewState.isMemberSelected($0) },
+                                     "There should be a regular user available to promote.")
+        
+        // First toggle on
+        context.send(viewAction: .toggleMember(firstUser))
```
pixlwave: I'm curious, is it not possible to call an existing test in Swift Testing or is this just Claude re-writing things for fun? I'm not sure I'm a fan given this way it doesn't actually asset the starting conditions before toggling the user off.

pixlwave: There's a few tests in this file where it has done the same thing.

UnitTests/Sources/RoomMembersFlowCoordinatorTests.swift
```diff
@@ -36,18 +38,20 @@ class RoomMembersFlowCoordinatorTests: XCTestCase {
...
-    private func setUp(entryPoint: RoomMembersFlowCoordinatorEntryPoint) async throws {
+
+    // MARK: - Helpers
+
+    private mutating func setup(entryPoint: RoomMembersFlowCoordinatorEntryPoint) async throws {
         stateMachineFactory = .init()
         navigationStackCoordinator = NavigationStackCoordinator()
         navigationStackCoordinator.setRootCoordinator(PlaceholderScreenCoordinator(hideBrandChrome: false))
         navigationStackCoordinator.push(BlankFormCoordinator())
-
```
pixlwave, line 47: Probably worth re-indenting this file to restore these removed spaces.

pixlwave (approved):
LGTM! 👌

compound-ios/Tests/CompoundTests/PreviewTests.swift
Velin92: yeah which is great, which proves the next point... once this migration is fully completed we need to make AppSettings behave like a protocol (or at least mockable) in a way that is not shared anymore between tests

UnitTests/Sources/RoomChangeRolesScreenViewModelTests.swift
Velin92: I'll give it a try

Velin92: Yeah is Claude req-writing for fun, are there any other places where you spotted this? I'll fix them in this file

Velin92: Okay fixed it, please let me know if you find other instances of this.

UnitTests/Sources/RoomMembersFlowCoordinatorTests.swift
Velin92: AIs tend to do that a lot... that's why I added a swiftformat rule, so that even if it happens without noticing the hook will always fix it for us.

PR #5121, opened 2026-02-18
Voice message: variable play back speed

Medformatik:
## Content

Add playback speed control for voice messages with support for 0.5×, 1×, 1.5×, and 2× playback speeds. The speed button is displayed above the timestamp and cycles through the available speeds when tapped. This PR alignes element-x-ios with element-x-android, where this feature is already in the app. That was PR [element-x-android/#5504](https://github.com/element-hq/element-x-android/pull/5504) and the follow-up PR [element-x-android/#5963](https://github.com/element-hq/element-x-android/pull/5963).

## Motivation and context

Allows users to adjust voice message playback speed for improved accessibility and user experience. This is a common feature in messaging apps that helps users consume voice content more efficiently.

### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Other/VoiceMessage/WaveformInteractionModifier.swift
+++ b/ElementX/Sources/Other/VoiceMessage/WaveformInteractionModifier.swift
@@ -41,6 +41,7 @@ private struct WaveformInteractionModifier: ViewModifier {
                                 onSeek(max(0, min(progress, 1.0)))
                             })
                         .offset(x: -cursorInteractiveSize / 2, y: 0)
+                        .allowsHitTesting(showCursor)
                 }
                 .gesture(SpatialTapGesture()
                     .onEnded { tapGesture in
--- a/ElementX/Sources/Screens/Timeline/TimelineModels.swift
+++ b/ElementX/Sources/Screens/Timeline/TimelineModels.swift
@@ -115,6 +115,7 @@ struct TimelineViewState: BindableState {
     var isViewSourceEnabled: Bool
     var areThreadsEnabled: Bool
     var linkPreviewsEnabled: Bool
+    var voiceMessagePlaybackSpeed = VoiceMessagePlaybackSpeed.default.rawValue
     
     let hasPredecessor: Bool
         
--- a/ElementX/Sources/Screens/Timeline/View/TimelineItemViews/VoiceMessageRoomTimelineView.swift
+++ b/ElementX/Sources/Screens/Timeline/View/TimelineItemViews/VoiceMessageRoomTimelineView.swift
@@ -23,7 +23,7 @@ struct VoiceMessageRoomTimelineView: View {
 }
 
 struct VoiceMessageRoomTimelineContent: View {
-    @Environment(\.timelineContext) private var context
+    @EnvironmentObject private var context: TimelineViewModel.Context
     @State private var resumePlaybackAfterScrubbing = false
     
     let timelineItem: VoiceMessageRoomTimelineItem
@@ -34,31 +34,32 @@ struct VoiceMessageRoomTimelineContent: View {
                                      onPlayPause: onPlaybackPlayPause,
                                      onSeek: { onPlaybackSeek($0) },
                                      onScrubbing: { onPlaybackScrubbing($0) },
+                                     playbackSpeed: context.viewState.voiceMessagePlaybackSpeed,
                                      onPlaybackSpeedChange: onPlaybackSpeedChange)
             .fixedSize(horizontal: false, vertical: true)
     }
 
     private func onPlaybackPlayPause() {
-        context?.send(viewAction: .handleAudioPlayerAction(.playPause(itemID: timelineItem.id)))
+        context.send(viewAction: .handleAudioPlayerAction(.playPause(itemID: timelineItem.id)))
     }
 
     private func onPlaybackSpeedChange() {
-        context?.send(viewAction: .handleAudioPlayerAction(.changePlaybackSpeed(itemID: timelineItem.id)))
+        context.send(viewAction: .handleAudioPlayerAction(.changePlaybackSpeed(itemID: timelineItem.id)))
     }
     
     private func onPlaybackSeek(_ progress: Double) {
-        context?.send(viewAction: .handleAudioPlayerAction(.seek(itemID: timelineItem.id, progress: progress)))
+        context.send(viewAction: .handleAudioPlayerAction(.seek(itemID: timelineItem.id, progress: progress)))
     }
     
     private func onPlaybackScrubbing(_ dragging: Bool) {
         if dragging {
             if playerState.playbackState == .playing {
                 resumePlaybackAfterScrubbing = true
-                context?.send(viewAction: .handleAudioPlayerAction(.playPause(itemID: timelineItem.id)))
+                context.send(viewAction: .handleAudioPlayerAction(.playPause(itemID: timelineItem.id)))
             }
         } else {
             if resumePlaybackAfterScrubbing {
-                context?.send(viewAction: .handleAudioPlayerAction(.playPause(itemID: timelineItem.id)))
+                context.send(viewAction: .handleAudioPlayerAction(.playPause(itemID: timelineItem.id)))
                 resumePlaybackAfterScrubbing = false
             }
         }
--- a/ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
+++ b/ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
@@ -102,6 +102,7 @@ class TimelineViewModel: TimelineViewModelType, TimelineViewModelProtocol {
                                                        isViewSourceEnabled: appSettings.viewSourceEnabled,
                                                        areThreadsEnabled: appSettings.threadsEnabled,
                                                        linkPreviewsEnabled: appSettings.linkPreviewsEnabled,
+                                                       voiceMessagePlaybackSpeed: appSettings.voiceMessagePlaybackSpeed.rawValue,
                                                        hasPredecessor: roomProxy.predecessorRoom != nil,
                                                        pinnedEventIDs: roomProxy.infoPublisher.value.pinnedEventIDs,
                                                        emojiProvider: emojiProvider,
@@ -541,6 +542,11 @@ class TimelineViewModel: TimelineViewModelType, TimelineViewModelProtocol {
         
         appSettings.$threadsEnabled
             .weakAssign(to: \.state.areThreadsEnabled, on: self)
+            .store(in: &cancellables)
+
+        appSettings.$voiceMessagePlaybackSpeed
+            .map(\.rawValue)
+            .weakAssign(to: \.state.voiceMessagePlaybackSpeed, on: self)
             .store(in: &cancellables)
         
         userSession.clientProxy.timelineMediaVisibilityPublisher
--- a/ElementX/Sources/Screens/Timeline/View/VoiceMessageRoomPlaybackView.swift
+++ b/ElementX/Sources/Screens/Timeline/View/VoiceMessageRoomPlaybackView.swift
@@ -19,6 +19,7 @@ struct VoiceMessageRoomPlaybackView: View {
     let onPlayPause: () -> Void
     let onSeek: (Double) -> Void
     let onScrubbing: (Bool) -> Void
+    let playbackSpeed: Float
     let onPlaybackSpeedChange: () -> Void
 
     var body: some View {
@@ -27,7 +28,7 @@ struct VoiceMessageRoomPlaybackView: View {
                                size: .medium,
                                action: onPlayPause)
             VStack(spacing: 2) {
-                PlaybackSpeedButton(speed: playerState.playbackSpeed, onTap: onPlaybackSpeedChange)
+                PlaybackSpeedButton(speed: playbackSpeed, onTap: onPlaybackSpeedChange)
                 Text(timeLabelContent)
                     .lineLimit(1)
                     .font(.compound.bodyXSSemibold)
@@ -134,6 +135,7 @@ struct VoiceMessageRoomPlaybackView_Previews: PreviewProvider, TestablePreview {
                                      onPlayPause: { },
                                      onSeek: { value in Task { await playerState.updateState(progress: value) } },
                                      onScrubbing: { _ in },
+                                     playbackSpeed: playerState.playbackSpeed,
                                      onPlaybackSpeedChange: { })
             .fixedSize(horizontal: false, vertical: true)
     }


Velin92:
Could you provide some screenshots, or even better, a screen recording of the feature?

ElementX/Resources/Localizations/en.lproj/Localizable.strings
```diff
@@ -18,6 +18,7 @@
 "a11y_page_n" = "Page %1$d";
 "a11y_pause" = "Pause";
 "a11y_paused_voice_message" = "Voice message, duration: %1$@, current position: %2$@";
+"a11y_playback_speed" = "Playback speed %1$@";
```
Velin92: We don't accept translations set directly in our main localizable strings, due to the fact that we systematically sync them with Localazy.

If you need to add new strings for a contribution, and before our product team approves them, we suggest adding them to the Untraslated lproj, which is used for all the temporary non product approved strings, that are then migrated (or improved) to Localazy once we come up with the final copy (if the feature is ofc approved)

ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
```diff
@@ -417,7 +417,17 @@ class TimelineInteractionHandler {
...
+
+    private static let availablePlaybackSpeeds: [Float] = [1.0, 1.5, 2.0, 0.5]
+
+    func changePlaybackSpeed(for itemID: TimelineItemIdentifier) {
+        let currentIndex = appSettings.voiceMessagePlaybackSpeedIndex
+        let nextIndex = (currentIndex + 1) % Self.availablePlaybackSpeeds.count
+        appSettings.voiceMessagePlaybackSpeedIndex = nextIndex
+        let speed = Self.availablePlaybackSpeeds[nextIndex]
+        audioPlayerState(for: itemID)?.setPlaybackSpeed(speed)
+    }
```
Velin92: So the UserDefaults stores a value that is just an index to fetch this array?
This feels quite unsafe and prone to possible crashes.

Medformatik:
> Could you provide some screenshots, or even better, a screen recording of the feature?


https://github.com/user-attachments/assets/56c6173b-bcc2-4335-886f-e3cfcccf1d5a

Looks and works exactly the same as on Android.

I will fix the other issues you mentioned in review.

ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -359,7 +361,10 @@ final class AppSettings {
     
     @UserPreference(key: UserDefaultsKeys.optimizeMediaUploads, defaultValue: true, storageType: .userDefaults(store))
     var optimizeMediaUploads
-    
+
+    @UserPreference(key: UserDefaultsKeys.voiceMessagePlaybackSpeedIndex, defaultValue: 0, storageType: .userDefaults(store))
+    var voiceMessagePlaybackSpeedIndex: Int
```
Velin92: Why in the AppSettings? is the feature supposed to work across the whole app, and also persist between sessions or if the app has been killed?

It could make sense, but the idea of using an Int as index is very dangerous would be better to instead have

```
enum PlaybackSpeed: Float, CaseIterable {
  case `default` = 1.0
  case fast = 1.5
  case  fastest = 2.0
  case slow = 0.5
}
```

Medformatik: This was based on the decision of @bmarty in https://github.com/element-hq/element-x-android/pull/5963#issue-3772544303. I tried to implement the feature 1:1 as it is on Android.

ElementX/Resources/Localizations/en.lproj/Localizable.strings
Medformatik: Fixed in 6b6c26609b700a271aecdde7227e4af7b1ec0184

ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
Velin92: You could now use the value from the enum, to resolve it's position in the allCases, and increase the index by 1 and then use the mod operator % on `allCases.count`. And then set the speed using the value.speedRatio.

This does two things:

- Makes iterating on the cases way safer because, the index is directly taken from the enum, instead of behind decoupled from it, which may cause in a crash if the logic of storing the value changes.
- The value is just an enum we can tweak and change later in how they are ordered.

ElementX/Resources/Localizations/en.lproj/Localizable.strings
Velin92: If this is already on Android, is it possible that this `a11y_playback_speed` (or that is named differently on Android) already exists somewhere? Android and iOS share the same copies

ElementX/Sources/Other/VoiceMessage/PlaybackSpeedButton.swift
```diff
@@ -0,0 +1,47 @@
...
+        HStack(spacing: 8) {
+            PlaybackSpeedButton(speed: 0.5) { }
+            PlaybackSpeedButton(speed: 1.0) { }
+            PlaybackSpeedButton(speed: 1.5) { }
+            PlaybackSpeedButton(speed: 2.0) { }
+        }
+        .padding()
+        .background(Color.gray)
+    }
+}
```
Velin92, line 53: If this is a testable preview, you need to run the preview tests to add this new test. You might probably also need to update some existing screenshots of the existing tests

UnitTests/Sources/MediaPlayerProviderTests.swift
```diff
@@ -39,25 +39,27 @@ class MediaPlayerProviderTests: XCTestCase {
...
+
         let audioPlayerStates = Array(repeating: AudioPlayerState(id: .timelineItemIdentifier(.randomEvent), title: "", duration: 0), count: 10)
         for audioPlayerState in audioPlayerStates {
             mediaPlayerProvider.register(audioPlayerState: audioPlayerState)
             audioPlayerState.attachAudioPlayer(audioPlayer)
             let isAttached = audioPlayerState.isAttached
             XCTAssertTrue(isAttached)
         }
-        
+
```
Velin92: don't clear indentations please

UnitTests/Sources/MediaPlayerProviderTests.swift
```diff
@@ -39,25 +39,27 @@ class MediaPlayerProviderTests: XCTestCase {
...
         mediaPlayerProvider.detachAllStates(except: nil)
         for audioPlayerState in audioPlayerStates {
             let isAttached = audioPlayerState.isAttached
             XCTAssertFalse(isAttached)
         }
     }
-    
+
```
Velin92: same as above

ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
Velin92, line 492: Thanks for spotting this @pixlwave 

Not sure how these would interact with changing the playback speed that is done few lines above

ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
Velin92: Also if the index is not found for the value provided in the app settings (in case they are changed and one is removed in the future), we need to set the value of the speed as .default

ElementX/Sources/Other/VoiceMessage/PlaybackSpeedButton.swift
pixlwave: Please make sure the snapshots are committed using LFS when doing so :)

ElementX/Sources/Services/Audio/Player/AudioPlayer.swift
```diff
@@ -128,7 +129,14 @@ class AudioPlayer: NSObject, AudioPlayerProtocol {
...
         await internalAudioPlayer.seek(to: CMTime(seconds: time, preferredTimescale: 60))
     }
-    
+
+    func setPlaybackSpeed(_ speed: Float) {
+        playbackSpeed = speed
+        if internalAudioPlayer?.rate != 0 {
+            internalAudioPlayer?.rate = speed
+        }
+    }
```
Velin92, line 138: maybe check if the state is `playing` instead of the playback speed? Would that also work?

ElementX/Resources/Localizations/en.lproj/Localizable.strings
Medformatik: > If this is already on Android, is it possible that this `a11y_playback_speed` (or that is named differently on Android) already exists somewhere? Android and iOS share the same copies

I checked Android PR [element-x-android#5963](https://github.com/element-hq/element-x-android/pull/5963): Android does not define a dedicated accessibility label for the playback-speed button, so there is no corresponding shared string key for it there yet.

Medformatik:
I’ve gone through all review comments and addressed each requested change in this PR. I believe all review feedback is now covered. Please re-check, and I’ll follow up on anything else needed.

ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -36,6 +36,15 @@ enum AppBuildType {
...
 /// Store Element specific app settings.
 final class AppSettings {
+    enum VoiceMessagePlaybackSpeed: Float, CaseIterable, Codable {
+        case `default` = 1.0
+        case fast = 1.5
+        case fastest = 2.0
+        case slow = 0.5
+
+        var speedRatio: Float { rawValue }
+    }
```
Velin92: nice but please let's put this at the end of the file outside the AppSettings.

Also no need for speedRatio now that I am thinking about it, rawValue is sufficient.

Velin92 (changes requested):
I found a bug:

https://github.com/user-attachments/assets/3d471502-a766-453f-9bcf-85e065844b3f

As you can see the label doesn't update for each item, but only for the current one, but when playing it updates, which is different from Android

https://github.com/user-attachments/assets/d863f60c-6ee9-466a-abbf-76bacdc591c1

This requires a different approach, one suggestion would be better to store the playbackspeed in the timeline view state, have it listen to the publisher of the value from the app settings, no need for the onPlaybackSpeedChange then since the change in the context will automatically update it.

What the TimelineInteraction handle would then do is just update the value on the app settings which will update the label.

Then when playing we just check for the appSettings or timeline value (whichever is easier). And set the playback speed.

ElementX/Sources/Other/VoiceMessage/PlaybackSpeedButton.swift
```diff
@@ -0,0 +1,47 @@
...
+    let onTap: () -> Void
+
+    var body: some View {
+        Button(action: onTap) {
+            Text(speedLabel)
+                .font(.compound.bodyXSSemibold)
+                .foregroundColor(.compound.iconSecondary)
+                .padding(.horizontal, 8)
+                .padding(.vertical, 2)
+                .background(.compound.bgCanvasDefault, in: RoundedRectangle(cornerRadius: 12))
```
Velin92: I know this is the same as Android, but it's definitely a bug, the size of the button should not change.
However this is an easy fix, just use a ZStack in the button that includes a copy of the Text below the original, but where the content tis 0.0x and that is hidden.
This will make the button always the biggest possible size, that will also be reactive to the font size.

Medformatik:
With commit eec72720666de5e5ab78ffa5588b175136b5c8aa the speed label now correctly updates for each item and I have also addressed your two other review-comments.

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/VoiceMessageRoomTimelineView.swift
```diff
@@ -23,7 +23,7 @@ struct VoiceMessageRoomTimelineView: View {
 }
 
 struct VoiceMessageRoomTimelineContent: View {
-    @Environment(\.timelineContext) private var context
+    @EnvironmentObject private var context: TimelineViewModel.Context
```
Velin92, line 26: Why has this changed? Let's absolutely avoid this

ElementX/Sources/Other/VoiceMessage/WaveformInteractionModifier.swift
```diff
@@ -41,6 +41,7 @@ private struct WaveformInteractionModifier: ViewModifier {
                                 onSeek(max(0, min(progress, 1.0)))
                             })
                         .offset(x: -cursorInteractiveSize / 2, y: 0)
+                        .allowsHitTesting(showCursor)
```
Velin92, line 44: what is this for?

ElementX/Sources/Screens/Timeline/TimelineModels.swift
```diff
@@ -115,6 +115,7 @@ struct TimelineViewState: BindableState {
     var isViewSourceEnabled: Bool
     var areThreadsEnabled: Bool
     var linkPreviewsEnabled: Bool
+    var voiceMessagePlaybackSpeed = VoiceMessagePlaybackSpeed.default.rawValue
```
Velin92, line 118: no need for a default value if you are setting it in the init below

ElementX/Sources/Other/VoiceMessage/WaveformInteractionModifier.swift
Medformatik: This avoids a tap-conflict before playback starts. The cursor view has a larger interactive frame than the visible 2pt marker (for scrubbing), and when progress is at 0 that hit area sits at the leading edge and can intercept taps intended for the speed button. We only need cursor hit-testing when the cursor is actually shown/interactive, so .allowsHitTesting(showCursor) disables that hidden hit area until playback starts.

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/VoiceMessageRoomTimelineView.swift
```diff
@@ -34,31 +34,32 @@ struct VoiceMessageRoomTimelineContent: View {
                                      onPlayPause: onPlaybackPlayPause,
                                      onSeek: { onPlaybackSeek($0) },
                                      onScrubbing: { onPlaybackScrubbing($0) },
+                                     playbackSpeed: context.viewState.voiceMessagePlaybackSpeed,
```
Velin92, line 37: if you are concerned for the optional here (which no worries will always be available) just do `context.viewState.voiceMessagePlaybackSpeed ? PlaybackSpeed.default.rawValue`

Velin92:
No worries I'll take it from here, thanks a lot for you contribution!

Medformatik:
With commit [33d1dfdbb7fb17122e67fafe77faca2f0599729a](https://github.com/Medformatik/element-x-ios/commit/33d1dfdbb7fb17122e67fafe77faca2f0599729a) I addressed the latest review feedback with minimal changes:

- `VoiceMessageRoomTimelineView.swift`
  - Removed `@EnvironmentObject` usage in VoiceMessageRoomTimelineContent.
  - Injected timelineContext from `@Environment(\.timelineContext)` and observed it explicitly as `@ObservedObject` to keep speed-label updates reactive without relying on `@EnvironmentObject`.
- `VoiceMessageMediaEventsTimelineView.swift`
  - Passed the same `timelineContext` explicitly to `VoiceMessageRoomTimelineContent` (and preview wiring), so behavior is consistent.
- `TimelineModels.swift`
  - Removed the inline default for `voiceMessagePlaybackSpeed` and made it a required init field.

About `.allowsHitTesting(showCursor)` in `WaveformInteractionModifier`: this prevents the hidden cursor hit area at progress 0 from intercepting taps intended for the speed button before playback starts.

PR #5124, opened 2026-02-19
Playback speed button

Velin92:
Refinement of the original: https://github.com/element-hq/element-x-ios/pull/5121 authored by @Medformatik 

Alongside the original code it includes:
- Localisation
- Formatted numbers (so that they can change their representation based on the user's region)
- An improved and more efficient way to listen to the app settings change to update the UI

### DIFF
--- a/ElementX/Sources/Application/Settings/AppSettings.swift
+++ b/ElementX/Sources/Application/Settings/AppSettings.swift
@@ -362,8 +362,8 @@ final class AppSettings {
     @UserPreference(key: UserDefaultsKeys.optimizeMediaUploads, defaultValue: true, storageType: .userDefaults(store))
     var optimizeMediaUploads
 
-    @UserPreference(key: UserDefaultsKeys.voiceMessagePlaybackSpeed, defaultValue: VoiceMessagePlaybackSpeed.default, storageType: .userDefaults(store))
-    var voiceMessagePlaybackSpeed: VoiceMessagePlaybackSpeed
+    @UserPreference(key: UserDefaultsKeys.voiceMessagePlaybackSpeed, defaultValue: AudioPlaybackSpeed.default, storageType: .userDefaults(store))
+    var voiceMessagePlaybackSpeed: AudioPlaybackSpeed
 
     /// Whether or not to show a warning on the media caption composer so the user knows
     /// that captions might not be visible to users who are using other Matrix clients.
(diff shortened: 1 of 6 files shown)


ElementX/Sources/Application/Settings/AppSettings.swift
pixlwave: Please could we also:
- Call this type `AudioPlaybackSpeed` given it now basically lives in `AudioPlayerState` etc.
- Move it to it's own file in this same directory similar to `AnalyticsConsentState`, `AppAppearance` etc that are also types used in `AppSettings`.

ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
```diff
@@ -422,14 +422,14 @@ class TimelineInteractionHandler {
         let availableSpeeds = VoiceMessagePlaybackSpeed.allCases
         guard let currentIndex = availableSpeeds.firstIndex(of: appSettings.voiceMessagePlaybackSpeed) else {
             appSettings.voiceMessagePlaybackSpeed = .default
-            audioPlayerState(for: itemID)?.setPlaybackSpeed(VoiceMessagePlaybackSpeed.default.rawValue)
+            audioPlayerState(for: itemID)?.setPlaybackSpeed(VoiceMessagePlaybackSpeed.default)
             return
         }
```
pixlwave: Completely optional: At this stage it seems simpler to add a `var nextSpeed` on the enum that simply switches over self and returns the next value.

ElementX/Sources/Services/Audio/Player/AudioPlayerState.swift
```diff
@@ -58,15 +58,22 @@ class AudioPlayerState: ObservableObject, Identifiable {
...
+         playbackSpeedPublisher: AnyPublisher<VoiceMessagePlaybackSpeed, Never>? = nil) {
         self.id = id
         self.title = title
         self.duration = duration
         self.waveform = waveform ?? EstimatedWaveform(data: [])
         self.progress = progress
+        self.playbackSpeed = playbackSpeed
         playbackState = .stopped
         playerButtonPlaybackState = .stopped
+        playbackSpeedPublisher?.assign(to: &$playbackSpeed)
```
pixlwave, line 76: You are right, this is indeed a cool trick!

pixlwave (approved):

PR #5125, opened 2026-02-19
Add an `AGENTS.md` file to the project.

pixlwave:
Built with the help of Opus. It's hard to know if it goes into too much detail or how well this is going to work, I'm not used to instructing machines via markdown 🙈

I guess we'll see how this goes and either expand or shrink it accordingly.

Additionally includes a CLAUDE.md file that simply imports AGENTS.md:
https://code.claude.com/docs/en/memory#claude-md-imports

Sidenote: Being able to use imports like Claude has e.g. `See @CONTRIBUTING.md`  instead of duplicating things in AGENTS.md seems like it would be a whole lot nicer way of creating this reference, but apparently Xcode uses the agents file 🤷‍♂️

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,6 +1,6 @@
 # AGENTS.md — Element X iOS
 
-> **Audience:** AI coding agents. **Repo:** `element-hq/element-x-ios` — iOS Matrix client (SwiftUI + `matrix-rust-sdk`).
+> **Repo:** `element-hq/element-x-ios` — iOS Matrix client (SwiftUI + `matrix-rust-sdk`).
 
 ---
 
@@ -59,6 +59,8 @@ PRs must meet these rules. Prefer Xcode MCP tools over terminal commands.
 ### Build System
 
 Initial setup: `swift run tools setup-project`
+
+**Git hooks** are installed by `swift run tools setup-project` and run SwiftLint/SwiftFormat on commit — if a hook fails, **do not abandon your changes**, fix the reported issues and recommit.
 
 | Tool | Command | Notes |
 |------|---------|-------|


AGENTS.md
```diff
@@ -0,0 +1,398 @@
...
+- Include screenshots/videos for any visual changes.
+- Keep PRs under 1000 additions when possible; split large changes.
+
+---
+
+## Project Structure
+
+### Build System
+
+- **XcodeGen** generates the Xcode project from `project.yml`. Never edit `*.xcodeproj` directly, but make sure to still commit it.
```
manuroe: Should we give the exact command lines for those functions?
If we are not directive, agents will be creative on the way to call them using sometimes weird params. They will not call them the same way on each session.

AGENTS.md
```diff
@@ -0,0 +1,398 @@
...
+- Create previews for **all main states** a screen/view will be in.
+- Use `PreviewProvider`s (not `#Preview`) — this is how snapshot tests are hooked up:
+  - Adding a second conformance `TestablePreview` indicates the preview should have snapshot and accessibility tests generated.
+
+---
+
+## Pull Request Guidelines
+
+- Do not create conventional-commits, prefer the use of a single sentence instead.
+- Every PR must reference a well-defined issue. Feature/UI PRs **must** have Product team input.
```
manuroe: Do you know it can verify the product input?

AGENTS.md
```diff
@@ -0,0 +1,398 @@
...
+- Cross-platform changes need an [element-meta](https://github.com/element-hq/element-meta) issue.
```
manuroe: How can it know it is a cross-platform change?

AGENTS.md
```diff
@@ -0,0 +1,398 @@
...
+- Apply exactly **one** `pr-` label to categorise the changelog entry (mapping defined in `.github/release.yml`).
+- The PR title becomes the changelog entry — make it descriptive and complete (no "Fixes #…" or conventional-commit prefixes).
+- Include screenshots/videos for any visual changes.
+- Keep PRs under 1000 additions when possible; split large changes.
+
+---
+
+## Project Structure
+
+### Build System
```
manuroe, line 59: Should we mention `swift run tools setup-project` here?

Maybe I am wrong on this. We can suppose the developer has already the setup/

pixlwave: It would theoretically be useful for creating Copilot tasks on the web, but that's assuming they execute in a macOS environment which I have no idea if that's true or not.

pixlwave:
Feedback on the file from Gemini 😆:

Screenshot 2026-02-20 at 5 56 05 pm

I guess it's time to stop iterating and merge this.

PR #5126, opened 2026-02-20
Parallelize Tests

Velin92:


Velin92:
Was a great learning experience but to proceed with this we need:
- To remove `ServiceLocator.shared` entirely
- Probably tag some specific tests and run then with a different xctestplan since they need to strictly run in serial by their own like localization tests or user preference tests
- A way to force AppSettings to use `.volatile` storage, in this way we can get rid 
- Probably a full Swift 6 migration would help spot and fix some other minor concurrency issues that make the tests work not so well in parallel.

PR #5127, opened 2026-02-20
Improved Defer Fulfillment

Velin92:
The new version uses async stream instead of yielding until the condition is met, and then it dispatches the stream alongside the timeout, and returns the result of whichever completes first.
Also added logging the source location

### DIFF
--- a/ElementX/Sources/Other/DeferredFulfillment.swift
+++ b/ElementX/Sources/Other/DeferredFulfillment.swift
@@ -42,25 +42,19 @@ struct DeferredFulfillmentError: Error {
 ///   - sourceLocation: The source location to attach to any recorded issues.
 ///   - until: callback that evaluates outputs until some condition is reached
 /// - Returns: The deferred fulfilment to be executed after some actions and that returns the result of the publisher.
-func deferFulfillment<P: Publisher>(_ publisher: P,
-                                    timeout: Duration = .seconds(10),
-                                    message: String? = nil,
-                                    sourceLocation: SourceLocation = #_sourceLocation,
-                                    until condition: @escaping (P.Output) -> Bool) -> DeferredFulfillment<P.Output> {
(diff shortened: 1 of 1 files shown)


ElementX/Sources/Other/DeferredFulfillment.swift
```diff
@@ -18,109 +19,121 @@ struct DeferredFulfillment<T> {
...
             }
+            group.addTask {
+                try await Task.sleep(for: timeout)
+                throw DeferredFulfillmentError.noOutput(message: message, sourceLocation: sourceLocation)
+            }
+            
+            defer { group.cancelAll() }
+            
+            // swiftlint:disable:next force_unwrapping
+            return try await group.next()!
```
pixlwave: We can `return try #require(try await group.next())` here like we do above and get rid of the force unwrap.

ElementX/Sources/Other/DeferredFulfillment.swift
```diff
@@ -169,35 +184,44 @@ func deferFulfillment<Value: Equatable>(_ asyncSequence: any AsyncSequence<Value
...
-            throw DeferredFulfillmentError.unexpectedFulfillment(message: message)
+            // Timeout elapsing without the condition firing = success.
+            group.addTask {
+                try await Task.sleep(for: timeout)
+            }
+            
+            defer { group.cancelAll() }
+            
+            // swiftlint:disable:next force_unwrapping
+            try await group.next()!
```
pixlwave: I guess the same applies here without the `return`.

ElementX/Sources/Other/DeferredFulfillment.swift
```diff
@@ -206,37 +230,44 @@ func deferFailure<P: Publisher>(_ publisher: P,
...
```
pixlwave: And here.

PR #5130, opened 2026-02-21
Add a `waitForConfirmation` function for testing.

Velin92:
`confirmation` is cool, but we were really missing the powerful expectation of `XCTests` and the capacity of being able to await a specific condition with a timeout.

So I made my own version of `waitConfirmation`.
It exists in two versions:
- The most optimal one which will have the stream listen to all the incoming confirmations until the expected count is meant, this one does not have a timeout, but is technically the most optimal if you know the confirmation is going to happen regardless
- A timeout version that uses `Task.sleep` and `waitTaskGroup` slightly less optimal but still very good, which can be used for testing the if a condition is not trigger within a timeout (for example an expectedCount: 0) or just if you think the confirmation may fail, and you don't want to wait 60 seconds for the test to automatically kill itself.

Both of them solve the real issue we have confirmation, which is that the standard confirmation does not wait as `expectation` to have met the condition, but once the code block reaches its end it will complete and check how many counts it reached, which in some specific test cases might not be enough like for example when testing a function call over a mock that gets called from a different Task or that does not reflect any change in the view model.
Take note that given the fact that both rely on testing what is usually synchronous code that triggers asynchronous task, and await for these to give their confirmation, *this version of the function accepts non async body* (even if it could be implemented also in this way) since is
- Not really required given out our testing contexts, which is synchronous code that sets up mocks, and trigger some specific view Model action or call that will internally dispatch asynchronous Tasks that will call the mock.
- It would not make sense from an API standpoint since the timeout should only account for awaiting the stream to complete NOT for the body to be awaited.

The async body case can already be handled quite well with the standard `confirmation`

The awesome documentation is courtesy of Claude Somnet 4.6, the rest is pretty much the result of myself having a lot of fun this weekend while studying Swift 6 concurrency.

This PR is definitely needed to fix some hard or just bad confirmation handling here: https://github.com/element-hq/element-x-ios/pull/5129


UnitTests/Sources/EmojiPickerScreenViewModelTests.swift
```diff
@@ -25,27 +25,18 @@ struct EmojiPickerScreenViewModelTests {
...
             context.send(viewAction: .emojiTapped(emoji: .init(id: "wave", value: reaction)))
-            
-            try await deferred.fulfill()
-            
-            // Since the reaction is called asynchronously after dismissing the picker
-            // We need to actively wait for the function to be called before fulfilling the test.
-            while !toggleReactionCalled {
-                await Task.yield()
-            }
         }
```
Velin92: if we want we can even use non timeout version for this one, the pro is that we avoid a Task.sleep which in a future where we want parellalization might create some issues.
The con is that if something breaks the test will fail after 60 seconds (the max run time for a test)

According to Swift Testing maintainers we should use Task.sleep sparingly if we want to have parellalization, because they make the parallel code deadlock when used too much, so when possible if we know the wait is going to resolve, maybe we should avoid them

I even thought about implementing a `deferredFulfill` version that does not take a timeout parameter and that uses confirmation and the async stream.

Ofc this is something we can do in the future when we have done a full swift 6 mgiration and ServiceLocator removal, so maybe is fine to keep this with a timeout for now that we use serial running.

UnitTests/Sources/PollFormScreenViewModelTests.swift
```diff
@@ -162,26 +162,17 @@ struct PollFormScreenViewModelTests {
...
             }
             context.alertInfo?.secondaryButton?.action?()
-            
-            try await deferred.fulfill()
-            
-            // Since the redactReasonClosure is called asynchronously after closing the alert
-            // We need to actively wait for the redactReasonClosure to be called before fulfilling the test.
-            while !redactReasonCalled {
-                await Task.yield()
-            }
```
Velin92: same comment as above

UnitTests/Sources/NavigationTabCoordinatorTests.swift
```diff
@@ -168,17 +168,17 @@ struct NavigationTabCoordinatorTests {
...
         let overlayCoordinator = SomeTestCoordinator()
         
-        try await confirmation("Callback should not be called when just changing mode",
-                               expectedCount: 0) { confirmation in
+        await waitConfirmation("Callback should not be called when just changing mode",
+                               expectedCount: 0,
+                               timeout: .seconds(1)) { confirmation in
             navigationTabCoordinator.setOverlayCoordinator(overlayCoordinator) {
                 confirmation()
             }
```
Velin92: This is a negative test and is the only case where I see having the timeout is required even with paralelization

ElementX.xcodeproj/project.pbxproj
```diff
@@ -4848,6 +4850,7 @@
 				7583EAC171059A86B767209F /* MediaProvider */,
 				7DBC911559934065993A5FF4 /* NotificationManager */,
 				1C62F5382CC9D9F7DCEC344A /* UserDiscoveryService */,
+				F5239D81C3F9D4A88937EC94 /* Utils */,
```
pixlwave: Let's call this directory `TestUtilities` and put DeferredFulfillment.swift in there too?

We could then also update the XcodeGen templates to simply include the directory rather than specific files.

UnitTests/Sources/TestUtilities/WaitingConfirmation.swift
```diff
@@ -0,0 +1,206 @@
...
+///
+/// `WaitConfirmation` is used in conjunction with ``waitConfirmation(_:expectedCount:isolation:sourceLocation:_:)``
+/// to synchronize async test expectations. It bridges between your test code and
+/// Swift Testing's `confirmation` mechanism using an `AsyncStream` under the hood.
+///
+/// You typically interact with this type via its `callAsFunction()` sugar:
+/// ```swift
+/// await waitConfirmation { confirmation in
+///     sut.onEvent = { confirmation() }
+///     sut.triggerEvent()
```
pixlwave, line 22: Can we go with something more standard like `model` `myObject` or even just `foo` here rather than `sut`?

UnitTests/Sources/TestUtilities/WaitingConfirmation.swift
```diff
@@ -0,0 +1,206 @@
...
+    /// This method is thread-safe — the count increment and the finish check are performed
+    /// atomically inside a `Mutex` lock.
+    func confirm() {
+        confirmationsCount.withLock { value in
+            continuation.yield()
+            value += 1
+            if value == expectedCount {
+                continuation.finish()
+            }
+        }
```
pixlwave, line 50: 👌

UnitTests/Sources/TestUtilities/WaitingConfirmation.swift
```diff
@@ -0,0 +1,206 @@
...
+/// Waits for a confirmation to be triggered an expected number of times within a synchronous body.
+///
+/// This is a wrapper around Swift Testing's `confirmation` that removes the need to manually
+/// manage an `AsyncStream` at the call site. The body receives a ``WaitConfirmation`` instance
+/// which can be called directly to signal that the expected event occurred.
+///
+/// The body is synchronous by design — it is intended for setting up mocks and triggering
+/// actions that schedule async work, rather than performing async work itself. The async
+/// waiting happens internally once the body returns, by draining the stream until all
+/// confirmations are received.
```
pixlwave, line 70: Documentation like this are really helpful when reviewing 🙌

UnitTests/Sources/Utils/WaitConfirmation.swift
```diff
@@ -0,0 +1,206 @@
...
+///
+/// The body is synchronous by design — it is intended for setting up mocks and triggering
+/// actions that schedule async work, rather than performing async work itself. The async
+/// waiting happens internally once the body returns, by draining the stream until all
+/// confirmations are received.
+///
+/// Unlike the timeout variant, this overload does not escape the body closure, which means
+/// you can safely capture mutable structs — a common pattern in Swift Testing.
+///
+/// > Warning: This overload has no timeout. If ``WaitConfirmation/confirm()`` is never called,
```
pixlwave:
```suggestion
/// > **Warning:** This overload has no timeout. If ``WaitConfirmation/confirm()`` is never called,
```

UnitTests/Sources/TestUtilities/WaitingConfirmation.swift
```diff
@@ -0,0 +1,206 @@
...
+/// you can safely capture mutable structs — a common pattern in Swift Testing.
+///
+/// > Warning: This overload has no timeout. If ``WaitConfirmation/confirm()`` is never called,
+/// > the test will hang indefinitely. Prefer the timeout variant when the confirmation
+/// > depends on asynchronous work that could silently fail.
+///
+/// Example:
+/// ```swift
+/// await waitConfirmation(expectedCount: 2) { confirmation in
+///     sut.onEvent = {
```
pixlwave, line 82: Same comment on `sut`

UnitTests/Sources/TestUtilities/WaitingConfirmation.swift
```diff
@@ -0,0 +1,206 @@
...
+/// > Note: Because this overload uses `withTaskGroup` internally to race the stream against
+/// > the timeout, the `body` closure is implicitly `@escaping`. This is why this is a separate
+/// > overload rather than a single function with an optional timeout — keeping them separate
+/// > allows the non-timeout variant to avoid `@escaping`, which lets you capture mutable structs
+/// > in `body` as is common in Swift Testing.
+///
+/// Example:
+/// ```swift
+/// await waitConfirmation(expectedCount: 1, timeout: .seconds(2)) { confirmation in
+///     sut.onNetworkResponse = { confirmation() }
```
pixlwave, line 148: Same comment again.

UnitTests/Sources/TestUtilities/WaitingConfirmation.swift
```diff
@@ -0,0 +1,206 @@
...
+    return try await confirmation(comment,
+                                  expectedCount: expectedCount,
+                                  isolation: isolation,
+                                  sourceLocation: sourceLocation) { confirmation in
+        let result = try body(.init(continuation: continuation,
+                                    expectedCount: expectedCount))
+        
+        // The reason why I don't add to the task group directly the non timeout implementation
+        // is that I do not want the body to be marked as @escaping and thus to be able to capture
+        // even mutable structs which is common in Swift Testing.
```
pixlwave, line 189: Probably worth wording this to point out that the task group makes this one `@escaping` meaning it can't use used with mutable structs, rather than talking about the other implementation in this function.

PR #5144, opened 2026-02-26
Set CI RunTests command retries default to 0, use 3 only for running unit tests

stefanceriu:


pixlwave (approved):
👍

Tools/Sources/Commands/CI/RunTests.swift
```diff
@@ -25,7 +25,7 @@ struct RunTests: AsyncParsableCommand {
     var osVersion = "26.1"
     
     @Option(help: "Number of times to retry failed tests. Only the failing tests are re-run, not the entire suite.")
```
Velin92, line 27: Maybe given the fact that there is this bug with swift testing we should edit this help comment, saying that it does indeed run the entire suite until they release support for it.

PR #5146, opened 2026-02-26
More fastlane to swift tool conversions

stefanceriu:
This does the following:
* make the pipeline a dependency of tools and as such can be invoked directly through `swift run pipeline`
* replaces all pipeline fastlane calls with swift ones
* adds a new CI command for configuring the nightly
* updates all nightly calls to swift tools

### DIFF
--- a/Tools/Sources/Commands/CI/ConfigureNightly.swift
+++ b/Tools/Sources/Commands/CI/ConfigureNightly.swift
@@ -1,0 +1,63 @@
+import ArgumentParser
+import CommandLineTools
+import Foundation
+import Yams
+
+struct ConfigureNightly: AsyncParsableCommand {
+    static let configuration = CommandConfiguration(abstract: "Configures the project for a Nightly build.",
+                                                    discussion: "Adds the Nightly variant to project.yml, updates secrets, runs xcodegen, and generates the app icon banner.")
+
+    @Option(help: "The build number to display on the app icon banner.")
+    var buildNumber: String
+
+    func run() async throws {
+        guard !buildNumber.isEmpty else {
+            throw ValidationError("Invalid build number.")
+        }
+
+        try addNightlyVariant()
+        
+        try Zsh.run(command: "swift run pipeline update-foss-secrets")
+        try Zsh.run(command: "xcodegen")
+
+        let releaseVersion = try readMarketingVersion()
+        try await generateAppIconBanner(version: releaseVersion, buildNumber: buildNumber)
+    }
+
+    /// Adds the Nightly variant include path to `project.yml` if it isn't already present.
+    private func addNightlyVariant() throws {
+        let projectURL = URL.projectDirectory.appending(component: "project.yml")
+        let projectString = try String(contentsOf: projectURL)
+        guard var projectConfig = try Yams.compose(yaml: projectString) else {
+            throw ValidationError("Failed to parse project.yml.")
+        }
+        
+        // Check if the nightly variant is already included
+        if projectConfig["include"]?.sequence?.contains(where: { $0.mapping?["path"] == "Variants/Nightly/nightly.yml" }) == false {
+            projectConfig["include"]?.sequence?.append(["path": "Variants/Nightly/nightly.yml"])
+        }
+
+        let updatedYAMLString = try Yams.serialize(node: projectConfig)
+        try updatedYAMLString.write(to: projectURL, atomically: true, encoding: .utf8)
+    }
+
+    /// Reads the MARKETING_VERSION from `project.yml`.
+    private func readMarketingVersion() throws -> String {
+        let projectURL = URL.projectDirectory.appending(component: "project.yml")
+        let projectString = try String(contentsOf: projectURL)
+
+        let marketingVersionRegex = /MARKETING_VERSION:\s*([^\s]+)/
+        guard let match = projectString.firstMatch(of: marketingVersionRegex) else {
+            throw ValidationError("Could not find MARKETING_VERSION in project.yml.")
+        }
+
+        return String(match.1)
+    }
+
+    /// Generates the app icon banner with version and build number.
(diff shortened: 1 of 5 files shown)


Tools/Sources/Commands/CI/ConfigNightly.swift
```diff
@@ -0,0 +1,74 @@
...
+        try Zsh.run(command: "swift run pipeline update-foss-secrets")
+        try Zsh.run(command: "xcodegen")
+
+        let releaseVersion = try readMarketingVersion()
+        try generateAppIconBanner(version: releaseVersion, buildNumber: buildNumber)
+    }
+
+    /// Adds the Nightly variant include path to `project.yml` if it isn't already present.
+    private func addNightlyVariant() throws {
+        let projectURL = URL.projectDirectory.appendingPathComponent("project.yml")
```
pixlwave: This one is deprecated.
```suggestion
        let projectURL = URL.projectDirectory.appending(component: "project.yml")
```

Tools/Sources/Commands/CI/ConfigureNightly.swift
```diff
@@ -0,0 +1,74 @@
...
+            let alreadyIncluded = includes.contains { node in
+                node["path"]?.string == nightlyPath
+            }
+
+            if !alreadyIncluded {
+                var updatedIncludes = Array(includes)
+                try updatedIncludes.append(Yams.compose(yaml: "path: \(nightlyPath)")!)
+                projectConfig["include"] = .sequence(.init(updatedIncludes))
+            }
+        }
```
pixlwave, line 38: This can be simplified a bit:

```suggestion
        if projectConfig["include"]?.sequence?.contains(where: { $0.mapping?["path"] == nightlyPath }) == false {
            projectConfig["include"]?.sequence?.append(["path": nightlyPath])
        }
```

Tools/Sources/Commands/CI/ConfigureNightly.swift
```diff
@@ -0,0 +1,74 @@
...
+        let projectURL = URL.projectDirectory.appendingPathComponent("project.yml")
+        let projectString = try String(contentsOf: projectURL)
+
+        let marketingVersionRegex = /MARKETING_VERSION:\s*([^\s]+)/
+        guard let match = projectString.firstMatch(of: marketingVersionRegex) else {
+            throw ValidationError("Could not find MARKETING_VERSION in project.yml.")
+        }
+
+        return String(match.1)
+    }
```
pixlwave, line 55: How about we do this a more structured way by reading it from the yml in `addNightlyVariant` and returning it from that function?

Tools/Sources/Commands/CI/ConfigNightly.swift
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        return String(match.1)
+    }
+
+    /// Generates the app icon banner with version and build number.
+    private func generateAppIconBanner(version: String, buildNumber: String) throws {
+        let bannerText = "\(version) (\(buildNumber))"
+        let iconPath = "Variants/Nightly/Resources/NightlyAppIcon.icon/Assets/Version.png"
+        try Zsh.run(command: "swift run tools app-icon-banner \(iconPath) --banner-text '\(bannerText)'")
```
pixlwave: I think we should use the command directly here, no need to `swift run` from a `swift run` :)

Tools/Sources/Commands/CI/ConfigNightly.swift
```diff
@@ -0,0 +1,74 @@
+import ArgumentParser
+import CommandLineTools
+import Foundation
+import Yams
+
+struct ConfigNightly: ParsableCommand {
```
pixlwave: *Please* lets be more Swifty now and call this `ConfigureNightly` 🥺

Tools/Sources/Commands/CI/ConfigNightly.swift
```diff
@@ -0,0 +1,74 @@
...
+    static let configuration = CommandConfiguration(abstract: "Configures the project for a Nightly build.",
+                                                    discussion:
+                                                    "Adds the Nightly variant to project.yml, updates secrets, runs xcodegen, and generates the app icon banner.")
```
pixlwave:
```suggestion
                                                    discussion: "Adds the Nightly variant to project.yml, updates secrets, runs xcodegen, and generates the app icon banner.")
```

pixlwave (commented):
This is cool! *waves bye bye to more Ruby*

Tools/Sources/Commands/CI/ConfigureNightly.swift
stefanceriu: Nice one 👍

Tools/Sources/Commands/CI/ConfigNightly.swift
stefanceriu: <del>Ah right, this is before we had `pipeline` exposed through `tools`</del> wrong comment 😁

Tools/Sources/Commands/CI/ConfigNightly.swift
stefanceriu: Suuure 😅

Tools/Sources/Commands/CI/ConfigureNightly.swift
stefanceriu: Would love to but chances are this whole thing is going to move to CI because other commands will need it (for tagging release par example)

pixlwave: In which case it would be nice to fetch it using the yaml so it doesn't break if we add a `$(MARKETING_VERSION)` somewhere in the project.yml 😇

stefanceriu: Yeah fair, I just couldn't be bothered to figure out the typecasting :))

PR #5147, opened 2026-02-26
Remove space FFs

Velin92:
About to add a new FF for live location sharing, so I thought would be nice to do a bit of a cleanup first


stefanceriu (approved):
🙌

PR #5149, opened 2026-02-27
Flaky test possible fix

Velin92:


UnitTests/Sources/RoomNotificationSettingsScreenViewModelTests.swift
```diff
@@ -221,6 +221,8 @@ struct RoomNotificationSettingsScreenViewModelTests {
         #expect(notificationSettingsProxyMock.setNotificationModeRoomIdModeReceivedArguments?.1 == .mute)
         #expect(notificationSettingsProxyMock.setNotificationModeRoomIdModeCallsCount == 2)
         
+        try await Task.sleep(for: .milliseconds(10)) // Workaround for flaky test
```
pixlwave, line 224: Personally I'd go for at least 100 here on CI, but we can see how it goes in practice.

PR #5150, opened 2026-02-27
Move more fastlane lanes to swift tool CI commands

stefanceriu:
- a swift command for running accessibility tests (successful run [here](https://github.com/element-hq/element-x-ios/actions/runs/22477984046/job/65109040011))
- and one for UI tests (successful run [here](https://github.com/element-hq/element-x-ios/actions/runs/22485507140))

### DIFF
--- a/ci_scripts/ci_post_clone.sh
+++ b/ci_scripts/ci_post_clone.sh
@@ -7,7 +7,7 @@ setup_xcode_cloud_environment
 install_xcode_cloud_brew_dependencies
 
 if [ "$CI_WORKFLOW" = "Nightly" ]; then
-    swift run tools ci configure-nightly
+    swift run tools ci configure-nightly --build-number $CI_BUILD_NUMBER
 elif [ "$CI_WORKFLOW" = "Element Pro" ]; then
     # Xcode Cloud automatically fetches the submodules.
     swift run pipeline configure-element-pro


.github/workflows/ui_tests.yml
```diff
@@ -28,69 +28,47 @@ jobs:
...
 
       - name: Run tests
         run: |
-          if [[ -z "${{ github.event.inputs.test_name }}" ]]; then
-            bundle exec fastlane ui_tests device:${{ matrix.device }} 
-          else
-            bundle exec fastlane ui_tests device:${{ matrix.device }} test_name:${{ github.event.inputs.test_name }}
+          args=(--device-type "${{ matrix.device }}")
+          if [[ -n "${{ github.event.inputs.test_name }}" ]]; then
+            args+=(--test-name "${{ github.event.inputs.test_name }}")
```
pixlwave: I think this needs an extra ` ` in it to separate the device name from the test name?

.github/workflows/unit_tests_enterprise.yml
```diff
@@ -28,36 +28,21 @@ jobs:
...
-
       - name: Run tests
-        run: bundle exec fastlane unit_tests skip_previews:true
-        
-      - name: Zip results # for faster upload
-        if: failure()
-        working-directory: fastlane/test_output
-        run: zip -r UnitTests.zip UnitTests.xcresult
-      
+        run: swift run -q tools ci unit-tests --skip-previews
```
pixlwave, line 38: Not sure if this is intentional but highlighting that this is the only one using `-q`

ci_scripts/ci_common.sh
```diff
@@ -40,9 +40,6 @@ setup_github_actions_environment() {
     export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1
     
     brew update && brew install xcodegen swiftlint swiftformat git-lfs pkl a7ex/homebrew-formulae/xcresultparser
-
-    bundle config path vendor/bundle
-    bundle install --jobs 4 --retry 3
```
pixlwave, line 45: Is it worth moving this into the Xcode Cloud function for now?

.github/workflows/unit_tests_enterprise.yml
stefanceriu: Unit tests are doing as well 😅

ci_scripts/ci_common.sh
pixlwave: Ignore me, I thought this was on a common function 🤦‍♂️

.github/workflows/ui_tests.yml
stefanceriu: Nice catch, thanks 👍

.github/workflows/unit_tests_enterprise.yml
stefanceriu: We agreed to do it everywhere

PR #5155, opened 2026-02-27
Refactor and redesign Location Sharing Screen

Velin92:
This is the first part to implemente live location sharing
- Adds an FF that for now only adds a single button that does nothing in the sharing location screen
- Removed "description" from the location item content, this was decided alongside design and product since its usage was a bit unclear and ambiguous, and not used a lot in the app itself
- Redesigned the location marker, and also added the possibility of having an avatar in it
- Now is possible to distinguish between location that are shared as a sender or as a simple pin and display an avatar of the sender in the first case
- Just overall redesign of the map view to match also closely with iOS 26
- Found a way (experimental) to update annotations
- Updated the alert copy when the permissions are missing
- Added a spinner to the center user location button when the location is not loaded yet.

The next part is to actually implement live location sharing.

Better to review this commit by commit.

### DIFF
--- a/ElementX/Sources/Generated/Strings.swift
+++ b/ElementX/Sources/Generated/Strings.swift
@@ -586,6 +586,8 @@ internal enum L10n {
   internal static var commonOpticIdIos: String { return L10n.tr("Localizable", "common_optic_id_ios") }
   /// or
   internal static var commonOr: String { return L10n.tr("Localizable", "common_or") }
+  /// Other options
+  internal static var commonOtherOptions: String { return L10n.tr("Localizable", "common_other_options") }
   /// Password
   internal static var commonPassword: String { return L10n.tr("Localizable", "common_password") }
   /// People
@@ -2244,6 +2246,8 @@ internal enum L10n {
   internal static var screenOnboardingSignInWithQrCode: String { return L10n.tr("Localizable", "screen_onboarding_sign_in_with_qr_code") }
   /// Create account
   internal static var screenOnboardingSignUp: String { return L10n.tr("Localizable", "screen_onboarding_sign_up") }
+  /// Welcome back
+  internal static var screenOnboardingWelcomeBack: String { return L10n.tr("Localizable", "screen_onboarding_welcome_back") }
   /// Welcome to the fastest %1$@ ever. Supercharged for speed and simplicity.
   internal static func screenOnboardingWelcomeMessage(_ p1: Any) -> String {
     return L10n.tr("Localizable", "screen_onboarding_welcome_message", String(describing: p1))
@@ -3707,22 +3711,10 @@ internal enum L10n {
     internal static var yourAvatar: String { return L10n.tr("Localizable", "a11y.your_avatar") }
   }
 
-  internal enum Common {
-    /// Other options
-    internal static var otherOptions: String { return L10n.tr("Localizable", "common.other_options") }
-  }
-
   internal enum Error {
     /// You're already logged in on this device as %1$@.
     internal static func accountAlreadyLoggedIn(_ p1: Any) -> String {
       return L10n.tr("Localizable", "error.account_already_logged_in", String(describing: p1))
-    }
-  }
-
-  internal enum Screen {
-    internal enum Onboarding {
-      /// Welcome back
-      internal static var welcomeBack: String { return L10n.tr("Localizable", "screen.onboarding.welcome_back") }
     }
   }
 }
--- a/ElementX/Sources/Other/SwiftUI/Views/LocationMarkerView.swift
+++ b/ElementX/Sources/Other/SwiftUI/Views/LocationMarkerView.swift
@@ -21,7 +21,7 @@ struct LocationMarkerView: View {
     private let circleRadius: CGFloat = 6 // in SVG space
     
     var body: some View {
-        // Generated through the SVG
+        // Generated from the SVG
         Canvas { context, canvasSize in
             let scaleX = canvasSize.width / 42
             let scaleY = canvasSize.height / 50
@@ -106,7 +106,8 @@ struct LocationMarkerView_Previews: PreviewProvider, TestablePreview {
                                mediaProvider: MediaProviderMock(configuration: .init()))
                 .colorScheme(.dark)
         }
-        .background(.red)
+        .padding(16)
+        .background(Color(red: 0.9, green: 0.85, blue: 0.8))
         .previewLayout(.sizeThatFits)
     }
 }
--- a/PreviewTests/Sources/__Snapshots__/PreviewTests/locationMarkerView.iPad-en-GB-0.png
+++ b/PreviewTests/Sources/__Snapshots__/PreviewTests/locationMarkerView.iPad-en-GB-0.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:fdc18ec0d17be9509576e2ededa9e4994fdfd79519cd852e8395e58a5b47877c
-size 65345
+oid sha256:c00139208711b767a8f19c563bc7cef80314c3382ce5ce7d481a71d5ef70df1a
+size 69042
--- a/ElementX/Resources/Localizations/en-US.lproj/Localizable.strings
+++ b/ElementX/Resources/Localizations/en-US.lproj/Localizable.strings
@@ -263,6 +263,7 @@
 "common_open_source_licenses" = "Open source licenses";
 "common_optic_id_ios" = "Optic ID";
 "common_or" = "or";
+"common_other_options" = "Other options";
 "common_password" = "Password";
 "common_people" = "People";
 "common_permalink" = "Permalink";
@@ -365,7 +366,6 @@
 "common_waiting_for_decryption_key" = "Waiting for this message";
 "common_world_readable_history" = "Anyone can see history";
 "common_you" = "You";
-"common.other_options" = "Other options";
 "common_unable_to_decrypt_insecure_device" = "Sent from an insecure device";
 "common_unable_to_decrypt_verification_violation" = "Sender's verified identity was reset";
 "confirm_recovery_key_banner_message" = "Confirm your recovery key to maintain access to your key storage and message history.";
@@ -668,6 +668,7 @@
 "screen_media_upload_preview_item_count" = "Item %1$d of %2$d";
 "screen_media_upload_preview_optimize_image_quality_title" = "Optimize image quality";
 "screen_media_upload_preview_processing" = "Processing...";
+"screen_onboarding_welcome_back" = "Welcome back";
 "screen_pinned_timeline_empty_state_description" = "Press on a message and choose “%1$@” to include here.";
 "screen_pinned_timeline_empty_state_headline" = "Pin important messages so that they can be easily discovered";
 "screen_push_history_title" = "Push history";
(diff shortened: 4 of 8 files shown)


ElementX/Sources/Other/MapLibre/MapLibreMapView.swift
```diff
@@ -80,17 +85,52 @@ struct MapLibreMapView: UIViewRepresentable {
...
+        
+        // Update existing annotations that are still present
+        let keptIDs = existingIDs.intersection(updatedIDs)
+        for id in keptIDs {
+            guard let existingAnnotation = existingByID[id],
+                  let updatedAnnotation = options.annotations[id] else {
+                continue
+            }
+            existingAnnotation.coordinate = updatedAnnotation.coordinate
+            if let annotationView = mapView.view(for: existingAnnotation) as? LocationAnnotationView {
```
Velin92, line 123: This function can definitely be improved in a later PR. I think instead of having the LocationAnnotation be directly used in the model, we could instrad pass a LocationAnntationData (as an independent struct not related to the MLN stuff) and the LocationAnnotationView  inside the MapLibreMapView, as a dictionary and then instantiate in the setup the LocationAnnotation, using the data, and in the update annotation function find the corresponding annotation by id and update its content given the data itself, and set the view (or re-render/recreate it) if anything has changed.

Just a comment to be used as a reminder or as an idea for the next PR, for now it does serve well the purpose of refreshing the view when is needed.

ElementX/Sources/Screens/LocationSharing/View/LocationSharingScreen.swift
```diff
@@ -0,0 +1,261 @@
...
+        }
+        .overlay(alignment: .topTrailing) {
+            centerToUserLocationButton
+        }
+        .overlay(alignment: .bottom) {
+            // We can't use a sheet because it might interfer with the alerts.
+            if context.viewState.isLocationPickerMode {
+                sharingOptionsOverlay
+            }
+        }
```
Velin92: I don't link this I would rather have this implemented as a sheet, so I don't need all that custom code to make it look like one, and find a way to present the alert on top of the sheet when permissions are not available. Maybe by nesting the .alert call when the sheet is present?
Could do but it does feel a bit ugly too.
The other solution I think would be to extend the indicator for the UserIndicatorController so that we can have a persistent indicator with an additional button to go the settings.

But I am not sure, open to suggestions.

Velin92: solved!

ElementX/Sources/Generated/Strings.swift
pixlwave, line 1: I've nested those 2 keys so they get properly flattened out.

ElementX/Sources/Other/MapLibre/LocationAnnotation.swift
```diff
@@ -11,23 +11,32 @@ import MapLibre
 import SwiftUI
 
 final class LocationAnnotation: NSObject, MLNAnnotation {
-    let coordinate: CLLocationCoordinate2D
+    let id: String
```
pixlwave, line 14: Can we document what this is/what it is for seeing as the object isn't identifiable and I see it is a UUID.

ElementX/Sources/Other/MapLibre/MapLibreMapView.swift
```diff
@@ -20,10 +20,10 @@ struct MapLibreMapView: UIViewRepresentable {
         /// The initial map center
         let mapCenter: CLLocationCoordinate2D
         
-        /// Map annotations
-        let annotations: [LocationAnnotation]
+        /// Map annotations keyed by a stable identifier (e.g. sender userID for user pins, UUID string for generic pins)
```
pixlwave, line 23: Ah ok the doc is here. Maybe we could use an enum instead for the ID, then it would be clear everywhere it is used.

ElementX/Sources/Other/MapLibre/MapLibreMapView.swift
pixlwave: Ok cool 👍

ElementX/Sources/Other/SwiftUI/Views/LocationMarkerView.swift
```diff
@@ -10,38 +10,103 @@ import Compound
...
     
     var body: some View {
-        CompoundIcon(\.locationPinSolid)
-            .dynamicTypeSize(.large)
-            .foregroundStyle(pinColor)
-            .padding(pinInsets)
-            .background {
-                backgroundShape
-                    .shadow(color: .black.opacity(0.2), radius: 4.1129, x: 0, y: 4.93548)
+        // Generated through the SVG
```
pixlwave:
```suggestion
        // Generated from the SVG
```

ElementX/Sources/Other/SwiftUI/Views/LocationMarkerView.swift
```diff
@@ -10,38 +10,103 @@ import Compound
...
                 .colorScheme(.dark)
+            
+            LocationMarkerView(userProfile: UserProfileProxy.mockDan,
+                               mediaProvider: MediaProviderMock(configuration: .init()))
+            
+            LocationMarkerView(userProfile: UserProfileProxy.mockDan,
+                               mediaProvider: MediaProviderMock(configuration: .init()))
+                .colorScheme(.dark)
         }
+        .background(.red)
```
pixlwave: Maybe we could use something like `.green.opacity(0.4)/`.blue.opacity(0.3)`/`.yellow.opacity(0.1)` here so the preview background colour looks somewhat close to the normal colours found on a map?

ElementX/Sources/Other/SwiftUI/Views/ToolbarButton.swift
pixlwave, line 1: 👌

ElementX/Sources/Other/Avatars.swift
pixlwave, line 1: Finally, this is waaaay better. Thank you 🙌

PreviewTests/Sources/__Snapshots__/PreviewTests/locationMarkerView.iPad-en-GB-0.png
pixlwave, line 1: It would also be nice to drop a `.padding()` on this preview so the border doesn't touch the white colour.

ElementX/Sources/Screens/LocationSharing/View/LocationSharingScreen.swift
```diff
@@ -0,0 +1,264 @@
...
+    @ToolbarContentBuilder
+    private var toolbar: some ToolbarContent {
+        ToolbarItem(placement: context.viewState.showShareAction ? .topBarLeading : .topBarTrailing) {
+            ToolbarButton(role: .close) {
+                context.send(viewAction: .close)
+            }
+        }
+                
+        if context.viewState.showShareAction {
+            ToolbarItem(placement: .topBarTrailing) {
```
pixlwave, line 71: We should probably use `.cancellationAction` and `.primaryAction` for these placements like we do throughout the rest of the app.

ElementX/Sources/Screens/LocationSharing/View/LocationSharingScreen.swift
```diff
@@ -0,0 +1,264 @@
...
+        if #available(iOS 26.0, *) {
+            .none // ShareLinks use a popover presentation on iOS 26, let it match that.
+        } else {
+            .sheet
+        }
+    }
+    
+    @State private var sharingOptionsSheetHeight: CGFloat = .zero
+    
+    private var sharingOptionsSheet: some View {
```
pixlwave, line 157: Were you going to extract this out into its own view so we could generate snapshots of it?

ElementX/Sources/Screens/LocationSharing/View/LocationSharingScreen.swift
```diff
@@ -0,0 +1,264 @@
...
+private struct LocationSharingLabel: View {
+    let text: String
+    let icon: KeyPath<CompoundIcons, Image>
+    let iconColor: Color
+    
+    var body: some View {
+        Label {
+            Text(text)
+                .frame(maxWidth: .infinity, alignment: .leading)
+                .padding(.vertical, 14)
```
pixlwave, line 201: I think this vertical padding is meant to be on the icon rather than the text (which isn't as tall).

ElementX/Sources/Screens/LocationSharing/View/LocationSharingScreen.swift
```diff
@@ -0,0 +1,264 @@
...
+    // MARK: - Private
+    
+    private var mainContent: some View {
+        mapView
+            .ignoresSafeArea(edges: .bottom)
+            .track(screen: context.viewState.isLocationPickerMode ? .LocationSend : .LocationView)
+            .navigationTitle(L10n.screenViewLocationTitle)
+            .navigationBarTitleDisplayMode(.inline)
+            .toolbar { toolbar }
+    }
```
pixlwave, line 37: Entirely optional (and completely un-tested), but I was curious what this would look like with custom bindings:

```suggestion
    /// Forwards `alertInfo` only when in picker mode, so alerts appear on the sheet.
    private var sheetAlertInfo: Binding<AlertInfo<LocationSharingViewError>?> {
        Binding(get: { context.viewState.isLocationPickerMode ? context.alertInfo : nil },
                set: { context.alertInfo = $0 })
    }
    
    /// Forwards `alertInfo` only when not in picker mode, so alerts appear on the main content.
    private var mainAlertInfo: Binding<AlertInfo<LocationSharingViewError>?> {
        Binding(get: { context.viewState.isLocationPickerMode ? nil : context.alertInfo },
                set: { context.alertInfo = $0 })
    }
    
    var body: some View {
        mapView
            .ignoresSafeArea(edges: .bottom)
            .track(screen: context.viewState.isLocationPickerMode ? .LocationSend : .LocationView)
            .navigationTitle(L10n.screenViewLocationTitle)
            .navigationBarTitleDisplayMode(.inline)
            .toolbar { toolbar }
            .sheet(isPresented: .constant(context.viewState.isLocationPickerMode)) {
                sharingOptionsSheet
                    .alert(item: sheetAlertInfo)
            }
            .alert(item: mainAlertInfo)
    }
```

ElementX/Sources/Screens/LocationSharing/View/LocationSharingScreen.swift
Velin92: doing that in the next PR where I am making another sheet

ElementX/Sources/Screens/LocationSharing/View/LocationSharingScreen.swift
Velin92: tried that but messes up the divider

ElementX/Sources/Other/MapLibre/MapLibreMapView.swift
Velin92: as I said this part is more of a test/experimental to test updating the annotations, will definitely improve it when I start working on live location

ElementX/Sources/Screens/LocationSharing/View/LocationSharingScreen.swift
Velin92: I guess it may work, but the thing is that I realized that the sheet appear staticlaly based on the mode, so I think have the view just set up the sheet differently with constant(true) is fine,

PR #5159, opened 2026-03-02
Fix notification and call delivery reliability

Svyathor:
Several bugs caused push notifications and incoming calls to be silently dropped or immediately dismissed under common real-world conditions:

- NSE: guard `withCheckedContinuation` in `handleCallNotification` against multiple resumes when rapid room-info updates arrive with hasRoomCall=true (undefined behaviour / crash in debug builds)

- NSE: call `contentHandler(mutableContent)` immediately when `NSEUserSession` init throws, so the notification is delivered right away instead of waiting up to 30 s for the iOS extension timeout

- ElementCallService: guarantee a minimum ring duration of 15 s so that calls whose expiration date was reached during NSE processing still appear on-screen rather than vanishing the instant CallKit shows them

- ElementCallServiceConstants: raise `ElementCallServiceNotificationDiscardDelta` from 15 s to 25 s — NSE session creation can take 10-20 s on a loaded device, leaving no margin before valid calls were downgraded to plain notifications

- ElementCallService: use `[weak self]` in the `asyncAfter` closure of `CXAnswerCallAction` to avoid a 1-second retain cycle

- ElementCallService: in `reportEndedCall`, cancel `incomingCallRoomInfoCancellable` and nil out `incomingCallID` so that duplicate `reportCall` invocations and stale state after a remote cancellation / answer-elsewhere are prevented

https://claude.ai/code/session_018tw6cP7FTXXUSF5YiPSEAV

### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Services/ElementCall/ElementCallService.swift
+++ b/ElementX/Sources/Services/ElementCall/ElementCallService.swift
@@ -194,7 +194,8 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
             return
         }
         
-        let ringDuration: Duration = .seconds(min(expirationDate.timeIntervalSince1970 - nowDate.timeIntervalSince1970, 90))
+        let rawDuration = expirationDate.timeIntervalSince1970 - nowDate.timeIntervalSince1970
+        let ringDuration: Duration = .seconds(min(max(rawDuration, 15), 90))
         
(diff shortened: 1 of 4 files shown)


stefanceriu:
I do realise this is mostly Claude but most of them seem reasonable. Would you like to push them through to the main app?

PR #5171, opened 2026-03-06
Static Location Sheet

Velin92:
This also refactored some code to pass all required info to display the sheet.
Screenshot 2026-03-06 at 15 25 51


ElementX/Sources/Screens/LocationSharing/View/LocationSharingScreen.swift
```diff
@@ -13,15 +13,20 @@ struct LocationSharingScreen: View {
...
+        switch context.viewState.interactionMode {
+        case .picker:
             mainContent
                 .sheet(isPresented: .constant(true)) {
-                    sharingOptionsSheet
+                    LocationPickerSheet(context: context)
                         .alert(item: $context.alertInfo)
                 }
-        } else {
+        case .viewStatic:
```
stefanceriu, line 23: Will there be a difference between `viewStatic` and `viewDynamic` or can we just call these `pick` and `view`?

ElementX/Sources/Screens/LocationSharing/View/LocationSharingScreen.swift
```diff
@@ -13,15 +13,20 @@ struct LocationSharingScreen: View {
...
             mainContent
-                .alert(item: $context.alertInfo)
+                .sheet(isPresented: .constant(true)) {
```
stefanceriu, line 25: Heh, cheeky!

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
```diff
@@ -76,17 +71,9 @@ struct LocationSharingScreenViewState: BindableState {
             false
         }
     }
-
-    var showShareAction: Bool {
```
stefanceriu, line 80: No more sharing? 😢

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
```diff
@@ -76,17 +71,9 @@ struct LocationSharingScreenViewState: BindableState {
...
-        switch interactionMode {
-        case .picker:
-            return false
-        case .viewStatic:
-            return true
-        }
-    }
     
     var isLocationLoading: Bool {
+        // May not work as intended on simulator
```
stefanceriu: Can you expand on this a bit please?

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
```diff
@@ -105,12 +92,12 @@ struct LocationSharingScreenViewState: BindableState {
     
     var userProfile: UserProfileProxy?
     
-    var shownUserProfile: UserProfileProxy? {
+    var staticLocationMarkerUserProfile: UserProfileProxy? {
         switch interactionMode {
         case .picker:
             isSharingUserLocation ? userProfile : nil
```
stefanceriu: Why is it called a staticLocation user profile if it works on the `picker` mode as well? 🤔

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
```diff
@@ -172,3 +159,10 @@ extension AlertInfo where T == LocationSharingViewError {
         }
     }
 }
+
+struct StaticLocationData: Hashable {
```
stefanceriu: Not entirely sure I like how this type is used within the timeline models. Maybe you should make it standalone and move it somewhere within /Services. Or perhaps in /Other together with LocationAnnotation and LocationMarkerView ¯\_(ツ)_/¯

ElementX/Sources/Screens/LocationSharing/View/LocationSharingScreen.swift
```diff
@@ -61,32 +66,21 @@ struct LocationSharingScreen: View {
     
     @ToolbarContentBuilder
     private var toolbar: some ToolbarContent {
-        ToolbarItem(placement: context.viewState.showShareAction ? .topBarLeading : .topBarTrailing) {
+        ToolbarItem(placement: .primaryAction) {
             ToolbarButton(role: .close) {
                 context.send(viewAction: .close)
             }
```
stefanceriu, line 72: Now honestly these placements always confuse me: shouldn't a big X button be shown as `.cancellationAction`?

ElementX/Sources/Screens/LocationSharing/View/LocationSharingScreen.swift
Velin92: yeah there will likely be differences in the data, so I'll introduce `.viewDynamic` or `.viewLive` in the follow up

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
Velin92: yeah but is only for the `viewStatic` case so you can see the popover directly added in the view when handling such case. The action is then handled in the sheet.

ElementX/Sources/Screens/LocationSharing/View/LocationSharingScreen.swift
Velin92: sadly is the only way to have alerts and sheet work together without having one dismiss the other (since the sheet needs to be always be presented)

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
Velin92: since the location may take time to load, we added a spinner in the view. On the simulator MapLibre is a bit weird and sometimes even with permissions on doesn't display your own location. It works well on device. Will add a better comment.

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
Velin92: my bad this should be called `locationMarkerUserProfile`, will update it

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
Velin92: oh yeah I'll move in the LocationAnnotation's folder

ElementX/Sources/Screens/LocationSharing/View/LocationSharingScreen.swift
Velin92: for this view we want to display it on the right now, design specifically asked for this.

PR #5181, opened 2026-03-09
Add support for reading accounts from Element Classic.

pixlwave:
This PR allows Element X to read accounts from Element Classic. To achieve this it:
- Adds 3 new settings to `app.yml` to configure Element X for access to Classic's App Group and Keychain Access Group.
- Ports the relevant code from Element Classic to decode models from the App Group.

| Element X | Element Classic |
| - | - |
| `ClassicAppAccountManager` |  `MXKAccountManager` and `MXFileStore` |
| `ClassicAppAES` | `MXAES` |
| `ClassicAppMXAccount` | `MXAccount` |
| `ClassicAppMXUser` | `MXUser` |

- Uses this code in a new `ClassicAppManager` which provides the keys from the Keychain Access Group.

Note: Whilst there are 2 commits in this PR, it would be better to review as a whole – I just wanted to keep the first commit in place as a reference as we may need more data from the store when implementing the feature.


ElementX/Sources/Services/Authentication/ClassicApp/ClassicAppAccountManager.swift
pixlwave, line 1: Now that I've tidied everything up, part of me wonders if this should simply become part of `ClassicAppManager` as that one is pretty small now.

pixlwave: Ignore this, turns out the separation is actually useful for unit tests.

ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -628,6 +630,19 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
...
+              let classicAppKeychainServiceIdentifier = InfoPlistReader.main.classicAppKeychainServiceIdentifier,
+              let classicAppKeychainAccessGroupIdentifier = InfoPlistReader.main.classicAppKeychainAccessGroupIdentifier else {
+            MXLog.info("Classic App IDs not set, manager disabled.")
+            return nil
+        }
+            
+        return ClassicAppManager(classicAppGroupIdentifier: classicAppGroupIdentifier,
+                                 classicAppKeychainServiceIdentifier: classicAppKeychainServiceIdentifier,
+                                 classicAppKeychainAccessGroupIdentifier: classicAppKeychainAccessGroupIdentifier)
+    }
```
Velin92: small nit, but this feels like can be a static func or even a static computed var

ElementX/Sources/Other/InfoPlistReader.swift
```diff
@@ -115,15 +119,34 @@ struct InfoPlistReader {
...
+        infoPlistValue(forKey: Keys.classicAppKeychainServiceIdentifier)
+    }
+    
+    var classicAppKeychainAccessGroupIdentifier: String? {
+        infoPlistValue(forKey: Keys.classicAppKeychainAccessGroupIdentifier)
+    }
+    
     // MARK: - Private
     
+    @_disfavoredOverload // Make sure optional types default to the optional version below.
```
Velin92, line 138: noice

ElementX/Sources/Services/Authentication/ClassicApp/ClassicAppAES.swift
```diff
@@ -0,0 +1,78 @@
...
+// Copyright 2026 Element Creations Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+import CommonCrypto
+import Foundation
+
+enum ClassicAppAES {
```
Velin92, line 11: I wonder why are we doing this client side? I mean since the crypto store on the classic app is still done by the rust sdk, couldn't we use a function directly from the SDK itself?

ElementX/Sources/Services/Authentication/ClassicApp/ClassicAppMXAccount.swift
```diff
@@ -0,0 +1,121 @@
...
+    let displayName: String?
+    let avatarURL: URL?
+    let serverName: String
+    let cryptoStoreURL: URL
+    let cryptoStorePassphrase: Data
+}
+
+// MARK: NSCoding Types
+
+class ClassicAppMXAccount: NSObject, NSCoding {
```
Velin92: small nit, use final.

UnitTests/Sources/ClassicAppAccountManagerTests.swift
```diff
@@ -0,0 +1,83 @@
...
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+@testable import ElementX
+import Foundation
+import Testing
+
+class ClassicAppAccountManagerTests {
```
Velin92: I see no deinit, why don't use a struct here?

ElementX/Sources/Services/Authentication/ClassicApp/ClassicAppMXAccount.swift
```diff
@@ -0,0 +1,121 @@
...
+        super.init()
+    }
+    
+    func encode(with coder: NSCoder) {
+        fatalError("Not available")
+    }
+}
+
+/// `MXUser` represents a user in Matrix.
+class ClassicAppMXUser: NSObject, NSCoding {
```
Velin92: same here

UnitTests/Sources/ClassicAppAccountManagerTests.swift
pixlwave: Discussed this, it's because it uses `Bundle(for Self.self)`, have marked as final.

ElementX/Sources/Application/AppCoordinator.swift
pixlwave: We discussed and moved it to the init.

ElementX/Sources/Services/Authentication/ClassicApp/ClassicAppAES.swift
pixlwave: So first we need to read Classic's normal file store to find the `activeAccount`, and only then can we get the URL to its crypto store. The `accountsV2` file in the file store is encrypted with a key in the Keychain, so this is necessary to decrypt that. The next PR will use the `cryptoStoreURL` and `cryptoStorePassphrase` to do all the magic.

I was hoping to use CryptoKit for this instead of all this CommonCrypto stuff, but as far as I can tell it doesn't support the AES CTR mode that Classic uses in the file store so I had to bring all this across.

I've added some docs to the 2 managers to explain this flow.

PR #5186, opened 2026-03-10
Replace the last of the fastlane lanes with swift tooling

stefanceriu:
- move more sharable code to CI static methods
- merge `release_to_github` and `prepare_next_release` into one single command as they had dependencies on each other
- remove all traces of ruby and fastlane


ci_scripts/ci_post_clone.sh
```diff
@@ -1,8 +1,14 @@
 #!/bin/sh
 
+# Return on failures
+# Fail when expanding unset variables
+# Trace each command before executing it
+set -eEu
+
```
pixlwave: Silly question… If this can go here on both post-clone and post-build, can we put it in the same place on the common script and automatically get it when sourcing that script?

fastlane/Fastfile
pixlwave, line 1: Image

Tools/Sources/Commands/CI/CI.swift
```diff
@@ -123,4 +124,41 @@ struct CI: ParsableCommand {
...
     }
+    
+    // MARK: - Git
+    
+    static func gitConfigureGlobals() async throws {
+        try await CI.run(.name("git"), ["config", "--global", "user.name", "Element CI"])
+        try await CI.run(.name("git"), ["config", "--global", "user.email", "ci@element.io"])
+    }
+    
+    static func gitGetRepoURL() async throws -> String {
```
pixlwave: Perhaps

```suggestion
    static func gitRepositoryURL() async throws -> String {
```

Tools/Sources/Commands/CI/ReleaseToGithub.swift
```diff
@@ -0,0 +1,169 @@
...
+
+        let yamlURL = URL.projectDirectory.appendingPathComponent(targetFilePath)
+        let yamlString = try String(contentsOf: yamlURL, encoding: .utf8)
+
+        let marketingVersionRegex = /MARKETING_VERSION:\s*(\S+)/
+        guard let match = yamlString.firstMatch(of: marketingVersionRegex) else {
+            throw ValidationError("Could not find MARKETING_VERSION in \(targetFilePath)")
+        }
+
+        let currentVersion = String(match.1)
```
pixlwave: Errrr, isn't this the same as `releaseVersion` above?

Tools/Sources/Commands/CI/ReleaseToGithub.swift
```diff
@@ -0,0 +1,169 @@
...
+
+        guard let newVersion = bumpPatchVersion(currentVersion) else {
+            throw ValidationError("Invalid version format: \(currentVersion)")
+        }
+
+        // Bump the patch version using sed (preserves file formatting)
+        try await CI.run(.path("/bin/zsh"), ["-cu", "sed -i '' 's/MARKETING_VERSION: \(currentVersion)/MARKETING_VERSION: \(newVersion)/g' \(targetFilePath)"])
+        logger.info("Version updated from \(currentVersion) to \(newVersion)")
```
pixlwave, line 54: Wellll booo, but yeah makes sense. Can we use `CI.run(.name("sed"), …)` here at least?

Tools/Sources/Commands/CI/ReleaseToGithub.swift
```diff
@@ -0,0 +1,169 @@
...
+        try await CI.run(.name("git"), ["commit", "-m", "Prepare next release"])
+        
+        try await CI.gitPush()
+
+        try await rebaseMainOntoCurrentBranch()
+    }
+
+    // MARK: - Private
+
+    private func createGitHubRelease(version: String) async throws -> String {
```
pixlwave, line 70: I feel like we should be re-using the command line tools more, rather than re-writing a bunch of stuff, e.g.

https://github.com/element-hq/swift-command-line-tools/blob/c085ad85c27970e4bba2dd1180c2a53ab77022ba/Sources/CommandLineTools/Release/Package.swift#L64

and more generally

https://github.com/element-hq/swift-command-line-tools/blob/main/Sources/CommandLineTools/Release/Git.swift

ci_scripts/ci_post_clone.sh
stefanceriu: We sure can, that's what we used to do. Claude decided to move them and I to see what you guys think. Guess we're going back to what we had 😅

Tools/Sources/Commands/CI/ReleaseToGithub.swift
stefanceriu: Ups, copy pasta 😞

Tools/Sources/Commands/CI/ReleaseToGithub.swift
stefanceriu: The release is bundled with xcframework uploading and the rest are using Zsh instead of Subprocess. I think we should tackle these separately after we get a better feel for CI work.

Tools/Sources/Commands/CI/ReleaseToGithub.swift
```diff
@@ -0,0 +1,159 @@
...
+
+        guard (200...299).contains(httpResponse.statusCode) else {
+            let errorBody = String(data: data, encoding: .utf8) ?? "Unknown error"
+            throw ReleaseError.failedToCreateRelease("HTTP \(httpResponse.statusCode): \(errorBody)")
+        }
+
+        guard let json = try JSONSerialization.jsonObject(with: data) as? [String: Any],
+              let releaseBody = json["body"] as? String else {
+            throw ReleaseError.failedToParseResponse
+        }
```
pixlwave, line 102: Should we use Codable here?

Tools/Sources/Commands/CI/ReleaseToGithub.swift
```diff
@@ -0,0 +1,159 @@
...
+        
+        return releaseBody
+    }
+
+    private func updateChangelog(version: String, generatedNotes: String) throws {
+        let changesURL = URL.projectDirectory.appendingPathComponent("CHANGES.md")
```
pixlwave:
```suggestion
        let changesURL = URL.projectDirectory.appending(component: "CHANGES.md")
```

Tools/Sources/Commands/CI/ReleaseToGithub.swift
```diff
@@ -0,0 +1,159 @@
...
+            .replacingOccurrences(of: "### ", with: "\n")
+            .replacingOccurrences(of: "## ", with: "### ")
+
+        guard !cleanedNotes.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty else {
+            throw ReleaseError.missingReleaseNotes
+        }
+
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "yyyy-MM-dd"
+        let releaseDate = dateFormatter.string(from: Date())
```
pixlwave:
```suggestion
        let releaseDate = Date().formatted(.iso8601.year().month().day())
```

Tools/Sources/Commands/CI/ReleaseToGithub.swift
```diff
@@ -0,0 +1,159 @@
...
+        logger.info("Updated CHANGES.md with release notes.")
+    }
+    
+    private func bumpPatchVersion(_ version: String) -> String? {
+        let regex = /^(\d{2})\.(\d{2})\.(\d+)$/
+        guard let match = version.firstMatch(of: regex) else { return nil }
+
+        let year = String(match.1)
+        let month = String(match.2)
+        let patch = Int(match.3)! + 1
```
pixlwave: Maybe we should add this one to the `guard let` rather than force unwrapping?

Tools/Sources/Commands/CI/ReleaseToGithub.swift
stefanceriu: Prooobably but seeing as we have all that stuff in CommandLine tools let's just wait until we merge the 2

PR #5192, opened 2026-03-11
Add default sound to APNS payload for eventIdOnly pushes

Reaster0:
*vibecoded pr | i do not know what i'm doing exactly*
following the issue https://github.com/element-hq/element-x-ios/issues/5132
and their counterpart on continuwuity
https://forgejo.ellis.link/continuwuation/continuwuity/issues/1533
https://forgejo.ellis.link/continuwuation/continuwuity/issues/1424

This PR ensures that incoming push notifications utilizing the eventIdOnly format are no longer silent by explicitly including the sound: "default" key in the APNS payload during pusher registration.

Previously, the APSInfo structure omitted the sound property, causing iOS to treat these background pushes as silent notifications, especially since some homeservers (continuwuity) discard the tweaks field when the event_id_only flag is used.

### DIFF
--- a/ElementX/Sources/Services/Notification/Manager/APNSPayload.swift
+++ b/ElementX/Sources/Services/Notification/Manager/APNSPayload.swift
@@ -21,10 +21,12 @@ struct APSAlert: Encodable {
 struct APSInfo: Encodable {
     let mutableContent: Int
     let alert: APSAlert
+    let sound: String?
 
     enum CodingKeys: String, CodingKey {
         case mutableContent = "mutable-content"
         case alert
+        case sound
     }
 }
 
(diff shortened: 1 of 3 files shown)


pixlwave:
Hey, thanks for your PR but as [mentioned](https://github.com/element-hq/element-x-ios/issues/5132#issuecomment-3960958253) in the issue, this isn't the right fix. The SDK loads the event and determines whether or not the notifications should be noisy and then the app adds the sound based on that computation.

The necessary fix is related to the SDK's `NotificationClient` and its computation of the [is_noisy](https://github.com/matrix-org/matrix-rust-sdk/blob/791f4bc87102de038d1ddb9352a8745370d59f1c/crates/matrix-sdk-ui/src/notification_client.rs#L1022) property based on your push rules.

PR #5209, opened 2026-03-13
Render the live location timeline item

Velin92:
This PR introduces a new timeline item to handle the Live Location sharing.

As of right now this PR only handles the rendering of such item, but doesn't provide any actions when it is tapped.
This includes:
- 2 new location marker states, one for the live item, and one for the placeholder/stopped state
- The LLS timeline model, which is used to translated the SDK content of the LLS timeline item into the data required to render the timeline item view.
- the LLS timeline item view which either be in a loading, in a stopped or in a live state (which is rendered slightly differently between the outgoing and the received state, since in the outgoing state you are able to also stop it, but this action is still not implemented yet)
- A way to hide the timestamp/sendinfo since we don't want to render it when in a both live and outgoing state
- Updated preview tests to handle all the possible states 
- The FF is also used to filter or not the beacon info events from the timeline.
- PreviewScrollView which is a nice way to keep the scrollable preview but have also a sizeThatFits preview snapshot.

NOTE: There are some bugs within the SDK that make the item no work properly sometimes like:
- `isLive` is always false
- The item is always marked as sent in clear even if that's not entirely true (it's due to beacon info being a non encrypted state event) but a PR with the fix on the SDK is already open: https://github.com/matrix-org/matrix-rust-sdk/pull/6308
- the timeline item disappears completely after the set expiration time, instead of remaining in the timeline with the stopped state.- 

These are entirely SDK issues, so they will be fixed in a future update

### DIFF
--- a/ElementX/Sources/Other/SwiftUI/PreviewScrollView.swift
+++ b/ElementX/Sources/Other/SwiftUI/PreviewScrollView.swift
@@ -7,6 +7,7 @@
 
 import SwiftUI
 
+#if DEBUG
 /// Only use in Previews! This useful scroll view allows you to still have a scroll view when previewing in Xcode
 /// but ignores it when running the tests, which allows you to still use it's content directly in preview tests
 /// and render the preview with `sizeThatFits` layout.
@@ -23,3 +24,4 @@ struct PreviewScrollView<Content: View>: View {
         }
     }
 }
+#endif


ElementX/Sources/Other/Extensions/Date.swift
```diff
@@ -64,6 +64,11 @@ extension Date {
     static var mock: Date {
         DateComponents(calendar: .current, year: 2007, month: 1, day: 9, hour: 9, minute: 41).date ?? .now
     }
+    
+    /// A fixed date used for mocks, previews etc.
+    static var inOneMinute: Date {
```
stefanceriu: I would call this `oneMinuteLater` but more importantly, why one minute and not any other value especiall as tests and previews are notoriously flakey on timer based components.

ElementX/Sources/Other/SwiftUI/Views/LocationMarkerView.swift
```diff
@@ -9,11 +9,85 @@
 import Compound
 import SwiftUI
 
+enum LocationMarkerKind {
```
stefanceriu: Does this really belong to the UI level or should it be moved to the models? 🤔

ElementX/Sources/Other/SwiftUI/Views/LocationMarkerView.swift
```diff
@@ -9,11 +9,85 @@
...
+}
+
 struct LocationMarkerView: View {
-    var userProfile: UserProfileProxy?
-    var fillColor = Color.compound.bgCanvasDefault
-    var strokeColor = Color.compound.iconSecondaryAlpha
-    var dotColor = Color.compound.iconPrimary
+    var kind: LocationMarkerKind
+    
+    var fillColor: Color {
```
stefanceriu: Let's make all of these private while at it and perhaps move them underneath the `body` to follow all our other components.

ElementX/Sources/Other/SwiftUI/Views/LocationMarkerView.swift
```diff
@@ -94,17 +168,27 @@ struct LocationMarkerView: View {
...
             
-            LocationMarkerView(userProfile: UserProfileProxy.mockDan,
+            // Static user without avatar
+            LocationMarkerView(kind: .staticUser(.init(userID: "@someone:matrix.org",
+                                                       displayName: "Someone")))
+            
+            // Live user with avatar
+            LocationMarkerView(kind: .liveUser(.mockDan),
                                mediaProvider: MediaProviderMock(configuration: .init()))
-                .colorScheme(.dark)
```
stefanceriu, line 107: I wonder if it's worth automating dark scheme preview tests. We do have a lot of screenshots already though..

ElementX/Sources/Screens/LocationSharing/View/LocationSharingScreen.swift
```diff
@@ -56,7 +56,7 @@ struct LocationSharingScreen: View {
             .ignoresSafeArea(.all, edges: mapSafeAreaEdges)
             
             if context.viewState.isLocationPickerMode {
-                LocationMarkerView(userProfile: context.viewState.locationMarkerUserProfile, mediaProvider: context.mediaProvider)
+                LocationMarkerView(kind: context.viewState.locationMarkerKind, mediaProvider: context.mediaProvider)
```
stefanceriu, line 59: Yeah no, you need to move it. I don't think view layer objects should end up in the view state generally.

ElementX/Sources/Other/SwiftUI/PreviewScrollView.swift
```diff
@@ -0,0 +1,22 @@
+//
+// Copyright 2026 Element Creations Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+import SwiftUI
+
+struct PreviewScrollView<Content: View>: View {
```
stefanceriu, line 14: Please add documentation here, it's not directly obvious what this does and why.

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
```diff
@@ -34,6 +34,12 @@ struct LocationSharingScreenViewState: BindableState {
         self.mapURLBuilder = mapURLBuilder
         self.showLiveLocationSharingButton = showLiveLocationSharingButton
         self.ownUserID = ownUserID
+        userProfile = switch interactionMode {
```
stefanceriu, line 38: Please add a new line here 🥹

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
```diff
@@ -34,6 +34,12 @@ struct LocationSharingScreenViewState: BindableState {
...
+        case .viewStatic(let locationData):
+            .init(sender: locationData.sender)
+        case .picker:
```
stefanceriu, line 41: We said we're gonna rename these https://github.com/element-hq/element-x-ios/pull/5171#discussion_r2906524858

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
stefanceriu, line 38: Umm.. why is the `var userProfile: UserProfileProxy` at the bottom at the file but set from the constructor at the top of the file? 🙃 Let's update the structure to match all the other components please 🙏

ElementX/Sources/Services/Timeline/TimelineItemContent/TimelineEventContent.swift
```diff
@@ -11,5 +11,6 @@ import Foundation
 enum TimelineEventContent: Hashable {
     case message(EventBasedMessageTimelineItemContentType)
     case poll(question: String)
+    case liveLocation
```
stefanceriu, line 14: Shouldn't we move this so it matches msgLike and the SDK? Same question for polls I guess..

ElementX/Sources/Services/Timeline/TimelineItems/Items/Messages/LiveLocationRoomTimelineItem.swift
```diff
@@ -0,0 +1,53 @@
...
+    let sender: TimelineItemSender
+    let content: LiveLocationRoomTimelineItemContent
+    var properties = RoomTimelineItemProperties()
+}
+
+struct LiveLocationRoomTimelineItemContent: Equatable {
+    let isLive: Bool
+    let timeoutDate: Date
+    
+    let lastLocation: BeaconInfo?
```
stefanceriu: Do we event care about other locations or can we just proxy the `geoURI` transparently?

ElementX/Sources/Services/Timeline/TimelineItems/Items/Messages/LiveLocationRoomTimelineItem.swift
```diff
@@ -0,0 +1,53 @@
...
+    let timestamp: Date
+    let geoURI: GeoURI?
+}
+
+extension BeaconInfo {
+    init(from beaconInfo: MatrixRustSDK.BeaconInfo) {
+        timestamp = Date(timeIntervalSince1970: Double(beaconInfo.ts) / 1000)
+        geoURI = GeoURI(string: beaconInfo.geoUri)
+    }
+}
```
stefanceriu, line 41: Seeing as they're not private, can we put these constructors in the actual struct instead of extensions?

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemFactory.swift
```diff
@@ -433,6 +432,28 @@ struct RoomTimelineItemFactory: RoomTimelineItemFactoryProtocol {
...
+                                     timestamp: eventItemProxy.timestamp,
+                                     isOutgoing: isOutgoing,
+                                     isEditable: eventItemProxy.isEditable,
+                                     canBeRepliedTo: eventItemProxy.canBeRepliedTo,
+                                     sender: eventItemProxy.sender,
+                                     content: .init(from: liveLocationContent, timestamp: eventItemProxy.timestamp),
+                                     properties: .init(replyDetails: buildTimelineItemReplyDetails(messageLikeContent.inReplyTo),
+                                                       isThreaded: messageLikeContent.threadRoot != nil,
+                                                       threadSummary: buildTimelineItemThreadSummary(messageLikeContent.threadSummary),
+                                                       isEdited: false, // Can never be edited
```
stefanceriu: I would let the SDK worry about that instead of taking decisions you don't need to take.

ElementX/Sources/Services/Timeline/GeoURI.swift
```diff
@@ -70,7 +70,7 @@ struct GeoURI: Hashable {
 private typealias RegexGeoURI = Regex<(Substring, latitude: Substring, longitude: Substring, uncertainty: Substring?)>
 
 private extension RegexGeoURI {
-    static let standard: Self = /geo:(?<latitude>-?\d+(?:\.\d+)?),(?<longitude>-?\d+(?:\.\d+)?)(?:;u=(?<uncertainty>\d+(?:\.\d+)?))?/
+    static let standard: Self = /geo:(?<latitude>-?\d+(?:\.\d+)?),(?<longitude>-?\d+(?:\.\d+)?)(?:,-?\d+(?:\.\d+)?)?(?:;u=(?<uncertainty>\d+(?:\.\d+)?))?/
```
stefanceriu, line 73: Uh what? This definitely deserved a commit of its own.

stefanceriu:
Oh, one more thing, please update the PR title: `LLS Timeline Item` doesn't  mean anything to anybody when reaching the changelog and the release notes.

ElementX/Sources/Other/Extensions/Date.swift
Velin92: no worries had to change this...

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemFactory.swift
Velin92: there is no message content so there is no `isEdited` value, that's why I need to take this decision myself.
`MessageLikeContent != MessageContent` (the latter contains `isEdited`)

at most I can decide to not specify it and leave it to the default constructor which already defaults to `isFalse` if you prefer, but is pretty much the same thing

ElementX/Sources/Services/Timeline/TimelineItems/Items/Messages/LiveLocationRoomTimelineItem.swift
Velin92: no because...
having an init in a struct overrides swift generated default init. So you can init with `LiveLocationRoomTimelineItemContent(isLive: timeoutDate: lastLocation:)` for example which is still useful for mocking and testing purposes
Instead extensions allow inits to be added on top of the default init.

Other solution is to have 2 inits, and you write the default init yourself, but feels a waste of code

ElementX/Sources/Services/Timeline/TimelineItems/Items/Messages/LiveLocationRoomTimelineItem.swift
Velin92: actually yeah I think we don't need lastLocation, lastGeoURI might be enough

ElementX/Sources/Services/Timeline/TimelineItemContent/TimelineEventContent.swift
Velin92: No because poll and liveLocation lack the `MessageContent` they have only have the `MessageLikeContent`

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
Velin92: yeah but I didn't touch the LocationSharingView yet, since I have just implemented the timeline item, not the live map. Please hold on a bit more to it.

ElementX/Sources/Services/Timeline/GeoURI.swift
Velin92: right, sorry slipped through my mind while I was fixing it, since live location can also includes an altitude value. So I added an optional regex for it when available.

ElementX/Sources/Other/SwiftUI/Views/LocationMarkerView.swift
stefanceriu: Any thoughts on this?

ElementX/Sources/Other/SwiftUI/PreviewScrollView.swift
stefanceriu: Perhaps we should wrap it in a #if DEBUG so it enforces the Only use in Previews?

ElementX/Sources/Services/Timeline/TimelineItems/Items/Messages/LiveLocationRoomTimelineItem.swift
stefanceriu: Riight, didn't know that's possible, nice 👍

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemFactory.swift
stefanceriu: Oh I'm sorry, I confused `isEdited` with `eventItemProxy.isEditable` which is set a bit above.

ElementX/Sources/Services/Timeline/TimelineItemContent/TimelineEventContent.swift
stefanceriu: Fair 👍

stefanceriu (approved):

ElementX/Sources/Other/SwiftUI/Views/LocationMarkerView.swift
Velin92: Like adding a new run to the test plan with dark mode?
Yeah we could do it in the future but this also means we are going to double the number of our snapshots

ElementX/Sources/Other/SwiftUI/PreviewScrollView.swift
Velin92: I'll give it a try

PR #5196, opened 2026-03-12
Remove check for allowed provider when signing in using QR code

hughns:
This removes the check for whether the server name is allowed or not. This is based on discussion with @mxandreas and @pixlwave.

It also adds support for the QR code containing a base URL as well as a server name which will be required for the 2025 version of MSC4108.

### Pull Request Checklist

- [x] I read the [contributing guide](https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md).
- [x] Pull request contains a [changelog label](https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md#changelog).

**UI changes have been tested with:**
- [ ] iPhone and iPad simulators in portrait and landscape orientations.
- [ ] Dark mode enabled and disabled.
- [ ] Various sizes of dynamic type.
- [ ] Voiceover enabled.

PR #5280, opened 2026-03-18
Rename Sign Out to Remove Device + Update SDK 26.03.18

Velin92:
Also regenerated preview tests

### DIFF
--- a/ElementX/Resources/Localizations/en-US.lproj/Localizable.strings
+++ b/ElementX/Resources/Localizations/en-US.lproj/Localizable.strings
@@ -226,6 +226,7 @@
 "common_empty_file" = "Empty file";
 "common_encryption" = "Encryption";
 "common_encryption_enabled" = "Encryption enabled";
+"common_ends_in_time_ios" = "Ends %1$@";
 "common_enter_your_pin" = "Enter your PIN";
 "common_error" = "Error";
 "common_everyone" = "Everyone";
@@ -242,7 +243,6 @@
 "common_image" = "Image";
 "common_in_reply_to" = "In reply to %1$@";
 "common_invite_unknown_profile" = "This Matrix ID can't be found, so the invite might not be received.";
-"common_ios_expiration_time" = "Ends %1$@";
 "common_leaving_room" = "Leaving room";
 "common_leaving_space" = "Leaving space";
 "common_light" = "Light";
--- a/ElementX/Sources/Other/Logging/Tracing.swift
+++ b/ElementX/Sources/Other/Logging/Tracing.swift
@@ -37,12 +37,10 @@ enum Tracing {
         // Log everything on integration tests to check whether
         // the logs contain any sensitive data. See `integration-tests.yml`
         let level: LogLevel = ProcessInfo.isRunningIntegrationTests ? .trace : logLevel
-        let sentryConfig: SentryConfig? = if let sentryDSN = sentryURL?.absoluteString {
-            SentryConfig(dsn: sentryDSN,
+        let sentryConfig = sentryURL.map { url in
+            SentryConfig(dsn: url.absoluteString,
                          appVersion: InfoPlistReader.app.bundleShortVersionString,
                          appPlatform: ProcessInfo.processInfo.platform)
-        } else {
-            nil
         }
         
         return .init(logLevel: level.rustLogLevel,
@@ -54,7 +52,7 @@ enum Tracing {
                                          fileSuffix: fileExtension,
                                          // Total compressed size needs to be under CloudFlare's max request size of 50Mb
                                          maxTotalSizeBytes: 100 * 1024 * 1024, // 100Mb
-                                         maxAgeSeconds: 7 * 24 * 60 * 60),
+                                         maxAgeSeconds: 7 * 24 * 60 * 60), // One week
                      sentryConfig: sentryConfig)
     }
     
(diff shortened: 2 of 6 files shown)


ElementX/Resources/Localizations/en-US.lproj/Localizable.strings
```diff
@@ -240,12 +242,15 @@
 "common_image" = "Image";
 "common_in_reply_to" = "In reply to %1$@";
 "common_invite_unknown_profile" = "This Matrix ID can't be found, so the invite might not be received.";
+"common_ios_expiration_time" = "Ends %1$@";
```
pixlwave: Side note: This doesn't match our rule of `common.`/`action.` having a key that matches the string. Also for iOS/Android specific strings they should end in `_ios`/`_android` to be filtered. Putting the platform in the middle doesn't do anything.

ElementX/Resources/Localizations/en-US.lproj/Localizable.strings
```diff
@@ -366,6 +372,7 @@
 "common_voice_message" = "Voice message";
 "common_waiting" = "Waiting…";
 "common_waiting_for_decryption_key" = "Waiting for this message";
+"common_waiting_live_location" = "Waiting for live location…";
```
pixlwave, line 375:
```suggestion
"common_waiting_for_live_location" = "Waiting for live location…";
```

ElementX/Sources/Other/Logging/Tracing.swift
```diff
@@ -37,6 +37,13 @@ enum Tracing {
         // Log everything on integration tests to check whether
         // the logs contain any sensitive data. See `integration-tests.yml`
         let level: LogLevel = ProcessInfo.isRunningIntegrationTests ? .trace : logLevel
+        let sentryConfig: SentryConfig? = if let sentryDSN = sentryURL?.absoluteString {
+            SentryConfig(dsn: sentryDSN,
+                         appVersion: InfoPlistReader.app.bundleShortVersionString,
+                         appPlatform: ProcessInfo.processInfo.platform)
+        } else {
+            nil
+        }
```
pixlwave: I don't think we really need that `if`/`else` here.
```suggestion
        let sentryConfig = sentryURL.map { url in
            SentryConfig(dsn: url.absoluteString,
                         appVersion: InfoPlistReader.app.bundleShortVersionString,
                         appPlatform: ProcessInfo.processInfo.platform)
        }
```

ElementX/Sources/Other/Logging/Tracing.swift
```diff
@@ -47,8 +54,8 @@ enum Tracing {
                                          fileSuffix: fileExtension,
                                          // Total compressed size needs to be under CloudFlare's max request size of 50Mb
                                          maxTotalSizeBytes: 100 * 1024 * 1024, // 100Mb
-                                         maxAgeSeconds: 7 * 24 * 60 * 60), // One week
```
pixlwave, line 50: What's the reason for removing the `One week` comment?

ElementX/Sources/Other/Extensions/ProcessInfo.swift
```diff
@@ -85,4 +85,14 @@ extension ProcessInfo {
...
         #endif
     }
+    
+    var platform: String {
+        if #available(iOS 26.1, *), ProcessInfo.processInfo.isiOSAppOnVision {
+            "visionOS"
+        } else if ProcessInfo.processInfo.isiOSAppOnMac {
+            "macOS"
+        } else {
+            "iOS"
```
pixlwave, line 95: Should we maybe use `ProcessInfo.processInfo.operatingSystemName()` here?

ElementX/Sources/Other/Extensions/ProcessInfo.swift
```diff
@@ -85,4 +85,14 @@ extension ProcessInfo {
         false
         #endif
     }
+    
+    var platform: String {
+        if #available(iOS 26.1, *), ProcessInfo.processInfo.isiOSAppOnVision {
```
pixlwave, line 90: ha I forgot they only added this property recently 🙃

ElementX/Sources/Services/Authentication/AuthenticationService.swift
```diff
@@ -281,7 +281,7 @@ private extension HumanQrLoginError {
             .qrCodeError(.slidingSyncNotAvailable)
         case .OtherDeviceNotSignedIn:
             .qrCodeError(.deviceNotSignedIn)
-        case .Unknown, .OidcMetadataInvalid, .CheckCodeAlreadySent, .CheckCodeCannotBeSent:
+        case .Unknown, .OidcMetadataInvalid, .CheckCodeAlreadySent, .CheckCodeCannotBeSent, .UnsupportedQrCodeType:
```
pixlwave: Let's use the mapping in #5196 perhaps?

ElementX/Sources/Services/Authentication/LinkNewDeviceService.swift
```diff
@@ -201,7 +201,7 @@ private extension QRCodeLoginError {
             .cancelled
         case .OtherDeviceAlreadySignedIn:
             .deviceAlreadySignedIn
-        case .Unknown, .MissingSecretsBackup, .DeviceIdAlreadyInUse:
+        case .Unknown, .MissingSecretsBackup, .DeviceIdAlreadyInUse, .UnsupportedQrCodeType:
```
pixlwave: Same here too #5196.

ElementX/Sources/Other/Extensions/ProcessInfo.swift
Velin92: that's deprecated, we should use `operatingSystemVersion` but that also includes the OS version which we don't care

pixlwave: Ah my bad, hadn't realised it was deprecated.

PR #5281, opened 2026-03-18
Use an iOS 26 close button and re-order some sections in SettingsScreen.

### LINKED ISSUE #5207: Global Settings: Update menu items
### What do you want to do?

1. Let's update the ordering of menu items shown in global Settings so that the most important options are at the top..
   - Top of list: Account, devices, blocked users
   - Middle: Notification, Screen lock, Encryption
   - Last: Everything else (no change)

2. While we're at it, can we...
   - Combine "Manage account" and "Manage devices" since they both take you to your external account
   - Rename "Manage account" to "Manage account & devices"
   - In the future we will introduce "Link new device" and possibly other features (like "Share your contact info")

### Why do you want to do this?

Currently Settings priorities Notification, Screen lock, and Encryption, which are not used very often. This change priorities the management of user account, devices, and blocked users, which is more important / more frequently used.

### Links

- [iOS Settings screen](https://www.figma.com/design/jvZ7TIyd7V9Vn8sIDB9vMd/Compound-iOS-Templates?node-id=3792-185312&t=sDLXBPuYg5jv8JGZ-4)
- [Android Settings screen](https://www.figma.com/design/pDlJZGBsri47FNTXMnEdXB/Compound-Android-Templates?node-id=8-20896&t=GEqBcUQr7BtQ4vJT-4)

americanrefugee:
Swaps the manage app section with manage account section.

Also merges Manage account and Manage devices.

Closes #5207


ElementX/Sources/Screens/Settings/SettingsScreen/View/SettingsScreen.swift
```diff
@@ -247,7 +247,7 @@ struct SettingsScreen: View {
     
     private var toolbar: some ToolbarContent {
         ToolbarItem(placement: .confirmationAction) {
```
stefanceriu: I believe this should be a `.cancellationAction`

```
/// Cancellation actions dismiss the modal interface without taking any
/// action, usually by tapping or clicking a Cancel button.
```

Velin92: I think however @americanrefugee  wants to keep the the close action on the right, so instead of using either .confirmation or .cancellation action use .primaryAction instead which is more generic and can be used  for both and will render on the right side by default

PR #5284, opened 2026-03-19
PillView and BadgeLabel design tweaks following compound updates

stefanceriu:
* the PillView now uses `textOnSolidPrimary` instead of `textPrimary` when not own mention
* the BadgeLabel now has a border for the default style


PreviewTests/Sources/__Snapshots__/PreviewTests/badgeLabel.iPhone-en-GB-0.png
bmarty, line 1: Do I see a glitch on the border sides here?

stefanceriu: They're just snapshotting artefacts, they don't show up in the real app.

PR #5305, opened 2026-03-25
Add support for starting voice calls from a DM

### LINKED ISSUE #5092: [Task] Add support for starting voice call from DM

When in a DM, the unique call button 🎥 should be replace with 2 different buttons 📞 🎥 .
One for voice only call and one for video call.

This is controlled by using a specific url parameter intent [StartCallDmVoice](https://github.com/matrix-org/matrix-rust-sdk/blob/70665a84aa995f6569f6ebd760edc1941d0f35b0/crates/matrix-sdk/src/widget/settings/element_call.rs#L146) 

EXA UX for ref

This issue only covers starting a voice call. 
The receiving part is another issue

BillCarsonFr:
### Pull Request Checklist

Depends on https://github.com/element-hq/element-x-ios/pull/5298
Fixes https://github.com/element-hq/element-x-ios/issues/5092

image

- [ ] Voiceover enabled.


ElementX/Sources/Services/ElementCall/ElementCallServiceProtocol.swift
```diff
@@ -10,7 +10,7 @@ import Combine
 
 enum ElementCallServiceAction {
     case receivedIncomingCallRequest
-    case startCall(roomID: String, isVoiceCall: Bool)
+    case startCall(roomID: String, voiceCall: Bool)
```
pixlwave: Honestly this seems like a weird change to me. `isVoiceCall` is a more Swifty name for a bool that belongs to something, personally I would use that everywhere.

ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
```diff
@@ -62,6 +63,9 @@ struct RoomScreenViewState: BindableState {
         isCallingEnabled && !isParticipatingInOngoingCall // Hide the join call button when already in the call
     }
     
+    /// Wheter the current room is a DM
```
pixlwave:
```suggestion
    /// Whether the current room is a DM
```

ElementX/Sources/Screens/RoomScreen/RoomScreenModels.swift
```diff
@@ -28,6 +28,7 @@ enum RoomScreenViewAction {
     case viewAllPins
     case displayRoomDetails
     case displayCall
+    case displayVoiceCall
```
pixlwave: It seems like it would be both simpler to handle and more consistent if these were also combined as a single action with a Bool.

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
```diff
@@ -250,7 +267,7 @@ struct RoomScreen_Previews: PreviewProvider, TestablePreview {
     static func makeViewModels(canSendMessage: Bool = true, hasSuccessor: Bool = false) -> ViewModels {
         let roomProxyMock = JoinedRoomProxyMock(.init(id: "stable_id",
                                                       name: "Preview room",
-                                                      hasOngoingCall: true,
+                                                      hasOngoingCall: false,
```
pixlwave: Was this intentional to commit? It might be worth adding it as a parameter, default it to `true` and then make one specific preview set it as `false` so that we don't need to change too many snapshots?

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
```diff
@@ -195,21 +192,41 @@ struct RoomScreen: View {
...
+                    .disabled(!context.viewState.canJoinCall)
+                }
+            }
+            ToolbarItem(placement: .primaryAction) {
+                Button {
+                    context.send(viewAction: .displayCall)
+                } label: {
+                    CompoundIcon(\.videoCallSolid)
+                }
+                .accessibilityLabel(L10n.a11yStartCall)
```
pixlwave: Do you think this is clear enough for VoiceOver users, or should we make it say "Start video call" now?

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
```diff
@@ -172,11 +172,8 @@ struct RoomScreen: View {
         }
         
         if !ProcessInfo.processInfo.isiOSAppOnMac {
-            ToolbarItem(placement: .primaryAction) {
-                if context.viewState.shouldShowCallButton {
-                    callButton
-                        .disabled(!context.viewState.canJoinCall)
-                }
+            if context.viewState.shouldShowCallButton {
+                callControls
```
pixlwave: As you have it on all 3 buttons you should be able to keep the `.disabled` here and remove the 3 duplicate uses
```suggestion
                callControls
                    .disabled(!context.viewState.canJoinCall)
```

ElementX/Sources/Screens/RoomMemberDetailsScreen/RoomMemberDetailsScreenModels.swift
```diff
@@ -94,6 +95,7 @@ enum RoomMemberDetailsScreenViewAction {
     case openDirectChat
     case createDirectChat
     case startCall(roomID: String)
+    case startVoiceCall(roomID: String)
```
pixlwave: Same comment, let's stay consistent and combine them if we can?

ElementX/Sources/Screens/RoomMemberDetailsScreen/RoomMemberDetailsScreenModels.swift
```diff
@@ -12,6 +12,7 @@ enum RoomMemberDetailsScreenViewModelAction {
     case openUserProfile
     case openDirectChat(roomID: String)
     case startCall(roomProxy: JoinedRoomProxyProtocol)
+    case startVoiceCall(roomProxy: JoinedRoomProxyProtocol)
```
pixlwave: And again

ElementX/Sources/Screens/RoomMemberDetailsScreen/RoomMemberDetailsScreenCoordinator.swift
```diff
@@ -21,6 +21,7 @@ enum RoomMemberDetailsScreenCoordinatorAction {
     case openUserProfile
     case openDirectChat(roomID: String)
     case startCall(roomProxy: JoinedRoomProxyProtocol)
+    case startVoiceCall(roomProxy: JoinedRoomProxyProtocol)
```
pixlwave: Ditto.

ElementX/Sources/Other/AccessibilityIdentifiers.swift
```diff
@@ -157,6 +157,8 @@ enum A11yIdentifiers {
         let attachmentPickerTextFormatting = "room-attachment_picker_text_formatting"
         let timelineItemActionMenu = "room-timeline_item_action_menu"
         let joinCall = "room-join_call"
+        let startCall = "room-start_call"
+        let startVoiceCall = "room-start_call"
```
pixlwave:
```suggestion
        let startVoiceCall = "room-start_voice_call"
```

ElementX/Sources/FlowCoordinators/SpacesTabFlowCoordinator.swift
```diff
@@ -178,7 +178,7 @@ class SpacesTabFlowCoordinator: FlowCoordinatorProtocol {
                 guard let self else { return }
                 
                 switch action {
-                case .presentCallScreen(let roomProxy):
+                case .presentCallScreen(let roomProxy, let voiceCall):
                     actionsSubject.send(.presentCallScreen(roomProxy: roomProxy))
```
pixlwave: It looks like this sent action should include the Bool on it now too?

ElementX/Sources/FlowCoordinators/SpaceSettingsFlowCoordinator.swift
```diff
@@ -395,7 +395,7 @@ final class SpaceSettingsFlowCoordinator: FlowCoordinatorProtocol {
             switch action {
             case .finished:
                 stateMachine.tryEvent(.stopMembersListFlow)
-            case .presentCallScreen(let roomProxy):
+            case .presentCallScreen(let roomProxy, let voiceCall):
                 actionsSubject.send(.presentCallScreen(roomProxy: roomProxy))
```
pixlwave, line 399: Ditto

ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
```diff
@@ -582,7 +582,7 @@ class SpaceFlowCoordinator: FlowCoordinatorProtocol {
                     stateMachine.tryEvent(.leftSpace)
                 }
             case .presentCallScreen(let roomProxy):
-                actionsSubject.send(.presentCallScreen(roomProxy: roomProxy))
+                actionsSubject.send(.presentCallScreen(roomProxy: roomProxy, voiceCall: false))
```
pixlwave: And then it can be handled here?

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -717,7 +717,7 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
                 case .presentMessageForwarding(let forwardingItem):
                     stateMachine.tryEvent(.presentMessageForwarding(forwardingItem: forwardingItem))
                 case .presentCallScreen:
-                    actionsSubject.send(.presentCallScreen(roomProxy: roomProxy))
+                    actionsSubject.send(.presentCallScreen(roomProxy: roomProxy, voiceCall: false))
```
pixlwave: This can pass the parameter from `.presentCallScreen` rather than hard coding it to `false`.

pixlwave (approved):
Very nice, I'm sure a lot of people are going to be happy to see this land 👏

Just little comments, nothing major 👍

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
BillCarsonFr: The thing is that the button is saying Call on figma. And not Video. I'd put video in voice other if we also put video on the label

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
BillCarsonFr: I tried that but callControls is a ToolbarContent and this do not have a disabled modifier

ElementX/Sources/Other/AccessibilityIdentifiers.swift
BillCarsonFr: good catch

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
pixlwave: Looking at it a bit closer I think this would be worth checking with Design. Blind users won't be able to see the icon for context and right now:

- The "Call" button on `RoomMemberDetailsScreen` starts an audio call.
- The "Start Call" button on `RoomScreen` starts a video call.

That seems far from ideal.

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
pixlwave: Ah, my bad I didn't realise that 🤦‍♂️

pixlwave (commented):
Code looks good, but you'll need updated snapshots for CI to be happy.

BillCarsonFr:
I am failing to run the screenshot tests

ElementX/Sources/Screens/RoomScreen/View/RoomCallControlsToolbar.swift
pixlwave, line 1: Very nice 👌

pixlwave (approved):

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
```diff
@@ -195,21 +195,41 @@ struct RoomScreen: View {
         }
     }
     
-    @ViewBuilder
-    private var callButton: some View {
+    @ToolbarContentBuilder
+    private var callControls: some ToolbarContent {
```
pixlwave: I think we can get rid of these controls now.

BillCarsonFr: arg right :/
/me stupid 
I expected xCoDe to tell me that there was some dead code ;)

pixlwave: ha, sadly no. There's a scheme called Periphery in the project to detect dead code but we only run that every 3-6 months or so.

PR #5310, opened 2026-03-26
Use single typeText call to speed up tests execution

manuroe:
## Summary
- Replace character-by-character typing with a single typeText call in clearAndTypeText
- Keep note documenting prior CI flakiness context

## Testing
- The CI will tell --> It fails but for already existing reasons.


manuroe:
This change comes from an internal [discussion](https://matrix.to/#/!NMoyFpQxIVYaQIfQHG:matrix.org/$-64FFIVrFEYR7k4MiATOz-6gWzboBypptnkFQap9gBY?via=matrix.org&via=element.io&via=jki.re).

Velin92 (approved):

PR #5312, opened 2026-03-27
Notification Decoration Revised

amshakal:
1. I removed the grey dots for when there is activity in rooms set to mentions or mute. I
2. For all unread rooms, I made the message preview text bold for better legibility
3. For rooms with no unreads, I made the font weight regular, again for legibility

Screenshot 2026-03-30 at 16 31 26


mxandreas:
@amshakal I think this is safe to try. I personally do not "effectively" see a difference, but it may be closer to what you see in other apps. The biggest argument is that Web is already doing that and having a different approach on mobile is probably the most confusing option.

(I also rebased the branch to get it compile against the latest RustSDK)

stefanceriu:
> to get it compile against the latest RustSDK

Why did you need the latest sdk? 🤔 It should just build directly after checkout..

PR #5313, opened 2026-03-27
Floating Date Pill

Velin92:
Essentially I just implemented a floating date system in EXI, a bit like WA, I used some of the existing contribution code of EXA, to base myself for this implementation, but added a few quirks to make it moe WA style:

- It automatically hides after 2 seconds the scrolling has stopped
- Reactively positions below any banner that is displayed at the top of the view

https://github.com/user-attachments/assets/51cc1dbb-1032-4ef8-9345-fd2f160443a6


ElementX/Sources/Screens/Timeline/View/FloatingDateBadge.swift
```diff
@@ -0,0 +1,58 @@
...
+    var body: some View {
+        ZStack {
+            if let dateText {
+                Text(dateText)
+                    .font(.compound.bodySMSemibold)
+                    .foregroundColor(.compound.textPrimary)
+                    .padding(.horizontal, 16)
+                    .padding(.vertical, 8)
+                    .background(Capsule()
+                        .fill(backgroundColor))
```
pixlwave: Super minor, but I'm starting to favour this when no stroke is involved
```suggestion
                    .background(backgroundColor, in: .capsule)
```

ElementX/Sources/Screens/Timeline/TimelineTableViewController.swift
```diff
@@ -505,6 +515,63 @@ extension TimelineTableViewController: UITableViewDelegate {
...
+            return
+        }
+        
+        // Before updating it already schedule it's removal or the future
+        scheduleFloatingDateHide()
+        
+        // Only update on changes to avoid needless SwiftUI recomputation.
+        if floatingDateText != dateText {
+            floatingDateText = dateText
+        }
```
pixlwave: Just to double check, presumably it is intentional that the scheduling doesn't happen inside the if? Might be worth adding a comment if so.

ElementX/Sources/Screens/Timeline/TimelineTableViewController.swift
```diff
@@ -505,6 +515,63 @@ extension TimelineTableViewController: UITableViewDelegate {
...
+        floatingDateHideWorkItem = workItem
+        DispatchQueue.main.asyncAfter(deadline: .now() + 2.0, execute: workItem)
+    }
+    
+    /// Returns the formatted date text for the topmost visible timeline item.
+    ///
+    /// The table view is flipped (unless VoiceOver is running), so the "topmost"
+    /// visible cell on screen is actually the *last* index path in `indexPathsForVisibleRows`
+    /// when the table is flipped, or the *first* when it is not.
+    private func topmostVisibleDateText() -> String? {
```
pixlwave: Can we call this `newestVisibleDateText` to match `newestVisibleItemID` so that we don't have to think about whether or not this "top" is pre or post-flip?

```suggestion
    private func newestVisibleDateText() -> String? {
```

ElementX/Sources/Screens/Timeline/TimelineTableViewController.swift
```diff
@@ -505,6 +515,63 @@ extension TimelineTableViewController: UITableViewDelegate {
...
+        }
+        
+        // In a flipped table view the last index path is the topmost item on screen;
+        // when VoiceOver is active the table is not flipped so the first is topmost.
+        let isFlipped = scaleY == -1
+        let orderedPaths = isFlipped ? visibleIndexPaths.reversed() : visibleIndexPaths
+        
+        // Walk from topmost downward and return the date of the first item that has one.
+        for indexPath in orderedPaths {
+            if let uniqueID = dataSource?.itemIdentifier(for: indexPath),
```
pixlwave, line 567: Should/can we use `newestVisibleItemID` here to reduce what might be similar code?

ElementX/Sources/Other/SwiftUI/Views/TopBannerModifier.swift
```diff
@@ -7,20 +7,53 @@
 
 import SwiftUI
 
+/// A single banner item to be displayed in a `topBanners` overlay.
+struct TopBannerItem {
+    var banner: AnyView
```
pixlwave, line 12: I think this is all we would need to not type erase the view so that its identity remains stable?

```suggestion
struct TopBannerItem<Banner: View> {
    var banner: Banner
```

ElementX/Sources/Other/SwiftUI/Views/TopBannerModifier.swift
```diff
@@ -7,20 +7,53 @@
...
+    /// slides from the top edge based on its own `isVisible` flag. Later items in
+    /// the array are overlayed on top of earlier ones. The footer is shared across
+    /// all banners and displayed below the topmost visible banner.
+    func topBanners(_ items: [TopBannerItem], footer: some View = EmptyView()) -> some View {
+        let anyBannerVisible = items.contains { $0.isVisible }
+        return overlay(alignment: .top) {
+            ZStack(alignment: .top) {
+                // Visible layout
+                VStack(spacing: 0) {
+                    ZStack(alignment: .top) {
```
pixlwave, line 38: Did you try it without this `ZStack` now that the view is full height?

pixlwave (approved):
Very nice 👌

ElementX/Sources/Screens/Timeline/TimelineTableViewController.swift
Velin92: yeah because every time we try to attend an update, it means it should still be shown (even if is not really getting updated), so the disappearence needs to be rescheduled.

ElementX/Sources/Screens/Timeline/TimelineTableViewController.swift
Velin92: sure thing

ElementX/Sources/Other/SwiftUI/Views/TopBannerModifier.swift
Velin92: doing that makes the animations a bit funky (like render them with a small clip and so on),, this seems instead to account for all the animations, I imagine this is because by doing so we reserve all the space behind without ever changing it  (in the other case, is all the space but is mutating the clear size), and makes the animations that are happening on top of it more stable

ElementX/Sources/Screens/Timeline/TimelineTableViewController.swift
Velin92: it's a bit different, in te first case we get the newest one we can see that is not a pagination indicator (and return nil in this case), instead here we get all the visible ones, check the topmost/latter and if it's not a valid date based item, we check the next one and so on.

ElementX/Sources/Other/SwiftUI/Views/TopBannerModifier.swift
Velin92: would make working with the array of different types impossible

pixlwave: Ah yeah, I remember you saying that when we talked about it now 🤦‍♂️

PR #5319, opened 2026-03-30
Add multiple window support for mac and iPad OS

stefanceriu:
This takes avantage of our neatly decoupled architecture to do the following:

Once the app starts the WindowManager is configured with SwiftUI's environment `OpenWindowAction`. It can then be used to register coordinators (that provide the toPresentable view) and an optional flow coordinator (as most of the screens are part of a flow, especially rooms). 
Once a coordinator is registed, the WindowManager invokes the `OpenWindowAction` which in turn makes the Application call its newly introduced `WindowManagerWindowType` `WindowGroup`'s block to instantiate a new visual window rooting that view.

The WindowManager is also responsible for wrapping the presentable in a disappearance block and clean up the coordinator stack. 

The rest lies in various tweaks like disabling multi window when the pin lock is enabled, closing windows when leaving rooms, focusing the main window when opening global search (cmd+k) etc.

As presented in http://youtube.com/watch?v=Bo7GaLRhtc0&t=1790s

Screenshot 2026-03-31 at 19 33 24

### DIFF
--- a/ElementX/Sources/Application/Windowing/WindowManager.swift
+++ b/ElementX/Sources/Application/Windowing/WindowManager.swift
@@ -20,6 +20,7 @@ class WindowManager: SecureWindowManagerProtocol {
     private(set) var alternateWindow: UIWindow!
     
     private(set) var openWindowAction: OpenWindowAction!
+    private(set) var dismissWindowAction: DismissWindowAction!
         
     var windows: [UIWindow] {
         [mainWindow, overlayWindow, globalSearchWindow, alternateWindow]
@@ -58,13 +59,23 @@ class WindowManager: SecureWindowManagerProtocol {
         delegate?.windowManagerDidConfigureWindows(self)
     }
     
-    func configure(with openWindowAction: OpenWindowAction) {
+    func configure(withOpenWinddowAction openWindowAction: OpenWindowAction,
+                   dismissWindowAction: DismissWindowAction) {
         self.openWindowAction = openWindowAction
+        self.dismissWindowAction = dismissWindowAction
     }
     
     func registerCoordinator(_ coordinator: CoordinatorProtocol, flowCoordinator: FlowCoordinatorProtocol?, forWindowType type: WindowManagerWindowType) {
         coordinators[type] = (coordinator, flowCoordinator)
         openWindowAction(value: type)
+    }
+    
+    func closeAllAuxiliaryWindows() {
+        for key in coordinators.keys {
+            dismissWindowAction(value: key)
+        }
+        
+        coordinators.removeAll()
     }
     
     func windowForType(_ type: WindowManagerWindowType) -> AnyView {
--- a/ElementX/Sources/Application/Windowing/WindowManager.swift
+++ b/ElementX/Sources/Application/Windowing/WindowManager.swift
@@ -74,31 +74,6 @@ class WindowManager: SecureWindowManagerProtocol {
         self.dismissWindowAction = dismissWindowAction
     }
     
-    func registerCoordinator(_ coordinator: CoordinatorProtocol, flowCoordinator: FlowCoordinatorProtocol?, forWindowType type: WindowManagerWindowType) {
-        coordinators[type] = (coordinator, flowCoordinator)
-        openWindowAction(value: type)
-    }
-    
-    func closeAllAuxiliaryWindows() {
-        for key in coordinators.keys {
-            dismissWindowAction(value: key)
-        }
-        
-        coordinators.removeAll()
-    }
-    
-    func windowForType(_ type: WindowManagerWindowType) -> AnyView {
-        guard let coordinator = coordinators[type]?.coordinator else {
-            return AnyView(InstantlyDismissingWindow())
-        }
-        
-        // This behaves strangely and gets called late but cleans up enough
-        // and is self contained enough to be just good .. enough
-        return AnyView(coordinator.toPresentable().onDisappear { [weak self] in
-            self?.coordinators[type] = nil
-        })
-    }
-    
     func handleRoute(_ appRoute: AppRoute, windowType: WindowManagerWindowType) {
         if let flowCoordinator = coordinators[windowType]?.flowCoordinator {
             flowCoordinator.handleAppRoute(appRoute, animated: true)
@@ -167,6 +142,8 @@ class WindowManager: SecureWindowManagerProtocol {
         mainWindow.makeKey()
     }
     
+    // MARK: - OrientationManager
+    
     func setOrientation(_ orientation: UIInterfaceOrientationMask) {
         mainScene?.requestGeometryUpdate(.iOS(interfaceOrientations: orientation))
     }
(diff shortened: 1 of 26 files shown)


ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
```diff
@@ -24,12 +25,12 @@ enum HomeScreenViewModelAction {
     case presentSettingsScreen
     case presentFeedbackScreen
     case presentStartChatScreen
-    case presentGlobalSearch
     case logout
 }
 
 enum HomeScreenViewAction {
     case selectRoom(roomIdentifier: String)
+    case detatchRoom(roomIdentifier: String)
```
pixlwave: I think it's just this one and the one above in this file, but there might be more instances of this.
```suggestion
    case detachRoom(roomIdentifier: String)
```

ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomList.swift
```diff
@@ -1,7 +1,7 @@
-//
 // Copyright 2025 Element Creations Ltd.
 // Copyright 2023-2025 New Vector Ltd.
 //
+//
```
pixlwave: Is this Claude at work…

ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomList.swift
```diff
@@ -34,7 +34,18 @@ struct HomeScreenRoomList: View {
                 let isSelected = context.viewState.selectedRoomID == room.id
                 
                 HomeScreenRoomCell(room: room, isSelected: isSelected, mediaProvider: context.mediaProvider, action: context.send)
+                    .simultaneousGesture(TapGesture(count: 2).onEnded {
+                        context.send(viewAction: .detatchRoom(roomIdentifier: room.id))
+                    })
                     .contextMenu {
+                        if !UIDevice.current.isPhone {
```
pixlwave: Did you try using `@Environment(\.supportsMultipleWindows)` on this instead?

I'm thinking e.g. iPhone Fold might mess this check up somewhat.

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -122,18 +122,22 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
         case .accountProvisioningLink:
             break // We always ignore this flow when logged in.
         case .settings, .chatBackupSettings:
-            if stateMachine.state != .settingsScreen {
-                stateMachine.tryEvent(.showSettingsScreen)
+            if ProcessInfo.processInfo.isiOSAppOnMac {
+                startSettingsFlow(detached: true)
```
pixlwave, line 126: Presumably this isn't going through the state machine because we're considering it a separate thing?

ElementX/Sources/Application/Windowing/WindowManager.swift
```diff
@@ -11,13 +11,25 @@ import SwiftUI
...
     private(set) var overlayWindow: UIWindow!
     private(set) var globalSearchWindow: UIWindow!
     private(set) var alternateWindow: UIWindow!
+    
+    private(set) var openWindowAction: OpenWindowAction!
+    private(set) var dismissWindowAction: DismissWindowAction!
+    
+    var auxiliaryWindowsEnabled = true {
+        didSet {
+            if auxiliaryWindowsEnabled == false {
```
pixlwave: I feel like this is a lint/format warning:

```suggestion
            if !auxiliaryWindowsEnabled {
```

ElementX/Sources/Application/Windowing/WindowManager.swift
```diff
@@ -96,8 +130,15 @@ class WindowManager: SecureWindowManagerProtocol {
             return
         }
         
+        if let mainSession {
+            let request = UISceneSessionActivationRequest(session: mainSession)
+            UIApplication.shared.activateSceneSession(for: request) { error in
+                MXLog.error("Failed to focus window with error: \(error)")
+            }
+        }
```
pixlwave, line 138: Oh that's cool 🤩

ElementX/Sources/Application/Windowing/WindowManager.swift
```diff
@@ -109,13 +150,51 @@ class WindowManager: SecureWindowManagerProtocol {
...
+    func windowForType(_ type: WindowManagerWindowType) -> AnyView {
+        guard let coordinator = coordinators[type]?.coordinator else {
+            return AnyView(InstantlyDismissingWindow())
+        }
+        
+        // This behaves strangely and gets called late but cleans up enough
+        // and is self contained enough to be just good .. enough
+        return AnyView(coordinator.toPresentable().onDisappear { [weak self] in
+            self?.coordinators[type] = nil
+        })
```
pixlwave, line 174: I'm still somewhat concerned this is going to lead to the coordinator being deallocated when you push a child view (like the room details).

ElementX/Sources/Application/Windowing/WindowManagerProtocol.swift
pixlwave, line 1: I feel like there might be a clearer name that helps differentiate a normal UIWindow window from an Application WindowGroup window. Auxiliary doesn't quite communicate that to me. I would've said WindowScene but we've already got a UIWindowScene in here, what a mess Apple's names are 🙈

I also think `WindowManagerWindowType` could probably be improved but likely based on 👆

pixlwave (approved):
Overall looks great to me, but let's wait to merge this until after we've made the RC :)

ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomList.swift
```diff
@@ -34,7 +34,18 @@ struct HomeScreenRoomList: View {
...
+                            Button {
+                                context.send(viewAction: .detatchRoom(roomIdentifier: room.id))
+                            } label: {
+                                Label("Open in new window", icon: \.popOut)
+                            }
```
pixlwave: I went over this with Aaron and we concluded on the following changes:
- Use the `\.spotlight` icon.
- Move the button beneath the Mark As Read action.

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
stefanceriu: 🤦‍♂️

ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomList.swift
stefanceriu: I didn't actually use claude on this one, no idea what happened 🤔

ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomList.swift
stefanceriu: Good idea 👍

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
stefanceriu: Yes, exactly. It lives outside of the HomeScreen/MainWindow state machine

ElementX/Sources/Application/Windowing/WindowManager.swift
stefanceriu: We do both in a bunch of places but yeah, maybe we should agree on one format. I personally like this one, it's harder to get it wrong.

ElementX/Sources/Application/Windowing/WindowManagerProtocol.swift
stefanceriu: I know what you're saying but I honestly don't have any better ideas 😬

ElementX/Sources/Application/Windowing/WindowManager.swift
stefanceriu: Well I tested it on both platforms and it seems top do the right thing so ¯\_(ツ)_/¯

ElementX/Sources/Application/Windowing/WindowManagerProtocol.swift
stefanceriu: We ended up going with "Secondary windows"

PR #5322, opened 2026-03-30
Live Location Timeline expiring timer check

Velin92:
The issue is that, if a user sends the live location update, but never sends any other update or the stop event, the item remains in a stagnant state, were is never recomputed and hence never updates its isLive value.

So we now schedule a timed publisher on the timeline item if is live, to check again the state of it.


ElementX/Sources/Other/Extensions/Date.swift
```diff
@@ -85,6 +85,18 @@ extension Date {
...
+        // swiftlint:disable:next force_unwrap
+        Calendar.current.date(bySettingHour: 4, minute: 10, second: 0, of: .now)!
+    }
+    
+    /// A fixed date representing today at 4:30 AM, used for mocks and previews.
+    static var mockToday430: Date {
+        // swiftlint:disable:next force_unwrap
+        Calendar.current.date(bySettingHour: 4, minute: 30, second: 0, of: .now)!
+    }
+
```
stefanceriu: I would've understood a different hour, meridian or seconds but why do we need 2 this similar to each other?

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/LiveLocationRoomTimelineView.swift
```diff
@@ -5,12 +5,41 @@
...
+        }
+        
+        return Just(())
+            .delay(for: .seconds(delay), scheduler: DispatchQueue.main)
+            .eraseToAnyPublisher()
+    }
+    
+    private var isLive: Bool {
+        timelineItem.content.isLive && !hasExpired
+    }
```
stefanceriu, line 42: Should this live in the view state? Perhaps as a binding you can listen to and with the actual timer logic in the view model?

ElementX/Sources/Other/Extensions/Date.swift
Velin92: testing purposes, one is to test the expiration, and the other is to test the time before expiration.
I use 4:20 as the expiration time, so to test a valid timeline item, I use 4:10 that has 10 mins left, and to test the expired one I use 4:30

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/LiveLocationRoomTimelineView.swift
Velin92: it's a timeline item so it doesn't have an observable context, we use the environment context injection (to solve the issue of not having a full update of the timeline item for any change that is not related to it). 

Also this is a failsafe logic for the view itself, to essentially check again the expiration state when such time is reached, in case for any reason an update to the timeline item doesn't arrive after expiration (otherwise the user would need to close and reopen the timeline).
Don't really see the need of having it anywhere else than in the view.

stefanceriu: > it's a timeline item so it doesn't have an observable context

Bah right, forgot about all of that. Fair then 👍

ElementX/Sources/Other/Extensions/Date.swift
stefanceriu: Talked about it offline, we'll move them to a private view extension

Velin92: @copilot please move mockToday410, mockToday430 and mockToday420 to a private extension Date inside the LiveLocationRoomTimelineView

PR #5326, opened 2026-03-31
LiveLocationManager permission handling

Velin92:
For now this PR only creates the basic skeleton for LiveLocationManager and implement the permission handling when trying to share the live location, and the fallbacks in case there no "always" permission set.

### DIFF
--- a/ElementX/Sources/Application/Settings/AppSettings.swift
+++ b/ElementX/Sources/Application/Settings/AppSettings.swift
@@ -48,7 +48,7 @@ final class AppSettings {
         case analyticsConsentState
         case hasRunNotificationPermissionsOnboarding
         case hasRunIdentityConfirmationOnboarding
-        case hasRequestedAlwaysLocationAuthorization
+        case hasRequestedLocationAlwaysLocationAuthorization
         
         case frequentlyUsedSystemEmojis
         
@@ -344,8 +344,8 @@ final class AppSettings {
     @UserPreference(key: UserDefaultsKeys.hasRunIdentityConfirmationOnboarding, defaultValue: false, storageType: .userDefaults(store))
     var hasRunIdentityConfirmationOnboarding
     
-    @UserPreference(key: UserDefaultsKeys.hasRequestedAlwaysLocationAuthorization, defaultValue: false, storageType: .userDefaults(store))
-    var hasRequestedAlwaysLocationAuthorization
+    @UserPreference(key: UserDefaultsKeys.hasRequestedLocationAlwaysLocationAuthorization, defaultValue: false, storageType: .userDefaults(store))
+    var hasRequestedLocationAlwaysLocationAuthorization
     
     @UserPreference(key: UserDefaultsKeys.frequentlyUsedSystemEmojis, defaultValue: [FrequentlyUsedEmoji](), storageType: .userDefaults(store))
     var frequentlyUsedSystemEmojis
--- a/ElementX/Sources/Screens/LocationSharing/LocationSharingScreenViewModel.swift
+++ b/ElementX/Sources/Screens/LocationSharing/LocationSharingScreenViewModel.swift
@@ -8,7 +8,7 @@
 
 import Combine
 import Foundation
-import UIKit
+import SwiftUI
 
 typealias LocationSharingScreenViewModelType = StateStoreViewModelV2<LocationSharingScreenViewState, LocationSharingScreenViewAction>
 
@@ -114,8 +114,8 @@ class LocationSharingScreenViewModel: LocationSharingScreenViewModelType, Locati
     
     private func startLiveLocationSharing() {
         authorizationStatusSubscription = nil
-        let authStatus = liveLocationManager.authorizationStatus.value
-        switch authStatus {
+        let authorizationStatus = liveLocationManager.authorizationStatus.value
+        switch authorizationStatus {
         case .authorizedAlways:
             // TODO: Start sending live location updates to the room
             break
@@ -123,7 +123,7 @@ class LocationSharingScreenViewModel: LocationSharingScreenViewModelType, Locati
             // This is to solve a race condition with map libre which always tries first
             // to request the when in use permission, we wait for it and then try again
             authorizationStatusSubscription = liveLocationManager.authorizationStatus
-                .filter { $0 != authStatus } // skip current status
+                .filter { $0 != authorizationStatus } // skip current status
                 .first() // this publisher only fires when there is an actual change, and if the user is done with permissions
                 .sink { [weak self] newValue in
                     guard newValue == .authorizedWhenInUse else { return }
@@ -137,7 +137,7 @@ class LocationSharingScreenViewModel: LocationSharingScreenViewModelType, Locati
             }
             
             authorizationStatusSubscription = liveLocationManager.authorizationStatus
-                .filter { $0 != authStatus } // skip current status
+                .filter { $0 != authorizationStatus } // skip current status
                 .first() // this publisher only fires when there is an actual change, and if the user is done with permissions
                 .sink { newValue in
                     guard newValue == .authorizedAlways else { return }
@@ -149,10 +149,9 @@ class LocationSharingScreenViewModel: LocationSharingScreenViewModelType, Locati
     }
     
     private func showMissingAlwaysAuthorizedAlert() {
-        let action: () -> Void = { [weak self] in self?.actionsSubject.send(.openSystemSettings) }
         state.bindings.alertInfo = .init(locationSharingViewError: .missingAlwaysAuthorization,
                                          primaryButton: .init(title: L10n.actionNotNow, role: .cancel, action: nil),
-                                         secondaryButton: .init(title: L10n.commonSettings, action: action))
+                                         secondaryButton: .init(title: L10n.commonSettings) { [weak self] in self?.actionsSubject.send(.openSystemSettings) })
     }
     
     private func sendLocation(_ geoURI: GeoURI, isUserLocation: Bool) async {
--- a/ElementX/Sources/Services/Location/LiveLocationManager.swift
+++ b/ElementX/Sources/Services/Location/LiveLocationManager.swift
@@ -39,8 +39,8 @@ class LiveLocationManager: NSObject, LiveLocationManagerProtocol, CLLocationMana
     
     @discardableResult
     func requestAlwaysAuthorizationIfPossible() -> Bool {
-        guard !appSettings.hasRequestedAlwaysLocationAuthorization else { return false }
-        appSettings.hasRequestedAlwaysLocationAuthorization = true
+        guard !appSettings.hasRequestedLocationAlwaysLocationAuthorization else { return false }
+        appSettings.hasRequestedLocationAlwaysLocationAuthorization = true
         locationManager.requestAlwaysAuthorization()
         return true
     }
@@ -51,14 +51,8 @@ class LiveLocationManager: NSObject, LiveLocationManagerProtocol, CLLocationMana
         // If the system resets authorization to notDetermined (e.g. after app reinstall or
         // settings reset), clear the flag so we can request again.
         if manager.authorizationStatus == .notDetermined {
-            appSettings.hasRequestedAlwaysLocationAuthorization = false
+            appSettings.hasRequestedLocationAlwaysLocationAuthorization = false
         }
         authorizationStatusSubject.send(manager.authorizationStatus)
     }
-    
-    // MARK: - Private
-    
-    // TODO: Add CLLocationManager location update handling to forward updates to rooms
-    // TODO: Track which rooms are currently sharing live location
-    // TODO: Send location updates to all active rooms via clientProxy
 }
--- a/ElementX/Sources/Services/Location/LiveLocationManagerProtocol.swift
+++ b/ElementX/Sources/Services/Location/LiveLocationManagerProtocol.swift
@@ -11,7 +11,7 @@ import Foundation
 
 // sourcery: AutoMockable
 protocol LiveLocationManagerProtocol: AnyObject {
-    /// Publishes the current "Always" location authorization status.
+    /// Publishes the current location authorization status.
     var authorizationStatus: CurrentValuePublisher<CLAuthorizationStatus, Never> { get }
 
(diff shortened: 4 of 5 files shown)


ElementX/Sources/Screens/LocationSharing/LocationSharingScreenViewModel.swift
```diff
@@ -8,32 +8,41 @@
 
 import Combine
 import Foundation
+import UIKit
```
pixlwave:
```suggestion
import SwiftUI
```

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenViewModel.swift
```diff
@@ -75,6 +86,13 @@ class LocationSharingScreenViewModel: LocationSharingScreenViewModelType, Locati
             self?.updateShownUserProfile(members: members)
         }
         .store(in: &cancellables)
+        
+        notificationCenter.publisher(for: UIApplication.didEnterBackgroundNotification)
+            .sink { [weak self] _ in
+                // Let's remove the subscription if the user backgrounds the app (maybe to change their location settings)
+                self?.authorizationStatusSubscription = nil
+            }
+            .store(in: &cancellables)
```
pixlwave, line 95: Pretty sure the app gets killed if you change privacy settings from the Settings app, so this might not be necessary.

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenViewModel.swift
```diff
@@ -94,6 +112,49 @@ class LocationSharingScreenViewModel: LocationSharingScreenViewModelType, Locati
         }
     }
     
+    private func startLiveLocationSharing() {
+        authorizationStatusSubscription = nil
+        let authStatus = liveLocationManager.authorizationStatus.value
```
pixlwave: We have all the characters :)
```suggestion
        let authorizationStatus = liveLocationManager.authorizationStatus.value
```

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenViewModel.swift
```diff
@@ -94,6 +112,49 @@ class LocationSharingScreenViewModel: LocationSharingScreenViewModelType, Locati
...
+                    guard newValue == .authorizedAlways else { return }
+                    // TODO: Start sending live location updates to the room
+                }
+        default:
+            showMissingAlwaysAuthorizedAlert()
+        }
+    }
+    
+    private func showMissingAlwaysAuthorizedAlert() {
+        let action: () -> Void = { [weak self] in self?.actionsSubject.send(.openSystemSettings) }
```
pixlwave: I would think we can inline this like we do elsewhere?

ElementX/Sources/Services/Location/LiveLocationManagerProtocol.swift
```diff
@@ -0,0 +1,23 @@
...
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Combine
+import CoreLocation
+import Foundation
+
+// sourcery: AutoMockable
+protocol LiveLocationManagerProtocol: AnyObject {
+    /// Publishes the current "Always" location authorization status.
```
pixlwave:
```suggestion
    /// Publishes the current location authorization status.
```

ElementX/Sources/Services/Location/LiveLocationManager.swift
```diff
@@ -0,0 +1,64 @@
...
+    }
+    
+    @MainActor
+    init(clientProxy: ClientProxyProtocol,
+         appSettings: AppSettings) {
+        self.clientProxy = clientProxy
+        self.appSettings = appSettings
+        // Very important, the CLLocationManager needs to be initialised on the main thread
+        // or the delegate functions won't be handled!
+        // https://developer.apple.com/documentation/corelocation/cllocationmanagerdelegate
```
pixlwave, line 29: Would it be worthwhile putting the whole manager on the MainActor?

ElementX/Sources/Services/Location/LiveLocationManager.swift
```diff
@@ -0,0 +1,64 @@
...
+            appSettings.hasRequestedAlwaysLocationAuthorization = false
+        }
+        authorizationStatusSubject.send(manager.authorizationStatus)
+    }
+    
+    // MARK: - Private
+    
+    // TODO: Add CLLocationManager location update handling to forward updates to rooms
+    // TODO: Track which rooms are currently sharing live location
+    // TODO: Send location updates to all active rooms via clientProxy
```
pixlwave: We probably don't need to introduce 3 todo warnings to the project here…

ElementX/Sources/Services/Location/LiveLocationManager.swift
```diff
@@ -0,0 +1,64 @@
...
+    }
+    
+    // MARK: - CLLocationManagerDelegate
+    
+    func locationManagerDidChangeAuthorization(_ manager: CLLocationManager) {
+        // If the system resets authorization to notDetermined (e.g. after app reinstall or
+        // settings reset), clear the flag so we can request again.
+        if manager.authorizationStatus == .notDetermined {
+            appSettings.hasRequestedAlwaysLocationAuthorization = false
+        }
```
pixlwave, line 55: I don't think this is possible. From the docs:

> the location manager doesn’t report the current authorization status to this method — the location manager only reports changes

So this flag reset probably wants to go in the `init`?

ElementX/Sources/Services/Session/UserSessionProtocol.swift
pixlwave, line 1: Ah nice, this is a good home for it 👍

ElementX/Sources/Services/UserSession/UserSessionStore.swift
```diff
@@ -103,16 +103,22 @@ class UserSessionStore: UserSessionStoreProtocol {
...
         let mediaProvider = MediaProvider(mediaLoader: clientProxy.mediaLoader,
                                           imageCache: .onlyInMemory,
                                           homeserverReachabilityPublisher: clientProxy.homeserverReachabilityPublisher)
         
         let voiceMessageMediaManager = VoiceMessageMediaManager(mediaProvider: mediaProvider)
         
+        let liveLocationManager = await MainActor.run {
+            LiveLocationManager(clientProxy: clientProxy,
+                                appSettings: appSettings)
+        }
```
pixlwave, line 116: Given the init is annotated as `@MainActor`, surely we don't need the `MainActor.run` here?

UnitTests/Sources/LocationSharingScreenViewModelTests.swift
```diff
@@ -7,33 +7,26 @@
...
 
 @MainActor
 struct LocationSharingScreenViewModelTests {
-    private let timelineProxy = TimelineProxyMock(.init())
-    private var viewModel: LocationSharingScreenViewModelProtocol
+    var timelineProxy: TimelineProxyMock!
+    var viewModel: LocationSharingScreenViewModel!
     
-    private var context: LocationSharingScreenViewModel.Context {
+    var context: LocationSharingScreenViewModel.Context {
```
pixlwave: I'm not particularly bothered by it, but I'm curious why the `private`s have been removed here?

UnitTests/Sources/LocationSharingScreenViewModelTests.swift
```diff
@@ -7,33 +7,26 @@
...
-                                                       mapURLBuilder: ServiceLocator.shared.settings.mapTilerConfiguration,
-                                                       liveLocationSharingEnabled: true,
-                                                       roomProxy: JoinedRoomProxyMock(.init()),
-                                                       timelineController: MockTimelineController(timelineProxy: timelineProxy),
-                                                       analytics: ServiceLocator.shared.analytics,
-                                                       userIndicatorController: UserIndicatorControllerMock(),
-                                                       mediaProvider: MediaProviderMock(configuration: .init()))
-        viewModel.state.bindings.isLocationAuthorized = true
-        self.viewModel = viewModel
+        AppSettings.resetAllSettings()
```
pixlwave, line 24: More importantly, we need this in a `deinit` too so that other successive tests have the expected defaults.

UnitTests/Sources/LocationSharingScreenViewModelTests.swift
```diff
@@ -120,4 +119,101 @@ struct LocationSharingScreenViewModelTests {
...
+                                                   mapURLBuilder: ServiceLocator.shared.settings.mapTilerConfiguration,
+                                                   liveLocationSharingEnabled: true,
+                                                   roomProxy: JoinedRoomProxyMock(.init()),
+                                                   timelineController: MockTimelineController(timelineProxy: timelineProxy),
+                                                   liveLocationManager: liveLocationManagerMock,
+                                                   analytics: ServiceLocator.shared.analytics,
+                                                   userIndicatorController: UserIndicatorControllerMock(),
+                                                   mediaProvider: MediaProviderMock(configuration: .init()))
+        viewModel.state.bindings.isLocationAuthorized = true
+    }
```
pixlwave: If we make the version above that takes the `LiveLocationManagerMock.Configuration` store the `liveLocationManager` like it does the `timelineProxy`, then we don't need this second version… (the configuration might need to take the publisher as a parameter too)

UnitTests/Sources/LocationSharingScreenViewModelTests.swift
```diff
@@ -120,4 +119,101 @@ struct LocationSharingScreenViewModelTests {
             try await deferred.fulfill()
         }
     }
+    
+    // MARK: - Live Location Authorization Tests
+    
+    @Test
+    mutating func startLiveLocationWithDeniedAuth() {
```
pixlwave:
```suggestion
    mutating func startLiveLocationWithDeniedAuthorization() {
```

And all the ones below too.

ElementX/Sources/Application/Settings/AppSettings.swift
pixlwave, line 1: Minor, but could we call this `hasRequestedLocationAlwaysAuthorization` so it matches the `alwaysAuthorization` name used by CoreLocation?

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenViewModel.swift
Velin92: no sadly it doesn't, app survives the location setting changes, that's why I had to add this. It does get killed for other changes, but not this one.

ElementX/Sources/Services/Location/LiveLocationManager.swift
Velin92: not for now, I think some of its parts are fine to run on other threads, but the initialisation and the initialisation only is mandatory to have here, if I see the need to handle other sutff on the MainActor I'll do that.

ElementX/Sources/Services/Location/LiveLocationManager.swift
Velin92: then Apple really needs to update their docs, because this thing runs IMMEDIATELY as you assign the delegate to the CLLocationManager. Tested it quite extensively the whole evening yesterday on a real device.
Here is a screenshot of the breakpoint as soon as the app build and runs.
Screenshot 2026-03-31 at 11 39 44

ElementX/Sources/Services/UserSession/UserSessionStore.swift
Velin92: we do because the UserSessionStore itself is not annotated with `@MainActor` so this function is not guaranteed to run on it, in fact this was the place where it was complaining after I added the @MainActor conformance, and had to wait for it. Ofc this is super fast, no need to worry about its performance.

UnitTests/Sources/LocationSharingScreenViewModelTests.swift
Velin92: Claude being dumb thanks for noticing, I think is time we introduce "Skills" in our project (a good talk for Friday), the agents.md might not be enough, and the skills would allow us to fine tune specific piece of work the AI does (and would also make our agents.md slender, by becoming an index of skills)

ElementX/Sources/Services/UserSession/UserSessionStore.swift
pixlwave: But the LiveLocationManager is annotated, so the `MainActor.run` does nothing here.

Doing this will have the same effect:

```
let liveLocationManager = await LiveLocationManager(clientProxy: clientProxy, appSettings: appSettings)
```

(the `await` being required because it is going to hop onto the main actor).

ElementX/Sources/Services/Location/LiveLocationManager.swift
pixlwave: *Reacts with pure shock that Apple's documentation is out of date* /s

Velin92: actually I double checked the docs and they do mention that this is called when instantiating the CLLocationManager:

> The system calls this method when the app creates the related object’s [CLLocationManager](https://developer.apple.com/documentation/corelocation/cllocationmanager) instance, and when the app’s authorization status changes. The status informs the app whether it can access the user’s location.

Velin92: The part you are referring to is this

> If the user’s choice doesn’t change the authorization status after you call the [requestWhenInUseAuthorization()](https://developer.apple.com/documentation/corelocation/cllocationmanager/requestwheninuseauthorization()) or [requestAlwaysAuthorization()](https://developer.apple.com/documentation/corelocation/cllocationmanager/requestalwaysauthorization()) method, the location manager doesn’t report the current authorization status to this method—the location manager only reports changes. For example, the location manager calls this method when the status changes from [CLAuthorizationStatus.notDetermined](https://developer.apple.com/documentation/corelocation/clauthorizationstatus/notdetermined) to [CLAuthorizationStatus.authorizedWhenInUse](https://developer.apple.com/documentation/corelocation/clauthorizationstatus/authorizedwheninuse).

Which means... that satus change are only reported when the user is done with selecting the permission option and only if the option is different from the previous existing permission option.

PR #5331, opened 2026-03-31
Prompt user when inviting users with uncached identities

kaylendog:
If the user attempts to invite someone (to a room or creating a DM) whose identity is not cached, we prompt them to make sure this was their intention.

Part of https://github.com/element-hq/element-meta/issues/3163

### Pull Request Checklist

- [ ] Voiceover enabled.


ElementX/Sources/Screens/InviteUsersScreen/View/ConfirmInviteUsersSheet.swift
```diff
@@ -0,0 +1,71 @@
...
+// Copyright 2026 Element Creations Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Compound
+import SwiftUI
+
+struct ConfirmInviteUsersSheetView: View {
```
kaylendog: I am wondering if it would be better to repurpose the existing `SendInviteConfirmationView` to support arbitrary numbers of members, but this seems to suffice for now.

ElementX/Sources/Screens/InviteUsersScreen/InviteUsersScreenViewModel.swift
```diff
@@ -67,12 +77,37 @@ class InviteUsersScreenViewModel: InviteUsersScreenViewModelType, InviteUsersScr
...
+                    return
+                }
+                guard identity == nil else {
+                    // If we have it cached, we implicity trust the user that this was intentional.
+                    return
+                }
+                Task { @MainActor in
+                    // If we do not, we will prompt the user to confirm they meant to invite them.
+                    self.state.usersToConfirm.append(user)
+                }
```
kaylendog: There is a potential race condition here, but I'm not sure how to fix it and since we're only loading from the store, these tasks should be very short-lived.

ElementX/Sources/Screens/InviteUsersScreen/InviteUsersScreenViewModel.swift
```diff
@@ -67,12 +77,37 @@ class InviteUsersScreenViewModel: InviteUsersScreenViewModelType, InviteUsersScr
...
         if state.selectedUsers.contains(user) {
             state.selectedUsers.removeAll { $0.userID == user.userID }
         } else {
             state.selectedUsers.append(user)
             withElementAnimation(.easeInOut) { state.bindings.selectedUsersPosition = user.userID }
+            Task.detached {
+                // Attempt to fetch the cached user identity.
+                guard case let .success(identity) = await self.clientProxy.userIdentity(for: user.userID, fallBackToServer: false) else {
+                    MXLog.error("Failed to get cached user identity for \(user.userID)")
+                    return
```
kaylendog: It might be worth appending here even in the case of failure, like the changes to the other view models do.

kaylendog:
I can't see why the Apple CI is failing, but otherwise I believe this is ready for review.

ElementX/Resources/Localizations/en.lproj/Untranslated.strings
```diff
@@ -14,3 +14,10 @@
 "soft_logout_clear_data_submit" = "Clear all data";
 "soft_logout_clear_data_dialog_title" = "Clear data";
 "soft_logout_clear_data_dialog_content" = "Clear all data currently stored on this device?\nSign in again to access your account data and messages.";
+
+// MARK: - History sharing
+
+"crypto_history_sharing_confirm_invite_dialog_title" = "Invite new contacts to this room?";
+"crypto_history_sharing_confirm_invite_dialog_content" = "You currently don’t have any chats with these contacts. Confirm inviting them to this room before continuing.";
+"crypto_history_sharing_confirm_start_chat_dialog_title" = "Start a chat with this new contact?";
+"crypto_history_sharing_confirm_start_chat_dialog_content" = "You currently don’t have any chats with this person. Confirm inviting them before continuing.";
```
stefanceriu: Should these have made it to localazy already?

ElementX/Sources/Screens/InviteUsersScreen/View/ConfirmInviteUsersSheet.swift
stefanceriu: I think it's fine too keep them separate but can we maybe name it `InviteUsersConfirmationView` so it somewhat matches the other one?

ElementX/Sources/Screens/InviteUsersScreen/InviteUsersScreenModels.swift
```diff
@@ -69,12 +70,17 @@ struct InviteUsersScreenViewStateBindings {
     var searchQuery = ""
     var selectedUsersPosition: String?
     
+    /// Whether we are showing the confirmation dialog.
+    var presentConfirmationDialog = false
```
stefanceriu, line 74: Can we infer this from usersToConfirm and have it be computed instead or at least set when that changes? 🤔

ElementX/Sources/Screens/InviteUsersScreen/InviteUsersScreenViewModel.swift
stefanceriu: > these tasks should be very short-lived

I agree so why not run them directly on the current actor?

ElementX/Sources/Screens/InviteUsersScreen/InviteUsersScreenViewModel.swift
stefanceriu: Makes sense to me

ElementX/Sources/Screens/InviteUsersScreen/View/InviteUsersScreen.swift
```diff
@@ -194,6 +198,11 @@ struct InviteUsersScreen_Previews: PreviewProvider, TestablePreview {
             viewModel.state.selectedUsers = [.mockAlice]
         }
         
+        if isConfirm {
```
stefanceriu: `isConfirm` sounds might weird, can we do `shouldConfirm` instead?

ElementX/Sources/Screens/InviteUsersScreen/InviteUsersScreenViewModel.swift
```diff
@@ -67,12 +77,37 @@ class InviteUsersScreenViewModel: InviteUsersScreenViewModelType, InviteUsersScr
 
     // MARK: - Private
     
+    private func presentInviteConfirmation() {
```
stefanceriu: This implies something will be presented. You can add `ifNecessary` but I would just inline it.

ElementX/Sources/Screens/RoomMemberDetailsScreen/RoomMemberDetailsScreenViewModel.swift
```diff
@@ -197,8 +200,20 @@ class RoomMemberDetailsScreenViewModel: RoomMemberDetailsScreenViewModelType, Ro
         case .success(let roomID):
             if let roomID {
                 actionsSubject.send(.openDirectChat(roomID: roomID))
+            } else if appSettings.enableKeyShareOnInvite {
+                Task.detached {
```
stefanceriu: Why are you reaching for detached tasks in all the places?

ElementX/Sources/Screens/RoomMemberDetailsScreen/View/RoomMemberDetailsScreen.swift
```diff
@@ -30,6 +30,8 @@ struct RoomMemberDetailsScreen: View {
         .alert(item: $context.alertInfo)
         .sheet(item: $context.inviteConfirmationUser) { user in
             SendInviteConfirmationView(userToInvite: user,
+                                       userIdentityKnown: context.inviteConfirmationUserIdentityKnown,
```
stefanceriu: This one doesn't need to be a binding but I appreciate wanting to have it next to the other one. Perhaps you should tuple them together, the type doesn't really matter.

ElementX/Sources/Screens/RoomMemberDetailsScreen/RoomMemberDetailsScreenViewModel.swift
```diff
@@ -197,8 +200,20 @@ class RoomMemberDetailsScreenViewModel: RoomMemberDetailsScreenViewModelType, Ro
...
+                    let identity = await self.userSession.clientProxy.userIdentity(for: roomMemberProxy.userID, fallBackToServer: false)
+                    Task { @MainActor in
+                        self.state.bindings.inviteConfirmationUser = .init(userID: roomMemberProxy.userID, displayName: roomMemberProxy.displayName, avatarURL: roomMemberProxy.avatarURL)
+                        self.state.bindings.inviteConfirmationUserIdentityKnown = switch identity {
+                        case .success(let identity): identity != nil
+                        default: false
+                        }
```
stefanceriu: Can also be written as

```
self.state.bindings.inviteConfirmationUserIdentityKnown = if case .success(let identity) = await self.userSession.clientProxy.userIdentity(for: roomMemberProxy.userID, fallBackToServer: false) {
    identity != nil
} else {
    false
}
```

ElementX/Sources/Screens/StartChatScreen/View/SendInviteConfirmationView.swift
```diff
@@ -11,6 +11,7 @@ import SwiftUI
 
 struct SendInviteConfirmationView: View {
     let userToInvite: UserProfileProxy
+    let userIdentityKnown: Bool
```
stefanceriu: I would rename this to `isUserIdentityKnown`

ElementX/Sources/Screens/StartChatScreen/View/SendInviteConfirmationView.swift
```diff
@@ -92,6 +105,13 @@ struct SendInviteConfirmationView: View {
 struct SendInviteConfirmationView_Previews: PreviewProvider, TestablePreview {
     static var previews: some View {
         SendInviteConfirmationView(userToInvite: .mockBob,
+                                   userIdentityKnown: true,
+                                   mediaProvider: nil) { }
+            .previewDisplayName("With Identity")
```
stefanceriu: Technically `With identity known` and unknown as you're actually passing the identity in.

ElementX/Sources/Screens/StartChatScreen/StartChatScreenModels.swift
```diff
@@ -45,6 +45,7 @@ struct StartChatScreenViewStateBindings {
     var alertInfo: AlertInfo<StartChatScreenErrorType>?
     
     var selectedUserToInvite: UserProfileProxy?
+    var selectedUserIdentityKnown = false
```
stefanceriu: Same question here, why is this a binding?

ElementX/Sources/Screens/StartChatScreen/StartChatScreenViewModel.swift
```diff
@@ -67,8 +67,24 @@ class StartChatScreenViewModel: StartChatScreenViewModelType, StartChatScreenVie
                 hideLoadingIndicator()
                 actionsSubject.send(.showRoom(roomID: roomId))
             case .success:
-                hideLoadingIndicator()
-                state.bindings.selectedUserToInvite = user
+                if appSettings.enableKeyShareOnInvite {
+                    Task.detached {
+                        let identity = await self.userSession.clientProxy.userIdentity(for: user.userID, fallBackToServer: false)
+                        Task { @MainActor in
```
stefanceriu: Same

stefanceriu (commented):
Looking good, nice start 👍

ElementX/Resources/Localizations/en.lproj/Untranslated.strings
kaylendog: It's on my list of things to do!

ElementX/Sources/Screens/InviteUsersScreen/InviteUsersScreenModels.swift
kaylendog: I'm not sure this is entirely possible, since we only want to prompt the user when they press `Invite` - I could add a `didSet { if usersToConfirm.isEmpty ... }`, which would let us avoid setting it to false in the view model?

ElementX/Sources/Screens/RoomMemberDetailsScreen/RoomMemberDetailsScreenViewModel.swift
kaylendog: Was using other code in the project, I'm a bit unfamiliar with Swift so don't entirely understand its concurrency model. Addressed in https://github.com/element-hq/element-x-ios/pull/5331/commits/b29a8bbc63168a29ebfd2ded6ee8da851a77d299

ElementX/Sources/Screens/StartChatScreen/StartChatScreenViewModel.swift
kaylendog: Addressed in https://github.com/element-hq/element-x-ios/pull/5331/commits/b29a8bbc63168a29ebfd2ded6ee8da851a77d299

ElementX/Sources/Screens/InviteUsersScreen/InviteUsersScreenViewModel.swift
kaylendog: Addressed in https://github.com/element-hq/element-x-ios/pull/5331/commits/b29a8bbc63168a29ebfd2ded6ee8da851a77d299

ElementX/Sources/Screens/RoomMemberDetailsScreen/View/RoomMemberDetailsScreen.swift
kaylendog: It looks like tuples aren't `Identifiable`, which prevents me from using them as arguments to `item` - I'll move the various `identityKnown` (now `identityUnknown` to avoid a double negative) out of bindings.

ElementX/Sources/Screens/InviteUsersScreen/View/InviteUsersConfirmationSheetView.swift
```diff
@@ -0,0 +1,71 @@
...
+                    ForEach(context.viewState.usersToConfirm, id: \.userID) { user in
+                        Divider()
+                        UserProfileListRow(user: user,
+                                           membership: nil,
+                                           mediaProvider: context.mediaProvider,
+                                           kind: .label)
+                            .accessibilityIdentifier(A11yIdentifiers.inviteUsersScreen.userProfile)
+                    }
+                }
+            }
```
pixlwave: 3 comments on this:
- The ScrollView isn't really needed here, `FullscreenDialog` automatically becomes scrollable when its content doesn't fit within the available space.
- `Divider` uses a system colour, whereas I assume that the Figma uses a colour from Compound so you probably want to either use `RowDivider` here (or build one yourself using a similar pattern).
- We should probably have a `VStack` to wrap the whole thing so that we can set a `spacing` that matches the Figma.

ElementX/Sources/Screens/InviteUsersScreen/View/InviteUsersScreen.swift
```diff
@@ -149,6 +152,7 @@ struct InviteUsersScreen_Previews: PreviewProvider, TestablePreview {
     static let viewModel = makeViewModel()
     static let searchingViewModel = makeViewModel(searchQuery: "Alice")
     static let selectedViewModel = makeViewModel(hasSelection: true)
+    static let confirmSelectedViewModel = makeViewModel(shouldConfirm: true)
```
pixlwave, line 155: From what I can tell, this is unused?

ElementX/Sources/Screens/InviteUsersScreen/View/InviteUsersScreen.swift
```diff
@@ -194,6 +198,11 @@ struct InviteUsersScreen_Previews: PreviewProvider, TestablePreview {
             viewModel.state.selectedUsers = [.mockAlice]
         }
         
+        if shouldConfirm {
+            viewModel.state.usersToConfirm = [.mockAlice, .mockAlice, .mockAlice, .mockAlice, .mockAlice, .mockAlice, .mockAlice, .mockAlice, .mockAlice]
```
pixlwave, line 207: Also unsure what we have so many Alice's here 😄

ElementX/Sources/Screens/InviteUsersScreen/InviteUsersScreenViewModel.swift
```diff
@@ -59,6 +61,20 @@ class InviteUsersScreenViewModel: InviteUsersScreenViewModelType, InviteUsersScr
         case .cancel:
             actionsSubject.send(.dismiss)
         case .proceed:
+            guard appSettings.enableKeyShareOnInvite,
+                  roomProxy.details.historySharingState != RoomHistorySharingState.hidden,
+                  !state.usersToConfirm.isEmpty,
+                  !state.isSkippable else {
+                return inviteUsers(state.selectedUsers.map(\.userID), roomProxy: roomProxy)
```
pixlwave: I get that the above works but it's also harder to read in a diff, I would suggest keeping it simple with

```suggestion
                inviteUsers(state.selectedUsers.map(\.userID), roomProxy: roomProxy)
                return
```

ElementX/Sources/Screens/InviteUsersScreen/InviteUsersScreenModels.swift
```diff
@@ -69,12 +70,17 @@ struct InviteUsersScreenViewStateBindings {
...
+    
     /// Information describing the currently displayed alert.
     var alertInfo: AlertInfo<InviteUsersScreenErrorType>?
 }
 
 enum InviteUsersScreenViewAction {
     case cancel
     case proceed
+    case recheck
+    case confirm
```
pixlwave: I would suggest that these are more descriptive names as to what the actions actually do.

```suggestion
    case removeUnknownUsers
    case confirmUnknownUsers
```

ElementX/Sources/Screens/InviteUsersScreen/InviteUsersScreenViewModel.swift
```diff
@@ -59,6 +61,20 @@ class InviteUsersScreenViewModel: InviteUsersScreenViewModelType, InviteUsersScr
...
+            }
+            state.bindings.presentConfirmationDialog = true
+        case .recheck:
+            state.bindings.presentConfirmationDialog = false
+            state.selectedUsers.removeAll { lhs in state.usersToConfirm.contains { rhs in lhs.userID == rhs.userID } }
```
pixlwave: Just a small nit for readability:

```suggestion
            state.selectedUsers.removeAll { user in
                state.usersToConfirm.contains { $0.userID == user.userID }
            }
```

ElementX/Sources/Screens/InviteUsersScreen/InviteUsersScreenViewModel.swift
```diff
@@ -73,6 +89,17 @@ class InviteUsersScreenViewModel: InviteUsersScreenViewModelType, InviteUsersScr
         } else {
             state.selectedUsers.append(user)
             withElementAnimation(.easeInOut) { state.bindings.selectedUsersPosition = user.userID }
+            Task {
+                let identityUnknown = if case .success(let identity) = await self.clientProxy.userIdentity(for: user.userID, fallBackToServer: true) {
```
pixlwave: Is `fallBackToServer: true` correct here for fetching the cached identity?

ElementX/Sources/Screens/RoomMemberDetailsScreen/RoomMemberDetailsScreenViewModel.swift
```diff
@@ -197,8 +200,19 @@ class RoomMemberDetailsScreenViewModel: RoomMemberDetailsScreenViewModelType, Ro
         case .success(let roomID):
             if let roomID {
                 actionsSubject.send(.openDirectChat(roomID: roomID))
+            } else if appSettings.enableKeyShareOnInvite {
```
pixlwave: How come this one doesn't also check the room's `RoomHistorySharingState`?

ElementX/Sources/Screens/StartChatScreen/View/SendInviteConfirmationView.swift
```diff
@@ -19,14 +20,26 @@ struct SendInviteConfirmationView: View {
     @State private var sheetHeight: CGFloat = .zero
     private let topPadding: CGFloat = 24
     
+    private var title: String {
+        return if isUserIdentityUnknown {
```
pixlwave: We can use an implicit return here.
```suggestion
       if isUserIdentityUnknown {
```

ElementX/Sources/Screens/StartChatScreen/StartChatScreenViewModel.swift
```diff
@@ -67,8 +67,22 @@ class StartChatScreenViewModel: StartChatScreenViewModelType, StartChatScreenVie
                 hideLoadingIndicator()
                 actionsSubject.send(.showRoom(roomID: roomId))
             case .success:
-                hideLoadingIndicator()
-                state.bindings.selectedUserToInvite = user
+                if appSettings.enableKeyShareOnInvite {
```
pixlwave, line 70: Same question about the `RoomHistorySharingState` (although this time presumably it would need to be inferred from the type of room being created).

ElementX/Sources/Screens/InviteUsersScreen/View/InviteUsersConfirmationSheetView.swift
```diff
@@ -0,0 +1,71 @@
...
+
+struct InviteUsersConfirmationSheetView: View {
+    @ObservedObject var context: InviteUsersScreenViewModel.Context
+    
+    var body: some View {
+        FullscreenDialog(topPadding: 24, horizontalPadding: 24) {
+            TitleAndIcon(title: UntranslatedL10n.cryptoHistorySharingConfirmInviteDialogTitle,
+                         subtitle: UntranslatedL10n.cryptoHistorySharingConfirmInviteDialogContent,
+                         icon: \.userAddSolid,
+                         iconStyle: .default)
```
pixlwave: I think you might want `.defaultSolid` here to get the right background colour.

ElementX/Sources/Screens/RoomMemberDetailsScreen/View/RoomMemberDetailsScreen.swift
pixlwave: Honestly I think the nicest solution here, for both the existing screen and the new one would be to make a `UserToInvite` or `InviteUser` struct/class that contains both the profile and an `isKnown` flag and use that on both of these screens. It would be cleaner that way from my perspective.

pixlwave (commented):
I think I've caught the full context from Stefan's review, here's some thoughts from me, but please shout if I put anything that contradicts what Stefan said.

ElementX/Sources/Screens/RoomMemberDetailsScreen/RoomMemberDetailsScreenViewModel.swift
kaylendog: Good catch!

ElementX/Sources/Screens/InviteUsersScreen/View/InviteUsersScreen.swift
kaylendog: I wanted to make sure scrolling would behave as expected, so she is throwing a party with her clones! I could use `.mockBob` and co.?

pixlwave: haha, nah it's fine for her to party away. Was only double-checking it wasn't a mistake 😄

ElementX/Sources/Screens/InviteUsersScreen/InviteUsersScreenViewModel.swift
kaylendog: ack, another good catch, will fix.

ElementX/Sources/Screens/StartChatScreen/StartChatScreenViewModel.swift
kaylendog: I think we want to prompt users regardless of the room history in the case of DM creation, which I believe is exclusively handled by this code path?

ElementX/Sources/Screens/InviteUsersScreen/View/InviteUsersConfirmationSheetView.swift
kaylendog: Attempted to apply much of this in https://github.com/element-hq/element-x-ios/pull/5331/commits/5a38fcbfc43d90314ef1c62cbc0c2a7f1625e641, but probably warrants another pass. I *think* I'm matching the Figma? It has some extra padding components which make it a bit tricky.

ElementX/Sources/Screens/InviteUsersScreen/View/InviteUsersScreen.swift
kaylendog: Yep, missing preview test added in https://github.com/element-hq/element-x-ios/pull/5331/commits/f88f17101a893aaef363d21715cbec49dafc84e9

UnitTests/Sources/RoomMemberDetailsViewModelTests.swift
```diff
@@ -151,6 +152,58 @@ struct RoomMemberDetailsViewModelTests {
...
+        ServiceLocator.shared.settings.enableKeyShareOnInvite = true
+        clientProxy.directRoomForUserIDReturnValue = .success(nil)
+        clientProxy.userIdentityForFallBackToServerReturnValue = .success(UserIdentityProxyMock(configuration: .init(verificationState: .notVerified)))
+        
+        // The user identity becomes known, i.e. not unknown.
+        let deferred = deferFulfillment(viewModel.context.$viewState) { viewState in
+            guard case let .some(isUnknown) = viewState.bindings.inviteConfirmationUser.map(\.isUnknown) else {
+                return false
+            }
+            return !isUnknown
```
kaylendog: I'm not a huge fan of this pattern since it makes the intent behind the assertion less clear. Perhaps there's something more conventional?

UnitTests/Sources/StartChatViewModelTests.swift
```diff
@@ -85,6 +85,46 @@ struct StartChatScreenViewModelTests {
...
+        ServiceLocator.shared.settings.enableKeyShareOnInvite = true
+        clientProxy.directRoomForUserIDReturnValue = .success(nil)
+        clientProxy.userIdentityForFallBackToServerReturnValue = .success(UserIdentityProxyMock(configuration: .init(verificationState: .notVerified)))
+        
+        // User identity becomes known, i.e. not unknown
+        let deferred = deferFulfillment(viewModel.context.$viewState) { viewState in
+            guard case let .some(isUnknown) = viewState.bindings.selectedUserToInvite.map(\.isUnknown) else {
+                return false
+            }
+            return !isUnknown
```
kaylendog: Same as above.

ElementX/Sources/Screens/StartChatScreen/StartChatScreenViewModel.swift
pixlwave: Ok got it, sgtm then 👍

ElementX/Sources/Screens/StartChatScreen/StartChatScreenViewModel.swift
```diff
@@ -69,19 +69,18 @@ class StartChatScreenViewModel: StartChatScreenViewModelType, StartChatScreenVie
             case .success:
                 if appSettings.enableKeyShareOnInvite {
                     Task {
-                        hideLoadingIndicator()
-                        self.state.bindings.selectedUserToInvite = user
                         // If an error occured while fetching the identity, assume they are unknown.
-                        self.state.selectedUserIdentityUnknown = if case .success(let identity) = await self.userSession.clientProxy.userIdentity(for: user.userID, fallBackToServer: false) {
+                        var isUnknown = if case .success(let identity) = await self.userSession.clientProxy.userIdentity(for: user.userID, fallBackToServer: false) {
```
pixlwave: This can be a `let` (and is presumably raising a warning about not being mutated).

ElementX/Sources/Screens/UserProfileScreen/UserProfileScreenViewModel.swift
```diff
@@ -124,7 +124,7 @@ class UserProfileScreenViewModel: UserProfileScreenViewModelType, UserProfileScr
             if let roomID {
                 actionsSubject.send(.openDirectChat(roomID: roomID))
             } else {
-                state.bindings.inviteConfirmationUser = userProfile
+                state.bindings.inviteConfirmationUser = .init(user: userProfile, isUnknown: false)
```
pixlwave: Do we want to fetch the profile for this too? Amongst other things, this screen is shown when opening permalinks from outside of the app, so there's probably a higher chance of unknown users being encountered here generally.

ElementX/Sources/Services/Room/UserToInvite.swift
```diff
@@ -0,0 +1,41 @@
...
+    /// The profile of the user being invited.
+    var user: UserProfileProxy
+    
+    /// Whether we have the cryptographic identity of this user cached locally.
+    var isUnknown: Bool
+    
+    /// The ID of the user being invited.
+    var userID: String {
+        user.userID
+    }
```
pixlwave, line 17: Let's remove this one, the `id` from the `Identifiable` conformance is perfectly clear enough (but you can move that here rather than being in an extension if you want).

ElementX/Sources/Services/Room/UserToInvite.swift
```diff
@@ -0,0 +1,41 @@
...
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Foundation
+
+struct UserToInvite {
+    static func == (lhs: UserToInvite, rhs: UserToInvite) -> Bool {
+        lhs.user == rhs.user
+    }
```
pixlwave: Does this do anything useful without an `Equatable` conformance?

Also if you need it to be `Equatable` then `UserProfileProxy` is already `Equatable` so if you add the conformance here and remove this function you should have everything you need for it to be automatically synthesised.

UnitTests/Sources/RoomMemberDetailsViewModelTests.swift
pixlwave: Ah yeah you can write this whole thing as

```swift
guard let user = viewState.bindings.inviteConfirmationUser else { return false }
return !user.isUnknown
```

or if you add a `var isKnown: Bool { !isUnknown }` to the type then you could use this 1-liner here:

```swift
viewState.bindings.inviteConfirmationUser?.isKnown ?? false
```

orrrrr

```swift
let deferred = deferFulfillment(viewModel.context.$viewState.compactMap(\.bindings.inviteConfirmationUser)) { !$0.isUnknown }
```

Any of those work from my perspective, feel free to pick your favourite :)

pixlwave (commented):
Looking good 👏

ElementX/Sources/Screens/RoomMemberDetailsScreen/View/RoomMemberDetailsScreen.swift
pixlwave: Thank you, that looks much neater now 👌

ElementX/Sources/Screens/UserProfileScreen/UserProfileScreenViewModel.swift
kaylendog: Good catch! Fixed in https://github.com/element-hq/element-x-ios/pull/5331/commits/7ed8ef6bd811294f8715f098ab1dc9a3a5671e3b

kaylendog:
@bmarty raised a valid concern in the mirror Android PR about plural translations that I think should be done in this PR too, will do this tomorrow/recording it now in case I don't finish before the weekend.

kaylendog:
I've added the plural translations to Localazy, but haven't figured out how to sync them into the project. Might be best to wait for the next translation update PR.

pixlwave:
> I've added the plural translations to Localazy, but haven't figured out how to sync them into the project. Might be best to wait for the next translation update PR.

Instructions added [here](https://github.com/element-hq/element-x-ios/commit/001a389f5132effde1aec3b3a3093a0e8f08641e), but #5388 was merged this morning anyway.

andybalaam:
OK @stefanceriu @pixlwave I have rebased this on top of the localazy update (I hope). I think Skye has addressed all your comments but let me know if we missed anything.

stefanceriu:
> let me know if we missed anything

Just the translations I guess, what's the plan there?

stefanceriu:
Great stuff, thank you for taking care of it! 👏

andybalaam:
@stefanceriu thank you for getting this over the line. You are a star.

PR #5334, opened 2026-03-31
Replace KeychainAccess with a custom keychain wrapper

mgcm:
[KeychainAccess](https://github.com/kishikawakatsumi/keychainaccess) hasn't seen an update for 5 years now and seems largely unmaintained for over 3 years.

This PR removes the dependency and introduces a custom Keychain wrapper that maintains mostly the same API.


pixlwave:
Hey @mgcm, thanks for the PR. We're a bit mixed within the team about whether it's worth adding the maintenance burden of our own component here. Are you able to share an overview of what this custom wrapper brings us over the existing KeychainAccess implementation?

On the one hand we can see that there haven't been any updates to KeychainAccess like you say, but then we're not aware of any changes to the underlying keychain API that may or may not have happened over the past 5 years. The package is [well tested](https://github.com/kishikawakatsumi/KeychainAccess/tree/master/Lib/KeychainAccessTests), Kishikawa Katsumi is still very much active within the Swift community and it still shows as the most popular package on https://swiftpackageindex.com/search?query=keychain so seems likely that any critical fixes would be applied if they were needed.

Again, I'm not saying we're not interested, but it would be helpful to know what technical/implementation benefits this adds over the third-party dependency.

pixlwave:
I'm going to close this as stale for now. Feel free to re-open @mgcm if you think there is good evidence for a reason for us to switch to managing the Keychain ourselves 🙂

PR #5336, opened 2026-04-01
Allow presenting threads from the new room thread list

stefanceriu:
.. and hook everything up to the RoomFlowCoordinator state machine.

### DIFF
--- a/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
@@ -394,6 +394,11 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
             case (.room, .presentPinnedEventsTimeline, .pinnedEventsTimeline):
                 startPinnedEventsTimelineFlow()
                 
+            // Thread List
+
+            case (.room, .presentThreadList, .threadList):
+                Task { await self.presentThreadList(animated: animated) }
+
             // Thread
                 
             case (_, .presentThread(let threadRootEventID, let focusEventID), .thread):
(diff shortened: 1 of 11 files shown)


ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -394,6 +394,11 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
             case (.room, .presentPinnedEventsTimeline, .pinnedEventsTimeline):
                 startPinnedEventsTimelineFlow()
                 
+            // Thread List
+
+            case (_, .presentThreadList, .threadList):
```
pixlwave: Considering the state machine only allows `.room` → `.threadList` should this be like so?

```suggestion
            case (.room, .presentThreadList, .threadList):
```

ElementX/Sources/Screens/RoomThreadListScreen/View/RoomThreadListScreen.swift
pixlwave, line 1: I feel like this should have a button style to set a background colour when pressed (like the home screen does) rather than the default style that reduces the opacity.

ElementX/Sources/Services/Room/JoinedRoomProxy.swift
```diff
@@ -204,7 +204,7 @@ class JoinedRoomProxy: JoinedRoomProxyProtocol {
     }
     
     func threadListService() async -> RoomThreadListServiceProxyProtocol {
-        await RoomThreadListProxy(threadListService: room.threadListService(), eventStringBuilder: eventStringBuilder)
+        RoomThreadListProxy(threadListService: room.threadListService(), eventStringBuilder: eventStringBuilder)
```
pixlwave: That `async` can go too then :)

pixlwave (approved):
Wooo, we're getting there! 🧵

PR #5341, opened 2026-04-01
Update reply view border corner radii and color

stefanceriu:
And also update relating preview test snapshots


americanrefugee (commented):
Why are there double sentences for sharing a location?

stefanceriu:
> Why are there double sentences for sharing a location?

That's expected, the tests also run in Double-Length Pseudolanguage to check how layouts work with longer words.

PreviewTests/Sources/__Snapshots__/PreviewTests/timelineItemBubbledStylerView.Thread-decorator-iPhone-en-GB.png
pixlwave, line 1: Is it intentional to use the same border colour as the outgoing message bubble background? This basically means that incoming messages have a border around the reply view but outgoing messages don't?

americanrefugee:
> Is it intentional to use the same border colour as the outgoing message bubble background? This basically means that messages have a border around the reply view but outgoing messages don't?

@pixlwave Yes, that's fine to me. There wasn't a lot of contrast between the reply contents and a received message, but there's certainly enough contrast between the reply contents and a sent message.

This way the border overall isn't too stark. I personally like this update. Do you agree?

pixlwave (approved):
> Yes, that's fine to me.

In which case, LGTM 👍

PR #5344, opened 2026-04-02
Remove the `shouldDisambiguateDisplayNames` parameter from the `RoomEvent` and `RoomStateEvent` string builders

stefanceriu:
Doug and I had a think about it and we can't come up with a reason why names should ever not be disambiguated


pixlwave (approved):

pixlwave:
Let's see 🤞

PR #5348, opened 2026-04-02
Start send and stop functions of the LiveLocationManager

### LINKED ISSUE #5197: [Task] EX LLS: Resume LLS

Velin92:
This PR implements inside the LiveLocationManager the following functions
- Start live location sharing
- Send LLS updates (all done in background the user doesn't need to worry about anything)
- Stop LLS

It also implements a store that keeps track of what are the rooms this device is sending LLS to, in this way in case of a crash, kill or reopen we can restore the LLS updates.
Also when there are no more rooms to send LLS events to, we just stop the CLLocationUpdates entirely.

We also try to stop events once they have expired in conformance with the MSC.
It's important to note that this is a best effort thing, if it's not possible we don't try again (usually clients implement an internal check to consider an LLS expired, so that's generally fine).

To review commit by commit.

Also the send functions will not work properly on the current SDK release, since we need a fix for the sync over the `beaconInfo` event, which is being worked on here: https://github.com/matrix-org/matrix-rust-sdk/pull/6385

The following PR is the inclusion of these functions inside the UI + tests: https://github.com/element-hq/element-x-ios/pull/5349

fixes #5197

### DIFF
--- a/ElementX/Sources/Application/Settings/AppSettings.swift
+++ b/ElementX/Sources/Application/Settings/AppSettings.swift
@@ -80,6 +80,7 @@ final class AppSettings {
         case focusEventOnNotificationTap
         case linkNewDeviceEnabled
         case liveLocationSharingEnabled
+        case liveLocationSharingTimeoutDatesByRoomID
         case floatingTimelineDateEnabled
         
         // Doug's tweaks 🔧
@@ -348,6 +349,9 @@ final class AppSettings {
     
     @UserPreference(key: UserDefaultsKeys.hasRequestedLocationAlwaysLocationAuthorization, defaultValue: false, storageType: .userDefaults(store))
     var hasRequestedLocationAlwaysLocationAuthorization
+    
+    @UserPreference(key: UserDefaultsKeys.liveLocationSharingTimeoutDatesByRoomID, defaultValue: [String: Date](), storageType: .userDefaults(store))
+    var liveLocationSharingTimeoutDatesByRoomID
     
     @UserPreference(key: UserDefaultsKeys.frequentlyUsedSystemEmojis, defaultValue: [FrequentlyUsedEmoji](), storageType: .userDefaults(store))
     var frequentlyUsedSystemEmojis
--- a/ElementX/Sources/Application/Settings/AppSettings.swift
+++ b/ElementX/Sources/Application/Settings/AppSettings.swift
@@ -66,6 +66,7 @@ final class AppSettings {
         case elementCallBaseURLOverride
         
         case voiceMessagePlaybackSpeed
+        case liveLocationSharingTimeoutDatesByRoomID
         
         // Feature flags
         case publicSearchEnabled
@@ -80,7 +81,6 @@ final class AppSettings {
         case focusEventOnNotificationTap
         case linkNewDeviceEnabled
         case liveLocationSharingEnabled
-        case liveLocationSharingTimeoutDatesByRoomID
         case floatingTimelineDateEnabled
         
         // Doug's tweaks 🔧
--- a/ElementX/Sources/Services/Location/LiveLocationManager.swift
+++ b/ElementX/Sources/Services/Location/LiveLocationManager.swift
@@ -59,20 +59,20 @@ class LiveLocationManager: NSObject, LiveLocationManagerProtocol, CLLocationMana
         return true
     }
     
-    func startLiveLocation(roomID: String, durationMillis: UInt64) async -> Result<Void, LiveLocationManagerError> {
+    func startLiveLocation(roomID: String, duration: Duration) async -> Result<Void, LiveLocationManagerError> {
         guard case .joined(let roomProxy) = await clientProxy.roomForIdentifier(roomID) else {
             MXLog.error("Failed to resolve joined room for identifier: \(roomID)")
             return .failure(.roomNotJoined)
         }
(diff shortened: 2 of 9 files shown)


ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -80,6 +80,7 @@ final class AppSettings {
         case focusEventOnNotificationTap
         case linkNewDeviceEnabled
         case liveLocationSharingEnabled
+        case liveLocationSharingTimeoutDatesByRoomID
```
stefanceriu: Surely this shouldn't go in the feature flags group

ElementX/Sources/Services/Location/LiveLocationManagerProtocol.swift
```diff
@@ -20,4 +25,18 @@ protocol LiveLocationManagerProtocol: AnyObject {
     ///            `false` if the request was already made before and iOS would silently ignore it.
     @discardableResult
     func requestAlwaysAuthorizationIfPossible() -> Bool
+    
+    /// Starts sharing live location in a room.
+    ///
+    /// - Parameters:
+    ///   - roomID: The identifier of the room to share live location in.
+    ///   - durationMillis: The duration in milliseconds for how long the live location should be shared.
+    func startLiveLocation(roomID: String, durationMillis: UInt64) async -> Result<Void, LiveLocationManagerError>
```
stefanceriu: Why not an actual `Duration`?

ElementX/Sources/Services/Location/LiveLocationManager.swift
```diff
@@ -33,6 +42,11 @@ class LiveLocationManager: NSObject, LiveLocationManagerProtocol, CLLocationMana
         super.init()
```
stefanceriu, line 42: Why do you need this constructor to be a main actor or `await MainActor.run` when you use it? 🤔

ElementX/Sources/Services/Room/RoomProxyProtocol.swift
```diff
@@ -194,6 +194,12 @@ protocol JoinedRoomProxyProtocol: RoomProxyProtocol {
     func saveDraft(_ draft: ComposerDraft, threadRootEventID: String?) async -> Result<Void, RoomProxyError>
     func loadDraft(threadRootEventID: String?) async -> Result<ComposerDraft?, RoomProxyError>
     func clearDraft(threadRootEventID: String?) async -> Result<Void, RoomProxyError>
+    
+    // MARK: - Live Location
+    
+    func startLiveLocationShare(durationMillis: UInt64) async -> Result<Void, RoomProxyError>
```
stefanceriu: Same question here I guess, we should use Duration when we can, the semantics are way clearer.

ElementX/Sources/Services/Location/LiveLocationManager.swift
```diff
@@ -53,6 +100,114 @@ class LiveLocationManager: NSObject, LiveLocationManagerProtocol, CLLocationMana
...
+    private func syncActiveRoomProxies(with sessions: [String: Date]) {
+        // Remove proxies for rooms no longer in the dictionary.
+        let activeRoomIDs = Set(sessions.keys)
+        for roomID in activeRoomProxies.keys where !activeRoomIDs.contains(roomID) {
+            activeRoomProxies.removeValue(forKey: roomID)
+        }
+    }
+    
+    private func startLocationUpdatesIfNeeded() {
+        guard locationUpdatesTask == nil else { return }
```
stefanceriu, line 149: Does this check mean that `locationUpdatesTask` doesn't need to be a `@CancellableTask`?

stefanceriu (commented):
Very nice, not long now! 🚀

ElementX/Sources/Services/Location/LiveLocationManagerProtocol.swift
Velin92: and then we convert it in milliseconds internally for the SDK? yeah sounds good why not.

ElementX/Sources/Services/Location/LiveLocationManager.swift
Velin92: https://github.com/element-hq/element-x-ios/pull/5348/changes#diff-70748cb900c5961ae09993ed60223c23d4f7b2adf2bd732c0a479d4dba4d9241R36-R38

> Core Location calls the methods of your delegate object on the runloop from the thread on which you initialized [CLLocationManager](https://developer.apple.com/documentation/corelocation/cllocationmanager). That thread must itself have an active run loop, like the one found in your app’s main thread.

ElementX/Sources/Services/Location/LiveLocationManager.swift
Velin92: `@CancellableTask` is just nice because when you reassign or set it to nil it automatically calls .cancel() over  it. Nothing big, but you save one line of code every time there is a reassignment (since a reassign task doesn't necessarily get cancelled)

ElementX/Sources/Services/Location/LiveLocationManager.swift
stefanceriu: Huh, that's such a weird API

stefanceriu (approved):
Noice 🚀

PR #5349, opened 2026-04-02
Implement live location sharing start and stop actions in the UI

Velin92:
Okay so this PR finally connects the CTAs that were already in the app about LLS with its manager.

- It adds a bunch of alerts after the permissions are determined when selecting the live location sharing option from the location screen (a disclaimer about your privacy + a duration selector) and then calls the start function for the room
- it connects the stop function to the live location timeline item view stop button
- Adds a bunch of tests to the `LocationSharingScreenViewModel`

### DIFF
--- a/ElementX/Sources/Screens/LocationSharing/LocationSharingScreenViewModel.swift
+++ b/ElementX/Sources/Screens/LocationSharing/LocationSharingScreenViewModel.swift
@@ -72,7 +72,7 @@ class LocationSharingScreenViewModel: LocationSharingScreenViewModelType, Locati
                 state.bindings.showsUserLocationMode = .showAndFollow
             case .some(false):
                 let action: () -> Void = { [weak self] in self?.actionsSubject.send(.openSystemSettings) }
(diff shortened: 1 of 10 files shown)


ElementX/Sources/Screens/LocationSharing/LocationSharingScreenViewModel.swift
```diff
@@ -139,17 +145,59 @@ class LocationSharingScreenViewModel: LocationSharingScreenViewModelType, Locati
...
+                                             // Delay so SwiftUI finishes dismissing the current alert
+                                             // before presenting the next one.
+                                             DispatchQueue.main.async {
+                                                 self?.showLiveLocationDurationPicker()
+                                             }
+                                         })
+    }
+    
+    private func showLiveLocationDurationPicker() {
+        let durations: [Duration] = [.seconds(15 * 60), // 15 minutes
```
stefanceriu: Use Duration? 😅

stefanceriu (approved):
lgtm, nice 👍

PR #5350, opened 2026-04-03
Add an action to the newly introduced FloatingDateBadge that scrolls the timeline to that day.

stefanceriu:


ElementX/Sources/Screens/Timeline/View/FloatingDateBadge.swift
```diff
@@ -13,6 +13,7 @@ import SwiftUI
 /// fades out shortly after scrolling stops.
 struct FloatingDateBadge: View {
     let dateText: String?
+    var onTap: (() -> Void)?
```
Velin92, line 16: can also be non optional

ElementX/Sources/Screens/Timeline/TimelineTableViewController.swift
```diff
@@ -546,6 +553,21 @@ extension TimelineTableViewController {
...
+        for uniqueID in timelineItemsDictionary.keys {
+            if let timestamp = timelineItemsDictionary[uniqueID]?.timestamp,
+               timestamp.formattedDateSeparator() == dateText,
+               let indexPath = dataSource?.indexPath(for: uniqueID) {
+                // The table view is flipped, so .bottom aligns the cell to the visual top.
+                tableView.scrollToRow(at: indexPath, at: .bottom, animated: true)
+                return
+            }
+        }
+    }
```
Velin92, line 569: this seems overly complicated, from what I understand we are scrolling with the earliest entry, which means the topmost/oldest that we can render right? This would match the current entry the floating date is getting the date from.
Then instead of determining the item again with this reverse look function, when storing the floatingDateText let's also store the indexPath of the corresponding item, so that when tapping we can just get there. It doesn't even need to be a binding or anything can just be another private variable inside the table view that is updated as soon as floatingDateText is updated.

In this way this function is just as simple as calling `tableView.scrollToRow(at: floatingDateIndexPath, at: .bottom, animated: true)`

Velin92 (changes requested):

ElementX/Sources/Screens/Timeline/TimelineTableViewController.swift
stefanceriu: Well not quite, we need to scroll to the earliest entry *for the day that the pill is showing when you tap on it* which does not necessarily match everything you have loaded up in the timeline. 
Perhaps I don't understand the suggestion, feel free to just add to this PR

ElementX/Sources/Screens/Timeline/View/FloatingDateBadge.swift
stefanceriu: Just so I didn't have to specify an empty block for the previews ¯\_(ツ)_/¯

ElementX/Sources/Screens/Timeline/View/FloatingDateBadge.swift
```diff
@@ -28,14 +29,18 @@ struct FloatingDateBadge: View {
...
-                    .shadow(color: Color(red: 0.11, green: 0.11, blue: 0.13).opacity(0.1), radius: 12, x: 0, y: 4)
-                    .transition(.opacity)
+                Button { onTap?() } label: {
+                    Text(dateText)
+                        .font(.compound.bodySMSemibold)
+                        .foregroundColor(.compound.textPrimary)
+                        .padding(.horizontal, 16)
+                        .padding(.vertical, 8)
+                        .background(backgroundColor, in: .capsule)
+                        .shadow(color: Color(red: 0.11, green: 0.11, blue: 0.13).opacity(0.1), radius: 12, x: 0, y: 4)
```
Velin92, line 39: this needs to be moved outside the button

ElementX/Sources/Screens/Timeline/TimelineTableViewController.swift
Velin92: you mean the oldest visible right? The it's exactly the value that the floating date is getting the date from.

Velin92: Oh okay tested it and now I see, we want the oldest item that is loaded (but not necessarily visible) with that timestamp got it. Can we then change the docs for it? something like
    `/// Scrolls to the oldest loaded item whose formatted date matches the current floating date.`

Velin92 (approved):
LGTM, just address my other small nit comments please 🙏

ganfra:
I think this should/could automatically back paginate, so you really are at the beginning of the day, wdyt?

ElementX/Sources/Screens/Timeline/TimelineTableViewController.swift
```diff
@@ -546,6 +553,21 @@ extension TimelineTableViewController {
         DispatchQueue.main.asyncAfter(deadline: .now() + 2.0, execute: workItem)
     }
     
+    /// Scrolls to the first (oldest) item whose formatted date matches the current floating date.
+    private func scrollToFirstItemForCurrentDate() {
+        guard let dateText = floatingDateText else { return }
+        // timelineItemsDictionary is ordered oldest-first; the first match is the earliest item for that day.
+        for uniqueID in timelineItemsDictionary.keys {
+            if let timestamp = timelineItemsDictionary[uniqueID]?.timestamp,
+               timestamp.formattedDateSeparator() == dateText,
```
Velin92: Another small thing, but since we don't need this timestamp to get formatted to be actually shown on the UI, maybe we can actually have the floatingDateText, not be a floatingDateText at all, but just a date.
And then the FloatingDateBadge itself can do the formatting. Would just make the comparison faster + avoid an additional formatting that is not really going to be shown here.

stefanceriu: Yeah, I think you're right, let's do that!

stefanceriu:
> I think this should/could automatically back paginate, so you really are at the beginning of the day, wdyt?

I thought about it but it would've complicated this PR significantly and I also don't know whether it would be good UX or not to have this thing randomly scroll in steps.

Velin92:
> I think this should/could automatically back paginate, so you really are at the beginning of the day, wdyt?

This could mean a lot of backpagination however... (in the context of big and spammy rooms) I like the idea a lot but maybe let's start small.

I imagine that if we had an API from the SDK that gave at least from the store, the timeline identifier of the matching date separator, we could instead use the focus event API to get there faster, but yeah just an idea.

ElementX/Sources/Screens/Timeline/TimelineTableViewController.swift
stefanceriu: Bah no, that doesn't work as the `floatingDateText` holds the newest and we need the oldest 🤔

stefanceriu: .. but what we can do is still iterate but check if they're `inSameDayAs`

bmarty:
Element Web has the `/jumptodate` slash command and each date separator has a popup like this:

image

So it may be possible to have the same thing?

Note: this is a FF on EleWeb:

image

ElementX/Sources/Screens/Timeline/View/FloatingDateBadge.swift
stefanceriu: Does it make a difference? 🤔

ganfra:
> Element Web has the `/jumptodate` slash command and each date separator has a popup like this:
> 
> image
> So it may be possible to have the same thing?
> 
> Note: this is a FF on EleWeb:
> 
> image

This is not exactly the same thing, tapping the FloatingDate and the DateDivider.

Telegram is handling both scenarios : 
- Clicking FloatingDate do the same than what you did in this PR
- Clicking the date divider opens a date picker where you can jump to any date

bmarty:
I know that this is not the same thing, but clicking on the floating date is just a particular case of focusing the timeline at a specific date.

stefanceriu:
> I know that this is not the same thing, but clicking on the floating date is just a particular case of focusing the timeline at a specific date.

Yup, I agree. The problem is that focused timelines are not currently going through the event cache and changing the focus ends up being pretty slow and backend bound. Let's see how this goes in practice and if need be we can improve it in the future.

PR #5352, opened 2026-04-03
Fix testUserFlow integration tests, at least locally

manuroe:
Fixes three issues causing `testUserFlow` integration tests to fail:

- Use `Close` instead of `Cancel` button label when dismissing the location picker (button label changed)
- Scroll up before tapping the People button in room details to ensure it's visible on screen
- Use `Remove this device` instead of log out

I splitted the tests into 4 with so that they are better clustered. The login test is mandatory for others.
I also made a change to log out any existing previous MAS session to avoid tests to fail if you relaunch them.

Tests locally pass:
image

Then, I tried to make them less flakey on the CI by:
- Removing the `sleep()` calls and use `waitForExistence()` instead 
- Adding retry mechanism

I also tried to reduce the flakiness on the CI but with no success. The most important here is that they pass locally.

You can check review commits one by one but be aware I made a back and forth on `tap(.center)`. I expect a squash and merge at the end.

### DIFF
--- a/IntegrationTests/Sources/Common.swift
+++ b/IntegrationTests/Sources/Common.swift
@@ -125,13 +125,13 @@ extension XCUIApplication {
         // Make the logout button visible
         swipeUp()
         
-        // Logout
+        // Remove this device (was "Sign out" before)
         let logoutButton = buttons[A11yIdentifiers.settingsScreen.logout]
         XCTAssertTrue(logoutButton.waitForExistence(timeout: 10.0))
         logoutButton.tap(.center)
         
-        // Confirm logout
-        let alertLogoutButton = alerts.firstMatch.buttons["Sign out"].firstMatch
+        // Confirm Remove this Device
+        let alertLogoutButton = alerts.firstMatch.buttons[A11yIdentifiers.alertInfo.primaryButton].firstMatch
         XCTAssertTrue(alertLogoutButton.waitForExistence(timeout: 10.0))
         alertLogoutButton.tap(.center)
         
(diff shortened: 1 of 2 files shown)


pixlwave (approved):
LGTM, thanks! 👍

Edit: Ah but I see it didn't work on CI 🙁

manuroe:
Closing it for the moment. I will continue to run the integration CI on a branch (if I don't forget it).

manuroe:
This PR is actually valid to run tests locally. CI is another story and should be addressed separetely.

pixlwave (approved):
Thanks Manu, I've tested this locally and split it into 2 parts. What remains in this PR is the base changes required for the tests to pass locally. And then https://github.com/element-hq/element-x-ios/tree/manu/fix-user-flow-tests contains the remaining the changes the AI made to attempt to fix CI.

IntegrationTests/Sources/Common.swift
```diff
@@ -74,6 +74,14 @@ extension XCUIApplication {
...
         XCTAssertTrue(webAuthenticationView.waitForExistence(timeout: 10.0))
         webAuthenticationView.tap(.top) // Tap the web view to properly focus the app again.
+
+        // The user may already be authenticated on MAS. Sign them out.
+        // The button label varies by MAS version: "Sign out" or "Use another account".
+        let webLogoutPredicate = NSPredicate(format: "label == 'Sign out' OR label == 'Use another account'")
+        let webLogoutButton = webAuthenticationView.buttons.matching(webLogoutPredicate).firstMatch
+        if webLogoutButton.waitForExistence(timeout: 2.0) {
+            webLogoutButton.tap(.center)
+        }
```
pixlwave, line 84: Nice quality of life improvement 🙌

PR #5354, opened 2026-04-04
Back off sync restarts after repeated sync failures

microchipster:
### Pull Request Checklist

- [ ] Voiceover enabled.

## Summary

Client side mitigation for this: https://github.com/element-hq/synapse/pull/19651

Android version: https://github.com/element-hq/element-x-android/pull/6532

This adds bounded backoff to the sync restart path in `ClientProxy` so the app stops hammering the same failing sync state over and over.

This is a client-side mitigation, not the root-cause fix. The underlying server-side failure is addressed in.

## Motivation and context

The reported issue was that after leaving a room in a bad server state, Element X could get stuck syncing forever until local state was reset.

The useful diagnosis came from synapse logs, not the iOS app log:

- incremental Sliding Sync was returning HTTP 500
- Synapse was asserting in `synapse/handlers/sliding_sync/room_lists.py` while reconstructing newly-left rooms

On the iOS side, `ClientProxy` restarts sync on every `.error` state. If the same persisted sync position keeps reproducing the same server failure, the app just re-enters the loop immediately.

This change adds bounded backoff of `250ms`, `1s`, `3s`, `10s`, and `30s`, and resets that backoff once sync genuinely reaches `.running` again.

## Testing

- Reviewed the `ClientProxy` sync error and restart flow after the change.
- Verified the retry path now backs off based on consecutive failures and resets after `.running`.
- Don't have or want access to Apple devices to test.

### DIFF
--- a/ElementX/Sources/Services/Client/ClientProxy.swift
+++ b/ElementX/Sources/Services/Client/ClientProxy.swift
@@ -147,6 +147,9 @@ class ClientProxy: ClientProxyProtocol {
     /// before the client is released which ends up running in a loop. This is a workaround until the sync service
     /// can tell us *what* error occurred so we can handle restarts more gracefully.
     private var hasEncounteredAuthError = false
+
+    /// Tracks consecutive sync failures so we can slow down retries if the same persisted sync state keeps failing.
+    private var consecutiveSyncFailureCount = 0
     
     deinit {
         stopSync { [delegateHandle] in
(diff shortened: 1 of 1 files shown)


stefanceriu:
Thanks for raising this! Ideally it would happen in the SDK itself but I think it's fine to have it in the client until/if that happens. Please sign the CLA so I can go ahead and merge it 🙏

stefanceriu:
New plan: we raised this in the rust team sync meeting today (cc @jmartinesp) and we agreed to implemented it only once, on the RustSDK side within the SyncService, instead of in each client. 
As such I'll close this PR, thank you for the idea!

PR #5357, opened 2026-04-07
Translations update

RiotRobot:
- Translations update


pixlwave:
Looking at the diffs on this one it appears that Localazy is now sorting the strings in a different order on download.

PR #5361, opened 2026-04-08
Add a view to show the user's account from Element Classic.

pixlwave:
Same as #5359, this doesn't contain any presentation logic, it is just to start integrating parts of my working branch in smaller chunks.

Screenshot 2026-04-08 at 11 19 13 am


ElementX/Sources/Screens/Authentication/StartScreen/View/AuthenticationClassicAppAccountView.swift
```diff
@@ -0,0 +1,120 @@
...
+            String(localPart)
+        } else {
+            userID
+        }
+    }
+}
+
+// MARK: - Previews
+
+struct AuthenticationClassicAppAccountView_Previews: PreviewProvider { // Not Testable – snapshots generated by main screen.
```
pixlwave, line 103: Already done in my working branch so will happen in a follow-up PR – doesn't seem worth adding snapshots just for the sake of coverage.

ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
```diff
@@ -263,6 +263,7 @@ class AuthenticationFlowCoordinator: FlowCoordinatorProtocol {
                                                              provisioningParameters: provisioningParameters,
                                                              isBugReportServiceEnabled: bugReportService.isEnabled,
                                                              appSettings: appSettings,
+                                                             mediaProvider: nil, // Currently unused.
```
pixlwave, line 266: Looking at this now, I might actually replace `ClassicAppMediaLoader` with a `ClassicAppClient` as there's a request to get the key backup version that is elsewhere and it would be nice to combine the two. That would mean this optional parameter could go and instead be passed directly to the `LoadableAvatarImage` from the `ClassicAppAccount`, but I need to explore that a bit more to decide if it's worthwhile.

ElementX/Sources/Mocks/ClassicAppManagerMock.swift
```diff
@@ -0,0 +1,41 @@
...
+// Copyright 2026 Element Creations Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Foundation
+import MatrixRustSDK
+
+extension ClassicAppManagerMock {
```
pixlwave, line 11: This is just a placeholder to include the mock accounts below. There's more to include once the SDK PR is merged.

ElementX/Sources/Screens/Authentication/StartScreen/View/AuthenticationClassicAppAccountView.swift
```diff
@@ -0,0 +1,120 @@
...
+        }
+    }
+    
+    var profile: some View {
+        VStack(spacing: 16) {
+            LoadableAvatarImage(url: classicAppAccount.avatarURL,
+                                name: classicAppAccount.displayName,
+                                contentID: classicAppAccount.userID,
+                                avatarSize: .user(on: .classicAppAccount),
+                                mediaProvider: context.mediaProvider)
```
Velin92, line 53: so this is nil now however right? which means we are going to have a different media loader for the classic app data?

ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
Velin92: but why pass nil and not passing the Classic one here?

Velin92 (approved):
LGTM but I have a few questions

ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
pixlwave: Ah, because the classic one needs a `ClassicAppAccount` so it has a homeserver URL and an access token. So this will actually be passed in like so:

```swift
let mediaProvider = authenticationService.classicAppAccount.map { account in
    MediaProvider(mediaLoader: ClassicAppMediaLoader(classicAppAccount: account),
                  imageCache: .onlyInMemory,
                  homeserverReachabilityPublisher: appMediator.networkMonitor.reachabilityPublisher) // Close enough approximation
}
```

ElementX/Sources/Screens/Authentication/StartScreen/View/AuthenticationClassicAppAccountView.swift
pixlwave: And then when passed in as mentioned above, it will be non-nil when there's a Classic account, meaning in this view it will always be non-nil as it is only reachable when there's an account.

PR #5362, opened 2026-04-08
Fix voice over not reading voice messages properly

### LINKED ISSUE #5353: A11y: Voiceover incorrectly reading voice messages
### Steps to reproduce

1. Turn on Voiceover in iOS accessibility settings.
2. Open Element X app, open a chat and try reading messages.
3. Focus on a voice message. Instead of "voice message, duration x seconds", it just says "*1".
4. Now leave the chat, and under the only chat member I'm chatting with, i heard an unlabeled button.

#### What did you expect?
Awesome accessibility, very plesent to use with voiceover, buttons are labeled, chat messages are correctly ordered!
#### What happened instead?
Some unlabeled buttons that are impossible to understand using voiceover. Also, chat message order is reversed or viewed in reverse by voiceover. I mean last messages on top.

### Your phone model

iPhone 14

### Operating system version

IOs 26.4

### Application version

26.03.3 (215)

### Homeserver

It is on synapse last version that was available in APT for 1st of april

### Will you send logs?

Yes

Velin92:
fixes #5353

### DIFF
--- a/ElementX/Sources/Screens/Timeline/View/VoiceMessageRoomPlaybackView.swift
+++ b/ElementX/Sources/Screens/Timeline/View/VoiceMessageRoomPlaybackView.swift
@@ -48,6 +48,7 @@ struct VoiceMessageRoomPlaybackView: View {
         .onChange(of: isDragging) { _, newValue in
             onScrubbing(newValue)
         }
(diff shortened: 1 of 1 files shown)


pixlwave (approved):
Nice 👌

PR #5366, opened 2026-04-09
Require video for callkit even if voice call only to workaround a bug answering calls on the lock screen

BillCarsonFr:
### Pull Request Checklist

Work-around for https://github.com/element-hq/element-x-ios/issues/5335

Apparently we are facing a limitation of compatibility between callkit and webviews :/
If we want great call support, we would need to implement native Element Call.

The only work-around I see for now is to still tell CallKit that we want video, like that it will properly start the app.
Then video will be disabled on EC.

Downsides:
- The call will be presented to the user as a video call in CallKit UI.
- Requires a FaceID to answer

It partially reverts the changes from https://github.com/element-hq/element-x-ios/pull/5298 (regarding callkit UI)

- [ ] Voiceover enabled.


pixlwave (approved):
LGTM, thanks for the fix!

pixlwave:
Ah looks like there's a test that needs fixing too.

PR #5370, opened 2026-04-09
Live Location Sharing Banner

### LINKED ISSUE #5202: [Task] EX LLS: Show LLS banner below room header

### LINKED ISSUE #5203: [Task] EX LLS: Start live location sharing

Velin92:
This PR implementes the live location sharing banner view.
It improves the existing top banner modifier to also render two banners one after another in the vertical axis.
Adds and updates preview tests for the various banners, and adds tests to the LiveLocationManager.
And also improves the rendering of the expired state when tapping stop inside the live location timeline item.

fixes #5203 
fixes #5202


ElementX/Resources/Localizations/en-US.lproj/Localizable.strings
Velin92, line 1: Not sure what was happening while working on this, but seems there was a massive change of strings on localazy, which in turn caused the strings to get re-arranged in their order.
@pixlwave

ElementX/Sources/Screens/RoomScreen/View/LiveLocationSharingBannerView.swift
```diff
@@ -0,0 +1,42 @@
...
+            CompoundIcon(\.locationPinSolid, size: .medium, relativeTo: .compound.bodyMDSemibold)
+                .foregroundColor(Color.compound.iconSuccessPrimary)
+                .accessibilityHidden(true)
+            Text(L10n.screenRoomLiveLocationBanner)
+                .font(.compound.bodyMDSemibold)
+                .foregroundColor(.compound.textPrimary)
+            Spacer()
+            Button(L10n.actionStop, role: .destructive) {
+                onStop()
+            }
```
pixlwave: Given it the action doesn't take any parameters you don't necessarily need the closure:

```suggestion
            Button(L10n.actionStop, role: .destructive, action: onStop)
```

ElementX/Sources/Screens/RoomScreen/View/LiveLocationSharingBannerView.swift
```diff
@@ -0,0 +1,42 @@
...
+            .buttonStyle(.compound(.primary, size: .small))
+        }
+        .padding(.vertical, 16)
+        .padding(.horizontal, 15)
+        .background(Color.compound.bgCanvasDefault)
+        .overlay(Rectangle()
+            .inset(by: 0.5)
+            .stroke(.compound.separatorPrimary))
```
pixlwave: Just double checking, is it intentional that we're also drawing vertical separators on both sides of the view too, I thought it was only meant to be horizontal separators at the top/bottom?

ElementX/Sources/Services/Location/LiveLocationManager.swift
```diff
@@ -60,6 +60,10 @@ class LiveLocationManager: NSObject, LiveLocationManagerProtocol, CLLocationMana
     }
     
     func startLiveLocation(roomID: String, duration: Duration) async -> Result<Void, LiveLocationManagerError> {
+        if appSettings.liveLocationSharingTimeoutDatesByRoomID[roomID] != nil {
+            await stopLiveLocation(roomID: roomID)
+        }
```
pixlwave, line 67: Is this in case you're sharing from a different device, or something else I can't think of? Might be nice to add a comment as to why we're stopping and then restarting rather than leaving it to run.

UnitTests/Sources/LiveLocationManagerTests.swift
```diff
@@ -0,0 +1,180 @@
...
+final class LiveLocationManagerTests {
+    private var clientProxy: ClientProxyMock!
+    private var manager: LiveLocationManager!
+    
+    private var appSettings: AppSettings {
+        ServiceLocator.shared.settings
+    }
+    
+    init() {
+        AppSettings.resetAllSettings()
```
pixlwave, line 20: I don't think we really need to add a new dependency on the `ServiceLocator` here…

```suggestion
    private let appSettings: AppSettings
    
    init() {
        AppSettings.resetAllSettings()
        appSettings = AppSettings()
```

UnitTests/Sources/LiveLocationManagerTests.swift
```diff
@@ -0,0 +1,180 @@
...
+//
+
+@testable import ElementX
+import Foundation
+import Testing
+
+@MainActor
+final class LiveLocationManagerTests {
+    private var clientProxy: ClientProxyMock!
+    private var manager: LiveLocationManager!
```
pixlwave, line 15: These 2 can be `let` and not force unwrapped too I think 🤔

UnitTests/Sources/LiveLocationManagerTests.swift
```diff
@@ -0,0 +1,180 @@
...
+    func startLiveLocationWithNoExistingSession() async {
+        let roomProxy = makeRoomProxy(roomID: "!room:matrix.org")
+        clientProxy.roomForIdentifierClosure = { _ in .joined(roomProxy) }
+        
+        let result = await manager.startLiveLocation(roomID: "!room:matrix.org", duration: .seconds(300))
+        
+        guard case .success = result else {
+            Issue.record("Expected success but got \(result)")
+            return
+        }
```
pixlwave: Just learnt/realised we can actually do this for the case where we don't care about the success's value:

```
try #require(throws: Never.self) { try result.get() }
```

or this when we do want the value:

```
let value = try #require { try result.get() }
```

UnitTests/Sources/LiveLocationManagerTests.swift
```diff
@@ -0,0 +1,180 @@
...
+    @Test
+    func startLiveLocationWhenRoomNotJoined() async {
+        clientProxy.roomForIdentifierClosure = { _ in nil }
+        
+        let result = await manager.startLiveLocation(roomID: "!room:matrix.org", duration: .seconds(300))
+        
+        guard case .failure(.roomNotJoined) = result else {
+            Issue.record("Expected roomNotJoined failure but got \(result)")
+            return
+        }
```
pixlwave: And can do this here:

```
#expect(throws: LiveLocationManagerError.roomNotJoined) { try result.get() }
```

UnitTests/Sources/LiveLocationManagerTests.swift
```diff
@@ -0,0 +1,180 @@
...
+    }
+    
+    @Test
+    func startLiveLocationDoesNotStopSessionForOtherRoom() async {
+        let roomProxy = makeRoomProxy(roomID: "!room1:matrix.org")
+        clientProxy.roomForIdentifierClosure = { _ in .joined(roomProxy) }
+        
+        appSettings.liveLocationSharingTimeoutDatesByRoomID["!room2:matrix.org"] = Date().addingTimeInterval(300)
+        
+        _ = await manager.startLiveLocation(roomID: "!room1:matrix.org", duration: .seconds(300))
```
pixlwave, line 75: Could also `#require` the `.get` on this one too.

UnitTests/Sources/LiveLocationManagerTests.swift
```diff
@@ -0,0 +1,180 @@
...
+    }
+    
+    @Test
+    func startLiveLocationStoresTimeoutDate() async throws {
+        let roomProxy = makeRoomProxy(roomID: "!room:matrix.org")
+        clientProxy.roomForIdentifierClosure = { _ in .joined(roomProxy) }
+        
+        let duration = Duration.seconds(300)
+        let beforeStart = Date()
+        _ = await manager.startLiveLocation(roomID: "!room:matrix.org", duration: duration)
```
pixlwave, line 110: And this one.

UnitTests/Sources/LiveLocationManagerTests.swift
```diff
@@ -0,0 +1,180 @@
...
+        let afterStart = Date()
+        
+        let storedTimeout = appSettings.liveLocationSharingTimeoutDatesByRoomID["!room:matrix.org"]
+        let expectedMinTimeout = beforeStart.addingTimeInterval(TimeInterval(duration.seconds))
+        let expectedMaxTimeout = afterStart.addingTimeInterval(TimeInterval(duration.seconds))
+        
+        #expect(storedTimeout != nil)
+        #expect(try #require(storedTimeout) >= expectedMinTimeout)
+        #expect(try #require(storedTimeout) <= expectedMaxTimeout)
```
pixlwave: I think you can do this with a single range check:

```suggestion
        #expect((expectedMinTimeout...expectedMaxTimeout).contains(try #require(storedTimeout)))
```

or apparently even

```suggestion
        #expect(expectedMinTimeout...expectedMaxTimeout ~= try #require(storedTimeout))
```

ElementX/Sources/Other/SwiftUI/Views/TopBannerModifier.swift
pixlwave, line 1: Ok now I've seen this in action how about this for the naming:

- Keep `TopBannerItem` as is, it is a banner item given to the topBanner modifier.
- Rename `ZBannerItem` to `TopBannerLayer` but keep it's `verticalBanners` property named that way.

This way you would call

```swift
.topBanners([
    TopBannerLayer(verticalBanners: [
        TopBannerItem(pinnedItemsBanner, isVisible: shouldShowPinnedEventsBanner),
        TopBannerItem(liveLocationBanner, isVisible: isSharingLiveLocation)
    ]),
    // This can overlay on top of the stacked banners
    TopBannerLayer(knockRequestsBanner, isVisible: shouldSeeKnockRequests)
], footer: dateBadge)
```

Which seems to describe how it works pretty clearly to me. WDYT?

pixlwave (approved):
Love it 🙌

ElementX/Sources/Other/SwiftUI/Views/TopBannerModifier.swift
Velin92: I loik it!

PR #5374, opened 2026-04-10
Verify Element X with an existing Element Classic account.

### LINKED ISSUE #5053: [Task] Sign in with Element Classic

pixlwave:
Ties together #5181, #5359 and #5361 with the logic for a seamless login of an account from Element Classic (so long as it is verified and has key backup enabled).

Closes #5053.

https://github.com/user-attachments/assets/3b0583f3-2cad-4539-9207-e75d82faf71b


ElementX/Sources/Screens/Authentication/StartScreen/View/AuthenticationClassicAppBackupInstructionsView.swift
```diff
@@ -44,10 +53,20 @@ struct AuthenticationClassicAppBackupInstructionsView: View {
     }
     
     var buttons: some View {
-        Button(L10n.screenMissingKeyBackupOpenElementClassic) {
+        Button {
             UIApplication.shared.open("element://open")
```
Velin92: just noticed this, but could we handle it at the view model level using the app mediator?

ElementX/Sources/Screens/Authentication/StartScreen/AuthenticationStartScreenViewModel.swift
```diff
@@ -121,6 +152,25 @@ class AuthenticationStartScreenViewModel: AuthenticationStartScreenViewModelType
...
+        
+        reloadClassicAppSecretsTask = Task { [weak self] in
+            await self?.authenticationService.refreshClassicAppAccountState()
+            
+            guard !Task.isCancelled else { return }
+            
+            if let availableSecrets = classicAppAccount.state.availableSecrets, availableSecrets != .requiresBackup {
+                Task { // The sheet doesn't dismiss unless we call this on the next loop 🤷‍♂️
+                    self?.state.bindings.showClassicAppBackupInstructions = false
+                }
```
Velin92: could you try doing instead `await MainActor.run { self?.state.bindings.showClassicAppBackupInstructions = false }` I think this would solve the problem. Or at least add a `@MainActor` to the internal task if you want to keep the handling async from this code.

UnitTests/Sources/AuthenticationServiceTests.swift
```diff
@@ -90,22 +91,113 @@ struct AuthenticationServiceTests {
...
+        try await setup(classicAppAccounts: [.mockAlice])
+        guard case .success = await service.configure(for: "matrix.org", flow: .login) else {
+            Issue.record("The service should be configured successfully.")
+            return
+        }
+        #expect(service.flow == .login)
+        #expect(service.classicAppAccount?.state.availableSecrets == .complete)
+        
+        // When logging in as Alice.
+        guard case .success = await service.login(username: "alice", password: "12345678", initialDeviceName: nil, deviceID: nil) else {
```
Velin92: you can also do `try await service.login(...).get()` but up to you no preferences, same for all the other cases (if you want to avoid writing the Issue yourself, and have the code just throw the error directly)

UnitTests/Sources/AuthenticationStartScreenViewModelTests.swift
```diff
@@ -139,19 +141,176 @@ final class AuthenticationStartScreenViewModelTests {
...
+        setupViewModel(classicAppAccount: classicAppAccount, availableSecrets: .requiresBackup)
+        guard case .welcomeBack(let account) = context.viewState.classicAppMode else {
+            Issue.record("Expected classicAppMode to be .welcomeBack")
+            return
+        }
+        
+        if account.state.availableSecrets != .requiresBackup {
+            let deferredSecrets = deferFulfillment(account.state.observe(\.availableSecrets)) { $0 == .requiresBackup }
+            try await deferredSecrets.fulfill()
+        }
```
Velin92: should this be an `#expect` or just use the fulfillment immediately? why this check?

Velin92 (approved):
LGTM, few things

UnitTests/Sources/AuthenticationServiceTests.swift
pixlwave: Oh yeah nice, done them all.

UnitTests/Sources/AuthenticationStartScreenViewModelTests.swift
pixlwave: My bad, this can just be the fulfilment, the `if` isn't doing anything.

ElementX/Sources/Screens/Authentication/StartScreen/AuthenticationStartScreenViewModel.swift
pixlwave: Ok yeah that fixed it (and it actually animates now too). I tried an `@MainActor` on the task before but it didn't work so removed it.

ElementX/Sources/Screens/Authentication/StartScreen/View/AuthenticationClassicAppBackupInstructionsView.swift
pixlwave: Oh yeah fair, it should also be set alongside the other Classic app configuration too. I'm going to do this in a separate PR.

pixlwave: Done in https://github.com/element-hq/element-x-ios/pull/5374/commits/6bddf8e0d6afcba1570fdb7e4f866543ff8c2abb whilst fixing the tests.

Velin92 (approved):

pixlwave:
The test failures are some flakey location sharing tests so I'm going to ignore them for now and merge.

PR #5377, opened 2026-04-10
Order localazy strings alphabetically by key

Velin92:
Localazy has done some weird changes lately in the way the localazation projects are generated, messing up the current order every single time, and making diffs for PRs insanely big, for no real reason.
This PR just updates our tool, so that the strings in the files are always ordered alphabetically by key.


Tools/Sources/Commands/DownloadStrings.swift
```diff
@@ -10,13 +10,67 @@ struct DownloadStrings: ParsableCommand {
...
+    private func sortStringsFiles() throws {
+        let localizationsURL = URL(fileURLWithPath: "ElementX/Resources/Localizations")
+        let fileManager = FileManager.default
+        
+        guard let enumerator = fileManager.enumerator(at: localizationsURL,
+                                                      includingPropertiesForKeys: nil) else {
+            return
+        }
+        
+        for case let fileURL as URL in enumerator where fileURL.pathExtension == "strings" {
```
pixlwave: Do you think we'll need to do something for the `stringsdict`s too?

Tools/Sources/Commands/DownloadStrings.swift
```diff
@@ -10,13 +10,67 @@ struct DownloadStrings: ParsableCommand {
...
+        let sortedContent = sortedLines.joined(separator: "\n") + "\n"
+        try sortedContent.write(to: url, atomically: true, encoding: .utf8)
+    }
+    
+    private func extractKey(from line: String) -> String? {
+        guard let openQuote = line.firstIndex(of: "\"") else { return nil }
+        let afterOpen = line.index(after: openQuote)
+        guard let closeQuote = line[afterOpen...].firstIndex(of: "\"") else { return nil }
+        return String(line[afterOpen..<closeQuote])
+    }
```
pixlwave, line 77: lol seems odd it didn't use a regex in this one 😅

pixlwave (approved):
Very nice. Might be good to try a more modern implementation, but either way, this is way better 👏

Tools/Sources/Commands/DownloadStrings.swift
Velin92: I guess makes sense, the key should be only ASCII characters and is always the first string between two quotes, so I guess that's okay.

Tools/Sources/Commands/DownloadStrings.swift
Velin92: yeah we should, is also quite easy since the `PropertyListSerialization `does all of the work for us.

PR #5391, opened 2026-04-13
Restore the .oidcCallback route for external authentication.

pixlwave:
Partially reverts https://github.com/element-hq/element-x-ios/pull/3461 so that if MAS was to open the authorisation page in an external app, Element X is once again able to forward the redirect to the OIDC presenter.

### DIFF
--- a/ElementX/Sources/Application/Navigation/AppRoutes.swift
+++ b/ElementX/Sources/Application/Navigation/AppRoutes.swift
@@ -14,6 +14,9 @@ import MatrixRustSDK
 enum AppRoute: Hashable {
     /// An account provisioning link generated externally.
     case accountProvisioningLink(AccountProvisioningParameters)
+    /// An external callback used to complete login with OIDC. This is only used when authentication
+    /// requires an external app so cannot be done within the built in web authentication session.
+    case oidcCallback(url: URL)
     
     /// The app's home screen.
     case roomList
@@ -58,6 +61,7 @@ enum AppRoute: Hashable {
     var isAuthenticationRoute: Bool {
         switch self {
         case .accountProvisioningLink: true
+        case .oidcCallback: true
         default: false
         }
     }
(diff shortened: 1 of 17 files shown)


ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
```diff
@@ -424,10 +417,14 @@ class AuthenticationFlowCoordinator: FlowCoordinatorProtocol {
         }
     }
     
-    private func showOIDCAuthentication(oidcData: OIDCAuthorizationDataProxy, presentationAnchor: UIWindow, fromState: State) {
+    /// **Note:** We have intentionally excluded this presentation from the state machine as it doesn't mutate our navigation stack and there
+    /// isn't a robust way to detect why the user returned to the app when the MAS URL directly opens an external app for authentication without
+    /// presenting a web authentication session.
```
stefanceriu, line 422: 👌

ElementX/Sources/Application/Navigation/AppRoutes.swift
```diff
@@ -198,3 +203,13 @@ private struct AccountProvisioningURLParser: URLParser {
         return .accountProvisioningLink(.init(accountProvider: serverName, loginHint: loginHint))
     }
 }
+
+/// The parser for the OIDC callback URL. This always returns a `.oidcCallback`.
+struct OIDCCallbackURLParser: URLParser {
+    let redirectURL: URL
```
stefanceriu, line 209:
```suggestion
    private let redirectURL: URL
```

stefanceriu (approved):
Nice, no surprises!

ElementX/Sources/Application/Navigation/AppRoutes.swift
pixlwave: Good spot, I'll do it in #5525 after the rebase so this one can merge right away.

pixlwave: Oh no wait, it's a struct so it is non-private to get the synthesised init.

PR #5393, opened 2026-04-14
Implement APIs for Live Location Sharing observation

Velin92:
In the next PR I'll have the location sharing screen react to the observation of LLS.

For now this just adds all the mocks and the API functions to allow observing LLS (which also includes a refactor of the screen to work reactively)


ElementX/Sources/Services/Location/LiveLocationShareProxy.swift
```diff
@@ -0,0 +1,29 @@
...
+// Copyright 2026 Element Creations Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Foundation
+import MatrixRustSDK
+
+struct LiveLocationShareProxy: Hashable {
```
pixlwave: This isn't proxying anything so let's not call it one. We can do `MatrixRustSDK.LiveLocationShare` whenever there's a conflict.

```suggestion
struct LiveLocationShare: Hashable {
```

ElementX/Sources/Services/Location/LiveLocationShareProxy.swift
```diff
@@ -0,0 +1,29 @@
...
+}
+
+extension LiveLocationShareProxy {
+    init(liveLocationShare: LiveLocationShare) {
+        userID = liveLocationShare.userId
+        if let geoURI = liveLocationShare.lastLocation?.location.geoUri {
+            self.geoURI = GeoURI(string: geoURI)
+        } else {
+            geoURI = nil
+        }
```
pixlwave: I don't think we need the if let here…

```suggestion
        geoURI = liveLocationShare.lastLocation?.location.geoUri.map(GeoURI.init)
```

ElementX/Sources/Services/Room/JoinedRoomProxy.swift
```diff
@@ -754,6 +754,10 @@ class JoinedRoomProxy: JoinedRoomProxyProtocol {
     
     // MARK: - Live Location
     
+    func getLiveLocationSharesService() async -> LiveLocationSharesServiceProtocol {
+        await LiveLocationSharesService(liveLocationShares: room.liveLocationShares())
+    }
```
pixlwave: Let's be super clear here that we're making one on each call rather than getting an existing one perhaps?

```suggestion
    func makeLiveLocationSharesService() async -> LiveLocationSharesServiceProtocol {
        await LiveLocationSharesService(liveLocationShares: room.liveLocationShares())
    }
```

ElementX/Sources/Services/Location/LiveLocationSharesServiceProtocol.swift
```diff
@@ -0,0 +1,14 @@
...
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Combine
+import Foundation
+
+// sourcery: AutoMockable
+protocol LiveLocationSharesServiceProtocol {
+    var liveLocationSharesPublisher: AnyPublisher<[LiveLocationShareProxy], Never> { get }
+}
```
pixlwave: I find the naming of this very confusing and it made me ask "what's the difference between the `LiveLocationManager` and `LiveLocationSharesService`.

We should definitely include `Room` in the name of this one given it is scoped to a specific room, and maybe remove the `Shares` since it might be possible to use it for other things in the future?

Also I wonder if we really need to repeat that the locations publisher is a `liveLocationSharesPublisher`, most of that seems implied by the name of the service?

```suggestion
protocol RoomLiveLocationServiceProtocol {
    var locationsPublisher: AnyPublisher<[LiveLocationShareProxy], Never> { get }
}
```

ElementX/Sources/Services/Location/RoomLiveLocationService.swift
```diff
@@ -0,0 +1,72 @@
...
+    private var observationToken: TaskHandle?
+    
+    private let liveLocationSharesSubject = PassthroughSubject<[LiveLocationShareProxy], Never>()
+    var liveLocationSharesPublisher: AnyPublisher<[LiveLocationShareProxy], Never> {
+        liveLocationSharesSubject.eraseToAnyPublisher()
+    }
+    
+    private var previousLiveLocationShares: [LiveLocationShareProxy] = []
+    
+    init(liveLocationShares: LiveLocationShares) {
```
pixlwave, line 25: Wait, I don't understand this. It takes a `LiveLocationsShares` and subscribes to it? What is a `LiveLocationShares`?

pixlwave: Ok just found this. I change my idea on naming. The SDK object should be called `RoomLiveLocationService` and then on our side this should be called a `RoomLiveLocationServiceProxy`.

pixlwave (approved):
Fundamentally makes sense to me but the naming is confusing. I think the SDK needs a tweak too to fully make sense.

PR #5394, opened 2026-04-14
Render LiveLocations in the map

Velin92:
This PR listens to the LLS APIs that were previously implemented, and allows the map to reactively update to them and display markers on the map based on the current annotations that are generated by the LLS updates.

To do this the location sharing screen and its view model have been somewhat refactored, to support this implementation.

Did not add preview tests since the changes to the view are not done yet.
I suggest to review this commit by commit.

Screenshot 2026-04-14 at 15 09 32

### DIFF
--- a/ElementX/Sources/Other/MapLibre/LocationAnnotation.swift
+++ b/ElementX/Sources/Other/MapLibre/LocationAnnotation.swift
@@ -6,31 +6,23 @@
 // Please see LICENSE files in the repository root for full details.
 //
 
+import CoreLocation
 import Foundation
 import MapLibre
 import SwiftUI
 
-final class LocationAnnotation: NSObject, MLNAnnotation {
+final class LocationAnnotation: NSObject, MLNAnnotation, Identifiable {
     let id: String
-    var coordinate: CLLocationCoordinate2D
-    let anchorPoint: CGPoint
-    var view: AnyView
+    dynamic var coordinate: CLLocationCoordinate2D
+    var kind: LocationMarkerKind
     
     // MARK: - Setup
     
-    init(id: String = UUID().uuidString,
-         coordinate: CLLocationCoordinate2D,
-         anchorPoint: CGPoint = .init(x: 0.5, y: 0.5),
-         @ViewBuilder label: () -> some View) {
+    init(id: String, coordinate: CLLocationCoordinate2D, kind: LocationMarkerKind) {
         self.id = id
         self.coordinate = coordinate
-        self.anchorPoint = anchorPoint
-        view = AnyView(label())
+        self.kind = kind
         super.init()
-    }
-    
-    func updateView(@ViewBuilder label: () -> some View) {
-        view = AnyView(label())
     }
 }
 
@@ -44,19 +36,21 @@ final class LocationAnnotationView: MLNUserLocationAnnotationView {
             reuseIdentifier)
     }
     
-    convenience init(annotation: LocationAnnotation) {
+    convenience init(annotation: LocationAnnotation, mediaProvider: MediaProviderProtocol?) {
         self.init(annotation: annotation, reuseIdentifier: "\(Self.self)")
-        let hostingController = UIHostingController(rootView: annotation.view)
+        let markerView = LocationMarkerView(kind: annotation.kind, mediaProvider: mediaProvider)
+        let hostingController = UIHostingController(rootView: AnyView(markerView))
         self.hostingController = hostingController
         let view: UIView = hostingController.view
         view.backgroundColor = .clear
-        view.anchorPoint = annotation.anchorPoint
+        view.anchorPoint = .init(x: 0.5, y: 1.0)
         addSubview(view)
         view.bounds.size = view.intrinsicContentSize
     }
-        
-    func updateContent(with view: AnyView) {
-        hostingController?.rootView = view
+    
+    func updateContent(with kind: LocationMarkerKind, mediaProvider: MediaProviderProtocol?) {
+        let markerView = LocationMarkerView(kind: kind, mediaProvider: mediaProvider)
+        hostingController?.rootView = AnyView(markerView)
         if let hostedView = hostingController?.view {
             hostedView.bounds.size = hostedView.intrinsicContentSize
         }
--- a/ElementX/Sources/Other/MapLibre/LocationAnnotation.swift
+++ b/ElementX/Sources/Other/MapLibre/LocationAnnotation.swift
@@ -13,7 +13,7 @@ import SwiftUI
 
 final class LocationAnnotation: NSObject, MLNAnnotation, Identifiable {
     let id: String
-    dynamic var coordinate: CLLocationCoordinate2D
+    var coordinate: CLLocationCoordinate2D
     var kind: LocationMarkerKind
     
     // MARK: - Setup
--- a/ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
+++ b/ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
@@ -26,6 +26,7 @@ enum LocationSharingScreenViewModelAction {
 enum LocationSharingInteractionMode: Hashable {
     case picker
     case viewStatic(StaticLocationData)
+    case viewLive(sender: TimelineItemSender, liveLocationShare: LiveLocationShare)
 }
 
 struct LocationSharingScreenViewState: BindableState {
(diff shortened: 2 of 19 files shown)


ElementX/Sources/Screens/PinnedEventsTimelineScreen/PinnedEventsTimelineScreenCoordinator.swift
```diff
@@ -27,6 +27,7 @@ enum PinnedEventsTimelineScreenCoordinatorAction {
     case dismiss
     case displayUser(userID: String)
     case presentLocationViewer(StaticLocationData)
+    case presentLiveLocationViewer(sender: TimelineItemSender, liveLocationShare: LiveLocationShare)
```
pixlwave: Do you think we should just merge these 2 together with the interaction mode parameter? If so then presumably the same would apply in the view model too.

ElementX/Sources/Screens/RoomScreen/RoomScreenCoordinator.swift
```diff
@@ -40,6 +40,7 @@ enum RoomScreenCoordinatorAction {
     case presentLocationPicker
     case presentPollForm(mode: PollFormMode)
     case presentLocationViewer(StaticLocationData)
+    case presentLiveLocationViewer(sender: TimelineItemSender, liveLocationShare: LiveLocationShare)
```
pixlwave: Ditto?

ElementX/Sources/Screens/ThreadTimelineScreen/ThreadTimelineScreenCoordinator.swift
```diff
@@ -34,6 +34,7 @@ enum ThreadTimelineScreenCoordinatorAction {
     case presentMediaUploadPreviewScreen(mediaURLs: [URL])
     case presentLocationPicker
     case presentLocationViewer(StaticLocationData)
+    case presentLiveLocationViewer(sender: TimelineItemSender, liveLocationShare: LiveLocationShare)
```
pixlwave: Ditto 😄

ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
```diff
@@ -540,12 +540,21 @@ class TimelineInteractionHandler {
...
+                                                      timeoutDate: item.content.timeoutDate)
+            return .displayLiveLocation(sender: item.sender, liveLocationShare: liveLocationShare)
+        case let item as ImageRoomTimelineItem:
+            return await mediaPreviewAction(for: item, messageTypes: [.image, .video])
+        case let item as VideoRoomTimelineItem:
+            return await mediaPreviewAction(for: item, messageTypes: [.image, .video])
+        case let item as AudioRoomTimelineItem:
+            return await mediaPreviewAction(for: item, messageTypes: [.audio, .file])
+        case let item as FileRoomTimelineItem:
+            return await mediaPreviewAction(for: item, messageTypes: [.audio, .file])
```
pixlwave, line 557: Errr, it looks like it's split these cases up for no reason?

ElementX/Sources/Screens/Timeline/TimelineModels.swift
```diff
@@ -24,6 +24,7 @@ enum TimelineViewModelAction {
     case displayMessageForwarding(forwardingItem: MessageForwardingItem)
     case displayMediaPreview(TimelineMediaPreviewViewModel)
     case displayLocation(StaticLocationData)
+    case displayLiveLocation(sender: TimelineItemSender, liveLocationShare: LiveLocationShare)
```
pixlwave: Ditto

ElementX/Sources/Services/Timeline/TimelineController/TimelineControllerProtocol.swift
```diff
@@ -26,6 +26,7 @@ enum TimelineControllerAction {
     
     case displayMediaPreview(item: EventBasedMessageTimelineItemProtocol, timelineViewModel: TimelineViewModelKind)
     case displayLocation(StaticLocationData)
+    case displayLiveLocation(sender: TimelineItemSender, liveLocationShare: LiveLocationShare)
```
pixlwave: Ditto

ElementX/Sources/Other/MapLibre/LocationAnnotation.swift
pixlwave, line 1: Nice, this was always weird how the annotation owned a view. 👏

ElementX/Sources/Other/MapLibre/LocationAnnotation.swift
```diff
@@ -6,32 +6,24 @@
...
 import MapLibre
 import SwiftUI
 
-final class LocationAnnotation: NSObject, MLNAnnotation {
+final class LocationAnnotation: NSObject, MLNAnnotation, Identifiable {
     let id: String
-    var coordinate: CLLocationCoordinate2D
-    let anchorPoint: CGPoint
-    var view: AnyView
+    dynamic var coordinate: CLLocationCoordinate2D
```
pixlwave: Why does this one need to be dynamic? Would be useful to add a small comment to explain it.

ElementX/Sources/Screens/LocationSharing/View/LocationSharingScreen.swift
```diff
@@ -127,9 +121,9 @@ struct LocationSharingScreen: View {
     }
     
     @ViewBuilder
-    private var shareSheet: some View {
-        let location = context.viewState.initialMapCenter
-        let senderName = context.viewState.locationMarkerKind.displayName ?? context.viewState.locationMarkerKind.userProfile?.userID
+    private func shareSheet(for annotation: LocationAnnotation) -> some View {
```
pixlwave: Minor, but might be nicer to extract this one out to be its own LocationSharingShareSheetView now.

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
```diff
@@ -38,27 +39,59 @@ struct LocationSharingScreenViewState: BindableState {
...
 
     let interactionMode: LocationSharingInteractionMode
     let mapURLBuilder: MapTilerURLBuilderProtocol
     let showLiveLocationSharingButton: Bool
     let ownUserID: String
-    var userProfile: UserProfileProxy
+    var userProfiles: [String: UserProfileProxy]
+    var liveLocationShares: [LiveLocationShare] = []
+    
+    var annotations: [LocationAnnotation] {
```
pixlwave, line 69: I have to be honest, I find this weird. I realise we're updating profiles, but I think it would be nicer to remove the `userProfiles` dictionary here and instead combine the profiles into the interaction mode's associated value (which I presume would need a new struct) and then just do all the computation up front in the view model. Do you think that would work/make sense?

pixlwave (approved):
Oh nice, this is exciting! 📍

ElementX/Sources/Screens/PinnedEventsTimelineScreen/PinnedEventsTimelineScreenCoordinator.swift
Velin92: I don't know, I think making them separates makes easy to understand what are the entry points that support which interaction mode (usually some of them support all of three, others only static and live, and others only the picker one)

ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
Velin92: oh nono, that's deliberate... before we were only checking for EventBased**Message** items, now we need to check for EventBased ones (check few lines above), so this means we need to do more casting sadly.

ElementX/Sources/Other/MapLibre/LocationAnnotation.swift
Velin92: oh yeah that's not required anymore, sorry gotta remove this.

ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
pixlwave: Ahhhh, right yeah I see that now 👍

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
Velin92: why? I think the profiles are kind of a separate thing from the state of the interaction, and we only care about the initial state when passing around the interaction.

Then what can really change are the `profiles` and the `liveLocationShares` and any changes on them should recompute the annotations, since now their positions or profile content might get updated.


I guess if we really want to go this way at most we could have the liveLocationShares be an associated value of the .viewLive case (which I guess can make sense at that point).
But yeah not sure if that's worth it.

PR #5399, opened 2026-04-15
Support for ongoing voice vs video call + joining

### LINKED ISSUE #5400: Update Active call indicators to reflect if it is a Voice or a Video Call (parity with web and android)
When there is an ongoing call, there are several places in the UX where an active call indicator is shown.
It is currently using a camera icon, but since https://github.com/element-hq/element-x-android/pull/5995 we have information of if the call is audio or video.
We have to update the indicator accordingly with a new variant for audio.
Joining a voice call should start EC without video.

Image

Image

BillCarsonFr:
### Pull Request Checklist

Fixes https://github.com/element-hq/element-x-ios/issues/5400

EXI counter part of this android PR  https://github.com/element-hq/element-x-android/pull/6574
It is now possible to know if an ongoing call is voice or video (https://github.com/matrix-org/matrix-rust-sdk/pull/6274).
This PR use this information to update the active calls indicators in the app

image

- [ ] Voiceover enabled.


ElementX/Sources/Other/SwiftUI/Views/JoinCallButton.swift
```diff
@@ -10,8 +10,17 @@ import Compound
...
+    let isVoiceCall: Bool
     let action: () -> Void
     
+    var icon: KeyPath<CompoundIcons, Image> {
+        if isVoiceCall {
+            return \.voiceCallSolid
+        } else {
+            return \.videoCallSolid
+        }
+    }
```
pixlwave: This one is simple enough that we can just inline it below as

```
CompoundIcon(isVoiceCall ? \.voiceCallSolid : \.videoCallSolid)
```

ElementX/Sources/Other/SwiftUI/Views/JoinCallButton.swift
```diff
@@ -66,9 +75,17 @@ struct JoinCallButton_Previews: PreviewProvider {
...
-        }
+        }.previewDisplayName("Join Video Call")
+        ElementNavigationStack {
+            Color.clear
+                .toolbar {
+                    ToolbarItem(placement: .confirmationAction) {
+                        JoinCallButton(isVoiceCall: true) { }
+                    }
+                }
+        }.previewDisplayName("Join Audio Call")
```
pixlwave:
```suggestion
        }
        .previewDisplayName("Join Video Call")
        
        ElementNavigationStack {
            Color.clear
                .toolbar {
                    ToolbarItem(placement: .confirmationAction) {
                        JoinCallButton(isVoiceCall: true) { }
                    }
                }
        }
        .previewDisplayName("Join Audio Call")
```

BillCarsonFr:
@pixlwave I don't understand what is wrong with the screenshots preview. If I ran `roomMemberDetailsScreen` locally (screens that I have not touched in this PR), then the snapshot are exactly the same.

It is hard for me to ran the preview test locally because Xcode complains about some file access problem (all looks fine if I look at the permissions). So I always have to delete the snapshot before beeing able to run the tests.

I deleted all in order to re-record all in case that CI was reporting the wrong file? 
But then git is seeing 128 changed screenshots (that looks very similar to me)

ElementX/Sources/Services/Room/RoomInfoProxyProtocol.swift
```diff
@@ -51,6 +51,7 @@ protocol RoomInfoProxyProtocol: BaseRoomInfoProxyProtocol {
     var notificationCount: Int { get }
     var cachedUserDefinedNotificationMode: RoomNotificationMode? { get }
     var hasRoomCall: Bool { get }
+    var activeRoomCallIntent: RtcCallIntent? { get }
```
pixlwave: Could we make our own Swift `CallIntent` enum type so that we don't need to `import MatrixRustSDK` in views like `RoomCallControlsToolbar.swift`? It can just be a really simple mapping like e.g. `TraceLogPack` does.

ElementX/Sources/Services/Room/RoomSummary/RoomSummary.swift
```diff
@@ -57,6 +57,7 @@ struct RoomSummary {
     let alternativeAliases: Set<String>
     
     let hasOngoingCall: Bool
+    let activeCallIntent: RtcCallIntent?
```
pixlwave: Same comment here to use the Swift type.

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
```diff
@@ -190,7 +190,8 @@ struct HomeScreenRoom: Identifiable, Equatable {
         let isDotShown: Bool
         let isMentionShown: Bool
         let isMuteShown: Bool
-        let isCallShown: Bool
+        let isVoiceCallShown: Bool
+        let isVideoCallShown: Bool
```
pixlwave: This feels like it should be an enum rather than 2 bools as I don't think a room can have both types going on?

pixlwave (commented):
Urgh, I wrote this and then didn't hit submit 🤦‍♂️

ElementX/Sources/Screens/HomeScreen/HomeScreenModels.swift
BillCarsonFr: done

ElementX/Sources/Services/Room/RoomInfoProxyProtocol.swift
BillCarsonFr: done

pixlwave (approved):
LGTM, thanks Valere.

PR #5402, opened 2026-04-15
Fix the space filters button on iPad.

pixlwave:
There's currently a bug on iPad when a space filter is selected where the background isn't visible (and the icon is white so it can't be seen). This PR switches the `SpaceFiltersButton` to use the default `Toggle` instead of the custom `Button` and let iOS do the styling automatically:

| Selected Before | Select After |
| - | - |
| Simulator Screenshot - iPad Pro 11-inch (M5) - 2026-04-15 at 20 14 59 | Simulator Screenshot - iPad Pro 11-inch (M5) - 2026-04-15 at 20 08 06 |

Note that a side-effect of this is that we have a slightly different style when selected on iPhone:

| Before | After |
| - | - |
| Simulator Screenshot - iPhone 17 - 2026-04-15 at 20 13 51 | Simulator Screenshot - iPhone 17 - 2026-04-15 at 20 11 29 |

However this means that the style now matches e.g. Messages and Mail:

Screenshot 2026-04-15 at 8 32 22 pm

@americanrefugee Are you happy with this change of style when the button is using Liquid Glass on iPhone?

### DIFF
--- a/ElementX/Sources/Application/Navigation/NavigationCoordinators.swift
+++ b/ElementX/Sources/Application/Navigation/NavigationCoordinators.swift
@@ -326,6 +326,11 @@ import SwiftUI
     }
 }
 
+extension EnvironmentValues {
+    /// Whether or not the current view is part of the sidebar module of a `NavigationSplitCoordinator`.
+    @Entry var isInSidebar = false
+}
+
 private struct NavigationSplitCoordinatorView: View {
     @State private var columnVisibility = NavigationSplitViewVisibility.all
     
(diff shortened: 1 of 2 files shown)


Velin92 (approved):

pixlwave:
Ok I looked a bit further and added an environment value to determine whether or not the view is in the sidebar or not so it can fallback to the iOS 18 rendering in that instance. It looks like this now:

Simulator Screenshot - iPad Pro 11-inch (M5) - 2026-04-17 at 14 33 39

Velin92 (approved):

PR #5403, opened 2026-04-15
Use a VStack for invite buttons when the HStack doesn't fit.

pixlwave:
The Accept/Decline labels were being wrapped onto 2 lines. This fixes both iPhone with a large Dynamic Type size, but also iPad/Mac when the sidebar is narrow enough.


ElementX/Sources/Screens/HomeScreen/View/HomeScreenInviteCell.swift
```diff
@@ -74,12 +74,14 @@ struct HomeScreenInviteCell: View {
             .fixedSize(horizontal: false, vertical: true)
             .accessibilityElement(children: .combine)
             
-            buttons
-                .padding(.top, 14)
-                .padding(.trailing, 22)
+            ViewThatFits {
+                HStack(spacing: 12) { buttons }
+                VStack(spacing: 16) { Group(subviews: buttons) { ForEach($0.reversed()) { $0 } } }
```
stefanceriu, line 79: First time I'm seeing SectionCollection be used, cool stuff, but I need to ask: why? If it's on the right horizontally why not be at the bottom vertically? 🤔

pixlwave: The Figma is [here](https://www.figma.com/design/n7t0R3BvpCUeKUM4lefa43/Element-X-on-iPad?node-id=9-11256&m=dev). It's similar to `Cancel` being on the left of an Alert with 2 buttons, and the bottom of an Alert with 3 I guess.

stefanceriu: Hmm, I see, fair enough!

stefanceriu (approved):

PR #5407, opened 2026-04-16
Remove DTCoreText/Foundation references and warning fixes

stefanceriu:


ElementX/Sources/FlowCoordinators/SpaceFlowCoordinator.swift
pixlwave, line 1: I think we're meant to be passing these along instead of those hard-coded `isVoiceCall: false`s so the call screen knows the expected intent.

ElementX/Sources/FlowCoordinators/SpaceSettingsFlowCoordinator.swift
```diff
@@ -395,7 +395,7 @@ final class SpaceSettingsFlowCoordinator: FlowCoordinatorProtocol {
             switch action {
             case .finished:
                 stateMachine.tryEvent(.stopMembersListFlow)
-            case .presentCallScreen(let roomProxy, let isVoiceCall):
+            case .presentCallScreen(let roomProxy, _):
                 actionsSubject.send(.presentCallScreen(roomProxy: roomProxy))
```
pixlwave, line 399: And similar, ideally this should have an `isVoiceCall` added to this action's enum.

ElementX/Sources/Other/HTMLParsing/UIFont+AttributedStringBuilder.m
pixlwave, line 1: Given this was a swizzle, have you tested the RTE was relying on the app doing this?

ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
pixlwave, line 1: Errrr, I think this is a genuine error that we should be fixing rather than ignoring no?

pixlwave (approved):
Nice tidy up 🧹

ElementX/Sources/Other/HTMLParsing/UIFont+AttributedStringBuilder.m
stefanceriu: Well it never had anything to do with the rich text editor and if it does it definitely should be part of the ex repo https://github.com/element-hq/element-x-ios/commit/cf42d61d279ea4311331414a32a9388db7c95328

ElementX/Sources/FlowCoordinators/SpaceSettingsFlowCoordinator.swift
stefanceriu: I imagine @BillCarsonFr is handling these somewhere else? 🤔

ElementX/Sources/Services/ElementCall/ElementCallWidgetDriver.swift
stefanceriu: It's an SDK type, it's never going to be sendable, right?

pixlwave: My understanding is that the warning is being raised because it's marked as Sendable:

Screenshot 2026-04-16 at 1 23 16 pm

stefanceriu: Oooh I wasn't aware uniffi does that, I completely misread what the problem is!

PR #5416, opened 2026-04-16
Add a workflow that runs on top of Renovate XcodeGen PRs to update the project and resolved packages.

stefanceriu:


Velin92 (approved):
LGTM, but I think we should double check those security issues that the bot flagged, maybe we could ask to the security team?

.github/workflows/renovate-xcodegen.yml
```diff
@@ -0,0 +1,52 @@
+name: Update project following Renovate-XcodeGen update
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    paths:
+      - project.yml
+
+permissions:
+  contents: write
```
pixlwave: This should be `permissions: {}` and then add `write` only to the step that needs it I believe.

.github/workflows/renovate-xcodegen.yml
```diff
@@ -0,0 +1,52 @@
...
+      group: renovate-xcodegen-${{ github.head_ref }}
+      cancel-in-progress: true
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          ref: ${{ github.head_ref }}
+          token: ${{ secrets.ELEMENT_BOT_TOKEN }}
+          persist-credentials: false
```
pixlwave:
```suggestion
          # we need to push any changes to the project.
          persist-credentials: true
```

pixlwave (approved):

renovate.json
```diff
@@ -8,24 +8,34 @@
...
     {
       "managers": ["swift", "cocoapods"],
       "matchPaths": ["compound-ios/Package.swift"],
       "groupName": "Compound"
     },
     {
-      "matchManagers": ["bundler"],
-      "groupName" : "Ruby"
-    }
+      // PostHog iOS uses continuous integration so limit upgrades to one PR per month
```
pixlwave, line 31: Waaaait This is red… I don't think comments are allowed in json.

pixlwave: Yeah, Android is using json5

https://github.com/element-hq/element-x-android/blob/develop/.github/renovate.json5

stefanceriu: Renovate supports JSONC for .json files and any config files without file extension (e.g. .renovaterc). We also recommend you prefer using JSONC within a .json file to using a .json5 file if you want to add comments. - https://docs.renovatebot.com/configuration-options/

PR #5420, opened 2026-04-16
Live Location Sharing Sheet implementation

### LINKED ISSUE #5199: [Task] EX LLS: View all LLS sessions on the map

### LINKED ISSUE #5200: [Task] EX LLS: Stop LLS

Velin92:
Added a sheet in live location mode, to track all the users that are sharing their location.
You can also tap on a user to higlight (go to) their location, or share through the OS share options their location in one of the map apps.
You can also stop your own live location if you are currently sharing it through such sheet.
The users are ordered from newer share to older share, but own user is always at the top.

Added also some preview tests.

The video is taken on the simulator (that's why the own user location is always in a loading state)

https://github.com/user-attachments/assets/6431c256-5905-4587-b902-b4989dcd00fb

fixes #5199 
fixes #5200


ElementX/Sources/Other/MapLibre/MapLibreMapView.swift
```diff
@@ -81,6 +88,11 @@ struct MapLibreMapView: UIViewRepresentable {
         showUserLocation(in: mapView)
     }
     
+    private func coordinatesAreEqual(_ lhs: CLLocationCoordinate2D, _ rhs: CLLocationCoordinate2D?) -> Bool {
+        guard let rhs else { return false }
+        return lhs.latitude == rhs.latitude && lhs.longitude == rhs.longitude
+    }
+
```
pixlwave: Would it make sense to use a retroactive conformance here instead of this method?

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenViewModel.swift
```diff
@@ -83,11 +83,21 @@ class LocationSharingScreenViewModel: LocationSharingScreenViewModelType, Locati
...
     }
     
     // MARK: - Private
     
+    private func stopLiveLocation() {
+        state.isStoppingLiveLocation = true
+        if let index = state.liveLocationShares.firstIndex(where: { $0.userID == roomProxy.ownUserID }) {
+            state.liveLocationShares.remove(at: index)
+        }
+        Task { await liveLocationManager.stopLiveLocation(roomID: roomProxy.id) }
```
pixlwave, line 98: Do we not need to set isStoppingLiveLocation back to `false` once it has stopped?

pixlwave (approved):

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenViewModel.swift
Velin92: nah, it's okay is a best effort thing since even if the stop has completed the remote echo of the server is not going to make the liveLocationShare disappear immediately (so it may pop back up for some seconds before being gone on the server too), and in theory, you can't start a new session from the same device until you close and reopen the screen.

pixlwave: Got it, probably worth adding a comment about that last bit.

PR #5439, opened 2026-04-17
Add a new workflow that runs the preview tests and re-records the snapshots

stefanceriu:
..  for PRs that are labeled with `record-snapshots`

### DIFF
--- a/.github/workflows/record_snapshots.yml
+++ b/.github/workflows/record_snapshots.yml
@@ -1,0 +1,63 @@
+name: Record Snapshots
+
+on:
+  # Triggered when a maintainer adds the "record-snapshots" label to a PR.
+  pull_request_target: # zizmor: ignore[dangerous-triggers]
+    types: [labeled]
+
+permissions: {}
+
+jobs:
+  record-snapshots:
+    name: Record Preview Snapshots
+    runs-on: macos-26
+    timeout-minutes: 45
+
+    if: >
+      contains(github.event.issue.labels.*.name, 'record-snapshots') &&
+      github.event.pull_request.head.repo.full_name == github.repository 
+
+    permissions:
+      contents: write
+      pull-requests: write
+
+    concurrency:
+      group: record-snapshots-${{ github.event.pull_request.number }}
+      cancel-in-progress: true
+
+    steps:
+      - name: Remove label
+        run: gh pr edit "${{ github.event.pull_request.number }}" --remove-label 'record-snapshots'
+        env:
+          GH_TOKEN: ${{ github.token }}
+
+      - name: Checkout PR
+        uses: nschloe/action-cached-lfs-checkout@385a8ecc719e50b8c71af6ab01a624b486b7c3bc # v1.2.5
+        with:
+          # Check out the PR head commit (not the base branch).
+          ref: ${{ github.event.pull_request.head.sha }}
+          persist-credentials: true
+
+      - name: Setup environment
+        run: source ci_scripts/ci_common.sh && setup_github_actions_environment
+
+      - name: Run preview tests in recording mode
+        run: swift run -q tools ci preview-tests --record
+
+      - name: Commit and push updated snapshots
+        run: |
+          git config user.name "ElementRobot"
+          git config user.email "releases@riot.im"
+
+          # Stage only snapshot changes.
+          git add PreviewTests/Sources/__Snapshots__/
+
+          if git diff --cached --quiet; then
+            echo "✅ No snapshot changes detected."
+          else
+            git commit -m "Record preview snapshots"
+            git push origin HEAD:refs/heads/${GITHUB_EVENT_PULL_REQUEST_HEAD_REF}
+            echo "✅ Snapshots recorded and pushed."
+          fi
+        env:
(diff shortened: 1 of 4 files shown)


.github/workflows/record_snapshots.yml
```diff
@@ -0,0 +1,74 @@
...
+          GITHUB_EVENT_PULL_REQUEST_HEAD_REF: ${{ github.event.pull_request.head.ref }}
+
+      - name: Upload snapshots artifact
+        # Always upload as a fallback (especially useful for fork PRs).
+        if: always()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: PreviewSnapshots
+          path: PreviewTests/Sources/__Snapshots__/
+          retention-days: 7
```
pixlwave: If we keep it, I'd be tempted to set this to just 1 day personally, as I doubt we need a whole weeks worth of snapshots.

.github/workflows/record_snapshots.yml
```diff
@@ -0,0 +1,74 @@
+name: Record Snapshots
+
+on:
+  # Triggered when a maintainer adds the 'record-snapshots' label to a PR.
+  # Uses pull_request_target so the workflow has write permissions to push
+  # snapshot commits back to same-repo PR branches.
+  # Safety: only maintainers can add labels, so checked-out PR code is trusted.
+  pull_request_target: # zizmor: ignore[dangerous-triggers]
+    types: [labeled]
```
pixlwave: If I'm honest, I don't think I like this justification, especially looking at e.g. https://github.com/search?q=org%3Aelement-hq%20zizmor%3A%20ignore%5Bdangerous-triggers%5D&type=code

I think we should gate the entire job to only work against our repo, at least for this first version.

.github/workflows/record_snapshots.yml
```diff
@@ -0,0 +1,74 @@
...
+
+permissions: {}
+
+jobs:
+  record-snapshots:
+    name: Record Preview Snapshots
+    if: github.event.label.name == 'record-snapshots'
```
pixlwave: So something like this on the job itself.

```suggestion
    if: >
      github.event.label.name == 'record-snapshots' &&
      github.event.pull_request.head.repo.full_name == github.repository # 🚨 Never run external code in a pull_request_target
```

.github/workflows/record_snapshots.yml
```diff
@@ -0,0 +1,74 @@
...
+
+      - name: Setup environment
+        run: source ci_scripts/ci_common.sh && setup_github_actions_environment
+
+      - name: Run preview tests in recording mode
+        run: swift run -q tools ci preview-tests --record
+
+      - name: Commit and push updated snapshots
+        # Only push for same-repo PRs; fork PRs lack push permissions.
+        if: github.event.pull_request.head.repo.full_name == github.repository
```
pixlwave: And then remove this and probably the artefacts too.

Tools/Sources/Commands/CI/PreviewTests.swift
```diff
@@ -0,0 +1,90 @@
...
+                "--create-simulator-type", Self.simulatorType
+            ]).run()
+        } catch {
+            if record {
+                // In recording mode, test failures are expected — swift-snapshot-testing marks
+                // recording runs as failed. Check whether the xcresult bundle was created to
+                // distinguish genuine failures (compilation error, simulator issue) from the
+                // expected snapshot-recording "failures".
+                let resultBundleURL = URL.projectDirectory
+                    .appendingPathComponent("\(CI.testOutputDirectory)/\(Self.scheme).xcresult")
```
pixlwave:
```suggestion
                    .appending(path: "\(CI.testOutputDirectory)/\(Self.scheme).xcresult")
```

Tools/Sources/Commands/CI/PreviewTests.swift
```diff
@@ -0,0 +1,90 @@
...
+        guard var plan = try JSONSerialization.jsonObject(with: data) as? [String: Any],
+              var defaultOptions = plan["defaultOptions"] as? [String: Any],
+              var envVars = defaultOptions["environmentVariableEntries"] as? [[String: Any]] else {
+            throw ValidationError("Could not parse test plan at \(Self.testPlanPath).")
+        }
+
+        for i in envVars.indices where envVars[i]["key"] as? String == "RECORD_FAILURES" {
+            envVars[i]["enabled"] = enabled
+            break
+        }
```
pixlwave: It's lovely when the AI just ignores things in the AGENTS.md file…

```suggestion
        for index in envVars.indices where envVars[index]["key"] as? String == "RECORD_FAILURES" {
            envVars[index]["enabled"] = enabled
            break
        }
```

pixlwave (approved):
Very cool, can't wait to try this one out 😎

.github/workflows/record_snapshots.yml
stefanceriu: Done 👍

Tools/Sources/Commands/CI/PreviewTests.swift
stefanceriu: We'll get there .. eventually 😅

PR #5441, opened 2026-04-20
Add xcresult-analysis Copilot skill

manuroe:
Adds a new `xcresult-analysis` Copilot CLI skill for analysing Xcode xcresult bundles to diagnose failing tests.
The skill accepts a local `.xcresult` path or a GitHub Actions run URL, and surfaces failing test details, activity logs, and diffs.

I used to improve integration tests in https://github.com/element-hq/element-x-ios/pull/5352. I successful tested it locally on unit tests by breaking them first. It was able to fix them.

⚠️ This skill has to be used with copilot. Apparently, it burns tokens too quickly with an Anthropic plan.

### DIFF
--- a/.github/skills/xcresult-analysis/.gitignore
+++ b/.github/skills/xcresult-analysis/.gitignore
@@ -1,0 +1,2 @@
+work/
+scripts/analyze-xcresult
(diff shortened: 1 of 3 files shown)


Velin92:
This is quite cool! I think we should also have a file for humans that explains how to use the skill (and all the skills we are going to build for the future). Could we make something like that?

stefanceriu:
I believe skills like these should live in a separate repository at least until they’re proven to work reliably,  especially as it contains a lot of swift code that would need to be managed, tested etc. I would also help with separating concerns and making sure skills rely on each other and other definitions to do the right thing. 

For example 

```
#### Snapshot & Preview tests (`PreviewTests.xcresult`)

- `Snapshot does not match reference` — a UI change caused a visual diff. Re-record snapshots on the correct device/OS: `swift run tools ci unit-tests` or open the test in Xcode and enable record mode.
```

this is entirely wrong, it will most certainly not record anythign new and will just fail again.

Iterating on problems like these within the EX repo will only bring noise and is better done separately, either open or in private.

Also worth noting, installing skills for Copiliot is now easier than ever with `gh skill install repo` - https://github.blog/changelog/2026-04-16-manage-agent-skills-with-github-cli/

PR #5446, opened 2026-04-20
Fix session verification SAS flow not automatically starting

stefanceriu:
With #5116 removing the `verificationRequestAccepted` view state and its `.startSasVerification` button action we now need another way of starting the SAS flows whenever the verification requests gets accepted. 
As such a state machine transition from `acceptingVerificationRequest` to `verificationRequestAccepted` will now automatically call `startSasVerification` on the SessionVerificationProxy.


Hywan:
Ship it 🚢!

Velin92 (approved):

PR #5448, opened 2026-04-20
Update InviteUsersConfirmationSheet strings 

Velin92:
name says it all


ElementX/Resources/Localizations/en.lproj/Localizable.strings
```diff
@@ -552,6 +552,10 @@
 "screen_advanced_settings_hide_invite_avatars_toggle_title" = "Hide avatars in room invite requests";
 "screen_advanced_settings_hide_timeline_media_toggle_title" = "Hide media previews in timeline";
 "screen_advanced_settings_labs" = "Labs";
+"screen_advanced_settings_live_location_section_description" = "The distance you have to move to trigger an update.";
+"screen_advanced_settings_live_location_section_footer" = "Make sure \"Precise Location” is enabled for this app. To change the permission go to %1$@.";
```
pixlwave, line 556: Unrelated but looks like this should be
```suggestion
"screen_advanced_settings_live_location_section_footer" = "Make sure “Precise Location” is enabled for this app. To change the permission go to %1$@.";
```

pixlwave (approved):
LGTM 🙏

PR #5449, opened 2026-04-20
Live Location Sharing update filtering based on time and distance

### LINKED ISSUE #5198: [Task] EX LLS: LLS update threshold

Velin92:
- Updates will always be throttled by 3 seconds
- Updates are also filtered by distanced specified by the users, which also dynamically adapt accuracy to save battery
- Use the full CLLocationManager delegate based approach instead of CLLocation.liveUpdates() for live updates since:
  - liveUpdates() was unreliable in background mode
  - liveUpdates() doesn't support accuracy and distance filter since it used a different CLLocationManager instance under the hood, that doesn't allow for the fine tuned implementation we want, which makes it good for simple apps that just need to track the location for brief periods of time, but not for our use case.

fixes #5198 
Screenshot 2026-04-21 at 12 02 06

### DIFF
--- a/ElementX/Sources/Screens/Settings/AdvancedSettingsScreen/View/AdvancedSettingsScreen.swift
+++ b/ElementX/Sources/Screens/Settings/AdvancedSettingsScreen/View/AdvancedSettingsScreen.swift
@@ -37,6 +37,7 @@ struct AdvancedSettingsScreen: View {
             
             moderationAndSafetySection
             timelineMediaSection
+            liveLocationSection
         }
         .compoundList()
         .navigationTitle(L10n.commonAdvancedSettings)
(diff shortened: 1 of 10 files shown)


ElementX/Sources/Services/Location/LiveLocationManager.swift
```diff
@@ -145,7 +167,15 @@ class LiveLocationManager: NSObject, LiveLocationManagerProtocol, CLLocationMana
...
                 appSettings.liveLocationSharingTimeoutDatesByRoomID.removeAll()
                 activeRoomProxies.removeAll()
-                locationUpdatesTask = nil
+                self.stopUpdatingLocation()
+            }
+            .store(in: &cancellables)
+        
+        appSettings.$liveLocationMinimumDistanceUpdate
+            .removeDuplicates()
+            .debounce(for: .seconds(3), scheduler: DispatchQueue.main)
```
stefanceriu: Why are you debouncing the app setting? 🤔

ElementX/Sources/Services/Location/LiveLocationManager.swift
```diff
@@ -158,26 +188,40 @@ class LiveLocationManager: NSObject, LiveLocationManagerProtocol, CLLocationMana
...
-            }
+    /// Sets up the distance filter and the most optimal accuracy given the minimum distance to save battery.
+    private func setupMinimumDistance(_ minimumDistance: Int) {
+        switch minimumDistance {
+        case 0..<10:
+            locationManager.desiredAccuracy = kCLLocationAccuracyBest
+        case 10..<100:
+            locationManager.desiredAccuracy = kCLLocationAccuracyNearestTenMeters
+        default:
+            locationManager.desiredAccuracy = kCLLocationAccuracyHundredMeters
```
stefanceriu, line 199: Huh, you would think this would be automatic on apple's side..

ElementX/Sources/Services/Location/LiveLocationManager.swift
Velin92: Because this value it's set through a slider (just pushed new designs), So I just don't want the value to be costantly get updated, even while the user is sliding it.

ElementX/Sources/Screens/Settings/AdvancedSettingsScreen/View/AdvancedSettingsScreen.swift
```diff
@@ -84,6 +85,30 @@ struct AdvancedSettingsScreen: View {
                 .compoundListSectionFooter()
         }
     }
+    
+    private var liveLocationSection: some View {
+        Section {
+            ListRow(kind: .custom {
+                Stepper(L10n.screenAdvancedSettingsLiveLocationUpdateDistance(context.liveLocationMinimumDistanceUpdate),
```
stefanceriu, line 92: I'm honestly not a fan of this. I'm sure people don't care about precision to the meter, we should just give them the 3 options.

Velin92: We changed this to a slider now (design didn't like the stepper, but product still wanted a fine tuned value)

ElementX/Sources/Services/Location/LiveLocationManager.swift
Velin92: Wait to see my PR about managing LLS with precision location turned off... is the next one... you even need to reupdate this after re-enabling permissions

stefanceriu (approved):
lgtm 👍

ElementX/Sources/Services/Location/LiveLocationManager.swift
stefanceriu: Ah, now I get it, it was a stepper when I looked.

PR #5454, opened 2026-04-21
Throttle capabilities updates to once per day (or when the app is next launched).

pixlwave:
Fixes a bug introduced by #5421 where `/capabilities` was being called far too much.

The `.running` state is deceiving. It is sent as part of every single iteration of the sync loop 🤦‍♂️


stefanceriu (approved):
Noice, ship it!

PR #5455, opened 2026-04-21
Handle reduced accuracy authorization case

Velin92:
Element Classic still sent live location updates even if the user turned off "Precise Location" permissions. So we should definitely support this too on EXI.

Essentially instead of using the `start/stopLocationUpdates()` (which do not work with reduced accuracy) functions we use `start/stopMonitoringSignificantLocationChanges()` which work also in reduced accuracy mode (but update on very long intervals of time and space, about 500m).

It's important to note that the distance filter and the desiredAccuracy do not matter in this mode, because under the hood they get set to specific values, however we need to restore them once we transition back to fullAccuracy mode.

Also added some tests to verify that the appropriate start/stop functions are called according to the right accuracy mode.


pixlwave (approved):
Nice, LGTM 👍

PR #5458, opened 2026-04-21
Update the Deactivate Account strings (and icon) to call it Delete Account.

### LINKED ISSUE #5432: Global settings > Update menu list items
### Description

- Change "Deactivate account" to "Delete account"
- Change icon to "Delete"/trashcan

Image

### Why do I want to do this?

This is to match what we're changing on iOS and what already exists on MAS.

pixlwave:
The strings were pulled in by #5452 and so the snapshots have already been updated. But there was an icon to update  to do (not in any snapshots) and we the `action.` strings have been renamed to follow the naming guidelines.

Closes #5432


stefanceriu (approved):
👍

PR #5459, opened 2026-04-21
Fix a potential crash caused by using old Sentry transactions.

pixlwave:
The trigger looked like this:

- You login and we start syncing and create the transaction to track the sync length.
- You see the Analytics prompt which triggers [setupSentry](https://github.com/element-hq/element-x-ios/blob/4e1c855467c2d4ad80f14b9b24dd0bcddb128399/ElementX/Sources/Application/AppCoordinator.swift#L979) which calls `SentrySDK.start`. Presumably this clears any in-flight transactions on their side.
- Eventually syncing is done and we finish the transaction.
- 💥

At least for now we can fix this by resetting the transactions any time the user opts in/out of analytics.

Also fixes a deprecation since upgrading the Sentry dependency, along with removing an unused method.

### DIFF
--- a/ElementX/Sources/Services/Analytics/Signposter.swift
+++ b/ElementX/Sources/Services/Analytics/Signposter.swift
@@ -75,6 +75,10 @@ class Signposter {
         transactions[transactionName] = nil
     }
     
+    func resetTransactions() {
+        transactions.removeAll()
+    }
+    
     // MARK: - Spans
     
     func addSpan(_ spanName: SpanName, toTransaction transactionName: TransactionName) -> Span? {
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Services/Analytics/Signposter.swift
```diff
@@ -75,6 +75,10 @@ class Signposter {
         transactions[transactionName] = nil
     }
     
+    func resetTransactions() {
```
stefanceriu, line 78: Now thinking about it.. perhaps we should do it the other way and ignore `startTransaction` requests if sentry is not enabled. Perhaps even crash in debug mode?

pixlwave: The reason I went this way was that we call `setupSentry` whenever the user opts in/out and whenever they sign in/out. So fundamentally just because you were opted in when calling `startTransaction`, doesn't necessarily mean you will be when calling `finishTransaction`

Not saying that we couldn't change it, but this seemed the safest way of thinking about it to me.

stefanceriu: Oh I see, yeah that makes sense 👍

stefanceriu (approved):

Velin92 (approved):

PR #5461, opened 2026-04-21
Improve Live Location Sharing with reduced accuracy mode

Velin92:
Reminder to myself to not fully trust the legacy app old code 😅

`startMonitoringSignificantgUpdates` DOES NOT work in background mode...
But most importantly `startUpdatingLocation` does indeed work fine also in reduced accuracy mode (granted you change the desiredAccuracy and not allow for pausing updates), and will work also in background with such mode on.
The only thing we need to do to make sure is to change the desiredAccuracy according to the accuract authorization status, and that's all.

Also updated the tests.

### DIFF
--- a/ElementX/Sources/Services/Location/LiveLocationManager.swift
+++ b/ElementX/Sources/Services/Location/LiveLocationManager.swift
@@ -9,12 +9,6 @@ import Combine
 import CoreLocation
 
 class LiveLocationManager: NSObject, LiveLocationManagerProtocol, CLLocationManagerDelegate {
-    enum LiveState {
-        case full
-        case limited
-        case off
-    }
-    
     private let clientProxy: ClientProxyProtocol
     private let locationManager: CLLocationManagerProtocol
     private let appSettings: AppSettings
@@ -32,7 +26,7 @@ class LiveLocationManager: NSObject, LiveLocationManagerProtocol, CLLocationMana
     
     private var cancellables = Set<AnyCancellable>()
     
-    private var liveState = LiveState.off
+    private var isUpdatingLocation = false
     
     @MainActor
     init(clientProxy: ClientProxyProtocol,
@@ -52,7 +46,9 @@ class LiveLocationManager: NSObject, LiveLocationManagerProtocol, CLLocationMana
         self.locationManager.delegate = self
         self.locationManager.allowsBackgroundLocationUpdates = true
         self.locationManager.showsBackgroundLocationIndicator = true
-        setupMinimumDistance(appSettings.liveLocationMinimumDistanceUpdate)
+        // Since unpausing location updates is not trivial, let's not pause them.
+        self.locationManager.pausesLocationUpdatesAutomatically = false
+        setupMinimumDistanceUpdatesAndAccuracy(minimumDistance: appSettings.liveLocationMinimumDistanceUpdate)
         setupSubscriptions()
     }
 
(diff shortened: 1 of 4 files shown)


ElementX/Sources/Services/Location/LiveLocationManager.swift
```diff
@@ -52,7 +46,9 @@ class LiveLocationManager: NSObject, LiveLocationManagerProtocol, CLLocationMana
         self.locationManager.delegate = self
         self.locationManager.allowsBackgroundLocationUpdates = true
         self.locationManager.showsBackgroundLocationIndicator = true
-        setupMinimumDistance(appSettings.liveLocationMinimumDistanceUpdate)
+        // Since unpausing location updates is not trivial, let's not pause them.
```
stefanceriu: I think this deserves a bit more explanation

ElementX/Sources/Services/Location/LiveLocationManager.swift
```diff
@@ -209,42 +195,36 @@ class LiveLocationManager: NSObject, LiveLocationManagerProtocol, CLLocationMana
...
-            liveState = .full
-            locationManager.startUpdatingLocation()
-        } else {
-            MXLog.info("Starting live location updates with significant changes (reduced accuracy)")
-            liveState = .limited
-            locationManager.startMonitoringSignificantLocationChanges()
-        }
+        MXLog.info("Starting live location updates")
+        isUpdatingLocation = true
+        locationManager.startUpdatingLocation()
```
stefanceriu, line 223: Wouldn't it be safer to hook the `startUpdatingLocation` and `stopUpdatingLocation` to an `isUpdatingLocation` setter?

ElementX/Sources/Services/Location/LiveLocationManager.swift
```diff
@@ -266,7 +246,7 @@ class LiveLocationManager: NSObject, LiveLocationManagerProtocol, CLLocationMana
             
             switch await roomProxy.sendLiveLocation(geoURI: geoURI) {
             case .success:
-                MXLog.debug("Sent live location to room: \(roomID)")
+                MXLog.info("Sent live location to room: \(roomID)")
```
stefanceriu: Oh I don't know about this, info is the default log level for all installs. Even if we're only sending one every 3 seconds it might still result in quite a lot of lines, especially if you decided to share over hours. Is it really useful?

Velin92: nah it's okay, I'll set it back to Debug

ElementX/Sources/Services/Location/LiveLocationManager.swift
Velin92: I'll ad a link to this in the comment: https://developer.apple.com/documentation/corelocation/cllocationmanager/pauseslocationupdatesautomatically

because even the explaination is quite long, but essentially when pausing happens the app would be responsible of restoring the session, and you would need to schedule a notification with  [UNLocationNotificationTrigger](https://developer.apple.com/documentation/UserNotifications/UNLocationNotificationTrigger) when the user has left its current region (which I assume is the current zone where everything has paused), and then wake up the app, and restore the location updates. As I said, not trivial at all, and maybe completely out of scope for our use case. In theory the distance filter is already enough, if the user doesn't move within those boundaries didUpdateLocation will not be called so we are still saving battery and processing

ElementX/Sources/Services/Location/LiveLocationManager.swift
Velin92: I mean is kinda the same thing, in the end we either simply call the start and stop functions or update the internal value with true or false and use a did set would be the same. As you prefer, however my personal preference to use a clear start and stop function.

stefanceriu (approved):

PR #5466, opened 2026-04-22
Use CurrentValuePublisher for `[LiveLocationShare]` updates

Velin92:
The reason is that the Passthrough was sending so fast, that it may have made the subscribers miss the first update.


pixlwave (approved):
LGTM, that's a nice little tidy-up regardless of the bug 👌

PR #5468, opened 2026-04-22
Bold the room cell when highlighted if not showing all activity.

pixlwave:
Small fix for #5312, the bolding should still be applied for highlighted rooms when not showing all activity.


stefanceriu (approved):
Nice, thank you 👍

pixlwave:
The failing test is an Element Call one.

PR #5469, opened 2026-04-22
LocationSharingScreen: fix indefinite loading on the center to location button in non picker mode

### LINKED ISSUE #5462: [Task] iOS: Pending location when permissions are Ask Next Time Or When I Share
1. Alice starts sharing location.
2. Bob's location permissions are set to _Ask Next Time Or When I Share_
3. Bob taps on the timeline to open the map.

Expected: Bob sees the current location button at the top right of the map, and when tapping is asked to give location permissions.
Actual: There is a spinner at the place of the current location button which does not seem to go away.

Image

Velin92:
In picker mode since `showAndFollow` is the default state and the permissions are asked immediately to the user we display a spinner in the center to location button while the user grants permission or if they are granted, while the system gives the first user position.

In other modes however, permissions are not asked immediately, and would only be asked if the user wants the map to center on his own location, so we don't need to be so strict with the loading check, so we only show a loader if permissions ARE granted AND the location has not been loaded yet.

fixes #5462

### DIFF
--- a/UnitTests/Sources/LocationSharingScreenViewModelTests.swift
+++ b/UnitTests/Sources/LocationSharingScreenViewModelTests.swift
@@ -169,6 +169,7 @@ final class LocationSharingScreenViewModelTests {
         #expect(!context.viewState.isLocationLoading)
     }
     
+    @Test
     func isLocationLoadingInNonPickerModeWithAuthorizationGiven() {
         let aliceShare = makeLiveLocationShare(userID: "@alice:matrix.org")
         let sender = TimelineItemSender(id: "@alice:matrix.org", displayName: "Alice")
--- a/UnitTests/Sources/LocationSharingScreenViewModelTests.swift
+++ b/UnitTests/Sources/LocationSharingScreenViewModelTests.swift
@@ -124,6 +124,61 @@ final class LocationSharingScreenViewModelTests {
         }
     }
     
+    // MARK: - isLocationLoading Tests
+
+    @Test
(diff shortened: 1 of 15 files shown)


stefanceriu (approved):
Alrighty, I'll trust you on this one 🚀

UnitTests/Sources/LocationSharingScreenViewModelTests.swift
```diff
@@ -124,6 +124,61 @@ final class LocationSharingScreenViewModelTests {
...
+        let aliceShare = makeLiveLocationShare(userID: "@alice:matrix.org")
+        let sender = TimelineItemSender(id: "@alice:matrix.org", displayName: "Alice")
+        let liveLocationsSubject = CurrentValueSubject<[LiveLocationShare], Never>([aliceShare])
+        setupViewModelForViewLive(sender: sender, initialShare: aliceShare, liveLocationsSubject: liveLocationsSubject)
+        context.isLocationAuthorized = nil
+        context.hasLoadedUserLocation = false
+        #expect(!context.viewState.isLocationLoading)
+    }
+    
+    func isLocationLoadingInNonPickerModeWithAuthorizationGiven() {
```
pixlwave, line 173: Not a review, just skimmed and I think this one is missing an `@Test`.

Velin92: whoopsie doopsie

PR #5471, opened 2026-04-22
Stop showing live location disclaimer once accepted.

### LINKED ISSUE #5463: [Task] iOS: Show disclaimer on LLS only once (per device)
Currently on iOS the disclaimer is shown each time you start sharing live location.

Velin92:
Name says it all.

fixes #5463

### DIFF
--- a/ElementX/Sources/Services/Location/LiveLocationManager.swift
+++ b/ElementX/Sources/Services/Location/LiveLocationManager.swift
@@ -58,6 +58,15 @@ class LiveLocationManager: NSObject, LiveLocationManagerProtocol, CLLocationMana
 
     // MARK: - LiveLocationManagerProtocol
     
+    var hasDisplayedLiveLocationDisclaimer: Bool {
+        get {
+            appSettings.liveLocationDisclaimerDisplayed
+        }
+        set {
+            appSettings.liveLocationDisclaimerDisplayed = newValue
+        }
+    }
+    
     @discardableResult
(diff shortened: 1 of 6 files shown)


ElementX/Sources/Services/Location/LiveLocationManager.swift
```diff
@@ -58,6 +58,15 @@ class LiveLocationManager: NSObject, LiveLocationManagerProtocol, CLLocationMana
...
     // MARK: - LiveLocationManagerProtocol
     
+    var hasDisplayedLiveLocationDisclaimer: Bool {
+        get {
+            appSettings.liveLocationDisclaimerDisplayed
+        }
+        set {
+            appSettings.liveLocationDisclaimerDisplayed = newValue
+        }
+    }
```
pixlwave, line 68: Minor nit: It seems more in line with out overall architecture to use `appSettings.liveLocationDisclaimerDisplayed` directly in the view model than forwarding it to the settings here. Amongst other things, if you ever needed to subscribe to it in the future, it wouldn't be possible this way.

pixlwave (approved):
Make sense to me.

ElementX/Sources/Services/Location/LiveLocationManager.swift
Velin92: I know, but at the same time we don't need a subscription for this and also it's tied to starting LLS. And didn't want to refactor a lot of code and tests just for one simple quick check that is done once per install and never again.

PR #5472, opened 2026-04-23
Add a Developer Options button to the AuthenticationStartScreen on Nightly builds.

pixlwave:
iOS equivalent to https://github.com/element-hq/element-x-android/pull/6587

I've deviated slightly by using the same icon as we use in the SettingScreen and have put it at the bottom rather than the top. The store sizes section won't be shown as there isn't a ClientProxy at this stage.

https://github.com/user-attachments/assets/41189503-9c06-4313-81ab-e79a22c9dbea

### DIFF
--- a/ElementX/Sources/Screens/Authentication/StartScreen/View/AuthenticationStartScreen.swift
+++ b/ElementX/Sources/Screens/Authentication/StartScreen/View/AuthenticationStartScreen.swift
@@ -123,12 +123,15 @@ struct AuthenticationStartScreen: View {
             versionText
                 .font(.compound.bodySM)
                 .foregroundColor(.compound.textSecondary)
-                .frame(maxWidth: .infinity)
-                .padding(.top, 16)
                 .onTapGesture(count: 7) {
                     context.send(viewAction: .reportProblem)
                 }
                 .accessibilityIdentifier(A11yIdentifiers.authenticationStartScreen.appVersion)
+                .overlay(alignment: .trailing) {
+                    developerOptionsButton
+                        .scaledOffset(x: 32, y: -0.5, relativeTo: .compound.bodySM)
+                }
+                .padding(.top, 16)
         }
         .padding(.horizontal, verticalSizeClass == .compact ? 128 : 24)
         .readableFrame()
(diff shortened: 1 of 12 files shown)


ElementX/Sources/Screens/Authentication/StartScreen/View/AuthenticationStartScreen.swift
```diff
@@ -123,12 +123,15 @@ struct AuthenticationStartScreen: View {
             versionText
                 .font(.compound.bodySM)
                 .foregroundColor(.compound.textSecondary)
-                .frame(maxWidth: .infinity)
-                .padding(.top, 16)
                 .onTapGesture(count: 7) {
                     context.send(viewAction: .reportProblem)
                 }
                 .accessibilityIdentifier(A11yIdentifiers.authenticationStartScreen.appVersion)
+                .overlay(alignment: .trailing) {
```
pixlwave, line 130: This might seem a bit weird not using an `HStack` but I didn't really want the label to move over when the button was visible.

ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/View/DeveloperOptionsScreen.swift
```diff
@@ -187,6 +190,19 @@ struct DeveloperOptionsScreen: View {
...
+        if context.viewState.isPresentedModally {
+            ToolbarItem(placement: .primaryAction) {
+                if #available(iOS 26.0, *) {
+                    Button(role: .close, action: dismiss.callAsFunction)
+                } else {
+                    Button(L10n.actionDone, action: dismiss.callAsFunction)
+                }
+            }
+        }
+    }
```
pixlwave, line 205: Seeing as we don't use Compound in this screen I opted for doing this manually rather than using `ToolbarButton`.

Velin92 (approved):

ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/View/DeveloperOptionsScreen.swift
Velin92: why can't we use Compound in this screen?

pixlwave: It's not that we can't. But we haven't, I think simply because it's an internal screen and not meant to be user facing so it's just pure SwiftUI in here.

pixlwave:
The test failures are unrelated (calls and chats flow)

PR #5473, opened 2026-04-23
Combine DM call actions into one menu item

amshakal:
This helps give more space to the room header. Specially in the scenario when threads is enabled.

### Pull Request Checklist

- [ ] Voiceover enabled.


amshakal:
I realised this doesn't take into account if threads developer option is turnd on or off, which is what was the motivation to create this PR in the first place.

PR #5475, opened 2026-04-23
Live Location Sharing: force a stop before starting a new session

### LINKED ISSUE #5464: [Task] iOS: Previous session not stopped when starting sharing again
For the scenario when Bob is already sharing in the room, and starts sharing again, it works fine if the Bob starts on the same device. However, timeline shows that both sessions are active when you start it from another device. Note that Web seems to be handling this correctly:

1. Start sharing from Web, then start sharing from EXI - EXI shows 2 active sessions, Web shows one (as expected).
2. Start sharing from EXI, then start sharing from Web - EXI shows 2 active sessions, Web shows one (as expected).

Velin92:
Always force a stop before starting a new session, in this way if a session is present on another client, we can force a stop on it.

Also added a re-send of the last location when a new session starts, while existing sessions are ongoing, otherwise the new session would not receive an update until the user moves over the distance filter.

fixes #5464

### DIFF
--- a/ElementX/Sources/Services/Location/LiveLocationManager.swift
+++ b/ElementX/Sources/Services/Location/LiveLocationManager.swift
@@ -28,6 +28,8 @@ class LiveLocationManager: NSObject, LiveLocationManagerProtocol, CLLocationMana
     
     private var isUpdatingLocation = false
     
+    private var lastLocation: CLLocationCoordinate2D?
+    
     @MainActor
     init(clientProxy: ClientProxyProtocol,
          appSettings: AppSettings,
@@ -76,10 +78,11 @@ class LiveLocationManager: NSObject, LiveLocationManagerProtocol, CLLocationMana
     }
     
     func startLiveLocation(roomID: String, duration: Duration) async -> Result<Void, LiveLocationManagerError> {
-        // Stop any existing session for this room first (e.g. one started from a different device)
-        // before starting a new one.
+        // Stop any existing session for this room first
+        var didAlreadyStopLocalSession = false
         if appSettings.liveLocationSharingTimeoutDatesByRoomID[roomID] != nil {
             await stopLiveLocation(roomID: roomID)
+            didAlreadyStopLocalSession = true
         }
         
         guard case .joined(let roomProxy) = await clientProxy.roomForIdentifier(roomID) else {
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Services/Location/LiveLocationManager.swift
```diff
@@ -28,6 +28,8 @@ class LiveLocationManager: NSObject, LiveLocationManagerProtocol, CLLocationMana
     
     private var isUpdatingLocation = false
     
+    private var lastLocation: CLLocationCoordinate2D?
```
pixlwave, line 31: I wonder if it would be better to combine these into an enum at this stage:

```
enum State { case stopped, starting, sending(CLLocationCoordinator2D), error }
```

It might make the flow a bit easier to follow.

ElementX/Sources/Services/Location/LiveLocationManager.swift
```diff
@@ -145,6 +160,7 @@ class LiveLocationManager: NSObject, LiveLocationManagerProtocol, CLLocationMana
         
         MXLog.verbose("Received location update via delegate, sending to rooms")
         locationUpdateSubject.send(location.coordinate)
+        lastLocation = location.coordinate
     }
     
     func locationManager(_ manager: CLLocationManager, didFailWithError error: Error) {
```
pixlwave, line 166: Do we need to handle the lastLocation via this error path too? (would probably be sorted with the enum suggestion above though).

pixlwave (approved):
LGTM 👍

ElementX/Sources/Services/Location/LiveLocationManager.swift
Velin92: the error will stop all session, which will eventually stop the manager and the lastLocation will be set to nil.

ElementX/Sources/Services/Location/LiveLocationManager.swift
Velin92: given the fact that error just stops (so they are kinda the same state) and that there is not yet a huge difference between starting and sending (since the didUpdateLocation delegate method is called instantly as you call start), I think a clear state enum is not needed... yet! But might change once maybe we get a better error management from the SDK.

pixlwave: Makes sense 👍

PR #5477, opened 2026-04-23
Open Live Locations map on tapping on the banner

### LINKED ISSUE #5474: [Task] iOS: Open map by tapping on the banner
When tapping on the banner that is shown on the device which shares its location, it should open the map. iOS currently does not do that.

Velin92:
fixes #5474 

This will make the banner open the map and always focus on the first LLS that arrives (which we can assume given the sorting algorithm, is the one for the own user of the app).


stefanceriu (approved):
lgtm 👍

PR #5478, opened 2026-04-23
Update SDK to 26.04.23

Velin92:
This PR updates the SDK.

It also updates some API to match the new names from the SDK, and adds a simple nice error management block in the sendLiveLocation API to handle cases where the session can't be considered active anymore.


pixlwave (approved):
LGTM 👍

PR #5479, opened 2026-04-24
update sim version to 26.4.1

Velin92:
name says it all


pixlwave (approved):
LGTM, but there's also UI/Integration/Accessibility tests which we should probably update too.

stefanceriu (approved):

PR #5484, opened 2026-04-24
Fix a bug where calls don't ring when quiet notifications are hidden.

pixlwave:
The feature flag should be applied after pre-processing so that calls do the right thing (at least while call notifications don't come through as noisy anyway).


stefanceriu (approved):
👌

PR #5486, opened 2026-04-25
Allow the mac app to chose room avatars directly from files instead of the photo library

stefanceriu:


pixlwave:
Honestly, I think if we're adding support for Files on one platform we should do it on both. Both of these seem perfectly reasonable on either platform to me:
- Someone wants to use a nice photo they just took as their profile picture → pick from Photos.
- Someone wants to use an image from a folder (in the Cloud) as their profile picture → pick from Files.

stefanceriu:
Right well I came at this from the perspective of "nobody in their right mind would chose a room avatar from their photo library on the mac", but sure, we can do both. I would however still fix this PR and add the rest later as it might get delayed in product team discussions and this fixes a real issue.

pixlwave:
> "nobody in their right mind would chose a room avatar from their photo library on the mac"

It's the same photo library as on their phone though? So if that was true then it would also be true on their phone.

stefanceriu:
I would argue file system interactions are completely opposite on both platforms, with Files on iOS being a second class citzen

Velin92 (approved):

pixlwave (approved):
😎

PR #5501, opened 2026-04-27
Avoid logouts when resetting the caches

stefanceriu:
The SDK is reporting a `Failed to run migrations: locking protocol` but in practice we tear down the session before using said session to access the `clearCaches` method on its client, which is obviously wrong.

This patch removes that step and the delay as everything seems to work properly without them.


pixlwave (approved):
If it works, then sgtm 👍

PR #5502, opened 2026-04-27
Fix UI Tests

Velin92:
name says it all.

I brought them to a point where they pass all locally at least, but might still need to check if there is any CI flakyness .


UITests/Sources/StartChatTests.swift
```diff
@@ -59,8 +59,14 @@ class StartChatTests: XCTestCase {
...
         inviteUsersSearchField.clearAndTypeText("Bob\n", app: app)
+        
         let cells = app.collectionViews.firstMatch.cells
-        XCTAssertEqual(cells.count, 2)
+        // The update of the cells is not immediate, and may take a bunch of milliseconds
+        // so let's await it.
+        let predicate = NSPredicate(format: "count == 2")
+        let expectation = XCTNSPredicateExpectation(predicate: predicate, object: cells)
+        await fulfillment(of: [expectation], timeout: 5.0)
+
```
stefanceriu, line 69: 👌

stefanceriu (approved):
Nice one 👍

PR #5506, opened 2026-04-28
Timeline: add a Jump to Unread button and a new-message badge on the scroll-to-bottom button.

danderson-cont:
## Summary

Two related UX additions for navigating unread messages in a room.

- **Jump to Unread button** — top-trailing button that appears when the read marker (NEW banner) exists in the timeline but isn't visible. Tapping scrolls the marker to the middle of the viewport. Hides once the marker is on screen or the user has scrolled past it.
- **New-message badge** — count badge on the existing scroll-to-bottom button. Increments as new message-content items arrive while the user is scrolled away from the bottom. Clears when they return to the bottom (organically or via tap).

## Test plan

- [x] Unit tests pass (22/22, includes new \`TimelineStateTests\` covering \`unreadMessageCount\` with mixed item types and absent read marker).
- [x] Manual smoke: open a busy room with unreads, top-right button appears with correct count.
- [x] Scroll up / down past the read marker, visibility flips correctly.
- [x] Receive new messages while scrolled up, badge increments.
- [x] Tap scroll-to-bottom (or scroll back manually), badge clears.
- [ ] Dark mode + iPad orientations to spot-check before merge.


danderson-cont:
@manuroe Can you also add the label pr-feature ?

I think Danger needs it?

ElementX/Sources/Other/SwiftUI/Views/TimelineScrollToBottomButton.swift
stefanceriu, line 1: Would've been nice to rename this in a separate commit to see how it evolved.

PreviewTests/Sources/GeneratedPreviewTests.swift
```diff
@@ -1587,6 +1587,14 @@ extension PreviewTests {
         }
     }
 
+    @Test
+    func timelineScrollButton() async throws {
```
stefanceriu, line 1591: You should run the preview tests and generate a new snapshot for this. You'll need an SE 3rd gen us-en simulator if you do. 

I don't think I've seen LFS images being PRs from a fork before, does that even work? 🤔

ElementX/Resources/Localizations/en.lproj/Untranslated.strings
```diff
@@ -1,3 +1,4 @@
+"a11y_jump_to_unread" = "Jump to unread";
```
stefanceriu, line 1: Note to self: move this to localazy. Element web is using `keyboard.jump_to_read_marker`/`Jump to oldest unread message` but that might be a bit too much.

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
```diff
@@ -27,8 +27,17 @@ struct RoomScreen: View {
...
     var body: some View {
         TimelineView(timelineContext: timelineContext)
+            .overlay(alignment: .topTrailing) {
+                if timelineContext.viewState.shouldShowJumpToUnread {
+                    TimelineScrollButton(direction: .up,
+                                         badgeCount: timelineContext.viewState.timelineState.unreadMessageCount) {
+                        timelineContext.send(viewAction: .scrollToFirstUnread)
+                    }
+                }
+            }
```
stefanceriu: Seems this gets hidden by the pinned events banner. I don't know what the right solution would be, perhaps putting it on top of the bottom one instead? 🤷‍♂️ 

https://github.com/user-attachments/assets/87741ddb-efe8-42ae-bfaf-873117a5e1f2

ElementX/Sources/Screens/Timeline/TimelineModels.swift
```diff
@@ -58,6 +58,7 @@ enum TimelineViewAction {
     case paginateForwards
     case scrollToBottom
     case scrollToFirstItemForCurrentDate
+    case scrollToFirstUnread
```
stefanceriu: First/last tends to get confusing over time, can we use oldest/newest everywhere instead?

ElementX/Sources/Other/SwiftUI/Views/TimelineScrollButton.swift
```diff
@@ -0,0 +1,102 @@
...
+        }
+        .opacity(isHidden ? 0 : 1)
+        .allowsHitTesting(!isHidden)
+        .accessibilityHidden(isHidden)
+        .accessibilityLabel(direction == .down ? L10n.a11yJumpToBottom : UntranslatedL10n.a11yJumpToUnread)
+        .animation(.elementDefault, value: isHidden)
+    }
+
+    private var icon: some View {
+        Image(systemName: direction == .down ? "chevron.down" : "chevron.up")
```
stefanceriu: It's probably time to move to compound `chevronDown` and `chevronUp` here.

ElementX/Sources/Other/SwiftUI/Views/TimelineScrollButton.swift
```diff
@@ -0,0 +1,102 @@
...
+    static var previews: some View {
+        VStack(spacing: 16) {
+            TimelineScrollButton(direction: .down) { }
+            TimelineScrollButton(direction: .down, badgeCount: 3) { }
+            TimelineScrollButton(direction: .down, badgeCount: 42) { }
+            TimelineScrollButton(direction: .up, badgeCount: 7) { }
+            TimelineScrollButton(direction: .up, isHidden: true, badgeCount: 7) { }
+        }
+        .padding()
+        .background(Color.compound.bgCanvasDefault)
```
stefanceriu, line 138: You might want a `.previewLayout(.sizeThatFits)` here.

ElementX/Sources/Other/SwiftUI/Views/TimelineScrollButton.swift
```diff
@@ -0,0 +1,102 @@
...
+         isHidden: Bool = false,
+         badgeCount: Int = 0,
+         callback: @escaping () -> Void) {
+        self.direction = direction
+        self.isHidden = isHidden
+        self.badgeCount = badgeCount
+        self.callback = callback
+    }
+
+    var body: some View {
```
stefanceriu, line 54: Layout is a tad off on larger dynamic types.

Image

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemViewState.swift
```diff
@@ -203,4 +203,18 @@ enum RoomTimelineItemType: Equatable {
             return nil
         }
     }
+
+    /// Whether this item represents a user-facing message that should be counted towards unread
+    /// totals. Excludes virtual items (separators, read marker, pagination indicators), redacted
+    /// messages, state events and unsupported items.
+    var isMessageContent: Bool {
```
stefanceriu: I wonder if it would be useful to move these one level above into the uderlying `EventBasedTimelineItemProtocol`.

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
```diff
@@ -646,7 +648,12 @@ class TimelineViewModel: TimelineViewModelType, TimelineViewModelProtocol {
             focusLive()
         }
     }
-    
+
+    private func scrollToFirstUnread() {
+        guard let readMarkerID = state.timelineState.readMarkerUniqueID else { return }
```
stefanceriu: Presumably the button shouldn't be shown if the identifier isn't available, should this be treated as a programmer/fatal error?

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
```diff
@@ -875,6 +885,33 @@ class TimelineViewModel: TimelineViewModelType, TimelineViewModelProtocol {
             && otherEventTimelineItem.timestamp.timeIntervalSince(eventTimelineItem.timestamp) < 5 * 60 // As does the passage of time.
     }
 
+    /// Increments `newMessagesAtBottomCount` by the number of newly arrived message-content items
+    /// when the user is scrolled up in a live timeline. Skips initial load and timeline switches.
+    private func updateNewMessagesAtBottomCount(with newItemsDictionary: OrderedDictionary<TimelineItemIdentifier.UniqueID, RoomTimelineItemViewState>,
+                                                isSwitchingTimelines: Bool) {
```
stefanceriu: Why pass this one in instead of reading it from the state like the others? 🤔

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
```diff
@@ -841,7 +848,10 @@ class TimelineViewModel: TimelineViewModelType, TimelineViewModelProtocol {
         if isSwitchingTimelines {
             state.timelineState.isSwitchingTimelines = true
         }
-        
+
+        // Read the previous dictionary before we overwrite it on the next line.
+        updateNewMessagesAtBottomCount(with: timelineItemsDictionary, isSwitchingTimelines: isSwitchingTimelines)
```
stefanceriu: Perhaps it would be better to be explicit and call the parameter `newTimelineItems`, you can then probably drop the comment too.

ElementX/Sources/Screens/Timeline/TimelineModels.swift
```diff
@@ -248,17 +258,45 @@ struct TimelineState {
     /// These can be removed when we have full swiftUI and moved as @State values in the view
     var scrollToBottomPublisher = PassthroughSubject<Void, Never>()
     var scrollToFirstItemForDatePublisher = PassthroughSubject<Void, Never>()
-    
+    var scrollToFirstUnreadPublisher = PassthroughSubject<TimelineItemIdentifier.UniqueID, Never>()
```
stefanceriu: Should we maybe stick with "readMarker" everywhere for consistency? Scroll to read marker, is read marker visible etc.

ElementX/Sources/Screens/Timeline/TimelineModels.swift
```diff
@@ -248,17 +258,45 @@ struct TimelineState {
...
+    var readMarkerUniqueID: TimelineItemIdentifier.UniqueID? {
+        itemsDictionary.first { _, viewState in
+            if case .readMarker = viewState.type { return true }
+            return false
+        }?.key
+    }
+
+    /// The number of message-content items that follow the read marker. Returns 0 if there is no
+    /// read marker.
+    var unreadMessageCount: Int {
```
stefanceriu: Does this need to be computed on demand or can we do it in the view model and just store it the state similar to `newMessagesAtBottomCount`? 🤔

ElementX/Sources/Screens/Timeline/TimelineModels.swift
```diff
@@ -248,17 +258,45 @@ struct TimelineState {
...
+        var foundMarker = false
+        for (uniqueID, viewState) in itemsDictionary {
+            if uniqueID == readMarkerID {
+                foundMarker = true
+                continue
+            }
+            if foundMarker, viewState.type.isMessageContent {
+                count += 1
+            }
+        }
```
stefanceriu: Can this be replaced by :?

```
return itemsDictionary.split { (key, value) in
    key == readMarkerUniqueID
}.first?.count ?? 0
```

(haven't actually tested it but something along those line should work)

ElementX/Sources/Screens/Timeline/TimelineModels.swift
```diff
@@ -270,6 +308,17 @@ enum ScrollDirection: Equatable {
...
 
 extension TimelineViewState {
+    /// The user is at the bottom of a live timeline (no jump-to-bottom button needed).
+    var isAtBottomAndLive: Bool {
+        bindings.isScrolledToBottom && timelineState.isLive
+    }
+
+    /// Whether the jump-to-unread button should be shown: a read marker exists in the timeline
+    /// and isn't currently visible in the viewport.
+    var shouldShowJumpToUnread: Bool {
```
stefanceriu: Again here, jumpToReadMarker maybe?

ElementX/Sources/Other/SwiftUI/Views/TimelineScrollButton.swift
```diff
@@ -0,0 +1,102 @@
...
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Compound
+import SwiftUI
+
+/// A small floating circular button used by the timeline to jump to either the bottom (newest
+/// item) or to the read marker (NEW banner). Optionally displays a count badge.
+struct TimelineScrollButton: View {
```
stefanceriu, line 14: Can't really comment on the UI, looks like this @americanrefugee.

Image

Image

stefanceriu (commented):
Very nice start, thank you for looking into this!

You touched on the button not showing up in gappy syncs under `edge cases not covered` but that same situation occurs within sdk event cache paginated timeline initializations, which happen for performance reasons. They're quite easy to reproduce in busy enough rooms.

That being said, perhaps the simplest way forward here is to put it under a feature flag and see how it behaves in practice, perhaps it's just good enough for most of the users.

Another option would be to drop unread counts entirely (as I imagine that's not straight forward on the sdk) and only do the event based timeline focus. 

https://github.com/user-attachments/assets/2a49157e-bb87-465b-a86b-cb31beda9f8c

danderson-cont:
Thanks for the review.

Took the feature flag path. `jumpToReadMarkerEnabled` lives in `AppSettings` with a Developer Options toggle, default off. With the flag off the existing scroll-to-bottom button is unchanged from `develop`.

On the "mark all as read" affordance from the meta ticket: I don't think iOS needs it. Leaving the room already advances `m.fully_read`, so the room is marked fully read when the user backs out. And the simplicity of the current single-button control is nice; bolting "mark all as read" onto it would clutter the affordance without much practical gain.

Pushing back gently on dropping unread counts. The down-arrow badge is intentionally session-local ("messages that arrived while you were looking away") rather than SDK unread count, and the up-arrow count exists because both buttons can be visible at the same time (you can have unreads above your current position in the room AND new messages arrived while partway scrolled). They answer different questions; dropping one would lose information.

The gappy-sync / paginated-init edge case is real and not covered. Cleanest upgrade path I can see, in three steps:

1. Land this.
2. Expose `firstUnreadEventID` from the SDK.
3. Use it to drive jump-to-read-marker visibility, and route the tap through the existing `focusOnEventID` path when the local read marker isn't loaded.

ElementX/Sources/Other/SwiftUI/Views/TimelineScrollToBottomButton.swift
danderson-cont: Fair. Will keep that in mind for future PRs.

PreviewTests/Sources/GeneratedPreviewTests.swift
danderson-cont: Skipped for two reasons:

1. `PreviewTests.swift` pins `requiredOSVersion` to (26, 4) and I'm on Xcode 26.3 without the 26.4 simulator runtime installed.
2. Your LFS-from-fork question made me uncertain whether the PNG would even survive the push.

Happy to record locally and push if you can confirm LFS-from-fork works for you, or you can record on your end. The preview itself renders without crashing.

ElementX/Resources/Localizations/en.lproj/Untranslated.strings
danderson-cont: Kept it short. Your "might be a bit too much" hedge convinced me. Happy to flip to web's longer phrasing during the Localazy import if you'd rather have cross-platform consistency.

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
danderson-cont: Fixed. Added a `trailingAccessory:` slot to `topBanners(_:footer:trailingAccessory:)` and put the jump-to-read-marker button there. Layout-positioned below the banner stack via the same VStack the footer uses, so it slides correctly with the banner animation rather than relying on measured height. Tried a PreferenceKey-based measurement first but the read-back didn't track the slide-in cleanly. Going through layout was simpler.

ElementX/Sources/Screens/Timeline/TimelineModels.swift
danderson-cont: Renamed `case scrollToFirstUnread` to `case scrollToReadMarker`.

ElementX/Sources/Other/SwiftUI/Views/TimelineScrollButton.swift
danderson-cont: Done. Replaced `Image(systemName: "chevron.down/up")` with `CompoundIcon(\.chevronDown / \.chevronUp, size: .custom(iconSize), relativeTo: .compound.bodyLG)`. Dropped `.fontWeight(.semibold)` (vector icons don't take a weight) and the 1pt vertical offset that was compensating for SF Symbol's centering.

ElementX/Sources/Other/SwiftUI/Views/TimelineScrollButton.swift
danderson-cont: Done.

ElementX/Sources/Other/SwiftUI/Views/TimelineScrollButton.swift
danderson-cont: Done. Wrapped badge dimensions in `@ScaledMetric(relativeTo: .compound.bodyLG)` so they scale with the chevron at large dynamic types.

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemViewState.swift
danderson-cont: Investigated but kept it on the enum. The catch: `.callInvite` and `.callNotification` only conform to `RoomTimelineItemProtocol`, not `EventBasedTimelineItemProtocol`, so moving the helper to that protocol would silently drop them from the count. Pushing it down to `RoomTimelineItemProtocol` instead works but spreads the classification across most of the timeline item types with overrides flipping the default in both directions (true on `EventBased`, false on the non-message subtypes State/Redacted/Unsupported/LiveLocation, true on Call*). The current single-switch on the enum keeps the logic auditable in one place and gets exhaustive-switch enforcement for new cases. Happy to revisit if you'd prefer the spread version.

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
danderson-cont: Rather not crash here. `readMarkerUniqueID == nil` can happen legitimately. Room marked read on another device while the action is in flight, or the marker briefly absent during a timeline switch. The button is hidden when the ID is nil so the action shouldn't normally fire, but I'd treat the race as a soft no-op rather than a `fatalError`. Happy to add an `MXLog.error` if it'd help diagnose unexpected occurrences.

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
danderson-cont: Because at this point `state.timelineState.itemsDictionary` still holds the *previous* dictionary. We read it as `oldDictionary` for the diff against the new one. The rename you suggested in the other comment (`newTimelineItems`) makes that explicit. Happy to flip the call ordering instead so state is written first and a snapshot of the old is held in a local. Same result either way, let me know if you'd prefer.

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
danderson-cont: Done. Renamed to `newTimelineItems` and dropped the explanatory comment.

ElementX/Sources/Screens/Timeline/TimelineModels.swift
danderson-cont: Done. `scrollToFirstUnreadPublisher` to `scrollToReadMarkerPublisher`, `shouldShowJumpToUnread` to `shouldShowJumpToReadMarker`, `case scrollToFirstUnread` to `case scrollToReadMarker`. Single rename commit.

ElementX/Sources/Screens/Timeline/TimelineModels.swift
danderson-cont: Done. Converted to stored `private(set) var readMarkerUniqueID` and `unreadMessageCount` on `TimelineState`, populated by a new `mutating func recomputeReadMarkerState()` that runs after `itemsDictionary` is assigned. Tests call it too.

ElementX/Sources/Screens/Timeline/TimelineModels.swift
danderson-cont: I prefer the explicit loop as it reads more clearly and includes the message-content filter, as well as the split version should have last instead of first due to the table being flipped.

ElementX/Sources/Screens/Timeline/TimelineModels.swift
danderson-cont: Done in the rename commit alongside the readMarker consistency rename above.

ElementX/Sources/Other/SwiftUI/Views/TimelineScrollButton.swift
danderson-cont: Sized to match the classic-mobile reference screenshots in element-meta#3056. Badge color is `.compound.iconAccentTertiary` (same green as the home cell unread dot). Eyeballed in sim across default and large dynamic-type sizes.

Danbana: Simulator Screenshot - iPhone 17 - 2026-05-03 at 23 06 13

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
stefanceriu: Oh sorry, I was referring only to the `isSwitchingTimelines` parameter which is already set at this point.

ElementX/Sources/Screens/RoomScreen/View/RoomScreen.swift
stefanceriu: I don't think this is particularly better, we need design input.


https://github.com/user-attachments/assets/b2409c48-d6b1-46bf-8205-1cfd71d9d992

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemViewState.swift
stefanceriu: No, it's fine, thanks. I was mostly thinking out loud with this one.

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
stefanceriu: I'm not convinced, how can this action race against its own `timelineState`? They should both be run on the main actor 🤔

ElementX/Sources/Other/SwiftUI/Views/TimelineScrollButton.swift
stefanceriu: I like it but do see https://github.com/element-hq/element-x-ios/pull/5506#discussion_r3181323648

PreviewTests/Sources/GeneratedPreviewTests.swift
stefanceriu: The only thing I can confirm at the moment is that pushing to fork PRs doesn't work from my side (we tried that in https://github.com/element-hq/element-x-ios/pull/5384#issuecomment-4251431274). 
We could record it later but it would be nice to see if you can push to lfs or not. 
Try adding the snapshots from this zip (not sure what's happening with the pseudo double ones 🤔)

[Archive.zip](https://github.com/user-attachments/files/27359424/Archive.zip)

stefanceriu:
Great, thanks for the tweaks. We're getting there!

> Pushing back gently on dropping unread counts

> 2) Expose firstUnreadEventID from the SDK & 3) Use it to drive jump-to-read-marker visibility

Not sure these 2 fit together as the SDK wouldn't provide the unread counts, at least not as easily as the read marker identifier.

ElementX/Sources/Other/SwiftUI/Views/TimelineScrollButton.swift
```diff
@@ -0,0 +1,110 @@
...
+
+    let direction: Direction
+    let isHidden: Bool
+    let badgeCount: Int
+    let callback: () -> Void
+
+    @ScaledMetric(relativeTo: .compound.bodyLG) private var iconSize: CGFloat = 22
+    @ScaledMetric(relativeTo: .compound.bodyLG) private var badgeSize: CGFloat = 20
+    @ScaledMetric(relativeTo: .compound.bodyLG) private var badgeDotSize: CGFloat = 14
+    @ScaledMetric(relativeTo: .compound.bodyLG) private var badgeOffsetY: CGFloat = -10
```
pixlwave: Just a small thought on accessibility, not a review:
- We have some modifiers to help here `scaledFrame`/`scaledPadding`/`scaledOffset`. Should let you inline the values rather than having to reach for `@ScaledMetric` manually.
- `CompoundIcon` already scales with DT size automatically so no need for `@ScaledMetric` there.
- Maybe the `badgeDotSize` (and possibly offset) should be relative to the font used for the badge count rather than to `bodyLG` (it'll probably only have an impact on `.small` and `.xSmall` sizes).

ElementX/Sources/Screens/Timeline/TimelineModels.swift
```diff
@@ -248,17 +259,44 @@ struct TimelineState {
     /// These can be removed when we have full swiftUI and moved as @State values in the view
     var scrollToBottomPublisher = PassthroughSubject<Void, Never>()
     var scrollToFirstItemForDatePublisher = PassthroughSubject<Void, Never>()
-    
+    var scrollToReadMarkerPublisher = PassthroughSubject<TimelineItemIdentifier.UniqueID, Never>()
+
     var itemsDictionary = OrderedDictionary<TimelineItemIdentifier.UniqueID, RoomTimelineItemViewState>()
-    
+
```
pixlwave, line 266: I see quite a few of these in the PR – we all run Xcode with "Trim whitespace-only lines" disabled to stop this, although annoyingly Xcode Intelligence doesn't seem to honour that setting and we haven't yet found a way to lint/format the project like this to raise warnings. Either way, once the setting above is disabled, re-indenting the file with Xcode should fix these for you (although SwiftFormat may then fix a thing or two afterwards).

Image

pixlwave: Ok I did something about it [here](https://github.com/nicklockwood/SwiftFormat/pull/2527) so 🤞 this won't be an issue going forward.

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
danderson-cont: Just defensive programming, checking the pre-conditions. I haven't seen this happen, but also I wouldn't want to crash the app if it does. How about an assertionFailure() so it would be obvious when debugging, but a no-op in production?

mxandreas:
I did some very quick tests from design perspective and it looks good. It worked functionally as well, it is just that my testing was very basic.

danderson-cont:
Pushed the SDK side work: SDK PR https://github.com/matrix-org/matrix-rust-sdk/pull/6569 exposes `fully_read_event_id` on `RoomInfo`, and on the iOS side `shouldShowJumpToReadMarker` falls back to that field when the marker isn't in the loaded window. Tap routes through `loadOrFetchEventDetails` then `focusOnEvent`, the same flow same-room permalinks use, so the detached timeline opens at the marker with the green focused-event highlight and the user can scroll forward through unread events. 

There's also an optimistic clear of `state.timelineState.fullyReadEventID` after `markAsRead(.fullyRead)` so the button hides immediately rather than waiting for the SDK echo. I went with the iOS-side optimistic marking as read, rather than asking the SDK to push an optimistic local update, but we could swap if the SDK ends up doing it.

 The two `(fix)` mock/proxy commits accommodate the matrix-rust-sdk surface from the SDK PR's branch (new `RoomMember.isServiceMember` / `SpaceRoom.isDm` / `RoomPowerLevelsValues.beacon`/`beaconInfo` init params and the now-async `subscribeToRoomUpdate`) so the local build worked while iterating; expect those to be redundant once the components-swift pin is bumped.

stefanceriu:
I merged your SDK PR and then went ahead and built a new version of the SDK bindings so you can directly use here. Please update and rebase on develop / fix whatever these conflicts github is complaining about are.

```
 project.yml | 2 +-
- exactVersion: 26.04.28
+ exactVersion: 26.05.20
```

Edit: nevermind, did it myself in the end.

stefanceriu:
Okay, I think this will have to do for now, we can address eventual feedback separately. 
Thanks for making it happen!

danderson-cont:
> I merged your SDK PR and then went ahead and built a new version of the SDK bindings so you can directly use here. Please update and rebase on develop / fix whatever these conflicts github is complaining about are.
> 
> ```
>  project.yml | 2 +-
> - exactVersion: 26.04.28
> + exactVersion: 26.05.20
> ```
> 
> Edit: nevermind, did it myself in the end.

Thanks! I was working through the issues here, but looks like this is all wrapped up.

PR #5514, opened 2026-04-29
Live Location Sharing - handle server echoes

### LINKED ISSUE #5465: [Bug] Banner does not disappear when stopping from another device
1. Bob starts sharing on device A (banner appears below room header).
2. Bob goes to device B and stops sharing from there.

Expected: Banner goes away on device A.
Actual: Banner stays.

Velin92:
This includes two important changes:
- Only starts sending a location when we are actually getting the start event echoed by the server, so that we are sure the SDK is not going to discard the sent location since it did not find the start event echo, this will then promote a starting session to an active session, and handle the send
- Check if a new event arrives in the room while an existing active local session is present, and if that's the case we can safely assume is an external event, and then remove the session locally.
- Update the tests to include testing the own beacon info updates.

fixes #5465

### DIFF
--- a/ElementX/Sources/Services/Client/ClientProxyProtocol.swift
+++ b/ElementX/Sources/Services/Client/ClientProxyProtocol.swift
@@ -90,9 +90,13 @@ enum TimelineMediaVisibility: Decodable {
     case never
 }
 
-struct OwnBeaconInfoUpdate: Equatable {
+/// Represents a server-echoed update about the current user's own beacon info state in a room.
+struct LiveLocationOwnInfoUpdate: Equatable {
+    /// The room where the beacon info event was sent.
     let roomID: String
+    /// The event ID of the beacon info state event.
     let eventID: String
+    /// Whether the beacon is currently active (live) or has been stopped.
     let isLive: Bool
 }
 
@@ -272,7 +276,8 @@ protocol ClientProxyProtocol: AnyObject {
     
     // MARK: - Live Location
 
-    var ownBeaconInfoUpdatesPublisher: AnyPublisher<OwnBeaconInfoUpdate, Never> { get }
+    /// Publishes updates about the current user's own live location beacon info state changes (start/stop) as echoed by the server.
+    var liveLocationOwnInfoUpdatesPublisher: AnyPublisher<LiveLocationOwnInfoUpdate, Never> { get }
 
     // MARK: - Moderation & Safety
     
--- a/ElementX/Sources/Services/Client/ClientProxyProtocol.swift
+++ b/ElementX/Sources/Services/Client/ClientProxyProtocol.swift
@@ -90,6 +90,12 @@ enum TimelineMediaVisibility: Decodable {
     case never
 }
 
+struct OwnBeaconInfoUpdate: Equatable {
+    let roomID: String
+    let eventID: String
+    let isLive: Bool
+}
+
 // sourcery: AutoMockable
 protocol ClientProxyProtocol: AnyObject {
     var actionsPublisher: AnyPublisher<ClientProxyAction, Never> { get }
@@ -264,6 +270,10 @@ protocol ClientProxyProtocol: AnyObject {
     
     func userIdentity(for userID: String, fallBackToServer: Bool) async -> Result<UserIdentityProxyProtocol?, ClientProxyError>
     
+    // MARK: - Live Location
+
+    var ownBeaconInfoUpdatesPublisher: AnyPublisher<OwnBeaconInfoUpdate, Never> { get }
+
     // MARK: - Moderation & Safety
     
     func setTimelineMediaVisibility(_ value: TimelineMediaVisibility) async -> Result<Void, ClientProxyError>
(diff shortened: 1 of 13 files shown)


ElementX/Sources/Services/Client/ClientProxyProtocol.swift
```diff
@@ -90,6 +90,12 @@ enum TimelineMediaVisibility: Decodable {
     case never
 }
 
+struct OwnBeaconInfoUpdate: Equatable {
```
pixlwave: We should probably add some docs describing what this type is as the name is pretty generic. Potentially we could also rename it to include LiveLocation.

ElementX/Sources/Services/Client/ClientProxyProtocol.swift
```diff
@@ -264,6 +270,10 @@ protocol ClientProxyProtocol: AnyObject {
     
     func userIdentity(for userID: String, fallBackToServer: Bool) async -> Result<UserIdentityProxyProtocol?, ClientProxyError>
     
+    // MARK: - Live Location
+
+    var ownBeaconInfoUpdatesPublisher: AnyPublisher<OwnBeaconInfoUpdate, Never> { get }
```
pixlwave: Same comments here too.

ElementX/Sources/Services/Location/LiveLocationManager.swift
```diff
@@ -20,7 +20,11 @@ class LiveLocationManager: NSObject, LiveLocationManagerProtocol, CLLocationMana
     
     /// Cached joined room proxies keyed by room ID, kept in sync with the active sessions dictionary.
     private var activeRoomProxies = [String: JoinedRoomProxyProtocol]()
-    
+
+    /// Sessions that have been requested but not yet confirmed by the server echo.
+    /// Once the server acknowledges the beacon info, sessions are promoted to the persistent store.
+    private var startingLiveLocationSharingSessionsByRoomID = [String: LiveLocationSession]()
+
```
pixlwave: The whitespace indentation has been stripped.

ElementX/Sources/Services/Location/LiveLocationManager.swift
```diff
@@ -104,28 +109,22 @@ class LiveLocationManager: NSObject, LiveLocationManagerProtocol, CLLocationMana
...
-        if isUpdatingLocation, let lastLocation {
-            // To make sure the newly started session is in sync with the existing ones,
-            // we re-send the last location received by the manager.
-            // Otherwise we would need to wait a distance filtered update.
-            locationUpdateSubject.send(lastLocation)
-        }
-        
+        let expirationDate = Date().addingTimeInterval(TimeInterval(duration.seconds))
+        startingLiveLocationSharingSessionsByRoomID[roomID] = LiveLocationSession(eventID: eventID, expirationDate: expirationDate)
+
```
pixlwave: Same, probably best to re-indent the whole file as Claude has done it's work 🙈

ElementX/Sources/Services/Location/LiveLocationManager.swift
```diff
@@ -209,7 +216,34 @@ class LiveLocationManager: NSObject, LiveLocationManagerProtocol, CLLocationMana
...
     }
     
-    private func syncActiveRoomProxies(with sessions: [String: Date]) {
+    private func handleBeaconInfoUpdate(_ update: OwnBeaconInfoUpdate) {
+        // A new beaconInfo has been received in a room with existing active session
+        // This is either a new start or a new stop from a different device, so we
+        // should remove the session from the current local one.
+        if appSettings.liveLocationSharingSessionsByRoomID[update.roomID] != nil {
+            appSettings.liveLocationSharingSessionsByRoomID.removeValue(forKey: update.roomID)
+        }
```
pixlwave: FWIW, I don't think we actually need that `if != nil` check.

UnitTests/Sources/LiveLocationManagerTests.swift
```diff
@@ -165,15 +177,35 @@ final class LiveLocationManagerTests {
...
+    // MARK: - Beacon info updates
+
+    @Test
+    func beaconInfoUpdateFromAnotherDeviceRemovesActiveSession() async throws {
+        setUp()
+        let roomProxy = makeRoomProxy(roomID: "!room:matrix.org")
+        clientProxy.roomForIdentifierClosure = { _ in .joined(roomProxy) }
+
+        let result = await manager.startLiveLocation(roomID: "!room:matrix.org", duration: .seconds(300))
+        try result.get()
```
pixlwave: I might be missing something but given we're not interested in the value couldn't this just be a single line:

```suggestion
        try await manager.startLiveLocation(roomID: "!room:matrix.org", duration: .seconds(300)).get()
```

pixlwave (approved):
LGTM! 😎

ElementX/Sources/Services/Location/LiveLocationManager.swift
Velin92: weird btw in VSCode it did not do this, but weird enough it behaves like this when using the Xcode agent.

PR #5517, opened 2026-04-30
add dev option to re-register for push notifications

mredig:
In testing my other PR, I'm running into notification issues. This is a dev feature to help with that troubleshooting that should also be useful in the long run.

### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/DeveloperOptionsScreenModels.swift
+++ b/ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/DeveloperOptionsScreenModels.swift
@@ -48,6 +48,7 @@ struct DeveloperOptionsScreenViewStateBindings {
 
 enum DeveloperOptionsScreenViewAction {
     case clearCache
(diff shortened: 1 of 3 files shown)


mredig:
(I am unable to add labels)

stefanceriu (commented):
This doesn't seem to be much different than disabling and then re-enabling notifications from the app's settings, can you work with that?

mredig:
> This doesn't seem to be much different than disabling and then re-enabling notifications from the app's settings, can you work with that?

I did not think of that. That should work in most cases, but technically it's different in that this new button is just refreshing the existing registration with APNs while the toggle off and back on is unregistering and creating a new registration.


I *think* I can work with the toggle for my needs, but I'll let you decide if that's enough differentiated value or not. Feel free to close if you don't. :)

stefanceriu:
Okay so if you don't mind I'm going to go ahead and close it. These developer flags tend to pile up and before long nobody knows if they can be cleaned up or not.

PR #5527, opened 2026-05-01
Custom Notification Tones

mredig:
### Pull Request Checklist

- [ ] Voiceover enabled.

## Add notification tone selection

Adds a tone picker to the notification settings screen, letting users choose the sound played when a notification is delivered.

### What's included

- A new **Alert Tone** section in notification settings, showing a collapsible list of available tones
- **Element X bundled tones** -- 13 tones shipped with the app
- **iOS system tones** -- 26 standard system sounds
- **Custom tone import** -- users can import their own audio files (CAF, MP3, AIFF, WAV, M4A); non-CAF formats are automatically converted to CAF for compatibility with the notification system
- **Custom tone deletion** via swipe-to-delete
- Tapping any tone **previews** it immediately via `AVAudioPlayer`
- The selection is **persisted** in app settings and picked up by the Notification Service Extension when delivering notifications

### Screenshots

IMG_0065
IMG_0066
IMG_0067

### Notes

- The tone list is only shown when the tone manager can be initialised (i.e. when required library directories are writable); if setup fails the section is hidden entirely rather than showing a broken state
- Non-CAF conversion runs on a dedicated `ConversionActor` to keep audio processing off the main thread
- The active tone is written to a fixed path (`Library/Sounds/currentAlert.caf`) that the NSE reads at notification delivery time
- Snapshot tests will need re-recording (I do not have the rights to add labels as requested in Contributing.md)

### Open questions

- The `DisclosureGroup` in the tone picker is styled ad hoc (manual `foregroundStyle`, `listRowBackground`, `listRowSeparatorTint`). Is this acceptable as-is, or should I create a reusable style for it?


mxandreas:
I played around with this one a little bit. A couple of questions and notes:

> Element X bundled tones -- 13 tones shipped with the app
> iOS system tones -- 26 standard system sounds

1) What are those 13 bundled tones? I mean where do they come from?

2) Propose to rename the Element ones as follows:

Default Message > Element Default
Pro Message > Element Fade (open for suggestions what better illustrates this sound)

3) The option for the custom one are at the very end of a long list. You may not find it. Perhaps it should be at the top?

4) The custom sound did not seem to work reliably in my simulator. The first time I selected a file, it was fine but at some point nothing happened when I tried to select a file. It just stayed on that screen. It was not frozen either because I could close the modal.

mredig:
> 1. What are those 13 bundled tones? I mean where do they come from?

We had discussions regarding a tone change in the past and our designer supplied us with a bunch of options. I took those, selected the most unique ones, and edited them down to have the fastest attack and cut as soon as it fades. I'm not 100% sure where they were sourced from, but I have a hunch it was https://freesound.org. If you need more detail, we can dig deeper.

> 2. Propose to rename the Element ones as follows:

done

> (open for suggestions what better illustrates this sound)

It reminds me of a doorbell chime, but I think 'fade' is a cleaner name.

> 3. The option for the custom one are at the very end of a long list. You may not find it. Perhaps it should be at the top?

I'll try it

> 4. The custom sound did not seem to work reliably in my simulator. The first time I selected a file, it was fine but at some point nothing happened when I tried to select a file. It just stayed on that screen. It was not frozen either because I could close the modal.

Can you elaborate on reproduction of the issue? I've tested quite thoroughly on both device and simulator and will try again, but I've not encountered anything similar on my end. Was it perhaps a specific file? I tested with a variety of files, covering the gamut of the filetypes I allowlisted, but maybe there's a specific formatting of a filetype that isn't compatible?

mxandreas:
> We had discussions regarding a tone change in the past and our designer supplied us with a bunch of options. I took those, selected the most unique ones, and edited them down to have the fastest attack and cut as soon as it fades. I'm not 100% sure where they were sourced from, but I have a hunch it was https://freesound.org/. If you need more detail, we can dig deeper.

Yes, we have to confirm what is their license, even if they do not cost anything, their license may require adding credits to the code or to the app, etc. My first preference is not to include them at all and have just the iOS standard sounds, unless there's a specific reasoning why we need them at once.

ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -16,8 +16,12 @@ import SwiftUI
 /// Common settings between app and NSE
 protocol CommonSettingsProtocol: AnyObject {
     var lastNotificationBootTime: TimeInterval? { get set }
-    var notificationSoundName: RemotePreference<UNNotificationSoundName> { get }
-    
+    #if !IS_SHARE_EXT
+    var selectedNotificationTone: NotificationAlertTone? {
```
stefanceriu: This is unnecessary, the share extension can have access to the whole thing, it doesn't matter. Presumably the problem is here is `NotificationAlertTone` file access. That whole folder should be moved outside of the Screens folder and into Services. NotificationAlertTone can be added to the ShareExtension from there or it can go into Application/Settings

ElementX/Resources/Localizations/en.lproj/Untranslated.strings
stefanceriu, line 1: Do we want to translate these maybe? cc @mxandreas

ShareExtension/SupportingFiles/target.yml
```diff
@@ -75,6 +75,11 @@ targets:
         MARKETING_VERSION: $(MARKETING_VERSION)
         CURRENT_PROJECT_VERSION: $(CURRENT_PROJECT_VERSION)
         DEVELOPMENT_TEAM: $(DEVELOPMENT_TEAM)
+      configs:
+        debug:
+          SWIFT_ACTIVE_COMPILATION_CONDITIONS: IS_SHARE_EXT DEBUG
+        release:
+          SWIFT_ACTIVE_COMPILATION_CONDITIONS: IS_SHARE_EXT RELEASE
```
stefanceriu: Let's get rid of these.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/AlertTones/NotificationToneManager.swift
```diff
@@ -0,0 +1,160 @@
...
+    /// Copies the tone's audio file to `selectedToneLocation` and persists the selection in app settings.
+    func setSelectedTone(_ alertTone: NotificationAlertTone) {
+        do {
+            try? FileManager.default.removeItem(at: NotificationAlertTone.selectedToneLocation)
+            try FileManager.default.copyItem(at: alertTone.location, to: NotificationAlertTone.selectedToneLocation)
+            appSettings.selectedNotificationTone = alertTone
+        } catch {
+            let userIndicator = UserIndicator(type: .toast,
+                                              title: UntranslatedL10n.screenNotificationSettingsSoundSetSoundErrorTitle,
+                                              iconName: "exclamationmark.triangle.fill")
```
stefanceriu: Where did this `exclamationmark.triangle.fill` come from? Here and in the other 3 places.

ElementX/Resources/Localizations/en.lproj/Untranslated.strings
```diff
@@ -11,3 +11,37 @@
...
+"screen_notification_settings_sound_system_telegraph" = "Telegraph";
+"screen_notification_settings_sound_system_tiptoes" = "Tiptoes";
+"screen_notification_settings_sound_system_typewriters" = "Typewriters";
+"screen_notification_settings_sound_system_update" = "Update";
+"screen_notification_settings_sound_system_swish" = "Swish";
+"screen_notification_settings_sound_system_tweet" = "Tweet";
+"screen_notification_settings_sound_element_default" = "Element Default";
+"screen_notification_settings_sound_element_fade" = "Element Fade";
+"screen_notification_settings_sound_custom_sound_button_title" = "Custom sound...";
+"screen_notification_settings_sound_section_title" = "Sound";
```
stefanceriu, line 43: How come we ended up calling it tone through the code but "sound" in the app? Even Whatsapp has `Alert tones` under `Notification/Sound` 😅

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/AlertTones/NotificationAlertTone.swift
```diff
@@ -0,0 +1,200 @@
...
+    static func createBundledSound(label: String?, filename: String) -> NotificationAlertTone {
+        NotificationAlertTone(labelOverride: label, storageLocationRoot: .appBundle, relativePath: [filename])
+    }
+
+    /// Creates a tone backed by a user-imported file in the app library.
+    static func createCustomUserSound(filename: String) -> NotificationAlertTone {
+        NotificationAlertTone(labelOverride: nil, storageLocationRoot: .appLibrary, relativePath: [filename])
+    }
+
+    #if IS_MAIN_APP // localization is only available in the main app
```
stefanceriu: That's not true, L10n is available just fine in the NSE, it's `UntranslatedL10n` that isn't, which in this case shouldn't even be used I imagine.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/AlertTones/NotificationTonePreviewer.swift
```diff
@@ -0,0 +1,46 @@
...
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+import AVFoundation
+
+/// Plays a preview of a notification tone using `AVAudioPlayer`.
+/// Players are retained for the duration of playback and released automatically on completion.
+class NotificationTonePreviewer: NSObject, AVAudioPlayerDelegate {
```
stefanceriu: Let's remove this whole component and use the `AudioPlayer` instead, letting the `NotificationSettingsScreenViewModel` drive it and the eventual error notifications / interaction with the UserIndicatorController

ElementX/Sources/Services/AlertTones/NotificationToneManagerProtocol.swift
```diff
@@ -0,0 +1,18 @@
...
+// Copyright 2026 Element Creations Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Foundation
+
+// sourcery: AutoMockable
+protocol NotificationToneManagerProtocol {
```
stefanceriu, line 11: Please leave some more white spaces, maybe move the actor bound method to the bottom.

stefanceriu, line 11: Let's converge on one single naming scheme, either `NotificationTone` or `NotificationAlertTone`

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/AlertTones/NotificationToneManager.swift
```diff
@@ -0,0 +1,160 @@
...
+    /// Errors that can occur during tone import.
+    enum ImportError: Error {
+        /// The source file could not be accessed due to sandbox restrictions.
+        case couldNotAccessSandboxedResource
+    }
+
+    @globalActor
+    actor ConversionActor {
+        static let shared = ConversionActor()
+    }
```
stefanceriu: Move these to the top of the file please. Also please introduce a `Private` section and reorder methods accordingly.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/AlertTones/NotificationToneManager.swift
```diff
@@ -0,0 +1,160 @@
...
+
+    /// Errors that can occur during audio file conversion.
+    enum ConversionError: Error {
+        /// `AVAudioConverter` could not be initialised for the given format pair.
+        case converterSetupFailed
+        /// A tone with the same filename already exists in the library.
+        case fileAlreadyExists
+        /// An `AVAudioPCMBuffer` could not be allocated.
+        case bufferCreationFailed
+    }
```
stefanceriu: Why do we need so many error types? One single `NotificationToneManagerError` is more than enough for everything in here. Also please move to the top.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/AlertTones/NotificationToneManager.swift
```diff
@@ -0,0 +1,160 @@
...
+//
+
+@preconcurrency import AVFoundation
+import Foundation
+import UniformTypeIdentifiers
+
+/// Manages notification tone selection, import, conversion, and deletion.
+struct NotificationToneManager: NotificationToneManagerProtocol {
+    private let appSettings: AppSettings
+    private let userIndicatorController: UserIndicatorControllerProtocol
```
stefanceriu: Let's not have the tone manager reach up and directly use UI level components.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/AlertTones/NotificationToneManager.swift
```diff
@@ -0,0 +1,160 @@
...
+        do {
+            try? FileManager.default.removeItem(at: NotificationAlertTone.selectedToneLocation)
+            try FileManager.default.copyItem(at: alertTone.location, to: NotificationAlertTone.selectedToneLocation)
+            appSettings.selectedNotificationTone = alertTone
+        } catch {
+            let userIndicator = UserIndicator(type: .toast,
+                                              title: UntranslatedL10n.screenNotificationSettingsSoundSetSoundErrorTitle,
+                                              iconName: "exclamationmark.triangle.fill")
+            userIndicatorController.submitIndicator(userIndicator)
+            MXLog.error("Error setting selected alert tone to designated location in filesystem: \(error)")
```
stefanceriu: Let the user/viewModel decide how ToneManager errors need to be displayed to the user.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/AlertTones/NotificationToneManager.swift
```diff
@@ -0,0 +1,160 @@
...
+        try FileManager.default.createDirectory(at: NotificationAlertTone.selectedToneLocation.deletingLastPathComponent(), withIntermediateDirectories: true)
+    }
+
+    /// Sets the given tone as the active notification alert tone.
+    ///
+    /// Copies the tone's audio file to `selectedToneLocation` and persists the selection in app settings.
+    func setSelectedTone(_ alertTone: NotificationAlertTone) {
+        do {
+            try? FileManager.default.removeItem(at: NotificationAlertTone.selectedToneLocation)
+            try FileManager.default.copyItem(at: alertTone.location, to: NotificationAlertTone.selectedToneLocation)
```
stefanceriu: Why does the NotificationAlertTone define these paths instead of the manager itself?

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/AlertTones/NotificationAlertTone.swift
```diff
@@ -0,0 +1,200 @@
...
+// Copyright 2026 Element Creations Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Foundation
+
+/// Represents a notification alert tone and its backing audio file location.
+struct NotificationAlertTone: Hashable, Comparable, Codable {
```
stefanceriu: This struct feels too complicated. I would move most of it to the manager and let that build simple, data based NotificationAlertTones accordingly.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/AlertTones/NotificationAlertTone.swift
```diff
@@ -0,0 +1,200 @@
...
+        NotificationAlertTone(labelOverride: nil, storageLocationRoot: .appLibrary, relativePath: [filename])
+    }
+
+    #if IS_MAIN_APP // localization is only available in the main app
+    /// The default Element X bundled message tone.
+    static let defaultElementXMessageTone: Self = .createBundledSound(label: UntranslatedL10n.screenNotificationSettingsSoundElementDefault,
+                                                                      filename: "message.caf")
+
+    /// Pre-defined iOS system tones available for selection, sorted by name.
+    static let defaultSystemAlerts: [Self] = [
```
stefanceriu: Aren't these something the manager should provide?

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/NotificationSettingsScreenCoordinator.swift
```diff
@@ -37,10 +38,20 @@ final class NotificationSettingsScreenCoordinator: CoordinatorProtocol {
     
     init(parameters: NotificationSettingsScreenCoordinatorParameters) {
         self.parameters = parameters
-        
+
+        let toneManager: NotificationToneManagerProtocol?
+        do {
+            toneManager = try NotificationToneManager(appSettings: parameters.appSettings, userIndicatorController: parameters.userIndicatorController)
+        } catch {
+            MXLog.error("Catastrophic error setting up tone manager: \(error)")
```
stefanceriu: Then let's fatalError, no?

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/NotificationSettingsScreenViewModel.swift
```diff
@@ -16,26 +16,46 @@ class NotificationSettingsScreenViewModel: NotificationSettingsScreenViewModelTy
...
         self.appSettings = appSettings
         self.userNotificationCenter = userNotificationCenter
         self.notificationSettingsProxy = notificationSettingsProxy
+        self.userIndicatorController = userIndicatorController
         
         let bindings = NotificationSettingsScreenViewStateBindings(enableNotifications: appSettings.enableNotifications)
-        super.init(initialViewState: NotificationSettingsScreenViewState(bindings: bindings, isModallyPresented: isModallyPresented))
-                
+        tonePreviewer = NotificationTonePreviewer(userIndicatorController: userIndicatorController)
+        toneManager = notificationToneManager
```
stefanceriu: Why not keep consistency and have this next to the other constructor injected deps, also name wise?

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/NotificationSettingsScreenViewModel.swift
```diff
@@ -16,26 +16,46 @@ class NotificationSettingsScreenViewModel: NotificationSettingsScreenViewModelTy
...
+         isModallyPresented: Bool) {
         self.appSettings = appSettings
         self.userNotificationCenter = userNotificationCenter
         self.notificationSettingsProxy = notificationSettingsProxy
+        self.userIndicatorController = userIndicatorController
         
         let bindings = NotificationSettingsScreenViewStateBindings(enableNotifications: appSettings.enableNotifications)
-        super.init(initialViewState: NotificationSettingsScreenViewState(bindings: bindings, isModallyPresented: isModallyPresented))
-                
+        tonePreviewer = NotificationTonePreviewer(userIndicatorController: userIndicatorController)
```
stefanceriu: Same for this, why stick it in between the bindings and the constructor call?

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/NotificationSettingsScreenViewModel.swift
```diff
@@ -16,26 +16,46 @@ class NotificationSettingsScreenViewModel: NotificationSettingsScreenViewModelTy
...
+        toneManager = notificationToneManager
+        super.init(initialViewState: NotificationSettingsScreenViewState(bindings: bindings,
+                                                                         isModallyPresented: isModallyPresented,
+                                                                         selectedAlertTone: appSettings.selectedNotificationTone ?? .defaultElementXMessageTone,
```
stefanceriu: Why is `selectedNotificationTone` even optional, can't we have it always default to something?

stefanceriu (changes requested):

ElementX/Sources/Application/Settings/AppSettings.swift
mredig: > Presumably the problem is here is NotificationAlertTone file access.

That is correct.

I've made the change, though I'd push back slightly on the direction. The conditional was intentional - including `NotificationAlertTone` in the Share Extension's compilation target when it has no relevance there sends a misleading signal about the target's responsibilities. A compiler conditional in a settings file feels like a smaller cost than that. Open to discussion if you see it differently.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/AlertTones/NotificationToneManager.swift
mredig: This is an SF Symbol. Is there a preferred way or alternative for these?

ElementX/Resources/Localizations/en.lproj/Untranslated.strings
mredig: I originally had it as "Tone" for user strings to match platform nomenclature, but it was previously flagged as differing from Android where "Sound(s)" was used, matching Android platform nomenclature. It was decided to converge on "Sound(s)".

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/AlertTones/NotificationAlertTone.swift
mredig: I'll need direction on what to do about this, if you want anything other than fixing the comment.

ElementX/Sources/Services/AlertTones/NotificationToneManagerProtocol.swift
mredig: I made some adjustments - let me know if it's what you mean.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/AlertTones/NotificationAlertTone.swift
mredig: Same principle as above, just plural - the type as namespace for its own well-known collections. Moving to the manager as requested.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/AlertTones/NotificationAlertTone.swift
mredig: The intent was to follow the pattern of types like `URL.homeDirectory` - using the type as its own namespace for contextually relevant values. Moving to the manager as requested.

Leaving `systemLocation`, `libraryLocation`, and `bundledLocation` on NotificationTone, though, because it's inherent to the type that it's a location relative to one of these locations. Additionally, if these move to the manager, it will also need to be added to the ShareExtension for NotificationTone to compile.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/AlertTones/NotificationToneManager.swift
mredig: These paths are intrinsic to where tones live, analogous to `Locale.availableIdentifiers` living on `Locale` rather than on a locale manager. Moving to the manager as requested.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/NotificationSettingsScreenCoordinator.swift
mredig: The tone manager is non-critical - the app functions fine without it, just without tone selection. Graceful degradation felt more appropriate than a crash for a feature that's purely additive.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/NotificationSettingsScreenViewModel.swift
mredig: also addressed

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/NotificationSettingsScreenViewModel.swift
mredig: This was an oversight - fixing

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/NotificationSettingsScreenViewModel.swift
mredig: It's optional as a consequence of the graceful fallback in the event that the manager fails to initialize. What would it default to? If it defaulted to a mock (which would be an ick unto itself), it would then render the ui as if the feature was working, but still broken underneath.

ElementX/Sources/Application/Settings/AppSettings.swift
stefanceriu: Yeah, that's fine. The share extension is simple enough for me to chose simplicity in the app settings instead, especially as we're looking into splitting it up. We can worry about the share extension later.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/AlertTones/NotificationToneManager.swift
stefanceriu: > This is an SF Symbol

Yep, I got that but compound icons should be used whenever possible. I do realise the `UserIndicatorToastView` doesn't know about them but perhaps it's better to not use an icon at all. 

Raised a ticket here to address the problem more thoroughly https://github.com/element-hq/element-x-ios/issues/5590

ElementX/Resources/Localizations/en.lproj/Untranslated.strings
stefanceriu: Fair enough, thanks!

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/AlertTones/NotificationAlertTone.swift
stefanceriu: Well we'll add them to Localazy pending https://github.com/element-hq/element-x-ios/pull/5527#discussion_r3234138786, then import them as proper translatables and get rid of this check.

ElementX/Sources/Services/AlertTones/NotificationToneManagerProtocol.swift
stefanceriu: Yup, that works.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/NotificationSettingsScreenCoordinator.swift
stefanceriu: The way I think about these things is that crashes show up in your bug tracker instantly and the developer response is pretty fast. Features that are non-critical but also don't work then to stay unreported for ages and only make the users believe the product is flakey. That's why we fatal error a lot and assert in production.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/NotificationSettingsScreenViewModel.swift
stefanceriu: Then maybe this links up to that fatal error we talked about before. I was thinking it can default to the `defaultElementXMessageTone` like it does here in the VM.

stefanceriu:
This is shaping up quite nicely, good stuff. 

On a more general note: once we're in the reviewing stage please try not to override history, it tends to break comments and make things harder to review as you no longer know what's been changed and what's old. Github reports 41 commits that have been added since I first reviewed this. What I do is use fixup commits or just normal commits that I squash/fixup later, after the PR has been approved and before shipping it. If history isn't particularly important you can even skip that step and squash merge it at the end.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/AlertTones/NotificationToneManager.swift
mredig: I definitely think the icon is preferable, but removed as requested.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/NotificationSettingsScreenCoordinator.swift
mredig: Updated as requested.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/NotificationSettingsScreenViewModel.swift
mredig: Non issue with the fatal error now.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/AlertTones/NotificationTonePreviewer.swift
stefanceriu: You might also want to rename the `notificationTonePreviewer` to just `audioPlayer`.

ElementX/Sources/Services/AlertTones/NotificationToneManager.swift
```diff
@@ -0,0 +1,258 @@
...
+    /// File URL of the active tone copied/linked for use by the system.
+    static let selectedToneLocation = NotificationTone.libraryLocation.deletingLastPathComponent().appending(component: selectedToneFilename)
+    
+    #if IS_MAIN_APP
+    /// The default Element X bundled message tone.
+    static let defaultElementXMessageTone: NotificationTone = .createBundledSound(label: UntranslatedL10n.screenNotificationSettingsSoundElementDefault,
+                                                                                  filename: "message.caf")
+    
+    /// Pre-defined iOS system tones available for selection, sorted by name.
+    static let defaultSystemAlerts: [NotificationTone] = [
```
stefanceriu: I reckon this one can stay here and be made private while the other 2 moved to the top of the file as they're publicly used?

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/AlertTones/NotificationAlertTone.swift
stefanceriu: I appreciate the intention but that doesn't feel at all natural to me. The `NotificationTone` as I understand it is supposed to provide a fileURL and a label.

What business does it have defining system paths, interacting with ProcessInfo and knowing about simulators, accessing bundle locations and defining override labels? 

Isn't it the manager's responsibility to well.. manage.. all of that? You can have different managers define the rules, perhaps depending on the platform, testing scenario etc.

mredig:
> I appreciate the intention but that doesn't feel at all natural to me. The NotificationTone as I understand it is supposed to provide a fileURL and a label.


I've complied with this change. Worth flagging the real costs though, in case you want to reconsider:

- `NotificationTone` can no longer store or vend a `URL` directly. This is a fundamental consequence of sandbox instability - paths change on every app relaunch, so a stored `URL` would be stale on next launch. This can't be worked around by delegating object creation to the manager each time either, since the tone is serialized and persisted in `AppSettings` - any stored URL would be invalid by the next launch.
- As a result, `NotificationTone` loses its natural URL interface. `filename` is now `relativePath.last ?? ""` and `label` falls back to constructing a throwaway `URL` just to call `.deletingPathExtension()`. Both are awkward workarounds for the absence of `.location`.
- The `storageLocationRoot` and `relativePath` properties are now `internal` rather than `private`, since the manager needs to read them. The model's internals are now exposed to satisfy an external dependency.

I'd recommend reverting this. I've isolated the changes to a single commit so it can be easily reverted if you agree.

stefanceriu:
> I've complied with this change

To be honest I wasn't expecting you to directly make the change, I was going for more of a conversation but now that I see it I prefer it this way. 

I need to come in and add the translations plus some other tweaks so I'll think about about it some more while I do that.

This PR is getting uncomfortably long so I'll just merge as it is and we can address anything else separately. Thank you for getting it here!

PR #5530, opened 2026-05-01
Re-enable the AppLock UI tests on iPhone only.

pixlwave:
These were disabled in #5502 but as I understand it, it was only iPad that had an issue, so let's keep running them on iPhone.

Test run here: https://github.com/element-hq/element-x-ios/actions/runs/25217436220


stefanceriu (approved):
Alrighty then 👍

PR #5531, opened 2026-05-01
Show which test suite failed in the UnitTest logs.

pixlwave:
Should help with a little less hunting through the logs.


stefanceriu (approved):
Noice 👌

PR #5542, opened 2026-05-05
Live Location Sharing: Show alert if user has no LLS permissions in the room.

### LINKED ISSUE #5544: [Task] Show alert if user has no permissions for live location sharing

Velin92:
We now check if the user has both the permissions to send `beacon` and `beaconInfo` events, before starting the LLS, in this way if either or both of the two permission are not available, we can show an alert to the user, to make this clear.

fixes #5544

### DIFF
--- a/ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
+++ b/ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
@@ -13,6 +13,7 @@ import MatrixRustSDK
 enum LocationSharingViewAlert: Hashable {
     case missingAuthorization
     case missingAlwaysAuthorization
+    case missingLiveLocationSharingPermission
     case liveLocationDisclaimer
     case liveLocationDurationSelection
     case mapError(MapLibreError)
@@ -213,6 +214,10 @@ extension AlertInfo where T == LocationSharingViewAlert {
                       message: L10n.dialogPermissionLiveLocationDescriptionIos(InfoPlistReader.main.bundleDisplayName),
                       primaryButton: primaryButton,
                       secondaryButton: secondaryButton)
+        case .missingLiveLocationSharingPermission:
+            self.init(id: alertID,
+                      title: L10n.screenShareLocationLiveLocationMissingPermissions,
+                      primaryButton: primaryButton)
         case .liveLocationDisclaimer:
             self.init(id: alertID,
                       title: L10n.screenShareLocationLiveLocationDisclaimerTitle,
--- a/ElementX/Sources/Screens/LocationSharing/LocationSharingScreenViewModel.swift
+++ b/ElementX/Sources/Screens/LocationSharing/LocationSharingScreenViewModel.swift
@@ -68,7 +68,7 @@ class LocationSharingScreenViewModel: LocationSharingScreenViewModelType, Locati
         case .close:
             actionsSubject.send(.close)
         case .startLiveLocation:
-            checkAlwaysShareLocationPermission()
+            startLiveLocation()
         case .selectLocation:
             guard let coordinate = state.bindings.mapCenterLocation else { return }
             let uncertainty = state.isSharingUserLocation ? context.geolocationUncertainty : nil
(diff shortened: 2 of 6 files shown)


ElementX/Sources/Screens/LocationSharing/LocationSharingScreenModels.swift
```diff
@@ -213,6 +214,11 @@ extension AlertInfo where T == LocationSharingViewAlert {
                       message: L10n.dialogPermissionLiveLocationDescriptionIos(InfoPlistReader.main.bundleDisplayName),
                       primaryButton: primaryButton,
                       secondaryButton: secondaryButton)
+        case .missingLiveLocationSharingPermission:
+            self.init(id: alertID,
+                      title: L10n.screenShareLocationLiveLocationMissingPermissions,
+                      message: nil,
```
pixlwave: This is the default value anyway
```suggestion
```

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenViewModel.swift
```diff
@@ -231,6 +235,13 @@ class LocationSharingScreenViewModel: LocationSharingScreenViewModelType, Locati
         }
     }
     
+    private static let durationFormatter: DateComponentsFormatter = {
+        let formatter = DateComponentsFormatter()
+        formatter.unitsStyle = .full
+        formatter.allowedUnits = [.hour, .minute]
+        return formatter
+    }()
```
pixlwave, line 245: Unrelated to this PR specifically, but how come we're using this and not `.formatted(.whatever)` like we do elsewhere?

UnitTests/Sources/LocationSharingScreenViewModelTests.swift
```diff
@@ -467,4 +483,16 @@ struct LocationSharingScreenViewModelTests {
...
+        if let roomInfo = roomProxy.infoPublisher.value as? RoomInfoProxyMock,
+           let powerLevels = roomInfo.powerLevels as? RoomPowerLevelsProxyMock {
+            powerLevels.canOwnUserSendStateEventClosure = { event in
+                event == .beaconInfo ? canShareLiveLocation : false
+            }
+            powerLevels.canOwnUserSendMessageClosure = { messageType in
+                messageType == .beacon ? canShareLiveLocation : true
+            }
+        }
+    }
```
pixlwave: Can we use the existing `RoomPowerLevelsProxyMockConfiguration.powerLevelsConfiguration` instead of adding this? It doesn't look like there would be too much additional work involved and would keep things more reusable.

UnitTests/Sources/LocationSharingScreenViewModelTests.swift
```diff
@@ -172,8 +172,18 @@ struct LocationSharingScreenViewModelTests {
...
+    mutating func startLiveLocationWithoutPermission() {
+        setupViewModel(liveLocationManagerConfiguration: .init(authorizationStatus: .authorizedAlways),
+                       canShareLiveLocation: false)
+        context.send(viewAction: .startLiveLocation)
+        #expect(context.alertInfo?.id == .missingLiveLocationSharingPermission)
+    }
+
     // MARK: - Live Location Authorization Tests
-    
+
```
pixlwave: Don't forget to re-indent this file ;)

pixlwave (approved):
LGTM, only one main thought on the mock configuration, the rest is minor.

ElementX/Sources/Screens/LocationSharing/LocationSharingScreenViewModel.swift
Velin92: There is no way to get the same exact strings we have in the formatter here, using a modified `.formatted` of Duration.
The closest thing is this:
```
Duration.seconds(15 * 60).formatted(.units(allowed: [.hours, .minutes], width: .wide))
// "0 hours, 15 minutes"
```

Velin92: otherwise we would need to apply only minutes for the first and only hours for the latter two... which would work, but this means we would need to treat the three durations separately, and so hardcode the conversion instead of just using the formatter on the value directly... that's fine, is a bit less scalable but if you really prefer this approach we could do that.

PR #5545, opened 2026-05-05
Add an OAuthPresenterHook.

pixlwave:
This PR adds a new hook used within our OAuth presenters.


stefanceriu (approved):
it simple, it gud, I like 👍

PR #5548, opened 2026-05-05
Replace the ServiceLocator with direct dependency injection

mredig:
Summary

Removes the `ServiceLocator` singleton and replaces it with direct dependency injection for better control over instantiation and clearer visibility into how dependencies flow through the app.

## What changed

- Deleted `ServiceLocator.swift`. Its three dependencies (`UserIndicatorController`, `AppSettings`, `AnalyticsService`) are now owned directly as properties on `AppCoordinator` and passed down explicitly wherever they were previously accessed via `ServiceLocator.shared`.
- `TimelineControllerFactory` now receives `AppSettings` via `init` rather than reaching into the service locator.
- All preview/mock call sites use `UserIndicatorControllerMock()` and `AppSettings()` directly.
- Added `AnalyticsService.mock` — a static mock backed by its own `AppSettings()` instance — for use in previews and tests, following the same pattern as other services in the project.
- `Analytics+SwiftUI.swift`'s `@Entry` default is now `nil`; `AppCoordinator.toPresentable()` always injects the real value, so omitting injection simply skips screen tracking.

## Tests

- UITestsAppCoordinator` owns `AppSettings`, `AnalyticsService.mock`, and `UserIndicatorController` as direct properties and injects them explicitly down to each mock screen. `UnitTestsAppCoordinator` and `AccessibilityTestsAppCoordinator` only configure `AppSettings` (suite name + reset); unit tests construct their own analytics and indicator dependencies directly.
- Unit tests that previously registered with `ServiceLocator` now construct their dependencies directly and pass them via `init`.
- Tests that previously relied on `ServiceLocator` setup for `AppSettings` generally called `AppSettings.resetAllSettings()`; that pattern is preserved where applicable.
### Pull Request Checklist

- [ ] Voiceover enabled.


pixlwave:
Oh this is amazing to see the `ServiceLocator` finally being destroyed, thanks for taking this on 🙌.

This isn't quite what we had in mind though, replacing it with a `Dependencies` struct. Could we aim for the following architecture instead please?
- The `{*Tests}AppCoordinator`s own whichever of `AppSettings`, `AnalyticsService` and `UserIndicatorController` they need separately rather than combined in a struct.
    - These types don't particularly have anything to do with each other and in the future we will need at the very least to have an `AnalyticsService` per user session rather than one for the app so it makes sense to break them up now whilst we're working on all of this.
- In the tests use `UserIndicatorControllerMock()` and `AppSettings()` directly like we do in the previews etc.
- Add an `AnalyticsService.mock` to use in the tests like we do for other services/dependencies in the project.
    - For the purposes of analytics within tests this can probably just use its own `AppSettings()` internally as all the settings it reads are either hard-coded or are backed by the user defaults store.

Hopefully that all makes sense, shout if not and we can discuss further.

mredig:
> thanks for taking this on 🙌

🙌!


> Hopefully that all makes sense, shout if not and we can discuss further.

I think it makes sense - basically just decompose the dependencies protocol/struct and instead inject each service individually.

mredig:
Changes made per requests

pixlwave (commented):
Super thanks. Any chance you could tidy up the PR before we do a proper review on it? There are a lot of lines that are simply whitespace changes which add noise to the diffs and make commenting harder.

Note: See https://github.com/element-hq/element-x-ios/pull/5506#discussion_r3189637734 for more info why.

Additionally, I don't think we need 16 commits worth of history here if there's anything you can do about that easily? It's perfectly fine to force push for this given that most of the content has changed since the initial PR.

mredig:
> There are a lot of lines that are simply whitespace changes which add noise to the diffs and make commenting harder.

Can and will do, but just fyi, this exists for exactly this purpose:
image

> Additionally, I don't think we need 16 commits worth of history here if there's anything you can do about that easily? It's perfectly fine to force push for this given that most of the content has changed since the initial PR.

On it.

ElementX/Sources/Screens/CreateRoomScreen/View/CreateRoomScreen.swift
```diff
@@ -456,13 +456,15 @@ struct CreateRoom_Previews: PreviewProvider, TestablePreview {
         let spaces = [SpaceServiceRoom].mockJoinedSpaces2
         clientProxy.spaceService = SpaceServiceProxyMock(.init(editableSpaces: spaces))
         let userSession = UserSessionMock(.init(clientProxy: clientProxy))
-        
+
+        let analytics = AnalyticsService.mock(settings: appSettings)
```
pixlwave: We can probably just inline this one?

ElementX/Sources/Screens/RoomScreen/ComposerToolbar/View/RoomAttachmentPicker.swift
pixlwave, line 1: Not particularly important in this PR, but our preferred pattern here would be to add a `makeViewModel` below the previews  and then keep the top clean with `static let viewModel = makeViewModel(…)`

ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/View/DeveloperOptionsScreen.swift
pixlwave, line 1: Let's inline this too to avoid the need for the closure.

ElementX/Sources/Screens/Spaces/SpaceSettingsScreen/View/SpaceSettingsScreen.swift
pixlwave, line 1: Given there are 2 here, let definitely follow the `makeViewModel` pattern.

ElementX/Sources/AccessibilityTests/AccessibilityTestsAppCoordinator.swift
pixlwave, line 1: Seeing as the AnalyticsService is now optional, personally I don't see any advantage of putting a mock into the Accessibility Tests environment, seeing as it won't do anything anyway.

ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -1087,18 +1088,18 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
             .dropFirst()
             .removeDuplicates()
             .receive(on: DispatchQueue.main)
-            .sink { [weak self] state in
+            .sink { [weak self, userIndicatorController, analyticsService] state in
```
pixlwave: Do we need to capture these directly when we could already get them through `self` like we do for other dependencies like `userSession` and `appMediator`? It seems like it would probably be more sensible to `guard let self` at the start now instead.

ElementX/Sources/Services/Authentication/AuthenticationService.swift
pixlwave, line 1: Purely out of interest can you share what caused these `@MainActor`s to be required now?

ElementX/Sources/UnitTests/UnitTestsAppCoordinator.swift
pixlwave, line 1: Nice 😎

UnitTests/Sources/AnalyticsSettingsScreenViewModelTests.swift
```diff
@@ -11,20 +11,19 @@ import Testing
 
 @MainActor
 final class AnalyticsSettingsScreenViewModelTests {
-    private var appSettings: AppSettings!
+    private let analytics: AnalyticsService
+    private let appSettings: AppSettings
```
pixlwave: Let's keep the order consistent :)

```suggestion
    private let appSettings: AppSettings
    private let analytics: AnalyticsService
```

UnitTests/Sources/AuthenticationServiceTests.swift
```diff
@@ -153,7 +158,7 @@ struct AuthenticationServiceTests {
                                         encryptionKeyProvider: encryptionKeyProvider,
                                         classicAppManager: classicAppManager,
                                         clientFactory: clientFactory,
-                                        appSettings: ServiceLocator.shared.settings,
+                                        appSettings: appSettings,
```
pixlwave: Can we just inline this one and not store them please? – It helps to know whether or not a test is mutating the settings (as in which case we need to include resets in the `init`/`deinit`).

UnitTests/Sources/BlockedUsersScreenViewModelTests.swift
```diff
@@ -13,13 +13,19 @@ import Testing
 
 @MainActor
 struct BlockedUsersScreenViewModelTests {
+    private let userIndicatorController: UserIndicatorControllerProtocol
```
pixlwave: Again let's just inline this one, there's no need to store it here if we don't use it for anything.

UnitTests/Sources/ChatsTabFlowCoordinatorTests.swift
```diff
@@ -19,6 +19,8 @@ struct ChatsTabFlowCoordinatorTests {
     var splitCoordinator: NavigationSplitCoordinator!
     var notificationManager: NotificationManagerMock!
     let stateMachineFactory = PublishedStateMachineFactory()
+    private let appSettings: AppSettings
+    private let analytics: AnalyticsService
```
pixlwave: I'm going to stop repeating the same comment for the unit tests. We can continue to let the coordinators/view models etc own these dependencies rather than adding a reference in the tests that aren't used for changing conditions/confirming expectations during the actual tests.

And any time `appSettings` are used for this we should make sure to reset them on `init`/`deinit`.

pixlwave (commented):
Looks good, no major comments, just suggestions to help follow our existing patterns.

ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/View/DeveloperOptionsScreen.swift
mredig: That would involve instantiating two AppSettings instead of one. Is that your intent?

ElementX/Sources/Screens/Spaces/SpaceSettingsScreen/View/SpaceSettingsScreen.swift
mredig: Please give this a second set of eyes comparing the differences between the two variants. I think the only thing was the members array, but if I missed something, I can fix it.

ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/View/DeveloperOptionsScreen.swift
pixlwave: They’re only previews so it’s perfectly fine from my perspective. In this particular case that second parameter probably shouldn’t even be on the init seeing as it can be read directly from the appSettings anyway.

ElementX/Sources/Services/Authentication/AuthenticationService.swift
mredig: The previous `DependenciesProtocol` iteration required `@MainActor` because one of the provided services had a @MainActor init. It's not strictly required anymore, so I'll remove it.

UnitTests/Sources/AuthenticationServiceTests.swift
mredig: Will do, but that underlying issue is going away in my next PR 😎

UnitTests/Sources/ChatsTabFlowCoordinatorTests.swift
mredig: I think you're asking for me to not store these as properties? I can do a pass doing that - The reason I did this was because it is closer, conceptually, to the old ServiceLocator lifecycle behavior. But I agree that there shouldn't be instance properties on the test suites in general, but should instead have a localized instantiated instance to each individual test.

mredig: Based on my understanding of what you're saying, my plan of attack is:

- leave all the resets/not reset as is
- where app settings gets reset, leave it as a test instance.
- where app settings/analytics is simply passed as a dependency, remove the instance properties

pixlwave: > where app settings gets reset, leave it as a test instance.

I think more specifically for this one I was suggesting the below.

- where app settings are changed or asserted in the tests leave it as a test instance (i.e. it actually has a purpose within the test).

But otherwise yes

> where app settings/analytics is simply passed as a dependency, remove the instance properties

is exactly what I was suggesting.

mredig: I am very low on sleep, but I believe I applied the changes consistently as requested

ElementX/Sources/Mocks/AnalyticsService+Mock.swift
```diff
@@ -0,0 +1,21 @@
+//
+// Copyright 2026 Element Creations Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+extension AnalyticsService {
+    static func mock(_ clientOverride: AnalyticsClientProtocol? = nil, settings: AppSettings? = nil) -> AnalyticsService {
```
stefanceriu, line 9: We might want to later change this `appSettings` instead of plain `settings` as it might get confused with settings for the mock itself.

ElementX/Sources/Mocks/AnalyticsService+Mock.swift
```diff
@@ -0,0 +1,21 @@
+//
+// Copyright 2026 Element Creations Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+extension AnalyticsService {
```
stefanceriu, line 8: At this point it feels awkward to me to not have this service under it's own protocol we can automock and use the same constructs and behaviors as all the other mocks. Did you happen to look into that and decide against it for some reason? 
We wouldn't even need to care about the `AnalyticsClientMock` anymore if we do that. Its only reason for existing is to pass an `isRunning` flag to the service.

mredig: I did not even investigate the possibility. Apart from trying to maintain a minimal surface area in the PR, I also initially found the pattern to be completely valid. Looking closer, I can see advantages to this suggestion, but it would be a relatively much larger PR to touch all the code that ingests the AnalyticsService as well as updating call sites to mock the new service mock instead of the client mock.

I'd be happy to do it, but I'd suggest that goes in a followup PR to keep both of them focused.

stefanceriu: Works for me 👍

stefanceriu:
I think this one's good to go now but there's build errors unfortunately, you'll need to use the `interactionMode: .picker(shouldShowLiveLocationOption: true)` after rebasing.

I would fix them myself but unfortunately can't push to your fork even though the PR has been configured so that `Maintainers are allowed to edit this pull request.` ¯\_(ツ)_/¯ 

```
Pushing to github.com:mredig/element-x-ios.git
error: Authentication error: Authentication required: You must have push access to verify locks
error: failed to push some refs to 'github.com:mredig/element-x-ios.git'
```

mredig:
> there's build errors unfortunately, you'll need to use the `interactionMode: .picker(shouldShowLiveLocationOption: true)` after rebasing.

fixed!

stefanceriu (approved):
So happy to see this land! Nicely done and thank you for making it happen! 🚀

stefanceriu:
P.S. I took the liberty of renaming the PR title (as that's what ends up in the release notes) and rewriting the commit/description while squashing so it matches what's benig delivered. If you find the time please try to update the rest of the PR description so we leave a clean history behind.

PR #5550, opened 2026-05-06
Update SDK to 26.05.06

Velin92:
Name says it all

### DIFF
--- a/ElementX/Sources/Mocks/InvitedRoomProxyMock.swift
+++ b/ElementX/Sources/Mocks/InvitedRoomProxyMock.swift
@@ -84,6 +84,7 @@ private extension RoomMember {
                   powerLevel: proxy.powerLevel.rustPowerLevel,
                   isIgnored: proxy.isIgnored,
                   suggestedRoleForPowerLevel: proxy.role.rustRole,
-                  membershipChangeReason: proxy.membershipChangeReason)
+                  membershipChangeReason: proxy.membershipChangeReason,
+                  isServiceMember: false)
     }
 }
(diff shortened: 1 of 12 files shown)


ElementX/Sources/Mocks/InvitedRoomProxyMock.swift
```diff
@@ -84,6 +84,7 @@ private extension RoomMember {
                   powerLevel: proxy.powerLevel.rustPowerLevel,
                   isIgnored: proxy.isIgnored,
                   suggestedRoleForPowerLevel: proxy.role.rustRole,
-                  membershipChangeReason: proxy.membershipChangeReason)
+                  membershipChangeReason: proxy.membershipChangeReason,
+                  isServiceMember: false)
```
Velin92, line 88: I imagine this is to differentiate between a bot and a normal user, guess we should not care for this as of right now

pixlwave: Maybe we should create an issue to check if we need it anywhere rather than forgetting about it?

pixlwave (approved):
LGTM, but can we handle this as part of the upgrade?

- https://github.com/element-hq/element-x-ios/issues/5538

Velin92:
Did not handle the updated DM check from https://github.com/element-hq/element-x-ios/issues/5538 since the SDK implementation is not fully completed, we still lack the check in the space service and in the NSE.

ElementX/Sources/Mocks/InvitedRoomProxyMock.swift
Velin92: https://github.com/element-hq/element-x-ios/issues/5551

PR #5560, opened 2026-05-07
call: Only group voice/video call if threads developer option is on

BillCarsonFr:
### Pull Request Checklist

Amend https://github.com/element-hq/element-x-ios/pull/5473 that has removed the voice shortcut to make place for another header button that is only present in a rare case of lab + developer option.

Now if the developer option is not available we keep two buttons.

**Default**
image

**With dev option** 

image
image

(dev option)
image

@amshakal I believe this adress your comment here https://github.com/element-hq/element-x-ios/pull/5473#issuecomment-4303729177

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Screens/RoomScreen/View/RoomCallControlsToolbar.swift
+++ b/ElementX/Sources/Screens/RoomScreen/View/RoomCallControlsToolbar.swift
@@ -23,24 +23,45 @@ struct RoomCallControlsToolbar: ToolbarContent {
             }
         } else {
             if viewState.isDirectOneToOneRoom {
-                ToolbarItem(placement: .primaryAction) {
-                    Menu {
-                        Button {
-                            onCallTap(true)
+                if viewState.roomThreadListEnabled {
+                    // If the developer mode room thread list option is enabled there
+                    // is not enough place for 2 calls buttons
+                    ToolbarItem(placement: .primaryAction) {
+                        Menu {
+                            Button {
+                                onCallTap(true)
+                            } label: {
+                                Label(L10n.a11yStartVoiceCall, icon: \.voiceCallSolid)
+                            }
+                                
+                            Button {
+                                onCallTap(false)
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Screens/RoomScreen/View/RoomCallControlsToolbar.swift
```diff
@@ -23,24 +23,45 @@ struct RoomCallControlsToolbar: ToolbarContent {
...
-                            onCallTap(true)
+                if viewState.roomThreadListEnabled {
+                    // If the developer mode room thread list option is enabled there
+                    // is not enough place for 2 calls buttons
+                    ToolbarItem(placement: .primaryAction) {
+                        Menu {
+                            Button {
+                                onCallTap(true)
+                            } label: {
+                                Label(L10n.a11yStartVoiceCall, icon: \.voiceCallSolid)
```
stefanceriu, line 34: ~~Why is the label an accessibility identifier? 🤔~~ Oh sorry, it's not an identifier, got confused by the a11y prefix.

ElementX/Sources/Screens/RoomScreen/View/RoomCallControlsToolbar.swift
stefanceriu, line 81: How come we don't generate previews (and accessibility tests) for this? 🤔

BillCarsonFr: I think I remember having been told that there is a limitation with toolbar stuff / liquid glass? 
But have no details

stefanceriu: I went and found out the hard way, the limitation is that it doesn't work and generates just a one pixel line :))

stefanceriu (approved):
Works for me! 👍

PR #5563, opened 2026-05-07
Link new device with QR code tweaks.

### LINKED ISSUE #5483: [Task] Replace the "sign in was not completed in time" QR code error with an automatic retry.
This:

Image

Should do this:

https://github.com/user-attachments/assets/83ce45ab-1770-48ae-ab80-c7ae9b8e1781

### LINKED ISSUE #5508: Missing "The numbers don’t match" error screen when linking new device
When the `.InvalidCheckCode` error is raised we should show the screen from the the [Figma design](https://www.figma.com/design/jvZ7TIyd7V9Vn8sIDB9vMd/Compound-iOS-Templates?node-id=4759-51026&t=AK2CYUayjoBcjjcy-4). 

Current|Expected
-|-
Image|Image

I think the place the mapping would happen is here:

https://github.com/element-hq/element-x-ios/blob/3a8e6ea0f793acf4fa99002a5dc4cd08173cdd9e/ElementX/Sources/Services/Authentication/LinkNewDeviceService.swift#L198

I note that the strings/translations do exist but they appear to be unused.

pixlwave:
This PR addresses the following 2 issues:
- Fixes https://github.com/element-hq/element-x-ios/issues/5508
- Fixes https://github.com/element-hq/element-x-ios/issues/5483

### DIFF
--- a/ElementX/Sources/Screens/QRCodeLoginScreen/QRCodeLoginScreenModels.swift
+++ b/ElementX/Sources/Screens/QRCodeLoginScreen/QRCodeLoginScreenModels.swift
@@ -39,7 +39,7 @@ enum QRCodeLoginScreenMode {
     /// Configures the screen to link another device by scanning a QR code.
     case linkDesktop(LinkNewDeviceServiceProtocol)
     /// Configures the screen to link another device by showing it a QR code.
-    case linkMobile(LinkNewDeviceService.LinkMobileProgressPublisher)
+    case linkMobile(LinkNewDeviceService.LinkMobileProgressPublisher, ClientProxyProtocol)
 }
 
 struct QRCodeLoginScreenViewState: BindableState {
@@ -107,7 +107,7 @@ enum QRCodeLoginState: Equatable {
     case displayCode(DisplayCodeState)
     
     /// Initial state where the user can link another device using the shown QR code.
-    case displayQR(UIImage)
+    case displayQR(DisplayQRState)
     /// The user needs to enter the two digit code to confirm the channel is secure
     case confirmCode(CheckCodeState)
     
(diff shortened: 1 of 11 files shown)


ElementX/Sources/Screens/QRCodeLoginScreen/QRCodeLoginScreenModels.swift
```diff
@@ -39,7 +39,7 @@ enum QRCodeLoginScreenMode {
     /// Configures the screen to link another device by scanning a QR code.
     case linkDesktop(LinkNewDeviceServiceProtocol)
     /// Configures the screen to link another device by showing it a QR code.
-    case linkMobile(LinkNewDeviceService.LinkMobileProgressPublisher)
+    case linkMobile(LinkNewDeviceService.LinkMobileProgressPublisher, ClientProxyProtocol)
```
stefanceriu, line 42: I would not have expected this to work 😅

ElementX/Sources/Screens/QRCodeLoginScreen/QRCodeLoginScreenViewModel.swift
```diff
@@ -232,6 +232,39 @@ class QRCodeLoginScreenViewModel: QRCodeLoginScreenViewModelType, QRCodeLoginScr
...
+            async let minimumDelay = Task.sleep(for: .seconds(1)) // To show the spinner for long enough so there isn't a flicker.
+            async let qrReadyProgress = progressPublisher.values
+                .first { progress in
+                    switch progress {
+                    case .qrReady: true
+                    default: false
+                    }
+                }
+
+            guard case .qrReady(let image) = try await (qrReadyProgress, minimumDelay).0 else {
```
stefanceriu, line 253: I appreciate the wittiness here but can we maybe use an async algo debounce instead?

stefanceriu (approved):
Nice, lgtm 👍

ElementX/Sources/Screens/QRCodeLoginScreen/QRCodeLoginScreenViewModel.swift
pixlwave: Honestly I think this expresses the intent most clearly. If the tuple syntax isn't favourable I can split this into 2 lines

```
let (progress, _) = try await (qrReadyProgress, minimumDelay)
guard case let .qrReady(image) = …
```

But I don't think that adds much.

PR #5570, opened 2026-05-11
Various UI test fixes

stefanceriu:
- fix settings UI test by making sure the analytics button is visible before attempting to tap it.
- fix start chat UI flow after allowing avatars to be selected from files too
- fix start chat flow UI test snapshots after enabling the room directory by default
- fix room roles and permissions UI test snapshots after adding live location sharing


pixlwave (approved):
LGTM 👍

PR #5571, opened 2026-05-11
Fix an accessibility test failure.

pixlwave:
The test was failing due to the avatar not having a label. I've hidden it as it isn't interactive anyway.


stefanceriu (approved):
Great, thank you! 👏

PR #5574, opened 2026-05-11
New timestamp  rendering system that is layout direction independent

Velin92:
The timestamp trick did play around with layout directions, creating some major issues with RTL or mixed bidi strings. This PR fixes the issue by avoiding adding some random whitespaces and braille spaces alongside direction unicode string. What is does is simply add a `NSTextAtachment` that is rendered as a clear image that matches exactly the size of the timestamp label, so that instead of modifying the bidirectionality of the content to calculate if the line can render or not the timestamp inlined, we just modify the size.
This PR also fixes 3 other major issues we had:
- the last word of the message was always being dragged to a new line if the last word + timestamp did not fit the bubble.
- It fixes the issue where bidi content was not being rendered correctly (like ? being redered at the start of an LTR string in the RTL context, even if it was actually at its end)
- The spacing was a bit higher than normal because we were using fixed size whitespace characters to do the spacing not the actual size of the timestamp label, and this got progressively worse the longer the string was.

It comes with just a small tradeoff:
- when the last line/paragraph is in direction that is different from the layout of the app, we just render the timestamp always on a new line, since there is no actual way to do that without messing around with the directionality of the contentl, which just leads to some inconsistencies. So for example if we are in RTL mode, but the text is only LTR we rendee the timestamp in a new line, viceversa if we have RTL text in LTR mode. However when the app and the text use the same directionality then the timestamp is placed inline if possible.

Review commit by commit


Velin92:
@stefanceriu this might break UI Tests again so feel free to merge that PR first before reviewing this, and I'll generate the UITests again

UnitTests/Sources/ElementCallServiceTests.swift
```diff
@@ -148,13 +148,6 @@ final class ElementCallServiceTests {
             await testClock.advance(by: .seconds(100))
         }
     }
-    
-    @Test
-    func callIntentRawValues() {
-        // Test to ensure that the implicit rawValue of the string enum matches the MSC values
-        #expect(CallIntent.audio.rawValue == "audio")
-        #expect(CallIntent.video.rawValue == "video")
-    }
```
stefanceriu, line 157: What's this about?

ElementX/Sources/Other/Extensions/String.swift
```diff
@@ -101,6 +101,25 @@ extension String {
...
+    var firstStrongCharacterIsRTL: Bool {
+        for scalar in unicodeScalars {
+            let value = scalar.value
+            // Strong RTL: Hebrew, Arabic, Syriac, Thaana, NKo, Samaritan, Mandaic,
+            // Arabic Extended, and their presentation forms.
+            let isStrongRTL = (0x0590...0x08FF).contains(value) ||
+                (0xFB1D...0xFDFF).contains(value) ||
+                (0xFE70...0xFEFF).contains(value)
+            if isStrongRTL { return true }
+            if scalar.properties.isAlphabetic { return false }
```
stefanceriu, line 117: I understand the `isStrongRTL` part but what does `isAlphabetic` do here, early bailout on the first alphabet character it encounters? From what I see `Alphabetic scalars are the primary units of alphabets and/or syllabaries`. Do RTL language scalars not have alphabetic scalars or what?

ElementX/Sources/Other/Pills/MessageText.swift
```diff
@@ -55,17 +55,77 @@ final class MessageTextView: UITextView, PillAttachmentViewProviderDelegate, UIG
     }
 }
 
+/// An `NSTextAttachment` that takes up its declared `bounds` but draws nothing — used
+/// as an invisible spacer so a message bubble's text reserves room for the overlaid
+/// timestamp. Without this subclass, an attachment with no image would render TextKit's
+/// default "missing image" glyph.
```
stefanceriu, line 61: Now that's a cool trick 😎

ElementX/Sources/Other/Pills/MessageText.swift
stefanceriu, line 1: Why so many change whitespaces throughout this file? 🤔

ElementX/Sources/Other/Pills/MessageText.swift
```diff
@@ -55,17 +55,77 @@ final class MessageTextView: UITextView, PillAttachmentViewProviderDelegate, UIG
...
+
+    private func lastParagraphDirectionMatchesLayout(in attributedText: NSAttributedString) -> Bool {
+        let string = attributedText.string as NSString
+        guard string.length > 0 else { return true }
+
+        let lastParagraphRange = string.paragraphRange(for: NSRange(location: string.length - 1, length: 0))
+        let lastParagraph = string.substring(with: lastParagraphRange)
+
+        let textIsRTL = lastParagraph.firstStrongCharacterIsRTL
+        return textIsRTL == (layoutDirection == .rightToLeft)
```
stefanceriu: What's happening here, double double equals? Perhaps we can rewrite it in a way that's less surprising.

ElementX/Sources/Services/Timeline/TimelineItems/EventBasedTimelineItemProtocol.swift
```diff
@@ -64,18 +64,27 @@ extension EventBasedTimelineItemProtocol {
...
-        }
-
-        // To account for the extra spacing created by the status icon
+    /// The inline area to reserve at the trailing edge of the message text so the
+    /// bubble's natural size accommodates the timestamp (which is overlaid on the
+    /// bubble's bottom-trailing corner). Text wraps around this area: if it fits on
+    /// the last line, the timestamp tucks inline; otherwise it sits on a new line.
+    /// Sized using the same dynamic-type-aware font as the rendered timestamp.
+    var trailingReservedSize: CGSize {
+        let textStyle: UIFont.TextStyle = .caption1 // matches compound.bodyXS
```
stefanceriu, line 73: I find it very weird to have things like these on the model layer. Presumably any small missmatch here would results in impredictable behavior in the on the UI layer as the calculations would be ever so slightly different?

ElementX/Sources/Services/Timeline/TimelineItems/EventBasedTimelineItemProtocol.swift
```diff
@@ -64,18 +64,27 @@ extension EventBasedTimelineItemProtocol {
...
+        let font = UIFont.preferredFont(forTextStyle: textStyle)
+        let textSize = (localizedSendInfo as NSString).size(withAttributes: [.font: font])
+        var width = textSize.width
+        var height = textSize.height
         if hasStatusIcon {
-            whiteSpaces += 3
+            // CompoundIcon at .xSmall is 16pt at the base size; scale with the same UIFontMetrics
+            // as the icon's `relativeTo: .compound.bodyXS` modifier, plus the 4pt HStack spacing.
+            let iconSize = UIFontMetrics(forTextStyle: textStyle).scaledValue(for: 16)
```
stefanceriu, line 81: Same as above here. Can we maybe move all of this to the UI layer, even if working with attributed string it's at least going to be closer to the final usage.

ElementX/Sources/Screens/Timeline/View/Style/TimelineStyler.swift
stefanceriu, line 1: What's happening here, why make these changes?

stefanceriu (commented):
This is looking super promising but left some comments inline.

UnitTests/Sources/ElementCallServiceTests.swift
Velin92: I think some stuff that was removed by mistake, will revert it

ElementX/Sources/Other/Pills/MessageText.swift
Velin92: Xcode tool sucks (it's actually very good but...), and does not preserve blank lines whitespace indentation... that's why Doug is working on making a new swiftformat rule for it. Once we have it we will get rid of them, and integrate them in our githook.

ElementX/Sources/Services/Timeline/TimelineItems/EventBasedTimelineItemProtocol.swift
Velin92: I agree, this could be a private extension on the styler view, will move it.

ElementX/Sources/Screens/Timeline/View/Style/TimelineStyler.swift
Velin92: Need to update these you're right, there are still some mismatches.
The hard part is that code is LTR only, so it places ! on the right always, even if the string is LTR since punctuation respects the context. The bubbles will instead get the context from the first characters and use the appropriate directionality they are supposed to have.
Was playing around with these to understand this better, but I guess I just forgot to fix some of the string before pushing

UnitTests/Sources/ElementCallServiceTests.swift
Velin92: Oh wait, checked these have already all been fixed. The code should be 100% related to this PR only.

ElementX/Sources/Other/Extensions/String.swift
Velin92: The RTL block check runs first and returns immediately, so Hebrew/Arabic letters never reach the isAlphabetic line. 
So when isAlphabetic is evaluated, we've already proven the scalar isn't in any of the strong-RTL Unicode blocks.

ElementX/Sources/Services/Timeline/TimelineItems/EventBasedTimelineItemProtocol.swift
Velin92: as agreed in DM, we decided to keep at as it is.

stefanceriu (approved):
Woo, good stuff 👌

PR #5575, opened 2026-05-12
Add media upload preview cropping capability for images

stefanceriu:
Start using [Mantis](https://github.com/guoyingtao/Mantis) to crop and rotate images before upload.

https://github.com/user-attachments/assets/e573e7bf-4ab5-4cf7-8c09-31493a657d12

### DIFF
--- a/ElementX/Sources/Screens/MediaUploadPreviewScreen/MediaUploadPreviewScreenModels.swift
+++ b/ElementX/Sources/Screens/MediaUploadPreviewScreen/MediaUploadPreviewScreenModels.swift
@@ -7,6 +7,7 @@
 //
 
 import Foundation
+import UIKit
 
 enum MediaUploadPreviewScreenViewModelAction {
     case dismiss
@@ -42,5 +43,5 @@ enum MediaUploadPreviewAlertType: Hashable {
 enum MediaUploadPreviewScreenViewAction {
     case send
     case cancel
-    case editedMedia(index: Int)
+    case editedMedia(image: UIImage, index: Int)
 }
--- a/ElementX/Sources/Screens/MediaUploadPreviewScreen/MediaUploadPreviewScreenModels.swift
+++ b/ElementX/Sources/Screens/MediaUploadPreviewScreen/MediaUploadPreviewScreenModels.swift
@@ -19,6 +19,8 @@ struct MediaUploadPreviewScreenViewState: BindableState {
     let isRoomEncrypted: Bool
     var shouldDisableInteraction = false
     
+    var mediaEditVersion = 0
+
     var bindings = MediaUploadPreviewScreenBindings()
 }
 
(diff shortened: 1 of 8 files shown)


ElementX/Sources/Screens/MediaUploadPreviewScreen/View/MediaUploadPreviewScreen.swift
```diff
@@ -59,6 +61,15 @@ struct MediaUploadPreviewScreen: View {
             .preferredColorScheme(colorSchemeOverride)
             .onAppear(perform: focusComposerIfHardwareKeyboardConnected)
             .alert(item: $context.alertInfo)
```
Velin92, line 63: beware because if an alert is presented while another sheet is presented, it won't be displayed (learnt the hard way while working on LLS), can any of these alert be displayed while the editing flow is happening? If not, no need to worry about

stefanceriu: Oh very good point, thanks! I put another alert on the sheet itself which seems to do the right thing whenever there's an out of bound alert.

ElementX/Sources/Screens/MediaUploadPreviewScreen/View/MediaUploadPreviewScreen.swift
```diff
@@ -149,6 +163,46 @@ struct MediaUploadPreviewScreen: View {
...
+    private func saveAndApplyCroppedImage(_ image: UIImage, at index: Int) {
+        guard let data = image.jpegData(compressionQuality: 1.0) else {
+            return
+        }
+        
+        let originalURL = context.viewState.mediaURLs[index]
+        
+        do {
+            try data.write(to: originalURL, options: .atomic)
+            context.send(viewAction: .editedMedia(index: index))
```
Velin92: Couldn't we have the data action be handled on the view model? feels like this should be handled there.

ElementX/Sources/Screens/MediaUploadPreviewScreen/MediaUploadPreviewScreenModels.swift
```diff
@@ -19,6 +19,8 @@ struct MediaUploadPreviewScreenViewState: BindableState {
     let isRoomEncrypted: Bool
     var shouldDisableInteraction = false
     
+    var mediaEditVersion = 0
```
Velin92, line 23: I imagine this is used to update the view after an edit has taken place, but could we add a comment to this for clarity?
Also why this value couldn't just be a `@State` var on the view itself?

ElementX/Sources/Screens/MediaUploadPreviewScreen/View/MediaUploadPreviewScreen.swift
stefanceriu: Fair, I'll move it 👍

Velin92 (approved):
LGTM just a few comments

ElementX/Sources/Screens/MediaUploadPreviewScreen/MediaUploadPreviewScreenModels.swift
stefanceriu: > I imagine this is used to update the view after an edit has taken place

Yep, exactly, because there's no direct link to the underlying QLPreview. It's in the models so that the view model can ask for it manually, which perhaps wasn't particularly useful before but now is now that we're moving the file writing to it.

PR #5576, opened 2026-05-12
Prevent replies to live location timeline items, and live location sharing in threads

### LINKED ISSUE #5572: [Task] EX LLS: reply and thread

Velin92:
fixes #5572 

This is due to a limitation that prevents the SDK to reply (or show in a thread) state events, and since the LLS timeline item is an aggregation happening on the beacon info state event, this creates a lot of issues if you try to reply to one, or if you try to send one as a reply or in a thread, so we're fully preventing it.


stefanceriu (approved):
lgtm 👍

PR #5580, opened 2026-05-13
Fix for cropped boosted emoji

### LINKED ISSUE #5579: Large emojis get cut off
### Steps to reproduce

In the app, when you send a small number of emojis without any text, the emojis will be displayed larger than usual.

#### What did you expect?
When sending emojis as described above, I would expect them to be displayed in full.

#### What happened instead?
The emoji gets cut off at approximately the size it would be if it was the usual size that would happen if you send more than 5 emojis or if you send text along with it in the same message:

Image

### Your phone model

iPhone 12 mini

### Operating system version

iOs 26.4.2

### Application version

26.05.1 (228)

### Homeserver

synapse

### Will you send logs?

No

Velin92:
essentially we just apply the last character font size to the attachment
fixes #5579


stefanceriu (commented):
🤞

stefanceriu (approved):

PR #5581, opened 2026-05-13
Bump the RustSDK to v26.05.13

stefanceriu:
Not much to it.


Velin92 (approved):

Velin92:
does this include the LLS replied to fix? In case it does, could you replace the `false` for the `canBeRepliedTo` paremter in the builder, to use back again the eventItemProxy value?

PR #5583, opened 2026-05-13
Use isDM from the SDK

### LINKED ISSUE #5538: [Task] Use `DmRoomDefinition` and `RoomInfo.isDm`

### LINKED ISSUE #5551: [Task] Include isServiceMember in the RoomMemberProxy
Could be useful if in the future we need to differentiate how we render a normal user from a service member (such as a bot).
We just need to include `isServiceMember` in our mocks and in the RoomMemberProxy, which comes from the `RoomMember` SDK object.

Velin92:
This changes the way the JoinedRoom, the SpaceService room and the NotificationItem determine if a room is a DM, by using the isDM check directly from the SDK instead of determine it client side.

This also includes using the dm specification in the client builder (otherwise it defaults to the matrix spec which considers a dm any simple direct room instead for also checking the number of active non service members).

I also added an isServiceMember var to the RoomMemberProxy but as of right now it's not used anywhere, but might be useful in the future.

fixes #5538 
fixes #5551

### DIFF
--- a/ElementX/Sources/Mocks/JoinedRoomProxyMock.swift
+++ b/ElementX/Sources/Mocks/JoinedRoomProxyMock.swift
@@ -182,6 +182,7 @@ extension RoomInfoProxyMock {
         activeMembersCount = configuration.members.filter { $0.membership == .join || $0.membership == .invite }.count
         invitedMembersCount = configuration.members.filter { $0.membership == .invite }.count
         joinedMembersCount = configuration.members.filter { $0.membership == .join }.count
+        isDM = isDirect && activeMembersCount <= 2
         highlightCount = 0
         notificationCount = 0
         cachedUserDefinedNotificationMode = .allMessages
--- a/ElementX/Sources/Mocks/NotificationItemProxyMock.swift
+++ b/ElementX/Sources/Mocks/NotificationItemProxyMock.swift
@@ -52,5 +52,6 @@ extension NotificationItemProxyMock {
         underlyingIsNoisy = configuration.isNoisy
         underlyingHasMention = configuration.hasMention
         threadRootEventID = configuration.threadRootEventID
+        underlyingIsDM = configuration.isRoomDirect && configuration.roomJoinedMembers <= 2
     }
 }
--- a/ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenViewModel.swift
+++ b/ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenViewModel.swift
@@ -228,7 +228,7 @@ class JoinRoomScreenViewModel: JoinRoomScreenViewModelType, JoinRoomScreenViewMo
         if case .space(let spaceServiceRoom) = source {
             switch spaceServiceRoom.state {
             case .invited:
-                state.mode = .invited(isDM: spaceServiceRoom.isDirect == true && spaceServiceRoom.joinedMembersCount == 1)
+                state.mode = .invited(isDM: spaceServiceRoom.isDM == true)
             case .knocked:
                 state.mode = .knocked
             case .banned:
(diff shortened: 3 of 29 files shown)


ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
```diff
@@ -420,6 +420,7 @@ class TimelineViewModel: TimelineViewModelType, TimelineViewModelProtocol {
     
     private func updateRoomInfo(_ roomInfo: RoomInfoProxyProtocol) {
         state.pinnedEventIDs = roomInfo.pinnedEventIDs
+        state.isDirectOneToOneRoom = roomInfo.isDM
```
stefanceriu: Should we maybe rename `isDirectOneToOneRoom` to `isDM` everywhere at this point? 🤔

ElementX/Sources/Screens/JoinRoomScreen/JoinRoomScreenViewModel.swift
```diff
@@ -228,7 +228,7 @@ class JoinRoomScreenViewModel: JoinRoomScreenViewModelType, JoinRoomScreenViewMo
         if case .space(let spaceServiceRoom) = source {
             switch spaceServiceRoom.state {
             case .invited:
-                state.mode = .invited(isDM: spaceServiceRoom.isDirect == true && spaceServiceRoom.joinedMembersCount == 1)
+                state.mode = .invited(isDM: spaceServiceRoom.isDM == true)
```
stefanceriu, line 231: So much better!

ElementX/Sources/Mocks/NotificationItemProxyMock.swift
```diff
@@ -52,5 +52,6 @@ extension NotificationItemProxyMock {
         underlyingIsNoisy = configuration.isNoisy
         underlyingHasMention = configuration.hasMention
         threadRootEventID = configuration.threadRootEventID
+        underlyingIsDM = configuration.isRoomDirect && configuration.roomJoinedMembers <= 2
```
stefanceriu, line 55: Random q: when am I supposed to set the `underlying` and when can I just use the normal property? 🤔

ElementX/Sources/Mocks/JoinedRoomProxyMock.swift
```diff
@@ -182,6 +182,7 @@ extension RoomInfoProxyMock {
         activeMembersCount = configuration.members.filter { $0.membership == .join || $0.membership == .invite }.count
         invitedMembersCount = configuration.members.filter { $0.membership == .invite }.count
         joinedMembersCount = configuration.members.filter { $0.membership == .join }.count
+        isDM = isDirect && activeMembersCount <= 2
```
stefanceriu, line 185: Hmm, shouldn't `isDirect` be the only thing controlling this? I wouldn't expect it to not be direct because I didn't know I had to set active members on the mock.

ElementX/Sources/Mocks/NotificationItemProxyMock.swift
Velin92: they are both fine to use! the underlying is always optional, while the same value named property is force unwrapped so it is consistent with the protocol definition (if such was not optional). ideally underlying makes more sense to use however imo, but doesn't matter too much especially when the mocked value is already optional, like below

ElementX/Sources/Mocks/JoinedRoomProxyMock.swift
Velin92: since before all our code was working on the protocol extension that has been deleted, all of our tests did assume this already. 

https://github.com/element-hq/element-x-ios/commit/c2357cfe8ee2043e2045b1c2138f0365242696b7#diff-97ab3c6ab477005c96401723176f61a3ac1b73033693067fa5461dfcfc418a52R224-L227

so this was always used essentially everytime this check was used since it was a protocol extension.
And all of our tests are coded assuming that isDirect && activeMembersCount <= 2.
Can change this to use a isDM flag directly but is a bit more work.

stefanceriu: Riiight, okay then, leave it alone I guess

stefanceriu (approved):
🙌

PR #5584, opened 2026-05-13
Fix avatar not updating on existing timeline messages

mredig:
Fixes the same issue tracked on Android in element-hq/element-x-android#5176.

### Problem

When a room member changes their avatar, existing messages they sent do not update to show the new avatar. A "profile changed" event appears at the bottom of the timeline, but scrolling up reveals all previous messages still showing the old avatar. Backing out and re-entering the room shows the correct avatar.

The root cause is in the Rust SDK: when a profile change event arrives via sync, the SDK correctly appends a new timeline item for the change, but does not re-emit existing timeline items with updated sender profile data. Ideally the SDK would emit timeline items with only a sender ID and the app would resolve the current profile at render time from a single source of truth -- avoiding this class of staleness entirely.

### Fix

Two changes work together to address this:

**1. `TimelineItemProvider` now publishes on `roomMemberEventPublisher` for profile changes**

Previously `isRoomMemberEvent` (formerly `isMembershipChange`) only evaluated true for join/leave/ban events. Profile changes (`profileChange` content) are also `m.room.member` events per the Matrix spec and now correctly evaluate true, triggering the publisher.

**2. `TimelineSenderAvatarView` reads from the live member cache**

The avatar URL is now resolved from `context.viewState.members` (the live member cache already maintained by `TimelineViewModel`) rather than the stale snapshot baked into the timeline item at creation time, with the snapshot kept as a fallback. This effectively treats the timeline item sender as no more than an identifier, as suggested earlier.

`TimelineController` subscribes to `roomMemberEventPublisher` and calls `roomProxy.updateMembers()` when it fires, refreshing the cache and triggering a view update.

### Note

`@EnvironmentObject` is used in `TimelineSenderAvatarView` instead of `@Environment(\.timelineContext)` to ensure SwiftUI's observation tracking picks up changes to `viewState.members`. The `environmentObject` is always injected alongside the environment key at the call site in `TimelineTableViewController`.

### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Screens/Timeline/View/TimelineSenderAvatarView.swift
+++ b/ElementX/Sources/Screens/Timeline/View/TimelineSenderAvatarView.swift
@@ -10,16 +10,16 @@ import Foundation
 import SwiftUI
 
 struct TimelineSenderAvatarView: View {
-    @Environment(\.timelineContext) private var context
-
+    @EnvironmentObject private var context: TimelineViewModel.Context
+    
     let timelineItem: EventBasedTimelineItemProtocol
-        
+    
     var body: some View {
-        LoadableAvatarImage(url: timelineItem.sender.avatarURL,
+        LoadableAvatarImage(url: context.viewState.members[timelineItem.sender.id]?.avatarURL ?? timelineItem.sender.avatarURL,
                             name: timelineItem.sender.displayName,
                             contentID: timelineItem.sender.id,
                             avatarSize: .user(on: .timeline),
-                            mediaProvider: context?.mediaProvider)
+                            mediaProvider: context.mediaProvider)
             .overlay {
                 Circle().stroke(Color.compound.bgCanvasDefault, lineWidth: 3)
             }
--- a/ElementX/Sources/Services/Timeline/TimelineItemProviderProtocol.swift
+++ b/ElementX/Sources/Services/Timeline/TimelineItemProviderProtocol.swift
@@ -31,10 +31,10 @@ protocol TimelineItemProviderProtocol {
     var paginationState: TimelinePaginationState { get }
     /// The kind of the timeline
     var kind: TimelineKind { get }
-    /// A publisher that signals when changes to the room's membership have occurred through `/sync`.
+    /// A publisher that signals when member changes have occurred in the room through `/sync`.
     ///
     /// This is temporary and will be replace by a subscription on the room itself.
-    var membershipChangePublisher: AnyPublisher<Void, Never> { get }
+    var roomMemberEventPublisher: AnyPublisher<Void, Never> { get }
 }
 
 // sourcery: AutoMockable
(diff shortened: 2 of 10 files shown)


ElementX/Sources/Services/Timeline/TimelineItemProviderProtocol.swift
```diff
@@ -31,10 +31,10 @@ protocol TimelineItemProviderProtocol {
     var paginationState: TimelinePaginationState { get }
     /// The kind of the timeline
     var kind: TimelineKind { get }
-    /// A publisher that signals when changes to the room's membership have occurred through `/sync`.
+    /// A publisher that signals when member changes have occurred in the room through `/sync`.
     ///
     /// This is temporary and will be replace by a subscription on the room itself.
```
mredig, line 36: This was written in 2024 - I'm assuming this ship has sailed?

ElementX/Sources/Screens/Timeline/View/TimelineSenderAvatarView.swift
```diff
@@ -10,16 +10,16 @@ import Foundation
 import SwiftUI
 
 struct TimelineSenderAvatarView: View {
-    @Environment(\.timelineContext) private var context
-
+    @EnvironmentObject private var context: TimelineViewModel.Context
+
```
Velin92, line 14: We can't have this change in the codebase, we actually already had this implementation in the past and we ended up going for the `@Environment` solution. The reason is that by making this an `EnvironmentObject` every timeline item will observe the changes of any other timeline item, causing multiple UI updates that slowed down the timeline.
Instead this allows us to inject the timeline contect in a non observable way, but to use then the     `RoomTimelineItemViewState`  as an `ObservedObject` (which is unique per timeline item) to only observe changes for that specific timeline item, so that the updates only happen for that timeline item when necessary.
However this is on the avatar so I am not sure what effect this might have but easy enough to test. If you send a new messae, where only 1 new avatar is displayed does this get called multiple times or just once?

ElementX/Sources/Screens/Timeline/View/TimelineSenderAvatarView.swift
```diff
@@ -10,16 +10,16 @@ import Foundation
...
-    @Environment(\.timelineContext) private var context
-
+    @EnvironmentObject private var context: TimelineViewModel.Context
+    
     let timelineItem: EventBasedTimelineItemProtocol
-        
+    
     var body: some View {
-        LoadableAvatarImage(url: timelineItem.sender.avatarURL,
+        LoadableAvatarImage(url: context.viewState.members[timelineItem.sender.id]?.avatarURL ?? timelineItem.sender.avatarURL,
```
Velin92, line 18: for curiosity sake does this also happen on Android?
The thing is that I think this issue is more of an SDK one, because the timeline item from the SDK contains the sender and its avatar, and this value should indeed be updated once a profile changes internally. We should not rely on the members directly to get the avatar, but on the value provided by timeline item itself. I think this is something that was probably recently broken with the changes made by the event cache.

Velin92:
I have a feeling this is an SDK issue, that is not correctly updating anymore the sender item from a timeline item once it gets changed, I'll check with the SDK team if they can tell me more about it.
Does it also happen on android?
Also because i just checked on develop that if I change my display name, all the timeline items get their display name updated nicely as soon as the event is received, so the bug it's definitely not related to having to listen to the `.profileChange` event, and the way the timeline is supposed to work, we should always rely on using the avatar from the sender directly if possible.

Velin92:
Okay we found the issue in the SDK: https://github.com/matrix-org/matrix-rust-sdk/issues/6575
We are looking if this is easily fixable or it would be better to have this client side fix in the meantime while we find a more robust longterm solution.

Otherwise if you would like to try to solve the issue on the sdk side, that would benefit both platforms at the same time.

Velin92:
This will be fixed with: https://github.com/matrix-org/matrix-rust-sdk/pull/6576 thanks for the contribution however, this helped us a lot in finding the bug.

PR #5589, opened 2026-05-15
AppSettings are Mockable

mredig:
Summary

Makes `AppSettings` mockable by introducing `UserDefaultsProtocol` and `UserDefaultsMock`, eliminating the need for tests and previews to clobber a shared `UserDefaults` suite name and manually reset state between runs.

## Approach

1. Introduced `UserDefaultsProtocol` and replaced all concrete `UserDefaults` usage with it. `UserDefaults` conforms via an extension.
2. Moved static methods/properties on `AppSettings` to instance level. Removed ones that are no longer relevant -- e.g. `configureWithSuiteName` is superseded by passing a `UserDefaultsProtocol` instance directly to `AppSettings.init`.
3. Refactored `AppSettings.UserDefaultsKey` from a private enum to a non-private `RawRepresentable` struct. The key space isn't exhaustive, so an enum isn't appropriate. As a struct it's extensible (e.g. tests can define their own keys) while retaining dot syntax.
4. Because property wrappers don't have access to the enclosing `self`, all `UserPreference` instances are initialized explicitly in `AppSettings.init`. The Apple-private `_enclosingInstance` mechanism works but is too risky for production code. The proper future path would be an `@attached(accessor)` macro, but that's a significantly heavier lift.
5. Removed `StorageType` and volatile storage -- the only uses were in tests, and `UserDefaultsMock` is inherently in-memory, making explicit volatile storage redundant. If per-property volatile storage is needed in future, a fresh `UserDefaultsMock` can be passed for that specific property.
6. Updated all call sites.

### Pull Request Checklist

- [ ] Voiceover enabled.


ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -107,24 +117,14 @@ final class AppSettings {
         #endif
     }
     
-    static func resetAllSettings() {
+    func resetAllSettings() {
         MXLog.warning("Resetting the AppSettings.")
-        store.removePersistentDomain(forName: suiteName)
+        store.removePersistentDomain(forName: Self.suiteName)
```
stefanceriu: With this now being an instance method what happens if this instance's store uses a different suite?

ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -411,45 +411,91 @@ final class AppSettings {
...
+        _enableOnlySignedDeviceIsolationMode = UserPreference(key: .enableOnlySignedDeviceIsolationMode, defaultValue: false, storage: store)
+        _knockingEnabled = UserPreference(key: .knockingEnabled, defaultValue: false, storage: store)
+        _threadsEnabled = UserPreference(key: .threadsEnabled, defaultValue: false, storage: store)
+        _roomThreadListEnabled = UserPreference(key: .roomThreadListEnabled, defaultValue: false, storage: store)
+        _focusEventOnNotificationTap = UserPreference(key: .focusEventOnNotificationTap, defaultValue: false, storage: store)
+        _linkPreviewsEnabled = UserPreference(key: .linkPreviewsEnabled, defaultValue: false, storage: store)
+        _linkNewDeviceEnabled = UserPreference(key: .linkNewDeviceEnabled, defaultValue: false, storage: store)
+        _automaticBackPaginationEnabled = UserPreference(key: .automaticBackPaginationEnabled, defaultValue: false, storage: store)
+        _developerOptionsEnabled = UserPreference(key: .developerOptionsEnabled, defaultValue: Self.appBuildType != .release, storage: store)
```
stefanceriu, line 499: I find this to be a step back from what we had in terms of ergonomy, where the definition of a setting also contained the default value and volatility.

ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -411,45 +411,91 @@ final class AppSettings {
...
+    var developerOptionsEnabled: Bool
+	
+    init(store: UserDefaultsProtocol? = nil) {
+        // UserDefaults to be used on reads and writes.
+        guard
+            let store: UserDefaultsProtocol = store ?? UserDefaults(suiteName: Self.suiteName)
+        else { fatalError("Catastrophic error - UserDefaults could not be instantiated with suite name: \(Self.suiteName)") }
+        self.store = store
+		
+        _lastVersionLaunched = UserPreference(key: .lastVersionLaunched, storage: store)
```
stefanceriu, line 460: How does setting volatility work with this new version?

ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -411,45 +411,91 @@ final class AppSettings {
...
+    
+    @UserPreference
+    var developerOptionsEnabled: Bool
+	
+    init(store: UserDefaultsProtocol? = nil) {
+        // UserDefaults to be used on reads and writes.
+        guard
+            let store: UserDefaultsProtocol = store ?? UserDefaults(suiteName: Self.suiteName)
+        else { fatalError("Catastrophic error - UserDefaults could not be instantiated with suite name: \(Self.suiteName)") }
+        self.store = store
```
stefanceriu, line 458: I would much rather we make this mandatory and passed from the outside instead of hidden inside here. The only real user is the AppCoordinator which configures everything already.

ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -35,64 +35,74 @@ enum AppBuildType {
...
 /// Store Element specific app settings.
 final class AppSettings {
-    private enum UserDefaultsKeys: String {
-        case lastVersionLaunched
-        case seenInvites
-        case hasSeenNewSoundBanner
-        case appLockNumberOfPINAttempts
-        case appLockNumberOfBiometricAttempts
-        case timelineStyle
+    struct UserDefaultsKey: RawRepresentable, ExpressibleByStringLiteral {
```
stefanceriu: This is a massive change, what's the reason behind it? 🤔 The commit only says `AppSettings.UserDefaultsKeys is now a RawRep struct, making it extensible`, but why, why do we need that?

stefanceriu (commented):
Drive by review, I have some questions..

ElementX/Sources/Application/Settings/AppSettings.swift
mredig: Good catch -- `removePersistentDomain(forName:)` passing `Self.suiteName` is indeed a bug when the backing store isn't the standard suite. Unfortunately `UserDefaults` doesn't expose its domain names publicly (the API for that is deprecated and the docs explicitly say to track domains yourself), which puts us in an awkward spot given the abstraction goal here.

A few options I can think of:
- Iterate over all known keys and remove them individually -- not elegant but effective
- Subclass `UserDefaults` and conform the subclass to the protocol instead of `UserDefaults` directly. The subclass can track its own suite name and expose a proper reset
- Extend the protocol and mock to have an explicit concept of suite/domain tracking, then surface that through `AppSettings`

Do you have a preference?

ElementX/Sources/Application/Settings/AppSettings.swift
mredig: Two points here: first, volatile storage wasn't used anywhere in the existing open source codebase except in tests. Second, if per-property volatile storage is still needed, it's still achievable by passing a fresh `UserDefaultsMock()` for that specific property -- `UserPreferenceTests` already demonstrates this pattern.

ElementX/Sources/Application/Settings/AppSettings.swift
mredig: Happy to make it non-optional -- will update.

ElementX/Sources/Application/Settings/AppSettings.swift
mredig: The PR description covers the rationale for this in point 5 -- the root cause is that property wrappers can't access the enclosing `self`, so with a per-instance store there's no way to wire up the storage at the declaration site. The previous code worked because the store was static. An `@attached(accessor)` macro is the proper long-term solution but a significantly heavier lift for this PR.

ElementX/Sources/Application/Settings/AppSettings.swift
mredig: The change serves a few concrete purposes: it provides semantic typing instead of raw strings, enables dot syntax at call sites (particularly useful in the `init`), and gives `UserPreference` a typed non-string key parameter rather than an arbitrary `String`. Most importantly it makes the key space extensible -- tests already take advantage of this by defining their own keys, and the enterprise module could do the same without touching the core type.

ElementX/Sources/Application/Settings/AppSettings.swift
Velin92: Sometimes we do use volatility for specific FF that only should live per session, and while using UserDefaultsMock would be fine, I don't really like semantically the idea of using a mock type in production code, but would be fine to use a volatile type inside a test, so maybe we should just have the mock be called volatile, and just have volatile be used in tests?

ElementX/Sources/Application/Settings/UserPreference.swift
```diff
@@ -139,9 +119,9 @@ protocol KeyedStorage<Value> {
 /// When used with a `Value` that conforms to `PlistRepresentable` the Codable encode/decode
 /// phase is skipped, and values are stored natively in the plist.
 final class UserDefaultsStorage<Value: Codable>: KeyedStorage {
-    private let userDefaults: UserDefaults
+    private let userDefaults: UserDefaultsProtocol
     
-    init(userDefaults: UserDefaults) {
+    init(userDefaults: UserDefaultsProtocol) {
         self.userDefaults = userDefaults
     }
```
Velin92, line 130: for brevity sake, let's also add a static mock() function/var that essentially does: appSettings(store: UserDefaultsMock()), because I see is going to be used a LOT. Or even better let's call it `.volatile`

ElementX/Sources/Mocks/UserDefaultsMock.swift
```diff
@@ -0,0 +1,45 @@
+//
+// Copyright 2026 Element Creations Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Foundation
+
+public final class UserDefaultsMock: UserDefaultsProtocol, Sendable {
```
Velin92: As I said in one of the previous comments if we want to have a volatile storage for a property in production code (which is a thing that rarely happens but it does happen), then I would rather have this be called `VolatileDefaults` and be used both in production if we need a volatile storage and in tests. I just don't like the idea of using something that is called mock in production.

ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -35,64 +35,74 @@ enum AppBuildType {
 
 /// Store Element specific app settings.
 final class AppSettings {
```
Velin92: If we can make this `Sendable` I think we can have solve one real issue that plagues Swift 6 migration. Doesn't need to be done now though, so don't bother for now.

ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -35,64 +35,74 @@ enum AppBuildType {
...
+            self.rawValue = rawValue
+        }
+        
+        init(stringLiteral value: StringLiteralType) {
+            self.init(rawValue: value)
+        }
     }
     
-    private static var suiteName: String = InfoPlistReader.main.appGroupIdentifier
+    private static let suiteName: String = InfoPlistReader.main.appGroupIdentifier
```
Velin92: this is good making this static let will make easier to use `Sendability`

ElementX/Sources/Application/Settings/AppSettings.swift
Velin92: This is already a big change, I see the point of making the keys extensible, but I don't see its purpose since, we then can't really extend the AppSettings to accomodate for the additional keys. Since this change doesn't really impact, could we revert it? I think the enum solution just looks more readable.

ElementX/Sources/Application/Settings/AppSettings.swift
Velin92: Actually if we can at least try to mark this as `Sendable` and if everything compiles fine that would be great (but I think since there are variable inside, we would probably need make the settings a struct), since each instance of the settings  will still acess the same underlying store that should be fine, but I wonder how that would work with volatile storage/mocks since we want them to be shared across different copies of the same instance.
As I said if it proves too complicated don't bother for now, but we definitely need the settings to be Sendable at one point.

ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -35,64 +35,74 @@ enum AppBuildType {
...
 
     /// UserDefaults to be used on reads and writes.
-    private static var store: UserDefaults! = UserDefaults(suiteName: suiteName)
+    private let store: UserDefaultsProtocol
```
Velin92, line 100: even better with this not being static

ElementX/Sources/Application/Settings/AppSettings.swift
mredig: I agree - I'll do that change.

mredig: ~~I don't have a better location for it than the `Mocks` folder, but I'm open to suggestions.~~ Nvm I just moved to the Settings folder

ElementX/Sources/Application/Settings/AppSettings.swift
mredig: I ended up subclassing UserDefaults and it appears to be performing well.

ElementX/Sources/Application/Settings/AppSettings.swift
mredig: Marking as Sendable has some cascading consequences. I definitely want to do this, but I think it deserves its own PR.

Velin92: absolutely let's align on this since I am also kinda working on it on my Swift 6 migration project

ElementX/Sources/Application/Settings/TrackedUserDefaults.swift
```diff
@@ -0,0 +1,50 @@
+//
+// Copyright 2026 Element Creations Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Foundation
+
+class TrackedUserDefaults: UserDefaults, UserDefaultsProtocol {
```
Velin92: let's put final when we define a new class we are not yet subclassing

ElementX/Sources/AccessibilityTests/AccessibilityTestsAppCoordinator.swift
```diff
@@ -40,8 +40,7 @@ class AccessibilityTestsAppCoordinator: AppCoordinatorProtocol {
         
         MXLog.configure(currentTarget: "accessibility-tests")
         
-        AppSettings.configureWithSuiteName("io.element.elementx.accessibilitytests")
-        AppSettings.resetAllSettings()
+        let appSettings = AppSettings.volatile()
```
Velin92, line 43: nice well done!

Velin92 (approved):
LGTM, just use final when defining a new class, also could you resolve the conflicts and make tests pass? When everything is green I'll merge it

PR #5594, opened 2026-05-15
Render and send gallery messages with multi-attachment support

svetoslav-sportinno:
## Summary

Adds end-to-end support for Matrix gallery messages — a single event that carries multiple media attachments with a shared caption.

- **Receive:** A new `GalleryRoomTimelineItem` is built from `GalleryMessageContent` in `RoomTimelineItemFactory`. Galleries render in a fixed-width bubble (280pt, 12pt corners) using hero grid layouts: 1 hero, 2 full-width rows, 3 = hero + two squares, 4 = 2x2, 5 = 2+3, 6+ shows a `+N` overlay on the last tile. Video tiles overlay a play icon and duration. If any attachment lacks a thumbnail (documents/audio), the gallery falls back to a vertical file-row list with dividers between rows and before the caption.
- **Tap routing:** Tapping a gallery tile opens a media preview scoped to that gallery's attachments only. Tapping a regular image/video/audio/file from the timeline now also includes gallery attachments in the preview, paged individually (each gallery sub-item has a synthetic preview ID but inherits the parent gallery's `TimelineItemIdentifier`, so reply/redact/forward target the gallery event).
- **Send:** Wraps the SDK's `sendGallery(params:itemInfos:)` in `TimelineProxy` and `TimelineController`. `MediaUploadPreviewScreenViewModel` now sends a single gallery (with one shared caption) when more than one attachment is selected, and falls back to the per-item path for single uploads.
- **Composer:** Replaces the `QLPreviewController` preview with a SwiftUI peek carousel that previews adjacent items on either side, with a centred `X of Y` pill at the top. Documents and audio render their real content via `QLPreviewController` per tile instead of a black fallback.
- **Across the timeline:** edits, replies, threads, and menu actions learn the new `.gallery` content type so captions can be edited and reply/thread previews show the first item's thumbnail.

Note on size: the PR exceeds the 1000-additions guideline because the feature spans models, factory, view layer, send path, composer, and tests as one coherent unit; splitting would leave half-wired states in `develop`.

## Test plan

- [ ] Receive a gallery message and confirm rendering for 1, 2, 3, 4, 5, and 7+ items (overflow `+N`).
- [ ] Receive a mixed gallery (image + document + audio) and confirm the file-row list layout with dividers, and a thumbnail in place of the icon for the image row.
- [ ] Tap a gallery tile -> media preview only pages within the gallery's items.
- [ ] Tap a regular image elsewhere in the timeline -> preview pages through all images including gallery sub-items.
- [ ] Send a multi-attachment selection from the composer with a caption -> a single gallery event is sent (one `sendGallery` call), the caption appears once, and `X of Y` pill shows during preview.
- [ ] Send a single-attachment selection -> still uses the per-type path.
- [ ] Reply/edit/forward/redact on a gallery message -> actions target the gallery event; caption edit updates the gallery caption.
- [ ] Verify with iPhone and iPad simulators in portrait and landscape.
- [ ] Verify in dark mode.
- [ ] Verify with the largest dynamic type size.
- [ ] Verify with VoiceOver — galleries announce "N attachments" and gallery file rows expose filename/extension.

### DIFF
--- a/AccessibilityTests/Sources/GeneratedAccessibilityTests.swift
+++ b/AccessibilityTests/Sources/GeneratedAccessibilityTests.swift
@@ -183,6 +183,10 @@ extension AccessibilityTests {
         try await performAccessibilityAudit(named: "FullscreenDialog_Previews")
     }
 
+    func testGalleryRoomTimelineView() async throws {
+        try await performAccessibilityAudit(named: "GalleryRoomTimelineView_Previews")
+    }
+
     func testGlobalSearchScreenListRow() async throws {
         try await performAccessibilityAudit(named: "GlobalSearchScreenListRow_Previews")
     }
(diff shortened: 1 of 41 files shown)


svetoslav-sportinno:
Wrong target repo — re-opening on the fork.

PR #5597, opened 2026-05-15
Improve TopBanner sliding animation

Velin92:
This should fix the problem with the pinned events banner, where elements composing it slide in while others appear with fade in.


stefanceriu (approved):
👌

PR #5598, opened 2026-05-15
Add local dev script hooks

mredig:
Adds optional local script execution points to each git hook. If a script exists at `~/.local/elementx_local_dev/<hook-name>`, it will be run automatically at the appropriate hook stage. The repo path is exported as `ELEMENTX_REPO` for convenience.

This allows developers to maintain local automation (e.g. managing personal build configuration) without risk of accidentally committing those changes upstream. The local scripts live entirely outside the repo and never need to be checked in.

### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/.githooks/post-checkout
+++ b/.githooks/post-checkout
@@ -7,3 +7,9 @@ export PATH="$PATH:/opt/homebrew/bin"
 
 # ignores updates of 'Secrets.swift' to avoid pushing sensitive data by mistake
 git update-index --assume-unchanged Secrets/Secrets.swift
+
+LOCAL_SCRIPT_PATH="$HOME/.local/elementx_local_dev/post-checkout"
+if [ -x "$LOCAL_SCRIPT_PATH" ]; then
+	export ELEMENTX_REPO="$(pwd)"
+	"$LOCAL_SCRIPT_PATH"
+fi
--- a/.githooks/post-commit
+++ b/.githooks/post-commit
@@ -1,3 +1,9 @@
 #!/bin/sh
 command -v git-lfs >/dev/null 2>&1 || { echo >&2 "\nThis repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, remove this hook by deleting '.git/hooks/post-commit'.\n"; exit 2; }
 git lfs post-commit "$@"
+
+LOCAL_SCRIPT_PATH="$HOME/.local/elementx_local_dev/post-commit"
+if [ -x "$LOCAL_SCRIPT_PATH" ]; then
+	export ELEMENTX_REPO="$(pwd)"
+	"$LOCAL_SCRIPT_PATH"
+fi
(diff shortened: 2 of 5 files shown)


.githooks/post-checkout
```diff
@@ -7,3 +7,9 @@ export PATH="$PATH:/opt/homebrew/bin"
 
 # ignores updates of 'Secrets.swift' to avoid pushing sensitive data by mistake
 git update-index --assume-unchanged Secrets/Secrets.swift
+
+LOCAL_SCRIPT_PATH="$HOME/.local/elementx_local_dev/post-checkout"
+if [ -x "$LOCAL_SCRIPT_PATH" ]; then
+	export ELEMENTX_REPO="$(pwd)"
+	"$LOCAL_SCRIPT_PATH"
```
stefanceriu, line 14: Does this just silently fail? 🤔

.githooks/post-checkout
```diff
@@ -7,3 +7,9 @@ export PATH="$PATH:/opt/homebrew/bin"
 
 # ignores updates of 'Secrets.swift' to avoid pushing sensitive data by mistake
 git update-index --assume-unchanged Secrets/Secrets.swift
+
+LOCAL_SCRIPT_PATH="$HOME/.local/elementx_local_dev/post-checkout"
```
stefanceriu, line 11: Cool idea but what if we do this instead:
* we get setup-project to create a `.githooks/local` if it doesnt exist and add it to `.gitignore`
* we name those local hooks the same as the normal ones
* we stop exporting `ELEMENTX_REPO` because everything is run from within this repo. (See `git config core.hooksPath .githooks` in `SetupProject.swift`)

.. something along these lines in each one of them.
```
if [ -x "$LOCAL_SCRIPT" ]; then
    "$LOCAL_SCRIPT" || { echo "Local hook failed: $LOCAL_SCRIPT"; exit 1; }
fi
```

This way it's per repo and more intentional and in line with what exists already.

.githooks/post-checkout
mredig: If the file doesn't exist, it won't run/is silent. If it exists and fails, it is not silent. This is intentional.

- missing execution permission -> feedback to fix
- if there's a regex present in the code that shouldn't be committed -> hold up commit (just like the formatting does now)

.githooks/post-checkout
mredig: It's not a bad approach, but speaking for myself, I use worktrees extensively. I would need to recreate this and any scripts in each worktree, every time, basically defeating the purpose.

I believe the home directory is the best solution.

mredig:
I should add, I considered potential security risks and I don't think there is one, at least no more than there is already in executing the local githook scripts. Exporting `ELEMENTX_REPO` additionally isn't providing any more information that's not already there from `PWD`, it's just an explicit convenience. 

To perform an attack, from scratch, you'd need to plant a script at the explicit location in your home directory with the correct permissions and then unwittingly force the user to use git with this project (additionally doing the `swift run setup-project` step along the way). If it's a vector where the user already is contributing to the project, they'd need to compromise the system to get the script in place, at which point they already have the system under control.

stefanceriu (approved):
Allrighty, let's roll with it and see how it goes then 👍

PR #5599, opened 2026-05-16
Make `AnalyticsService` mockable via `AnalyticsServiceProtocol

mredig:
Extracts `AnalyticsServiceProtocol` from `AnalyticsService` and updates all injection points to use the protocol type, enabling Sourcery to generate `AnalyticsServiceMock` for use in tests and previews.

`AnalyticsService+Mock.swift` (a hand-written extension that constructed a lightly configured `AnalyticsService` for use in mocking contexts) is replaced by `AnalyticsServiceMock.default()`, generated from the new protocol.

Default-argument overloads that previously lived on `AnalyticsService` are moved to a protocol extension so all conforming types, including the mock, inherit them.

## Tests

Most tests and previews that previously depended on `AnalyticsService` now use `AnalyticsServiceMock.default()`. Two exceptions use a real `AnalyticsService` (injecting an `AnalyticsClientMock`) by necessity:

- **`AnalyticsTests`**: directly tests the service's own behaviour, so it must use the concrete type.
- **`AnalyticsSettingsScreenViewModelTests`**: the screen it tests exists specifically to let the user opt in/out of analytics. Its view model observes `appSettings.$analyticsConsentState` and reflects that through `AnalyticsService.isEnabled`. Using a mock would stub out that entire feedback loop, making the tests verify nothing meaningful. The real service is used with a mock client so the internal logic runs without hitting PostHog.

### Pull Request Checklist

- [ ] Voiceover enabled.


ElementX/Sources/Mocks/AnalyticsServiceMock.swift
```diff
@@ -0,0 +1,16 @@
+//
+// Copyright 2026 Element Creations Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+extension AnalyticsServiceMock {
```
stefanceriu, line 8: One of the later commits needs to bypass this default and use the mock directly. I would've instead kept consistency with the other mocks and replaced `default` with a `convenience init(_ configuration..`

UnitTests/Sources/AnalyticsSettingsScreenViewModelTests.swift
```diff
@@ -20,7 +20,9 @@ final class AnalyticsSettingsScreenViewModelTests {
     init() {
         AppSettings.resetAllSettings()
         appSettings = AppSettings()
-        analytics = .mock(settings: appSettings)
+        let client = AnalyticsClientMock()
+        client.isRunning = false
```
stefanceriu, line 24: .. as in here, `AnalyticsClientMock.init(_ config ...`

ElementX/Sources/Services/Analytics/AnalyticsServiceProtocol.swift
```diff
@@ -0,0 +1,160 @@
+//
+// Copyright 2026 Element Creations Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+import AnalyticsEvents
+
+// sourcery: AutoMockable
```
stefanceriu: Let's move this comment to an extension as we do in other places, it looks weird at the top of the documentation i.e. 

```
// sourcery: AutoMockable
extension AnalyticsServiceProtocol { }
```

stefanceriu (changes requested):
What a beaut, love it! Thanks for [not forgetting about it](https://github.com/element-hq/element-x-ios/pull/5548/changes#r3223975191)! 🙏

ElementX/Sources/Mocks/AnalyticsServiceMock.swift
mredig: I can do that, but...

UnitTests/Sources/AnalyticsSettingsScreenViewModelTests.swift
mredig: This is a real AnalyticsService with a client mock, not a service mock. The tests appear to be testing genuine interaction between the service and the view model, so the mock is ineffective here.

ElementX/Sources/Services/Analytics/AnalyticsServiceProtocol.swift
mredig: I moved it to the extension where default parameter method variants are living - if you want it on a dedicated extension instead, lemme know.

UnitTests/Sources/AnalyticsSettingsScreenViewModelTests.swift
mredig: re both comments: I was modeling after `UserIndicatorControllerMock.default`, tho technically i guess i made it a method and not a property. Oh well. 

If you want it refactored to your suggestion, I can, but I'd prefer to make that effort explicit, if desired.

stefanceriu: > This is a real AnalyticsService with a client mock

Oh I'm sorry, I misread that

> tho technically i guess i made it a method and not a property

Yeah, no worries, I doesn't matter.

stefanceriu:
🚀 👏

PR #5600, opened 2026-05-16
Resolve Some Warnings

mredig:
Resolves a number of compiler warnings across the project, primarily around `Sendable` conformance and a few test correctness issues.

## SendableBox / WeakSendableBox

Introduces two new utility types in `Other/SendableConformance/`:

- `SendableBox<Wrapped>` — wraps a non-`Sendable` type and synchronizes all access through an `NSRecursiveLock`. Conforms to `@unchecked Sendable`, which is safe because all mutations are lock-isolated. The recursive lock (vs. `NSLock`) is necessary because `@Observable`'s `withMutation` fires synchronously while the lock is held, and the same thread may re-enter on observation callbacks. Uses `@dynamicMemberLookup` for ergonomic property access. Also `@Observable` so wrapped values participate in SwiftUI's observation system.
- `WeakSendableBox<Wrapped: AnyObject>` — same concept with `weak` reference semantics, for breaking retain cycles in delegate patterns.

## Sendable conformances

- `JoinRule` / `AllowRule` — pure value types, no-op addition
- `AudioConverterProtocol` — stateless protocol, straightforward
- `TimelineControllerProtocol` — already `@MainActor` (implying `Sendable`), but the compiler still warned; explicit conformance resolved it
- `ClassicAppAccount` — gaining `Sendable` required `ClassicAppAccount.State` to also be `Sendable`. Making it `@MainActor` caused a cascade of `Equatable` conformance errors. Instead, each mutable property is backed by a `SendableBox`, with computed properties providing the same public interface. The indirect observation chain (through `SendableBox`'s own `@Observable` registrar) was validated in a playground.
- `ElementCallWidgetDriver`, `NotificationSettingsProxy`, `SessionVerificationControllerProxy` — non-`Sendable` stored properties wrapped in `SendableBox` / `WeakSendableBox`

## Other

- `BugReportService`: migrated from deprecated `SentrySDK.crashedLastRun` to `SentrySDK.lastRunStatus == .didCrash`
- Disabled `redundantSendable` in both SwiftLint and SwiftFormat — both were producing false positives
- `AttributedStringBuilderTests`: `NSTextAttachment` is marked `@_nonSendable(_assumed)` in the UIKit SDK, generating a warning at every `.attachment` access site. Consolidated all ~20 call sites behind two private `suppressedAttachment` extension properties that take the warning hit in one place. Also fixed `#expect(foundBlockquoteAndLink != nil)` (comparing a `Bool` to `nil` is invalid in Swift Testing) to `!= false`, and assigned away an unused `#require` result.

### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Other/SendableConformance/SendableBox.swift
+++ b/ElementX/Sources/Other/SendableConformance/SendableBox.swift
@@ -7,7 +7,7 @@
 
 import Foundation
 
-/// Wraps a non-Sendable type and assures most of it's interactions are done in isolation with locking mechanisms
+/// Wraps a non-Sendable type and assures most of its interactions are done in isolation with locking mechanisms
 ///
 /// The exception is reference types. If multiple SendableBox instances are created with a reference to the same class,
 /// each SendableBox operates in isolation and ignorance of the others. When using with reference types (or value
@@ -17,7 +17,7 @@ import Foundation
 @Observable
 @dynamicMemberLookup
 final class SendableBox<Wrapped>: @unchecked Sendable {
-    private let isolationLock = NSLock()
+    private let isolationLock = NSRecursiveLock()
     
     private var _value: Wrapped
     var value: Wrapped {
--- a/ElementX/Sources/Other/SendableConformance/SendableBox.swift
+++ b/ElementX/Sources/Other/SendableConformance/SendableBox.swift
@@ -1,0 +1,45 @@
+//
+// Copyright 2026 Element Creations Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Foundation
+
+/// Wraps a non-Sendable type and assures most of it's interactions are done in isolation with locking mechanisms
+///
+/// The exception is reference types. If multiple SendableBox instances are created with a reference to the same class,
+/// each SendableBox operates in isolation and ignorance of the others. When using with reference types (or value
+/// types that point to reference types), make sure to only create one Box instance per instance of Wrapped.
+///
+/// Leverages the `@dynamicMemberLookup` feature in swift to be able to access properties directly on the Box.
+@dynamicMemberLookup
+final class SendableBox<Wrapped>: @unchecked Sendable {
+    private let isolationLock = NSLock()
+    
+    private var _value: Wrapped
+    var value: Wrapped {
+        get { isolationLock.withLock { _value } }
+        set { isolationLock.withLock { _value = newValue } }
+    }
+    
+    subscript<T>(dynamicMember member: WritableKeyPath<Wrapped, T>) -> T {
+        get { value[keyPath: member] }
+        set { value[keyPath: member] = newValue }
+    }
+    
+    subscript<T>(dynamicMember member: KeyPath<Wrapped, T>) -> T {
+        value[keyPath: member]
+    }
+    
+    init(_ value: Wrapped) {
+        self._value = value
+    }
+}
+
+extension SendableBox: ExpressibleByNilLiteral where Wrapped: ExpressibleByNilLiteral {
+    convenience init(nilLiteral: ()) {
+        self.init(nil)
+    }
+}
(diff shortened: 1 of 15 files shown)


ElementX/Sources/Other/SendableConformance/SendableBox.swift
```diff
@@ -0,0 +1,46 @@
...
+/// Wraps a non-Sendable type and assures most of its interactions are done in isolation with locking mechanisms
+///
+/// The exception is reference types. If multiple SendableBox instances are created with a reference to the same class,
+/// each SendableBox operates in isolation and ignorance of the others. When using with reference types (or value
+/// types that point to reference types), make sure to only create one Box instance per instance of Wrapped.
+///
+/// Leverages the `@dynamicMemberLookup` feature in swift to be able to access properties directly on the Box.
+@Observable
+@dynamicMemberLookup
+final class SendableBox<Wrapped>: @unchecked Sendable {
```
Velin92, line 19: I don't think this a real solution to the Sendable problem, the fact that this object is an unchecked Sendable, feels more like we are just trying to find out a way to suppress the warning in a general matter insstead of going through what could be the real fix for a specific possible race condition

ElementX/Sources/Other/SendableConformance/SendableBox.swift
```diff
@@ -0,0 +1,46 @@
...
+import Foundation
+
+/// Wraps a non-Sendable type and assures most of its interactions are done in isolation with locking mechanisms
+///
+/// The exception is reference types. If multiple SendableBox instances are created with a reference to the same class,
+/// each SendableBox operates in isolation and ignorance of the others. When using with reference types (or value
+/// types that point to reference types), make sure to only create one Box instance per instance of Wrapped.
+///
+/// Leverages the `@dynamicMemberLookup` feature in swift to be able to access properties directly on the Box.
+@Observable
```
Velin92, line 17: A bit concerned about the observable conformance here, since this means the observation will inform of the change only when the the whole internal wrapped object changes which won't be applied for reference type internal changes (even if the type is itself Observable), and will probably be triggering for any change in a value type (since the value type is regenerated). At least these are my feelings, I am not 100% sure, and the lack of testing is probably also why I am quite hesitant on this PR.

UnitTests/Sources/AttributedStringBuilderTests.swift
```diff
@@ -1006,10 +1006,46 @@ struct AttributedStringBuilderTests {
...
-        for run in attributedString.runs where run.attachment != nil {
+        for run in attributedString.runs where run.suppressedAttachment != nil {
             return
         }
         
         Issue.record("Couldn't find expected value.")
     }
 }
+
+// warning suppression
```
Velin92, line 1017: As I said in a comment above, this feels more like a patch just to suppress warnings, but we're not really handling the underlying issues marked the compiler, and even if they are indeed safe in these contexts, i would rather have warnings tha to do nothing than overcomplicate the codebase just to suppress them

ElementX/Sources/Other/SendableConformance/SendableBox.swift
```diff
@@ -0,0 +1,46 @@
...
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Foundation
+
+/// Wraps a non-Sendable type and assures most of its interactions are done in isolation with locking mechanisms
+///
+/// The exception is reference types. If multiple SendableBox instances are created with a reference to the same class,
```
Velin92, line 12: Okay but this is not the only issue of reference types... another issue I see here with the lock is the fact that we are locking only the **pointer** of a reference type, not any of its content, so in the end we would still need to make such object Sendable add locks internally and so on, and according to how we are accessing the values of this reference type we might (due to warning suppressions) not be aware of concurrency issues, which creates more issues than positives... so this wrapper is kinda useless for a reference type... Which make me think it could only have sense for a value type.

Velin92:
Sorry but I feel this PR, suppresses warnings more than resolving them, also I think these Box objects are too limited in usage (they won't work for reference types at all), would require more tests to be effectively prove they are dealing concurrency in the right way or just hiding it.
I would suggest ands encourage instead explore more modern concurrency types and approaches of Swift 6, like "actors".
We are already working on a Swift 6 migration internally, that we are doing step by step in small increases (and where your contribution with the removal of the dependency injector will help immensely).

Feel free to reopen this PR or makew new one if we can find a way to solve these warnings in a more specific way instead of relying on a general object like the box, which will probably work okayish for value types (but I would argue that an actor, or dedicated solution in the codebase of that value would already be enough), but would not do anything for reference types.

PR #5608, opened 2026-05-19
Swift 6 migration for NSE

### LINKED ISSUE #4992: [Task] NSE - Swift 6 Migration

Velin92:
requires: https://github.com/element-hq/element-ios-enterprise/pull/13

So this PR essentially uses swift 6.2 and approachable concurrency in NSE.
NSE by itself did not support Swift 6 concurrency so I had to resort to use an actor that did handle the main logic in a concurency safe manner, while suppressing the NSE unsupported concurrency using `@preconcurrency import`.

The result is very good, but I am not happy with the current AppSettings implementation which I had to mark as `@unchecked Sendable` which is fine as of right now since they were always unsafe, but I think in a follow up PR, or maybe through the full implementation of: https://github.com/element-hq/element-x-ios/pull/5589 we can in the future make the AppSettings safely @Sendable.

This PR also includes some concurrency changes in the app, like the usage of asyncstream for the diffs in the Timeline builder, or the usage of a Mutex for the string caches (which were all required due to some enforcing Sendable conformance)

fixes #4992

### DIFF
--- a/ElementX/Sources/Application/Settings/AppSettings.swift
+++ b/ElementX/Sources/Application/Settings/AppSettings.swift
@@ -14,7 +14,7 @@ import Foundation
 import SwiftUI
 
 /// Common settings between app and NSE
-protocol CommonSettingsProtocol: AnyObject {
+protocol CommonSettingsProtocol: AnyObject, Sendable {
     var lastNotificationBootTime: TimeInterval? { get set }
     var selectedNotificationTone: NotificationTone? { get set }
 
@@ -34,7 +34,9 @@ enum AppBuildType {
 }
 
 /// Store Element specific app settings.
-final class AppSettings {
+///
+/// State is persisted in `UserDefaults`, which is thread-safe per Apple's documentation, hence `@unchecked`.
+final class AppSettings: @unchecked Sendable {
     private enum UserDefaultsKeys: String {
         case lastVersionLaunched
         case seenInvites
@@ -92,10 +94,10 @@ final class AppSettings {
         case developerOptionsEnabled
     }
     
-    private static var suiteName: String = InfoPlistReader.main.appGroupIdentifier
+    private nonisolated(unsafe) static var suiteName: String = InfoPlistReader.main.appGroupIdentifier
 
-    /// UserDefaults to be used on reads and writes.
-    private static var store: UserDefaults! = UserDefaults(suiteName: suiteName)
+    /// UserDefaults to be used on reads and writes. `UserDefaults` is thread-safe per Apple's documentation.
+    private nonisolated(unsafe) static var store: UserDefaults! = UserDefaults(suiteName: suiteName)
     
     static var appBuildType: AppBuildType {
         #if DEBUG
--- a/ElementX/Sources/Other/CurrentValuePublisher.swift
+++ b/ElementX/Sources/Other/CurrentValuePublisher.swift
@@ -10,7 +10,9 @@ import Combine
 
 /// A wrapper of CurrentValueSubject.
 /// The purpose of this type is to remove the possibility to send new values on the underlying subject.
-struct CurrentValuePublisher<Output, Failure: Error>: Publisher {
+///
+/// `CurrentValueSubject` is documented as thread-safe but is not formally `Sendable`, hence `@unchecked`.
+struct CurrentValuePublisher<Output, Failure: Error>: Publisher, @unchecked Sendable {
     private let subject: CurrentValueSubject<Output, Failure>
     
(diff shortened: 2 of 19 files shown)


ElementX/Sources/Other/Extensions/Bundle.swift
```diff
@@ -25,10 +25,11 @@ public extension Bundle {
     // MARK: - Localisation
     
     /// Overrides `Bundle.app.preferredLocalizations` for testing translations.
-    static var overrideLocalizations: [String]?
-    
+    nonisolated(unsafe) static var overrideLocalizations: [String]?
```
Velin92, line 30: need to add some docs here, the reason why this is not a big deal is that we're doing this only in tests, so it can be nonisolated(unsafe)

ElementX/Sources/Other/Logging/Tracing.swift
```diff
@@ -22,7 +22,7 @@ enum Tracing {
     /// Set this to temporarily override the directory from which logs will be collected.
     /// This basically only affects ``logFiles``, and doesn't inform the SDK to write
     /// the logs to a different directory, which should be done before setting this.
-    static var logsDirectoryOverride: URL?
+    nonisolated(unsafe) static var logsDirectoryOverride: URL?
```
Velin92, line 25: also here, this is done only once for overrides so it's fine

NSE/Sources/NotificationHandler.swift
```diff
@@ -10,7 +10,9 @@ import CallKit
 import MatrixRustSDK
 import UserNotifications
 
-class NotificationHandler {
+/// `@unchecked Sendable`: each instance handles a single notification on whatever thread the OS
+/// dispatched `didReceive` on; instance state is only touched from within that single async flow.
+final class NotificationHandler: @unchecked Sendable {
```
Velin92: WIP here, not sure about this, I feel a better handling without unchecked is possible, but is also true that seems that Apple did not update a lot of the native struct/classes/methods for the NSE so we might still resort to some unchecked stuff here.

NSE/SupportingFiles/target.yml
```diff
@@ -111,7 +111,7 @@ targets:
     - path: ../../ElementX/Sources/Other/InfoPlistReader.swift
     - path: ../../ElementX/Sources/Other/Logging
     - path: ../../ElementX/Sources/Other/MatrixEntityRegex.swift
-    - path: ../../ElementX/Sources/Other/NetworkMonitor
+    - path: ../../ElementX/Sources/Other/NetworkMonitor/NetworkMonitorProtocol.swift
```
Velin92, line 115: not sure about this import

ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -34,7 +34,9 @@ enum AppBuildType {
 }
 
 /// Store Element specific app settings.
-final class AppSettings {
+///
+/// State is persisted in `UserDefaults`, which is thread-safe per Apple's documentation, hence `@unchecked`.
+final class AppSettings: @unchecked Sendable {
```
Velin92, line 39: This is kinda true... but not completely, we also have volatile states which may not be entirely thread safe. It's good enough but would prefer to not have the `@unchecked Sendable` at all here. Not sure how to achieve this without a very big refactor by making the AppSettings both `@MainActor` and `Sendable` which would be also correct in my opinion.

ElementX/Sources/Other/CurrentValuePublisher.swift
```diff
@@ -10,7 +10,9 @@ import Combine
 
 /// A wrapper of CurrentValueSubject.
 /// The purpose of this type is to remove the possibility to send new values on the underlying subject.
-struct CurrentValuePublisher<Output, Failure: Error>: Publisher {
+///
+/// `CurrentValueSubject` is documented as thread-safe but is not formally `Sendable`, hence `@unchecked`.
+struct CurrentValuePublisher<Output, Failure: Error>: Publisher, @unchecked Sendable {
```
Velin92, line 15: true, but I don't like it, and the only instancr where this is a problem is the MediaLoader because it does some shenanigans with the networkReachabilityPublisher which is of this type here.

ElementX/Sources/Application/Settings/AppSettings.swift
Velin92: Will probably be fixed in another PR

stefanceriu (approved):
We talked about all of these over a call yesterday, still looks good me 👍

PR #5627, opened 2026-05-22
Tidy up mock creation.

pixlwave:
We have an established pattern of creating mocks like so `XYZMock(.init())`, however the code diverges from this in a few places:
- There are a handful of `XYZMock.default` uses, these have been converted to the normal pattern.
- There are some places where the init looks like `XYZMock(configuration: .init())` (mainly my doing 🙈), these have had the parameter label removed.


ElementX/Sources/Mocks/UserIndicatorControllerMock.swift
pixlwave, line 1: This one was pointless – the closures are safely unwrapped in the mock so there's no need to set them to no-op values.

Velin92 (approved):

PR #5629, opened 2026-05-22
Fix for overlapping timestamp when last block is not plainText

Velin92:
Name says it all


ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FormattedBodyText.swift
```diff
@@ -63,7 +63,15 @@ struct FormattedBodyText: View {
         // component so that the bubble's natural size accommodates the overlaid
         // timestamp, and TextKit decides whether to keep the timestamp on the last
         // line or push it to a new one.
-        let components = attributedComponents
+        var components = attributedComponents
+        // When the body ends in a block component (blockquote/codeBlock) the overlaid
+        // timestamp would land on top of it, so append an empty trailing plain-text
+        // component that exists solely to reserve space for the timestamp underneath.
+        if trailingReservedSize != .zero, let last = components.last, last.type != .plainText {
+            components.append(AttributedStringBuilderComponent(id: "trailingTimestampSpacer",
```
Velin92: I mean not sure what this id is about, but I think is safe to use this? Or maybe I should use an UUID ?

Velin92: No wait an UUID would change every single time and we don't want that

pixlwave: I think this is just to allow `ForEach` to iterate over an identifiable collection, so should be fine as a constant.

pixlwave (approved):

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FormattedBodyText.swift
Velin92: why we don't just use the offset value of the enumerated array as the id then?

PR #5630, opened 2026-05-22
Resolve More Warnings (revolving around Sendable)

mredig:
Following up on #5600. This PR tackles some `Sendable` warnings that require synchronized access to mutable stored properties.

The warning reduction goal was made in the previous PR - reiterating briefly here that this continues that work, now down to 11 warnings in the main target and 13 overall.

## `LockBox` / `WeakLockBox`

Introduces two utility types in `Other/SendableConformance/`. The rename is intentional - these are about synchronized concurrent access, not about making types `Sendable`. `Sendable` conformance is a consequence, not the goal.

Key design decisions:
- `Sendable` conformance is **conditional on `Wrapped: Sendable`** -- the box synchronizes access to a type that is already safe to share. It cannot be used to make an unsafe type appear safe.
	- the original `@unchecked Sendable` was a brain fart, not intentional
- `withLock { }` is provided for cases where an operation must be atomic with the pointer read, or where a compound read-modify-write must be uninterruptible
- `WeakLockBox` directly conforms to delegate protocols (rather than wrapping in a separate class), eliminating the `WeakNotificationSettingsProxy` and `WeakSessionVerificationControllerProxy` wrapper classes entirely

The conditional `Sendable` requirement had a useful side effect: `NotificationSettingsProxy` and `SessionVerificationControllerProxy` both needed to earn `Sendable` conformance explicitly. Both did so cleanly - no `@unchecked` required.

## `ClassicAppAccount.State`

The properties themselves are pure value types (`Bool?`, `AvailableSecrets?`), inherently `Sendable` and already satisfying the conditional conformance requirement. The issue is that they're mutable properties on a reference type. `@MainActor` isolation was attempted first since the type is `@Observable` and UI-facing, but caused a cascade of `Equatable` conformance errors propagating into ~20 unrelated sites. Making `State` an actual `actor` type is also off the table since `@Observable` can only be applied to `class` types.

As for `@Observable`, it propagates through reference types - this was validated in a playground because I wasn't certain at first myself - happy to share it. This is one of the subtler improvements `@Observable` brings over `ObservableObject`, and easy to miss when reasoning about the change.

## `ElementCallWidgetDriver`

`WidgetDriverAndHandle` is a `uniffi::Record` wrapping two `Arc`-backed Rust types that both implement `Send + Sync`, verified in the SDK docs. Their Swift wrappers are correctly `@unchecked Sendable`, satisfying the conditional conformance. The warning is on `ElementCallWidgetDriver` itself holding a mutable `var` of that type. `LockBox` provides synchronized access with exclusive ownership of the wrapped pointer.

## `NotificationSettingsProxy` / `SessionVerificationControllerProxy`

Both delegate protocols require `Sendable` conformance, but the delegate holder needed a `weak var` for memory safety - putting the two requirements at odds. `WeakLockBox` resolves this by synchronizing access to the weak reference, leaving the containing type with no mutable state of its own and enabling genuine (non-`@unchecked`) `Sendable` conformance on both proxy types.

---

The feedback on #5600 suggested actors as the preferred direction. Actors are a great tool and worth pursuing in the broader Swift 6 migration - but they require `await` at every external access point, meaning any synchronous call site can't use them without becoming async - a change that cascades architecturally and introduces suspension overhead that, while small individually, can accumulate. Some degree of manual thread safety is simply inherent to Swift 6 conformance, as [noted by Apple's own engineers](https://forums.swift.org/t/overhead-of-using-actors-at-scale/79466/3). For the cases in this PR, I believe `LockBox` is the practical solution. If there's a preferred alternative for these specific patterns - particularly `weak var` delegate holders and mutable stored properties on `@Observable` classes - happy to discuss and adapt. On that note, it would be appreciated if future PRs could receive some discussion before being closed, even if briefly - it helps make sure the work can be steered in the right direction rather than redone from scratch.

P.S. In earlier commits, I disabled redundant_sendable in both swiftlint and swiftformat as it was previously firing false positives. I tested it once more before I finalized the PR and, while it was no longer firing false positives, I found that it reverted one of the places I set Sendable: ClassicAppAccount. 

I reverted because that is what the project status quo is currently, but I'd encourage you to remove this formatting rule. `ClassicAppAccount` is apparently implicit `Sendable` now, but that means that its signature is identical to before when it wasn't, when `ClassicAppAccount.State` was not `Sendable`. By marking `ClassicAppAccount` explicitly `Sendable`, it clearly delineates intent that it *needs* to be `Sendable`, and any future additions should not downgrade its status.

### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Other/SendableConformance/LockBox.swift
+++ b/ElementX/Sources/Other/SendableConformance/LockBox.swift
@@ -36,7 +36,7 @@ import Foundation
 /// Leverages `@dynamicMemberLookup` for ergonomic property access directly on the box.
 @Observable
 @dynamicMemberLookup
-final class LockBox<Wrapped>: @unchecked Sendable {
+final class LockBox<Wrapped> {
     private let isolationLock = NSRecursiveLock()
     
     private var _value: Wrapped
@@ -77,3 +77,5 @@ extension LockBox: ExpressibleByNilLiteral where Wrapped: ExpressibleByNilLitera
         self.init(nil)
     }
 }
+
+extension LockBox: @unchecked Sendable where Wrapped: Sendable { }
--- a/ElementX/Sources/Other/SendableConformance/LockBox.swift
+++ b/ElementX/Sources/Other/SendableConformance/LockBox.swift
@@ -7,31 +7,29 @@
 
 import Foundation
 
-/// Synchronizes access to a stored value across concurrency boundaries by protecting it with a lock, enabling
-/// `Sendable` conformance on the containing type.
+/// Synchronizes access to a `Sendable` stored value across concurrency boundaries by protecting it with a lock.
 ///
 /// This addresses one specific concern: the container's unsynchronized hold on its stored property. It does not
 /// address the internal thread-safety of the wrapped value itself -- that remains a separate responsibility. For
 /// reference types, this means the box synchronizes the pointer, not the internals of the referenced object. This
 /// is intentional and consistent with how other synchronization primitives work.
 ///
-/// **Important:** `box.value.doThing()` retrieves the value under the lock but executes `doThing()` outside it.
-/// When the operation itself must be atomic with the lock, use `withLock { }` instead.
+/// `Sendable` conformance is conditional on `Wrapped: Sendable` -- the box synchronizes access to a type that
+/// is already safe to share, it does not make an unsafe type safe.
+///
+/// **Important:** `box.value.doThing()` retrieves the value under the lock but executes `doThing()` outside it,
+/// relying on `Wrapped`'s `Sendable` conformance for safety. When the operation itself must be atomic with the
+/// lock, use `withLock { }` instead.
 ///
 /// **Disciplines required:**
-/// - For reference types: the wrapped type should itself be `Sendable`, or internal mutations must be
-/// otherwise synchronized
-/// - Only one `LockBox` instance should exist per instance of a wrapped reference type (multiple boxes
-///  do not synchronize with each other)
-/// - Mutations to the wrapped value should go through the box, not via a retained reference to the wrapped object
 /// - For reference types, use `withLock { }` when the operation on the referenced object must be atomic with
 ///   the pointer read
 ///
 /// **Benefits attained:**
 /// - Access to the stored property (pointer/reference for reference types) is synchronized -- no unsynchronized
-///   reassignment of the stored property can occur
-/// - When used as a property on a container, each wrapped property carries its own synchronized conformance,
-/// reducing the need for broad `@unchecked Sendable` on the containing type
+///   reassignment can occur
+/// - Each wrapped property carries its own synchronized conformance, reducing the need for broad
+///   `@unchecked Sendable` on the containing type
 ///
 /// Leverages `@dynamicMemberLookup` for ergonomic property access directly on the box.
 @Observable
@@ -58,6 +56,37 @@ final class LockBox<Wrapped> {
         self._value = value
     }
     
+    /// Executes `block` with exclusive access to the wrapped value.
+    ///
+    /// Use this instead of `value` when:
+    /// - The operation on a referenced object must be atomic with the pointer read -- i.e. the reference should
+    /// not be able to change between reading it and acting on it. This is the typical use case for
+    /// wrapped reference types:
+    ///
+    ///     ```swift
+    ///     // Not atomic -- reference could be replaced between read and method call
+    ///     box.value?.doThing()
+    ///
+    ///     // Atomic -- pointer read and method call are a single unit
+    ///     box.withLock { value in
+    ///         value?.doThing()
+    ///     }
+    ///     ```
+    ///
+    /// - A compound operation (read + modify, or multiple mutations) must execute as an uninterruptible unit,
+    /// preventing other threads from observing intermediate state. This is more typical for wrapped value
+    /// types, though neither use case is exclusive to one or the other:
+    ///
+    ///     ```swift
+    ///     // Not atomic -- another thread could modify value between the read and write
+    ///     let current = box.value
+    ///     box.value = transform(current)
+    ///
+    ///     // Atomic -- the entire operation is a single unit
+    ///     box.withLock { value in
+    ///         value = transform(value)
+    ///     }
+    ///     ```
     func withLock<Success, Failure: Error>(_ block: (inout Wrapped) throws(Failure) -> Success) throws(Failure) -> Success {
         do {
             return try isolationLock.withLock {
--- a/ElementX/Sources/Other/SendableConformance/LockBox.swift
+++ b/ElementX/Sources/Other/SendableConformance/LockBox.swift
@@ -107,4 +107,10 @@ extension LockBox: ExpressibleByNilLiteral where Wrapped: ExpressibleByNilLitera
     }
 }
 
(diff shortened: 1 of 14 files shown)


Velin92 (changes requested):
As I said already in a previous PR I am not a fan of the LockBox implementation. Could we try other solutions? Like Mutexes, actors and so on.

ElementX/Sources/Other/SendableConformance/LockBox.swift
```diff
@@ -0,0 +1,116 @@
...
+/// **Benefits attained:**
+/// - Access to the stored property (pointer/reference for reference types) is synchronized -- no unsynchronized
+///   reassignment can occur
+/// - Each wrapped property carries its own synchronized conformance, reducing the need for broad
+///   `@unchecked Sendable` on the containing type
+///
+/// Leverages `@dynamicMemberLookup` for ergonomic property access directly on the box.
+@Observable
+@dynamicMemberLookup
+final class LockBox<Wrapped> {
```
Velin92, line 44: I mean why use this over a Mutex, which is the new Swift API to manage locked access?
Is it for the recursive lock? Recursive locks are very inefficient and if the code requires a recursive lock without the code being intentionally recursive there is definitely something wrong.

ElementX/Sources/Other/SendableConformance/WeakLockBox.swift
```diff
@@ -0,0 +1,53 @@
...
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Foundation
+
+/// See ``LockBox`` - Mostly the same, but the value stored with `weak` reference semantics
+@dynamicMemberLookup
+final class WeakLockBox<Wrapped: AnyObject> {
```
Velin92, line 12: same as above

ElementX/Sources/Other/SendableConformance/LockBox.swift
mredig: Thanks for the feedback - the previous PR's feedback directly inspired some meaningful design improvements worth walking through.

**`LockBox` vs `Mutex`:**

`LockBox` offers things `Mutex` doesn't: stricter compiler-enforced `Sendable` safety, ergonomics that make correct usage the path of least resistance, `@Observable` conformance, and a weak variant. The first two are worth unpacking.

On ergonomics - `@dynamicMemberLookup` allows safe ergonomic access while retaining lock safety. `Mutex` requires `withLock { }` for every access with no shorthand. When `Wrapped: Sendable`, `.value` and full mutable/immutable property access are available. For non-`Sendable` wrapped types the API is tiered - read-only access to individual `Sendable` properties is still available, mutation and `.value` are compiler errors, and `withLock { }` remains the full-access path with `sending` on the return type, similar to `Mutex`.

This is demonstrated concretely. First, how `LockBox` and `Mutex` behave as stored properties on a `Sendable` container:

```swift
class NotSafe {
    let asdf = "baz"
    var foo = "bar"
}

final class Safe: Sendable {
    let asdf = "baz"
    nonisolated(unsafe) var foo = "bar" // pretend this is otherwise safe for this simplistic demo
}

final class ContainerExample: Sendable {
    let safeRaw: Safe                          // works
//  let notSafeRaw: NotSafe                    // compiler error

//  var safeRawMutable: Safe                   // compiler error - mutable stored property
//  var notSafeRawMutable: NotSafe             // compiler error - mutable stored property

    let boxSafeMutable: LockBox<Safe>          // works - mutable access via box
//  let boxNotSafeMutable: LockBox<NotSafe>    // compiler error - LockBox not Sendable

    let mutexSafeMutable: Mutex<Safe>          // works
    let mutexNotSafeMutable: Mutex<NotSafe>    // works - silently accepted
}
```

And the tiered access in practice:

```swift
let safeTest = LockBox(Safe())
safeTest.foo              // readable
safeTest.foo = "asdf"     // writable
safeTest.asdf             // readable
// safeTest.asdf = "a"    // compiler error - 'asdf' is a 'let' constant
safeTest.value            // accessible

let unsafeTest = LockBox(NotSafe())
unsafeTest.foo            // readable (String is Sendable)
// unsafeTest.foo = "asdf"  // compiler error - NotSafe not Sendable
unsafeTest.asdf           // readable (String is Sendable)
// unsafeTest.asdf = "a"  // compiler error - 'asdf' is a 'let' constant
// unsafeTest.value       // compiler error - NotSafe not Sendable
```

This tiered design came directly out of the previous feedback.

**Where `Mutex`'s approach has limitations:**

The job of both `Mutex` and `LockBox` in this context is to make a mutating reference pointer safe across concurrency boundaries. By default, both protect the pointer itself - though `withLock` and discipline can extend that protection to the internals when needed. When `Wrapped: Sendable`, escaping the value outside the lock is a non-issue - `Sendable` guarantees thread safety beyond the lock, so the caller can use it freely.

The difference is what happens when `Wrapped` is *not* `Sendable`. `Mutex` accepts this silently - the container retains its `Sendable` claim with no compiler signal that the wrapped type needs more care. `LockBox` surfaces it: to use `LockBox<NotSafe>` on a `Sendable` container, the property must be marked `nonisolated(unsafe)`:

```swift
nonisolated(unsafe)
let boxNotSafeMutable: LockBox<NotSafe>   // loud signal - wrapped type needs attention
```

That's a deliberate, visible flag that the wrapped type requires more scrutiny. Compare that to the existing `AppHooks` in this codebase, which wraps non-`Sendable` protocols in `Mutex` and exposes them via computed properties - the compiler sees nothing wrong, and future refactors adding mutable state to any of those protocol implementations would be completely unprotected behind a valid `Sendable` claim. As it is *currently*, I believe everything is okay - but it's the same concern raised about the previous PR's `@unchecked Sendable` usage: the conformance is satisfied on paper, but the safety is entirely dependent on discipline.

That said - if the `nonisolated(unsafe)` signal isn't considered sufficient and a harder line is preferred, there's the option of requiring `Wrapped: Sendable` unconditionally on `LockBox` itself, making the non-`Sendable` case a compiler error outright. That would cleanly separate `LockBox`'s territory from `Mutex`'s, at the cost of losing the intermediate signal the current design provides. Happy to make that change if preferred.

**On `NSRecursiveLock`:**

`WeakLockBox` has already been updated to use `NSLock` - it has no `@Observable` conformance and no reentrancy requirement. `LockBox` uses `NSRecursiveLock` specifically because `@Observable`'s `withMutation` fires synchronously while the lock is held, and the same thread can reenter. The choice is deliberate and scoped to that requirement. `NSRecursiveLock` is ~20-50% slower than `NSLock` in uncontended scenarios, but both sit in the ~50-150ns range - the absolute difference is small and the real cost cliff for any lock is contention, not the choice between recursive and non-recursive. Happy to revisit if profiling shows it matters. If preferred, `LockBox` could also be made generic over `NSLocking`, defaulting to `NSLock` and allowing `NSRecursiveLock` to be opted into only where `@Observable` reentrancy is needed.

Velin92:
Thanks for the contribution, but since we're currently working on implementing Swift 6 migration, we're also creating a good and safe framework of patterns to handle concurrency (more info in the link below)
https://github.com/element-hq/element-x-ios/issues/4990

So I'll close this PR, since it's kinda clashing with our migration work, which is currently happening slowly behind curtains trying to find te best solution for each specific concurrency problem we may find in the app.

PR #5633, opened 2026-05-25
Set map tiler URL from well known if available

### LINKED ISSUE #5623: [Task] Support `map_style_url` configuration from `.well-known/matrix/client`

Velin92:
Name says it all. If the well known (taken from the sdk directly) owns a TileServerInfo value, we store it, and we use that url in place of the MapTilerConfiguration generated one.
I also use an enum called MapTilerSettings to differentiate between the usage of a configuration or the usage of an overridden URL.

fixes #5623


ElementX/Sources/Services/Client/Client.swift
```diff
@@ -16,17 +16,28 @@ extension ClientProtocol {
...
+    /// Reads the homeserver-advertised map tile server (`tile_server.map_style_url`
+    /// from the matrix client well-known) and applies it to ``AppSettings``.
+    ///
+    /// Clears any previously applied URL when none is advertised (or the well-known
+    /// is unavailable), so each session start reflects the latest server-side
+    /// configuration.
+    func updateMapTilerSettings(in appSettings: AppSettings) async {
+        let url = await tileServer().flatMap { URL(string: $0.mapStyleUrl) }
+        appSettings.remoteMapStyleURL = url
+    }
```
Velin92, line 45: Since I apply this when authenticating (which happens before the ClientProxy is created) I added the handling here as an extension to the ClientProtocol, a bit like we do for the element well known

ElementX/Sources/Other/MapLibre/MapURLs.swift
```diff
@@ -8,44 +8,68 @@
...
+            pathComponents.removeLast()
+            pathComponents.append(staticComponent)
+            components.path = pathComponents.joined(separator: "/")
+
+            var queryItems = components.queryItems ?? []
+            queryItems.append(.init(name: "attribution", value: attribution.rawValue))
+            components.queryItems = queryItems
+
+            return components.url
+        }
```
stefanceriu, line 58: Why oh why does this need to be so complex? 😢

ElementX/Sources/Application/Settings/AppSettings.swift
pixlwave, line 1: I kind of wonder if there's a way we can adapt `RemotePreference` to also work here too. It feels annoying to basically have the same logic, but with different types as the output.

Perhaps we could rename `MapTilerSettings.overrideURL` to simply `MapTilerSettings.url` and just know that we set `.configuration` in the app and `.url` from the well-known?

ElementX/Sources/Application/Settings/MapTilerSettings.swift
```diff
@@ -30,13 +30,13 @@ enum MapTilerSettings {
         let darkStyleID: String
     }
 
-    case configuration(Configuration)
-    case overrideURL(URL)
+    case config(Configuration)
```
pixlwave: `configuration` here still surely? 😄

ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -398,12 +398,18 @@ final class AppSettings: @unchecked Sendable {
...
-                                                                   lightStyleID: "9bc819c8-e627-474a-a348-ec144fe3d810",
-                                                                   darkStyleID: "dea61faf-292b-4774-9660-58fcef89a7f3")
+
+    /// The locally-bundled MapTiler configuration. Enterprise builds layer their MDM overrides
+    /// on top of this via ``override(...)``.
+    static let bundledMapTilerConfiguration = MapTilerSettings.Configuration(baseURL: "https://api.maptiler.com/maps",
+                                                                             apiKey: Secrets.mapLibreAPIKey,
+                                                                             lightStyleID: "9bc819c8-e627-474a-a348-ec144fe3d810",
+                                                                             darkStyleID: "dea61faf-292b-4774-9660-58fcef89a7f3")
+
```
pixlwave: This can be inlined now right?

ElementX/Sources/Other/MapLibre/MapURLs.swift
Velin92: sadly the style.json urls usually have the API Key as a query param, so to get the static image url we need to strip it, but preserve the api key query param (and generally any other param if the url provided it, even if the only one I have seen so far is the api key). Also checked... there a single url per light/dark mode (where the color scheme key is  apath value before the style.json like `light_key/style.json&key=apiKey`) so there is nothing we can do about selecting dark and light mode dynamically without the MSC being extended...

ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -398,12 +398,17 @@ final class AppSettings: @unchecked Sendable {
...
-                                                                   darkStyleID: "dea61faf-292b-4774-9660-58fcef89a7f3")
+
+    /// The locally-bundled MapTiler configuration.
+    static let bundledMapTilerConfiguration = MapTilerSettings.Configuration(baseURL: "https://api.maptiler.com/maps",
+                                                                             apiKey: Secrets.mapLibreAPIKey,
+                                                                             lightStyleID: "9bc819c8-e627-474a-a348-ec144fe3d810",
+                                                                             darkStyleID: "dea61faf-292b-4774-9660-58fcef89a7f3")
+
+    /// The resolved map tile settings. Defaults to ``MapTilerSettings/config(_:)`` with the
+    /// bundled configuration and is remotely overridden with ``MapTilerSettings/url(_:)`` when
```
pixlwave: Shouldn't these be `.` rather than `/`?

UnitTests/Sources/MatrixClientWellKnownTests.swift
```diff
@@ -0,0 +1,75 @@
...
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+@testable import ElementX
+import Foundation
+import MatrixRustSDKMocks
+import Testing
+
+/// Exercises ``ClientProtocol/updateMapTilerSettings(in:)``. The SDK is now the
```
pixlwave: Same

ElementX/Sources/Other/MapLibre/MapURLs.swift
```diff
@@ -8,44 +8,68 @@
...
+
         var url: URL = baseURL
         url.appendPathComponent(styleID(for: style), conformingTo: .item)
         url.append(queryItems: [URLQueryItem(name: "key", value: apiKey)])
         return url
     }
-    
-    private func styleID(for style: MapTilerStyle) -> String {
+
+    func styleID(for style: MapTilerStyle) -> String {
```
pixlwave, line 74: Are these changes intentional? We should at least keep the `// MARK: - Private` but tbh I don't see why this has changed.

ElementX/Sources/Other/MapLibre/MapURLs.swift
stefanceriu: Oh, sorry, I wasn't clear. It was just rhetorical.

ElementX/Sources/Application/Settings/MapTilerSettings.swift
```diff
@@ -0,0 +1,42 @@
...
+//
+
+import Foundation
+
+/// The map tile settings used to build interactive and static map URLs.
+///
+/// Either resolves to a locally-bundled MapTiler configuration, or to a remote
+/// `style.json` URL advertised by the homeserver through the matrix client well-known
+/// (`tile_server.map_style_url`). When the override URL is present it fully replaces
+/// the bundled configuration and is used regardless of light/dark appearance.
```
pixlwave: I don't think we need any of this. If anything a 1-line comment on each of the cases would be better.

pixlwave (approved):
LGTM overall 👍

pixlwave:
Sorry, I had to rebase the submodule for #5654, so I've force-pushed that last commit.

UnitTests/Sources/MatrixClientWellKnownTests.swift
Velin92: > That's DocC symbol-link syntax. The double backticks (vs single backticks for inline code) tell
>   Xcode/DocC to resolve the contents as a symbol reference and turn it into a clickable link in Quick
>    Help and generated docs.

according to Claude Code, and i tested it, if you do Product -> Build Documentation it does actually work it's quite cool, but not too important can remove if you wish

ElementX/Sources/Application/Settings/AppSettings.swift
Velin92: answered below

Issue #5636, opened 2026-05-26
edit message not taking effect in ui, caused a ban


simonmichael:
### Steps to reproduce

Recently with Element X IOS test flight version, I have noticed that intermittently, while editing a message, tapping the confirm/accept message has no visual effect. Sometimes it works if you tap it again. This leads the user to tap it multiple times. And this can easily trigger an automated ban. Eg, about 48 hours ago I was editing a message in Matrix HQ, and probably pressed the accept button a bunch of times trying to get it to accept the edit.

#### What did you expect?

I expected the edit to be accepted on first tap of the confirm button.

#### What happened instead?

Tapping the button had no effect. I probably tapped it a bunch of times trying to make it go through. I was instantly banned from this and other rooms. Some no longer appeared in room lists; some appeared and could be read but not posted to. I was lucky that the ban was reversed an hour or two later.

My edited message was removed so I couldn't inspect it with Element, but a gomuks user provided these details:

<img width="1070" height="1224" alt="Image" src="https://github.com/user-attachments/assets/0b776868-290c-45ac-bda3-73a5c58a5e2b" />

> appears your client spammed edits to the message in question. after 10, it looks like an automated protection nuked you.

<img width="2384" height="990" alt="Image" src="https://github.com/user-attachments/assets/8d89998f-51e1-47f0-88ad-185aad9f87e5" />

> there are 10 events like this, each making this edit
> with different IDs and timestamps
>
> if it is at all helpful, here is one of those PDUs:
> https://paste.zirco.dev/q/FtfuF_FZ6wWmBO4PeDatV#az1MeHVOWE5ET1l5U0sxWUdwbWMlMkZ3Vk93ZzY2UjNVRndxWktkY1RnZSUyQnRPYyUzRA==

### Your phone model

iphone SE 2nd edition

### Operating system version

IOS 26.4.2

### Application version

26.05.4 (1222)

### Homeserver

matrix.org

### Will you send logs?

Yes

simonmichael:
Possibly related, but a different symptom: I have also noticed that sometimes an edit appears to complete, but the new text is not shown until I exit to the room list and then re-enter the room screen.

stefanceriu:
> Possibly related, but a different symptom: I have also noticed that sometimes an edit appears to complete, but the new text is not shown until I exit to the room list and then re-enter the room screen.

[...]

PR #5637, opened 2026-05-26
Bump the calendar version ready for the next release

RiotRobot:
- Version bump


pixlwave:
Blocking this one until we fix #5636 so that the public Nightly doesn't get stuck behind a review.

pixlwave (approved):

PR #5649, opened 2026-05-26
Fix for the timeline item not updating the content after an edit

### LINKED ISSUE #5636: edit message not taking effect in ui, caused a ban

Velin92:
Reverts some changes that were made, and adds some comments so that now we know why we should keep using the content based identifier in the ForEach of the FormattedBody.

The commit that created the issue: https://github.com/element-hq/element-x-ios/commit/de6529afac7e117776bbd

Fixes https://github.com/element-hq/element-x-ios/issues/5636

### DIFF
--- a/ElementX/Sources/Other/Extensions/AttributedString.swift
+++ b/ElementX/Sources/Other/Extensions/AttributedString.swift
@@ -37,7 +37,8 @@ extension AttributedString {
                 .plainText
             }
                     
-            components.append(AttributedStringBuilderComponent(attributedString: attributedString,
+            components.append(AttributedStringBuilderComponent(id: String(attributedString.characters),
+                                                               attributedString: attributedString,
                                                                type: componentType))
         }
         
--- a/ElementX/Sources/Other/HTMLParsing/AttributedStringBuilderProtocol.swift
+++ b/ElementX/Sources/Other/HTMLParsing/AttributedStringBuilderProtocol.swift
@@ -8,13 +8,15 @@
 
 import Foundation
 
-struct AttributedStringBuilderComponent: Hashable {
+struct AttributedStringBuilderComponent: Hashable, Identifiable {
     enum ComponentType {
         case plainText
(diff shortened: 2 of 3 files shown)


stefanceriu (commented):
Please reference the commit that broke it and the bug you're fixing.

More importantly though, this is not about an edit animation but about the timeline item not updating properly to show the edited content, right?

stefanceriu (approved):
👌

PR #5651, opened 2026-05-27
Add message sending end editing steps to the user session flow UI test

stefanceriu:
Should catch regressions like the ones in https://github.com/element-hq/element-x-ios/issues/5636

### DIFF
--- a/ElementX/Sources/Services/Timeline/TimelineController/MockTimelineController.swift
+++ b/ElementX/Sources/Services/Timeline/TimelineController/MockTimelineController.swift
@@ -13,6 +13,9 @@ import Foundation
 import MatrixRustSDK
 
 class MockTimelineController: TimelineControllerProtocol {
+    private let lastInsertedMessageID: TimelineItemIdentifier = .event(uniqueID: .init("last_message"),
+                                                                       eventOrTransactionID: .eventID("last_message"))
+    
     /// An array of timeline item arrays that will be inserted in order for each back pagination request.
     var backPaginationResponses: [[RoomTimelineItemProtocol]] = []
     /// An array of timeline items that will be appended in order when ``simulateIncomingItems()`` is called.
@@ -35,7 +38,12 @@ class MockTimelineController: TimelineControllerProtocol {
         }
     }
     
-    var timelineItems: [RoomTimelineItemProtocol] = RoomTimelineItemFixtures.default
+    var timelineItems: [RoomTimelineItemProtocol] = RoomTimelineItemFixtures.default {
+        didSet {
+            callbacks.send(.updatedTimelineItems(timelineItems: timelineItems, isSwitchingTimelines: false))
+        }
+    }
+
     var timelineItemsTimestamp: [TimelineItemIdentifier: Date] = [:]
     
     private var client: UITestsSignalling.Client?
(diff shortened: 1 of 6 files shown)


ElementX/Sources/Services/Timeline/TimelineController/MockTimelineController.swift
```diff
@@ -35,7 +35,12 @@ class MockTimelineController: TimelineControllerProtocol {
         }
     }
     
-    var timelineItems: [RoomTimelineItemProtocol] = RoomTimelineItemFixtures.default
+    var timelineItems: [RoomTimelineItemProtocol] = RoomTimelineItemFixtures.default {
+        didSet {
+            callbacks.send(.updatedTimelineItems(timelineItems: timelineItems, isSwitchingTimelines: false))
```
pixlwave, line 43: Errr, I think we should update simulateIncomingItem and simulateBackPagination too otherwise we’re sending double updates for all those timeline UI tests.

ElementX/Sources/Services/Timeline/TimelineController/MockTimelineController.swift
```diff
@@ -122,7 +127,19 @@ class MockTimelineController: TimelineControllerProtocol {
     func edit(_ eventOrTransactionID: TimelineItemIdentifier.EventOrTransactionID,
               message: String,
               html: String?,
-              intentionalMentions: IntentionalMentions) async { }
+              intentionalMentions: IntentionalMentions) async {
+        callbacks.send(.messageSentOrEdited)
+        
+        timelineItems.removeLast()
+        timelineItems.append(TextRoomTimelineItem(id: .event(uniqueID: .init(UUID().uuidString), eventOrTransactionID: .eventID(UUID().uuidString)),
```
pixlwave: Does this actually test what we want to test? 🤔
- The items ID changes so the ForEach won’t come into play as it’ll be rendered as a new item.
- The change happens in 2 steps so if the remove is processed first, then even if the ID remains stable, it’ll still do a fresh render.

ElementX/Sources/Services/Timeline/TimelineController/MockTimelineController.swift
```diff
@@ -259,6 +298,8 @@ class MockTimelineController: TimelineControllerProtocol {
     // MARK: - Polls
     
     func createPoll(question: String, answers: [String], pollKind: Poll.Kind) async -> Result<Void, TimelineControllerError> {
+        callbacks.send(.messageSentOrEdited)
```
pixlwave, line 303: Hmmmm kind of annoying that we’re testing the mock here, this wouldn’t have actually helped save us from the first bug anyway would it?

pixlwave (commented):
Just a couple of thoughts.

ElementX/Sources/Services/Timeline/TimelineController/MockTimelineController.swift
stefanceriu: Ah, good catch, sorry about that

ElementX/Sources/Services/Timeline/TimelineController/MockTimelineController.swift
stefanceriu: Yeah well ¯\_(ツ)_/¯

ElementX/Sources/Services/Timeline/TimelineController/MockTimelineController.swift
stefanceriu: Not really, but I was mostly focusing on whether I can get this working to being with.

pixlwave (approved):
Nice, thank you 🙏.

PR #5654, opened 2026-05-27
Use SwiftFormat from `develop` and enable the new `indent-blank-lines` option.

pixlwave:
As there doesn't appear to be an obvious release cycle for SwiftFormat, let's install it with `--HEAD` and enable the `--indent-blank-lines` option now given the amount of AI generated code is going up and is introducing a lot of empty blank lines.


stefanceriu (approved):
Love it! 💚 Makes perfect sense to me 👍

pixlwave:
Squashed a few commits to make it easier to revert once the next SwiftFormat release is available.

PR #5662, opened 2026-05-29
Add Swift Concurrency Pro skill to the project.

pixlwave:
This PR adds the [Swift Concurrency Pro](https://github.com/twostraws/Swift-Concurrency-Agent-Skill) skill as a first test to the project. It was added using [`gh skill`](https://github.blog/changelog/2026-04-16-manage-agent-skills-with-github-cli/) (only for Copilot, I manually created the `.claude` symlink) so we can `gh skill update` in the future.


stefanceriu (approved):
Looks good, let's see it work in practice. Thanks for taking care of it!

PR #5664, opened 2026-05-29
Restrict the dynamic range of avatars to prevent them from displaying HDR content.

pixlwave:
It makes no sense to scroll through the room list/timeline and have the screen going in and out of HDR. It doesn't appear that SwiftUI does any tone-mapping to the image, but it seems such a rare occurrence that it probably doesn't matter.


stefanceriu (approved):
😎

PR #5665, opened 2026-05-29
Use the same icon in the TimelineReplyView as shown in the File timeline view.

pixlwave:
Seems we missed this one when updating all the icons.


stefanceriu (approved):
Nice one, thanks 👌

PR #5666, opened 2026-05-29
Fix empty state overlapping the banner

Velin92:
Since the empty state used an overlay this would happen if the banner was present:

Simulator Screenshot - iPhone-26 4 - 2026-05-29 at 10 27 31

To fix this issue I have removed the usage of the overlay (it's not needed that much since we are already calculating the expected size of the empty state), and in addition subtract the height of the banner so that the empty state always looks centered below it (if present)

### DIFF
--- a/ElementX/Sources/Screens/HomeScreen/View/HomeScreenContent.swift
+++ b/ElementX/Sources/Screens/HomeScreen/View/HomeScreenContent.swift
@@ -16,6 +16,8 @@ struct HomeScreenContent: View {
     @ObservedObject var context: HomeScreenViewModel.Context
     let scrollViewAdapter: ScrollViewAdapter
     
+    @State private var topSectionHeight: CGFloat = 0
+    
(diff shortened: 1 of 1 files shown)


ElementX/Sources/Screens/HomeScreen/View/HomeScreenContent.swift
```diff
@@ -130,6 +128,7 @@ struct HomeScreenContent: View {
                 }
             }
             .background(Color.compound.bgCanvasDefault)
+            .onGeometryChange(for: CGFloat.self, of: \.size.height) { topSectionHeight = $0 }
```
pixlwave, line 131: You can use `.readHeight` here as a convenience (although TIL that you can use a key path for `of:`).

pixlwave (approved):
LGTM, have you checked that it scrolls properly with accessibility dynamic type sizes?

Issue #5669, opened 2026-06-01
MAS/OAuth login on iOS stops after MAS login page loads; no token or device created


baumar73:
### Steps to reproduce

1. Use Element X iOS 26.05.3 with a self-hosted Matrix server using MAS/OAuth.
2. Homeserver: `<redacted self-hosted homeserver>`
3. Server stack:
   - Synapse 1.153.0
   - Matrix Authentication Service 1.17.0
   - Element Web behind the same domain
4. In Element X, choose the homeserver and continue with the MAS/OAuth login flow.
5. The web authentication sheet loads the MAS login page, but the login does not complete and Element X shows the same generic error as before.

#### What did you expect?

Element X should complete the MAS/OAuth login flow, exchange the authorization code, and create a new Matrix device for the iPhone.

#### What happened instead?

From the server side, the Element X login starts correctly but never progresses past the MAS login page. No iPhone access token or Synapse device is created.

Element Web on the same Mac/browser and same account works successfully against the same MAS/Synapse server, including the consent step and `/oauth2/token` exchange.

### Server-side evidence

Relevant MAS/Synapse events from the failed Element X attempt:

```text
2026-06-01T00:12:25Z GET /_matrix/client/v3/login 200
User-Agent: Element X/26.05.3 (iPhone 16 Pro Max; iOS 26.5; Scale/3.00)

2026-06-01T00:12:27Z POST /oauth2/registration 201
Client registration:
  redirect_uris: ["https://element.io/oauth/ios/io.element.elementx"]
  application_type: "native"
  client_name: "Element X"

2026-06-01T00:12:29Z GET /authorize 303
User-Agent: Mozilla/5.0 (iPhone; CPU iPhone OS 18_7 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/26.5 Mobile/15E148 Safari/604.1

2026-06-01T00:12:29Z GET /login 200
User-Agent: Mozilla/5.0 (iPhone; CPU iPhone OS 18_7 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/26.5 Mobile/15E148 Safari/604.1
```

After that there is no iPhone `POST /login`, no `/consent`, no `/oauth2/token`, and no new Synapse device for the account.

The MAS authorization grant row for this attempt remains unfulfilled and unexchanged:

```text
client_name:      Element X
application_type: native
redirect_uri:     https://element.io/oauth/ios/io.element.elementx
[...]

PR #5670, opened 2026-06-01
Add password fallback for OAuth login

baumar73:
## Summary
- keep OAuth as the default login path, but track whether the selected homeserver also supports password login
- show a manual password-login fallback on the server confirmation screen when both OAuth and password login are available
- route non-cancelled OAuth failures to password login when password login is supported

## Context
This is intended to avoid a dead end for self-hosted MAS homeservers where the native OAuth web authentication flow does not complete, while preserving OAuth as the preferred path.

Related issue: #5669

## Tests
- swift build
- xcodebuild test -project ElementX.xcodeproj -scheme UnitTests -destination id=CE2BDF04-317A-4509-97BA-B0BCE012285D -only-testing:UnitTests/ServerConfirmationScreenViewModelTests CODE_SIGNING_ALLOWED=NO

### DIFF
--- a/ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
@@ -433,8 +433,12 @@ class AuthenticationFlowCoordinator: FlowCoordinatorProtocol {
             switch await presenter.authenticate(using: oAuthData) {
             case .success(let userSession):
                 stateMachine.tryEvent(.signedIn, userInfo: userSession)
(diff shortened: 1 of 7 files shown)


pixlwave:
Sorry, this is not what we want the application to do. If the MAS flow is broken then the admin should fix it. The password flow is legacy and we don't support the same feature set when logging in that way.

Also note: As mentioned in the PR template, please read and follow the [CONTRIBUTING.md](https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md) file before working on any features.

PR #5673, opened 2026-06-01
Allow voice messages to be sent as replies

### LINKED ISSUE #5485: Allow replying to an audio message
Android has it: https://github.com/element-hq/element-x-android/issues/4053

(Proving that everything's there in the SDK to support it 😎)

Android has a feature that iOS doesn't have.

Keep my MacOS machine and being able to reply with audio messages there too.

Throw away my MacOS machine, leave the field of computer and become a farmer.

baumar73:
## Summary
- Preserves the replied-to event while recording and previewing a voice message.
- Passes the reply event ID through voice-message sending so the SDK sends it as a reply instead of a normal voice message.
- Adds regression coverage for the composer action and recorder send path.

Fixes #5485

## Tests
- `xcodebuild test -quiet -project ElementX.xcodeproj -scheme UnitTests -destination 'platform=iOS Simulator,name=iPhone 17,OS=26.5' '-only-testing:UnitTests/ComposerToolbarViewModelTests/sendsVoiceMessageAsReply()' '-only-testing:UnitTests/VoiceMessageRecorderTests/sendVoiceMessage()' CODE_SIGNING_ALLOWED=NO`


pixlwave:
Please read the [CONTRIBUTING.md](https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md) file before working on any issues.

PR #5676, opened 2026-06-01
Avoid compressing pinned message bubbles

### LINKED ISSUE #5668: [Aesthetic issue] Pinned messages are horizontally compressed without reason
This is especially noticeable when using a phone with a smaller form factor such as the iPhone 13 Mini.

To reproduce, pin a message (which contains substantial text, like a paragraph) and notice that it is horizontally compressed to fit the pushpin indicator next to it despite there already being enough room beside the message to fit the indicator.

This leads to pinned messages sometimes being very narrow and overall looking weird. In my opinion it is not good from a design perspective.

baumar73:
Fixes #5668

Summary:
- Renders the pinned indicator as an overlay outside the message bubble instead of as a sibling in the same horizontal layout.
- Preserves the existing visual spacing while allowing long pinned message bubbles to use the available width.

Tests:
- `xcodebuild test -quiet -project ElementX.xcodeproj -scheme UnitTests -destination 'platform=iOS Simulator,name=iPhone 17,OS=26.5' '-only-testing:UnitTests/TimelineViewModelTests/pinnedEvents()' CODE_SIGNING_ALLOWED=NO`
- `git diff --check`

### DIFF
--- a/ElementX/Sources/Screens/Timeline/View/Style/TimelineItemBubbledStylerView.swift
+++ b/ElementX/Sources/Screens/Timeline/View/Style/TimelineItemBubbledStylerView.swift
@@ -293,19 +293,18 @@ private extension EdgeInsets {
 private struct PinnedIndicatorViewModifier: ViewModifier {
     let isPinned: Bool
     let isOutgoing: Bool
+    private let indicatorSpacing: CGFloat = 8
     
+    @ViewBuilder
     func body(content: Content) -> some View {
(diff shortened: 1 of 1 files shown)


pixlwave:
Please read the [CONTRIBUTING.md](https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md) file before working on any issues. We can't make changes to things like this without input from Design/Product.

Issue #5075, opened 2026-02-09
EXI wrongly believe the call is not going on anymore and thus discard the notification (`The room no longer has an ongoing call`)


BillCarsonFr:
See rageshake https://github.com/element-hq/element-x-ios-rageshakes/issues/5474

```
INFO nse: The room no longer has an ongoing call, handling as push notification | NotificationHandler.swift:200 | spans: root
```

toger5:
How can we determin that it is "wrongly" thinking the call is not running anymore.
It could be that the notification arrived after the user actually left the call?

Can you provide more information how this occurs?

Issue #5074, opened 2026-02-09
Element call is sometimes not ringing


BillCarsonFr:
Creating this issue to track ringing reliability problems

## Reported Issues

- [ ] #5075

PR #5682, opened 2026-06-01
Report a phantom call to CallKit on PushKit early-return paths

danderson-cont:
There is another background crash, if a VOIP call is reported to the main app by the notification extension, but then the main app never reports the incoming call to CallKit.

There were some early-return paths without this wired up, which causes the app to be terminated with an NSInternalInconsistencyException

Possibly related: #5074, #5075

### Pull Request Checklist
- [x] I read the contributing guide.
- [x] I am aware of the etiquette.
- [x] Pull request contains a changelog label.

### DIFF
--- a/ElementX/Sources/Services/ElementCall/ElementCallService.swift
+++ b/ElementX/Sources/Services/ElementCall/ElementCallService.swift
@@ -162,19 +162,25 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
     func pushRegistry(_ registry: PKPushRegistry, didReceiveIncomingPushWith payload: PKPushPayload, for type: PKPushType, completion: @escaping () -> Void) {
         guard let roomID = payload.dictionaryPayload[ElementCallServiceNotificationKey.roomID.rawValue] as? String else {
             MXLog.error("Something went wrong, missing room identifier for incoming voip call: \(payload)")
-            completion()
+            assertionFailure("Incoming voip push is missing the room identifier")
+            reportImmediatelyEndedCall(reason: .failed, completion: completion)
             return
         }
         
         guard let rtcNotificationID = payload.dictionaryPayload[ElementCallServiceNotificationKey.rtcNotifyEventID.rawValue] as? String else {
             MXLog.error("Something went wrong, missing rtc notification event identifier for incoming voip call: \(payload)")
-            completion()
+            assertionFailure("Incoming voip push is missing the rtc notification event identifier")
+            reportImmediatelyEndedCall(reason: .failed, completion: completion)
             return
         }
         
+        let roomDisplayName = payload.dictionaryPayload[ElementCallServiceNotificationKey.roomDisplayName.rawValue] as? String
+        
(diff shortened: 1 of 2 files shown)


ElementX/Sources/Services/ElementCall/ElementCallService.swift
```diff
@@ -162,19 +162,19 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
...
+            reportImmediatelyEndedCall(completion: completion)
             return
         }
         
         guard ongoingCallID?.roomID != roomID else {
             MXLog.warning("Call already ongoing for room \(roomID), ignoring incoming push")
-            completion()
+            reportImmediatelyEndedCall(completion: completion)
             return
         }
```
stefanceriu, line 185: Realistically we shouldn't ever reach this point without those fields so perhaps it's better to just fatal error? Otherwise it might be jarring to see a missed call in your phone log without you ever being informed of it, no?

ElementX/Sources/Services/ElementCall/ElementCallService.swift
```diff
@@ -185,15 +185,15 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
...
         
         let nowDate = timeProvider.now()
         
         guard nowDate < expirationDate else {
             MXLog.warning("Call expired for room \(roomID), ignoring incoming push")
-            completion()
+            reportImmediatelyEndedCall(completion: completion)
             return
```
stefanceriu, line 208: And here, shouldn't this be reported as missed instead? Presumably the call was real but you just didn't receive the notification in time, right?

stefanceriu (approved):
Nice, thanks! I wonder if the crashes are actually only caused by the `Call expired for room` branch.

ElementX/Sources/Services/ElementCall/ElementCallService.swift
danderson-cont: Yeah, it would show in Recents since the provider has `includesCallsInRecents = true`. If we want zero phantom entries, a second `CXProvider` with `includesCallsInRecents = false` just for these is feasible. 

Sure I can add an assertionFailure for those cases where the fields should be there from the NSE

Danbana:
OK so having 2 providers to avoid showing a call in the Recents was a dead end in testing. Almost done here though

ElementX/Sources/Services/ElementCall/ElementCallService.swift
Danbana: changed this one to .unanswered

Danbana:
@stefanceriu Heads up that I force-pushed a reworked version. Each early-return path now reports with a reason that fits the case. I also tried a second CXProvider to keep the duplicate/malformed reports out of the call history, but on device that suppressed call logging for the whole app, so this version uses the single provider and just labels the calls. Tested on device: normal calls, missed calls, and a duplicate ring during an active call all behave correctly.

stefanceriu (approved):
Nice one, thank you very much! 🚀

PR #5686, opened 2026-06-02
Various mocking improvements

stefanceriu:
* Convert the old MockTimelineController to the generated `TimelineControllerMock`
* Remove all underlyingFoo mock property usage because they are not necessary
* Switch the mock stencil templates to use NSLock instead of dispatch queues


ElementX/Sources/Mocks/TimelineControllerMock.swift
```diff
@@ -12,318 +12,238 @@ import Combine
...
-        return .success(())
+    func update(paginationState: TimelinePaginationState) {
+        self.paginationState = paginationState
+        callbacks.send(.paginationState(paginationState))
     }
     
-    func endPoll(pollStartID: String, text: String) async -> Result<Void, TimelineControllerError> {
-        if let timelineProxy {
-            _ = await timelineProxy.endPoll(pollStartID: pollStartID, text: text)
+    /// Configures `paginateBackwardsRequestSizeClosure` to simulate back pagination from the given response queue,
```
pixlwave, line 220: Smart!

ElementX/Sources/Mocks/TimelineControllerMock.swift
pixlwave, line 1: There's a bunch of

```
self.whatever
```

in here where we can remove the `self.`

pixlwave (approved):
Very nice 👏

PR #5687, opened 2026-06-02
Update ListRowLabel with the correct icon spacing on iOS 26.

pixlwave:
Not sure if this was missed in #5029 or the Figma was changed afterwards but I noticed while working on the User Status views that the label spacing in `ListRowLabel` should be smaller on iOS 26.

Loooots of snapshots 🙈


stefanceriu (approved):
👌

PR #5695, opened 2026-06-03
Fix for the timestamp state change creating clipping

Velin92:
When the timestamp changed after the bubble was rendered, the content of the formatted body wasn't catching any update in the content, which created clipping if the new timeastamp contained any aditional icon (like the normal -> error state transition) so we essentially include in the last `plainText` block the size as part of the id itself.


pixlwave (approved):
😎

PR #5697, opened 2026-06-03
Various testing tweaks

stefanceriu:
* Switch to using proper SDK Mocks everywhere.
* Rename RoomTimelineItemFixtures to just TimelineFixtures

Left over from https://github.com/element-hq/element-x-ios/pull/5663


pixlwave (approved):
LGTM, I wonder if we should add a SwiftLint rule that detects the usage of `.init(noHandle:` or similar pattern but not massively important.

stefanceriu:
> we should add a SwiftLint rule that detects the usage of .init(noHandle)

Don't think it's really worth it tbh, we only use these things once in a blue moon.

PR #5698, opened 2026-06-03
Do not show membership/profile events in public rooms

bxdxnn:
Port of https://github.com/element-hq/element-x-android/pull/6360
Sorry for the AI slop, didn't test it, though there are unit tests. This is pretty simple so I assume it should work as-is without problems.
The expected behavior is: "user joined/left/changed display name/avatar" are not shown if the room's join rule isn't `invite`, the current code seems to do it.

### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Services/Timeline/Fixtures/RoomTimelineItemFixtures.swift
+++ b/ElementX/Sources/Services/Timeline/Fixtures/RoomTimelineItemFixtures.swift
@@ -273,7 +273,7 @@ extension TimelineItemProxy {
 private extension TimelineItemProxy {
     func buildRoomTimelineItem() -> RoomTimelineItemProtocol {
         guard case .event(let eventProxy) = self,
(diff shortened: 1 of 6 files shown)


pixlwave:
Sorry, this should be done in the SDK, we shouldn't be filtering out things like this at the application layer. There's already a filtering mechanism in place that could be used, although I've been told it would need extending slightly. It would also mean that we only have to maintain the code (and tests) in a single place and have better guarantees about the behaviour on both platforms.

bxdxnn:
@pixlwave that was also my suspicion, but it seems like it was OK in Android? @jmartinesp

jmartinesp:
After taking a closer look, it seems like this could *mostly* be done in the SDK as is now, it would just need some extra filtering for the type of membership changes to exclude using `TimelineItemFilter`. If I have the time I'll come back to this and try to make it work through the SDK.

PR #5701, opened 2026-06-04
Automatically adjust the ElementCall PiP when changing orientations

stefanceriu:
Shouldn't make any difference until the change lands in EC itself.

### DIFF
--- a/ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
+++ b/ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
@@ -88,7 +88,7 @@ private struct CallView: UIViewRepresentable {
         init(viewModelContext: CallScreenViewModel.Context) {
             self.viewModelContext = viewModelContext
             pictureInPictureViewController = AVPictureInPictureVideoCallViewController()
-            pictureInPictureViewController.preferredContentSize = CGSize(width: 1920, height: 1080)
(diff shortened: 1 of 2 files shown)


Velin92 (approved):

ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
```diff
@@ -192,6 +192,16 @@ private struct CallView: UIViewRepresentable {
                 viewModelContext?.send(viewAction: .outputDeviceSelected(deviceID: deviceID))
             case .onBackButtonPressed:
                 viewModelContext?.send(viewAction: .navigateBack)
+            case .onPipMediaOrientationUpdate:
+                guard let orientation = message.body as? String else { return }
+                switch orientation {
+                case "portrait":
+                    pictureInPictureViewController.preferredContentSize = PiPSize.portrait.size
+                case "landscape":
+                    pictureInPictureViewController.preferredContentSize = PiPSize.landscape.size
```
Velin92, line 201: ugh this sucks, there's no better way? I guess we can't do better considering that this is a message from the js context.

stefanceriu: We only care about the aspect ratio and for the size not to be too small. Couldn't come up with anything better but I'm open to suggestions 🤷‍♂️

PR #5564, opened 2026-05-08
Add editorconfig to enforce style guides naturally in the editor

mredig:
Adds an `.editorconfig` file to enforce the project's existing code style preferences in editors that support it. No Xcode preference changes are required -- the file simply overrides settings per-project.

For all files:
- Line endings: LF
- Insert final newline

For Swift files:
- Space indentation, 4 spaces
- Trailing whitespace trimming disabled (preserves indentation on whitespace-only lines, per project convention)

### Pull Request Checklist

- [x] I read the [contributing guide](https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md).
- [ ] Pull request contains a [changelog label](https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md#changelog).

**UI changes have been tested with:**
- [ ] iPhone and iPad simulators in portrait and landscape orientations.
- [ ] Dark mode enabled and disabled.
- [ ] Various sizes of dynamic type.
- [ ] Voiceover enabled.

PR #5703, opened 2026-06-04
Remove the .editorconfig file

pixlwave:
Reverts #5564.

After using Xcode for a while I've noticed that `trim_trailing_whitespace = false` isn't actually what we wanted. We run with Xcode trimming non-blank lines but leaving blank lines alone and .editorconfig does not support this nuance:

Screenshot 2026-06-04 at 12 13 45 pm

We also have a proper SwiftFormat rule for indentation now so this is no longer an issue and as far as I can see, we're covered for the rest of this anyway so it doesn't seem worth keeping.


stefanceriu:
Oh I was wondering why xcode started doing weird things to trailing spaces, didn't make the connection though 🤦‍♂️ 
Thanks, kill it with 🔥 !

mredig:
The removal rationale only addresses `trim_trailing_whitespace`, but that was one of four settings. The others (LF line endings, final newline, 4-space indentation) configure the editor at typing time, not at commit time - SwiftFormat doesn't help there.

The practical effect is that contributors whose editors default to tabs see constant indentation warnings while working. The project already has a high warning count; adding noise that SwiftFormat will silently fix later makes warnings even less useful as a signal.

Wouldn't it make more sense to remove just the offending setting rather than the whole file?

pixlwave:
> contributors whose editors default to tabs see constant indentation warnings while working.

Ok good point, didn't think about that. Restored in #5712

PR #5705, opened 2026-06-04
Client pausing and resuming timing fixes

stefanceriu:
While the pausing and resuming feature works properly we did notice problems when deeplinking from notifications. Depending on timing, the underlying `handleAppRoute` would be handled before the `applicationWillEnterForeground` and implicitly the `startSync` call, which would in turn make the route (room timeline) encounter store closed errors.

This series of patches exposes the async nature of the startSync/resumeClient calls from within the ClientProxy, changes the semantics to pause/resume and ensure that everything gets started back up before routes are handled.

### DIFF
--- a/ElementX/Sources/Application/AppCoordinator.swift
+++ b/ElementX/Sources/Application/AppCoordinator.swift
@@ -41,8 +41,10 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
                 configureElementCallService()
                 configureNotificationManager()
                 observeUserSessionChanges()
-                startSync()
-                Task { await appHooks.configure(with: userSession) }
+                Task {
+                    await startSync()
+                    await appHooks.configure(with: userSession)
+                }
             }
         }
     }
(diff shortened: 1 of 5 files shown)


ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -41,8 +41,10 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
                 configureElementCallService()
                 configureNotificationManager()
                 observeUserSessionChanges()
-                startSync()
-                Task { await appHooks.configure(with: userSession) }
+                Task {
+                    await resumeClientServices()
+                    await appHooks.configure(with: userSession)
+                }
```
pixlwave, line 47: I wonder if these should be 2 separate tasks, I don't think the appHooks particularly needs a running client, it just needs to know the userSession.

pixlwave (approved):

ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -1060,24 +1072,24 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
     
     // MARK: - Application State
     
-    private func stopSync(isBackgroundTask: Bool, completion: (() -> Void)? = nil) {
+    private func pauseClientServices(isBackgroundTask: Bool, completion: (() -> Void)? = nil) {
```
pixlwave, line 1075: FWIW, I *think* we have everything in Swift now to get rid of this completion and make it async too.

ElementX/Sources/Application/AppCoordinator.swift
stefanceriu: I figured resuming services is/should aways be fast and it's safer this way.

ElementX/Sources/Application/AppCoordinator.swift
stefanceriu: Perhaps but I don't trust that thing enough to make that change in this PR 😁

PR #5707, opened 2026-06-04
Hide user dot and track own user marker in Live Location mode.

### LINKED ISSUE #5681: [Task] Hide current location dot when sharing location

Velin92:
This PR essentially does 2 things:
- Hides the dot for the current user location in case the user is actually sharing their live location and visualising the LLS map, this only applies when the user is currently sharing their live location from the current device, in all other cases their location dot will be displayed.
- When the dot is hidden in LLS mode, since the user is sharing their live location on the current device, the center to user button now implements a new tracking logic, that will follow the actual user marker (since the dot is now hidden), the tracking logic has been implemented with the aid of AI, and essentially is a customisation done through the usage of the APIs provided by MapLibre.

fixes #5681

### DIFF
--- a/ElementX/Sources/Other/MapLibre/MapLibreMapView.swift
+++ b/ElementX/Sources/Other/MapLibre/MapLibreMapView.swift
@@ -74,7 +74,9 @@ struct MapLibreMapView: UIViewRepresentable {
         }
         
         // If the center coordinate was updated externally (not by the map itself), move the map.
-        if let newCenter = mapCenterCoordinate,
+        // Not applied while following a marker, where the camera is driven by the marker's position.
+        if showsUserLocationMode.followedMarkerID == nil,
+           let newCenter = mapCenterCoordinate,
            newCenter != context.coordinator.lastReportedCenter {
             context.coordinator.lastReportedCenter = newCenter
             mapView.setCenter(newCenter, animated: true)
@@ -86,6 +88,8 @@ struct MapLibreMapView: UIViewRepresentable {
         updateAnnotations(in: mapView)
         
         showUserLocation(in: mapView)
+        
+        context.coordinator.updateMarkerFollowing(in: mapView, markerID: showsUserLocationMode.followedMarkerID)
     }
     
     func makeCoordinator() -> Coordinator {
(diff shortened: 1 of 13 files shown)


ElementX/Sources/Other/MapLibre/MapLibreMapView.swift
```diff
@@ -182,12 +187,62 @@ extension MapLibreMapView {
...
+            // so that it doesn't cancel the transition.
+            mapView.setCenter(annotation.coordinate,
+                              zoomLevel: mapView.zoomLevel,
+                              direction: -1, // negative value keeps the current direction
+                              animated: true) { [weak self, weak annotation, weak mapView] in
+                guard let self, let annotation, let mapView, annotation === followedAnnotation else { return }
+                
+                // Mirror every animated coordinate update to keep the camera as smooth as the marker.
+                followedAnnotationObservation = annotation.observe(\.coordinate) { [weak mapView] annotation, _ in
+                    mapView?.setCenter(annotation.coordinate, animated: false)
```
stefanceriu, line 236: > Mirror every animated coordinate update

Doesn't that mean `animated` is supposed to be true? 🤔

stefanceriu (approved):
Looks good to me, thanks! 👍

ElementX/Sources/Other/MapLibre/MapLibreMapView.swift
Velin92: for this one we want it to be false, becuse since the map markers are updated through the `CoordinateAnimator` which updates through interpolation at the same framerate of the screen rendering (about 60 times per seconds usually) to create the an aimation effect, we want to essentially use those same updates to sync the center with the marker, (check the CoordinateAnimator for more info). If we added animated to true, what would happen is that the MapLibre SDK would try to also interpolate the set center function, even if not required since is all done through the observation of the marker.

Velin92:
@stefanceriu @stefanceriu  if you're okay to make things a bit faster and also use a bit less code, I could try to replace the array of `LocationAnnotation` with an `OrderedDictionary` and key them by their ID, should be cleaner and have the access to the arrays be faster.

ElementX/Sources/Other/MapLibre/MapLibreMapView.swift
stefanceriu: Alrighty, makes sense 👍

stefanceriu:
> @stefanceriu @stefanceriu if you're okay to make things a bit faster and also use a bit less code, I could try to replace the array of `LocationAnnotation` with an `OrderedDictionary` and key them by their ID, should be cleaner and have the access to the arrays be faster.

Yeah that sounds like a good idea if you're willing to give it a shot, but perhaps for a separate PR?

PR #5710, opened 2026-06-05
Added a comments section to agents.md and use caveman

Velin92:
This should make AI less verbose, also rewritten the agents.md to use caveman so that we can save even more tokens.


pixlwave (approved):

stefanceriu (approved):
🪨

PR #5712, opened 2026-06-05
Bring back the .editorconfig file without the trim rule.

pixlwave:
Undo #5703 and just remove the `trim_trailing_whitespace` line.


stefanceriu (approved):
👌

PR #5713, opened 2026-06-05
Migrate the share extension to swift 6

### LINKED ISSUE #4993: [Task] ShareExtension - Swift 6 Migration

Velin92:
fixes #4993

Very simple, just made functions that could make self cross actors static, and to keep the behaviour use a different bg actor from the caller actor, marked the functions with `@concurrent`


ElementX/Sources/Other/Extensions/NSItemProvider.swift
```diff
@@ -108,12 +99,41 @@ extension NSItemProvider {
...
         
+        return await Self.writeData(shareData,
+                                    filename: filename,
+                                    withinAppGroupContainer: withinAppGroupContainer,
+                                    providerDescription: providerDescription)
+    }
+    
+    // The next 2 methods are static as NSItemProvider isn't Sendable. @concurrent
+    // keeps the image encoding and file writes off the caller's actor, which is
+    // the main one when sharing from the ShareExtension.
```
pixlwave: Do we need to document how `@concurrent` works?

ShareExtension/SupportingFiles/target.yml
```diff
@@ -75,6 +75,8 @@ targets:
         MARKETING_VERSION: $(MARKETING_VERSION)
         CURRENT_PROJECT_VERSION: $(CURRENT_PROJECT_VERSION)
         DEVELOPMENT_TEAM: $(DEVELOPMENT_TEAM)
+        SWIFT_VERSION: 6.2
+        SWIFT_APPROACHABLE_CONCURRENCY: YES
```
pixlwave, line 79: Out of interest, what's the plan for setting default actor isolation like we have [in Compound](https://github.com/element-hq/element-x-ios/blob/097d055dbe866ecabb8c4fd53832b465a2a24018/compound-ios/Package.swift#L39-L41)?

pixlwave (approved):

ShareExtension/SupportingFiles/target.yml
Velin92: For now I am following Apple's default project initialisation. A default ShareExtension target is nonisolated by default, same the NSE. While the main app is `@MainActor` isolarted by default. Compound makes sense, it's a UI module, so we keep it `@MainActor` isolated.

PR #5726, opened 2026-06-11
Add max selections support for polls

### LINKED ISSUE #2756: Support for the `max_selections` poll parameter
It is currently not possible to select multiple options on a poll.

It should be possible to select multiple options if `max_selections` allows for it.
Selecting another option shouldn't unselect previously selected options. Instead, selected options may be unselected by tapping them again.

**Additional context**
See also: https://github.com/element-hq/element-web/issues/23319

svetoslav-sportinno:
### Pull Request Checklist

- [x] iPhone simulator in portrait orientation.
- [ ] Voiceover enabled.

### Notes

- Adds `Votes allowed per person` to poll creation and editing.
- Sends multi-answer poll responses when `maxSelections` is greater than 1.
- Verified with focused unit tests and simulator testing.

Related:
- https://github.com/element-hq/element-meta/issues/2694
- https://github.com/element-hq/element-x-android/pull/6477

Fixes #2756


ElementX/Sources/Screens/CreatePollScreen/PollFormScreenViewModel.swift
```diff
@@ -102,15 +108,23 @@ class PollFormScreenViewModel: PollFormScreenViewModelType, PollFormScreenViewMo
...
         case .success:
             actionsSubject.send(.close)
         case .failure:
             userIndicatorController.submitIndicator(UserIndicator(title: L10n.errorUnknown))
         }
     }
     
+    private func clampMaxSelections() {
+        state.bindings.maxSelections = clampedMaxSelections
+    }
```
Velin92: this function feels a bit uselss, is used only once, and is essentially a 1 line function that saves only one line. Can be removed.

ElementX/Sources/Screens/CreatePollScreen/PollFormScreenModels.swift
```diff
@@ -78,6 +78,7 @@ enum PollFormMode: Hashable {
 struct PollFormScreenViewStateBindings: Equatable {
     var question = ""
     var options: [Option] = [.init(), .init()]
+    var maxSelections = 1
```
Velin92: In theory this value could be reactively be updated in two ways toi avoid having to use a dedicated function like campMaxSelection().
1. Could be to have the viewmodel listen to the viewstate with a subscription to the options count, and always clamp it whenever the number of options count changes
2. With a did or will set on options (but will not work when initialising, only after the first change)

ElementX/Sources/Screens/CreatePollScreen/PollFormScreenViewModel.swift
```diff
@@ -102,15 +108,23 @@ class PollFormScreenViewModel: PollFormScreenViewModelType, PollFormScreenViewMo
...
+    
+    private var clampedMaxSelections: Int {
+        min(max(state.bindings.maxSelections, 1), max(state.bindings.options.count, 1))
+    }
```
Velin92: I would probably move this to the model, but I don't this is actually necessary if we use this approach: https://github.com/element-hq/element-x-ios/pull/5726/changes#r3395028714 to make sure the value is always in sync with the options

ElementX/Sources/Screens/Timeline/View/Polls/PollView.swift
```diff
@@ -100,21 +100,37 @@ struct PollView: View {
...
+            return true
+        }
+        
+        if poll.maxSelections == 1 {
+            return option.isSelected
+        }
+        
+        if option.isSelected {
+            return poll.options.filter(\.isSelected).count == 1
+        }
```
Velin92, line 133: can you add a small comment for the reason of this check? It's a bit hard to read

Velin92 (changes requested):
Generally looks good added a few comments., but it would also need to have some preview tests.
You need to add an iPhone SE 3rd Gen simulator with iOS 26.5 (setup with only the US language and region) and then run the preview tests scheme over it, it will automatically generate the snapshots  (and if an existing snapshot fails, you need to check if is either an error or a change that requires to delete the existing snapshot and generate it again).
Remember to have git-lfs installed before committing any snapshot.

svetoslav-sportinno:
Addressed all review comments and added the required preview snapshots.

Preview tests were run locally on iPhone SE (3rd generation), iOS 26.5, with the required simulator setup:

```
swift run tools ci preview-tests
```

The preview tests passed. Note: local coverage/JUnit export logged missing `xcresultparser`, but the preview test run itself passed and the wrapper finished successfully.

Velin92:
> Addressed all review comments and added the required preview snapshots.
> 
> Preview tests were run locally on iPhone SE (3rd generation), iOS 26.5, with the required simulator setup:
> 
> ```
> swift run tools ci preview-tests
> ```
> 
> The preview tests passed. Note: local coverage/JUnit export logged missing `xcresultparser`, but the preview test run itself passed and the wrapper finished successfully.

Sorry but all the preview tests need to be regenerated.
As I mentioned in my comment:
> You need to add an iPhone SE 3rd Gen simulator with iOS 26.5 (setup with only the US language and region)

This is something that needs to be done manually on your simulator, the general settings should look like this:
simulator_screenshot_ADE970D3-4981-4682-8B83-7D0C76C6D4A8


This is required to maintain the right localisation for dates which otherwise would fail on CI.

svetoslav-sportinno:
Fixed the simulator localisation setup and regenerated the affected preview snapshots.

The simulator had Bulgarian retained as a second preferred language even though the command-line defaults reported only English. I removed the Bulgarian language/keyboard state and verified the iPhone SE (3rd generation), iOS 26.5 simulator is configured with:

- English as the only preferred language
- United States region
- Gregorian calendar
- Fahrenheit
- US measurement system
- Sunday as the first day of the week

Then I regenerated the snapshots and verified them with:

```
swift run tools ci preview-tests
```

Both preview test passes now succeed locally.

Velin92:
there is still something weird... like the am in the timestamp (since the test plan would run the english UK language), and seems like it's ignored. At this point don't worry with it, I'll take care of the rest, and generate the Preview tests myself and fix the conflict, no worries

Velin92 (approved):

PR #5733, opened 2026-06-12
Swift 6 Migration

Velin92:
This PR is not supposed to be merged as it is. It's more of an experiment to define a path to actually fully migrate the main app to swift 6 with default main actor isolation on, and has been through the usage of AI. It is indeeded as a study case of possble approaches, to use to actually complete the migration, it builds and runs fine, but has made some of the tests a bit flaky, so there is definitely room for some improvements. Some of these commits/changes could be taken individually, while not enabling swift 6 yet to just ease in the migration process and test some of its parts.

# Swift 6.2 Concurrency Migration — Commit-by-Commit Changelog

This PR migrates the **ElementX** app target to Swift 6.2 with *default `MainActor` isolation* and *approachable concurrency* (`nonisolatedNonsendingByDefault`), matching the NSE / Share Extension (which already use 6.2, but without the main-actor default since they don't need it).

The work is split into atomic commits, each one a single coherent category of change, ordered for **reviewability by concern**.

> **Note on intermediate commits:** because the very first commit flips the build settings, individual commits are *not* independently buildable — this is inherent to a default-isolation migration (nothing compiles until every category lands).

> **Why so much had to change:** with `SWIFT_DEFAULT_ACTOR_ISOLATION = MainActor`, every type now defaults to main-actor isolation. Anything that genuinely runs off the main actor — Rust SDK wrappers, value-type models, background services, and especially the NSE-shared files (the `ElementX` target also compiles `NSE/Sources` for testing, and the NSE has *no* main-actor default) — must now be *explicitly* annotated `nonisolated` / `Sendable`, or the two targets disagree and the build fails.

---

## 1. `Enable Swift 6.2 with default main actor isolation for the app target`
*5 files*

The root change that triggers everything else.

- **`ElementX/SupportingFiles/target.yml`**:
  ```yaml
  SWIFT_VERSION: 6.2
  SWIFT_APPROACHABLE_CONCURRENCY: YES
  SWIFT_DEFAULT_ACTOR_ISOLATION: MainActor
  ```
- **`ElementX.xcodeproj/project.pbxproj`**, **schemes** (`ElementX.xcscheme`, `UnitTests.xcscheme`), **`Package.resolved`** — regenerated/aligned.

**Problem solved:** turns on Swift 6 strict-concurrency checking for the app. Everything below is a consequence of this single switch.

---

## 2. `Update the code generation templates for Swift 6`
*6 files (incl. ~2,250 lines of regenerated output)*

The generated code has to be born Swift-6-correct, so the templates are fixed once and the output regenerated.

- **`Tools/Sourcery/AutoMockable.stencil`**:
  - mock classes → `nonisolated … : @unchecked Sendable`
  - stored vars → `nonisolated(unsafe)` (NSLock-protected / test-sequential)
  - **async mock methods → `@concurrent`** — so async mocks *suspend like the real implementations*. Without it, under caller-isolation-by-default the mocks complete inline and intermediate view-model states become unobservable in tests (this caused the `NotificationSettings…` flakes).
- **`Tools/SwiftGen/Templates/Strings/*.stencil`** (×2): generated `L10n` / `UntranslatedL10n` enums + extensions → `nonisolated`.
- Regenerated: **`GeneratedMocks.swift`**, **`Strings.swift`**, **`Strings+Untranslated.swift`**.

---

## 3. `Make the data layer and services nonisolated and Sendable`
*201 files*

The bulk of the migration: annotating everything that lives off the main actor. **~178 hand-authored types** + extensions + protocol conformances.

### Types made `nonisolated` — by subsystem
- **Timeline data layer (~66 types):** `RoomTimelineItemFactory`, the string builders (`RoomEventStringBuilder`, `RoomMessageEventStringBuilder`, `RoomStateEventStringBuilder`), every timeline item + content type (`Text`/`Emote`/`Notice`/`Image`/`Video`/`Audio`/`File`/`Location`/`LiveLocation`/`Sticker`/`Poll`/`Redacted`/`Encrypted`/`CallInvite`/`CallNotification`/`State`/`Separator`/`ReadMarker`/`PaginationIndicator`/`TimelineStart`/`Unsupported`), plus `TimelineItemIdentifier`, `TimelineItemSender`, `TimelineItemProxy`, `EventTimelineItemProxy`, `TimelineItemReply`/`ReplyDetails`, `TimelineItemSendFailure`, `TimelineItemDeliveryStatus`, `TimelineItemProperties`, `TimelineItemThreadSummary`, `TimelineItemDebugInfo`, `AggregatedReaction`, `ReactionSender`, `EventBasedMessageTimelineItemContentType`, `TimelineEventContent`. *(Built off-main in a detached task, so the whole graph must be nonisolated.)*
- **Rust SDK proxy wrappers:** `RoomMemberProxy`, `KnockRequestProxy`, `UserIdentityProxy`, `MediaSourceProxy`, `MediaFileHandleProxy`, `CheckCodeSenderProxy`, `SendHandleProxy`, `GeoURI`, `ImageInfoProxy`/`VideoInfoProxy`/`MediaInfoProxy`. *(Immutable, internally-synchronized.)*
- **Settings & preferences (NSE-shared):** `AppSettings`, `UserPreference<T>`, `UserDefaultsStorage<Value>`, `RemotePreference<T>`, `TrackedUserDefaults`, `RageshakeConfiguration`, `TraceLogPack`, `RestorationToken`, `SessionDirectories`, and the settings enums (`AppAppearance`, `LogLevel`, `AnalyticsConsentState`, `RoomListActivityVisibility`, `AudioPlaybackSpeed`, `NotificationTone`, `FrequentlyUsedEmoji`, `LiveLocationSession`, `MapTilerSettings`, `AnalyticsConfiguration`, `AppBuildType`).
- **Logging / infra (NSE-shared):** `MXLog`, `Tracing`, `ExtensionLogger`, `Target`, `InfoPlistReader`, `BootDetectionManager`, `NotificationConstants`.
- **NSE notification stack (shared):** `NotificationServiceExtension`, `NotificationHandler`, `NSEUserSession`, `NotificationContentBuilder`, `NotificationItemProxy`, `EmptyNotificationItemProxy`, `NotificationIcon`.
- **HTML/text parsing:** `AttributedStringBuilder`, `MentionBuilder`, `PlainMentionBuilder`, `PhishingDetector`, `MatrixEntityRegex`, `TextParsingMatch`.
- **Audio/voice:** `AudioConverter`, `VoiceMessageCache`, `NotificationToneManager`, `EstimatedWaveform`, `MediaPlayerState`. *(`AudioRecorder` and `AudioPlayer` are in the targeted-fixes commit.)*
- **Other/utilities:** `CurrentValuePublisher`, `Pagination`/`PaginationState`, `KeychainController`, `MediaProvider`, `EffectsScene`, `CollapsibleReactionLayout`, `CollapsibleTimelineItem`, `TimelineBubbleLayout`, the Pills value types (`PillTextAttachment`, `PillTextAttachmentData`, `PillType`, `PillUtilities`, `TransparentTextAttachment`), `JoinRule`/`AllowRule`, `RoomPermissions`/`RoomPowerLevel`/`RoomRole`, `RoomMemberDetails`, `BugReport`.

### Protocols conformed to `Sendable`
`AppHooksProtocol`, all 9 hook protocols (`AppSettingsHook`, `BugReportHook`, `ClientBuilderHook`, `CompoundHook`, `DeveloperOptionsScreenHook`, `OAuthPresenterHook`, `RoomScreenHook`, `TracingHook`), `CommonSettingsProtocol`, `AttributedStringBuilderProtocol`, `MentionBuilderProtocol`, `NotificationToneManagerProtocol`, `AudioConverterProtocol`, `AudioSessionProtocol`, `AudioRecorderProtocol`, `MediaLoaderProtocol`, `MediaProviderProtocol`, `KnockRequestProxyProtocol`, `RoomMemberProxyProtocol`, `RoomProxyProtocol`, `UserSessionProtocol`, `RoomTimelineItemProtocol`, `RoomTimelineItemFactoryProtocol`, `TimelineProxyProtocol`, `UserIdentityProxyProtocol`, `VoiceMessageCacheProtocol`, `NotificationItemProxyProtocol`, plus (for the `@concurrent` mocks) `SpaceRoomListProxyProtocol`, `RoomPreviewProxyProtocol`, `RoomPowerLevelsProxyProtocol`, `RoomLiveLocationServiceProtocol`, `RoomNotificationSettingsProxyProtocol`.

**Justified unsafe escapes carried by this commit** (they're part of making these types nonisolated):
- `AppSettings` → `@unchecked Sendable` + 41 × `nonisolated(unsafe)` on the `@UserPreference` stored vars (UserDefaults-backed, thread-safe).
- `AudioRecorder` → `@unchecked Sendable` + 1 × `nonisolated(unsafe)` (serial-queue-confined realtime audio).
- `CurrentValuePublisher` → `@unchecked Sendable` (thin wrapper over an already-Sendable subject).

**Problem solved:** lets these off-main types and their existentials cross isolation boundaries (into `Task`s, returned across `await`) without the compiler rejecting them.

---

## 4. `Adopt the new isolation across call sites, mocks and signatures`
*42 files*

The "glue" that makes the annotated data layer actually compile end-to-end — distinct from the type declarations themselves.

- **Call-site & signature adjustments:** services/proxies whose method signatures, closures or call sites needed updating for the new isolation (`ClientProxyProtocol` gains a `@Sendable` filter on `recentlyVisitedRooms`, `TimelineProxy`, `JoinedRoomProxy`, `SpaceServiceProxy`, `RoomDirectorySearchProxy`, `SecureBackupController`, `AuthenticationService`, `UserSessionStore`, etc.).
- **`@MainActor` *additions*** where a member in an otherwise-nonisolated type must stay on main (`SessionVerificationControllerProxy`, `RoomThreadListServiceProxy`, `KnockRequestsBannerView`, `KnockRequestCell`, `Avatars.scaledValue/scaledSize`, several view models).
- **Hand-written mock isolation:** `@MainActor` convenience inits / `nonisolated` on `ClientProxyMock`, `UserSessionMock`, `TimelineControllerMock`, `RoomPreviewProxyMock`, `RoomLiveLocationServiceMock`, `RoomNotificationSettingsProxyMock`, `RoomPowerLevelsProxyMock`, `SpaceRoomListProxyMock`, `SessionVerificationControllerProxyMock`.

---

## 5. `Remove redundant @MainActor annotations`
*136 files*

With the target now main-actor-by-default, explicit `@MainActor` on view models, coordinators, views and their protocols is redundant and removed.

**Problem solved:** removes noise and avoids the false impression that these types are *specially* isolated when they simply inherit the default. (Net across the PR: −175 `@MainActor` removed, +77 re-added where genuinely needed — the re-adds live in commits 4 and 6.)

---

## 6. `Fix concurrency hotspots surfaced by the migration`
*22 files*

The behavioural / structural fixes — the parts that required real code changes rather than annotations.

- **`RoomSummaryProvider`** — replaced the `serialDispatchQueue` + `DispatchSemaphore` sync-over-async bridge with two `AsyncStream` consumers; `fetchRoomDetails`/`buildRoomSummary`/`processDiff`/etc. made `async`. **Fixes the blank-screen main-thread deadlock** that appeared once diff processing moved onto the main actor.
- **`SDKListener`** — new `onMainActor` factory: an `AsyncStream` FIFO bridge so SDK callbacks (fired from arbitrary threads) are delivered one-at-a-time on the main actor. Replaced 17 call sites.
- **`AudioPlayer`** — KVO observers read values on the KVO thread then hop via `Task { @MainActor }` (only `Sendable` values cross); `isolated deinit`.
- **`ClientProxy`**, **`VoiceMessageRecorder`** — `isolated deinit` (touch main-actor state during teardown).
- **`ElementCallService`** — push completion hopped to `Task { @MainActor }` with a `@MainActor @Sendable` wrapper (push registry callback is main-bound, CallKit's is not).
- **`AppMediator`(+Protocol)** — `beginBackgroundTask` handler → `(@MainActor @Sendable () -> Void)?`.
- **`EmojiProvider`** — `@preconcurrency import Emojibase` (unannotated package).
- **`Observable.observe()`** — rewritten as `@MainActor func observe<Value: Sendable>` with a dedicated change-signal stream + synchronous initial read (closes a window where intermediate values were missed under load).
- **`MainActor.assumeIsolated`** at provably-main-but-unannotated Apple call sites: `MapLibreMapView` (KVO), `TimelineMediaPreviewController` (main-runloop timer), `BugReportScreen` (PhotosPicker label), `LoadableImage` (Kingfisher), `PillAttachmentViewProvider` ×2, `UserAgentBuilder` ×2 (+ a `DispatchQueue.main.sync` hop for the off-main background path), `ShareToMapsAppActivity`, `SpaceAddRoomsScreen`.
- **Justified `nonisolated(unsafe)`:** `QRCodeScannerView` ×2 (`AVCaptureSession` start/stop, documented off-main), `CallScreen` (WKWebView JS result), `AppCoordinator` (`BGAppRefreshTask` handler), `PillAttachmentViewProvider` ×2 (TextKit self-shadow).
- **`compound-ios/.../ListRow.swift`** — `case custom(@MainActor () -> CustomView)`. ⚠️ Separate package; needs its own commit/PR upstream.

---

## 7. `Fix the unit tests for Swift 6 concurrency`
*16 files (UnitTests target)*

- **`WaitingConfirmation.swift`** — routes the test body through a `runBody(on isolation: isolated (any Actor)?, …)` hop. Swift Testing's `confirmation` runs its closure on the concurrent executor *regardless* of the `isolation:` parameter; a body that synchronously poked main-actor state crashed with `EXC_BREAKPOINT` in `_dispatch_assert_queue_fail`. The hop runs the body on the caller's actor. Public API unchanged.
- **`ElementCallServiceTests`** — 4 sites converted `confirmation` → `waitForConfirmation`.
- **`LocationSharingScreenViewModelTests`** — `startLiveLocation…` rewritten to assert synchronously.
- Mechanical adjustments to the other suites for the new isolation / `async` signatures and the `UserPreference` API.

---

## 8. `Make the generated Secrets enum nonisolated`
*1 file*

- **`Secrets/Secrets.swift`** → `nonisolated enum Secrets`.

This file carries the git `assume-unchanged` bit (so developers' real secret values aren't committed); the committed **placeholder** lacked `nonisolated`, which is why **CI failed to build `AppSettings` even though local builds passed** — CI checks out the placeholder directly. The committed placeholder values are untouched; only the annotation is added.

---

## Escape-hatch summary (all justified)

| Mechanism | Count | Where |
|---|---|---|
| `@unchecked Sendable` (hand-written) | 4 | `AppSettings`, `AudioRecorder`, `CurrentValuePublisher`, `SignallingTimelineItemProviderMock` |
| `nonisolated(unsafe)` (hand-written) | 48 | `AppSettings` (41), `QRCodeScannerView` (2), `PillAttachmentViewProvider` (2), `AudioRecorder` (1), `CallScreen` (1), `AppCoordinator` (1) |
| `MainActor.assumeIsolated` | 10 sites | listed in commit 6 |

Everything else is `nonisolated` / `Sendable` / `@MainActor` / `isolated deinit` / `@concurrent` — checked, race-free annotations rather than escapes.

---


ElementX/Sources/Screens/CreateRoomScreen/CreateRoomScreenViewModel.swift
```diff
@@ -303,9 +303,10 @@ class CreateRoomScreenViewModel: CreateRoomScreenViewModelType, CreateRoomScreen
     
     private func addRoomToSpace(roomProxy: JoinedRoomProxyProtocol, spaceID: String) async {
         roomProxy.subscribeToRoomInfoUpdates()
-        let runner = ExpiringTaskRunner {
+        let runner = ExpiringTaskRunner { @MainActor in
             // Necessary to build the room cache so that the space can be added as a parent.
-            _ = await roomProxy.infoPublisher.values.first { $0.powerLevels != nil }
+            var iterator = roomProxy.infoPublisher.values.makeAsyncIterator()
+            while let info = await iterator.next(isolation: #isolation), info.powerLevels == nil { }
```
stefanceriu: What in god's name is this? 😁

ElementX/Sources/Screens/InviteUsersScreen/InviteUsersScreenViewModel.swift
```diff
@@ -156,9 +156,11 @@ class InviteUsersScreenViewModel: InviteUsersScreenViewModelType, InviteUsersScr
...
                 }
                 
-                return await group.first { inviteResult in
-                    inviteResult.isFailure
-                } ?? .success(())
+                for await inviteResult in group where inviteResult.isFailure {
+                    return inviteResult
+                }
+                
+                return .success(())
```
stefanceriu, line 163: Random claude change or real thing?

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/TextRoomTimelineView.swift
```diff
@@ -55,15 +55,9 @@ struct TextRoomTimelineView: View, TextBasedRoomTimelineViewProtocol {
...
-                taskGroup.addTask {
-                    if case let .success(metadata) = await context?.viewState.linkMetadataProvider?.fetchMetadataFor(url: url) {
-                        await MainActor.run {
-                            linkMetadata[url] = metadata
-                        }
-                    }
-                }
+        for url in timelineItem.links.prefix(Self.maxLinkPreviewsToRender) {
+            if case let .success(metadata) = await context?.viewState.linkMetadataProvider?.fetchMetadataFor(url: url) {
+                linkMetadata[url] = metadata
```
stefanceriu: Shouldn't this be concurrent or something?

ElementX/Sources/Services/Authentication/AuthenticationService.swift
```diff
@@ -198,12 +198,24 @@ class AuthenticationService: AuthenticationServiceProtocol {
...
+        // Bridge from the SDK's synchronous callback into Swift Concurrency. Yielding is safe from
+        // any thread; the single long-lived `for await` consumer forwards the progress updates on
+        // the main actor in FIFO order.
+        let (progressStream, progressContinuation) = AsyncStream<QrLoginProgress>.makeStream()
+        
         let listener = SDKListener { progress in
-            guard let progress = QRLoginProgress(rustProgress: progress) else { return }
-            progressSubject.send(progress)
+            progressContinuation.yield(progress)
+        }
```
stefanceriu: Any way to do this generically for all these different listeners?

ElementX/Sources/Services/Location/RoomLiveLocationService.swift
```diff
@@ -20,15 +20,34 @@ final class RoomLiveLocationService: RoomLiveLocationServiceProtocol {
...
     
+    /// Bridge from the SDK's synchronous callback into Swift Concurrency. Yielding is safe from any
+    /// thread; a single long-lived `for await` consumer (set up in `init`) applies the updates on the
+    /// main actor in FIFO order, guaranteeing one in-flight update at a time.
+    private let updatesContinuation: AsyncStream<[LiveLocationShareUpdate]>.Continuation
+    
+    deinit {
+        updatesContinuation.finish()
+    }
+
```
stefanceriu, line 31: What happens if we don't finish it? Why doesn't the AuthenticationService do it?

ElementX/Sources/Services/Spaces/SpaceRoomListProxy.swift
```diff
@@ -29,29 +29,70 @@ class SpaceRoomListProxy: SpaceRoomListProxyProtocol {
...
-        let paginationStateSubject = CurrentValueSubject<SpaceRoomListPaginationState, Never>(spaceRoomList.paginationState())
-        paginationStatePublisher = paginationStateSubject.asCurrentValuePublisher()
+        let (paginationStates, paginationStateContinuation) = AsyncStream<SpaceRoomListPaginationState>.makeStream()
+        self.paginationStateContinuation = paginationStateContinuation
+        
+        let (roomUpdates, roomUpdatesContinuation) = AsyncStream<[SpaceListUpdate]>.makeStream()
+        self.roomUpdatesContinuation = roomUpdatesContinuation
+        
+        let (spaceUpdates, spaceUpdatesContinuation) = AsyncStream<SpaceRoom>.makeStream()
+        self.spaceUpdatesContinuation = spaceUpdatesContinuation
```
stefanceriu: Did we really had to make these changes now or they could've waited? I would've tackled this together with moving away from Combine.

ElementX/Sources/AccessibilityTests/AccessibilityTestsAppCoordinator.swift
```diff
@@ -142,14 +139,17 @@ struct PreviewsWrapperView: View {
         
         switch fulfillmentSource {
         case .publisher(let publisher):
-            _ = await publisher
-                // Not sure whye byt some publisher seem to not properly comunicate their completion,
-                // so we added a timeout. Since we are going to migrate from publishers to stream,
-                // this is a temporary solution
+            // Not sure whye byt some publisher seem to not properly comunicate their completion,
```
stefanceriu: Typo(s)

ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomCell.swift
```diff
@@ -211,10 +211,20 @@ import MatrixRustSDKMocks
...
+    
+    static func mockRooms(from provider: RoomSummaryProviderProtocol) -> [HomeScreenRoom] {
+        var rooms = [HomeScreenRoom]()
+        for summary in provider.roomListPublisher.value {
+            if let room = mockRoom(summary: summary) {
+                rooms.append(room)
+            }
+        }
+        return rooms
+    }
```
stefanceriu: What's happening here?

ElementX/Sources/Services/NotificationSettings/NotificationSettingsProxy.swift
```diff
@@ -119,12 +120,18 @@ final class NotificationSettingsProxy: NotificationSettingsProxyProtocol {
...
             .timeout(.seconds(2.0), scheduler: DispatchQueue.main, options: nil, customError: nil)
-            .values.first { $0 == .settingsDidChange }
+            .values
+            .makeAsyncIterator()
+        
+        while let callback = await iterator.next(isolation: #isolation) {
+            if callback == .settingsDidChange {
+                break
+            }
+        }
```
stefanceriu, line 132: Why the change?

ElementX/Sources/UITests/UITestsAppCoordinator.swift
```diff
@@ -620,11 +619,16 @@ class MockScreen: Identifiable {
...
+            for spaceRoom in [SpaceServiceRoom].mockSpaceList {
+                roomPreviews.append(RoomPreviewProxyMock(spaceServiceRoom: spaceRoom))
+            }
+            
             let clientProxy = ClientProxyMock(.init(userID: "@mock:client.com",
                                                     deviceID: "MOCKCLIENT",
                                                     roomSummaryProvider: RoomSummaryProviderMock(.init(state: .loaded(roomSummaries))),
                                                     spaceServiceConfiguration: .init(topLevelSpaces: .mockSpaceList.filter(\.isSpace) + .mockSingleRoom),
-                                                    roomPreviews: [SpaceServiceRoom].mockSpaceList.map(RoomPreviewProxyMock.init),
+                                                    roomPreviews: roomPreviews,
```
stefanceriu: Why?

ElementX/Sources/Other/Extensions/Observable.swift
```diff
@@ -14,26 +14,34 @@ extension Observable {
     /// https://github.com/swiftlang/swift-evolution/blob/main/proposals/0475-observed.md
     ///
     /// - Parameter property: The key path to the property you would like to observe.
-    func observe<Value>(_ property: KeyPath<Self, Value>) -> AsyncStream<Value> {
-        AsyncStream { continuation in
-            let isActive = Mutex(true)
-            
-            @Sendable func observe() {
+    @MainActor func observe<Value: Sendable>(_ property: KeyPath<Self, Value>) -> AsyncStream<Value> {
```
stefanceriu, line 17: This was flakey to being with, right? Probably deserved a commit of its own, not entirely sure what's happening here.

ElementX/Sources/Other/MapLibre/MapLibreMapView.swift
```diff
@@ -233,7 +233,10 @@ extension MapLibreMapView {
                 
                 // Mirror every animated coordinate update to keep the camera as smooth as the marker.
                 followedAnnotationObservation = annotation.observe(\.coordinate) { [weak mapView] annotation, _ in
-                    mapView?.setCenter(annotation.coordinate, animated: false)
+                    // The coordinate is only ever updated by SwiftUI so KVO fires on the main actor.
```
stefanceriu, line 236: This feels unsafe given future changes.

ElementX/Sources/Other/MapLibre/MapLibreMapView.swift
```diff
@@ -303,7 +306,7 @@ extension MapLibreMapView {
             // is an option set and gestures can come combined with other reasons, so check for
             // containment instead of an exact match.
             let centerChangingGestures: MLNCameraChangeReason = [.gesturePan, .gesturePinch, .gestureRotate]
-            if !reason.intersection(centerChangingGestures).isEmpty {
+            if !reason.isDisjoint(with: centerChangingGestures) {
```
stefanceriu, line 309: Random warning fix, right?

ElementX/Sources/Other/Pills/PillAttachmentViewProvider.swift
```diff
@@ -23,44 +23,53 @@ final class PillAttachmentViewProvider: NSTextAttachmentViewProvider, NSSecureCo
...
-    override init(textAttachment: NSTextAttachment, parentView: UIView?, textLayoutManager: NSTextLayoutManager?, location: NSTextLocation) {
+    override nonisolated init(textAttachment: NSTextAttachment, parentView: UIView?, textLayoutManager: NSTextLayoutManager?, location: NSTextLocation) {
         super.init(textAttachment: textAttachment, parentView: parentView, textLayoutManager: textLayoutManager, location: location)
         
-        // Keep a reference to the parent text view for size adjustments and pills flushing.
-        delegate = parentView?.superview as? PillAttachmentViewProviderDelegate
+        // TextKit isn't annotated but creates the view providers on the main thread,
+        // the assumeIsolated below would trap otherwise.
+        nonisolated(unsafe) let provider = self
+        MainActor.assumeIsolated {
```
stefanceriu, line 32: We should probably use this `assumeIsolated` in more place to catch migration problems early.

ElementX/Sources/Other/Pills/PillAttachmentViewProvider.swift
```diff
@@ -23,44 +23,53 @@ final class PillAttachmentViewProvider: NSTextAttachmentViewProvider, NSSecureCo
...
+            // Keep a reference to the parent text view for size adjustments and pills flushing.
+            provider.delegate = parentView?.superview as? PillAttachmentViewProviderDelegate
+        }
         tracksTextAttachmentViewBounds = true
     }
     
-    @MainActor
-    override func loadView() {
+    override nonisolated func loadView() {
```
stefanceriu, line 39: Why change this to nonisolated? 🤔

ElementX/Sources/Other/UserAgentBuilder.swift
```diff
@@ -61,4 +61,18 @@ enum UserAgentBuilder {
...
+    /// built off the main thread (e.g. when the NSE builds its client), so hop over when needed.
+    private static func mainThreadDeviceInfo() -> (systemVersion: String, scale: CGFloat) {
+        if Thread.isMainThread {
+            MainActor.assumeIsolated { (UIDevice.current.systemVersion, UIScreen.main.scale) }
+        } else {
+            DispatchQueue.main.sync {
+                MainActor.assumeIsolated { (UIDevice.current.systemVersion, UIScreen.main.scale) }
+            }
+        }
+    }
```
stefanceriu: wut? 😵‍💫

UnitTests/Sources/UserPreferenceTests.swift
```diff
@@ -135,28 +135,28 @@ struct UserPreferenceTests {
     @Test
     func localOverRemoteValue() {
         let storage = VolatileUserDefaults()
-        @UserPreference(key: TestsKey.testKey, defaultValue: "", storage: storage) var preference
-        #expect(preference == "")
+        let preference = UserPreference(key: TestsKey.testKey, defaultValue: "", storage: storage)
+        #expect(preference.wrappedValue == "")
```
stefanceriu, line 139: What's happening here? 🤔

Tools/Sourcery/AutoMockable.stencil
```diff
@@ -21,7 +21,7 @@ import {{ import }}
 {% macro accessLevel level %}{% if level != 'internal' %}{{ level }} {% endif %}{% endmacro %}
 {% macro staticSpecifier method %}{% if method.isStatic and not method.isInitializer %}static {% endif %}{% endmacro %}
 {% macro methodThrowableErrorDeclaration method %}
-    {% call accessLevel method.accessLevel %}{% call staticSpecifier method %}var {% call swiftifyMethodName method.selectorName %}ThrowableError: Error?
+    {% call accessLevel method.accessLevel %}{% call staticSpecifier method %}nonisolated(unsafe) var {% call swiftifyMethodName method.selectorName %}ThrowableError: Error?
```
stefanceriu, line 24: Aren't these actually safe because of the inner locks?

ElementX/Sources/Application/Settings/AppSettings.swift
stefanceriu, line 1: Surely there must be a better way than making everything here unsafe, right?

ElementX/Sources/Screens/CreateRoomScreen/CreateRoomScreenViewModel.swift
Velin92: This is actually cool, essentially it's a way to convert the values published from a combine publisher into an async sequence, the problem with first is that was not concurrency aware, so we would get an issue because RoomInfoProxy is not sendable, and we are not sure if by awaiting with first we would be exiting the current actor or not. This solution instead allows us to specify the isolation in which we are performing the await, the next function is like first, but only if the condition is met. Instead the pattern `isolation: #isolation` means, run this on the current isolation whichever it is, which means that nothing is crossing actors, and RoomInfoProxy can stay non sendable.

However by pointing out this code you made me realise an important bit... RoomInfoProxy NEVER MUTATES, so it can be nonisolated Sendable regardless being a struct or a class! Which would mean a huge performance gain, alongside an improved ease to use, since we can pass it through any actor. AI ofc didn't notice it! But now that we did we can mark them as nonisolated sendable which will make them extremely easy to reuse. We can also revert this code.

ElementX/Sources/Screens/InviteUsersScreen/InviteUsersScreenViewModel.swift
Velin92: this is because the first isFailure check was going to require Sendability, instead in this way this still a task that is running on the main actor. Is not a big deal because then internally it will await tokio on the SDK and offload this main actor task. Essentially we are dispatching these task on the main actor one by one, but the main actor is still going to await them, and do the next while awaiting.

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/TextRoomTimelineView.swift
Velin92: You are right, this should be done concurrently.

ElementX/Sources/Services/Authentication/AuthenticationService.swift
Velin92: Maybe? But let's save this for another PR

ElementX/Sources/Services/Location/RoomLiveLocationService.swift
Velin92: since the Task iterates over the stream is not stored anywhere and cancelled, the task might leak and keep a thread busy awaiting even if its not going to do anything, so we need to stop it somehow, and the only way is to call finish on the continuation. The other way around is to store the task so that when deallocation happens, it gets cancelled and the for await in it gets cancelled too.

ElementX/Sources/Services/Spaces/SpaceRoomListProxy.swift
Velin92: solved with the onMainActor helper function

ElementX/Sources/Screens/HomeScreen/View/HomeScreenRoomCell.swift
Velin92: stuff that we can solve with some more senables

ElementX/Sources/Services/NotificationSettings/NotificationSettingsProxy.swift
Velin92: as above

ElementX/Sources/UITests/UITestsAppCoordinator.swift
Velin92: also here, solved

ElementX/Sources/Other/Extensions/Observable.swift
Velin92: will take care in a next PR

ElementX/Sources/Other/MapLibre/MapLibreMapView.swift
Velin92: yup, asked the AI to take care of the warnings that concurrency created it also took care of other minimal warnings

ElementX/Sources/Other/MapLibre/MapLibreMapView.swift
Velin92: let's also add a MainThread check I think it's probably difficult this is ever going to change.

ElementX/Sources/Other/Pills/PillAttachmentViewProvider.swift
Velin92: Before swift 6 you were allowed to change the actor of an override, in swift 6 not anymore since it wasn't really safe and guaranteed, so even if the whole provider is MainActor now, its overridden functions are not so this piece here is still nonisolated (needs to be annotated because new functions will be main actor instead). Still this is definitely running on the main actor.

PR #5745, opened 2026-06-16
Hide avatar in send invite confirmation view

### LINKED ISSUE #5239: A11y: [Start chat] Avatar letter announced without context in 'Send i...
- **Checkpoint ID:** 11.1.1.1b/5.1
- **View/Component:** Start chat

### Description
Avatar letter announced without context in 'Send invite?' overlay.

### Recommendation
Hide letter or provide meaningful alt text.

--- 
*Report Date: 20.02.2026*

Velin92:
fixes #5239

### DIFF
--- a/ElementX/Sources/Screens/StartChatScreen/View/SendInviteConfirmationView.swift
+++ b/ElementX/Sources/Screens/StartChatScreen/View/SendInviteConfirmationView.swift
@@ -63,6 +63,7 @@ struct SendInviteConfirmationView: View {
                                 contentID: userToInvite.id,
                                 avatarSize: .user(on: .sendInviteConfirmation),
                                 mediaProvider: mediaProvider)
+                .accessibilityHidden(true) // Decorative: the user's name and ID are read from the subtitle below.
             VStack(spacing: 8) {
                 Text(title)
                     .multilineTextAlignment(.center)


ElementX/Sources/Screens/StartChatScreen/View/SendInviteConfirmationView.swift
```diff
@@ -63,6 +63,7 @@ struct SendInviteConfirmationView: View {
                                 contentID: userToInvite.id,
                                 avatarSize: .user(on: .sendInviteConfirmation),
                                 mediaProvider: mediaProvider)
+                .accessibilityHidden(true) // Decorative: the user's name and ID are read from the subtitle below.
```
stefanceriu, line 66: Hmm, we do this quite a lot, should we disable it within the LoadableAvatarImage itself?

Velin92: in some cases in the existing code base we do hide it, while in others we apply labels to it, for example there are instances of applying onTap on one of the avatars, so we give an a11y label to communicate what the tap does, making it hidden would break it. So I guess we should keep this flexibility.

stefanceriu: Oh so an accessibilityLabel we would apply frmo the tapGesture level would break because of the internal `accessibilityHidden`? Wouldn't have expected that tbh 🤔

Velin92: If the view is not encapsulated yeah it will, if we instead wrap it in a button it would actually work, but if I remember we can't always wrap it in a button due to some constraints on the looks, like the nav bar

stefanceriu (approved):
👍

PR #5754, opened 2026-06-18
Search tab

stefanceriu:
Moves search out of the home screen and into its own dedicated tab.

- Adds a `Search` tab that filters rooms as you type, with an empty state and hardware-keyboard navigation (arrow keys to move through results, return to open).
- Wires the tab into the tab coordinator with support for search role tabs.
- Removes the old global search screen it replaces.
- Gated behind the **Global search** developer flag ("Moves search to a separate tab"), iOS 26+.

Screenshot 2026-06-19 at 13 59 01


ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -137,12 +162,17 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
...
              .roomDetails, .roomMemberDetails, .userProfile,
              .event, .eventOnRoomAlias, .childEvent, .childEventOnRoomAlias,
-             .share, .transferOwnership, .thread, .globalSearch:
+             .share, .transferOwnership, .thread:
             clearPresentedSheets(animated: animated) // Make sure the presented route is visible.
             chatsTabFlowCoordinator.handleAppRoute(appRoute, animated: animated)
             if navigationTabCoordinator.selectedTab != .chats {
                 navigationTabCoordinator.selectedTab = .chats
             }
+        case .globalSearch:
```
pixlwave: Personally I think we should rename this route to `search` to match `SearchScreen`. If we ever have a room search route we can call it `roomSearch`.

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -38,6 +38,12 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
     private let spacesTabFlowCoordinator: SpacesTabFlowCoordinator
     private let spacesTabDetails: NavigationTabCoordinator<HomeTab>.TabDetails
     
+    /// The search tab relies on the detached `.search` role tab, so it's only shown on iOS 26.
+    /// It's also gated behind the `globalSearchEnabled` flag.
```
pixlwave: I'm not sure we really need these that much but up to you 😄

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
```diff
@@ -49,6 +49,10 @@ class HomeScreenViewModel: HomeScreenViewModelType, HomeScreenViewModelProtocol
                                            bindings: .init(filtersState: .init(appSettings: appSettings))),
                    mediaProvider: userSession.mediaProvider)
         
+        if appSettings.globalSearchEnabled, !ProcessInfo.processInfo.isiOSAppOnMac, #available(iOS 26.0, *) {
```
pixlwave: I *think* it should be like this if I'm following correctly?

```suggestion
        if appSettings.globalSearchEnabled, #available(iOS 26.0, *) {
```

ElementX/Sources/Screens/SearchScreen/View/SearchScreen.swift
```diff
@@ -0,0 +1,226 @@
...
+                emptyState
+            } else {
+                roomList
+            }
+        }
+        .frame(maxHeight: .infinity, alignment: .top)
+        .searchable(text: $context.searchQuery, placement: .toolbarPrincipal)
+        .searchFocused($isSearchFieldFocused)
+        .autocorrectionDisabled(true)
+        .textInputAutocapitalization(.never)
```
pixlwave: Is this one necessary? We didn't have it before and generally I expect searches to start with a capital letter like they do in Messages/Mail/Notes etc.

ElementX/Sources/Screens/SearchScreen/View/SearchScreen.swift
```diff
@@ -0,0 +1,226 @@
...
+        .onSubmit(of: .search) {
+            if let selectedRoomID {
+                context.send(viewAction: .selectRoom(roomID: selectedRoomID))
+            }
+        }
+        // The search field is owned by the system, so we reach into it to forward the arrow keys
+        // to the results list (the same key handling the old global search did with its own field).
```
pixlwave: I don't think this comment adds any value here at all 😄

ElementX/Sources/Screens/SearchScreen/View/SearchScreen.swift
```diff
@@ -0,0 +1,226 @@
...
+            selectedRoomID = context.viewState.rooms.first?.id
+        }
+        // Reset to the first result when the results change (e.g. a new query), keyed on the top result
+        // so pagination and background updates that leave it untouched don't clobber the selection.
+        .onChange(of: context.viewState.rooms.first?.id) {
+            selectedRoomID = context.viewState.rooms.first?.id
+        }
+    }
+    
+    private func moveSelection(backwards: Bool) {
```
pixlwave: This should probably go nearer the bottom beneath the subviews.

ElementX/Sources/Screens/SearchScreen/View/SearchScreen.swift
```diff
@@ -0,0 +1,226 @@
...
+        
+        let nextIndex = backwards ? currentIndex - 1 : currentIndex + 1
+        guard rooms.indices.contains(nextIndex) else { return }
+        self.selectedRoomID = rooms[nextIndex].id
+    }
+    
+    @ViewBuilder
+    private var emptyState: some View {
+        if context.viewState.isSearching {
+            ContentUnavailableView.search
```
pixlwave: As much as I like Apple's view, presumably we have our own one?

ElementX/Sources/Screens/SearchScreen/View/SearchScreen.swift
```diff
@@ -0,0 +1,226 @@
...
+                    .listRowBackground(selectionBackground(for: room))
+                    .alignmentGuide(.listRowSeparatorLeading) { _ in 64 } // Inset the separator to the text.
+                    .onTapGesture { context.send(viewAction: .selectRoom(roomID: room.id)) }
+                    .onAppear {
+                        if room == context.viewState.rooms.first {
+                            context.send(viewAction: .reachedTop)
+                        } else if room == context.viewState.rooms.last {
+                            context.send(viewAction: .reachedBottom)
+                        }
+                    }
```
pixlwave: Can we follow our normal pattern and make a `SearchScreenRoomCell` here instead of using a func?

ElementX/Sources/Screens/SearchScreen/View/SearchScreen.swift
```diff
@@ -0,0 +1,226 @@
...
+        }
+    }
+    
+    private var roomList: some View {
+        List {
+            ForEach(context.viewState.rooms) { room in
+                roomCell(for: room)
+                    .listRowInsets(.init())
+                    .listRowBackground(selectionBackground(for: room))
+                    .alignmentGuide(.listRowSeparatorLeading) { _ in 64 } // Inset the separator to the text.
```
pixlwave: Is that 64 compatible with dynamic type? I think we should be following the same pattern we use elsewhere by hiding the separator (or using a `LazyVStack`) and manually applying it ourselves to the trailing portion of the view. I think we've got a modifier for it now too.

ElementX/Sources/Screens/SearchScreen/View/SearchScreen.swift
```diff
@@ -0,0 +1,226 @@
...
+                if !room.description.isEmpty {
+                    Text(room.description)
+                        .font(.compound.bodyMD)
+                        .foregroundStyle(.compound.textSecondary)
+                        .lineLimit(1)
+                }
+            }
+            
+            Spacer(minLength: 0)
+        }
```
pixlwave:
```suggestion
        }
        .frame(maxWidth: .infinity, alignment: .leading)
```

ElementX/Sources/Screens/SearchScreen/View/SearchScreen.swift
```diff
@@ -0,0 +1,226 @@
...
+    @Environment(\.dynamicTypeSize) private var dynamicTypeSize
+    @Environment(\.horizontalSizeClass) private var horizontalSizeClass
+    
+    @Bindable var context: SearchScreenViewModel.Context
+    
+    @FocusState private var isSearchFieldFocused: Bool
+    /// The room highlighted for hardware keyboard selection (arrow keys + return).
+    @State private var selectedRoomID: String?
+    
+    var body: some View {
```
pixlwave, line 22: We probably want a `bgCanvasDefault` background colour on here (and to hide the scrollContentBackground) for the expected result in dark mode.

ElementX/Sources/Screens/SearchScreen/View/SearchScreen.swift
```diff
@@ -0,0 +1,226 @@
...
+    ///
+    /// The `.search` role tab owns its text field, so there's no SwiftUI hook for its key presses. We reach the
+    /// field through the navigation stack (as ``SearchFieldStyle`` does) and swap its class for one that overrides
+    /// `pressesBegan` — the same key interception the old global search performed on its own field.
+    func searchResultsKeyboardNavigation(moveUp: @escaping () -> Void, moveDown: @escaping () -> Void) -> some View {
+        introspect(.navigationStack, on: .supportedVersions, scope: .ancestor) { navigationController in
+            guard let textField = navigationController.navigationBar.topItem?.searchController?.searchBar.searchTextField else { return }
+            
+            if !(textField is KeyNavigatingSearchTextField) {
+                object_setClass(textField, KeyNavigatingSearchTextField.self)
```
pixlwave, line 204: loool, yeah that works 😂

ElementX/Sources/Screens/SearchScreen/View/SearchScreen.swift
```diff
@@ -0,0 +1,226 @@
...
+                    return false
+                }
+                return true
+            }
+        }
+    }
+}
+
+/// A `UISearchTextField` whose class is installed at runtime via `object_setClass`, so it must not declare any
+/// stored properties — the key handler is kept as an associated object instead.
```
pixlwave: I'm not having anything to do with any of this 🙈

ElementX/Sources/Screens/SearchScreen/SearchScreenViewModel.swift
```diff
@@ -33,36 +34,41 @@ class GlobalSearchScreenViewModel: GlobalSearchScreenViewModelType, GlobalSearch
...
-        context.$viewState
-            .map(\.bindings.searchQuery)
-            .removeDuplicates()
-            .sink { [weak self] searchQuery in
+        searchQueryObservationTask = Task { [weak self] in
+            guard let stream = self?.context.observe(\.viewState.bindings.searchQuery) else { return }
+            var lastQuery: String?
+            for await searchQuery in stream {
+                guard searchQuery != lastQuery else { continue }
+                lastQuery = searchQuery
```
pixlwave: If you `import AsyncAlgorithms` then you can do this:


```suggestion
            guard let stream = self?.context.observe(\.viewState.bindings.searchQuery).removeDuplicates() else { return }
            for await searchQuery in stream {
```

ElementX/Sources/Screens/SearchScreen/SearchScreenViewModel.swift
```diff
@@ -33,36 +34,41 @@ class GlobalSearchScreenViewModel: GlobalSearchScreenViewModelType, GlobalSearch
...
     }
     
+    deinit {
+        searchQueryObservationTask?.cancel()
+    }
+    
     func stop() {
-        // This is a shared provider so we should reset the filtering when we are done with the view
+        searchQueryObservationTask?.cancel()
+        // This is a shared provider so we should reset the filtering when we are done with the view.
```
pixlwave: Errr, does it matter that we're never "done" with the view now?

ElementX/Sources/Screens/SearchScreen/SearchScreenViewModel.swift
pixlwave, line 1: It's only at this point I realise `SearchScreen` probably isn't entirely new and is just the `GlobalSearchScreen` renamed 🤦‍♂️

pixlwave: Probably should make the header say 2026 though if we're removing the old copyright.

pixlwave (approved):
Lots of little thoughts, but nothing major.

ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift
stefanceriu: Oh yeah, definitely, sorry about that. Got left over from when they were 2 different screens, before I bought keyboard navigation over to the new one.

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
stefanceriu: Just Claude being extra 😁

ElementX/Sources/Screens/SearchScreen/SearchScreenViewModel.swift
stefanceriu: Yeah, let's just treat it like a brand new screen, the old one was never properly reviewed

ElementX/Sources/Screens/SearchScreen/View/SearchScreen.swift
stefanceriu: Yeah, no idea where it got that from. It should be left to whatever you keyboard autocapitalization settings are.

ElementX/Sources/Screens/SearchScreen/View/SearchScreen.swift
stefanceriu: Yeah, definitely 👍

ElementX/Sources/Screens/SearchScreen/SearchScreenViewModel.swift
stefanceriu: Now that's pretty damn cool!

ElementX/Sources/Screens/SearchScreen/SearchScreenViewModel.swift
stefanceriu: I'm honestly just testing how far I can push this

> Comment #15 (never "done"): removed the dead `stop()` (never called for a persistent tab) from VM/protocol/coordinator. But deleting it alone regresses — the shared `alternateRoomSummaryProvider` gets its filter overwritten by other consumers (message forwarding, room selection…), so returning to the tab with an unchanged query would show stale results. Added an `.appeared` action that re-applies the filter on each appearance.

ElementX/Sources/Screens/SearchScreen/View/SearchScreen.swift
stefanceriu: Apparently

> No-results state now uses `Text(L10n.commonNoResults)` (our pattern from StartChat/RoomDirectory) instead of `ContentUnavailableView.search`

and it's not wrong it seems. Were you thinking of a screen/component in particular?

pixlwave: Well if there aren't any designs then presumably the same thing we use when you select enough filters to have no results:

Screenshot 2026-06-19 at 9 43 48 am

I have no idea if that's a re-usable component though.

ElementX/Sources/Screens/SearchScreen/View/SearchScreen.swift
```diff
@@ -107,45 +107,69 @@ struct SearchScreen: View {
...
+                            .lineLimit(1)
+                    }
                 }
+                .frame(maxWidth: .infinity, alignment: .leading)
+            }
+            .padding(.horizontal, 16)
+            .padding(.vertical, 8)
+            .overlay(alignment: .bottom) {
+                separator
             }
```
pixlwave: We can use `.rowDivider` here instead of this custom one :)

ElementX/Sources/Screens/SearchScreen/View/SearchScreen.swift
```diff
@@ -154,6 +178,29 @@ struct SearchScreen: View {
...
     }
+    
+    private var separator: some View {
+        Rectangle()
+            .fill(Color.compound.borderDisabled)
+            .frame(height: 1 / UIScreen.main.scale)
+    }
+}
+
+private struct SearchScreenRoomCellButtonStyle: ButtonStyle {
```
pixlwave, line 180: TBH, I think we should probably just use `HomeScreenRoomCellButtonStyle` directly looking at the implementations. I don't think this one calls for insets and a rounded style given it's inside of a plain List.

ElementX/Sources/Screens/SearchScreen/View/SearchScreen.swift
```diff
@@ -107,45 +107,69 @@ struct SearchScreen: View {
             }
         }
         .listStyle(.plain)
+        .scrollContentBackground(.hidden)
```
pixlwave: Ah you know what, I should've just pointed you to the `.compoundList(.plain)` parameter that I added the other day 🤦‍♂️

pixlwave (approved):
Very nice, sorry 3 small comments that just jumped out to me looking at these again.

PR #5765, opened 2026-06-22
Fix libdispatch main-thread trap when logging Sentry crash events

stefanceriu:
Sentry's beforeSend runs on a background queue, but Event.crashLog read UIDevice.current there, which carries a main-thread dispatch assertion and trapped libdispatch. Read the device values on the main thread at setup and pass them into the now-nonisolated crashLog.


pixlwave (approved):
👌

PR #5769, opened 2026-06-22
Various crash fixes

stefanceriu:
Address some of the Sentry crashes. Treat this as an experiment in Claude <-> SentryMCP interactions.

| Sentry | Commit | Problem | Fix |
|---|---|---|---|
| 60N | da934a545 | `NotificationToneManager.init` fatalError'd when `createDirectory` failed (sandbox permission, NSCocoaError 513) — e.g. background launch before the container is writable | Log the error instead of crashing; root cause acknowledged for separate treatment |
| 61D | 1568e28d9 | `LoadableImage` GIF provider used `MainActor.assumeIsolated` (libdispatch crash when Kingfisher called off-main) + fatalError when content wasn't GIF | Hop onto the main actor and return `.failure`; handler is `@Sendable` per Kingfisher's protocol |
| PG | c8489e2bd | Photo/document pickers emitted `.selectedMediaAtURLs([])` when every selected item failed to load/export — an empty "success" consumers fatalError'd on | Both `PhotoLibraryPicker` & `DocumentPicker` skip the empty emission (failures already go through `.error`); also report `.error` for the silent missing-content-type case |
| VV | d4edf0150 | `AudioRecorder.startRecording` crashed AVAudioEngine (IsFormatSampleRateAndChannelCountValid) when the input format was invalid (0 sample rate, no input route) | Guard `sampleRate > 0 && channelCount > 0` and bail with `.unsupportedAudioFormat` before connecting nodes / installing the tap |
| DS | 30e5d2f44 | `TimelineTableViewController.scrollToOldestItem` targeted `(timelineItemsIDs.count - 1, section: 1)` before the data source applied that section → out-of-bounds NSRangeException | Scroll against the table's actual `numberOfSections`/`numberOfRows(inSection:)` |

### DIFF
--- a/ElementX/Sources/Other/SwiftUI/Views/LoadableImage.swift
+++ b/ElementX/Sources/Other/SwiftUI/Views/LoadableImage.swift
@@ -234,21 +234,22 @@ private struct LoadableImageContent<TransformerView: View, PlaceholderView: View
         mediaSource.url.absoluteString
     }
     
-    nonisolated func data(handler: @escaping (Result<Data, Error>) -> Void) {
-        // Kingfisher isn't annotated but invokes the provider on the main thread.
-        let gifData = MainActor.assumeIsolated { () -> Data? in
+    nonisolated func data(handler: @escaping @Sendable (Result<Data, Error>) -> Void) {
+        // Kingfisher isn't annotated and doesn't guarantee the provider is invoked on the main
+        // thread so hop onto the main actor.
+        Task { @MainActor in
             guard case let .gifData(data) = contentLoader.content else {
-                return nil
+                handler(.failure(LoadableImageError.missingGIFData))
+                return
             }
-            return data
+            
+            handler(.success(data))
         }
-        
-        guard let gifData else {
-            fatalError("Shouldn't reach this point without any gif data")
-        }
-        
-        handler(.success(gifData))
     }
+}
+
+private enum LoadableImageError: Error {
+    case missingGIFData
 }
 
 private class ContentLoader: ObservableObject {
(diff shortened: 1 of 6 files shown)


Velin92 (approved):

ElementX/Sources/Other/SwiftUI/Views/LoadableImage.swift
```diff
@@ -234,23 +234,24 @@ private struct LoadableImageContent<TransformerView: View, PlaceholderView: View
         mediaSource.url.absoluteString
     }
     
-    nonisolated func data(handler: @escaping (Result<Data, Error>) -> Void) {
-        // Kingfisher isn't annotated but invokes the provider on the main thread.
-        let gifData = MainActor.assumeIsolated { () -> Data? in
+    nonisolated func data(handler: @escaping @Sendable (Result<Data, Error>) -> Void) {
+        // Kingfisher isn't annotated and doesn't guarantee the provider is invoked on the main
+        // thread, so hop onto the main actor rather than asserting isolation (which would crash).
```
pixlwave: I don't think we need the "rather than asserting isolation (which would crash)." part, just explaining the why and the how is enough, we don't need to note the previous setup.

ElementX/Sources/Services/Audio/Recorder/AudioRecorder.swift
```diff
@@ -166,7 +166,15 @@ nonisolated class AudioRecorder: AudioRecorderProtocol, @unchecked Sendable {
             
             let inputNode = audioEngine.inputNode
             let inputFormat = inputNode.outputFormat(forBus: 0)
-            let hardwareSampleRate = audioEngine.inputNode.outputFormat(forBus: 0).sampleRate
+            let hardwareSampleRate = inputFormat.sampleRate
```
pixlwave, line 169: Ok interesting, I wondered what this one was about given it isn't related to below and has changed from output to input, but I see the sample rate converter below so it was just plain wrong 😅

pixlwave (approved):
LGTM, that last one makes sense although also seems annoying that they can get out of sync 😅

PR #5771, opened 2026-06-22
Delete the composerAttachments UI test snapshot step

stefanceriu:
Because it keeps failing and because we have proper tests for all of those options


UITests/Sources/UserSessionScreenTests.swift
```diff
@@ -74,7 +74,6 @@ class UserSessionScreenTests: XCTestCase {
         XCTAssert(latestEditedMessage.waitForExistence(timeout: 5.0))
         
         app.buttons[A11yIdentifiers.roomScreen.composerToolbar.openComposeOptions].tap(.center)
```
pixlwave, line 76: Is there any point in continuing to tap the button if there are already other tests for this?

pixlwave (approved):
😎

UITests/Sources/UserSessionScreenTests.swift
stefanceriu: I figured it's good to at least try to find the identifier on screen ¯\_(ツ)_/¯

PR #5773, opened 2026-06-22
Ensure that the current user is the device owner before linking a new device.

pixlwave:
This PR uses `LAContext.evaluatePolicy(.deviceOwnerAuthentication)` to authenticate the current user before continuing the Link New Device flow.

Additionally it adds missing unit tests for `LinkNewDeviceScreenViewModel`.

A few remaining tasks:

- [x] Confirm authentication reason string and move to Localazy.
- [x] Confirm Info.plist string update.

| Link Mobile | Link Desktop | Fallback to PIN code |
| - | - | - |
| LinkMobile | LinkDesktop | LinkPINFallback |

### DIFF
--- a/ElementX/Sources/Screens/LinkNewDeviceScreen/LinkNewDeviceScreenModels.swift
+++ b/ElementX/Sources/Screens/LinkNewDeviceScreen/LinkNewDeviceScreenModels.swift
@@ -17,7 +17,7 @@ struct LinkNewDeviceScreenViewState: BindableState {
     nonisolated enum Mode: Equatable {
         enum ReadyState: Equatable {
             case idle
-            case authenticatingDeviceOwner
+            case verifyingDeviceOwner
             case generatingCode
         }
         
--- a/ElementX/Sources/Screens/LinkNewDeviceScreen/LinkNewDeviceScreenModels.swift
+++ b/ElementX/Sources/Screens/LinkNewDeviceScreen/LinkNewDeviceScreenModels.swift
@@ -15,8 +15,14 @@ enum LinkNewDeviceScreenViewModelAction {
 
 struct LinkNewDeviceScreenViewState: BindableState {
     nonisolated enum Mode: Equatable {
+        enum ReadyState: Equatable {
+            case idle
+            case authenticatingDeviceOwner
+            case generatingCode
+        }
+        
         case loading
-        case readyToLink(isGeneratingCode: Bool)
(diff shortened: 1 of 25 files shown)


ElementX/Sources/Screens/LinkNewDeviceScreen/LinkNewDeviceScreenModels.swift
```diff
@@ -15,8 +15,14 @@ enum LinkNewDeviceScreenViewModelAction {
...
     nonisolated enum Mode: Equatable {
+        enum ReadyState: Equatable {
+            case idle
+            case authenticatingDeviceOwner
+            case generatingCode
+        }
+        
         case loading
-        case readyToLink(isGeneratingCode: Bool)
+        case readyToLink(ReadyState)
```
stefanceriu, line 25: Way nicer 👌

ElementX/Sources/Screens/LinkNewDeviceScreen/LinkNewDeviceScreenViewModel.swift
```diff
@@ -89,12 +117,43 @@ class LinkNewDeviceScreenViewModel: LinkNewDeviceScreenViewModelType, LinkNewDev
...
             MXLog.error("Unexpected error action: \(action)")
             actionsSubject.send(.dismiss)
         case .cancel:
             actionsSubject.send(.dismiss)
         }
     }
+    
+    // MARK: - Authentication
+    
+    private func authenticateDeviceOwner() async throws -> Bool {
```
stefanceriu: Should this be a `isAuthenticatedDeviceOwner` instead maybe?

stefanceriu (approved):
Looking good! 👍

ElementX/Sources/Screens/LinkNewDeviceScreen/LinkNewDeviceScreenViewModel.swift
pixlwave: It's all moved to the AppLockService now, new name, new return type 😄

pixlwave:
I've updated this in preparation for falling back to an AppLockPIN screen (if possible) when the user doesn't have a passcode set on their device. That part ended up being enough work that I've split it into #5794 keep the reviews manageable.

stefanceriu (approved):
Noice, lgtm .. still 👍

Issue #5735, opened 2026-06-15
Crash (fatalError) opening Settings → Notifications on macOS (Designed for iPad) — NotificationToneManager init fails in 26.06.0


mdal-git:
## Steps to reproduce
1. Install Element X 26.06.0 (231) from the Mac App Store on an Apple Silicon Mac ("Designed for iPad" build).
2. Sign in to any homeserver.
3. Open Settings → Notifications.

**Result:** app crashes immediately, 100% reproducible.
**Expected:** notification settings screen opens (as on iOS).

## Root cause analysis (from local crash log + binary + source)
- Crash log: `EXC_BREAKPOINT (SIGTRAP)`, main thread, frame 0 `libswiftCore _assertionFailure`, triggered from SwiftUI ButtonAction.
- Disassembly of the shipped binary resolves the assertion arguments to **file `ElementX/NotificationToneManager.swift`, line 57**, message **"Catastrophic error setting up tone manager: "** — the `fatalError` in `NotificationToneManager.init` introduced with **Custom Notification Tones (PR #5527)**, first shipped in 26.06.0.
- `NotificationSettingsScreenCoordinator.swift:44` constructs `NotificationToneManager` when the screen is built — matching the trigger point.
- **Why it only crashes on macOS:** in an iOS-on-Mac app container, `Data/Library/Sounds` is a **symlink pointing outside the sandbox** (to `~/Library/Sounds`). `FileManager.createDirectory(.../Library/Sounds/AvailableSounds)` therefore throws, and `init` converts that into `fatalError`. On real iOS devices the directory is created normally, so the path is never exercised.
- Additional note: `/System/Library/Audio/UISounds` does not exist on macOS, so the system-tone list degrades to empty (handled safely via `compactMap`) — not the crasher, but worth handling for the Mac case too.

## Suggested fix
Treat tone-directory setup failure as non-fatal (disable custom tones / fall back to default) instead of `fatalError`, or special-case the iOS-on-Mac container layout.

## Environment
- Element X 26.06.0 (231), Mac App Store, Apple Silicon (macOS 26.x, Darwin 25.5.0)
- Homeserver: self-hosted (Tuwunel 1.7.1) — irrelevant to the crash; happens before any network interaction
- Crash log available on request (can attach `.ips`)

stefanceriu:
[...]

PR #5776, opened 2026-06-22
Remove custom tones menu on macOS

mredig:
Addresses #5735 by removing the custom tone selection item from the notification screen. 

Rationale: macOS lacks the default system sounds from iOS, macOS's sandbox contains an immutable `Sounds` directory, and most importantly, macOS appears to have flaky support for custom sounds in the first place.

This should probably be readdressed at some point in the future. In the meantime, it should at least work with the default tri-tone macOS notification sound, and, when its mood is right, perhaps the bundled element sound.

### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Screens/Settings/NotificationSettingsScreen/View/NotificationSettingsScreen.swift
+++ b/ElementX/Sources/Screens/Settings/NotificationSettingsScreen/View/NotificationSettingsScreen.swift
@@ -39,7 +39,7 @@ struct NotificationSettingsScreen: View {
                         additionalSettingsSection
                     }
                     
-                    if context.viewState.settings?.customToneSelectionEnabled == true {
+                    if context.viewState.customToneSelectionEnabled {
                         soundSelectionSection
                     }
                 }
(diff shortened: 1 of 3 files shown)


ElementX/Sources/Screens/Settings/NotificationSettingsScreen/View/NotificationSettingsScreen.swift
```diff
@@ -39,7 +39,9 @@ struct NotificationSettingsScreen: View {
                         additionalSettingsSection
                     }
                     
-                    soundSelectionSection
+                    if context.viewState.settings?.customToneSelectionEnabled == true {
```
pixlwave: That `NotificationSettingsScreenSettings` is intended for the push rules settings read from the SDK as I understand it. Instead this should go directly on the viewState similar to `isPresentedModally` etc.


```suggestion
                    if context.viewState.customToneSelectionEnabled == true {
```

pixlwave (commented):
Looks good to me, just one comment on where we store the boolean.

ElementX/Sources/Screens/Settings/NotificationSettingsScreen/View/NotificationSettingsScreen.swift
mredig: I see that all the properties on that object are from the SDK now. If I can make an observation, cross referencing is the only hint that that struct is for SDK settings, otherwise it's tribal knowledge. A more appropriate name for the structure or some documentation would easily alleviate that.

I'll make the change, just wanted to provide some feedback.

pixlwave (approved):
Thank you!

PR #5778, opened 2026-06-23
fix(ci): stabilise unit tests enterprise workflow

sarah-lacerda:
## Summary

- Add `|| github.token` fallback so `nschloe/action-cached-lfs-checkout` doesn't fail in forks where `ELEMENT_BOT_TOKEN` is not set
- Guard the "Zip results" step so it doesn't error when `fastlane/test_output` doesn't exist (e.g. when a prior step failed before tests ran)

## Test plan

- [ ] Trigger the workflow on a PR from a fork — checkout step should succeed with the `github.token` fallback
- [ ] Force a failure before tests run — zip step should exit cleanly (`|| true`) with no artifact rather than failing the step itself


sarah-lacerda:
Opened against wrong repo by mistake — please close.

PR #5779, opened 2026-06-23
Generate AppSettings preference accessors with Sourcery instead of @UserPreference

stefanceriu:
Problem
After the Swift 6 / strict concurrency / default-MainActor-isolation migration, AppSettings emits 41 "'nonisolated(unsafe)' has no effect on property" warnings. AppSettings must be `nonisolated` (it's read/written off the main actor by the NSE, background work, etc.), which forces `nonisolated` onto its stored properties. A `@UserPreference`-wrapped property is stored as a `var`, so it requires `nonisolated(unsafe)`, and the compiler then warns that the modifier has no effect on the wrapper's computed projection. The warning has no diagnostic group, so it can't be suppressed in place.

What we tried
- Dropping the annotation / using plain `nonisolated`: build error, the backing is a mutable stored property.
- Making AppSettings MainActor-isolated: warnings persist (they're about the projection, not isolation) and it breaks all off-main access.
- Per-file `-default-isolation nonisolated` / `-suppress-warnings` via XcodeGen compilerFlags: Xcode strips these (they're managed by build settings), so they never reach the compiler.
- A `@dynamicMemberLookup` facade over a Preferences holder: can't satisfy protocol requirements, and AppSettings conforms to four settings protocols (Common/Developer/Advanced/Labs) whose union needs 24 of the 41 preferences as real members - so it removed almost no boilerplate.

Solution
Drop the property wrapper. Each preference is now a plain `let _foo: UserPreference<T>`, and the `foo` value accessor (plus a `fooPublisher` for the ones annotated `// sourcery: publisher`) is code-generated by Sourcery into AppSettings+Preferences.swift. The generated accessors are plain computed properties, so there's no `nonisolated(unsafe)` and no warning, and they're real members so they still satisfy the four protocols.

Cost
`appSettings.$foo` is property-wrapper-only syntax that can't be generated, so the 23 publisher call sites now use `appSettings.fooPublisher`. All value usages are unchanged.

### Pull Request Checklist

- [ ] Voiceover enabled.


stefanceriu:
Decided to go with https://github.com/element-hq/element-x-ios/pull/5798 instead

PR #5781, opened 2026-06-23
Unit Tests migration to swift 6.2

Velin92:
Includes main actor isolation


UnitTests/Sources/ClassicAppMediaLoaderTests.swift
```diff
@@ -141,7 +141,7 @@ private class MockURLProtocol: URLProtocol {
     static let thumbnailData = Data("thumbnail data".utf8)
     
     /// The last request handled, for URL/header inspection in tests.
-    static var lastRequest: URLRequest?
+    nonisolated(unsafe) static var lastRequest: URLRequest?
```
pixlwave: I might be missing something, but can't we make this a Mutex instead of unsafe?

UnitTests/Sources/NSItemProviderTests.swift
pixlwave, line 1: I'm not following why these 2 have changed?

UnitTests/Sources/TestUtilities/DeferredFulfillment.swift
```diff
@@ -50,40 +65,32 @@ func deferFulfillment<P: Publisher<P.Output, Never>>(_ publisher: P,
...
-            group.addTask {
-                try await Task.sleep(for: timeout)
-                throw DeferredFulfillmentError.noOutput(message: message, sourceLocation: sourceLocation)
+            defer { group.cancelAll() }
+            
+            // This body inherits the caller's actor isolation (via `#isolation`), so `condition`
+            // is evaluated there rather than on the publisher's (possibly background) emission queue.
+            for await value in stream where condition(value) {
+                return value
             }
```
pixlwave: I'm not sure I follow why this isn't done as part of the task group but if it works then 👍

I guess my only question is, do we actually need the task group any more? It looks to me like it could be re-written with a single child task for the timeout and would be functionally the same.

UnitTests/Sources/TestUtilities/DeferredFulfillment.swift
```diff
@@ -195,36 +202,29 @@ func deferFailure<P: Publisher<P.Output, Never>>(_ publisher: P,
...
-                                                 until condition: @escaping (P.Output) -> Bool) -> DeferredFulfillment<Void> where P.Failure == Never {
-    let (stream, continuation) = AsyncStream<Void>.makeStream()
-    
-    let cancellable = publisher
-        .sink { value in
-            guard condition(value) else { return }
-            continuation.yield(())
-            continuation.finish()
-        }
+                                                 until condition: @escaping (P.Output) -> Bool) -> DeferredFulfillment<Void> where P.Failure == Never, P.Output: Sendable {
```
pixlwave, line 201: Same question on this one too :)

pixlwave (approved):
Nice, that's way smaller than I had expected 👍

UnitTests/Sources/ClassicAppMediaLoaderTests.swift
Velin92: doesn't  atter too much it's test code here

UnitTests/Sources/NSItemProviderTests.swift
Velin92: because require macro does some weird stuff behind the scenes, since it uses an autoclosure inside in the signature, and essentially violates the isolation of the NSItemProvider.
TLDR... doesn't compile

pixlwave: Riiight an autoclosure of course. Thanks that makes sense.

UnitTests/Sources/ClassicAppMediaLoaderTests.swift
pixlwave: Ok sure, but equally I don’t think we should normalise the use of unsafe code anywhere when the fix is trivial, it’s too easy to end up creating bad habits that way.

Velin92: Sure but in this case the access was on a test that was bound on the main actor so it didn't make a huge difference (it just needs to be unsafe because it's set in the override of the start loading which is nonisolated by default, but is always read by the MainActor). But okay I'll add a Mutex for it.

UnitTests/Sources/TestUtilities/DeferredFulfillment.swift
Velin92: you're right a single task is enough, I think this was done because the structured-concurrency.md said "prefer structured task groups over unstructured Task {}." Which I guess was interpreted to be done also in case of a single Task

UnitTests/Sources/TestUtilities/DeferredFulfillment.swift
Velin92: as above

UnitTests/Sources/TestUtilities/DeferredFulfillment.swift
pixlwave: Nice, this looks neat now 🙌

PR #5785, opened 2026-06-24
Fix media loading failures when retrying with a closed store

stefanceriu:
`homeserverReachabilityPublisher` was driven directly by the sync service state so stopping sync on pause still reported .reachable. MediaProvider.loadImageRetryingOnReconnection would then retry against a closed store, fail, and bail out permanently because it treated the reachable state as a non-recoverable error.

Make reachability a function of both the desired service state and the latest sync state, computed in one place: store the sync state in the listener and recompute reachability from it, reporting unreachable whenever the client isn't meant to be running. Recompute on suspend and on resume.

### DIFF
--- a/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
@@ -285,7 +285,7 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
             .store(in: &cancellables)
         
         let reachabilityNotificationID = "io.element.elementx.reachability.notification"
-        userSession.clientProxy.homeserverConnectivityPublisher.removeDuplicates()
+        userSession.clientProxy.homeserverReachabilityPublisher.removeDuplicates()
             .combineLatest(flowParameters.appMediator.networkMonitor.reachabilityPublisher.removeDuplicates())
             .receive(on: DispatchQueue.main)
             .sink { [weak self] homeserverReachability, networkReachability in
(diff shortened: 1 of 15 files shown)


pixlwave (commented):
I think we also need to take that `syncServiceState` into account when [computing the reachability banners](https://github.com/element-hq/element-x-ios/blob/80f747985f488cbbeab239823f42c37ef414d082/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift#L291-L307) otherwise we'll show a "Server unreachable" banner whenever we stop the client:

https://github.com/user-attachments/assets/80bc99bb-84d1-4375-b78f-6e7aa8c9a614

stefanceriu:
> I think we also need to take that `syncServiceState` into account when [computing the reachability banners](https://github.com/element-hq/element-x-ios/blob/80f747985f488cbbeab239823f42c37ef414d082/ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift#L291-L307) otherwise we'll show a "Server unreachable" banner whenever we stop the client:

Ah damn it, I only thought of the background suspension case. Problem is the 2 behaviors conflict with each other, we'll need to figure out some other way.

pixlwave (approved):
> we'll need to figure out some other way.

Maybe in a future PR we use a new enum `reachable, unreachable, paused` and then go back to a single publisher, but what you've got works for now 👍

stefanceriu:
Yeah no, I'm still messing around with it. I'm trying the new enum case as we speak.

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
pixlwave, line 1: 👌

UnitTests/Sources/UserSessionFlowCoordinatorTests.swift
```diff
@@ -228,6 +228,14 @@ struct UserSessionFlowCoordinatorTests {
...
         #expect(userIndicatorController.submitIndicatorDelayCallsCount == 3)
         #expect(retractReachabilityIndicatorCallsCount == 2)
+        
+        // When the client is suspended.
+        homeserverReachabilitySubject.send(.suspended)
+        try await Task.sleep(for: .milliseconds(100))
+        
+        // Then no unreachable indicator should be shown as the pause is intentional.
+        #expect(userIndicatorController.submitIndicatorDelayCallsCount == 3)
+        #expect(retractReachabilityIndicatorCallsCount == 3)
```
pixlwave, line 238: Nice 😎

pixlwave (approved):
Ah yeah that looks great, giving all the call sites the info to decide on a case-by-case basis 👍

PR #5790, opened 2026-06-24
Fix a crash on macOS when using QuickLook.

pixlwave:
Seems QuickLook on macOS uses the quicklook properties on a background thread so the default main actor is violated. This affected all instances of QuickLook including avatars, timeline media etc.


stefanceriu (approved):
👌

PR #5793, opened 2026-06-25
Limit conversations a message can be forwarded to

### LINKED ISSUE #5749: Limit number of selectable rooms to 10
In order to limit abuse, we decided to limit number of selectable rooms when sharing or forwarding message to 10.

When this number is reached, the RadioButton in the list of rooms are disabled. For instance, the limit is set to 1 in this screenshot:

Image

Context: https://github.com/element-hq/element-x-android/pull/7013

Creating the issue to ensure that iOS and Android are aligned, sorry if this has already been implemented!

mredig:
Closes #5749 - limits the number of rooms a message can be forwarded to within the app to 10 to avoid abuse. As mentioned in the issue, sharing from outside the app already only allows sending to a single user, so this only addresses in app forwarding.

### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Screens/MessageForwardingScreen/MessageForwardingScreenCoordinator.swift
+++ b/ElementX/Sources/Screens/MessageForwardingScreen/MessageForwardingScreenCoordinator.swift
@@ -10,10 +10,13 @@ import Combine
 import SwiftUI
 
 struct MessageForwardingScreenCoordinatorParameters {
+    static let defaultMaxRoomSelectionCount = 10
+    
     let forwardingItem: MessageForwardingItem
     let userSession: UserSessionProtocol
     let roomSummaryProvider: RoomSummaryProviderProtocol
     let userIndicatorController: UserIndicatorControllerProtocol
+    var maxRoomSelectionCount: Int = MessageForwardingScreenCoordinatorParameters.defaultMaxRoomSelectionCount
 }
 
 enum MessageForwardingScreenCoordinatorAction {
@@ -34,7 +37,8 @@ final class MessageForwardingScreenCoordinator: CoordinatorProtocol {
         viewModel = MessageForwardingScreenViewModel(forwardingItem: parameters.forwardingItem,
                                                      userSession: parameters.userSession,
                                                      roomSummaryProvider: parameters.roomSummaryProvider,
-                                                     userIndicatorController: parameters.userIndicatorController)
+                                                     userIndicatorController: parameters.userIndicatorController,
+                                                     maxRoomSelectionCount: parameters.maxRoomSelectionCount)
     }
     
     func start() {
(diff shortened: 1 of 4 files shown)


ElementX/Sources/Screens/MessageForwardingScreen/MessageForwardingScreenCoordinator.swift
```diff
@@ -10,10 +10,13 @@ import Combine
 import SwiftUI
 
 struct MessageForwardingScreenCoordinatorParameters {
+    static let defaultMaxRoomSelectionCount = 10
+    
     let forwardingItem: MessageForwardingItem
     let userSession: UserSessionProtocol
     let roomSummaryProvider: RoomSummaryProviderProtocol
     let userIndicatorController: UserIndicatorControllerProtocol
+    var maxRoomSelectionCount: Int = MessageForwardingScreenCoordinatorParameters.defaultMaxRoomSelectionCount
```
Velin92: I guess we can remove the static let, and just have this be a simple let = 10 for now, since there is no need to change it.

Velin92: Actually I guess we can remove this value entirely, and just rely on the value of the view state directly

ElementX/Sources/Screens/MessageForwardingScreen/MessageForwardingScreenModels.swift
```diff
@@ -16,8 +16,13 @@ enum MessageForwardingScreenViewModelAction {
 
 struct MessageForwardingScreenViewState: BindableState {
     var rooms: [MessageForwardingRoom] = []
-    var selectedRoomIDs = [String]()
+    var selectedRoomIDs: Set<String> = []
+    var maxRoomSelectionCount = MessageForwardingScreenCoordinatorParameters.defaultMaxRoomSelectionCount
```
Velin92: Let's have this be a let, and default it to 10, no need of a static value or passing the value from the coordinator.

Velin92 (changes requested):
Some minor changes requested, mostly LGTM

ElementX/Sources/Screens/MessageForwardingScreen/MessageForwardingScreenModels.swift
mredig: A couple thoughts on this:

1. Without a named constant, 10 becomes a magic number with no context about what it represents or why it's that value. Are you sure that's what you want?

2. This was intentionally designed to match Android's implementation -- the original issue explicitly links to the Android PR as context and asks for iOS/Android alignment. Android parameterizes `maxNumberOfRooms` all the way from entry point params through to state, with a named `DEFAULT_MAX_NUMBER_OF_ROOMS` constant. The current approach mirrors that.

Velin92: `let maxRoomSelectionCount = 10` for me is already giving out the context of what it is and what it does pretty well.
Also I do understand that this is designed to match Android implementation, but the two repos have different architecture and architectural design choices, even if they share designs copies and behaviour, internally we should follow the principles of the repo.

mredig: Updated as requested.

Velin92 (approved):

PR #5794, opened 2026-06-25
Make LinkNewDeviceScreen fall back to the App Lock PIN code when there's no passcode.

pixlwave:
Next part of the QR changes following on from #5773. This PR integrates the `AppLockScreen` into the `LinkNewDeviceFlow`, showing it when app lock is enabled and the user doesn't have a passcode set on their device.

Remaining:

- [ ] Potentially switch to `AppLockSetupPINScreen` instead of `AppLockScreen` (seems a better fit now that I think about it).
- [x] Verify the behaviour when denying access to Face ID and without a PIN set.
    - [x] And add specific tests (or rename existing tests if already covered).


stefanceriu (approved):
Nice, looks good to me! 👍

PR #5797, opened 2026-06-26
Use iOS 26 Observation for Unit and Preview tests.

Velin92:
Unit tests and preview tests now have iOS 26 as min target, and the preview tests that use observe will use iOS 26, the old function is still available since is used in production code but is a deprecated disfavored overload.
Had to tweak a bit the preview dictionary stencil to make a11y tests still work properly.


ElementX/Sources/Other/Extensions/Observable.swift
```diff
@@ -7,13 +7,26 @@
...
-    /// Creates an async stream for the specified property on this object. We probably won't need this once SE-0475 is available:
-    /// https://github.com/swiftlang/swift-evolution/blob/main/proposals/0475-observed.md
+    /// Creates an async sequence for the specified property on this object, emitting the current
+    /// value and then every subsequent change, backed by the Observation `Observations` sequence (SE-0475).
     ///
     /// - Parameter property: The key path to the property you would like to observe.
+    @available(iOS 26.0, *)
+    @MainActor func observe<Value: Sendable>(_ property: KeyPath<Self, Value>) -> Observations<Value, Never> {
+        Observations { self[keyPath: property] }
+    }
```
pixlwave: Can't we delete this new function and use `Observations` inlined directly?

Velin92: yes but this means changing the code in 200+ places, are we sure?

pixlwave (approved):
😎

ElementX/Sources/Other/Extensions/Observable.swift
```diff
@@ -7,13 +7,26 @@
...
 
 import Foundation
+import Observation
 import Synchronization
 
 extension Observable {
-    /// Creates an async stream for the specified property on this object. We probably won't need this once SE-0475 is available:
-    /// https://github.com/swiftlang/swift-evolution/blob/main/proposals/0475-observed.md
+    /// Creates an async sequence for the specified property on this object, emitting the current
+    /// value and then every subsequent change, backed by the Observation `Observations` sequence (SE-0475).
```
pixlwave: Maybe this should simply say "A convenience wrapper around `Observations`" given we've just learnt it doesn't really emit the current value.

PR #5798, opened 2026-06-26
Replace the `@UserPreference` property wrapper with a new macro version.

pixlwave:
An alternative to #5779, this PR shows what a macro version of the property wrapper would look like (and how it would integrate into the project).

- The new `BuildExtensions` package holds the macro.
    - It can obviously also be a place to store future macros all accessed via `import Macros`
    - It can also become a place to store build and command plugins.
- This package is added to all the targets that use `AppSettings`
- `AppSettings` then becomes much simpler:
    - It removes the `nonsolated(unsafe)` annotations entirely.
    - It removes the verbose init needed by the storage injection.
    - It removes the need for `UserDefaultsKeys` as these are derrived from the property name.
        - There's still a `key:` parameter escape hatch to allow us to change keys or for existing preferences that are backed by a different key name.

The expanded macro looks like this:
| UserDefaults | Volatile |
| - | - |
| Screenshot 2026-06-26 at 2 17 45 pm | Screenshot 2026-06-26 at 2 18 18 pm |


stefanceriu (approved):
🧑‍🍳 👌 !

PR #5804, opened 2026-06-29
Fix a bug where scanning a universal link as a QR code would open the app without handling the link.

pixlwave:
I'm not sure how long this has been the case but as mentioned by https://www.hackingwithswift.com/forums/swiftui/onopenurl-not-being-called/24714 we need to handle an `NSUserActivityTypeBrowsingWeb` activity to get the URL from a scanned QR code as `onOpenURL` isn't used for this path 🤷‍♂️


stefanceriu (approved):
🤦‍♂️🤦‍♂️🤦‍♂️

PR #5805, opened 2026-06-29
Fix MediaLoader EXC_BAD_ACCESS `ongoingRequests` crashes

stefanceriu:
and stop mutating the underlying dictionary from SDK threads.


pixlwave (approved):
I much prefer the shape of this, but I find it would be interesting to know why Rust isn't honouring the `actor` here. Seems like something we should figure out and try to fix.

stefanceriu:
> I much prefer the shape of this, but I find it would be interesting to know why Rust isn't honouring the `actor` here. Seems like something we should figure out and try to fix.

Presumably because it's all `preconcurrency` and all bets are off.

PR #5822, opened 2026-07-01
Stop reserving space for the non-existant icon background in ListRowLabel on iOS 26.

pixlwave:
When we adopted Liquid Glass we hid the icon background in `ListRowLabel`. However looking at the Figma while working on a custom row I noticed that we were also meant to stop reserving space for the background too. This PR makes those changes.

It also fixes NotificationSettingsEditScreenRoomCell to use the correct `.avatar` ListRowLabel style (and uses the standard avatar size).


stefanceriu (approved):
Sure! 🙈

PR #5825, opened 2026-07-01
User Status preparation.

pixlwave:
- Adds a feature flag (unused for now)
- Refactors `UserProfileProxy` and `SearchUsersResultsProxy` to drop the `Proxy` (they've never proxied anything and I've wanted to get round to this task for a very long time now).
- Replaces the separate `displayName` & `avatarURL` publishers (and load methods) on `ClientProxy` with a combined `userProfile` publisher (and load method).
    - The avatar URL remains cached although it is mega weird that we don't cache the display name. By the end of this project I will make sure that the whole profile is cached Rust side (I'm just not sure if it will be via SSS or manually like the avatar URL yet).


stefanceriu (approved):
Absolutely love it! 🦸‍♂️

PR #5826, opened 2026-07-01
Content Scanner Service

Velin92:
Based on Android's implementation, for now this is not wired yet to SwiftUI, but is just a system to essentially allow to scan a MediaSource and cache the scanning result, so that it can then be observed to render specific error placeholders on the UI side.

The implementations use the actor types, because they hold mutable states that can't be accessed externally, exist in a single instance per client, and offer async access to their functions, also this makes them sendable quite easily, so I thought in this specific case seemed the best choice, but happy to change it if is not convincing

### DIFF
--- a/ElementX/Sources/Other/SwiftUI/ViewModel/StateStoreViewModel.swift
+++ b/ElementX/Sources/Other/SwiftUI/ViewModel/StateStoreViewModel.swift
@@ -66,8 +66,9 @@ class StateStoreViewModel<State: BindableState, ViewAction> {
         /// Intentionally non-generic so that it doesn't grow uncontrollably
         let mediaProvider: MediaProviderProtocol?
         
-        /// An optional content scanning service so that views can validate media themselves,
-        /// `nil` when no content scanner is configured for the server.
+        /// An optional content scanning service so that views can validate media themselves.
+        /// `nil` when no content scanner is configured for the server, or when the screen
+        /// intentionally doesn't provide one because it has no media to scan.
         let contentScannerService: ContentScannerServiceProtocol?
         
         /// Set-able/Bindable access to the bindable state.
--- a/ElementX/Sources/Other/SwiftUI/ViewModel/StateStoreViewModel.swift
+++ b/ElementX/Sources/Other/SwiftUI/ViewModel/StateStoreViewModel.swift
@@ -30,8 +30,8 @@ class StateStoreViewModel<State: BindableState, ViewAction> {
         set { context.viewState = newValue }
     }
     
-    init(initialViewState: State, mediaProvider: MediaProviderProtocol? = nil) {
-        context = Context(initialViewState: initialViewState, mediaProvider: mediaProvider)
+    init(initialViewState: State, mediaProvider: MediaProviderProtocol? = nil, contentScannerService: ContentScannerServiceProtocol? = nil) {
+        context = Context(initialViewState: initialViewState, mediaProvider: mediaProvider, contentScannerService: contentScannerService)
         context.viewModel = self
     }
     
@@ -66,6 +66,10 @@ class StateStoreViewModel<State: BindableState, ViewAction> {
         /// Intentionally non-generic so that it doesn't grow uncontrollably
         let mediaProvider: MediaProviderProtocol?
         
+        /// An optional content scanning service so that views can validate media themselves,
+        /// `nil` when no content scanner is configured for the server.
+        let contentScannerService: ContentScannerServiceProtocol?
+        
         /// Set-able/Bindable access to the bindable state.
         subscript<T>(dynamicMember keyPath: WritableKeyPath<State.BindStateType, T>) -> T {
             get { viewState.bindings[keyPath: keyPath] }
@@ -78,9 +82,10 @@ class StateStoreViewModel<State: BindableState, ViewAction> {
             viewModel?.process(viewAction: viewAction)
         }
         
-        fileprivate init(initialViewState: State, mediaProvider: MediaProviderProtocol?) {
+        fileprivate init(initialViewState: State, mediaProvider: MediaProviderProtocol?, contentScannerService: ContentScannerServiceProtocol?) {
             self.viewState = initialViewState
             self.mediaProvider = mediaProvider
+            self.contentScannerService = contentScannerService
         }
     }
 }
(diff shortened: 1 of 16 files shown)


ElementX/Sources/Services/ContentScanner/ContentScannerProxy.swift
```diff
@@ -0,0 +1,35 @@
...
+//
+
+import Foundation
+
+@preconcurrency import MatrixRustSDK
+
+actor ContentScannerProxy: ContentScannerProxyProtocol {
+    private let contentScanner: ContentScanner
+    /// The client is held weakly for the same reason as in `MediaLoader` - to avoid keeping the
+    /// underlying `MatrixRustSDK.Client` alive longer than the owning `ClientProxy`.
```
pixlwave, line 15: Presumably SwiftFormat turned this into a doc comment instead of a regular one?

ElementX/Sources/Services/Client/ClientProxyProtocol.swift
```diff
@@ -135,6 +135,8 @@ protocol ClientProxyProtocol: AnyObject {
     
     var mediaLoader: MediaLoaderProtocol { get }
     
+    var contentScanner: ContentScannerProxyProtocol? { get }
```
pixlwave, line 136: Should this actually be `contentScannerService`? Otherwise that one is unused and I'm not yet sure where it would live.

ElementX/Sources/Services/ContentScanner/ContentScannerProxy.swift
```diff
@@ -0,0 +1,35 @@
+//
+// Copyright 2026 Element Creations Ltd.
+// Copyright 2026 New Vector Ltd.
```
pixlwave: Applies to all the files, I don't think Claude used the new file header properly…

```suggestion
// Copyright 2026 Element Creations Ltd.
```

ElementX/Sources/Services/ContentScanner/ContentScannerService.swift
```diff
@@ -0,0 +1,55 @@
...
+
+actor ContentScannerService: ContentScannerServiceProtocol {
+    private let contentScannerProxy: ContentScannerProxyProtocol
+    private let validationCache: EventContentValidationCacheProtocol
+    
+    /// The media source URLs that have already been scanned, per event.
+    private var scannedSources: [String: Set<String>] = [:]
+    /// The media source URLs whose scan is currently in flight, per event. Used to dedupe concurrent
+    /// scans of the same source (the actor doesn't hold isolation across the `await`).
+    private var inFlightSources: [String: Set<String>] = [:]
```
pixlwave: I don't understand the complexity here:
- Why are `scannedSources`/`inFlightSources` keyed by eventID?
    - Doesn't this mean that you will scan the same media twice if it is in multiple events (e.g. has been forwarded)
    - How will this work for avatars?
- Why do we need `scannedSources` and a `validationCache`?
    - It will always be in both so we checking 2 things for little benefit.
    - Surely we can just use the cache here?

ElementX/Sources/Services/ContentScanner/EventContentValidationCache.swift
```diff
@@ -0,0 +1,41 @@
...
+import Combine
+import Foundation
+import Synchronization
+
+final nonisolated class EventContentValidationCache: EventContentValidationCacheProtocol {
+    private let subjects = Mutex<[String: CurrentValueSubject<ContentValidation, Never>]>([:])
+    
+    func validationPublisher(for eventID: String) -> CurrentValuePublisher<ContentValidation, Never> {
+        subject(for: eventID).asCurrentValuePublisher()
+    }
```
pixlwave: Why is a cache responsible for coordinating the current status? Surely it should cache `safe: Bool` and the service should be responsible for coordinating the in-flight state?

pixlwave (changes requested):
Honestly, this feels rather over-engineered. Can rework it to follow existing patterns in the app. For example we do request coalescing in `MediaLoader` when loading media in a much simpler way. I appreciate there's an extra intermediate step of `scanning` here compared to loading and loaded/failed, but it should still be able to work that way.

ElementX/Sources/Services/Client/ClientProxyProtocol.swift
Velin92: so integrate together both the scanner proxy and the service into one object? The proxy is just the wraper on the SDK type but I guess we can fuse them together

Velin92:
@pixlwave after the talk we had on EC, I have update the PR, which is quite different now, give it another check

ElementX/Sources/Services/ContentScanner/ContentScannerServiceProtocol.swift
```diff
@@ -0,0 +1,27 @@
...
+/// Scans media sources through the content scanner, caching the verdicts by URL so that each source
+/// is only ever scanned once. Mirrors ``MediaProviderProtocol``'s shape: a synchronous cache lookup
+/// for views that already have a verdict and an async method to perform the scan.
+nonisolated protocol ContentScannerServiceProtocol: Sendable {
+    /// The cached scan verdict for the given source: `true` when the content is safe, `false` when
+    /// it is unsafe, or `nil` when the source hasn't been scanned yet.
+    func scanResultFromSource(_ source: MediaSourceProxy) -> Bool?
+    
+    /// Returns the cached verdict for the given source, scanning it first when needed.
+    /// Failed scans aren't cached so that they can be retried.
```
pixlwave: The comment on this one seems slightly backwards. Maybe it should be something more like:

"Starts a scan for the given source unless there's already a cached verdict, in which case that is returned".

ElementX/Sources/Services/ContentScanner/ContentScannerService.swift
pixlwave, line 1: Woah neat, that's even simpler than I expected 👌

ElementX/Sources/Other/SwiftUI/ViewModel/StateStoreViewModel.swift
```diff
@@ -66,6 +66,10 @@ class StateStoreViewModel<State: BindableState, ViewAction> {
         /// Intentionally non-generic so that it doesn't grow uncontrollably
         let mediaProvider: MediaProviderProtocol?
         
+        /// An optional content scanning service so that views can validate media themselves,
+        /// `nil` when no content scanner is configured for the server.
```
pixlwave: Small nit on this: It is `nil` when no content scanner is configured *or* when we didn't set it on the view model, intentionally or otherwise. Which is slightly concerning as when we forget to set a `mediaProvider` the app crashes, but here we'll silently not be scanning media on a particular screen.

(obviously applies to V2 as well)

pixlwave (approved):
Nice 😎

PR #5828, opened 2026-07-02
Remove Hyperion from the Compound Inspector app.

pixlwave:
Hyperion is barely working at this stage anyway and now the [repo](https://github.com/willowtreeapps/Hyperion-iOS) has been removed.

I've run `pod deintegrate` and deleted any usage of the pod from the inspector app.

Given that the `.xcworkspace` has also been deleted, I think we will also need to update XcC so that it builds the right thing.


stefanceriu (approved):
🔥

pixlwave:
CI failures are SwiftFormat. But they're not relevant here.

PR #5829, opened 2026-07-02
Add message support to the new search tab

stefanceriu:
Adds a Messages tab to global search, next to Rooms.

It's backed by a new `SearchServiceProxy` wrapping the SDK's `SearchService`, exposed on the `ClientProxy` and using the session's search index store.

Message results show the room's name and avatar with a sender-prefixed preview. Image/video/audio/file matches get a thumbnail (or type icon) with filename and size.

Polls, stickers, redactions and live locations are supported too but depend on https://github.com/matrix-org/matrix-rust-sdk/pull/6710


ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -98,6 +98,8 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
         
         if flowParameters.appSettings.globalSearchEnabled, #available(iOS 26.0, *) {
             let searchCoordinator = SearchScreenCoordinator(parameters: .init(roomSummaryProvider: flowParameters.userSession.clientProxy.alternateRoomSummaryProvider,
+                                                                              searchService: flowParameters.userSession.clientProxy.searchService,
+                                                                              clientProxy: flowParameters.userSession.clientProxy,
```
pixlwave: This one seems redundant seeing as we already pass in the client proxy (and we can easily set a mock on the ClientProxyMock)?

ElementX/Sources/Other/Avatars.swift
```diff
@@ -87,6 +87,7 @@ enum UserAvatarSizeOnScreen {
     case threadSummary
     case map
     case classicAppAccount
+    case globalSearch
```
pixlwave: Similar to the last PR I reviewed, should this simply be `search` (with `roomSearch` being added later if we ever have a screen for that)?

ElementX/Sources/Screens/SearchScreen/View/SearchScreen.swift
```diff
@@ -111,18 +158,54 @@ struct SearchScreen: View {
...
+                        if message == context.viewState.messages.last {
+                            context.send(viewAction: .reachedBottom)
+                        }
+                    }
+            }
+        }
+        .compoundList(.plain)
+    }
+    
+    /// Hidden buttons that switch tabs via ⌘1/⌘2, the common shortcut for jumping to tab N.
```
pixlwave, line 177: 👌

ElementX/Sources/Screens/SearchScreen/View/SearchScreen.swift
```diff
@@ -111,18 +158,54 @@ struct SearchScreen: View {
...
+    
+    private func selectCurrent() {
+        guard let selectedID else { return }
+        switch context.viewState.bindings.searchMode {
+        case .rooms:
+            context.send(viewAction: .selectRoom(roomID: selectedID))
+        case .messages:
+            guard let message = context.viewState.messages.first(where: { $0.id == selectedID }) else { return }
+            context.send(viewAction: .selectMessage(roomID: message.roomID, eventID: message.id))
+        }
```
pixlwave, line 208: Not important for right now, part of me feels like there should be a single `selectItem` view action and then this logic could be in the view model and tested.

ElementX/Sources/Screens/SearchScreen/View/SearchScreen.swift
```diff
@@ -192,6 +275,140 @@ private struct SearchScreenRoomCellButtonStyle: ButtonStyle {
...
+            icon(\.audio)
+        case .image(let thumbnail, let blurhash):
+            thumbnailView(thumbnail, blurhash: blurhash, isVideo: false)
+        case .video(let thumbnail, let blurhash):
+            thumbnailView(thumbnail, blurhash: blurhash, isVideo: true)
+        }
+    }
+    
+    private func icon(_ icon: KeyPath<CompoundIcons, Image>) -> some View {
+        CompoundIcon(icon, size: .medium, relativeTo: .body)
```
pixlwave: Those are the defaults so can just be

```suggestion
        CompoundIcon(icon)
```

ElementX/Sources/Screens/SearchScreen/View/SearchScreen.swift
```diff
@@ -192,6 +275,140 @@ private struct SearchScreenRoomCellButtonStyle: ButtonStyle {
...
+            RoomAvatarImage(avatar: message.roomAvatar,
+                            avatarSize: .room(on: .globalSearch),
+                            mediaProvider: context.mediaProvider)
+                .dynamicTypeSize(dynamicTypeSize < .accessibility1 ? dynamicTypeSize : .accessibility1)
+                .accessibilityHidden(true)
+        }
+    }
+}
+
+private struct SearchScreenMediaPreviewView: View {
```
pixlwave, line 349: Again not specifically for now, but would be worth seeing what this one is like with voiceover (I think there might need to be some hiding and some combining)

ElementX/Sources/Screens/SearchScreen/View/SearchScreen.swift
```diff
@@ -192,6 +275,140 @@ private struct SearchScreenRoomCellButtonStyle: ButtonStyle {
...
+    let preview: SearchScreenMediaPreview
+    let mediaProvider: MediaProviderProtocol?
+    
+    private static let mediaSize: CGFloat = 36
```
pixlwave: This should be a (non-static) `ScaledMetric` (or we should use `scaledFrame`).

ElementX/Sources/Screens/SearchScreen/View/SearchScreen.swift
```diff
@@ -242,13 +459,48 @@ private final class KeyNavigatingSearchTextField: UISearchTextField {
...
+                                                         initialSearchQuery: "Foundation",
+                                                         initialSearchMode: .messages)
+    static let loadingMessagesViewModel = SearchScreenViewModel(roomSummaryProvider: RoomSummaryProviderMock(.init(state: .loaded([]))),
+                                                                searchService: makeSearchService(paginationState: .loading),
+                                                                clientProxy: makeClientProxy(),
+                                                                mediaProvider: MediaProviderMock(.init()),
+                                                                initialSearchQuery: "Foundation",
+                                                                initialSearchMode: .messages)
+    
+    private static func makeSearchService(results: [SearchServiceResult] = [], paginationState: SearchServicePaginationState = .idle(endReached: true)) -> SearchServiceProxyMock {
```
pixlwave: Could we keep this beneath the `var previews` please 😄

ElementX/Sources/Screens/SearchScreen/SearchScreenModels.swift
```diff
@@ -39,3 +60,99 @@ struct SearchScreenRoom: Identifiable, Equatable {
...
+        roomName = roomSummary?.name ?? result.roomID
+        roomAvatar = roomSummary?.avatar ?? .room(id: result.roomID, name: roomSummary?.name, avatarURL: nil)
+        senderName = isOutgoing ? L10n.commonYou : result.sender.disambiguatedDisplayName ?? result.sender.id
+        content = result.content
+        timestamp = result.timestamp
+    }
+    
+    var preview: AttributedString? {
+        guard let messageBody else { return nil }
+        return AttributedString("\(senderName): ") + messageBody
```
pixlwave, line 85: I realise this is in the wrong place, but seems worth noting that we should probably be using the `RoomEventStringBuilder` with an appropriate configuration to build these.

ElementX/Sources/Screens/SearchScreen/SearchScreenViewModel.swift
```diff
@@ -35,18 +44,54 @@ class SearchScreenViewModel: SearchScreenViewModelType, SearchScreenViewModelPro
...
             for await searchQuery in stream {
                 self?.updateFilter(for: searchQuery)
             }
         }
         
+        // Flip the loading indicator on the moment the user starts typing, ahead of the debounced
+        // query above, so the empty state doesn't flash while the first search is still pending.
+        loadingObservationTask = Task { [weak self] in
+            guard let stream = self?.context.observe(\.viewState.bindings.searchQuery).removeDuplicates() else { return }
+            for await searchQuery in stream {
```
pixlwave: Generally I would set up the stream outside of the task so the task is *only* responsible for the `for await in`
```suggestion
        let searchQueryStream = context.observe(\.viewState.bindings.searchQuery).removeDuplicates()
        loadingObservationTask = Task { [weak self] in
            for await searchQuery in searchQueryStream {
```

ElementX/Sources/Screens/SearchScreen/SearchScreenViewModel.swift
```diff
@@ -56,15 +101,22 @@ class SearchScreenViewModel: SearchScreenViewModelType, SearchScreenViewModelPro
         
         switch viewAction {
         case .appeared:
-            // The provider is shared, so other consumers may have changed its filter while we were off-screen.
-            // Re-apply ours on every appearance to keep the displayed results in sync with the query.
```
pixlwave, line 60: Unless this is no longer relevant, I think it was a useful comment?

ElementX/Sources/Screens/SearchScreen/SearchScreenViewModel.swift
```diff
@@ -97,12 +156,6 @@ class SearchScreenViewModel: SearchScreenViewModelType, SearchScreenViewModelPro
         }
     }
     
-    /// The actual range values don't matter as long as they contain the lower
-    /// or upper bounds. updateVisibleRange is a hybrid API that powers both
-    /// sliding sync visible range update and list paginations.
-    /// For lists other than the home screen one we don't care about visible ranges,
-    /// we just need the respective bounds to be there to trigger a next page load or
-    /// a reset to just one page.
```
pixlwave, line 105: And this?

ElementX/Sources/Services/Search/SearchServiceProxy.swift
```diff
@@ -0,0 +1,144 @@
...
+    
+    // MARK: - Private
+    
+    private func handleResultUpdates(_ updates: [SearchServiceResultsUpdate]) {
+        var results = resultsSubject.value
+        
+        for update in updates {
+            switch update {
+            case .append(let values):
+                results.append(contentsOf: values.compactMap { self.makeResult($0) })
```
pixlwave: I *think* these maps can be like so?

```suggestion
                results.append(contentsOf: values.compactMap(makeResult))
```

ElementX/Sources/Services/Search/SearchServiceProxy.swift
```diff
@@ -0,0 +1,144 @@
...
+                case .message(let msg):
+                    content = .message(timelineItemFactory.buildMessageTimelineItemContent(messageType: msg.msgType,
+                                                                                           senderID: result.sender,
+                                                                                           senderDisplayName: sender.displayName))
+                case .poll(let question, _, _, _, _, _, _):
+                    content = .poll(question: question)
+                case .sticker(let body, _, _):
+                    content = .message(.text(.init(body: body)))
+                case .redacted:
+                    content = .redacted
```
pixlwave, line 127: Maybe return `nil` here? (Although presumably its never going to happen)

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemFactory.swift
```diff
@@ -895,7 +895,7 @@ nonisolated struct RoomTimelineItemFactory: RoomTimelineItemFactoryProtocol {
     
     // MARK: - Helpers
```
pixlwave: Might want to move this one down?

ElementX/Sources/Screens/SearchScreen/SearchScreenViewModel.swift
```diff
@@ -13,19 +13,28 @@ typealias SearchScreenViewModelType = StateStoreViewModelV2<SearchScreenViewStat
 
 class SearchScreenViewModel: SearchScreenViewModelType, SearchScreenViewModelProtocol {
     private let roomSummaryProvider: RoomSummaryProviderProtocol
+    private let searchService: SearchServiceProxyProtocol
```
pixlwave, line 16: Wasn't sure where to put this one, it obviously applies to the whole screen module:

Naming-wise it might be slightly weird to give the `SearchScreen` a `SearchService` that is only responsible for message search and has no part in room search.

pixlwave (approved):
Very cool, nothing major just small thoughts 🔍

ElementX/Sources/Screens/SearchScreen/SearchScreenViewModel.swift
stefanceriu: Yeah, it's not ideal yet but the SearchService will take care of all of it in the next iteration

pixlwave: You even told me this on the call and I completely forgot about it 🤦‍♂️

ElementX/Sources/Screens/SearchScreen/SearchScreenViewModel.swift
stefanceriu: Honestly didn't find them particularly interesting but I did have all of that in my head. I added them back.

PR #5832, opened 2026-07-03
Update the readme to fix the unit tests shield and remove the SonarCloud ones

stefanceriu:


pixlwave (approved):
Ha this could've been an edit directly to `develop` if you ask me 😅

PR #5833, opened 2026-07-03
Fix various project warnings

stefanceriu:


compound-ios/Sources/Compound/List/ListRowButtonStyle.swift
```diff
@@ -8,7 +8,7 @@
 
 import SwiftUI
 
-// TODO: Check if the primitive style is actually needed now the insets are part of ListRow.
+// TODO: @pixlwave Check if the primitive style is actually needed now the insets are part of ListRow.
```
Velin92, line 11: are we sure we want to leave the tags in the code?

stefanceriu: That bit was intentional, I wanted to see how annoying it gets 🙈

ElementX/Sources/Services/AppLock/AppLockService.swift
```diff
@@ -229,3 +229,12 @@ class AppLockService: AppLockServiceProtocol {
         appSettings.appLockNumberOfPINAttempts = 0
     }
 }
+
+extension LAContext {
```
pixlwave: Oh that's a neat way of doing it 👌

ElementX/Sources/Services/AppLock/AppLockService.swift
pixlwave, line 1: There's a few more tweaks to be made to this file as `evaluatedPolicyDomainState` is only available after evaluating (or checking if you can evaluate), whereas `domainState` is meant to be available all of the time, so there may be some simplification that can be done.

stefanceriu: Okay, talked about this separately, will drop it from here and it will get handled separately.

pixlwave (approved):

PR #5838, opened 2026-07-04
Report user presence on app lifecycle transitions

lhjt:
Sets the user's presence from the app lifecycle:

- Foreground active → `online` (only while the Share presence setting is on, `offline` otherwise)
- Backgrounded/inactive → `unavailable` (or `offline` when not sharing)
- The NSE and background app refresh set the process presence to `offline`, so their sync requests carry `set_presence=offline` and have no presence side effect. Currently, a push actually marks an offline user as idle, since the SDK's default is `unavailable`.

The goal for this change is to report accurate presence to homeservers to enable better push notification suppression handling. Ensuring that a device instantly exits the `online` state when suspended allows the remote homeserver to send any new push notifications without delay.

I previous authored https://github.com/matrix-org/matrix-rust-sdk/pull/6672 to support this change.

**Known limitation:** if the app is force-killed while online, the user stays online until the homeserver times the presence out. A follow-up could correct this from the NSE.

No UI changes.

### Pull Request Checklist

- [ ] Voiceover enabled.


NSE/Sources/NSEUserSession.swift
```diff
@@ -132,6 +143,18 @@ final nonisolated class NSEUserSession: NSEUserSessionProtocol {
     }
 }
 
+private nonisolated struct MatrixSDKNSEPresenceSetter: NSEPresenceSetterProtocol {
+    let client: Client
+    
+    func setOfflinePresence() async {
+        do {
+            try await client.setPresence(presence: .offline, immediate: true)
```
pixlwave: I don't think it makes sense to have this set to `immediate: true` – the NSE continues to run whilst the app is in the foreground, so this will potentially result in the user's status flipping from online to offline and back again when a notification arrives whilst the user is looking at the app?

NSE/Sources/NSEUserSession.swift
```diff
@@ -19,15 +19,19 @@ nonisolated protocol NSEUserSessionProtocol {
     func roomForIdentifier(_ roomID: String) -> Room?
 }
 
+/// Testable seam for the offline presence seed. The NSE must never let a sync mark the user online or idle,
+/// so it only ever seeds `.offline`.
+nonisolated protocol NSEPresenceSetterProtocol {
+    func setOfflinePresence() async
+}
```
pixlwave: From what I can tell this isn't tested (and as there's no logic it probably doesn't make any sense to test it anyway), so we don't really need this part.

ElementX/Sources/Application/PresenceReporter.swift
```diff
@@ -0,0 +1,93 @@
+//
+// Copyright 2025 Element Creations Ltd.
+// Copyright 2022-2025 New Vector Ltd.
```
pixlwave:
```suggestion
// Copyright 2026 Element Creations Ltd.
```

ElementX/Sources/Application/AppCoordinator.swift
```diff
@@ -1176,6 +1183,7 @@ class AppCoordinator: AppCoordinatorProtocol, AuthenticationFlowCoordinatorDeleg
     @objc
     private func applicationWillResignActive() {
         MXLog.info("Application will resign active")
+        presenceReporter?.applicationWillResignActive()
```
pixlwave: We can do this directly inside of the struct with e.g.

```
NotificationCenter.default.publisher(for: UIApplication.willResignActiveNotification)
```

And if we do that, this reporter would likely no longer need to live in the `AppCoordinator` but instead could live directly on the `ClientProxy` or in the `UserSessionFlowCoordinator`.

ElementX/Sources/Application/PresenceReporter.swift
```diff
@@ -0,0 +1,93 @@
...
+            await previousTask?.value
+            guard let self, !Task.isCancelled else {
+                return
+            }
+            _ = await clientProxy.setPresence(presence, sendImmediately: true)
+        }
+    }
+}
+
+extension PresenceReporter {
```
pixlwave: Do we need this, it's annoying to have test-specific helpers in the main app. We can use techniques like putting an `#expect` inside a mock's closure for example.

ElementX/Sources/Application/PresenceReporter.swift
```diff
@@ -0,0 +1,93 @@
...
+//
+
+import Combine
+import Foundation
+
+/// Owns the app's presence reporting. It maps the two axes that determine presence — whether the app is
+/// foreground-active and whether the user shares presence — onto a single desired `ClientProxyPresence` and
+/// pushes it to the server. Sending is deduped and serialised latest-wins so a late background update can never
+/// land after a newer foreground one.
+final class PresenceReporter {
```
pixlwave: Please could we call this `PresenceService`. That would be more consistent with how we make things like this in the rest of the app.

ElementX/Sources/Services/Client/ClientProxyProtocol.swift
```diff
@@ -283,4 +289,10 @@ protocol ClientProxyProtocol: AnyObject {
     
     func setTimelineMediaVisibility(_ value: TimelineMediaVisibility) async -> Result<Void, ClientProxyError>
     func setHideInviteAvatars(_ value: Bool) async -> Result<Void, ClientProxyError>
+    
+    // MARK: - Presence
+    
+    /// Updates the process presence that rides on every future sync request. When `sendImmediately` is `true` the SDK
+    /// also issues a direct presence update instead of waiting for the next sync request to carry the new value.
+    func setPresence(_ presence: ClientProxyPresence, sendImmediately: Bool) async -> Result<Void, ClientProxyError>
```
pixlwave: Looking at how this works SDK side, could we call this one `configurePresence` seeing as it doesn't actually set the user's presence unless you pass `sendImmediately: true`. This seems more in line with what happens, it configures the presence that will be used by the sync service the next time the sync loop happens.

pixlwave (changes requested):
Thanks @lhjt, we'd love to get this fixed. I've put some comments in line, but one there's an overarching point I haven't noted:

https://github.com/element-hq/element-x-ios/pull/5838/changes/19307fd05d73d6431e761619944a98c2485527c5 and https://github.com/element-hq/element-x-ios/pull/5838/changes/a15cfc6659a238627a61476f7da2e5b1ca101949 do not belong in this PR. It appears as though you might not have the latest Xcode installed and so you're using the Swift 6.2 compiler. Please can you fix your toolchain and drop these commits as everything compiles correctly on our side without those changes. You can always find the current version of Xcode that the project is expecting in the `ci_common.sh` script:

https://github.com/element-hq/element-x-ios/blob/a2c1ca06dd83b1f2a0a4ab67631d89b675cdd79b/ci_scripts/ci_common.sh#L42

NSE/Sources/NSEUserSession.swift
pixlwave: Which thinking about it still wouldn't be fixed by just `immediate: false`…

Shouldn't we configure the NSE to not send presence values on its sync loops, rather than explicitly marking the user as `.offline` here?

lhjt: I previously had a more in depth design for this problem, specifically setting state shared between the NSE and main app so that the NSE was aware of the main app's foreground/background state. I can reintroduce that here?

> Shouldn't we configure the NSE to not send presence values on its sync loops, rather than explicitly marking the user as .offline here?

From my understanding, the spec states that omitting the presence value on a sync should be interpreted as being online by default, which is not what we want here (though please do correct me if I incorrect)

lhjt:
Thanks for the detailed review @pixlwave, and apologies for the delay in updating this PR.

I have gone ahead and updated the PR accordingly, so all concerns should be addressed (except for the rebase to remove the toolchain changes)

ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -19,12 +19,13 @@ import SwiftUI
...
     var selectedNotificationTone: NotificationTone? { get set }
-    
+    var sharePresence: Bool { get }
+
     var logLevel: LogLevel { get }
     var traceLogPacks: Set<TraceLogPack> { get }
     var bugReportRageshakeURL: RemotePreference<RageshakeConfiguration> { get }
     var contentScannerURL: RemotePreference<URL?> { get }
-    
+
```
pixlwave: Please make sure to run SwiftFormat to avoid random changes in the diffs like this.

ElementX/Sources/Application/AppCoordinator.swift
pixlwave, line 1: Please don't remove code that isn't related to your changes.

pixlwave (changes requested):
> (except for the rebase to remove the toolchain changes)

It's fine, go ahead and drop those commits as requested please. We don't have the time to review unrelated code changes I'm afraid.

NSE/Sources/NSEUserSession.swift
lhjt: Updated to set `immediate: false` and `offline`. Synapse just ignores the presence setting if set to offline, so my understanding is that it should not override the main app's presence setting, and will not set the user to online at any point in time (when operating in the background).

pixlwave:
By the way this hasn’t been forgotten about, looks great from a code perspective. I just haven’t had a chance to test it out yet.

pixlwave:
Thanks I gave this ago and it works well!

https://github.com/user-attachments/assets/508255c0-e60a-4bcd-8189-8eadfab6dfc4

(Sometimes it is a little slower than others for the server/other client to see the change but thats by no way slow enough to impact the experience imo).

pixlwave:
@lhjt Would you be interested in opening a similar PR on [element-x-android](https://github.com/element-hq/element-x-android) too so that we have parity between the 2 platforms?

lhjt:
> Thanks I gave this ago and it works well!
> 
>  Screen.Recording.2026-07-20.at.2.45.45.pm.mov 
> (Sometimes it is a little slower than others for the server/other client to see the change but thats by no way slow enough to impact the experience imo).

Thanks for testing this out @pixlwave 🙂

As for the delays in the video, I believe it can be explained if you are using synapse:

- The presence transitions in the video are almost perfectly spaced at ~10s between each transition.
- Synapse defaults `rc_presence.per_user` to 0.1 updates/second with a burst of one: one accepted update every 10 seconds. 
- This explicitly applies to the presence PUT endpoint used by this PR: [Synapse configuration](https://element-hq.github.io/synapse/latest/usage/configuration/config_documentation.html#rc_presence)

> @lhjt Would you be interested in opening a similar PR on [element-x-android](https://github.com/element-hq/element-x-android) too so that we have parity between the 2 platforms?

I would be open to bringing this to Android too. I haven’t explored the Android codebase yet, though, and I’d prefer to get the iOS implementation merged first since that’s the platform I use day to day. 

Once this is merged, I can take a look at what Android parity would involve 🙂

pixlwave (approved):

pixlwave:
> As for the delays in the video, I believe it can be explained if you are using synapse:
> 
> * The presence transitions in the video are almost perfectly spaced at ~10s between each transition.

Ah, TIL!

> I would be open to bringing this to Android too. I haven’t explored the Android codebase yet, though, and I’d prefer to get the iOS implementation merged first since that’s the platform I use day to day.
> 
> Once this is merged, I can take a look at what Android parity would involve 🙂

Thanks, well this one looks good to me!

PR #5842, opened 2026-07-06
Content scanning view for reply preview

Velin92:
This PR allows also for the reply details to be scanned and show a failed state.


ElementX/Sources/Other/SwiftUI/Views/ContentScanningView.swift
```diff
@@ -43,6 +43,7 @@ struct ContentScanningView<SafeContent: View, ScanningContent: View, UnsafeConte
     
     private let contentScannerService: ContentScannerServiceProtocol?
     private let mediaSource: MediaSourceProxy?
+    private let shouldReportFailure: Bool
```
pixlwave: Nice trick! From an API perspective, I'm wondering if this could be called something more understandable like `isSelfContained`/`isSelfContainedMedia`/`isNestedMedia` or the inverse `isMainBubbleContent`. That way you don't really need to know what reporting a failure means.

ElementX/Sources/Screens/Timeline/View/Replies/TimelineReplyView.swift
```diff
@@ -189,8 +222,51 @@ struct TimelineReplyView: View {
...
+            .background {
+                ZStack {
+                    backgroundShape.fill(backgroundColor)
+                    backgroundShape.stroke(borderColor)
+                }
+            }
+    }
+}
+
+private extension View {
```
pixlwave: Given this whole thing is private we could probably just use the modifier directly rather than having the dedicated view extension, I don't think it adds much.

ElementX/Sources/Screens/Timeline/View/Replies/TimelineReplyView.swift
```diff
@@ -189,8 +222,51 @@ struct TimelineReplyView: View {
...
+private struct RoundedCardModifier: ViewModifier {
+    var cornerRadius: CGFloat = 12
+    var padding: CGFloat = 12
+    var maxWidth: CGFloat?
+    let backgroundColor: Color
+    let borderColor: Color
+    
+    private var backgroundShape: RoundedRectangle {
+        RoundedRectangle(cornerRadius: cornerRadius)
+    }
```
pixlwave: I might be missing something here, but as far as I can tell the corner radius doesn't need to be parameterised at all?

```suggestion
    var padding: CGFloat = 12
    var maxWidth: CGFloat?
    let backgroundColor: Color
    let borderColor: Color
    
    let backgroundShape = RoundedRectangle(cornerRadius: 8)
```

ElementX/Sources/Screens/Timeline/View/Replies/TimelineReplyView.swift
```diff
@@ -189,8 +222,51 @@ struct TimelineReplyView: View {
     }
 }
 
+/// Styles a view as a rounded card with a background fill and a border.
+private struct RoundedCardModifier: ViewModifier {
```
pixlwave: The `Card` name feels weird here to me. Generally you have a collection of cards, this is just a view being embedded within another one.

ElementX/Sources/Screens/Timeline/View/Replies/TimelineReplyView.swift
```diff
@@ -15,23 +15,56 @@ enum TimelineReplyViewPlacement {
...
-            .frame(maxWidth: maxWidth, alignment: .leading)
-            .padding(4.0)
-            .background {
-                ZStack {
-                    backgroundShape.fill(.compound.bgCanvasDefault)
-                    backgroundShape.stroke(.compound.separatorPrimary)
-                }
-            }
+        // The failure isn't reported up the view hierarchy as it should only affect
+        // the reply preview and not the whole bubble of the reply itself.
```
pixlwave: With the right parameter name we wouldn't need this comment.

pixlwave (approved):
LGTM, just some small thoughts inline.

ElementX/Sources/Screens/Timeline/View/Replies/TimelineReplyView.swift
Velin92: I just don't like direct modifiers they are bit non-idiomatic... I just like the syntactic sugar of them.

pixlwave: Tbh, I don't think we even need the modifier if you'd rather go that way… The entire formatting code could go directly in the view extension for this one? That would remove the ceremony of defining all the parameters twice.

ElementX/Sources/Other/SwiftUI/Views/ContentScanningView.swift
Velin92: the thing is that this will also be reused in the media browser where there is not actually a bubble, so I would keep the `shouldReportFailure` or at most the `shouldReportFailurePreference` name

PR #5843, opened 2026-07-06
Swift format enforce isEmpty rule

Velin92:
a new rule has been added

### DIFF
--- a/Dangerfile.swift
+++ b/Dangerfile.swift
@@ -25,7 +25,7 @@ if danger.github.pullRequest.body?.isEmpty ?? true {
 }
 
 /// Check for screenshots on view changes
-let hasChangedViews = !editedFiles.filter { $0.lowercased().contains("/view") }.isEmpty
+let hasChangedViews = editedFiles.contains(where: { $0.lowercased().contains("/view") })
 if hasChangedViews {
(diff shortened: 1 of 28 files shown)


ElementX/Sources/Screens/RoomScreen/ComposerToolbar/View/MessageComposerTextField.swift
```diff
@@ -279,7 +279,7 @@ private class ElementTextView: UITextView, PillAttachmentViewProviderDelegate {
             return false
         }
         
-        return UIPasteboard.general.itemProviders.filter { !$0.isSupportedForPasteOrDrop }.isEmpty
+        return !UIPasteboard.general.itemProviders.contains(where: { !$0.isSupportedForPasteOrDrop })
```
pixlwave: I think we want to resolve that Trailing Closure Violation warning.

Dangerfile.swift
pixlwave, line 1: I think this one can use trailing closures too, even if there isn't a warning.

ElementX.xcodeproj/xcshareddata/xcschemes/PreviewTests.xcscheme
pixlwave, line 1: Is this intentionally committed?

pixlwave (approved):

PR #5847, opened 2026-07-07
Scanning states for the media events timeline screen

Velin92:
This PR adds a scanning and a failure state for the elements of the MediaEventsTimeline, precisely for images and videos (since files audio and voice messages, will reuse the existing implementation). Here however it needs to be pointed out that for files, audio, images and videos in these views, we are for now allowing the tap actions, and the reason is that this opens up a scrollable preview of the elements, which in turn will display an error state when an unsafe file is displayed (so the download prevention is still happening, and we have an extra error state view in the preview), so that it's still possible to swipe between the elements in order, but prevent the download of such elements. However this last bit of the feature will be implemented in the next PR. Will however double check with design if we intend to prevent the tap altogether or not


ElementX/Sources/Screens/MediaEventsTimelineScreen/View/TimelineViews/ScanStateMediaEventsTimelineViews.swift
```diff
@@ -0,0 +1,59 @@
...
+//
+
+import Compound
+import SwiftUI
+
+/// The grid item shown in place of a media item whilst it's being scanned by the content scanner.
+struct ScanningMediaEventsTimelineView: View {
+    var body: some View {
+        Color.compound.bgSubtleSecondary
+            .opacity(0.3)
```
pixlwave, line 15: Is that right, we're using a token and then applying opacity to it manually? Fine if it is, just double checking.

ElementX/Sources/Screens/MediaEventsTimelineScreen/View/MediaEventsTimelineScreen.swift
```diff
@@ -321,4 +321,87 @@ struct MediaEventsTimelineScreen_Previews: PreviewProvider, TestablePreview {
                                  linkMetadataProvider: LinkMetadataProvider(),
                                  timelineControllerFactory: TimelineControllerFactoryMock(.init()))
     }
+    
+    // MARK: Content scanning
+    
+    private static let scanningMediaURL: URL = "mxc://matrix.org/scanning"
+    private static let unsafeMediaURL: URL = "mxc://matrix.org/unsafe"
```
pixlwave: Maybe we should add these to our MXC URL mocks?

https://github.com/element-hq/element-x-ios/blob/9e1d95db1c6892d145aa21c84a20dbe64a9f4f8e/ElementX/Sources/Other/Extensions/URL.swift#L136

pixlwave (approved):
LGTM!

ElementX/Sources/Screens/MediaEventsTimelineScreen/View/TimelineViews/ScanStateMediaEventsTimelineViews.swift
Velin92: I am just reusing what was already used as a placeholder for the LoadableImage (but also adding a spinner, since the scanning may take longer than loading te thumbnail so I thought it would be better to at least show a spinner)

PR #5851, opened 2026-07-08
Media Preview content scanning

Velin92:
Allows for scanning and presenting the result of the scan for medias also in the media preview carousel. I also added a way for the error state to also be swiped across the screen so that it feels natural when swiping between items that belong to QL, if one of the items has an unsafe error. (Otherwise it's hard to tell that the swipe is actually happening)

### DIFF
--- a/ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
+++ b/ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
@@ -158,6 +158,9 @@ enum TimelineMediaPreviewItem: Equatable {
         
         var downloadError: Error?
         
+        /// When set, the media failed content scanning and must not be downloaded or previewed.
+        var contentScanningFailure: ContentScanningFailure?
+        
         init(timelineItem: EventBasedMessageTimelineItemProtocol) {
             self.timelineItem = timelineItem
         }
--- a/ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
+++ b/ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
@@ -144,7 +144,29 @@ class TimelineMediaPreviewDataSource: NSObject, QLPreviewControllerDataSource {
 
 enum TimelineMediaPreviewItem: Equatable {
     case media(Media)
+    case contentScan(Scan)
     case loading(Loading)
+    
+    /// The media item this preview is for, regardless of its content scanning state.
+    var mediaItem: Media? {
+        switch self {
+        case .media(let mediaItem): mediaItem
+        case .contentScan(let scan): scan.media
+        case .loading: nil
+        }
+    }
+    
+    /// A media item that is being processed by the content scanner, or that failed a scan
+    /// and therefore must not be downloaded or previewed.
+    struct Scan: Equatable {
+        enum State: Equatable {
+            case scanning
+            case failure(ContentScanningFailure)
+        }
+        
+        let media: Media
+        let state: State
+    }
     
     /// Wraps a media file and title to be previewed with QuickLook.
     @Observable class Media: NSObject, QLPreviewItem, Identifiable {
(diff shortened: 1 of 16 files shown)


ElementX/Sources/Screens/FilePreviewScreen/View/TimelineMediaContentScanningFailureView.swift
```diff
@@ -0,0 +1,61 @@
...
+//
+
+import Compound
+import SwiftUI
+
+/// Shown fullscreen in place of the media preview when the media failed content scanning.
+struct TimelineMediaContentScanningFailureView: View {
+    let failure: ContentScanningFailure
+    
+    var body: some View {
```
pixlwave, line 15: We can use `TitleAndIcon` here.

ElementX/Sources/Screens/FilePreviewScreen/View/TimelineMediaPreviewController.swift
```diff
@@ -175,6 +185,36 @@ class TimelineMediaPreviewController: QLPreviewController {
         }
     }
     
+    /// Makes the centered overlay (scan failure, download error/indicator) track the page scroll view
+    /// when swiping between items, as it would otherwise float statically above the moving pages.
+    private func observePageScrollViewIfNeeded() {
+        guard pageScrollViewObservation == nil, let pageScrollView else { return }
+        
+        pageScrollViewObservation = pageScrollView.observe(\.contentOffset) { [weak self] _, _ in
```
pixlwave: Should we use `pageScrollView.publisher(for: \.contentOffset)` here instead of old-school KVO?

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
```diff
@@ -158,6 +158,12 @@ enum TimelineMediaPreviewItem: Equatable {
         
         var downloadError: Error?
         
+        /// When set, the media failed content scanning and must not be downloaded or previewed.
+        var contentScanningFailure: ContentScanningFailure?
+        
+        /// Whether the media is currently being scanned by the content scanner.
+        var isBeingScanned = false
+
```
pixlwave: I'm curious. Did you explore whether it worked to use an additional case (or 2) on the enum for the content scanner for these instead of bundling them into the displayable case? It feels like it would be safer.

pixlwave (commented):
Just some initial comments for the moment. I'd be interested to know whether that enum case would work here.

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
Velin92: like using an enum of scanning, failed, unsafe, safe? yeah I guess it's also more consistent with the other side of the implementation

ElementX/Sources/Screens/FilePreviewScreen/View/TimelineMediaContentScanningFailureView.swift
Velin92: However they sizing does differ a bit from figma and the generic title and icon even if the design looks similar,  figma wants LGSemibold, while the default is HeadingSemibold

pixlwave: My bad, it isn't using the component is it 🤦‍♂️

ElementX/Sources/Services/Client/ClientProxy.swift
pixlwave, line 1: Might be worth doing then 😅

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewViewModel.swift
```diff
@@ -109,28 +113,65 @@ class TimelineMediaPreviewViewModel: TimelineMediaPreviewViewModelType {
...
-                    mediaItem.fileHandle = handle
-                    state.previewControllerDriver.send(.itemLoaded(mediaItem.id))
-                case .failure(let error):
-                    MXLog.error("Failed loading media: \(error)")
-                    context.objectWillChange.send() // Manually trigger the SwiftUI view update.
-                    mediaItem.downloadError = error
-                }
+            guard mediaItem.fileHandle == nil, let source = mediaItem.mediaSource else { return }
+            
+            guard await checkSourceSafetyIfNeeded(for: mediaItem, source: source) else { return }
```
pixlwave: Given this returns a `Bool` maybe it should be `checkSourceIsSafeIfNeeded` as it wasn't immediately obvious to me that it was the place we gate on the scan result.

pixlwave (approved):
Looks great, nice job! 👌

PR #5853, opened 2026-07-09
Fix for the media browser showing also images and videos in file mode

Velin92:
- Uses a custom binding to update the view state
- Has the lazy vstack  recompute the content using the screen mode as a parameter instead of reading the state, since lazy stacks could not get the update recomputed properly
- adds test to validate the changes

### DIFF
--- a/ElementX/Sources/Screens/MediaEventsTimelineScreen/MediaEventsTimelineScreenModels.swift
+++ b/ElementX/Sources/Screens/MediaEventsTimelineScreen/MediaEventsTimelineScreenModels.swift
@@ -28,6 +28,9 @@ struct MediaEventsTimelineScreenViewState: BindableState {
     var isBackPaginating = false
     var shouldShowEmptyState = false
     
+    /// The mode that `groups` were built for, so that the layout and its items always change
+    /// together. Updated through the ``MediaEventsTimelineScreenViewAction/changeScreenMode(_:)`` action.
+    var screenMode: MediaEventsTimelineScreenMode
     var groups = [MediaEventsTimelineGroup]()
     
     var activeTimelineContext: TimelineViewModel.Context
@@ -36,13 +39,12 @@ struct MediaEventsTimelineScreenViewState: BindableState {
 }
 
 struct MediaEventsTimelineScreenViewStateBindings {
-    var screenMode: MediaEventsTimelineScreenMode
     var mediaPreviewViewModel: TimelineMediaPreviewViewModel?
     var mediaPreviewSheetViewModel: TimelineMediaPreviewViewModel?
 }
 
 enum MediaEventsTimelineScreenViewAction {
-    case changedScreenMode
+    case changeScreenMode(MediaEventsTimelineScreenMode)
     case oldestItemDidAppear
     case oldestItemDidDisappear
     case tappedItem(item: RoomTimelineItemViewState)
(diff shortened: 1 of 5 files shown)


ElementX/Sources/Screens/MediaEventsTimelineScreen/View/MediaEventsTimelineScreen.swift
```diff
@@ -153,18 +150,22 @@ struct MediaEventsTimelineScreen: View {
         }
     }
     
+    /// The mode check makes sure that items of the other mode never render, e.g. from a stale
+    /// `groups` that was built for the previous mode when switching between the two. It's passed
+    /// in by the caller (which statically knows its own mode) rather than read from the state, as
+    /// reading the state from within a lazy container's rows isn't reliable when the mode changes.
```
pixlwave: I think we can say all of this in 1 short sentence 😅

ElementX/Sources/Screens/MediaEventsTimelineScreen/MediaEventsTimelineScreenModels.swift
```diff
@@ -28,6 +28,9 @@ struct MediaEventsTimelineScreenViewState: BindableState {
     var isBackPaginating = false
     var shouldShowEmptyState = false
     
+    /// The mode that `groups` were built for, so that the layout and its items always change
+    /// together. Updated through the ``MediaEventsTimelineScreenViewAction/changeScreenMode(_:)`` action.
```
pixlwave: Not sure if this one is necessary?

UnitTests/Sources/MediaEventsTimelineScreenViewModelTests.swift
```diff
@@ -0,0 +1,67 @@
...
+        context.send(viewAction: .changeScreenMode(.media))
+        
+        #expect(context.viewState.screenMode == .media)
+        #expect(context.viewState.activeTimelineContext === mediaTimelineViewModel.context)
+        #expect(allItemsMatchScreenMode(.media, groups: context.viewState.groups))
+    }
+    
+    // MARK: - Helpers
+    
+    /// Both timelines are given the same mixed chunk of items so that the mode based filtering is what's under test.
```
pixlwave: Is this meant to be a doc comment?

UnitTests/Sources/MediaEventsTimelineScreenViewModelTests.swift
pixlwave, line 1: Nice!

pixlwave (approved):
🙌

ElementX/Sources/Screens/MediaEventsTimelineScreen/View/MediaEventsTimelineScreen.swift
```diff
@@ -239,7 +240,8 @@ struct MediaEventsTimelineScreen: View {
     }
     
     private var screenModePicker: some View {
-        Picker("", selection: $context.screenMode) {
+        Picker("", selection: Binding(get: { context.viewState.screenMode },
+                                      set: { context.send(viewAction: .changeScreenMode($0)) })) {
```
pixlwave, line 241: FWIW, I'm not really a fan of these because of https://talk.objc.io/episodes/S01E439-optional-bindings

> However, there's a big downside to using manually constructed bindings. By creating a binding from getter and setter closures, we prevent SwiftUI from checking if the binding's value has actually changed and whether or not the view using the binding needs to be rerendered, thereby forcing rerenders with every update.

Maybe there's another solution but I don't think it's the end of the world if not.

Velin92: the other solution was to have the mode both be a binding and a normal viewState var but I did not like the duplication (it allowed to listen and observe one value change before changing the other one, but yeah did not like it)... so I resorted to this

PR #5859, opened 2026-07-10
Display other users' User Status.

pixlwave:
Definitely review this commit by commit. This PR makes the following changes:
- Add a `UserStatus` type and include in on all of the various profile types in the app.
- Show the statuses in all the various places they should be shown (see exactly where by looking at the individual commits).
- Update the layout of `AvatarHeaderView` (the spacing didn't match the Figma).
- Update the layout of the message bubble's sender info (the font and alignment was wrong).
- Set the correct text colour on the `RoomHeaderView`.
- Add an extra pt padding in the composer suggestions to closer match Figma and allow room for the emoji.

The consequence of this is that a *lot* of snapshots were updated.

I didn't realise this was going to be so big. I think it's mainly the snapshots contributing to the size, but please do shout if so and I'll figure out a good place to split it in 2.


ElementX/Sources/Other/SwiftUI/Views/AvatarHeaderView.swift
```diff
@@ -133,6 +137,84 @@ struct AvatarHeaderView<Footer: View>: View {
...
+                    .multilineTextAlignment(.center)
+                    .textSelection(.enabled)
+                
+                if let displayedStatus = status?.displayed {
+                    HStack(spacing: 6) {
+                        Text(String(displayedStatus.emoji))
+                            .foregroundStyle(.compound.textPrimary)
+                        Text(displayedStatus.text)
+                            .foregroundStyle(.compound.textSecondary)
+                    }
```
Velin92, line 157: I wonder... but is just. minor thing, since both are actually text, maybe we could use some unicode character to create a fixed space that can in theory scalr with dynamic sizing? Or we are fine to keep 6 as the padding between the two?

ElementX/Sources/Screens/RoomScreen/ComposerToolbar/View/MentionSuggestionItemView.swift
```diff
@@ -6,26 +6,37 @@
...
+import Compound
 import SwiftUI
 
 struct MentionSuggestionItemView: View {
     let mediaProvider: MediaProviderProtocol?
     let item: SuggestionItem
     
     var body: some View {
-        HStack(alignment: .center, spacing: 16) {
+        HStack(alignment: .center, spacing: 12) {
```
Velin92, line 17: why has this been reduced? new design?

ElementX/Sources/Screens/Timeline/View/Style/TimelineItemBubbledStylerView.swift
```diff
@@ -54,7 +54,7 @@ struct TimelineItemBubbledStylerView<Content: View>: View {
     
     var body: some View {
         ZStack(alignment: .trailingFirstTextBaseline) {
-            VStack(alignment: alignment, spacing: -12) {
+            VStack(alignment: alignment, spacing: -8) {
```
Velin92, line 57: sae here why the reduction?

pixlwave: Yep, I don't know how long this has been like this in Figma (I suspect quite a long time), but the font is now larger and the overlap reduced:

https://www.figma.com/design/ux3tYoZV9WghC7hHT9Fhk0/Compound-iOS-Components?node-id=1782-15870&m=dev

Without those changes the emoji was clipping into the bubble.

ElementX/Sources/Screens/RoomScreen/ComposerToolbar/View/MentionSuggestionItemView.swift
pixlwave: Yep, I can't find the master component, but the user status design is here:

https://www.figma.com/design/IjNuVxaLRhe2MotMWJ9EtG/ER-237--User-status?node-id=326-35493&m=dev

ElementX/Sources/Services/Users/UserProfile.swift
```diff
@@ -55,6 +61,53 @@ nonisolated struct UserProfile: Hashable, Identifiable {
     }
 }
 
+nonisolated struct UserStatus: Hashable {
+    /// The status manually set by the user (if any).
+    let userSet: UserSet?
```
Velin92, line 66: UserSet is from the sdk? I don't really like this namesince Set (with capital S) is a container of unique items... it feels like a mistyped set of users. Couldn't be userDefined or something similar?

ElementX/Sources/Other/SwiftUI/Views/AvatarHeaderView.swift
pixlwave: I did wonder about this, but then we wouldn't get automatic RTL support and so it didn't seem worth it to me. We can always switch to a `ScaledMetric` here if it is ever flagged as an issue.

PreviewTests/Sources/__Snapshots__/PreviewTests/imageRoomTimelineView.iPad-pseudo-0.png
Velin92, line 1: I know this is not related but damn it's a bit ugly is it not?

PreviewTests/Sources/__Snapshots__/PreviewTests/roomDetailsScreen.DM-Room-iPad-en-GB.png
Velin92, line 1: is this slightly smaller?

PreviewTests/Sources/__Snapshots__/PreviewTests/roomDetailsScreen.DM-Room-iPhone-en-GB.png
Velin92, line 1: did something change in the spacing maybe ?

Velin92 (approved):
LGTM just few comments

ElementX/Sources/Other/SwiftUI/Views/AvatarHeaderView.swift
Velin92: dw about this just realised that the fact they are separate is indeed a good thing, since joined strings can always be replicated by users, while in this way they are indeed separate UI elements.

ElementX/Sources/Services/Users/UserProfile.swift
pixlwave: It's from Android 😄 I can't think of anything better though:
- `status` would mean you do `user.status.status` which is horrible
- `custom` would be confusing because you can select a preset status or input a custom status and both result in this `userSet` value being updated.
- … I have no other ideas

Open to suggestions.

PreviewTests/Sources/__Snapshots__/PreviewTests/roomDetailsScreen.DM-Room-iPad-en-GB.png
pixlwave: Yeah the spacing has reduced. But urgh I just noticed. That the spacing between the header and the buttons has reduced, *but* the buttons also have an additional 8pt vertical padding on them which means that one is now wrong. Will fix.

PreviewTests/Sources/__Snapshots__/PreviewTests/imageRoomTimelineView.iPad-pseudo-0.png
pixlwave: 🤷‍♂️

PreviewTests/Sources/__Snapshots__/PreviewTests/roomDetailsScreen.DM-Room-iPad-en-GB.png
pixlwave: https://github.com/element-hq/element-x-ios/pull/5859/commits/534be24bf4dcb4105c752e672a1a1d5af20f601d

PreviewTests/Sources/__Snapshots__/PreviewTests/roomDetailsScreen.DM-Room-iPhone-en-GB.png
pixlwave: As above

PreviewTests/Sources/__Snapshots__/PreviewTests/imageRoomTimelineView.iPad-pseudo-0.png
pixlwave: Well let's see what the reaction to it is like on Nightly.

ElementX/Sources/Services/Users/UserProfile.swift
pixlwave: Going to merge this for now, we can rename this in a follow-up.

PR #5860, opened 2026-07-10
Use dark theme for Element Call

toger5:
This PR updates EC to always use the dark theme. Additionally it will configure the EC Screens preferred theme to Dark so that the rest of the app also fits to the dark webview.
(The webview will not automatically use the system preferred theme. That is why we add the url param `theme='dark'`)
### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
+++ b/ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
@@ -26,9 +26,9 @@ struct CallScreen: View {
                 .toolbar { toolbar }
         }
         .alert(item: $context.alertInfo)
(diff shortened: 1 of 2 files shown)


toger5:
I am worried this is a trivial fix so I will stop spending time on this: `An additional repository requires authorization.`
and ask for help. What is breaking CI?

pixlwave:
I've updated the PR to set the colour scheme in the environment, otherwise the entire app is affected by the preferred colour scheme (or at least the root navigation stack is anyway), so when you put the call into a PiP you're presented with dark mode rather than whatever your system is set to.

pixlwave (approved):
So long light mode 😢

PR #5861, opened 2026-07-10
Use deferFulfillment in flow coordinator tests

Velin92:
Specifically in ChatTabsFlow and RoomFlow Coordinator tests

PR #5869, opened 2026-07-13
Fix for flaky tests

Velin92:
## Root causes & fixes

| Tests | Root cause → fix |
|---|---|
| `LoggingTests` (all 7 tests failing together in recent CI runs) | Per-test log directories in the app group container were never cleaned up. CI retries (`xcodebuild -retry-tests-on-failure`) re-run the suite on the same simulator, so every retry found stale log files and failed the `Tracing.logFiles.isEmpty` precondition — turning any unrelated flake into an unrecoverable job. → Delete the test directory in `setupTest()` before each test. |
| `MediaUploadingPreprocessorTests.landscapeMovVideoProcessing` / `portraitMp4VideoProcessing` | Two real AVFoundation transcodes per test exceed the 60s test execution allowance on a busy CI runner. → Split the optimised-upload halves into their own tests and raise the test plan's `defaultTestExecutionTimeAllowance` from 60s to 120s. |
| `MediaUploadPreviewScreenViewModelTests` video upload tests | The view model transcodes the video for real, but the test waited for `.dismiss` with the default 10s `deferFulfillment` timeout. → 60s timeout for the video upload tests. |
| `TimelineMediaPreviewViewModelTests.redactConfirmation` | The redaction runs in an unstructured `Task` racing the synchronously-sent dismiss action, so `redactCalled` could be asserted before the task ran. → Wait for the mock's redact call via `waitForConfirmation`. |
| `RoomDetailsScreenViewModelTests` ignore/unignore, `SpaceScreenViewModelTests` joining tests, `RoomNotificationSettingsScreenViewModelTests.deleteCustomSettingTappedFailure` | Transient states (`isProcessingIgnoreRequest`, `joiningRoomIDs`, `deletingCustomSetting`) are set and cleared around an instantly-returning mock, and the `Observations`-based `observe()` sequence samples values rather than queueing them — so the transition was often never seen. → Assert the in-flight state inside the mock closure (where it's guaranteed to be set) and only wait for durable outcomes from the outside. |
| `SpaceScreenViewModelTests.singlePagination` | Asserted view state right after the proxy publisher fired, racing the view model's own subscription; rooms and pagination state also arrive on separate publishers. → Wait on the view state itself for both the paginating and end-reached states, and for the rooms. |
| `ComposerToolbarViewModelTests.multipleViolation` | The view model consumes identity status change deltas via `AsyncPublisher`, which drops values emitted while the previous ones are still being processed — a resolution delta could be lost, leaving the composer disabled forever. → Synchronise each delta on the view state emission its processing produces. (The underlying delta-dropping is a real product bug, to be fixed separately.) |
| `UserSessionFlowCoordinatorTests` | The state machine's state changes before the coordinators update their stacks — the same race #5861 fixed for the chats tab / room flow coordinators. → Apply the same pattern: after the state, wait for the navigation side effect via `observe` + `deferFulfillment`, replacing the 100ms delay. Shared the coordinator ID-projection extensions in `TestUtilities/Extensions/` instead of duplicating them per file. |
| `ElementCallServiceTests` timer tests (the most frequent flake in recent CI runs) | Two mechanisms: (1) `TestClock.advance` raced the unstructured timer task — if it advanced before the task registered its sleep, the timer never fired; (2) `TestClock` internally awaits background-priority tasks, which get starved on loaded runners, so a single advance could exceed the execution time allowance. → Wait for the sleep to actually be scheduled (`checkSuspension()`) before advancing, and await the unanswered-call report through a stream instead of asserting right after the advance. |


UnitTests/Sources/TestUtilities/Extensions/NavigationCoordinators.swift
pixlwave, line 1: Maybe we should just have a single `NavigationCoordinators` file for these, similar to what we have in the main app (except I think it make more sense to put them *all* in the tests one rather than just split & stack like the main app has).

UnitTests/Sources/LoggingTests.swift
```diff
@@ -331,6 +331,12 @@ final class LoggingTests {
     /// to start with a fresh state (as calling ``Tracing.deleteLogFiles`` would trigger the bug).
     private func setupTest(name: String = #function, redirectTracingFileWriter: Bool = true) throws {
         let testDirectory = URL.appGroupLogsDirectory.appending(component: name, directoryHint: .isDirectory)
+        
+        // CI re-runs failing tests on the same simulator (xcodebuild -retry-tests-on-failure),
+        // so a previous attempt may have left log files behind which would break the assertion
+        // below. No file writer points at the directory at this stage, so deleting it is safe.
+        try? FileManager.default.removeItem(at: testDirectory)
```
pixlwave: Riiiight, so they likely pass the first time but then because of the Swift Testing's retry behaviour, the entire suite is re-run meaning these then fail the next 3 times 🤦‍♂️

That actually makes sense. Maybe we can shorten that comment however.

UnitTests/Sources/MediaUploadingPreprocessorTests.swift
```diff
@@ -74,8 +74,14 @@ final class MediaUploadingPreprocessorTests {
         #expect(isEqual(videoInfo.thumbnailInfo?.size ?? 0, 183_093, within: 100))
         #expect(videoInfo.thumbnailInfo?.width == 800)
         #expect(videoInfo.thumbnailInfo?.height == 450)
+    }
+    
+    // Split from landscapeMovVideoProcessing: each transcode is slow enough on a busy CI runner
+    // that running both in a single test can exceed the execution time allowance.
```
pixlwave: I don't think we really need this comment.

UnitTests/Sources/MediaUploadingPreprocessorTests.swift
```diff
@@ -129,8 +135,14 @@ final class MediaUploadingPreprocessorTests {
         #expect(isEqual(videoInfo.thumbnailInfo?.size ?? 0, 40976, within: 100))
         #expect(videoInfo.thumbnailInfo?.width == 337)
         #expect(videoInfo.thumbnailInfo?.height == 600)
+    }
+    
+    // Split from portraitMp4VideoProcessing: each transcode is slow enough on a busy CI runner
+    // that running both in a single test can exceed the execution time allowance.
```
pixlwave: Nor this.

UnitTests/Sources/MediaUploadingPreprocessorTests.swift
```diff
@@ -74,8 +74,14 @@ final class MediaUploadingPreprocessorTests {
         #expect(isEqual(videoInfo.thumbnailInfo?.size ?? 0, 183_093, within: 100))
         #expect(videoInfo.thumbnailInfo?.width == 800)
         #expect(videoInfo.thumbnailInfo?.height == 450)
+    }
+    
+    // Split from landscapeMovVideoProcessing: each transcode is slow enough on a busy CI runner
+    // that running both in a single test can exceed the execution time allowance.
+    @Test
+    func landscapeMovVideoProcessingOptimized() async throws {
```
pixlwave: Probably slightly clearer to understand: `landscapeMovVideoOptimizedProcessing`

UnitTests/Sources/MediaUploadingPreprocessorTests.swift
```diff
@@ -129,8 +135,14 @@ final class MediaUploadingPreprocessorTests {
         #expect(isEqual(videoInfo.thumbnailInfo?.size ?? 0, 40976, within: 100))
         #expect(videoInfo.thumbnailInfo?.width == 337)
         #expect(videoInfo.thumbnailInfo?.height == 600)
+    }
+    
+    // Split from portraitMp4VideoProcessing: each transcode is slow enough on a busy CI runner
+    // that running both in a single test can exceed the execution time allowance.
+    @Test
+    func portraitMp4VideoProcessingOptimized() async throws {
```
pixlwave: and `portraitMp4VideoOptimizedProcessing`

UnitTests/Sources/MediaUploadPreviewScreenViewModelTests.swift
pixlwave, line 1: 👌

UnitTests/Sources/ElementCallServiceTests.swift
```diff
@@ -11,6 +11,8 @@ import Clocks
 import PushKit
 import Testing
 
+// The TestClock advances by yielding to background tasks which can be starved on a busy CI runner.
```
pixlwave, line 14: I'm incredibly skeptical about this… We run all the tests in serial so there's no way there's other things that are taking 2 minutes worth of priority.

UnitTests/SupportingFiles/UnitTests.xctestplan
```diff
@@ -9,7 +9,7 @@
     }
   ],
   "defaultOptions" : {
-    "defaultTestExecutionTimeAllowance" : 120,
+    "defaultTestExecutionTimeAllowance" : 60,
```
pixlwave, line 12: We don't need this change. The default should be 1 minute and then we can use that `.timeLimit` to increase specific tests on a case-by-case basis.

UnitTests/Sources/ComposerToolbarViewModelTests.swift
pixlwave, line 1: This one seems wrong to me. It has changed the meaning of the test by removing that `mockSubject.send`. I think we should try this:

```
send(verificationViolation)
assertFullfillment(canSend == false)

send(pinned)
assertFailure(canSend == true)

send(pinned)
assertFullfillment(canSend == true)
```

UnitTests/Sources/ElementCallServiceTests.swift
```diff
@@ -235,6 +242,17 @@ final class ElementCallServiceTests {
         #expect(service.ongoingCallRoomIDPublisher.value == "!room:example.com")
     }
     
+    /// Waits until the service's unanswered-call timer has scheduled its sleep on the test clock.
+    ///
+    /// The timer runs in an unstructured task, so without this the clock could be advanced
+    /// before the task even started sleeping, leaving it asleep forever and the test hanging.
+    /// `checkSuspension()` throws when a sleep is scheduled, which is the state to wait for.
+    private func waitForScheduledSleep() async {
```
pixlwave, line 247: Have you tried these tests with *just* this fix? If it's right then this seems like the root cause of the failures we've been seeing.

UnitTests/Sources/RoomDetailsScreenViewModelTests.swift
pixlwave, line 1: Wait all of these have `transitionValues: [false, true, false]`.

Isn't the fix exactly the same as we did for the notification tests and do:

```
#expect(false)

let deferred = transitionValues: [true, false]
send(action)
deferred.fulfill()
```

UnitTests/Sources/RoomNotificationSettingsScreenViewModelTests.swift
pixlwave, line 1: Same here, the format is fine we just can't assert the starting transition value.

UnitTests/Sources/TimelineMediaPreviewViewModelTests.swift
pixlwave: Should we remove this one then given it is confirmed above.

UnitTests/Sources/UserSessionFlowCoordinatorTests.swift
```diff
@@ -148,6 +150,10 @@ struct UserSessionFlowCoordinatorTests {
         try await process(route: .share(sharePayload),
                           expectedChatsState: .roomList(detailState: .room(roomID: "2")))
         
+        let splitCoordinator = try #require(chatsSplitCoordinator)
+        let deferredSheet = deferFulfillment(splitCoordinator.observe(\.sheetCoordinatorID)) { $0 != nil }
+        try await deferredSheet.fulfill()
+        
         #expect(detailNavigationStack?.rootCoordinator is RoomScreenCoordinator)
         #expect(tabCoordinator?.sheetCoordinator == nil)
         #expect((chatsSplitCoordinator?.sheetCoordinator as? NavigationStackCoordinator)?.rootCoordinator is MediaUploadPreviewScreenCoordinator)
```
pixlwave:
```suggestion
        #expect((splitCoordinator.sheetCoordinator as? NavigationStackCoordinator)?.rootCoordinator is MediaUploadPreviewScreenCoordinator)
```

UnitTests/Sources/SpaceScreenViewModelTests.swift
```diff
@@ -48,22 +48,23 @@ struct SpaceScreenViewModelTests {
         #expect(!response.isEmpty, "There should be some test rooms.")
         
         // When the pagination is triggered.
-        var deferred = deferFulfillment(spaceRoomListProxy.paginationStatePublisher) { $0 == .loading }
+        // Wait on the view state (not the proxy's publisher) as the view model updates it asynchronously.
+        let deferredPaginating = deferFulfillment(context.observe(\.viewState.paginationState)) { $0 == .paginating }
+        let deferredEndReached = deferFulfillment(context.observe(\.viewState.paginationState)) { $0 == .endReached }
```
pixlwave: This first test seems over-complicated now. I think it can all be achieved using `transitionValues` and a `#expect(spaceRoomListProxy.paginateCallsCount == 1)` at the end. That would prove the same thing.

pixlwave (commented):
Overall this looks great, but I do have plenty of little thoughts:

UnitTests/Sources/ElementCallServiceTests.swift
Velin92: quoting Claude here, and also had it perform a test, where we removed the timelimit, and experienced again the failures:

>  it's not our tests competing, it's the rest of the machine. TestClock.advance internally calls Task.megaYield(), which awaits 20 Task.detached(priority: .background) hops per loop iteration (source (https://github.com/pointfreeco/swift-concurrency-extras/blob/main/Sources/ConcurrencyExtras/Task.swift)). Background QoS is throttled machine-wide, so on a loaded runner those hops stall regardless of our suite running serially — and the CI boxes are definitely loaded.

> Those are the same three tests with the same failure message we see on CI, and it happens even with the waitForScheduledSleep() fix in place — that fix removes the race (wrong results/hangs), while this covers the raw slowness of advancing the clock under load.
> 
> So I've kept the 2-minute limit but moved it from the suite onto just the three clock-advancing tests, as you suggested for the test plan — everything else in the suite stays on the 60s default.

UnitTests/Sources/ElementCallServiceTests.swift
Velin92: [check my answer to the comment above](https://github.com/element-hq/element-x-ios/pull/5869#discussion_r3579289219)

UnitTests/Sources/ComposerToolbarViewModelTests.swift
```diff
@@ -760,13 +756,19 @@ final class ComposerToolbarViewModelTests {
                                              analyticsService: AnalyticsServiceMock(.init()),
                                              composerDraftService: draftServiceMock)
         
+        // The consuming AsyncPublisher drops changes sent while earlier ones are still being
+        // processed, so only send after observing the previous change's view state update.
```
pixlwave: Looking at the diff it could equally be that it doesn't publish changes if nothing changed? Maybe we should remove the comment, it probably doesn't add anything useful here. But not massively important.

pixlwave (approved):
Very nice 👌

PR #5870, opened 2026-07-13
Fix CurrentValuePublisher.map dropping updates when the mapped publisher is chained inline

ara4n:
This fixes a bug which the E2E test rig showed up, where EXI doesn't show new knocks if you already have the room open.

Fable-generated, with its explanation below:

> The eager implementation bridged through a second CurrentValueSubject fed by a sink whose AnyCancellable was stored on the returned struct. Combine consumes publisher structs when building a subscription, so chaining the mapped publisher inline (e.g. `publisher.map { … }.removeDuplicates(…) .sink { … }`, as RoomScreenViewModel and RoomDetailsScreenViewModel do for knock requests) dropped the only copy holding that cancellable: the feeding sink was cancelled and subscribers only ever received the initial value.
>
> In practice this meant a room admin viewing a knockable room never saw the "wants to join" banner (or a live "Requests to join" badge) for a knock arriving while the room was open — it only appeared when the knock predated opening the room. Caught by a live cross-client knocking test; the double-execution of the transform at subscription time (eager transform + current-value replay) was the giveaway.
>
>Make map lazy instead: keep the transformed AnyPublisher and a value provider, so subscribers attach straight through to the source subject (which replays its current value itself) and no keep-alive is needed. Adds a regression test chaining a mapped publisher inline.

### DIFF
--- a/ElementX/Sources/Other/CurrentValuePublisher.swift
+++ b/ElementX/Sources/Other/CurrentValuePublisher.swift
@@ -13,33 +13,44 @@ import Combine
 ///
 /// `CurrentValueSubject` is documented as thread-safe but is not formally `Sendable`, hence `@unchecked`.
 nonisolated struct CurrentValuePublisher<Output, Failure: Error>: Publisher, @unchecked Sendable {
-    private let subject: CurrentValueSubject<Output, Failure>
-    /// Retains the upstream subscription feeding ``subject`` when this publisher is derived via ``map(_:)``.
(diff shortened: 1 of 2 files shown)


Velin92:
Updated formatting, and rewrote the comments because AI explaination was just so convoluted and long, when in reality it was way simpler. But great catch!

ara4n:
woo, thanks for merging - sorry for the AI spew, but glad it was a legit catch.

PR #5871, opened 2026-07-14
Update the SDK and handle API breaks.

pixlwave:
Also adds a couple of new methods to ClientProxy for setting/removing the user status.

The SDK is still being built using https://github.com/matrix-org/matrix-rust-sdk/pull/6714


Velin92 (approved):

ElementX/Sources/Services/Users/UserProfile.swift
```diff
@@ -104,7 +104,34 @@ nonisolated struct UserStatus: Hashable {
...
+        call = nil
+    }
+    
+    init(rustStatus: MatrixRustSDK.UserStatus?, rustCall: MatrixRustSDK.UserCall?) {
+        userSet = rustStatus.map(UserSet.init)
+        call = rustCall.map(Call.init)
+    }
+}
+
+nonisolated extension UserStatus.UserSet {
```
Velin92: I would just call it Status as a internal type of UserStatus, and just call the variable storing simply as status... and add a doc explaining that it's set by the user. There is no need for the fact of the value being set by the user to be so explicitly documented that needs to be in the name of both the type and the variable

pixlwave: We discussed this and landed on raw status to match the raw name that we already have for display names.

PR #5872, opened 2026-07-14
Implements push notifications for LLS via MSC4505

ara4n:
Will requires SDK >= 26.07.11 to build and pass CI, once released.

### DIFF
--- a/NSE/Sources/NotificationContentBuilder.swift
+++ b/NSE/Sources/NotificationContentBuilder.swift
@@ -59,28 +59,38 @@ nonisolated struct NotificationContentBuilder {
                                  notificationItem: notificationItem,
                                  mediaProvider: mediaProvider)
         case .timeline(let event):
-            guard case let .messageLike(messageContent) = try? event.content() else {
-                processEmpty(&notificationContent)
-                return
-            }
-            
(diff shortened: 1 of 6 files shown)


pixlwave:
@ara4n This one can be rebased to use an SDK built yesterday.

Also please can you make sure Claude runs `swiftformat` before opening a PR so that one of us doesn't need to do it to get the changes into a mergeable shape :)

NSE/Sources/NotificationContentBuilder.swift
```diff
@@ -59,28 +59,38 @@ nonisolated struct NotificationContentBuilder {
...
+                    notificationContent.body = L10n.commonUnsupportedCall
+                case .rtcNotification:
+                    notificationContent.body = L10n.notificationIncomingCall
+                default:
+                    processEmpty(&notificationContent)
+                }
+            case .state(let stateContent):
+                switch stateContent {
+                case .beaconInfo:
+                    await processMessageLike(notificationContent: &notificationContent,
```
pixlwave: It feels a bit weird to call `processMessageLike` when we're not `.messageLike` here maybe we could tidy up the naming a bit.

Velin92 (approved):

PR #5873, opened 2026-07-14
Constrained tests tool implementation + skill

Velin92:
Are you tired of writing a huge suite of tests (or have your AI write them), and then just find out they work fine locally, but miserably fail on CI? With this tool you can "kinda" simulate the CPU constraints of CI, by essentially running the tests, alongside some awaiting slow processes to create a starved CPU environments. It's not intended to be used always, like the skill suggested, should be tried once, if you have build a very heavy dependant suite of tests on awaiting a lot of long sleeps.

### DIFF
--- a/.agents/skills/constrained-tests/SKILL.md
+++ b/.agents/skills/constrained-tests/SKILL.md
@@ -1,0 +1,34 @@
+---
+name: constrained-tests
+description: Run unit tests on saturated CPU to reproduce CI-style flakiness locally. Suggest when new/reworked suite lean heavy on awaits (deferFulfillment, waitForConfirmation, TestClock, timeouts) or test fail only on CI. Not for every change.
+---
+
+# Constrained test runs
+
+CI runners shared + slow. Background tasks starve, awaits stretch, scheduler races flake there, pass locally. `--constrained` saturate CPU with hog tasks — same conditions, local.
+
+## When
+
+- PR add/rework suite heavy on timing-sensitive async: `deferFulfillment`, `waitForConfirmation`, `TestClock`, `transitionValues`, timeouts.
+- Test fail on CI, pass locally.
+- Not routine changes, not sync suites — slow, no gain.
+
+## How
+
+Full unit test workflow (what CI run):
+
+```bash
+swift run tools ci unit-tests --constrained
+```
+
+One suite, iterate on new tests (prefer while developing):
+
+```bash
+swift run tools ci run-tests --scheme UnitTests --test-name MyNewTests --constrained
+```
+
+Both take `--constrained-task-count <n>` (default 20 ≈ 2× cores). Suite survive ~10 constrained iterations → good shape.
+
+## Failures mean
+
+Failure under constraint = real bug in test (sometimes code under test): assertion race async work, transient state observation missed, timeout too tight. Fix synchronisation, no blind timeout bump. Helpers in `UnitTests/Sources/TestUtilities/` (`deferFulfillment`, `deferFailure`, `waitForConfirmation`). Wait on durable state, not transient.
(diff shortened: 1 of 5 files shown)


Tools/Sources/Commands/CI/RunTests.swift
```diff
@@ -39,6 +39,12 @@ struct RunTests: AsyncParsableCommand {
     @Option(help: "Only run a specific test (format: 'ClassName/testName').")
     var testName: String?
     
+    @Flag(help: "Run the tests on a saturated CPU to reproduce the flakiness of a busy CI runner.")
+    var constrained = false
+    
+    @Option(help: "The number of CPU hogging tasks to run when the tests are constrained.")
+    var constrainedTaskCount = 20
```
pixlwave: We can probably hardcode this using `ProcessInfo.processInfo.processorCount * 2` so it is automatic on any machine. Also should it be `- 1` so that there's a single "free" core for the tests like a 1 CPU test runner?

Edit: Actually presumably it could be hard-coded into `CPUConstraint`.

Edit 2: Although 19 busy cores doesn't make sense as Apple Silicon doesn't do Hyperthreading/SMT. But we should still use the processorCount in some way to avoid needing to think about this ourselves.

Tools/Sources/Commands/CI/CPUConstraint.swift
```diff
@@ -0,0 +1,39 @@
...
+
+/// Saturates the CPU with busy processes for the duration of a test run, approximating
+/// a loaded CI runner where timing-sensitive tests are starved of processor time.
+final class CPUConstraint {
+    private var processes = [Process]()
+    
+    /// Starts the given number of CPU hogging tasks. 2× the machine's cores is
+    /// enough to keep the scheduler permanently oversubscribed.
+    func start(taskCount: Int) {
+        logger.info("\n🐌 Constraining the CPU with \(taskCount) tasks…\n")
```
pixlwave, line 12: I love these logs 🐌

Tools/Sources/Commands/CI/UnitTests.swift
```diff
@@ -19,6 +25,13 @@ struct UnitTests: AsyncParsableCommand {
         // Run unit tests
         do {
             logger.info("\n🧪 Running unit tests…\n")
+            
+            let cpuConstraint = CPUConstraint()
+            if constrained {
+                cpuConstraint.start(taskCount: constrainedTaskCount)
+            }
+            defer { cpuConstraint.stop() }
+
```
pixlwave: Errr, this one is wrong isn't it? We can just pass the flag(s) to the `RunTests` below and not duplicate this code.

AGENTS.md
```diff
@@ -81,6 +81,7 @@ Initial setup: `swift run tools setup-project`
 
 CI test commands:
 - Unit tests: `swift run tools ci unit-tests`
+- Slow-CI simulation: add `--constrained` (hog count: `--constrained-task-count`, default 20). Also on `ci run-tests` for one suite. New async-heavy suite? Use `constrained-tests` skill — not for every change.
```
pixlwave: Is this actually necessary. If anything just make sure the skill's description says "Slow-CI" somewhere and that's a good as this line without adding more tokens to the context.

.agents/skills/constrained-tests/SKILL.md
```diff
@@ -0,0 +1,34 @@
...
+swift run tools ci unit-tests --constrained
+```
+
+One suite, iterate on new tests (prefer while developing):
+
+```bash
+swift run tools ci run-tests --scheme UnitTests --test-name MyNewTests --constrained
+```
+
+Both take `--constrained-task-count <n>` (default 20 ≈ 2× cores). Suite survive ~10 constrained iterations → good shape.
```
pixlwave: If we hard code it, then we can presumably remove this line too.

pixlwave (approved):
Very cool 👏

PR #5888, opened 2026-07-15
Add the UI for your own user status in the Settings screen.

pixlwave:
This PR makes 2 changes:
- Shows your own status emoji next to your name in the top row of the settings screen when set (doesn't work in production, we only request the display name and avatar URLs for the profile).
- Adds a new row for setting/clearing your own status (hidden behind the feature flag, only setting will work it won't ever show the status).


Velin92 (approved):

ElementX/Sources/Screens/Settings/SettingsScreen/View/SettingsScreenUserStatusRow.swift
```diff
@@ -0,0 +1,116 @@
...
+                        }
+                        .contentShape(Rectangle())
+                    }
+                    .buttonStyle(.plain)
+                    
+                    Button(L10n.actionClear) { action(.set(nil)) }
+                        .buttonStyle(.compound(.textLink))
+                }
+                .padding(ListRowPadding.insets)
+            })
```
Velin92, line 79: we probably need an a11y label for the action and an a11y value to tell the user the current status

ElementX/Sources/Screens/Settings/SettingsScreen/View/SettingsScreenUserStatusRow.swift
```diff
@@ -0,0 +1,116 @@
...
+    
+    var body: some View {
+        switch mode {
+        case .pick:
+            ListRow(label: .default(title: L10n.screenSettingsUserStatusPlaceholder, icon: \.reaction),
+                    kind: .button { action(.pickStatus) })
+        case .custom:
+            ListRow(kind: .custom {
+                HStack(spacing: ListRowPadding.labelIconSpacing) {
+                    Button { } label: {
```
Velin92: why use a button with no action here? Also needs some a11y magic

Velin92 (commented):
LGTM, just a few a11y concerns

ElementX/Sources/Screens/Settings/SettingsScreen/View/SettingsScreenUserStatusRow.swift
pixlwave: Ha no! This is meant to present the emoji picker. Good spot, thanks!

pixlwave: Have put in a placeholder action for now, will hook this up in the next PR.

PR #5890, opened 2026-07-15
Several small, independent accessibility fixes (VoiceOver labels/traits, contrast)

unclejay80:
## Content

This PR bundles eight small, independent accessibility fixes that were originally submitted as separate PRs and closed by a maintainer with a request to combine them into a single PR since they're all accessibility-related. Each fix touches a different screen or component and can be reviewed independently — they're combined here only because of that request, not because they share code:

1. **Onboarding start screen title as a header** (`AuthenticationStartScreen.swift`) — adds `.accessibilityAddTraits(.isHeader)` to the screen title so it's reachable via VoiceOver's heading rotor.
2. **Poll winner icon + vote count combined** (`PollOptionView.swift`) — adds `.accessibilityElement(children: .combine)` to the winning-option `HStack` so the trophy icon and vote count are announced together instead of as two separate elements.
3. **Poll option elements combined, progress bar hidden** (`PollOptionView.swift`) — combines each option's icon/text/vote-count into one VoiceOver element and hides the purely-visual `PollProgressView` with `.accessibilityHidden(true)`.
4. **Formatting toolbar buttons exposed as toggles** (`FormattingToolbar.swift`) — adds `.isToggle` (plus `.isSelected` when active) to the accessibility traits so VoiceOver announces formatting buttons as toggles with their current state, not plain buttons.
5. **Avatar edit button labelled** (`UserDetailsEditScreen.swift`) — reuses the existing `L10n.a11yEditAvatar` string (already used on the create-room screen) as the avatar button's `accessibilityLabel`.
6. **Text field border contrast** (`CompoundTextFieldStyles.swift`) — gives text fields a visible border in their resting (non-focused) state instead of the border only becoming visible once focused or erroring.
7. **Chat list reordering announced** (`HomeScreenContent.swift`) — adds `.accessibilityAddTraits(.updatesFrequently)` to the room list so VoiceOver knows to re-announce position changes when a room moves (e.g. after a new message) instead of silently reordering.
8. **Pinned items banner explicit label/value** (`PinnedItemsBannerView.swift`) — gives the "view all" button an explicit `accessibilityLabel` so it isn't silently unlabelled while loading (its visible text is emptied during that state), and gives the main pinned-item button an explicit label (the pinned message) and value (its position) instead of relying on `.accessibilityElement(children: .contain)` to merge its sub-views.

## Motivation and context

Each of these was flagged independently during an accessibility review — a missing VoiceOver label/trait or insufficient resting-state contrast. None of the fixes are large, and each is scoped to a single view or style file, but per maintainer feedback on the original individual PRs they should land as a single PR rather than as many.

## Testing

Each item was verified independently with VoiceOver enabled (or, for the contrast fix, with an automated contrast checker), in both light and dark mode:

- Onboarding start screen title is reachable via the VoiceOver heading rotor.
- Poll winner icon + vote count are announced together on a finished poll.
- Poll option icon/text/vote-count are announced as one element per option, with no separate progress-bar element.
- Formatting toolbar buttons are announced as toggles with their current state.
- Avatar edit button on the profile edit screen announces a meaningful label.
- Text field resting-state border/placeholder contrast verified with an automated contrast checker.
- Chat list reorder is communicated to VoiceOver after a new message moves a room up.
- Pinned items banner announces a meaningful label/value in both its loading and loaded states.

## Checklist

- [ ] Pull request contains a [changelog label](https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md#changelog) (`pr-a11y`).

- [ ] Voiceover enabled.


compound-ios/Sources/Compound/BaseStyles/CompoundButtonStyle.swift
```diff
@@ -117,6 +118,11 @@ public struct CompoundButtonStyle: ButtonStyle {
                 makeBackground(configuration: configuration)
             }
             .contentShape(contentShape)
+            .overlay {
+                if isFocused {
+                    contentShape.stroke(.compound.borderInteractivePrimary, lineWidth: 2)
+                }
+            }
     }
```
Velin92: is this for FKA or Voice Over, will this also affect a long tap that gets cancelled? Ideally it's okay if it only affects FKA or Voice Over but we don't want this to impact the UI for a normal user, has this been tested? 
The test case is:
- Long press the button so it's a hovered/focussed state
- Then cancel by releasing outside the button

Does the border appear also while being in the hovered state from a long tap?

ElementX/Sources/Screens/RoomScreen/View/PinnedItemsBanner/PinnedItemsBannerView.swift
```diff
@@ -40,9 +40,11 @@ struct PinnedItemsBannerView: View {
             }
         }
         .disabled(state.isLoading)
-        .accessibilityElement(children: .contain)
+        .accessibilityElement(children: .ignore)
+        .accessibilityLabel(Text(state.displayedMessage))
+        .accessibilityValue(state.count > 1 ? Text(state.bannerIndicatorDescription) : Text(""))
```
Velin92: I would use this: https://developer.apple.com/documentation/swiftui/view/accessibilityvalue(_:isenabled:) the enable can check the state.count > 1 (also could you use isEmpty on this if avaialble instead of checking the count?)

Velin92 (changes requested):

Velin92:
Thanks for the contribution, just asking for a small test and a small change, the rest is good, if this PR is related to some of our A11y issues, could you add them in the PR description with the "fixes #issue_number" so we can also keep track of what issues are being fixed?

ElementX/Sources/Screens/RoomScreen/View/PinnedItemsBanner/PinnedItemsBannerView.swift
unclejay80: I've added [9913af5](https://github.com/element-hq/element-x-ios/pull/5890/commits/9913af514e54de0994791e70b1b65a5ebfbe8f50) to fix the first thing.

What did you have in mind for `isEmpty` here? `state.count` is an Int (derived from the pinned-events dictionary size), not a collection, so there's no `isEmpty` on it directly happy to add a clearer named property (e.g. hasMultiplePinnedItems) if that's what you're after, just want to make sure I'm implementing the right thing.

unclejay80:
> Thanks for the contribution, just asking for a small test and a small change, the rest is good, if this PR is related to some of our A11y issues, could you add them in the PR description with the "fixes #issue_number" so we can also keep track of what issues are being fixed?

sorry these fixes are based on a a11y testing we have commisioned. we did not search for matching issues in your list

compound-ios/Sources/Compound/BaseStyles/CompoundButtonStyle.swift
Velin92: I see this issue was marked as a resolved, does it mean this scenario has been tested?

unclejay80: no. I did revert the changes to CompoundButtonStyle.swift. so this isn't still a part of this PR

unclejay80:
I've fixed the ci issues. hopefully everything is fine now

Velin92 (approved):

PR #5893, opened 2026-07-15
Periphery cleanup

Velin92:
This PR does 3 things
- Fixes periphery runs, and updates it
- Creates a skill to run periphery and have AI clean up the code, or mark with `periphery:ignore` any false positive
- Has run such skill to clean up the whole project


ElementX/Sources/AppHooks/Hooks/RemoteSettingsHook.swift
```diff
@@ -51,11 +51,11 @@ struct DefaultRemoteSettingsHook: RemoteSettingsHookProtocol {
     func updateCache(using client: ClientProtocol) async { }
     func reset(_ appSettings: any CommonSettingsProtocol) { }
     #endif
-
```
pixlwave, line 54: I think committed by mistake?

ElementX/Sources/AppHooks/Hooks/RemoteSettingsHook.swift
```diff
@@ -51,11 +51,11 @@ struct DefaultRemoteSettingsHook: RemoteSettingsHookProtocol {
     func updateCache(using client: ClientProtocol) async { }
     func reset(_ appSettings: any CommonSettingsProtocol) { }
     #endif
-    
     func loadCache(forHomeserver homeserver: String, applyingTo appSettings: CommonSettingsProtocol) { }
 }
 
 private struct ElementWellKnown: Decodable {
+    // periphery:ignore - documents the schema, parsed but not consumed yet
     var version: Int?
```
pixlwave: TBH, this one can go. We're not going to use it now we have a dedicated endpoint.

ElementX/Sources/Application/Windowing/WindowManager.swift
pixlwave, line 1: Heh, I thought we needed this, clearly not 😅

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -1152,8 +1146,7 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
     
     private func presentEmojiPicker(for itemID: TimelineItemIdentifier,
                                     selectedEmoji: Set<String>,
-                                    timelineController: TimelineControllerProtocol,
-                                    animated: Bool) {
```
pixlwave, line 1156: Maybe we should use the `animated` property instead of removing it?

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -1176,8 +1169,7 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
     }
     
     private func presentMapNavigator(interactionMode: LocationSharingInteractionMode,
-                                     timelineController: TimelineControllerProtocol,
-                                     animated: Bool) {
```
pixlwave, line 1180: Ditto, should we use the `animated` property instead of removing it?

ElementX/Sources/Mocks/SessionVerificationControllerProxyMock.swift
```diff
@@ -6,6 +6,8 @@
 // Please see LICENSE files in the repository root for full details.
 //
 
+// periphery:ignore:all
```
pixlwave: Should we just ignore the entire `Mocks` directory in the `periphery.yml` configuration rather than adding these to each file?

ElementX/Sources/Mocks/JoinedRoomProxyMock.swift
```diff
@@ -6,6 +6,8 @@
 // Please see LICENSE files in the repository root for full details.
 //
 
+// periphery:ignore:all
```
pixlwave: This `ignore:all` was ignored 😅

ElementX/Sources/Other/Extensions/CompoundIcon.swift
pixlwave, line 1: Niiiice, no more `ImageAsset`s used as icons 👏

ElementX/Sources/Other/Extensions/String.swift
```diff
@@ -10,21 +10,6 @@ import SwiftUI
 import UniformTypeIdentifiers
 
 nonisolated extension String {
-    /// Returns the string as an `AttributedString` with the specified character tinted in a different color.
```
pixlwave, line 13: This one seems potentially useful to keep for the future?

ElementX/Sources/Other/Extensions/Task.swift
pixlwave, line 1: I feel like we should either keep both `Task.dispatch` or get rid of them together. Having only the throwing variant feels unbalanced.

ElementX/Sources/Other/Extensions/XCTestCase.swift
pixlwave, line 1: Wow, cool to see we've fully migrated away from this :D

ElementX/Sources/Other/HTMLParsing/ElementXAttributeScope.swift
```diff
@@ -73,25 +75,32 @@ enum InlineCodeAttribute: AttributedStringKey {
     static let name = "MXInlineCodeAttribute"
 }
 
-// periphery: ignore - required to make NSAttributedString to AttributedString conversion even if not used directly
```
pixlwave, line 76: What was wrong with the single `ignore` here?

ElementX/Sources/Other/Logging/MXLog.swift
```diff
@@ -95,11 +95,13 @@ nonisolated enum MXLog {
     }
     
     #if DEBUG
+    // periphery:ignore - might be useful to have
     private static let devPrefix = URL.documentsDirectory.pathComponents[2].uppercased()
     /// A helper method for print debugging, only available on debug builds.
     ///
     /// When running on the simulator this will log `[USERNAME] message` so that
     /// you can easily filter the Xcode console to see only the logs you're interested in.
+    // periphery:ignore - might be useful to have
```
pixlwave:
```suggestion
    // periphery:ignore - used for debugging
```

ElementX/Sources/Other/Logging/MXLog.swift
```diff
@@ -95,11 +95,13 @@ nonisolated enum MXLog {
     }
     
     #if DEBUG
+    // periphery:ignore - might be useful to have
```
pixlwave: How come we need to ignore this one? Does it not end up being considered in use, by the fact that `MXLog.dev` is now ignored?

ElementX/Sources/Other/SwiftUI/Backports.swift
pixlwave, line 1: In some ways I imagine we can add this file to the periphery config's ignore too. We'll always clean this file up each year with OS drops as everything has as `if #available` on it.

ElementX/Sources/Screens/Settings/UserDetailsEditScreen/UserDetailsEditScreenCoordinator.swift
```diff
@@ -27,8 +26,8 @@ final class UserDetailsEditScreenCoordinator: CoordinatorProtocol {
     private var viewModel: UserDetailsEditScreenViewModelProtocol
     private var cancellables = Set<AnyCancellable>()
     
-    private let actionsSubject: PassthroughSubject<RoomDetailsEditScreenCoordinatorAction, Never> = .init()
-    var actions: AnyPublisher<RoomDetailsEditScreenCoordinatorAction, Never> {
+    private let actionsSubject: PassthroughSubject<UserDetailsEditScreenCoordinatorAction, Never> = .init()
+    var actions: AnyPublisher<UserDetailsEditScreenCoordinatorAction, Never> {
```
pixlwave, line 30: LOL!

ElementX/Sources/Screens/Timeline/View/Style/TimelineItemBubbledStylerView.swift
```diff
@@ -166,7 +166,7 @@ struct TimelineItemBubbledStylerView<Content: View>: View {
                 context.send(viewAction: .displayTimelineItemMenu(itemID: timelineItem.id))
             }
             .swipeRightAction {
-                SwipeToReplyView(timelineItem: timelineItem)
+                SwipeToReplyView()
```
pixlwave: We should probably just put this thing directly into the modifier give it has a tiny body.

ElementX/Sources/Services/Authentication/ClassicApp/ClassicAppMXAccount.swift
pixlwave, line 1: Heh, I intended to come back and tidy this up after knowing what we needed 🙈

ElementX/Sources/Services/Client/ClientProxyProtocol.swift
```diff
@@ -160,7 +159,6 @@ protocol ClientProxyProtocol: AnyObject {
     var capabilities: HomeserverCapabilitiesProxyProtocol { get }
     
     var isReportRoomSupported: Bool { get async }
-
```
pixlwave, line 163: Probably wasn't meant to be committed.

UnitTests/Sources/RestorationTokenTests.swift
```diff
@@ -157,7 +157,6 @@ struct SessionV1: Equatable {
 }
 
 enum SlidingSyncVersionV1: Equatable {
-    case none
```
pixlwave, line 160: Let's not touch these `Vx` types, they represent structures that the app used to write, and serves as documentation of possible migrations other things.

ElementX/Sources/Screens/KnockRequestsListScreen/KnockRequestsListScreenViewModel.swift
pixlwave, line 1: This feels like the wrong fix. I realise there aren't any actions, but it would follow the architecture much better if the coordinator was to `sink` this publisher instead.

ElementX/Sources/Screens/ManageRoomMemberSheet/ManageRoomMemberSheetViewModelProtocol.swift
pixlwave, line 1: Maybe we should change all the `manageMemberViewModel` bindings to use the protocol instead of the actual VM and then we can remove this ignore.

ElementX/Sources/Screens/RoomMemberListScreen/RoomMembersListScreenModels.swift
pixlwave, line 1: I question if we should be removing the `alertInfo` if there aren't any types of alert to be using with it?

ElementX/Sources/Services/Analytics/AnalyticsServiceProtocol.swift
```diff
@@ -41,7 +41,6 @@ protocol AnalyticsServiceProtocol: AnyObject {
     /// be called on sign-out to maintain opt-in status, whilst ensuring the next
     /// account used isn't associated with the previous one.
     /// Note: **MUST** be called before stopping PostHog or the reset is ignored.
```
pixlwave: We should probably remove these too if the function is going 😄

ElementX/Sources/Services/Media/Provider/MediaProviderProtocol.swift
```diff
@@ -20,7 +20,6 @@ enum MediaProviderError: Error {
 // sourcery: AutoMockable
 nonisolated protocol MediaProviderProtocol: Sendable {
     func imageFromSource(_ source: MediaSourceProxy?, size: CGSize?) -> UIImage?
-    func loadImageFromSource(_ source: MediaSourceProxy, size: CGSize?) async -> Result<UIImage, MediaProviderError>
```
pixlwave, line 23: Ok took me a while to understand this removal, but it's because we always use the retrying variant now. I think we should probably make the method `private` inside `MediaProvider` then?

ElementX/Sources/Services/Room/RoomInfoProxy.swift
pixlwave, line 1: These removals strike me like they should all have the `potentially useful` comment on. It's annoying to try use a proxy and find out that what you want exists but just isn't exposed so doesn't show up via auto-completion.

pixlwave (commented):
Alright, here's my review on the code changes, haven't looked at the skill yet.

ElementX/Sources/Screens/ManageRoomMemberSheet/ManageRoomMemberSheetViewModelProtocol.swift
Velin92: I looked into this on, it doesn't work with the current presentation unfortunately. The binding feeds .sheet(item:), which needs its Item to conform to Identifiable, and the any ManageRoomMemberSheetViewModelProtocol can't satisfy that generic constraint even if the protocol inherits it.
We can get around it but is a bit of a refactor that maybe is not worth doing here.

pixlwave: Ah ok yeah makes sense 👍

ElementX/Sources/Services/Media/Provider/MediaProviderProtocol.swift
pixlwave: I don't think this was made private?

ElementX/Sources/Services/Analytics/AnalyticsService.swift
```diff
@@ -57,7 +57,11 @@ class AnalyticsService: AnalyticsServiceProtocol {
         MXLog.info("Started.")
     }
     
-    func reset() {
+    /// Resets any IDs and event queues in the analytics client. This method should
+    /// be called on sign-out to maintain opt-in status, whilst ensuring the next
+    /// account used isn't associated with the previous one.
```
pixlwave, line 62: Technically this comment is incorrect as we don't maintain opt-in status in Element X, we ask each time. So we should probably update to:

```suggestion
    /// Resets any IDs and event queues in the analytics client. This method should
    /// be called on sign-out to ensure the next
    /// account used isn't associated with the previous one.
```

pixlwave (approved):
👌

ElementX/Sources/Services/Media/Provider/MediaProviderProtocol.swift
Velin92: Went with internal instead of private — MediaProviderTests calls loadImageFromSource(_:size:) directly in six places, so private breaks the build.

pixlwave: Riiight, didn't think about tests 🤦‍♂️

PR #5897, opened 2026-07-17
Group Element Pro changes into a dedicated release notes section

manuroe:
This is a proposal to avoid confusion between the features than land in EX and code changes that are made only for Element Pro.

It add a new tag `element-pro` which does not start with `pr-` to pass Danger checks.
All `element-pro` PRs will land on a dedicated section at the bottom for visibility. We are losing the granularity of feature vs bugfix vs etc but we could be a separate homemade script later for proper Element Pro release notes.

I am really not sure about 💎 .

### Pull Request Checklist

- [x] I read the [contributing guide](https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md).
- [x] I am aware of the [etiquette](https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md#etiquette).
- [x] Pull request contains a [changelog label](https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md#changelog).
- [x] This PR has been made with the help of an LLM.

PR #5899, opened 2026-07-17
Group Element Pro changes into a dedicated release notes section

manuroe:
This PR replaces #5897 with a simpler approach using a simple new [`pr-element-pro`](https://github.com/element-hq/element-x-ios/labels?q=pr-element-pro) label.

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -28,6 +28,10 @@ changelog:
       labels:
         - pr-wip
 
+    - title: 💼 Element Pro
+      labels:
+        - pr-element-pro
+
     - title: Others
       labels:
         - pr-misc


.github/release.yml
```diff
@@ -28,6 +28,10 @@ changelog:
       labels:
         - pr-wip
 
+    - title: 💼 Element Pro
```
manuroe, line 31: It's not really at the bottom because the "*" rule of `Others` would catch them.

pixlwave (approved):
Works for me, thanks for the update 🎉

.github/release.yml
pixlwave: I think that's fine. `Others` is usually just noise anyway.

PR #5902, opened 2026-07-20
Hook up the emoji picker to the custom status row.

pixlwave:
This PR makes 2 changes:
- Decouples the emoji picker from the timeline controller.
- Hooks up the emoji picker to the custom status flow.


Velin92 (approved):
LGTM, so you are passing around a continuation instead of a publisher to ensure non isolation of the value?

pixlwave:
> LGTM, so you are passing around a continuation instead of a publisher to ensure non isolation of the value?

Honestly I just started doing this in #4935 and have like the pattern ever since. Not sure I was specifically thinking about isolation but I guess it helps at times.

PR #5904, opened 2026-07-20
Enable client pausing and resuming for all builds

stefanceriu:
The mechanism has been running on nightly and debug builds since it was introduced and Sentry shows background teardown panic occurring far less on nightly than on production.

Flip the default on for release builds too so store users get the fix.


pixlwave (approved):
I feel like it would be reasonable to remove the setting/toggle at this stage so we don't forget about it later perhaps?

stefanceriu:
Baby steps, let's make sure nothing goes horribly wrong first.

PR #5905, opened 2026-07-20
Update the SDK and hook up the status from Room Heroes.

pixlwave:
This PR
- Updates the SDK.
- Includes the SDK's `RoomHero` status fields when mapping to our `UserProfile`
- Starts requesting profiles via /sync when the user status feature flag is enabled.
- Hooks up the new subscribeToOwnProfile method based on the feature flag (but still needs an API to additionally check if the server supports it).

### DIFF
--- a/ElementX/Sources/Services/Client/ClientProxy.swift
+++ b/ElementX/Sources/Services/Client/ClientProxy.swift
@@ -33,6 +33,9 @@ class ClientProxy: ClientProxyProtocol {
     private var syncServiceStateUpdateTaskHandle: TaskHandle?
     
     // periphery:ignore - required for instance retention in the rust codebase
+    private var userProfileListenerTaskHandle: TaskHandle?
+    
+    // periphery:ignore - required for instance retention in the rust codebase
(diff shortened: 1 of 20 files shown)


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -278,9 +281,13 @@ class ClientProxy: ClientProxyProtocol {
         
         try await client.setUtdDelegate(utdDelegate: ClientDecryptionErrorDelegate(actionsSubject: actionsSubject))
         
-        loadUserAvatarURLFromCache()
+        let canSubscribeToUserProfile = appSettings.userStatusEnabled // TODO: @pixlwave to add a /versions check too.
```
Velin92, line 284: is this coming in another PR?

Velin92 (approved):

ElementX/Sources/Services/Client/ClientProxy.swift
pixlwave: Sure is, will need https://github.com/matrix-org/matrix-rust-sdk/pull/6778

PR #5911, opened 2026-07-21
Content scanner also checks for thumbnails

Velin92:
The scanning now includes thumbnails too, and if one between the original source or the thumbnail fails, we flag the whole item as a failure AND cancel the other request.

Included also a small change for the dev options screen to keep track if such screen is used before or after a log in.

### DIFF
--- a/ElementX/Sources/Screens/Timeline/View/Replies/TimelineReplyView.swift
+++ b/ElementX/Sources/Screens/Timeline/View/Replies/TimelineReplyView.swift
@@ -25,6 +25,7 @@ struct TimelineReplyView: View {
     var body: some View {
         ContentScanningView(contentScannerService: timelineContext?.contentScannerService,
                             mediaSource: scannedMediaSource,
+                            thumbnailSource: scannedThumbnailSource,
                             containerShowsFailure: false) {
             content
                 .roundedContainer(padding: 4,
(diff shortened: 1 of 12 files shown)


ElementX/Sources/Screens/Timeline/View/Replies/TimelineReplyView.swift
```diff
@@ -63,6 +64,21 @@ struct TimelineReplyView: View {
...
+    private var scannedThumbnailSource: MediaSourceProxy? {
+        guard case .loaded(_, _, let eventContent) = timelineItemReplyDetails,
+              case .message(let message) = eventContent else {
+            return nil
+        }
+        
+        switch message {
+        case .image(let content): return content.thumbnailInfo?.source
+        case .video(let content): return content.thumbnailInfo?.source
+        default: return nil
```
pixlwave, line 78: We scan a file's thumbnail in `FileRoomTimelineView` but we don't in the reply view? Should we be consistent here (one way or the other, I'm not sure it matters much as we don't actually show the file's thumbnail in EXI right)?

pixlwave (approved):
Nice, LGTM 👍

ElementX/Sources/Screens/Timeline/View/Replies/TimelineReplyView.swift
Velin92: decided to go for the scan all never regret route.

PR #5912, opened 2026-07-21
Add support for showing the user's own status over their avatar on the app's Settings button.

pixlwave:
- Make a shared view modifier for an inverse mask and re-use everywhere.
- Switch from the old `luminanceToAlpha` trick to the `destinationOut` blend mode. Apparently this is the more common approach.
    - Nice little benefit is that the masks now render properly in our Preview snapshots 🎉
- Add support for showing the user's status over their avatar in the app's Settings button
    - Note: I've scaled things a bit compared to the design as we're currently using an inset 32pt avatar instead of a full-sized 40pt avatar.

avatarSettingsButtonLabel iPhone-0

- Bonus change: The room header decoration icons are now aligned with the room's title/status emoji when there is a subtitle.

Screenshot 2026-07-21 at 5 57 49 pm


ElementX/Sources/Other/SwiftUI/Views/AvatarSettingsButtonLabel.swift
```diff
@@ -0,0 +1,86 @@
...
+    }
+    
+    var emojiText: (some View)? {
+        statusEmoji.map {
+            Text(String($0))
+                // All values (inc font) scaled by 0.8 as our Liquid Glass avatar is 32pt and not 40pt.
+                .font(.compound.bodyXS)
+                .padding(.vertical, 1) // Match Figma's line height.
+                .alignmentGuide(.trailing) { $0.width - 6.5 }
+                .alignmentGuide(.bottom) { $0.height + 1.5 }
```
Velin92, line 55: how does this behave with dynamic type?

ElementX/Sources/Other/SwiftUI/InverseMask.swift
```diff
@@ -0,0 +1,24 @@
...
+// Please see LICENSE files in the repository root for full details.
+//
+
+import SwiftUI
+
+extension View {
+    /// Masks this view using the inverted alpha channel of the given view.
+    ///
+    /// The inverse of `mask(alignment:_:)`.
+    func inverseMask(alignment: Alignment = .center, @ViewBuilder _ inverseMask: () -> some View) -> some View {
```
Velin92, line 14: maybe I would change the name of this function, I know this should look as an overload of `inverseMask` but the existing `inverseMask` doesn't do so many things, so it's not really overloading the original function, but it's more like a custom one using it, so I would actually rename it to `inverseMasking` (since we are inverse masking as an overlay that content) or `elementInverseMask`

Velin92 (approved):
LGTM, just few comments

ElementX/Sources/Other/SwiftUI/InverseMask.swift
pixlwave: I don't understand this, there isn't an existing `inverseMask` that this is overloading, this is a new function that removes the duplication we had in multiple views.

ElementX/Sources/Other/SwiftUI/Views/AvatarSettingsButtonLabel.swift
pixlwave: Not great, but given this view lives in the toolbar, it basically grows by not much more that 6pt from `.small` to `.ax5` so the different is hardly noticable:

DT-AX5
DT-small

ElementX/Sources/Other/SwiftUI/InverseMask.swift
pixlwave: Merging this now, can do a follow-up if necessary.

PR #5914, opened 2026-07-22
Add signed in flag to the developer options screen hook

Velin92:
to differentiate between the signed in and the logged out state when presenting the dev options screen


ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/DeveloperOptionsScreenViewModel.swift
```diff
@@ -24,7 +24,7 @@ class DeveloperOptionsScreenViewModel: DeveloperOptionsScreenViewModelType, Deve
         self.clientProxy = clientProxy
         super.init(initialViewState: .init(appHooks: appHooks,
                                            shouldShowClearCache: clientProxy != nil,
```
pixlwave, line 26: tbf I didn't realise this, we can also remove `shouldShowClearCache` and use `isSignedIn` here as well.

pixlwave (approved):
Nice, thanks 😎

PR #5920, opened 2026-07-22
Gate user status behind server support as well as the feature flag.

pixlwave:
Uses https://github.com/matrix-org/matrix-rust-sdk/pull/6778

### DIFF
--- a/ElementX/Sources/Screens/Settings/SettingsScreen/SettingsScreenViewModel.swift
+++ b/ElementX/Sources/Screens/Settings/SettingsScreen/SettingsScreenViewModel.swift
@@ -29,7 +29,7 @@ class SettingsScreenViewModel: SettingsScreenViewModelType, SettingsScreenViewMo
         
         super.init(initialViewState: .init(deviceID: userSession.clientProxy.deviceID,
                                            userProfile: userSession.clientProxy.userProfilePublisher.value,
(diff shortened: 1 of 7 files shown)


pixlwave:
Rebased with the fixes for the call timeline item snapshots.

ElementX/Sources/Screens/Settings/SettingsScreen/SettingsScreenViewModel.swift
```diff
@@ -29,7 +29,7 @@ class SettingsScreenViewModel: SettingsScreenViewModelType, SettingsScreenViewMo
         
         super.init(initialViewState: .init(deviceID: userSession.clientProxy.deviceID,
                                            userProfile: userSession.clientProxy.userProfilePublisher.value,
-                                           showUserStatus: appSettings.userStatusEnabled,
+                                           showUserStatus: false,
```
Velin92: I don't think there is need for this parameter anymore to be passed in the init and can be just `var showUserStatus = false.`

Velin92 (approved):
LGTM

ElementX/Sources/Screens/Settings/SettingsScreen/SettingsScreenViewModel.swift
pixlwave: Good point 👍

PR #5909, opened 2026-07-21
Support sending gallery messages behind a feature flag

Velin92:
Adds sending of multi-media "gallery" messages, gated behind the new galleryEnabled developer option (Developer Options → Room, off by default). Rendering of received galleries ships in a separate PR.

- Wires sendGallery through TimelineProxy/TimelineController onto the SDK's Timeline.sendGallery.
- MediaUploadPreviewScreen sends selected items as one gallery when the flag is on.

Co-authored with the original author of #5601.

NOTE: This PR allows to send gallery messages, but they won't be rendered without the gallery timeline item PR.

PR #5601, opened 2026-05-16
Support gallery messages

bxdxnn:
See https://github.com/element-hq/element-meta/issues/3201

This PR is vibecoded, but it was tested in an iOS Simulator 

### Pull Request Checklist

- [x] I read the [contributing guide](https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md).
- [ ] Pull request contains a [changelog label](https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md#changelog).

**UI changes have been tested with:**
- [x] iPhone and iPad simulators in portrait and landscape orientations.
- [ ] Dark mode enabled and disabled.
- [ ] Various sizes of dynamic type.
- [ ] Voiceover enabled.

PR #5922, opened 2026-07-22
Support rendering gallery messages in the timeline

Velin92:
Renders received gallery messages (multiple media in a single event) in the timeline. Always on — sending is gated separately (#5909).

- Images/videos shown in a grid, files in a list, each tile/row wired through the content scanner with loading and unsafe states.
- Tapping opens the existing full-screen media preview with swipe navigation across the gallery's items.


ElementX/Sources/Screens/FilePreviewScreen/View/TimelineMediaPreviewController.swift
```diff
@@ -251,7 +251,7 @@ class TimelineMediaPreviewController: QLPreviewController {
         }
     }
     
-    private func handleFileLoaded(itemID: TimelineItemIdentifier.EventOrTransactionID) {
+    private func handleFileLoaded(itemID: String) {
```
pixlwave: Can we use an enum here so we stay strongly typed please?

ElementX/Sources/Services/Timeline/TimelineItems/Items/Messages/GalleryRoomTimelineItemContent.swift
```diff
@@ -0,0 +1,61 @@
...
+}
+
+nonisolated struct GalleryItem: Hashable, Identifiable {
+    enum Kind: Hashable {
+        case image
+        case video
+        case audio
+        case file
+        case other
+    }
```
pixlwave: Given the SDK gives us back the exact same content types as it does for individual messages, surely we should do the same. Something like:
```suggestion
nonisolated enum GalleryItem: Hashable, Identifiable {
    case image(id: GalleryItemID, ImageRoomTimelineItemContent)
    case video(id: GalleryItemID, VideoRoomTimelineItemContent)
    case audio(id: GalleryItemID, AudioRoomTimelineItemContent)
    case file(id: GalleryItemID, FileRoomTimelineItemContent)
    case other(id: GalleryItemID, *not sure*)
}
```

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryRoomTimelineView.swift
```diff
@@ -0,0 +1,272 @@
...
+    
+    private var hasMediaCaption: Bool {
+        timelineItem.content.caption != nil
+    }
+    
+    /// A gallery falls back to a vertical file-list layout when any of its items lacks a thumbnail
+    /// (typically audio or generic files). Mixed galleries (image/video + file) also use the list
+    /// — a grid of mostly-icons looks worse than a tidy list.
+    private var usesListLayout: Bool {
+        timelineItem.content.items.contains { !$0.hasThumbnail }
```
pixlwave: What happens if there's 2 files that have thumbnails? Do we hardcode `false` for those?

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryItemTileView.swift
```diff
@@ -0,0 +1,124 @@
...
+    private var overflowOverlay: some View {
+        ZStack {
+            Color.compound.bgCanvasDefault.opacity(0.7)
+            Text("+\(overflowCount)")
+                .font(.compound.headingSMSemibold)
+                .foregroundStyle(.compound.textPrimary)
+        }
+    }
+    
+    private func formatted(duration: TimeInterval) -> String {
```
pixlwave, line 111: Come on, we have Swift's `formatted` here, we shouldn't be manually assembling times like this.

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryItemTileView.swift
```diff
@@ -0,0 +1,124 @@
...
+            LinearGradient(colors: [.clear, .compound.bgCanvasDefault],
+                           startPoint: .top,
+                           endPoint: .bottom)
+        }
+    }
+    
+    private var overflowOverlay: some View {
+        ZStack {
+            Color.compound.bgCanvasDefault.opacity(0.7)
+            Text("+\(overflowCount)")
```
pixlwave, line 105: Should this be localised as there may be different ways of representing this is other languages (or simply just RTL support required).

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryTileView.swift
```diff
@@ -0,0 +1,124 @@
...
+// Copyright 2025 New Vector Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Compound
+import SwiftUI
+
+struct GalleryTileView: View {
```
pixlwave: With 4 of these views we should be specific here.

```suggestion
struct GalleryItemTileView: View {
```

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryTileView.swift
```diff
@@ -0,0 +1,124 @@
...
+    let item: GalleryItem
+    let uniqueID: TimelineItemIdentifier.UniqueID
```
pixlwave: We shouldn't need this with a strongly typed item ID.

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryTileView.swift
```diff
@@ -0,0 +1,124 @@
...
+    let onTap: () -> Void
+    
+    var body: some View {
+        // Each tile is gated on its own scan state. `containerShowsFailure` is false so that an
+        // unsafe item only turns its own tile critical rather than the whole gallery bubble.
+        ContentScanningView(contentScannerService: contentScannerService,
+                            mediaSource: item.mediaSource,
+                            containerShowsFailure: false) {
+            // `Color.clear` adopts the tile size set by the grid so the image fills and clips to it,
+            // and the overlays align to the tile bounds rather than the image's larger filled size.
```
pixlwave: Probably not necessary.

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryTileView.swift
```diff
@@ -0,0 +1,124 @@
...
+    private var placeholder: some View {
+        Color.compound.bgSubtleSecondary.opacity(0.9)
+    }
+    
+    private var iconFallback: some View {
+        ZStack {
+            Color.compound.bgSubtleSecondary
+            CompoundIcon(item.isAudio ? \.audio : \.attachment,
+                         size: .medium,
+                         relativeTo: .compound.bodyLG)
```
pixlwave: Those are the default values already.

```suggestion
            CompoundIcon(item.isAudio ? \.audio : \.attachment)
```

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryGridView.swift
```diff
@@ -0,0 +1,89 @@
...
+
+import Compound
+import SwiftUI
+
+struct GalleryGridView: View {
+    let items: [GalleryItem]
+    let uniqueID: TimelineItemIdentifier.UniqueID
+    let mediaProvider: MediaProviderProtocol?
+    let contentScannerService: ContentScannerServiceProtocol?
+    let onTap: (Int) -> Void
```
pixlwave: Could we be slightly clearer here API wise?

```suggestion
    let onItemTap: (Int) -> Void
```

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryGridView.swift
```diff
@@ -0,0 +1,89 @@
...
+    static let groupWidth: CGFloat = 264
+    static let maxVisible = 5
+    
+    private var visibleCount: Int {
+        min(items.count, Self.maxVisible)
+    }
+    
+    /// Number to show on the overflow tile when items.count > maxVisible.
+    /// The last visible tile becomes the overflow indicator, so it represents
+    /// itself + every hidden item.
```
pixlwave:
```suggestion
    /// Number shown on the overflow tile when items.count > maxVisible.
```

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryGridView.swift
```diff
@@ -0,0 +1,89 @@
...
+    private var overflowCount: Int {
+        items.count > Self.maxVisible ? items.count - Self.maxVisible + 1 : 0
```
pixlwave, line 28:
```suggestion
        // Represents the last visible item + the rest.
        items.count > Self.maxVisible ? items.count - Self.maxVisible + 1 : 0
```

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryGridView.swift
```diff
@@ -0,0 +1,89 @@
...
+            
+            GalleryTileView(item: items[index],
+                            uniqueID: uniqueID,
+                            mediaProvider: mediaProvider,
+                            contentScannerService: contentScannerService,
+                            overflowCount: overflow) {
+                onTap(index)
+            }
+            .frame(width: size.width, height: size.height)
+            .clipped()
```
pixlwave: The tile already has a `.clipped()` in its body, is this one strictly necessary?

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryGridView.swift
```diff
@@ -0,0 +1,89 @@
...
+                let totalSpacing = CGFloat(row.indices.count - 1) * Self.spacing
+                let tileWidth = (Self.groupWidth - totalSpacing) / CGFloat(row.indices.count)
+                HStack(spacing: Self.spacing) {
+                    ForEach(row.indices, id: \.self) { index in
+                        tile(at: index, size: CGSize(width: tileWidth, height: row.height))
+                    }
+                }
+            }
+        }
+        .frame(width: Self.groupWidth)
```
pixlwave: Do we need this given the GalleryRoomTimelineView already has it too?

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryGridView.swift
```diff
@@ -0,0 +1,89 @@
+//
+// Copyright 2025 Element Creations Ltd.
+// Copyright 2025 New Vector Ltd.
```
pixlwave: Not sure if all the new files are like this, but there's at least a few:

```suggestion
// Copyright 2026 Element Creations Ltd.
```

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryListView.swift
```diff
@@ -0,0 +1,145 @@
...
+
+/// Used for gallery messages that contain at least one attachment without a thumbnail
+/// (typically a document or audio). Renders a vertical list of file rows so each item is
+/// individually recognisable — the grid layout doesn't read well for non-visual content.
+struct GalleryListView: View {
+    let items: [GalleryItem]
+    let uniqueID: TimelineItemIdentifier.UniqueID
+    let mediaProvider: MediaProviderProtocol?
+    let contentScannerService: ContentScannerServiceProtocol?
+    let onTap: (Int) -> Void
```
pixlwave:
```suggestion
    let onItemTap: (Int) -> Void
```

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryListView.swift
```diff
@@ -0,0 +1,145 @@
...
+                    onTap(index)
+                }
+            }
+        }
+    }
+    
+    static var galleryDivider: some View {
+        Rectangle()
+            .fill(Color.compound.borderInteractiveSecondary)
+            .frame(height: 0.5)
```
pixlwave: We should use `@Environment(\.pixelLength)` for this as `0.5` doesn't work properly for 3x devices

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryListView.swift
```diff
@@ -0,0 +1,145 @@
...
+    }
+    
+    private var galleryDivider: some View {
+        Self.galleryDivider
+    }
```
pixlwave: How about we make a proper view rather than this weird static var stuff?

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryListView.swift
```diff
@@ -0,0 +1,145 @@
...
+
+private struct GalleryListRow: View {
+    let item: GalleryItem
+    let uniqueID: TimelineItemIdentifier.UniqueID
+    let mediaProvider: MediaProviderProtocol?
+    let contentScannerService: ContentScannerServiceProtocol?
+    let onTap: () -> Void
+    
+    private var fileDescription: String {
+        ".\(item.filename.validatedFileExtension.uppercased())"
```
pixlwave: Do we really need the `.` here, feels like we'd need to think about RTL or something like that? We don't put it on e.g. the `FileRoomTimelineView`?

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryListView.swift
```diff
@@ -0,0 +1,145 @@
...
+    }
+    
+    private var iconKeyPath: KeyPath<CompoundIcons, Image> {
+        item.isAudio ? \.audio : \.attachment
+    }
+    
+    var body: some View {
+        // The whole row is gated on its scan state. `containerShowsFailure` is false so an unsafe
+        // item only turns its own row critical rather than the whole gallery bubble.
```
pixlwave: I don't think this adds any value, it's completely clear by looking at the simple body.

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryListView.swift
```diff
@@ -0,0 +1,145 @@
...
+                          mediaProvider: mediaProvider) { imageView in
+                imageView
+                    .aspectRatio(contentMode: .fill)
+            } placeholder: {
+                Color.compound.bgSubtleSecondary
+            }
+            .frame(width: 36, height: 36)
+            .clipShape(RoundedRectangle(cornerRadius: 4, style: .continuous))
+        } else {
+            CompoundIcon(iconKeyPath, size: .medium, relativeTo: .body)
```
pixlwave: Default values again.

```suggestion
            CompoundIcon(iconKeyPath)
```

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryListView.swift
```diff
@@ -0,0 +1,145 @@
...
+                .foregroundColor(.compound.iconPrimary)
+                .scaledPadding(6)
+                .background(.compound.iconOnSolidPrimary,
+                            in: RoundedRectangle(cornerRadius: 4, style: .continuous))
```
pixlwave: And again
```suggestion
                            in: RoundedRectangle(cornerRadius: 4))
```

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryListView.swift
```diff
@@ -0,0 +1,145 @@
...
+            .fill(Color.compound.borderInteractiveSecondary)
+            .frame(height: 0.5)
+    }
+    
+    private var galleryDivider: some View {
+        Self.galleryDivider
+    }
+}
+
+private struct GalleryListRow: View {
```
pixlwave, line 48: General nit: This uses "accessory" everywhere but isn't the definition of an accessory, the view shown on the trailing edge of a view? It seems like we should copy the `Icon` naming from `ReplyView` here instead as it is basically the same format (mediaSource or Compound Icon).

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryRoomTimelineView.swift
```diff
@@ -0,0 +1,272 @@
...
+    var body: some View {
+        TimelineStyler(timelineItem: timelineItem) {
+            VStack(alignment: .leading, spacing: 8) {
+                if usesListLayout {
+                    GalleryListView(items: timelineItem.content.items,
+                                    uniqueID: timelineItem.id.uniqueID,
+                                    mediaProvider: context?.mediaProvider,
+                                    contentScannerService: context?.contentScannerService) { index in
+                        tap(index: index)
+                    }
```
pixlwave: We could do this here and below given the signatures match

```suggestion
                                    contentScannerService: context?.contentScannerService,
                                    onItemTap: tap)
```

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryRoomTimelineView.swift
```diff
@@ -0,0 +1,272 @@
...
+    }
+    
+    private func tap(index: Int) {
+        context?.send(viewAction: .galleryItemTapped(itemID: timelineItem.id, index: index))
+    }
+}
+
+// MARK: - Previews
+
+struct GalleryRoomTimelineView_Previews: PreviewProvider, TestablePreview {
```
pixlwave, line 72: So much boiler-plate. Can we move some of this (like all the content scanning results) to the mock instead and just configure it here with a parameter or something?

Oh or add them to `TimelineFixtures` if that works.

Anything to make this simpler *and* more reusable :)

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
```diff
@@ -242,45 +303,54 @@ enum TimelineMediaPreviewItem: Equatable {
         // MARK: Media details
         
         var mediaSource: MediaSourceProxy? {
+            if let galleryItem {
+                return galleryItem.mediaSource
+            }
             switch timelineItem {
```
pixlwave: This happens a lot in this file. We can just

```
return switch timelineItem {
    case:
       value
}
```

rather than adding all those `return`s.

pixlwave (commented):
Ok first review (because this is getting big enough already). I haven't looked at the preview controller stuff at all yet, will come back to that in round 2! 🙈

ElementX/Sources/Services/Timeline/TimelineItems/Items/Messages/GalleryRoomTimelineItemContent.swift
pixlwave: And again, let's use a typed ID here, not a string.

Velin92: Done — GalleryItem is now an enum over the SDK content types; .other is case other(id:, filename:) (no media to preview). IDs are a typed GalleryItemID (parent uniqueID + index)

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryRoomTimelineView.swift
Velin92: Changed usesListLayout to key off item type: any non-image/video → list.

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryItemTileView.swift
Velin92: left it as it is, just because we do the same for read receipts, maybe we can take care of this in the future

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryItemTileView.swift
Velin92: Switched to Duration.formatted(.time(pattern:)), keeping trimmed H:MM:SS (hours only when ≥1h) rather than the flat .minuteSecond.

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryListView.swift
Velin92: CompoundIcon has no size-only init, so dropping args would switch relativeTo from `.body` to `.compound.bodyLG`, diverging from FileRoomTimelineView. Instead extracted a shared FileTypeIconView

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryListView.swift
Velin92: as above

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryRoomTimelineView.swift
Velin92: Moved per-source scan results into `perSourceScanResult`

PR #5924, opened 2026-07-23
Remember camera and microphone state between Element Calls

IOrelI:
Content
[element-hq/element-meta#3234](https://github.com/element-hq/element-meta/issues/3234)

Saving camera and microphone state between group calls (for example if in the last group call the user used the camera, the camera will be enabled by default on the next call; if the user was muted in the last group call, the microphone will be disabled by default etc.)

Element Call always joined with the camera on regardless of how the user left it in their previous call. The last known audio/video state is now persisted and restored via a toWidget mediaState message once the widget reports content_loaded.

Motivation and context
In my organization on many meetings some users don't use microphone or camera most of the time so they are manually turning off camera and microphone before each call, and it's annoying.

Screenshots / GIFs
UI doesn't change.

Tests
Join a group call
Enable/disable microphone and/or camera during the call
Leave and join the next group call in the same room, see that camera and microphone are enabled/disabled as they were left in the last call

### Pull Request Checklist

- [x] Voiceover enabled.


pixlwave:
Sorry but as mentioned in the contributing guide we can't accept this PR. Whilst there might be an issue in Element Meta, there is no product feedback on this and so we can't determine if the approach is correct. Some questions for example:

- Is this something the Element Call team even want?
- Should it be remembered as a single state across all calls?
    - Maybe a user wants one way for group calls and another for DM calls.
    - Maybe a user only wants the camera on for their family, but not their colleagues.

We can't just decide these behaviours, this is why we have a product team. Please discuss this with them in the meta issue and when you have the answer and a sign-off, then we can look at a PR :)

bon3o:
> Sorry but as mentioned in the contributing guide we can't accept this PR. Whilst there might be an issue in Element Meta, there is no product feedback on this and so we can't determine if the approach is correct. Some questions for example:
> 
> * Is this something the Element Call team even want?
> * Should it be remembered as a single state across all calls?
>   
>   * Maybe a user wants one way for group calls and another for DM calls.
>   * Maybe a user only wants the camera on for their family, but not their colleagues.
> 
> We can't just decide these behaviours, this is why we have a product team. Please discuss this with them in the meta issue and when you have the answer and a sign-off, then we can look at a PR :)

Hello! The issue is that the discussion in the meta issue feels more like a monologue than an actual discussion. My message hasn't received any response. 

https://github.com/element-hq/element-meta/issues/3234

pixlwave:
I'm afraid that doesn't change the fact that we as developers can't accept changes like this without product input, so opening PR anyway isn't the right solution for the lack of activity on the issue.

The lack of a robust triage system on element-meta is known about, but for now I would suggest you bring up your issue in e.g. [#webrtc:matrix.org](https://matrix.to/#/#webrtc:matrix.org) and see if the Element Call team are able to provide any input at this time.

PR #5926, opened 2026-07-23
Add content scanner to the NSE Client

Velin92:
We won't pre-scan anything, is just to allow also the NSE to scan when a download is tried from it, and eventually have it fail when the scan fails too.


pixlwave (approved):
Ah yeah, right 👍

Issue #5451, opened 2026-04-21
Cannot Select Text in iOS App


f1yankees:
### Steps to reproduce

1. Attempt to double press / double tap on text of any message in a channel
2. Note that on the screen, nothing happens and no cursor appears

#### What did you expect?
I expected an ability to highlight text, similar to the standard behavior of an iOS application

#### What happened instead?
The Apple Intelligence button pops up instead

### Your phone model

iPhone 16 Pro

### Operating system version

iOS 26.3.1 (23D8133)

### Application version

26.04.01 (225)

### Will you send logs?

Yes

Mushy-Snugglebites-badonkadonk:
I’d like to add that this would be especially useful now that more people are using AI agents and assistants inside Matrix rooms.

Agent replies can be long, structured, and information-dense. Being able to select/copy a specific portion of a message makes it much easier to quote the relevant part, pull out a command or detail, reference a specific section in a follow-up, or move text into another tool without copying the entire reply.

This matters not just for convenience, but for accuracy: when working with long agent-generated responses, users often need to preserve exact wording from one part of a reply rather than paraphrasing or manually retyping it.

qike-ms:
I have an implementation in a public fork that follows the dedicated-selection-mode direction described here: the existing message action sheet gains **Select text**, which opens an isolated selectable sheet while leaving timeline gestures and whole-message Copy unchanged.

Fork commit: https://github.com/qike-ms/element-x-ios/commit/e68c557cab0b37b63e5f467c5c0540ba122b2fcc

I am validating it through independent review and macOS CI before requesting review. I will keep the upstream PR in draft until compilation, unit/preview/accessibility checks, and interactive selection behavior have evidence.

PR #5930, opened 2026-07-26
Add dedicated message text selection

qike-ms:
Add a dedicated text-selection sheet for timeline messages.

Long agent and human messages often contain a command, quote, identifier, or paragraph that needs to be copied exactly. The existing Copy action only copies the whole message, while inline UITextView selection was previously removed because it conflicts with timeline scrolling, links, and swipe-to-reply gestures.

This change follows element-hq/element-x-ios#5451:

- adds Select text beside the existing whole-message Copy action;
- presents the complete text or media caption in an isolated selection sheet;
- enables native substring selection only inside that sheet;
- keeps normal timeline gesture safeguards unchanged;
- supports formatted bodies, captions, mentions, links, blockquotes, code blocks, RTL text, and Dynamic Type through the existing attributed-text renderer;
- adds focused action-provider and selectable-content tests plus a testable preview.

Verification on the Linux implementation host:

- exact 13-file staged set verified;
- `git diff --cached --check` passed before commit;
- `git show --check` passes on the pushed commit;
- local and remote commit SHAs match;
- working tree is clean.

Xcode compilation, unit tests, previews, accessibility snapshots, and interactive selection behavior still require macOS CI/device verification. This PR remains draft until those checks pass.

Upstream issue: element-hq/element-x-ios#5451

### DIFF
--- a/ElementX.xcodeproj/project.pbxproj
+++ b/ElementX.xcodeproj/project.pbxproj
@@ -55,6 +55,7 @@
 		05FF0CD80EDAB3A7C0D4700A /* InfoPlistReader.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6A580295A56B55A856CC4084 /* InfoPlistReader.swift */; };
 		0638CBDE3098B1C3F23AFCFA /* MXLog.swift in Sources */ = {isa = PBXBuildFile; fileRef = 111B698739E3410E2CDB7144 /* MXLog.swift */; };
 		065EAB39F3F3AB4F6BD2A362 /* AppLockSetupTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 19DD166C3625EE426203FA29 /* AppLockSetupTests.swift */; };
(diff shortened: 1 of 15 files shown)


qike-ms:
Independent review found three blockers in the current draft: mention pills need flattening before native copy, link interactions must be suppressed in selection mode, and long selections need an internally scrolling text view for handle autoscroll. I am addressing these in a follow-up commit; this PR is not ready for review yet.

qike-ms:
The review findings are now addressed, including regenerating the checked-in Xcode project so the new view and both test files are in their correct Sources phases. The current head is `3ddd2d47f5e88dfb1712264211f74500a3f58e9a`.

Could a maintainer please approve the first-time contributor workflow runs when convenient? Unit Tests, Enterprise Unit Tests, Danger, and zizmor are currently `action_required` and have not executed. I will keep the PR in draft until those checks and device-sensitive selection behavior are validated.

pixlwave:
@qike-ms Thanks for this, I haven't looked through it yet, but it's worth noting that #5451 doesn't mention adding a dedicated sheet for selection, but some kind of "mode" change on the bubble so you can make the selection directly in the timeline (just like iMessage does). Did you try that first before reaching this implementation?

I think the sheet reference was about the "Select" action being placed in the `TimelineItemMenuAction` sheet, not actually showing a sheet for the view.

PR #5932, opened 2026-07-27
Render galleries with non-visual attachments as a list

Velin92:
### What this does
- Adds `GalleryListView` (+ `GalleryDivider`): a vertical file-row list used when a gallery contains a non-visual attachment (file/audio), each row content-scanned with safe/scanning/unsafe states.
- Extracts a shared `FileTypeIconView` from `FileRoomTimelineView` (the rounded file-type icon badge), reused by the list rows.
- `GalleryRoomTimelineView` now picks list vs grid via `usesListLayout` (any item that isn't an image/video → list), with the caption below.

### PRs
- [x] the gallery timeline item + model (placeholder rendering)
- [x] Grid/tile views for image & video galleries
- [x] **This PR** — File-list view for galleries with non-visual attachments
- [ ] Full-screen preview when tapping a gallery item

### DIFF
--- a/ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FileRoomTimelineView.swift
+++ b/ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FileRoomTimelineView.swift
@@ -102,21 +102,32 @@ struct MediaFileRoomTimelineContent: View {
             .foregroundStyle(.compound.textPrimary)
             .lineLimit(2)
         } icon: {
-            Group {
-                if isScanning {
-                    ProgressView()
-                        .scaledFrame(size: CompoundIcon.Size.medium.value, relativeTo: .body)
-                } else {
-                    CompoundIcon(icon, size: .medium, relativeTo: .body)
-                        .foregroundColor(.compound.iconPrimary)
-                }
-            }
-            .scaledPadding(6)
-            .background(.compound.iconOnSolidPrimary,
-                        in: RoundedRectangle(cornerRadius: 4, style: .continuous))
+            FileTypeIconView(icon: icon, isScanning: isScanning)
         }
         .labelStyle(.custom(spacing: 8, alignment: .center))
         .padding(.horizontal, 4) // Add to the styler's padding of 8, as we use the default insets for the caption.
+    }
+}
+
+/// The rounded icon badge used as the leading accessory of a file/audio row, either showing the
+/// file-type icon or a scanning spinner.
+struct FileTypeIconView: View {
+    let icon: KeyPath<CompoundIcons, Image>
+    var isScanning = false
+    
+    var body: some View {
+        Group {
+            if isScanning {
+                ProgressView()
+                    .scaledFrame(size: CompoundIcon.Size.medium.value, relativeTo: .body)
+            } else {
+                CompoundIcon(icon, size: .medium, relativeTo: .body)
+                    .foregroundColor(.compound.iconPrimary)
+            }
+        }
+        .scaledPadding(6)
+        .background(.compound.iconOnSolidPrimary,
+                    in: RoundedRectangle(cornerRadius: 4, style: .continuous))
     }
 }
 
--- a/ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FileRoomTimelineView.swift
+++ b/ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FileRoomTimelineView.swift
@@ -119,15 +119,15 @@ struct FileTypeIconView: View {
         Group {
             if isScanning {
                 ProgressView()
-                    .scaledFrame(size: CompoundIcon.Size.medium.value, relativeTo: .body)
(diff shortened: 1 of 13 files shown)


ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryListView.swift
```diff
@@ -0,0 +1,133 @@
...
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Compound
+import SwiftUI
+
+/// Used for gallery messages that contain at least one attachment without a thumbnail
+/// (typically a document or audio). Renders a vertical list of file rows so each item is
+/// individually recognisable — the grid layout doesn't read well for non-visual content.
```
pixlwave: Can we flip this documentation around? I think it's generally more important what the thing is, rather than what it is/can be used for.

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FileRoomTimelineView.swift
```diff
@@ -102,24 +102,35 @@ struct MediaFileRoomTimelineContent: View {
...
+            FileTypeIconView(icon: icon, isScanning: isScanning)
         }
         .labelStyle(.custom(spacing: 8, alignment: .center))
         .padding(.horizontal, 4) // Add to the styler's padding of 8, as we use the default insets for the caption.
     }
 }
 
+/// The rounded icon badge used as the leading accessory of a file/audio row, either showing the
+/// file-type icon or a scanning spinner.
+struct FileTypeIconView: View {
```
pixlwave, line 114: Not a blocker, but personally I'm not sure if the re-use here is worthwhile as it couples them together and the designs could potentially diverge (e.g. in size) in the future.

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FileRoomTimelineView.swift
```diff
@@ -102,24 +102,35 @@ struct MediaFileRoomTimelineContent: View {
...
+                ProgressView()
+                    .scaledFrame(size: CompoundIcon.Size.medium.value, relativeTo: .body)
+            } else {
+                CompoundIcon(icon, size: .medium, relativeTo: .body)
+                    .foregroundColor(.compound.iconPrimary)
+            }
+        }
+        .scaledPadding(6)
+        .background(.compound.iconOnSolidPrimary,
+                    in: RoundedRectangle(cornerRadius: 4, style: .continuous))
```
pixlwave: That's the default style now, we can remove this.

```suggestion
                    in: RoundedRectangle(cornerRadius: 4))
```

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FileRoomTimelineView.swift
```diff
@@ -102,24 +102,35 @@ struct MediaFileRoomTimelineContent: View {
...
+    let icon: KeyPath<CompoundIcons, Image>
+    var isScanning = false
+    
+    var body: some View {
+        Group {
+            if isScanning {
+                ProgressView()
+                    .scaledFrame(size: CompoundIcon.Size.medium.value, relativeTo: .body)
+            } else {
+                CompoundIcon(icon, size: .medium, relativeTo: .body)
```
pixlwave: I think I said this in the last PR, but these 2 are the default values, so we don't necessarily need to pass them in (but looking at this a second time I can also see the explicit consistency with the size above could be considered nice).

```suggestion
                CompoundIcon(icon)
```

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FileRoomTimelineView.swift
```diff
@@ -102,24 +102,35 @@ struct MediaFileRoomTimelineContent: View {
...
+/// file-type icon or a scanning spinner.
+struct FileTypeIconView: View {
+    let icon: KeyPath<CompoundIcons, Image>
+    var isScanning = false
+    
+    var body: some View {
+        Group {
+            if isScanning {
+                ProgressView()
+                    .scaledFrame(size: CompoundIcon.Size.medium.value, relativeTo: .body)
```
pixlwave: I know this was moved, but we should probably use Compound here:

```suggestion
                    .scaledFrame(size: CompoundIcon.Size.medium.value, relativeTo: .compound.bodyLG)
```

pixlwave (approved):
👌

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FileRoomTimelineView.swift
Velin92: as of right they were literally the same, i guess is fine to keep for now and remove it later if there are divergences in the future.

PR #5934, opened 2026-07-27
Add an `oAuthClientURIPath` option to AppSettings.

pixlwave:
This allows us to use unique client URIs for OAuth which MAS can use to distinguish Element on iOS from Web & Android. It is optional as forks would not necessarily need this. The value is `nil` for now until we have confirmation of the URL we can use on element.io


stefanceriu (approved):
👍

PR #5936, opened 2026-07-27
Open a full-screen preview when tapping a gallery item

Velin92:
Part 4 (final) of the gallery rendering stack (follows #5932).

### What this does
- Tapping a gallery tile/row opens the existing media preview, scoped to that gallery's attachments and paging between them.
- Introduces a typed `MediaPreviewItemID` (replacing the stringly-typed preview identifier) and a gallery-scoped `TimelineMediaPreviewDataSource`; the tap routes through `TimelineInteractionHandler` → `TimelineViewModel`.
- Non-previewable (`.other`) items are skipped without shifting the tapped index.

### PRs
- [x] the gallery timeline item + model (placeholder rendering)
- [x] Grid/tile views for image & video galleries
- [x] File-list view for galleries with non-visual attachments
- [x] **This PR** — Full-screen preview when tapping a gallery item

### DIFF
--- a/ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
+++ b/ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
@@ -216,13 +216,6 @@ enum TimelineMediaPreviewItem: Equatable {
             }
         }
         
-        var galleryItem: GalleryItem? {
-            switch content {
-            case .timelineItem: nil
-            case .galleryItem(_, let item): item
-            }
-        }
-        
         var fileHandle: MediaFileHandleProxy? {
             didSet { updatePreviewItemValues() }
         }
@@ -299,55 +292,63 @@ enum TimelineMediaPreviewItem: Equatable {
         enum Kind { case image, video, file }
         
         var kind: Kind {
-            if let galleryItem {
-                switch galleryItem {
-                case .image: return .image
-                case .video: return .video
-                case .audio, .file, .other: return .file
+            switch content {
+            case .galleryItem(_, let item):
+                switch item {
+                case .image: .image
+                case .video: .video
+                case .audio, .file, .other: .file
                 }
-            }
-            return switch timelineItem {
-            case is ImageRoomTimelineItem: .image
-            case is VideoRoomTimelineItem: .video
-            default: .file
+            case .timelineItem(let timelineItem):
+                switch timelineItem {
+                case is ImageRoomTimelineItem: .image
+                case is VideoRoomTimelineItem: .video
+                default: .file
+                }
             }
         }
         
         var mediaSource: MediaSourceProxy? {
-            if let galleryItem {
-                return galleryItem.mediaSource
-            }
-            return switch timelineItem {
-            case let audioItem as AudioRoomTimelineItem: audioItem.content.source
-            case let fileItem as FileRoomTimelineItem: fileItem.content.source
-            case let imageItem as ImageRoomTimelineItem: imageItem.content.imageInfo.source
-            case let videoItem as VideoRoomTimelineItem: videoItem.content.videoInfo.source
-            default: nil
+            switch content {
+            case .galleryItem(_, let item):
+                item.mediaSource
+            case .timelineItem(let timelineItem):
+                switch timelineItem {
+                case let audioItem as AudioRoomTimelineItem: audioItem.content.source
+                case let fileItem as FileRoomTimelineItem: fileItem.content.source
+                case let imageItem as ImageRoomTimelineItem: imageItem.content.imageInfo.source
+                case let videoItem as VideoRoomTimelineItem: videoItem.content.videoInfo.source
+                default: nil
+                }
             }
         }
         
         var thumbnailMediaSource: MediaSourceProxy? {
-            if let galleryItem {
-                return galleryItem.thumbnailSource
-            }
-            return switch timelineItem {
-            case let fileItem as FileRoomTimelineItem: fileItem.content.thumbnailSource
-            case let imageItem as ImageRoomTimelineItem: imageItem.content.thumbnailInfo?.source
-            case let videoItem as VideoRoomTimelineItem: videoItem.content.thumbnailInfo?.source
-            default: nil
+            switch content {
+            case .galleryItem(_, let item):
+                item.thumbnailSource
+            case .timelineItem(let timelineItem):
+                switch timelineItem {
+                case let fileItem as FileRoomTimelineItem: fileItem.content.thumbnailSource
+                case let imageItem as ImageRoomTimelineItem: imageItem.content.thumbnailInfo?.source
+                case let videoItem as VideoRoomTimelineItem: videoItem.content.thumbnailInfo?.source
+                default: nil
+                }
             }
         }
         
         var filename: String? {
-            if let galleryItem {
-                return galleryItem.filename
-            }
-            return switch timelineItem {
-            case let audioItem as AudioRoomTimelineItem: audioItem.content.filename
-            case let fileItem as FileRoomTimelineItem: fileItem.content.filename
-            case let imageItem as ImageRoomTimelineItem: imageItem.content.filename
-            case let videoItem as VideoRoomTimelineItem: videoItem.content.filename
-            default: nil
+            switch content {
+            case .galleryItem(_, let item):
+                item.filename
+            case .timelineItem(let timelineItem):
+                switch timelineItem {
+                case let audioItem as AudioRoomTimelineItem: audioItem.content.filename
+                case let fileItem as FileRoomTimelineItem: fileItem.content.filename
+                case let imageItem as ImageRoomTimelineItem: imageItem.content.filename
+                case let videoItem as VideoRoomTimelineItem: videoItem.content.filename
+                default: nil
+                }
             }
         }
         
@@ -356,15 +357,17 @@ enum TimelineMediaPreviewItem: Equatable {
         }
         
         private var expectedFileSize: UInt? {
-            if let galleryItem {
-                return galleryItem.fileSize
-            }
-            return switch timelineItem {
-            case let audioItem as AudioRoomTimelineItem: audioItem.content.fileSize
-            case let fileItem as FileRoomTimelineItem: fileItem.content.fileSize
-            case let imageItem as ImageRoomTimelineItem: imageItem.content.imageInfo.fileSize
-            case let videoItem as VideoRoomTimelineItem: videoItem.content.videoInfo.fileSize
-            default: nil
+            switch content {
+            case .galleryItem(_, let item):
+                item.fileSize
+            case .timelineItem(let timelineItem):
+                switch timelineItem {
+                case let audioItem as AudioRoomTimelineItem: audioItem.content.fileSize
+                case let fileItem as FileRoomTimelineItem: fileItem.content.fileSize
+                case let imageItem as ImageRoomTimelineItem: imageItem.content.imageInfo.fileSize
+                case let videoItem as VideoRoomTimelineItem: videoItem.content.videoInfo.fileSize
+                default: nil
+                }
             }
         }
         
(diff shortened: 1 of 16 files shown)


ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
```diff
@@ -58,6 +59,40 @@ class TimelineMediaPreviewDataSource: NSObject, QLPreviewControllerDataSource {
         self.paginationState = paginationState
     }
     
+    /// Builds a data source scoped to a single gallery's previewable attachments.
+    /// Used when the user taps a tile inside a gallery message — paging is local to that
+    /// gallery and there's no timeline pagination to drive.
```
pixlwave, line 66: We should probably document the other init too to distinguish the behaviour.

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
```diff
@@ -58,6 +59,40 @@ class TimelineMediaPreviewDataSource: NSObject, QLPreviewControllerDataSource {
...
+        let items = galleryItem.content.items
+        let media = items.compactMap { item in
+            TimelineMediaPreviewItem.Media(galleryParent: galleryItem, item: item)
+        }
+        
+        // `initialIndex` is an index into the full items array, but non-previewable items (`.other`)
+        // are dropped from `media`, so map it across by identity to avoid opening the wrong attachment.
+        let mediaIndex = items.indices.contains(initialIndex)
+            ? (media.firstIndex { $0.id == .galleryItem(items[initialIndex].id) } ?? 0)
+            : 0
```
pixlwave: Would it not make more sense to show a "cannot preview this item" or "unknown type" message/warning/error instead of skipping it?

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
```diff
@@ -58,6 +59,40 @@ class TimelineMediaPreviewDataSource: NSObject, QLPreviewControllerDataSource {
...
+            initialItemIndex = mediaIndex
+            currentItem = .media(media[mediaIndex])
+        } else {
+            // Fall back to a synthetic placeholder for empty galleries — shouldn't happen in practice.
+            previewItems = [.init(timelineItem: galleryItem)]
+            initialItemIndex = 0
+            currentItem = .media(previewItems[0])
+        }
+        
+        // No surrounding pagination — the gallery is the whole world.
```
Velin92: This comment is funny.

pixlwave: 🤣

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
```diff
@@ -172,41 +209,66 @@ enum TimelineMediaPreviewItem: Equatable {
             didSet { updatePreviewItemValues() }
         }
         
+        /// When non-nil, this preview item represents a single attachment from a gallery message.
+        /// Property accessors (mediaSource, filename, …) read from the gallery item instead of the
+        /// parent timeline item — but `timelineItem.id` still points at the parent event so menu
+        /// actions (redact, reply, …) target the gallery as a whole.
```
pixlwave: Might just be me, but this feels a bit weird to me. If I'm looking at a gallery item and choose to delete that gallery item, I would expect *only* that one to be removed, not the whole gallery. I don't think it's blocking, but we should probably get design/product input on this behaviour.

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
```diff
@@ -172,41 +209,66 @@ enum TimelineMediaPreviewItem: Equatable {
...
+                resolved = imageRoomTimelineItem
             case .video(let videoRoomTimelineItem):
-                timelineItem = videoRoomTimelineItem
+                resolved = videoRoomTimelineItem
             default:
                 return nil
             }
+            timelineItem = resolved
+            galleryItem = nil
+            id = Media.derivedID(for: resolved)
```
pixlwave: Can't we just use `timelineItem` here (now that it's set) and get rid of the `resolved`?

Otherwise I would vote to rename `resolved` to `timelineItem` and do `self.timelineItem = timelineItem` above (but I don't think that's necessary)

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
```diff
@@ -172,41 +209,66 @@ enum TimelineMediaPreviewItem: Equatable {
...
-        /// We're identifying items by this to ensure that all matching is made using only this part of the identifier. This is
+        /// We identify items by this to ensure that all matching is made using only this part of the identifier. This is
         /// because the unique ID will be different across timelines so when the initial item comes from a regular timeline and
         /// we build a filtered timeline to fetch the other media items, it is impossible to match by the `TimelineItemIdentifier`.
-        var id: TimelineItemIdentifier.EventOrTransactionID {
+        private static func derivedID(for timelineItem: EventBasedMessageTimelineItemProtocol) -> MediaPreviewItemID {
             guard let id = timelineItem.id.eventOrTransactionID else { fatalError("Virtual items cannot be previewed.") }
-            return id
+            return .timelineItem(id)
         }
```
pixlwave: I think this could be a convenience init on MediaPreviewItemID rather than a private static func here?

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
```diff
@@ -289,21 +346,20 @@ enum TimelineMediaPreviewItem: Equatable {
...
-                fileItem.content.fileSize
-            case let imageItem as ImageRoomTimelineItem:
-                imageItem.content.imageInfo.fileSize
-            case let videoItem as VideoRoomTimelineItem:
-                videoItem.content.videoInfo.fileSize
-            default:
-                nil
+            if galleryItem != nil {
+                return nil
+            } // The SDK doesn't surface individual gallery item sizes.
```
pixlwave: Again not blocking, but can it? Or is this a spec limitation?

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
```diff
@@ -289,21 +346,20 @@ enum TimelineMediaPreviewItem: Equatable {
...
+            case let audioItem as AudioRoomTimelineItem: audioItem.content.fileSize
+            case let fileItem as FileRoomTimelineItem: fileItem.content.fileSize
+            case let imageItem as ImageRoomTimelineItem: imageItem.content.imageInfo.fileSize
+            case let videoItem as VideoRoomTimelineItem: videoItem.content.videoInfo.fileSize
+            default: nil
             }
         }
         
         var hasCaption: Bool {
+            // Captions live on the gallery itself, not on individual items.
```
pixlwave: This comment only really makes sense in the context of this PR. More generally we could put:

```suggestion
            // No need to special-case gallery items here, captions live on the gallery event itself
```

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
```diff
@@ -172,41 +209,66 @@ enum TimelineMediaPreviewItem: Equatable {
             didSet { updatePreviewItemValues() }
         }
         
+        /// When non-nil, this preview item represents a single attachment from a gallery message.
+        /// Property accessors (mediaSource, filename, …) read from the gallery item instead of the
+        /// parent timeline item — but `timelineItem.id` still points at the parent event so menu
+        /// actions (redact, reply, …) target the gallery as a whole.
+        let galleryItem: GalleryItem?
```
pixlwave: Part of me would be interested to see what this would look like it `timelineItem` and `galleryItem` were combined into an enum here… But I can't decide if it would simplify the implementation, or merely move things around.

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewViewModel.swift
```diff
@@ -78,6 +78,36 @@ class TimelineMediaPreviewViewModel: TimelineMediaPreviewViewModelType {
             .store(in: &cancellables)
     }
     
+    /// Initialises the preview scoped to a single gallery's attachments. The data source is
+    /// built from the gallery's items directly and isn't kept in sync with the underlying
+    /// timeline — gallery contents don't change without the event being replaced or redacted.
```
pixlwave, line 89: Same comment, we should document both inits to help distinguish.

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
```diff
@@ -708,6 +710,8 @@ class TimelineViewModel: TimelineViewModelType, TimelineViewModelProtocol {
             
             let mediaPreviewViewModel = makeMediaPreviewViewModel(item: item, timelineViewModelKind: timelineViewModelKind)
             actionsSubject.send(.displayMediaPreview(mediaPreviewViewModel))
+        case .displayGalleryPreview(let galleryItem, let initialIndex):
+            displayGalleryPreview(galleryItem: galleryItem, initialIndex: initialIndex)
```
pixlwave: I'm confused, we have `.displayGalleryPreview` inside of `handleMediaTapped` and we have `.galleryItemTapped` separately. Shouldn't we pick one path or the other?

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
```diff
@@ -58,6 +59,42 @@ class TimelineMediaPreviewDataSource: NSObject, QLPreviewControllerDataSource {
...
+        // `initialIndex` is an index into the full items array, but non-previewable items (`.other`)
+        // are dropped from `media`, so map it across by identity to avoid opening the wrong attachment.
+        var mediaIndex = 0
+        if items.indices.contains(initialIndex),
+           let index = media.firstIndex(where: { $0.id == .galleryItem(items[initialIndex].id) }) {
+            mediaIndex = index
+        }
+        
+        if media.indices.contains(mediaIndex) {
+            previewItems = media
```
pixlwave, line 75:
```suggestion
            // We set the entire gallery here up front.
            previewItems = media
```

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
```diff
@@ -58,6 +59,42 @@ class TimelineMediaPreviewDataSource: NSObject, QLPreviewControllerDataSource {
...
+            // Fall back to a synthetic placeholder for empty galleries — shouldn't happen in practice.
+            previewItems = [.init(timelineItem: galleryItem)]
+            initialItemIndex = 0
+            currentItem = .media(previewItems[0])
+        }
+        
+        // No surrounding pagination — the gallery is the whole world.
+        backwardPadding = 0
+        forwardPadding = 0
+        self.paginationState = paginationState
```
pixlwave:
```suggestion
        // And we disable any use of the timeline by configuring the data source as though everything has paginated.
        backwardPadding = 0
        forwardPadding = 0
        self.paginationState = paginationState
```

pixlwave: Also, question: It seems wrong that we would set the external pagination state here as we don't paginate anyway. Does it break anything if we remove that parameter and hard code `.init(backward: .endReached, forward: .endReached)` here?

pixlwave (approved):
Quite a few comments about *how* we do this, but overall it seems decent to me 👍

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
Velin92: I guess QL also handles non previewable items with a default  OS message, so this could be done quite easily, I guess is however more of an edge case but if you feel is worth doing, we can add it.

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
Velin92: Android does the same too and the reason is that being the gallery 1 single matrix event, we can only fully redact it. Maybe we could try to edit it to delete a single item, but I guess how we do that is as you said a product question. I would say let's leave it as it is now, since we only care about rendering and previewing a gallery we can flag this to the product and have it handled in the future.

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
Velin92: I mean it will require some moving around but I also didn't really like the fact that this could be nil and unused in the non gallery context. Decidedo to not bother too much but if you have the same feeling, let's give it a try

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
pixlwave: Yeah makes sense. And sorry, to be clear I wasn't specifically thinking we should go full hog and support deleting of just one item, but that we could just hide actions that would be confusing for gallery items. Again, product/design decision for the future but there's a simple solution available for them to pick too.

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
Velin92: Oh wait... this was actualy later added to the SDK, so we can indeed get the sizes.

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
Velin92: actually the handleMedia path is not really required, I guess we can get rid of it.

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
```diff
@@ -205,15 +198,38 @@ enum TimelineMediaPreviewItem: Equatable {
...
-        /// parent timeline item — but `timelineItem.id` still points at the parent event so menu
-        /// actions (redact, reply, …) target the gallery as a whole.
-        let galleryItem: GalleryItem?
+        var timelineItem: EventBasedMessageTimelineItemProtocol {
+            content.timelineItem
+        }
+        
+        var galleryItem: GalleryItem? {
+            content.galleryItem
+        }
```
pixlwave: Ah, my assumption was that we wouldn't have these 2 anymore. But this definitely looks way better to me anyway, thanks 👌

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
```diff
@@ -346,9 +364,9 @@ enum TimelineMediaPreviewItem: Equatable {
...
         private var expectedFileSize: UInt? {
-            if galleryItem != nil {
-                return nil
-            } // The SDK doesn't surface individual gallery item sizes.
+            if let galleryItem {
+                return galleryItem.fileSize
+            }
             return switch timelineItem {
             case let audioItem as AudioRoomTimelineItem: audioItem.content.fileSize
             case let fileItem as FileRoomTimelineItem: fileItem.content.fileSize
```
pixlwave: Technically now we can combine these into a single `switch` on `content` for all of the properties in this file.

pixlwave (approved):

PR #5927, opened 2026-07-23
WIP: support for waitingForAuth continuation sender during Link new device

hughns:
For use with https://github.com/matrix-org/matrix-rust-sdk/pull/6711

### Pull Request Checklist

- [ ] I read the [contributing guide](https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md).
- [ ] I am aware of the [etiquette](https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md#etiquette).
- [ ] Pull request contains a [changelog label](https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md#changelog).
- [ ] This PR has been made with the help of an LLM.

**UI changes have been tested with:**
- [ ] iPhone and iPad simulators in portrait and landscape orientations.
- [ ] Dark mode enabled and disabled.
- [ ] Various sizes of dynamic type.
- [ ] Voiceover enabled.

PR #5938, opened 2026-07-28
update sdk to 26-7-28

Velin92:
name says it all


pixlwave (approved):

pixlwave:
I'll handle #5927 after the RC, given that the Link New Device flow is behind a feature flag.

PR #5939, opened 2026-07-28
Fix for some flaky tests part 2

Velin92:
Our tests have become significantly more consistent, but we still sometimes rely on multiple runs, so had AI go through some of the actions we have run recently that used more than 1 run to try to fix some of last flaky tests. Ideally we want to never rely on multiple runs, since they might end up eating CPU time + make a CI test fail due to timeout.

This is an experiment


stefanceriu (approved):
Looks good, let's try it out! 👍

PR #5942, opened 2026-07-29
Render replies to gallery items

Velin92:
Implemented (with also content scanning) a way to nicely render reply views for galleries


ElementX/Sources/Screens/Timeline/View/Replies/TimelineReplyView.swift
```diff
@@ -159,6 +157,48 @@ struct TimelineReplyView: View {
...
+        
+        private var icon: ReplyView.Icon? {
+            guard isMediaGallery, let item = content.items.first else {
+                return .init(kind: .icon(\.attachment))
+            }
+            
+            return switch item {
+            case .image(_, let itemContent): .init(kind: .mediaSource(itemContent.thumbnailInfo?.source ?? itemContent.imageInfo.source))
+            case .video(_, let itemContent): itemContent.thumbnailInfo.map { .init(kind: .mediaSource($0.source)) }
+            case .audio, .file, .other: nil
```
stefanceriu: Shouldn't we use the default icons here? 🤔

stefanceriu: Oh wait, it's not a gallery unless it only has images and videos?

ElementX/Sources/Screens/Timeline/View/Replies/TimelineReplyView.swift
```diff
@@ -75,6 +77,7 @@ struct TimelineReplyView: View {
         case .file(let content): return content.thumbnailSource
         case .image(let content): return content.thumbnailInfo?.source
         case .video(let content): return content.thumbnailInfo?.source
+        case .gallery(let content): return content.items.first?.thumbnailSource
```
stefanceriu, line 80: Hyst for the record I'm not a pariticularly big fan of these helpers (thumbnailSource) and the asymetry with the other content types

stefanceriu (approved):
lgtm 👍

ElementX/Sources/Screens/Timeline/View/Replies/TimelineReplyView.swift
Velin92: Yeah but actually I think you are right... Android behaves by just having this reply view when there are only medias in the gallery and no files... but honestly I think this is not the intended behaviour, we should just display the thumbnail/icon of the first element regardless, let me change this

PR #5943, opened 2026-07-29
Allow previews in the timeline to include filtered galleries

Velin92:
When a gallery is opened directly from the timeline, the preview should be self contained to the gallery.
However when the preview is opened outside the gallery, galleries should be included too when swiping between media, and the gallery should also be filtered following the same pattern used for the opening media (images and videos, or files and audios), essentially flattening the gallery and considering each single media element as a timeline item.

### DIFF
--- a/ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
+++ b/ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
@@ -36,7 +36,7 @@ class TimelineMediaPreviewDataSource: NSObject, QLPreviewControllerDataSource {
     var paginationState: TimelinePaginationState
     
     /// The gallery attachments to include, so that a gallery only contributes the media being browsed.
-    private let galleryFilters: [GalleryFilter]?
+    private let allowedGalleryItemTypes: [TimelineAllowedGalleryItemType]?
     
     /// Builds a data source spanning every previewable attachment in the timeline, paginating
     /// as the user swipes past the loaded range. Used when tapping a standalone media message.
@@ -44,10 +44,10 @@ class TimelineMediaPreviewDataSource: NSObject, QLPreviewControllerDataSource {
          initialItem: EventBasedMessageTimelineItemProtocol,
          initialPadding: Int = 100,
          paginationState: TimelinePaginationState,
-         galleryFilters: [GalleryFilter]? = nil) {
-        self.galleryFilters = galleryFilters
+         allowedGalleryItemTypes: [TimelineAllowedGalleryItemType]? = nil) {
+        self.allowedGalleryItemTypes = allowedGalleryItemTypes
         
-        previewItems = itemViewStates.flatMap { TimelineMediaPreviewItem.Media.items(from: $0, galleryFilters: galleryFilters) }
+        previewItems = itemViewStates.flatMap { $0.previewableMedia(allowedGalleryItemTypes: allowedGalleryItemTypes) }
         
         let initialPreviewID = TimelineMediaPreviewItem.Media(timelineItem: initialItem).id
         if let initialItemArrayIndex = previewItems.firstIndex(where: { $0.id == initialPreviewID }) {
@@ -72,7 +72,7 @@ class TimelineMediaPreviewDataSource: NSObject, QLPreviewControllerDataSource {
     /// gallery and there's no timeline pagination to drive.
     init(galleryItem: GalleryRoomTimelineItem,
          initialIndex: Int) {
-        galleryFilters = nil // All of the gallery's attachments are shown.
+        allowedGalleryItemTypes = nil // All of the gallery's attachments are shown.
         
         let media = galleryItem.content.items.map { item in
             TimelineMediaPreviewItem.Media(galleryParent: galleryItem, item: item)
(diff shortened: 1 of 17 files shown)


ElementX/Sources/Services/Timeline/TimelineItems/Items/Messages/GalleryRoomTimelineItemContent.swift
```diff
@@ -25,6 +25,16 @@ nonisolated struct GalleryItemID: Hashable {
     let mediaIndex: Int
 }
 
+/// The kinds of gallery attachment that a screen includes, mirroring the message types its timeline is
+/// filtered to. The SDK can only filter whole events, so a gallery's attachments are filtered by these
+/// client side. Unlike a message type there's no gallery case, as a gallery can't contain another one.
+nonisolated enum GalleryFilter {
```
pixlwave: Sorry, this is a pretty big change to this PR, but based on the documentation, the name of this feels fundamentally wrong to me. We already have `TimelineAllowedMessageType` which does the same thing for timelines, so this type should probably line up with that.

I would suggest something like `TimelineAllowedGalleryItemType` with `gallery.allowedItemTypes`, `allowedMessageTypes.allowedGalleryItemTypes`, `timelineController.allowedGalleryItemTypes` etc.

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
```diff
@@ -93,17 +101,17 @@ class TimelineMediaPreviewDataSource: NSObject, QLPreviewControllerDataSource {
...
     func updatePreviewItems(itemViewStates: [RoomTimelineItemViewState]) {
-        let newItems: [TimelineMediaPreviewItem.Media] = itemViewStates.compactMap { itemViewState in
-            guard let newItem = TimelineMediaPreviewItem.Media(roomTimelineItemViewState: itemViewState) else { return nil }
-            
-            // If an item already exists use that instead to preserve the file handle, download error etc.
-            if let oldItem = previewItems.first(where: { $0.id == newItem.id }) {
-                oldItem.content = newItem.content
-                return oldItem
+        let newItems: [TimelineMediaPreviewItem.Media] = itemViewStates
+            .flatMap { TimelineMediaPreviewItem.Media.items(from: $0, galleryFilters: galleryFilters) }
```
pixlwave: It would read clearer if this `items()` was a (private?) extension on `RoomTimelineItemViewState` so that this could be something like

```
.flatMap { $0.previewableMedia(allowedGalleryItemTypes: allowedGalleryItemTypes) }
```

ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
```diff
@@ -607,7 +614,10 @@ class TimelineInteractionHandler {
         case .pinned:
             newTimelineFocus = .pinned
             newTimelinePresentation = .pinnedEventsScreen
-        case .media, .thread:
+        case .thread(let rootEventID):
+            newTimelineFocus = .thread(eventID: rootEventID)
+            newTimelinePresentation = .roomScreenThread
```
pixlwave, line 619: 👏

ElementX/Sources/Services/Timeline/TimelineProxy.swift
```diff
@@ -663,15 +663,19 @@ private extension MatrixRustSDK.PollKind {
...
-        for item in self {
-            if case let .event(eventTimelineItem) = item {
-                if eventTimelineItem.id.uniqueID == id.uniqueID {
-                    return eventTimelineItem.item
-                }
-            }
+    /// The event matching the given identifier, found by the event itself as that is the only part of
+    /// the identifier that can be compared across timelines. Identifiers that don't represent an event
+    /// fall back to the unique ID.
+    func firstEventTimelineItem(matching id: TimelineItemIdentifier) -> EventTimelineItem? {
```
pixlwave, line 669: Nice! 😎

ElementX/Sources/Services/Timeline/TimelineItems/Items/Messages/GalleryRoomTimelineItemContent.swift
```diff
@@ -152,13 +163,24 @@ nonisolated enum GalleryItem: Hashable, Identifiable {
...
+        }
+    }
 }
 
 // MARK: - Mocks
 
 nonisolated extension GalleryItemID {
     static func mock(_ index: Int) -> GalleryItemID {
-        .init(timelineItemUniqueID: .init("gallery-mock"), mediaIndex: index)
+        .init(timelineItemID: .virtual(uniqueID: .init("gallery-mock")), mediaIndex: index)
```
pixlwave: We shouldn't use `.virtual` here, that's a lie. Can't we use the existing mock ID here instead?


```suggestion
        .init(timelineItemID: .randomEvent), mediaIndex: index)
```

Otherwise we should use `event(uniqueID: .init("gallery-mock"), eventOrTransactionID: .eventID(UUID().uuidString))` if the `gallery-mock` is important, but you wouldn't be able to put 2 mocks in the same timeline which could be annoying.

ElementX/Sources/Services/Timeline/TimelineItems/RoomTimelineItemFactory.swift
```diff
@@ -149,7 +149,7 @@ nonisolated struct RoomTimelineItemFactory: RoomTimelineItemFactoryProtocol {
         case .location(let content):
             .location(buildLocationTimelineItemContent(content))
         case .gallery(let content):
-            .gallery(buildGalleryTimelineItemContent(content, uniqueID: .init(senderID)))
+            .gallery(buildGalleryTimelineItemContent(content, timelineItemID: .virtual(uniqueID: .init(senderID))))
```
pixlwave: Same here:

```suggestion
            .gallery(buildGalleryTimelineItemContent(content, timelineItemID: .randomEvent))
```

pixlwave (approved):
Fundamentally this looks decent, but I have a big comment about the naming, sorry 😕

PR #5945, opened 2026-07-29
Escape search queries and keep results aligned with the SDK's diff stream

hayaksi1:
### Problem

Three bugs in `SearchServiceProxy`, all reachable today through the Messages tab behind the
`globalSearchEnabled` developer flag:

- `handleResultUpdates` applies the SDK's positional diffs to a local list that `makeResult` has
  already filtered, since it returns `nil` for content it cannot render. From the first dropped
  result onward every index addresses the wrong row, and `popFront`, `popBack` and an over-long
  `truncate` do not merely misbehave — they trap.
- `setQuery` never resets pagination state, so the previous query's `endReached` survives into the
  new one. A cursor that has never run a query reports `endReached = true`.
- The raw query reaches tantivy's parser: `"unclosed` or `(` fails the whole search, `AND`/`OR`/`TO`
  are read as operators, and the parser being OR-by-default, a multi-word query matches documents
  containing *any* word rather than all of them.

### Fix

- Map every result instead of dropping some — content with no view renders as the existing
  unsupported-event placeholder — and move filtering to the exposure boundary, so the local list
  stays index-parallel with the SDK's.
- Reset pagination state in `setQuery`, and return a `Result` from `paginate()` so a caller can
  tell a finished list from a failed page. It is `@discardableResult`; the existing call site is
  unchanged.
- Escape tantivy syntax in a single forward pass, marking every token containing a letter or digit
  as required (`+`). Tokens that index no terms at all, such as `:)`, stay optional.

Two additions a room-scoped surface will need, both inert for the shipped tab: an optional `roomID`
on `SearchServiceProxy`, applied only at the exposure boundary, and
`ClientProxy.makeSearchService(roomID:)`.

**This changes what the Messages tab returns**, deliberately: multi-word queries are now AND rather
than OR, and unrenderable results appear as a placeholder row instead of vanishing — which is what
desynced the list.

### Tests

19 Swift Testing cases. `TantivyEscapingTests` pins the escaper against the tantivy grammar in the
SDK we ship (`26.07.28`). `SearchServiceProxyTests` adds one regression per bug: a positional
update following an unsupported result, the `truncate` and `popFront` traps, the pagination reset,
escaping at the boundary, `paginate` failure propagation, and a room-filter case proving the filter
leaves index parallelism intact.

### Notes for reviewers

138 lines of non-generated, non-test source; the rest is tests and regenerated mocks. The
behaviour changes are described above.

No file overlap with #5910, which now has a `Result` from `paginate()` to surface as well.

Could a maintainer add the `pr-bugfix` label when you get a chance? Danger fails without one and
I cannot set it from a fork.

Relates to element-hq/element-x-ios#5829 — added the Messages mode and the index this hardens.
Relates to element-hq/element-x-ios#5910 — the set-query error indicator.
Relates to matrix-org/matrix-rust-sdk#5350 — the SDK-side full-text search work.

### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX.xcodeproj/project.pbxproj
+++ b/ElementX.xcodeproj/project.pbxproj
@@ -231,6 +231,7 @@
 		2435DCCBD4513B9FD053CAC9 /* target.yml in Resources */ = {isa = PBXBuildFile; fileRef = DD3C65634A34467CB407A061 /* target.yml */; };
 		244407B18B2F2D6466BA5961 /* RoomChangeRolesScreenCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 82DFA1B7B088D033E0794B82 /* RoomChangeRolesScreenCoordinator.swift */; };
 		2447FADEF13225BB6227B977 /* VerificationBadge.swift in Sources */ = {isa = PBXBuildFile; fileRef = D1D97BAF04AA150C0EF03021 /* VerificationBadge.swift */; };
+		249D15AAF13C7ED37BC504E0 /* TantivyEscapingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = C7F62FB4D78056B2BB3AFA79 /* TantivyEscapingTests.swift */; };
 		24A1BBADAC43DC3F3A7347DA /* AnalyticsPromptScreen.swift in Sources */ = {isa = PBXBuildFile; fileRef = E53BFB7E4F329621C844E8C3 /* AnalyticsPromptScreen.swift */; };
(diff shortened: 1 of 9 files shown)


stefanceriu:
You seem to be right about the first but (albeit Inconsequential) but I'm not so sure about the other 2. If they are indeed bugs they should be fixed on the Rust side. 
Either way, one commit for 3 different fixes plus a bunch of random changes will not do. Please re-read our [contributing guide etiquette](https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md#etiquette) and try again.

PR #5946, opened 2026-07-29
Add a room message search screen reachable from the room top bar

hayaksi1:
> **Depends on #5945**, which has not merged yet, so this branch carries that PR's commit too. It
> is not mergeable until #5945 lands, at which point I will rebase.

Raising this against the direction in element-meta#3252 rather than assuming it is wanted — happy
to rework, split further, or drop it if it cuts across work already under way.

### Problem

element-x-ios has a global Search tab with a Messages mode behind the `globalSearchEnabled`
developer flag, but no way to search inside a single room. Element X Android ships a magnifier in
the room top bar that does exactly that.

### Fix

A standard MVVM-C screen, `RoomMessageSearchScreen`, reachable from a magnifier in the room top bar
that renders only when the new **disabled** flag `roomMessageSearchEnabled` is on. It owns one
cursor from `clientProxy.makeSearchService(roomID:)`, debounces at 250 ms, and runs a room-scoped
auto-pagination walk: room scoping filters a globally ranked result set, so an empty room-scoped
list over a non-exhausted index means "keep walking", not "no results". Tapping a result pops back
to the room focused on that event.

Three supporting changes:

- **The search index is now attached at login.** Session restore already attached it, but a freshly
  logged-in client did not, so a new login had no index for its entire first run. This also fixes
  the shipped Messages tab for fresh logins.
- Two untranslated strings using **Android's exact keys** (`screen_message_search_title`,
  `screen_message_search_error`) so the shared Localazy project converges once they are imported.
- Text and media preview derivation moves out of `SearchScreenMessage` into `TimelineEventContent`,
  and `SearchScreenMediaPreviewView` into its own file, so the two result rows cannot drift. Pure
  refactor.

### Tests

`RoomMessageSearchScreenViewModelTests` — 13 Swift Testing cases, porting the applicable ones from
Android's `MessageSearchPresenterTest`: the debounce, a failed query never shown as empty, clearing
returning to the initial state, the walk paginating while empty and stopping once the room has a
hit, list-end visibility pulling a page, `endReached` and a failed page each stopping the walk, and
selecting a result.

Previews cover all six states and conform to `TestablePreview`; 24 snapshots are recorded and
`roomMessageSearchScreen()` passes against them. `searchScreen()` — the shipped tab's own snapshot
test — passes before and after, which is the evidence for the "pure refactor" claim above.

### Notes for reviewers

**On the diff size you will see.** This depends on #5945, which has not merged, so GitHub shows
**1,672 additions across 60 files** — that total includes #5945's commit and the 24 recorded
snapshots (LFS pointers, 72 lines). This PR's own contribution is **857 lines** of non-generated,
non-test source, and the diff drops to that once #5945 lands and this is rebased. Reviewing the
last two commits only gives you this PR.

857 is still well over the 200 guideline, and most of it is one screen. Split points if you would
prefer them: attach-at-login + flag + strings (~15), the preview extraction (~139), the toolbar and
flow wiring (~75), leaving the screen itself at ~629. Say the word and I will resplit.

Depends on #5945 for `makeSearchService(roomID:)`, the room filter, and `paginate()` returning a
`Result`.

Nothing in this stack turns the feature on — every flag stays off, so there is deliberately no
"enable the feature" PR at the end.

Attach-at-login has no unit test: `AuthenticationClientFactory` calls the static
`ClientBuilder.baseBuilder` with no injection seam, so it is verified manually. Adding a
builder-factory seam is the fix if you want one.

The two string keys need a Localazy import by the team.

cannot set it from a fork.

Relates to element-hq/element-x-ios#5620 — the earlier room-search PR; same surface, different
implementation.
Relates to element-hq/element-x-ios#5829 — added the Messages mode and the index this screen queries.
Relates to element-hq/element-meta#3252 — *[Story] Search for messages*.
Relates to element-hq/element-meta#2004 — the room-scoped request this answers (X-Needs-Design).
Relates to element-hq/element-meta#1044 — *Search in encrypted rooms*.

### Pull Request Checklist

- [ ] Voiceover enabled.


stefanceriu:
Nowhere in https://github.com/element-hq/element-meta/issues/3252 does it say we agreed to introduce a per room search and you clearly didn't read or respect our [contributing guide etiquette](https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md#etiquette). I have no option than to close this PR.

PR #5947, opened 2026-07-29
Backfill room history into the search index in the background

hayaksi1:
### Problem

The local message search index only ever sees events that arrive *after* it was attached, so
anything older is silently unsearchable. There is no path today that feeds it history.

### Fix

Walk rooms backwards through timeline back-pagination so history flows through the SDK's event
cache and reaches the indexer. Everything is behind a new developer flag,
**`searchBackfillEnabled`, default off**.

- **Queue** — the static room summary provider, most-recently-active first, excluding spaces,
  tombstoned rooms, invites and knocks, and rooms that never received a message. Capped at 100,
  which is element-meta#3252's "the 100 most recent rooms".
- **Walk** — per room, back-paginate until the SDK reports the start of the timeline or a budget
  stops it. Budgets are counted in pages and rooms, never events.
- **Cursor** — a JSON file in the session data directory, written after every room, so process
  death costs at most one room. It lives there rather than in `AppSettings` because every path
  that ends a session already deletes that directory; a cursor outliving a logout would claim
  rooms were swept using the previous account's room IDs.
- **Triggers** — a `BGProcessingTask` requiring external power and network, plus a foreground
  sweep when the user opens room search.

One API change outside the feature: `TimelineProxyProtocol.paginateBackwards` now returns
`Result<Bool, TimelineProxyError>` — whether the start of the timeline was reached. The SDK has
always returned this and the live path was discarding it. The only other way to observe
end-of-timeline is `PaginationState`, published exclusively through `subscribeForUpdates()`, and
the sweep must not subscribe: that calls `roomListService.subscribeToRooms`, and the FFI
`RoomListService` has no unsubscribe, so a subscribing sweep would permanently grow the session's
sliding-sync subscription set. No existing caller bound the success payload, so the widening is
source-compatible.

### Deliberate divergences from element-meta#3252 — please push back if you disagree

The spec's provisional approach is "back-paginate up to 3 months in the 100 most recent rooms;
process last week of all 100, then the week before". Two parts are not implemented:

- **No 3-month age horizon.** It needs the oldest loaded event's timestamp, and the only app-side
  source is the timeline item list — which reads *highest* exactly when nothing was indexed. A room
  whose recent history is already cached re-hydrates from disk in a few cheap pages, the horizon
  fires, and the room is recorded done having indexed zero events. That is precisely the population
  this exists to serve.
- **Depth-first per room, not weekly breadth-first rounds.** Parking a room after one week needs a
  resumable per-room pagination cursor; iOS has none, since each iteration builds a fresh
  `JoinedRoomProxy` and so a fresh timeline starting at the live end. Every weekly round would
  re-walk all previously-fetched history to reach its frontier.

iOS also has no Wi-Fi-only background constraint — `requiresExternalPower` exists, "and wifi" cannot
be expressed. The page budgets and inter-page delays are the data-cost governor instead.

### Tests

30 unit tests: 8 planner (ordering and exclusions), 17 runner (reached-start, page caps, per-room
and execution deadlines, failure abandonment, cursor resume, generation rollover, excluded rooms,
never-subscribes), 5 store (round-trip, absent, unreadable-is-discarded, clear, and that the
filename cannot collide with the SDK's store prefixes).

**They pin the loop shape and nothing more.** They cannot show that a single message became
searchable: the chain from `paginateBackwards` through the event cache into tantivy has no
app-level observation point and no index API over FFI. A green suite means the loop iterated
correctly over mocks. Device verification is the reason the flag ships off.

### Notes for reviewers

**On the diff size you will see.** This depends on #5945 and #5946, neither merged, so the branch
carries their commits and GitHub shows **3,213 additions across 85 files**. This PR's own
contribution is **759 lines** of non-generated, non-test source — the last commit only. The diff
drops to that once the two below it land and this is rebased.

759 splits cleanly into three and I am happy to resend it that way: cursor/store/budget/planner
(~258, pure data and a pure function with no SDK contact); the `paginateBackwards` seam and the
runner (~216); the scheduler, flag, `BGTask` registration and lifecycle wiring (~285).

Known costs, not designed away: each swept room builds a `JoinedRoomProxy`, whose initialiser fires
an uncancellable `updateMembers()` outside every budget; and the background task does not coordinate
with the existing background app refresh, so a sweep can be suspended mid-walk — mitigated by the
cursor being resumable and every step idempotent.

Depends on #5946 for the foreground trigger. That coupling is about six lines in
`RoomFlowCoordinator`; if you would rather this landed independently, it can be decoupled.

cannot set it from a fork.

Relates to element-hq/element-meta#3252 — *[Story] Search for messages*; the "100 most recent rooms"
budget comes from its provisional approach.
Relates to element-hq/element-x-ios#4906 — an earlier community attempt at this problem, closed
because the plan was to search via the SDK index. This sweep feeds that index rather than replacing
it.
Relates to element-hq/element-x-ios#3872 — gaps in timelines blocking older event-cache content.
Relates to matrix-org/matrix-rust-sdk#5350 — the SDK-side full-text search work.

### Pull Request Checklist

- [ ] Voiceover enabled.


stefanceriu:
Please re-read our [contributing guide etiquette](https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md#etiquette) and ask before raising 4 commit 3000 line PRs. This is all to be handled on the RustSDK level and steps in that direction have been made in https://github.com/matrix-org/matrix-rust-sdk/pull/6805

PR #5948, opened 2026-07-29
Repair the search index coverage when it can't cover cached history

hayaksi1:
### Problem

The SDK's message indexer only sees events as they pass through the event cache, and only from the
moment the index is attached. Back-paginating history already on disk re-hydrates it from the cache,
and those updates are deliberately dropped before broadcast
(`event_cache/caches/room/pagination.rs` — *"Let's not propagate the updates to the store!"*). So a
session whose event cache predates its index carries a band of history that no amount of paginating
will make searchable.

iOS has this in production today: `withSearchIndexStore` was added unconditionally in 26.07.x with
no coverage bookkeeping, so every session that already existed at that upgrade has a silently
unsearchable band. The backfill sweep cannot fix it — pagination over cached history reports exactly
the same success as a network fetch that indexed everything.

### Fix

Once per session, before the client is built, delete the event cache store and its write-ahead log
sidecars. History then re-enters over the network, the one path the indexer does observe. An empty
`search-index.covered` marker in the session data directory records that coverage holds, so a
session is never charged for it twice.

| marker | event cache store | flag | outcome |
|---|---|---|---|
| present | — | — | nothing |
| absent | absent | — | write marker (the index starts level with the cache) |
| absent | present | off | nothing, and deliberately no marker, so enabling the flag later still finds it |
| absent | present | on | delete store + sidecars, clear the backfill cursor, write marker |
| absent | present | on, deletion failed | **no marker** — the next launch tries again |

A fresh login writes the marker directly: its client was built with the index attached, so its
cache cannot hold anything the indexer missed. Without that write, the *second* launch after any
login would find no marker beside a populated cache and wipe a session that was never broken.

The destructive branch is behind **`searchIndexCoverageHealEnabled`, default off**. The
non-destructive bookkeeping runs regardless, so enabling the flag later repairs exactly the
sessions that still need it.

Neither existing primitive fits. `clearCaches(syncService:)` restarts the whole session flow, far
heavier than a pre-build file delete. `deleteTransientUserData()` also clears the state store, which
this must leave intact, and swallows its errors internally — breaking the "no marker when the
deletion fails" rule that keeps the repair retryable. So `deleteEventCacheData()` is a separate,
**throwing** primitive touching only `matrix-sdk-event-cache*` in the caches directory, sidecars
first so an interrupted deletion leaves a stale database rather than a fresh one beside a replayable
WAL.

**The cost, stated plainly:** re-downloading room history costs bandwidth once per session, and the
first timeline load after a repair is noticeably slower. The alternative worth having instead is an
SDK-side "reindex from the event cache store" capability, which belongs in matrix-rust-sdk#5350 — if
you would rather have that, say so and this should be dropped in favour of tracking the hole.

### Tests

13, all filesystem-level with no SDK involvement. `SearchIndexCoverageTests` (8) — one per row of
the table above, plus idempotence, the fresh-login mark, and an interrupted repair (orphaned
sidecars must be swept, not mistaken for a session with nothing to repair).
`SessionDirectoriesTests` (5) — event cache deletion leaves the state and crypto stores alone;
deletion without a caches directory succeeds rather than throwing; store detection; marker
read/write; and that the marker survives `deleteTransientUserData()`, pinning the naming rule that
keeps clearing the caches from silently re-arming the repair.

### Notes for reviewers

**On the diff size you will see.** This depends on #5945, #5946 and #5947, none merged, so the
branch carries their commits and GitHub shows **3,684 additions across 89 files**. This PR's own
contribution is **186 lines** of non-generated, non-test source — the last commit only, 9 files.
The diff drops to that once the three below it land and this is rebased.

Declared, not fixed here:

- **No cross-process lock against the NSE**, which opens the same store paths with
  `processSetup: .multipleProcesses`. The deletion happens once per session at launch, on a store
  that is disposable by design, but I cannot rule out worse failure modes than the extension
  re-fetching a notification's events — that judgement is the team's, not mine, and it is the main
  thing I would want checked before this lands.
- **The NSE does not attach the search index.** If the notification client's fetch writes into the
  event cache store, events arriving while the app is dead would land there unindexed and open a new
  band over time, which a one-off repair does not close. I cannot verify this from the Swift side —
  it is Rust below the FFI surface — so I raise it as a question rather than asserting it.

Depends on #5947: the repair clears that sweep's cursor, because those visits walked the history
just deleted and therefore indexed nothing. That coupling is about three lines; it can be decoupled
if you would rather this landed independently.

cannot set it from a fork.

Relates to element-hq/element-meta#3252 — *[Story] Search for messages*.
Relates to element-hq/element-x-ios#5829 — attached `withSearchIndexStore` unconditionally, which is
what leaves pre-existing sessions with a silently unsearchable band.
Relates to element-hq/element-x-ios#3872 — gaps in timelines blocking older event-cache content.
Relates to matrix-org/matrix-rust-sdk#5350 — where an SDK-side "reindex from the event cache store"
would belong, and would make this PR unnecessary.

### Pull Request Checklist

- [ ] Voiceover enabled.


hayaksi1:
**The result after four PRs is shown below:**

| encrypted-1 | encrypted-2 | encrypted-3 |
| :---: | :---: | :---: |
| encrypted-1 | encrypted-2 | encrypted-3 |

| normal-room-1 | normal-room-2 |
| :---: | :---: |
| normal-1 | normal-2 |

stefanceriu:
Please re-read our [contributing guide etiquette](https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md#etiquette) and ask before raising 5 commit 3500 line PRs. This is all to be handled on the RustSDK level and steps in that direction have been made in https://github.com/matrix-org/matrix-rust-sdk/pull/6805

PR #5949, opened 2026-07-30
Combine the Server Selection screen into the Server Confirmation screen and add autocompletion when entering an account provider.

### LINKED ISSUE #4500: Element X login page extra steps logic problem
The steps to log in to elementX are very long and tedious and time-consuming.
It must be ridiculous to go through 4 pages from clicking on the app to officially log in. This is another logic optimization.

Image

Image

Image

Image

The following is the login logic of fluffychat, which is relatively simple and convenient for users to use. element X can learn it

Image

mredig:
Merges the server confirmation screen with the server selection screen and adds autocomplete when connecting to a previous server again.

notable tidbits:
- leverages the adapter pattern with introspection from elsewhere in the app to intercept keystroke information from the UITextField backing the TextField.
- SwiftUI doesn't like it when a value mutation initiated from user input happens in the same SwiftUI update cycle as a programmatic mutation, so a debounced delay was implemented for the autocompletion.

https://github.com/user-attachments/assets/0d202f63-d431-416a-bf93-54604eff12ab

Closes #4500 and https://github.com/element-hq/element-meta/issues/3283.

### Pull Request Checklist

- [ ] Voiceover enabled.


ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
```diff
@@ -34,9 +34,6 @@ class AuthenticationFlowCoordinator: FlowCoordinatorProtocol {
         /// The screen used for the whole QR Code flow.
         case qrCodeLoginScreen
         
-        /// The screen to continue authentication with the current server.
-        case serverConfirmationScreen
-        /// The screen to choose a different server.
```
pixlwave, line 39: Shouldn't we be keeping this doc comment (or changing it to something like "The screen used to choose the server.")?

ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
```diff
@@ -67,11 +64,6 @@ class AuthenticationFlowCoordinator: FlowCoordinatorProtocol {
         /// The user aborted manual login.
         case cancelledServerConfirmation
         
-        /// The user would like to enter a different server.
-        case changeServer(AuthenticationFlow)
-        /// The user is no longer selecting a server.
-        case dismissedServerSelection
-
```
pixlwave, line 74: As we're deleting these, should we rename `confirmServer` to `selectServer` and `cancelledServerConfirmation` to `cancelledServerSelection` to match the `ServerSelectionScreen` naming?

ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
```diff
@@ -355,38 +336,8 @@ class AuthenticationFlowCoordinator: FlowCoordinatorProtocol {
...
-        .store(in: &cancellables)
-        
-        navigationStackCoordinator.push(coordinator) { [weak self] in
-            self?.stateMachine.tryEvent(.cancelledServerConfirmation)
-        }
-    }
-    
     private func showServerSelectionScreen(authenticationFlow: AuthenticationFlow) {
-        let navigationCoordinator = NavigationStackCoordinator()
+        authenticationService.reset()
```
pixlwave, line 346: Please can we keep this comment as it provides real value when working on the flow from my perspective.

```suggestion
        // Reset the service back to the default homeserver before continuing. This ensures
        // we check that registration is supported if it was previously configured for login.
        authenticationService.reset()
```

ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
```diff
@@ -399,17 +350,18 @@ class AuthenticationFlowCoordinator: FlowCoordinatorProtocol {
...
                 switch action {
-                case .updated:
-                    navigationStackCoordinator.setSheetCoordinator(nil)
+                case .continueWithOAuth(let oAuthData, let window):
+                    showOAuthAuthentication(oAuthData: oAuthData, presentationAnchor: window)
+                case .continueWithPassword:
+                    stateMachine.tryEvent(.continueWithPassword)
                 case .dismiss:
-                    navigationStackCoordinator.setSheetCoordinator(nil)
+                    navigationStackCoordinator.pop()
```
pixlwave: Is this `.dismiss` action actually needed now, given that we've removed the cancel button from the screen's toolbar?

ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
```diff
@@ -507,5 +459,18 @@ class AuthenticationFlowCoordinator: FlowCoordinatorProtocol {
     
     private func userHasSignedIn(userSession: UserSessionProtocol) {
         delegate?.authenticationFlowCoordinator(didLoginWithSession: userSession)
+        
+        // track server in previous servers, maintaining uniqueness and sorting by most recent
```
pixlwave: Is there an issue with the acceptance criteria for this task? The location of this made me wonder what our definition of "signed-in" is in order to store the server. It could be here but it could also be:
- In `OnboardingFlowCoordinator` when we've confirmed the user was able to verify their account (and so are about to actually see their room list): https://github.com/element-hq/element-x-ios/blob/33f8ce27597f1ea5cb4a61ac5a118988fec80c8b/ElementX/Sources/FlowCoordinators/OnboardingFlowCoordinator.swift#L228
- In `ServerSelectionScreenViewModel` when you successfully configure a server (a pure indication that a Matrix server exists for the entered account provider).

PreviewTests/Sources/GeneratedPreviewTests.swift
pixlwave, line 1: We also need to delete the snapshots for this screen.

ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -181,6 +181,14 @@ final nonisolated class AppSettings: @unchecked Sendable {
     @UserPreference(defaultValue: AppAppearance.system)
     var appAppearance: AppAppearance
     
+    @UserPreference(key: "previousServers", defaultValue: [])
+    private var _previousServers: [String]
+    /// Tracks previous servers the user connected to for autocompletion purposes. Entries are made lowercase on write.
+    var previousServers: [String] {
+        get { _previousServers }
+        set { _previousServers = newValue.map { $0.lowercased() } }
+    }
```
pixlwave: Rather than adding boilerplate here to lowercase the setting, could we make a little micro-service that wraps `AppSettings` and can be used to add a previous server and to provide search results for the auto-completion? That would keep things tidy here and allow us to unit test it end-to-end.

ElementX/Sources/Other/TextFieldAdapter.swift
```diff
@@ -0,0 +1,98 @@
...
+    
+    func textFieldDidEndEditing(_ textField: UITextField, reason: UITextField.DidEndEditingReason) {
+        passthroughDelegate?.textFieldDidEndEditing?(textField, reason: reason)
+    }
+    
+    func textFieldDidEndEditing(_ textField: UITextField) {
+        passthroughDelegate?.textFieldDidEndEditing?(textField)
+    }
+    
+    func textField(_ textField: UITextField, shouldChangeCharactersInRanges ranges: [NSValue], replacementString string: String) -> Bool {
```
pixlwave: Small nit: Please could we put the customised delegate methods above the ones that passthrough (and maybe put a `// MARK: - Passthough` separator between them?

ElementX/Sources/Services/Authentication/AuthenticationService.swift
```diff
@@ -278,6 +278,10 @@ class AuthenticationService: AuthenticationServiceProtocol {
         }
     }
     
+    private static func defaultServer(from appSettings: AppSettings) -> String {
+        appSettings.previousServers.first ?? appSettings.accountProviders[0]
+    }
```
pixlwave: Maybe we could make this a computed property on `AppSettings` rather than using a static func here?

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/ServerSelectionScreenViewModel.swift
```diff
@@ -49,22 +76,68 @@ class ServerSelectionScreenViewModel: ServerSelectionScreenViewModelType, Server
...
+        
+        switch await authenticationService.configure(for: accountProvider, flow: authenticationFlow) {
+        case .success:
+            MXLog.info("Selected server: \(accountProvider)")
+            await fetchLoginURLIfNeededAndContinue()
+            stopLoading()
+        case .failure:
+            MXLog.info("Invalid server: \(accountProvider)")
+            stopLoading()
+            state.bindings.alertInfo = AlertInfo(id: .unknownError)
```
pixlwave, line 98:
```suggestion
            // When the servers are hard-coded they should have a valid configuration, so show a generic error.
            state.bindings.alertInfo = AlertInfo(id: .unknownError)
```

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/ServerSelectionScreenModels.swift
```diff
@@ -6,21 +6,64 @@
...
     var bindings: ServerSelectionScreenBindings
+    
+    /// Upon introspection, this is the UITextField backing the TextField for server input
+    var textField: UITextField? {
+        get { adapter.textField }
+        set { adapter.textField = newValue }
+    }
+    
+    /// Adapts the text field for custom functionality beyond what's available in SwiftUI
+    let adapter = TextFieldAdapter()
```
pixlwave: Minor nit for clarity at the call-site:

```suggestion
    let textFieldAdapter = TextFieldAdapter()
```

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/View/ServerSelectionScreen.swift
```diff
@@ -89,38 +90,75 @@ struct ServerSelectionScreen: View {
...
         let authenticationService = AuthenticationService.mock
         
         let viewModel = ServerSelectionScreenViewModel(authenticationService: authenticationService,
+                                                       mode: mode,
                                                        authenticationFlow: .login,
                                                        appSettings: .volatile(),
                                                        userIndicatorController: UserIndicatorControllerMock())
-        viewModel.context.homeserverAddress = homeserverAddress
-        if homeserverAddress == "thisisbad" {
+        if case .userInput(let homeserverAddress) = mode, homeserverAddress == "thisisbad" {
```
pixlwave: This should also work right?

```suggestion
        if case .userInput("thisisbad") = mode {
```

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/View/ServerSelectionScreen.swift
```diff
@@ -22,64 +22,65 @@ struct ServerSelectionScreen: View {
                 serverForm
             }
             .readableFrame()
-            .padding(.horizontal, 16)
+        } bottomContent: {
+            continueButton
         }
         .background(Color.compound.bgCanvasDefault.ignoresSafeArea())
-        .toolbar { toolbar }
```
pixlwave, line 28: Looking at the designs I think we need a title on this screen now, although I've [asked](https://www.figma.com/design/jvZ7TIyd7V9Vn8sIDB9vMd?node-id=4032-209431&m=dev#1865272604) for clarification if it should change based on the flow.

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/ServerSelectionScreenViewModel.swift
```diff
@@ -49,22 +76,68 @@ class ServerSelectionScreenViewModel: ServerSelectionScreenViewModelType, Server
     
     // MARK: - Private
     
+    private func pickServer() async {
```
pixlwave, line 81: Thank you for adding the logs to this one 🙌

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/ServerSelectionScreenModels.swift
```diff
@@ -6,21 +6,64 @@
...
+    case continueWithOAuth(data: OAuthAuthorizationDataProxy, window: UIWindow)
+    /// Continue the flow using password authentication.
+    case continueWithPassword
     /// Dismiss the view without using the entered address.
     case dismiss
 }
 
+enum ServerSelectionScreenMode: Equatable {
+    /// The user is confirming the displayed account provider (or can enter their own).
+    case userInput(defaultValue: String)
```
pixlwave: As far as I can see `defaultValue` isn't actually used anywhere in the app, we're continuing to consider `authenticationService.homeserver.value.address` to be the source of truth (which makes sense to me).

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/ServerSelectionScreenViewModel.swift
```diff
@@ -49,22 +76,68 @@ class ServerSelectionScreenViewModel: ServerSelectionScreenViewModelType, Server
...
+        }
+        
+        startLoading()
+        
+        let homeserver = authenticationService.homeserver.value
+        guard homeserver.loginMode == .unknown || homeserver.address != accountProvider else {
+            await fetchLoginURLIfNeededAndContinue()
+            stopLoading()
+            return
+        }
```
pixlwave: Given that we don't include https://github.com/element-hq/element-x-ios/blob/ae795df53045c318c81c91eadc4464e7cdb25965/ElementX/Sources/Screens/Authentication/ServerConfirmationScreen/ServerConfirmationScreenViewModel.swift#L81-L86 in `configureHomeserver` (which makes perfect sense to me now that the modal selection has been removed), I think we can probably also remove this too.

All we're saving is a re-configure if you start, then cancel and then start again which is more of an edge-case in the flow.

UnitTests/Sources/ServerSelectionScreenViewModelTests.swift
pixlwave, line 1: Hmmm these don't seem ideal now. There appears to be a bunch of overlap between the existing tests and the ones brought over from the ServerConfirmationScreen. Personally I don't think any of the `userInput` tests really add any value.

UnitTests/Sources/ServerSelectionScreenViewModelTests.swift
```diff
@@ -141,16 +147,343 @@ struct ServerSelectionScreenViewModelTests {
...
+    
     // MARK: - Helpers
     
-    private mutating func setup(authenticationFlow: AuthenticationFlow) {
-        let appSettings = AppSettings.volatile()
+    private mutating func setup(authenticationFlow: AuthenticationFlow,
+                                mode: ServerSelectionScreenMode = .userInput(defaultValue: ""),
+                                supportsOAuth: Bool = true,
+                                supportsOAuthCreatePrompt: Bool = true,
+                                supportsPasswordLogin: Bool = true) throws {
```
pixlwave: And I *think* we could probably get rid of these parameters by making the picker array from some of these fixtures (but could be completely off there):
https://github.com/element-hq/element-x-ios/blob/783251184526bcd8c57a11f8c0a880e1a39c6349/ElementX/Sources/Mocks/AuthenticationClientBuilderFactoryMock.swift#L15-L43

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/View/ServerSelectionScreen.swift
pixlwave, line 1: Nice merging of the 2 modes btw, thanks 👌

UnitTests/Sources/ServerSelectionScreenViewModelTests.swift
```diff
@@ -141,16 +147,343 @@ struct ServerSelectionScreenViewModelTests {
...
+        // Given a view model with a previous server in history.
+        try setup(authenticationFlow: .login)
+        appSettings.previousServers = ["myserver.com"]
+        context.homeserverAddress = ""
+        let textField = UITextField()
+        context.send(viewAction: .updateTextField(textField))
+        
+        // When the user types a prefix that matches the previous server.
+        let typedSoFar = "my"
+        let nextChar = "s"
```
pixlwave: This isn't Rust, I think we've got enough characters to type out the full name for all of these 😅

```suggestion
        let nextCharacter = "s"
```

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/View/ServerSelectionScreen.swift
```diff
@@ -89,38 +90,75 @@ struct ServerSelectionScreen: View {
...
+            }
+        }
+        .clipShape(RoundedRectangle(cornerRadius: 10))
+    }
+}
+
 // MARK: - Previews
 
 @available(iOS 26.0, *)
 struct ServerSelection_Previews: PreviewProvider, TestablePreview {
```
pixlwave, line 142: We'll need to delete the old snapshots and regenerate these too.

pixlwave (commented):
Thanks, this is going to be such a nice improvement! Sorry for the amount of comments, but I don't think there's anything major in there 👍

pixlwave:
Note: I've added a recording to the PR description and I've removed one of the two `pr-` labels as only 1 is allowed (so Danger is failing).

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/View/ServerSelectionScreen.swift
```diff
@@ -22,64 +22,65 @@ struct ServerSelectionScreen: View {
                 serverForm
             }
             .readableFrame()
-            .padding(.horizontal, 16)
+        } bottomContent: {
+            continueButton
         }
         .background(Color.compound.bgCanvasDefault.ignoresSafeArea())
```
pixlwave: Oh also we need to use the conditional `backgroundColor` from the old screen here:

https://github.com/element-hq/element-x-ios/blob/743c22d980370bd8c414ce0d2bcf509af63d7d31/ElementX/Sources/Screens/Authentication/ServerConfirmationScreen/View/ServerConfirmationScreen.swift#L15-L20

And I think we need it via the `backgroundStyle` to make sure the `FullscreenDialog` can read it internally too.

```suggestion
        .background()
        .backgroundStyle(backgroundColor)
```

Without this the picker view looks like so

Image

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/View/ServerSelectionScreen.swift
```diff
@@ -22,64 +22,65 @@ struct ServerSelectionScreen: View {
...
+        .introspect(.window, on: .supportedVersions) { window in
+            context.send(viewAction: .updateWindow(window))
+        }
     }
     
     /// The title, message and icon at the top of the screen.
     var header: some View {
         VStack(spacing: 8) {
-            BigIcon(icon: \.host)
+            BigIcon(icon: \.userProfileSolid)
```
pixlwave: And then this needs the conditional icon style too:

https://github.com/element-hq/element-x-ios/blob/743c22d980370bd8c414ce0d2bcf509af63d7d31/ElementX/Sources/Screens/Authentication/ServerConfirmationScreen/View/ServerConfirmationScreen.swift#L29-L34

ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
mredig: I'm not sure what acceptance criteria you're referring to. My intent was to place it in a location where the server is validated as a real server and that the user has been authenticated. This was the earliest point that I could find that both the password and oidc flows converged. If you want it somewhere else, I'm open to it.

PreviewTests/Sources/GeneratedPreviewTests.swift
mredig: I've not dealt with snapshot testing - where would I look for direction?

ElementX/Sources/Application/Settings/AppSettings.swift
mredig: I like that - will do.

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/View/ServerSelectionScreen.swift
mredig: aha! Yes I missed that detail. good catch

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/ServerSelectionScreenModels.swift
mredig: good catch - it was carried over without much thought from the old screen. updated

UnitTests/Sources/ServerSelectionScreenViewModelTests.swift
mredig: I removed the `userInputLogin*` and `userInputRegister*` tests as they are largely redundant with the tests that were already here, but the `userInputPassword*` tests uniquely test the password flow, so I'm keeping them here unless you don't think that variation is enough to warrant its own test.

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/View/ServerSelectionScreen.swift
mredig: Updated, but I believe it's rendering correctly with just `.background(backgroundColor)` - if I'm missing something, I can update as suggested, but this is simpler code as is.

ElementX/Sources/Application/Settings/AppSettings.swift
mredig: I made these changes in a separate commit so it'll be easier to separate out from the other updates. There're a lot of ways this could be done, so I wanted to make sure it could be scrutinized individually more easily.

ElementX/Sources/Services/Client/HomeserverHistoryManager.swift
pixlwave, line 1: Thanks, this looks much neater now 👌

It feels like there may be some redundant calls to `lowercased()` in here. Clearly not a huge issue but thought it was worth pointing out in case there's some simple tweaks.

UnitTests/Sources/ServerSelectionScreenViewModelTests.swift
```diff
@@ -470,18 +338,12 @@ struct ServerSelectionScreenViewModelTests {
...
         
-        client = ClientSDKMock(.init(oAuthLoginURL: supportsOAuth ? "https://account.matrix.org/authorize" : nil,
-                                     supportsOAuthCreatePrompt: supportsOAuthCreatePrompt,
-                                     supportsPasswordLogin: supportsPasswordLogin))
-        var factoryConfig = AuthenticationClientFactoryMock.Configuration()
-        factoryConfig.homeserverClients["matrix.org"] = client
-        factoryConfig.homeserverClients["https://matrix-client.matrix.org"] = client
+        let factoryConfig = AuthenticationClientFactoryMock.Configuration()
+        // matrix.org: OAuth. example.com: password only. server.net: no login. secure.gov: OAuth + Element Pro required.
+        client = factoryConfig.homeserverClients["matrix.org"]!
```
pixlwave: This Force Unwrapping Violation can be fixed with a `#require`.

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/View/ServerSelectionScreen.swift
```diff
@@ -89,38 +106,79 @@ struct ServerSelectionScreen: View {
...
+        .previewDisplayName("Error")
+        
+        ElementNavigationStack {
+            ServerSelectionScreen(context: pickerViewModel.context)
+        }
+        .previewDisplayName("Picker")
     }
     
-    static func makeViewModel(for homeserverAddress: String) -> ServerSelectionScreenViewModel {
+    static func makeViewModel(mode: ServerSelectionScreenMode = .userInput, server: String) -> ServerSelectionScreenViewModel {
```
pixlwave: Would you mind leaving this parameter name as it was? At some stage I'd like to do a refactor of all of these to `accountProvider` (potentially with its own type), so this will just help with a find & replace.

```suggestion
    static func makeViewModel(mode: ServerSelectionScreenMode = .userInput, homeserverAddress: String) -> ServerSelectionScreenViewModel {
```

ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
```diff
@@ -507,5 +464,11 @@ class AuthenticationFlowCoordinator: FlowCoordinatorProtocol {
     
     private func userHasSignedIn(userSession: UserSessionProtocol) {
         delegate?.authenticationFlowCoordinator(didLoginWithSession: userSession)
+        
+        guard let newServer = userSession.clientProxy.userIDServerName else {
```
pixlwave: Ah, one annoyance here is that we will need this to store whatever the user entered rather than specifically the server name. Whilst it's nice to assume that the server should have a `.well-known/matrix/client` file set up, this isn't always the case.

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/ServerSelectionScreenCoordinator.swift
```diff
@@ -36,9 +37,17 @@ final class ServerSelectionScreenCoordinator: CoordinatorProtocol {
...
+            .userInput
+        } else {
+            .picker(parameters.appSettings.accountProviders)
+        }
+        
         viewModel = ServerSelectionScreenViewModel(authenticationService: parameters.authenticationService,
+                                                   mode: mode,
                                                    authenticationFlow: parameters.authenticationFlow,
-                                                   appSettings: parameters.appSettings,
+                                                   appSettings: parameters.appSettings, homeserverHistoryManager: parameters.homeserverHistoryManager,
```
pixlwave:
```suggestion
                                                   appSettings: parameters.appSettings,
                                                   homeserverHistoryManager: parameters.homeserverHistoryManager,
```

ElementX/Sources/Services/Client/HomeserverHistoryManager.swift
pixlwave: Also, did you want to add some tests for this?

UnitTests/Sources/ServerSelectionScreenViewModelTests.swift
```diff
@@ -159,8 +354,28 @@ struct ServerSelectionScreenViewModelTests {
                                         appHooks: AppHooks())
         
         viewModel = ServerSelectionScreenViewModel(authenticationService: service,
+                                                   mode: mode,
                                                    authenticationFlow: authenticationFlow,
                                                    appSettings: appSettings,
```
pixlwave, line 359: Looks like the tests don't compile any more.

UnitTests/Sources/ServerSelectionScreenViewModelTests.swift
```diff
@@ -141,16 +148,204 @@ struct ServerSelectionScreenViewModelTests {
...
+    
     // MARK: - Helpers
     
-    private mutating func setup(authenticationFlow: AuthenticationFlow) {
-        let appSettings = AppSettings.volatile()
+    private mutating func setup(authenticationFlow: AuthenticationFlow,
+                                mode: ServerSelectionScreenMode = .userInput) throws {
+        appSettings = AppSettings.volatile()
+        
+        let factoryConfig = AuthenticationClientFactoryMock.Configuration()
```
pixlwave:
```suggestion
        let factoryConfiguration = AuthenticationClientFactoryMock.Configuration()
```

UnitTests/Sources/ServerSelectionScreenViewModelTests.swift
pixlwave: That looks like the right choice to me, thanks 👌

ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
pixlwave: There's an issue now 🎉 https://github.com/element-hq/element-meta/issues/3283

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/View/ServerSelectionScreen.swift
pixlwave: Ah sorry I could have explained the problem a bit more. Without doing `.background().backgroundStyle(color)` the bottom section doesn't have access to the colour used to clip the main content, so yeah we do need it:

| `.background(color)` | `.background().backgroundStyle(color)` |
| - | - |
| Screenshot 2026-08-03 at 11 01 25 am | Screenshot 2026-08-03 at 11 01 51 am |

PreviewTests/Sources/GeneratedPreviewTests.swift
pixlwave: Our guide seems severely lacking 😕 https://github.com/element-hq/element-x-ios/blob/develop/CONTRIBUTING.md#snapshot-tests

There's a PreviewTests scheme in Xcode. The tests expect an iPhone SE on 26.5 with an en-US locale. If there's no snapshot for a particular `TestablePreview` it will record one. If there is it will compare them and fail if they don't match.

In this instance it would be better to delete all of the ServerSelection/ServerConfirmation snapshots and run the tests to re-record (due to screen removal and the addition of `.previewDisplayName`), but there's also a `RECORD_FAILURES` environment variable in the test plan that can be enabled to automatically re-record the snapshots for any failing tests.

pixlwave (commented):
Thanks for the changes, a few more comments and info about snapshots 👇

ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
mredig: Awesome! That said, I feel like this is still consistent with the acceptance criteria. Is there a nuance you're referring to that I'm overlooking?

PreviewTests/Sources/GeneratedPreviewTests.swift
mredig: I think I got it working

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/View/ServerSelectionScreen.swift
mredig: Ah I see - updated.

ElementX/Sources/Services/Client/HomeserverHistoryManager.swift
mredig: > may be some redundant calls

slightly, but as far as I can tell, necessary. I believe every call to it covers a case where it could introduce capitalization.

However, I was able to find a slight tweak that reduces one case of doubling up, offset by some required vigilance by the developer.

As for tests - wanted to make sure this approach is what you wanted first. Adding now.

ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
mredig: I updated to use `AuthenticationService.homeServer`. I think that retains verbatim-ish user input and doesn't update from the server connection.

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/ServerSelectionScreenCoordinator.swift
mredig: weird - I would have expected SwiftFormat to get that. FWIW though, this is a consequence of the parameter alignment style - everything gets pushed far right when argument labels are long and it becomes very hard to read.

mredig:
There was a small placeholder text change in figma I included in the latest round of updates

ElementX/Sources/Services/Client/HomeserverHistoryManager.swift
pixlwave: Nice on both counts (sorry I hadn't realised that about the tests as the approach came across as sound to me 😄).

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/View/ServerSelectionScreen.swift
pixlwave: Thank you!

ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
pixlwave: Perfect, that looks decent to me 👍

ElementX/Sources/FlowCoordinators/AuthenticationFlowCoordinator.swift
pixlwave: Not at all, it matches from my perspective. I was just looking somewhere for confirmation of the desired behaviour 👍

PreviewTests/Sources/GeneratedPreviewTests.swift
pixlwave: Yep, thanks 👌

ElementX/Resources/Localizations/en.lproj/Untranslated.strings
pixlwave, line 1: Alright, I've added these to Localazy. A `swift run tools download-strings` should pull them in. 

Note: I've kept them in the old namespace as it's not much fun for management if related strings become fragmented.

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/View/ServerSelectionScreen.swift
```diff
@@ -12,8 +12,22 @@ import SwiftUI
...
+        case .picker: .default
+        }
+    }
+    
     var body: some View {
-        ScrollView {
+        FullscreenDialog {
             VStack(spacing: 0) {
                 header
                     .padding(.top, UIConstants.iconTopPaddingToNavigationBar)
```
pixlwave: Now that we have both the `FullscreenDialog` and the `.navigationTitle` we don't need this padding anymore...

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/View/ServerSelectionScreen.swift
```diff
@@ -12,8 +12,22 @@ import SwiftUI
...
+    private var headerIconStyle: BigIcon.Style {
+        switch context.viewState.mode {
+        case .userInput: .defaultSolid
+        case .picker: .default
+        }
+    }
+    
     var body: some View {
-        ScrollView {
+        FullscreenDialog {
```
pixlwave: And instead we should customise the build in topPadding value here to match the Figma:


```suggestion
        FullscreenDialog(topPadding: 24) {
```

pixlwave (commented):
Ok looking good, hopefully this is the last round of changes:

- I've adding the strings to Localazy.
- I noticed a padding issue when looking through the snapshots (which unfortunately means you'll need to re-record them).
- We're also going to need to tweak the UI and Integration tests so they continue to run:
    - There's an `AuthenticationFlowCoordinatorTests.swift` in the UI tests that expects the server confirmation to exist (there shouldn't be any snapshots to update for these).
    - There's a `login()` function in the Integration tests' `Common.swift` that will need similar changes. You can set a homeserver in `Application.swift` to test the changes against.

mredig:
> * We're also going to need to tweak the UI and Integration tests so they continue to run:
>   
>   * There's an `AuthenticationFlowCoordinatorTests.swift` in the UI tests that expects the server confirmation to exist (there shouldn't be any snapshots to update for these).
>   * There's a `login()` function in the Integration tests' `Common.swift` that will need similar changes. You can set a homeserver in `Application.swift` to test the changes against.

login is fixed, but I can't validate the rest of the integration tests. I also found a couple ui tests that needed updated screenshots.

pixlwave (approved):
Thanks very much, I pushed a couple of snapshots to make sure the keyboard was included (the hardware keyboard isn't attached on CI), otherwise looks ready to me!

PR #5957, opened 2026-07-30
Small gallery tweaks

Velin92:
- Added the string for the room list last message
- The string for the pinned items banner
- Notification content builder handling for gallery pushes

### DIFF
--- a/NSE/Sources/NotificationContentBuilder.swift
+++ b/NSE/Sources/NotificationContentBuilder.swift
@@ -207,6 +207,13 @@ nonisolated struct NotificationContentBuilder {
                                      using: mediaProvider,
                                      mediaSource: .init(source: content.source,
                                                         mimeType: content.info?.mimetype))
+        case .gallery(content: let content):
+            // A notification can only show one attachment, so the gallery is represented by its first.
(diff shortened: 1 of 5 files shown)


NSE/Sources/NotificationContentBuilder.swift
```diff
@@ -362,3 +369,24 @@ private nonisolated struct NotificationIcon {
...
+        for itemType in self {
+            switch itemType {
+            case .image(let content):
+                return .init(source: content.source, mimeType: content.info?.mimetype)
+            case .video(let content):
+                return .init(source: content.source, mimeType: content.info?.mimetype)
+            case .audio(let content):
+                return .init(source: content.source, mimeType: content.info?.mimetype)
+            case .file, .other:
+                continue
```
stefanceriu, line 386: Surely audio and file/other should be handle the same way, no? Presumably the default icon?

Velin92: this is the reusing the same pattern we use for other notifications, and single file notifications do not display any icon or media, we just say "sent an attachment"

stefanceriu (approved):
🙌

PR #5958, opened 2026-07-30
Increase the multi-file upload limit to 25

beposec:
**What changed**

This PR increases the maximum number of files that can be selected or shared at once from 10 to 25.

**Why**

The original limit of 10 was chosen mainly because inline media galleries were not supported at the time (https://github.com/element-hq/element-x-ios/pull/4358). Sending a larger number of attachments would therefore have resulted in many individual timeline entries and could easily spam the room.

Now that media galleries are available, multiple attachments are displayed in a much more compact and user-friendly way. This removes the main reason for keeping the limit low.

Increasing the limit to 25 better supports common use cases such as sharing a larger set of photos, while still keeping a reasonable upper bound on the number of files processed in a single upload.

### DIFF
--- a/ElementX/Sources/Screens/MediaPickerScreen/DocumentPicker.swift
+++ b/ElementX/Sources/Screens/MediaPickerScreen/DocumentPicker.swift
@@ -79,7 +79,7 @@ struct DocumentPicker: UIViewControllerRepresentable {
             }
             
             var selectedURLs = [URL]()
-            for url in urls.prefix(10) {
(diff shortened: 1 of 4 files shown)


Velin92:
Hello, no worries we are going to handle this ourselves, also the suggested MSC cap is 60, and we are probably going to allow that many items to be added to the gallery at once! (And only when the gallery FF is on)

beposec:
Great! Thanks for you work!

PR #5959, opened 2026-07-30
Allow clearing your own call status.

pixlwave:
I forgot that we needed to support showing the Call status in the Settings screen status row (the profile row already shows it) so that the user can manually clear it if they want. This PR adds that state:

Screenshot 2026-07-30 at 4 59 16 pm

### DIFF
--- a/ElementX/Sources/Screens/Settings/SettingsScreen/View/SettingsScreenUserStatusRow.swift
+++ b/ElementX/Sources/Screens/Settings/SettingsScreen/View/SettingsScreenUserStatusRow.swift
@@ -9,7 +9,7 @@ import Compound
 import SwiftUI
 
 struct SettingsScreenUserStatusRow: View {
-    enum Mode: Equatable { case pick, custom(emoji: Character), show(UserStatus.Raw) }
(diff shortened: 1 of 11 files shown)


ElementX/Sources/Screens/Settings/SettingsScreen/View/SettingsScreenUserStatusRow.swift
```diff
@@ -9,7 +9,7 @@ import Compound
 import SwiftUI
 
 struct SettingsScreenUserStatusRow: View {
-    enum Mode: Equatable { case pick, custom(emoji: Character), show(UserStatus.Raw) }
+    enum Mode: Equatable { case pick, custom(emoji: Character), show(UserStatus.Displayed) }
```
stefanceriu: I'm not sure I understand how these fit together, or perhaps I just find the naming hard to digest. I would perhaps call `UserSelected` so it's extra clear, and why do we need a `var displayed` with its own type instead of exposing `text` and `emoji` properties directly from the status? 🤔

stefanceriu (approved):
👌

PR #5960, opened 2026-07-30
Set gallery sending limit to 60 when the gallery ff is on

Velin92:
just as the MSC 4274 suggests


beposec:
Wow, that was a really fast response to my suggestion! Thanks so much—I really appreciate it!

pixlwave (approved):
LGTM 👌

PR #5963, opened 2026-07-31
Fix for pagination getting stuck loading when swiping media previews in the timeline

Velin92:
When a pagination finished without adding any media, nothing re-evaluated the placeholder: `previewItemsPaginationPublisher` only fired when items were added, and the controller only refreshed on placeholder→media, never when one placeholder replaced another. So the spinner stayed even though the data source would now return the start of the timeline for that index. `paginationState` now publishes when it changes, and the controller refreshes whenever the item at the current index differs from the one on display.


stefanceriu (approved):
Nice 👍

PR #5964, opened 2026-07-31
Add a custom TabRail on iPad instead of the top tab bar.

### LINKED ISSUE #4617: Space/Chats tab bar looks fugly on Sequoia
### Steps to reproduce

Image

#### What did you expect?

pretty tab bar

#### What happened instead?

fugly tab bar

### Your phone model

mbp

### Operating system version

macos sequoia 15.6

### Application version

1042

### Homeserver

_No response_

### Will you send logs?

Yes

pixlwave:
This is a simplified version of the concept I built for the Hackathon based on a real design with a bunch of improvements and layout fixes applied. It's taken much longer to get over the line than I had hoped, but at this stage seems robust enough to me to be worth shipping.

Not enabled on iOS 18, and will need some additional tweaks for iOS 27 when that ships given the split view appearance has changed again (similar to iOS 18 but without a vertical separator for some reason).

| Fullscreen | Windowed | macOS |
| - | - | - |
| Simulator Screenshot - iPad Pro 11-inch (M5) - 2026-07-31 at 17 36 02 | Simulator Screenshot - iPad Pro 11-inch (M5) - 2026-07-31 at 17 36 14 | Screenshot 2026-07-31 at 6 06 19 pm |

Closes #4617 (if you ignore that it is for *OS 27+ only)

Disclosure: There is one remaining bug where if the window's width is exactly the point where the split view chooses to hide the sidebar, the simultaneous removal of the rail results in a flicker between the sidebar showing/hiding. While not ideal, this is easily resolved with a small adjustment in window size.

### DIFF
--- a/ElementX/Sources/Application/Navigation/NavigationCoordinators.swift
+++ b/ElementX/Sources/Application/Navigation/NavigationCoordinators.swift
@@ -335,7 +335,7 @@ extension EnvironmentValues {
 }
 
 private struct NavigationSplitCoordinatorView: View {
-    @Environment(\.horizontalSizeClass) private var horizontalSizeClass
+    @Environment(\.tabViewHorizontalSizeClass) private var horizontalSizeClass
     @Environment(\.scenePhase) private var scenePhase
     
     @Bindable var navigationSplitCoordinator: NavigationSplitCoordinator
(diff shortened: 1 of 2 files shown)


Velin92 (approved):

dwt:
I am a bit saddened that the extra sidebar approach is used here instead of doing this as a switcher at the top of the left pane, because that had the potential to save a lot of screen real estate. Some explainer: I have to have a lot of chat apps open, and like to keep them as small and as content focused (i.e. on the actual chats) as possible as I am using a 14" device. Element has historically not supported this very well and this will be another paper cut in that direction.

PR #5969, opened 2026-08-03
A11y for Galleries

Velin92:
- It allows to voice over each individual element of the gallery and select it
- It reads the index and the type of the gallery attachment
- The first time you select through voice over a new gallery it will also read sender timestamp and caption (if available), but if you select another element of the gallery it won't (until you select a new container) to eliminate unnecessary redundancy.

### DIFF
--- a/ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryItemTileView.swift
+++ b/ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryItemTileView.swift
@@ -10,10 +10,24 @@ import SwiftUI
 
 struct GalleryItemTileView: View {
     let item: GalleryItem
+    /// The item's position within the gallery, announced instead of its filename.
+    let position: Int
     let mediaProvider: MediaProviderProtocol?
     let contentScannerService: ContentScannerServiceProtocol?
     let overflowCount: Int
     let onTap: () -> Void
+    
+    private var accessibilityLabel: String {
+        if overflowCount > 0 {
+            L10n.a11yGalleryMoreMedia(overflowCount)
+        } else if item.isVideo {
+            L10n.a11yGalleryVideo(position)
+        } else if item.isImage {
+            L10n.a11yGalleryImage(position)
+        } else {
+            L10n.a11yGalleryAttachment(position)
+        }
+    }
     
     var body: some View {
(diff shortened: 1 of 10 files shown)


ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryItemTileView.swift
```diff
@@ -33,11 +47,15 @@ struct GalleryItemTileView: View {
...
-                    .accessibilityLabel(item.filename)
+                    .accessibilityLabel(accessibilityLabel)
             } scanningContent: {
                 ScanningMediaEventsTimelineView()
+                    .accessibilityElement(children: .ignore)
+                    .accessibilityLabel(L10n.a11yGalleryScanning(position))
             } unsafeContent: { failure in
                 UnsafeMediaEventsTimelineView(failure: failure)
+                    .accessibilityElement(children: .ignore)
+                    .accessibilityLabel(L10n.A11y.galleryFailedScan(position))
```
pixlwave: I think you need to enable the `nested` option on Localazy for this key (looks like there's some others nearby it too that need it as well.

```suggestion
                    .accessibilityLabel(L10n.a11yGalleryFailedScan(position))
```

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryRoomTimelineView.swift
```diff
@@ -40,7 +40,9 @@ struct GalleryRoomTimelineView: View {
                     if usesListLayout {
                         GalleryDivider()
                     }
+                    // Announced when entering the gallery, so not a stop of its own.
                     caption
+                        .accessibilityHidden(true)
```
pixlwave: Perhaps?

```suggestion
caption
                        .accessibilityHidden(true) // Announced when entering the gallery, so not a stop of its own.
```

pixlwave (approved):

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/GalleryItemTileView.swift
Velin92: whoops missed this one thanks.

PR #5970, opened 2026-08-03
Add timeouts after verifying the device owner when linking a new device.

### LINKED ISSUE #5954: Add timeout to Link new device flow after the biometric verification step
https://github.com/element-hq/element-x-ios/pull/5773 and https://github.com/element-hq/element-x-ios/pull/5794 added the biometric/PIN/app lock check to the Link new device flow.

After this step we want to enforce a maximum time in which the user must complete the linking.

Flow: https://www.figma.com/design/66PalsLxjI7v0UAWF1bGE0/Compound-Matrix-Auth-Service--MAS--Templates?node-id=2801-58063

pixlwave:
- Linking a new mobile device uses the timeout already emitted by the SDK (by removing the automatic regeneration of the QR code). This is currently 1-minute but will be extended to 2-minutes when the newer QR code MSC is available.
- Linking a new desktop computer schedules an expiration date on the `QRCodeLoginScreen` (shown immediately after verifying).

Also includes #5927 now that we have SDK support.

Easiest to review commit-by-commit. Closes #5954.

### DIFF
--- a/ElementX/Sources/Screens/QRCodeLoginScreen/QRCodeLoginScreenViewModel.swift
+++ b/ElementX/Sources/Screens/QRCodeLoginScreen/QRCodeLoginScreenViewModel.swift
@@ -262,12 +262,13 @@ class QRCodeLoginScreenViewModel: QRCodeLoginScreenViewModelType, QRCodeLoginScr
     private func handleExpiration() {
         // Don't override an error that's already being shown.
         if case .error = state.state {
+            currentTask = nil // Cancel the scan (just in case).
             return
         }
         
         MXLog.info("Timed out establishing a secure channel, expiring.")
         currentTask = nil // Cancel the scan.
-        state.state = .error(.expired)
+        handleError(.expired)
     }
     
     private func sendCheckCode() async {
@@ -290,10 +291,8 @@ class QRCodeLoginScreenViewModel: QRCodeLoginScreenViewModelType, QRCodeLoginScr
     }
     
     private func requestOAuthAuthorization(url: URL, continuationSender: ContinuationMessageSenderProxy) {
-        // We're meant to use the continuationSender to confirm when the URL has been opened or if the user decided
-        // against it (by cancelling the system's WAS prompt). Unfortunately the WAS only gives us a signal for
-        // success or failure (and we don't get success as there isn't a redirect at the end of the flow).
-        // Failure could be the user cancelling the prompt or closing the web view when they're done.
+        // There's no OAuth redirect, so the WAS always returns a cancellation failure (both for a declined system prompt and
+        // when closing after successful authorisation). We can't use the continuation as intended, so confirm it and continue.
         Task { await continuationSender.confirm() }
         
         let (stream, continuation) = AsyncStream<Result<Void, OAuthError>>.makeStream()
--- a/ElementX/Sources/Screens/QRCodeLoginScreen/QRCodeLoginScreenViewModel.swift
+++ b/ElementX/Sources/Screens/QRCodeLoginScreen/QRCodeLoginScreenViewModel.swift
@@ -13,6 +13,11 @@ typealias QRCodeLoginScreenViewModelType = StateStoreViewModel<QRCodeLoginScreen
 
 class QRCodeLoginScreenViewModel: QRCodeLoginScreenViewModelType, QRCodeLoginScreenViewModelProtocol {
     private let appMediator: AppMediatorProtocol
+    /// The time allowed between verifying the device's owner (on the previous screen), to having
+    /// scanned a QR code and established a secure channel for the `linkDesktopComputer` flow.
+    ///
+    /// The generated QR code handles this automatically when linking a mobile device and QR login doesn't need it.
+    private let linkDesktopTimeout: Duration
     
     private let actionsSubject: PassthroughSubject<QRCodeLoginScreenViewModelAction, Never> = .init()
     var actionsPublisher: AnyPublisher<QRCodeLoginScreenViewModelAction, Never> {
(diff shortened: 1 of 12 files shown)


ElementX/Sources/Screens/QRCodeLoginScreen/QRCodeLoginScreenViewModel.swift
```diff
@@ -232,37 +246,28 @@ class QRCodeLoginScreenViewModel: QRCodeLoginScreenViewModelType, QRCodeLoginScr
...
+    
+    private func handleExpiration() {
+        // Don't override an error that's already being shown.
+        if case .error = state.state {
+            return
         }
         
-        listenToDisplayQRProgress(progressPublisher: progressPublisher)
+        MXLog.info("Timed out establishing a secure channel, expiring.")
+        currentTask = nil // Cancel the scan.
```
Velin92, line 270: shouldn't we cancel the task also in the guard?

ElementX/Sources/Screens/QRCodeLoginScreen/QRCodeLoginScreenViewModel.swift
```diff
@@ -284,7 +289,13 @@ class QRCodeLoginScreenViewModel: QRCodeLoginScreenViewModelType, QRCodeLoginScr
         }
     }
     
-    private func requestOAuthAuthorization(url: URL) {
+    private func requestOAuthAuthorization(url: URL, continuationSender: ContinuationMessageSenderProxy) {
+        // We're meant to use the continuationSender to confirm when the URL has been opened or if the user decided
+        // against it (by cancelling the system's WAS prompt). Unfortunately the WAS only gives us a signal for
+        // success or failure (and we don't get success as there isn't a redirect at the end of the flow).
+        // Failure could be the user cancelling the prompt or closing the web view when they're done.
```
Velin92: I guess this could be a bit less verbose

ElementX/Sources/Screens/QRCodeLoginScreen/QRCodeLoginScreenViewModel.swift
```diff
@@ -341,6 +349,7 @@ class QRCodeLoginScreenViewModel: QRCodeLoginScreenViewModelType, QRCodeLoginScr
     /// Only for mocking initial states
     fileprivate init(state: QRCodeLoginState, mode: QRCodeLoginScreenMode, canSignInManually: Bool, checkCodeInput: String) {
         appMediator = AppMediatorMock(.init())
+        linkDesktopTimeout = .seconds(3600)
```
Velin92, line 351: 1 hour timeout seems a lot, is this some kind of standard of the flow? wouldn't 10 mins be okay too?

ElementX/Sources/Services/Authentication/LinkNewDeviceService.swift
```diff
@@ -256,6 +256,34 @@ nonisolated class CheckCodeSenderProxy: Equatable {
     }
 }
 
+class ContinuationMessageSenderProxy: Equatable {
```
Velin92: final nonisolated and Sendable?

It's technically only holding unmutable stuff in it, and I think the Protocol is also @unchecked Sendable. So the proxy should keep its sendability.

ElementX/Sources/Services/Authentication/LinkNewDeviceService.swift
```diff
@@ -256,6 +256,34 @@ nonisolated class CheckCodeSenderProxy: Equatable {
...
+    static func == (lhs: ContinuationMessageSenderProxy, rhs: ContinuationMessageSenderProxy) -> Bool {
+        lhs.underlyingSender === rhs.underlyingSender
+    }
```
Velin92, line 264: Uhm not a fan... why we need this? We are assuming two objects are equal when they are the same in reference (so they are the same object)... not necessarily the same content. It's technically okay, but do we actually 100% need for them to Equatable at this point? Since instead of requiring Equatability seems more like we need to require a singular instance.

Velin92 (changes requested):

ElementX/Sources/Screens/QRCodeLoginScreen/QRCodeLoginScreenViewModel.swift
pixlwave: This is the mock state, to make sure we don't time out in e.g. previews. The actual value is that default `120` in the real `init` above.

ElementX/Sources/Screens/QRCodeLoginScreen/QRCodeLoginScreenViewModel.swift
pixlwave: Fair, I thought it was handled elsewhere but found one instance where it wouldn't be.

ElementX/Sources/Screens/QRCodeLoginScreen/QRCodeLoginScreenViewModel.swift
pixlwave: Hey, those were actually all my own words that I spent a long time re-wording to actually make sense! 😂

But yeah fair, shortened it 👍

ElementX/Sources/Services/Authentication/LinkNewDeviceService.swift
pixlwave: Updated the `CheckCodeSenderProxy` too to match 👍

ElementX/Sources/Services/Authentication/LinkNewDeviceService.swift
pixlwave: Sadly we need the `Equatable` in order to `removeDuplicates` on the progress publishers. Have added comments to both to note why.

Velin92: Couldn't we just avoid having duplicates of the same undelrying sender in the first place?

Velin92 (approved):

ElementX/Sources/Services/Authentication/LinkNewDeviceService.swift
pixlwave: We discussed and seems this is similar to the login flow, where the QR scanner can potentially trigger duplicates after having first scanned.

PR #5976, opened 2026-08-04
Accessibility tests fixes

stefanceriu:


AccessibilityTests/SupportingFiles/AccessibilityTests.xctestplan
```diff
@@ -21,7 +21,7 @@
         }
       ]
     },
-    "defaultTestExecutionTimeAllowance" : 120,
+    "defaultTestExecutionTimeAllowance" : 240,
```
pixlwave, line 24: Each individual test shouldn't take 4 minutes surely?! 🙈

ElementX/Sources/Other/SwiftUI/Views/AvatarSettingsButtonLabel.swift
```diff
@@ -18,9 +18,9 @@ struct AvatarSettingsButtonLabel: View {
                             contentID: userProfile.id,
                             avatarSize: .user(on: .chats),
                             mediaProvider: mediaProvider)
-            .accessibilityIdentifier(A11yIdentifiers.homeScreen.userAvatar)
```
pixlwave, line 21: My bad, I assumed the outer label would trump this Image end of story… 🤦‍♂️

pixlwave (approved):

AccessibilityTests/SupportingFiles/AccessibilityTests.xctestplan
stefanceriu: Well not but the CI is sometimes stupidly so and aborts the whole thing

ElementX/Sources/Other/SwiftUI/Views/AvatarSettingsButtonLabel.swift
stefanceriu: No worries, that's why we have these tests

PR #5977, opened 2026-08-04
Disable the store pausing feature as we encountered cases in which it doesn't work properly

stefanceriu:
https://github.com/element-hq/element-x-ios/issues/5973


pixlwave (approved):
Is it worth providing a custom key too, so that it disables for anyone who had played with the toggle?

stefanceriu:
> Is it worth providing a custom key too, so that it disables for anyone who had played with the toggle?

Good idea, yeah! Pushed 👍

pixlwave:
Heh I see now it is `volatile: true` anyway, so it probably didn't matter.

PR #5980, opened 2026-08-04
Add hooks for the RoomDetails, RoomMemberDetails and UserProfile screens.

pixlwave:
Also tidies up the `AppHooks` file with a new `@AppHook` macro to remove the repetitive boilerplate. The history should make sense to review commit-by-commit if wanted (but not really necessary).


Velin92 (approved):
LGTM this is very cool.

PR #5983, opened 2026-08-05
Remove map tiler url from matrix wellknown

Velin92:


ElementX/Sources/Other/MapLibre/MapURLs.swift
```diff
@@ -27,47 +22,30 @@ extension MapTilerSettings: MapTilerURLBuilderProtocol {
...
+        return url
     }
 }
 
 // MARK: - Private
 
-private extension MapTilerSettings.Configuration {
+private extension MapTilerConfiguration {
     func styleURL(for style: MapTilerStyle) -> URL? {
         guard let apiKey else { return nil }
```
stefanceriu, line 42: Perhaps guarding here for empty is better? As I understand it the style won't work on mismatching API keys, will it?

stefanceriu (approved):
Looks great, just a small comment 👍

ElementX/Sources/Other/MapLibre/MapURLs.swift
Velin92: empty is fine, nil is disabled.
There are some cases where a base URL might internally proxy the apiKey or even use some self hosting solution where the apiKey is not required at all. So an empty key works while a nil one considers everything as disabled. It's mentioned in the config docs.

PR #5984, opened 2026-08-05
Add an HTTP_PROXY environment variable too (also disabled).

pixlwave:
We already have an `HTTPS_PROXY` environment variable available in the EX scheme, this PR adds an `HTTP_PROXY` which is helpful when testing against a servers that doesn't have TLS configured.

### DIFF
--- a/ElementX.xcodeproj/project.pbxproj
+++ b/ElementX.xcodeproj/project.pbxproj
@@ -7660,16 +7660,16 @@
 			projectDirPath = "";
 			projectRoot = "";
 			targets = (
-				C0C687DE1D270F9895FEE186 /* AccessibilityTests */,
 				C0FAEB81CFD9776CD78CE489 /* ElementX */,
-				D3DB351B7FBE0F49649171FC /* IntegrationTests */,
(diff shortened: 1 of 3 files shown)


ElementX.xcodeproj/project.pbxproj
pixlwave, line 1: Not sure what's going on here, my `xcodegen` did this though 🤷‍♂️

ElementX/SupportingFiles/target.yml
```diff
@@ -22,6 +22,9 @@ schemes:
         - variable: HTTPS_PROXY
           value: localhost:9090
           isEnabled: false
+        - variable: HTTP_PROXY
+          value: localhost:9090
```
pixlwave, line 26: Tested with Proxyman and it is happy to accept HTTP over the same port as HTTPS.

ElementX.xcodeproj/project.pbxproj
Velin92: maybe a version mismatch, that has slightly changed the order of stuff when generated?

Velin92 (approved):

PR #5986, opened 2026-08-06
Update the SDK and handle some API breaks.

pixlwave:
- There's a new trace log pack for Search.
- `clearCallStatus` has been removed (as `clearUserStatus` clears both).


stefanceriu (approved):
🙌

PR #5992, opened 2026-08-07
Set the finalised OAuth Client URI path to match the new EXI page.

pixlwave:
The page is live at https://element.io/apps/ios so we can now set this value to point at it.


stefanceriu (approved):
⭐

PR #5993, opened 2026-08-07
Show loading/failure states when setting/clearing user status.

pixlwave:
Nothing complex. An indeterminate progress toast while saving and a failure toast if it fails:

https://github.com/user-attachments/assets/5bee7c3f-58e1-4184-a338-1a35d21110ca


stefanceriu (approved):
Coolio 👍

PR #5994, opened 2026-08-07
Enable automatic call status under the User Status feature flag.

### LINKED ISSUE #5648: [Task] Use `m.call` for the user's status (instead of the `m.status`) when set.
In theory Element Call will be the one that sets `m.call` so we don't need to worry about that.

pixlwave:
Uses https://github.com/matrix-org/matrix-rust-sdk/pull/6825 to close https://github.com/element-hq/element-x-ios/issues/5648.

### DIFF
--- a/ElementX/Sources/Services/Client/ClientProxy.swift
+++ b/ElementX/Sources/Services/Client/ClientProxy.swift
@@ -310,6 +310,13 @@ class ClientProxy: ClientProxyProtocol {
             mediaPreviewConfigListenerTaskHandle = await createMediaPreviewConfigObserver()
         }
         
+        if appSettings.userStatusEnabled {
+            Task {
+                guard case .success(true) = await isUserStatusSupported() else { return }
+                client.enableAutomaticCallStatus(enabled: true)
+            }
+        }
+        
         liveLocationOwnInfoUpdatesListenerTaskHandle = createLiveLocationOwnInfoUpdatesObserver()
     }
     


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -310,6 +310,13 @@ class ClientProxy: ClientProxyProtocol {
             mediaPreviewConfigListenerTaskHandle = await createMediaPreviewConfigObserver()
         }
         
+        if appSettings.userStatusEnabled {
+            Task {
+                guard case .success(true) = await isUserStatusSupported() else { return }
+                client.enableAutomaticCallStatus(enabled: true)
```
stefanceriu, line 316: You could say that it's a tad weird that we do this at the very bottom but the `enableAutomaticBackpagination` at the very top. That one's weird though, error prone, reason why I'm replacing it with a ClientBuiler solution in the automatic back pagination SDK implementation.

stefanceriu (approved):

ElementX/Sources/Services/Client/ClientProxy.swift
pixlwave: Yeah, so I did actually start with this at the very top, but once I realised it was `async` to figure out server support (and that using `self.isSupported` for this was better for logs compared to `client.isSupported`) I just went looking for wherever we spawned other tasks and landed here.

PR #5996, opened 2026-08-08
Fix tracing span panic when accessed from multiple threads

stefanceriu:
L.E. Decided to remove the whole thing in the end because it's easier to live without it instead of dealing with the with the API redesign and/or misuse.

---

Brings it in sync with the TimelineItemProvider as it happens.

Fixes https://sentry.tools.element.io/organizations/element/issues/11257907/?project=63&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream

> assertion `left != right` failed: tried to clone a span (Id(790382009481428994)) that already closed

### DIFF
--- a/ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
+++ b/ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
@@ -251,15 +251,6 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
             return currentItems
         }
         
-        // The span guard is thread local, so it must not straddle a suspension
-        // point, otherwise the task could resume on a different thread and
-        // unbalance it, causing a panic and subsequent (intentional) crash.
-        let span = MXLog.createSpan("\(name).apply_room_list_diff")
-        span.enter()
-        defer {
-            span.exit()
-        }
-        
         guard let updatedItems = currentItems.applying(collectionDiff) else {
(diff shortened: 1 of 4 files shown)


ElementX/Sources/Services/Room/RoomSummary/RoomSummaryProvider.swift
```diff
@@ -257,6 +251,15 @@ class RoomSummaryProvider: RoomSummaryProviderProtocol {
             return currentItems
         }
         
+        // The span guard is thread local, so it must not straddle a suspension
+        // point, otherwise the task could resume on a different thread and
+        // unbalance it, causing a panic and subsequent (intentional) crash.
```
stefanceriu: I wonder if we should move this to the function documentation. Even then it would still be somewhat footgun-y

pixlwave: One other option is to restructure how `createSpan` works. We could try things like giving it a sync closure:

```swift
MXLog.withSpan("\(name).apply_room_list_diff") {
    let updatedItems = currentItems.applying(collectionDiff) // fine
    await something() // error: 'async' call in a function that does not support concurrency
}
```

Or TIL you can mark a function as `@available(*, noasync)`: https://www.hackingwithswift.com/swift/5.7/noasync

Both would require restructuring things but it would definitely be better than relying on documentation alone.

pixlwave (approved):

PR #6000, opened 2026-08-10
Improve integration test diagnostics in an attempt to catch run crashes

stefanceriu:


Tools/Sources/Commands/CI/IntegrationTests.swift
pixlwave, line 1: Smart! 😎

pixlwave (approved):

PR #6007, opened 2026-08-11
Reinforce text typing within XCUITests

stefanceriu:


pixlwave (approved):
Seems worth a shot but worth noting that this is now basically what the original implementation looked like (albeit with 3 iterations instead of 10):

https://github.com/element-hq/element-x-ios/pull/370/changes#diff-612a19b5482d7256f00a236b0c4715682ce55244f9d28dab8cd1205f99183334

PR #6008, opened 2026-08-11
Drop the `addressSanitizer` and `undefinedBehaviorSanitizer` from the Accessibility and UI tests

stefanceriu:
They cost a lot of extra memory to enable and cause problems on constrained CIs


pixlwave (approved):
Not sure I fully understand the rationale here, but I see no issue with it 👍

stefanceriu:
> Not sure I fully understand the rationale here, but I see no issue with it 👍

Sorry, it didn't pick up the commit description. It basically uses less memory and will maker the CI flake less.

PR #6009, opened 2026-08-11
Fix a simulcast bug in Element Call.

### LINKED ISSUE #6005: Masquerade as Safari so simulcast is supported
https://github.com/element-hq/element-x-ios/pull/4973/changes introduces a regression in livekit-js-sdk feature detection which is based in UA sniffing. Since element-x-ios sends an unknown UA, it turns off features that the embedded safari browser likely supports. 

I ran into this issue because video chatting was very laggy on a connection that works fine in other applications. I found that simulcast was off and required a lot of network bandwidth to support the first chosen case which was like 720@30hz

I tested simulcast by using "element call remote URL override" to enable simulcast with the altered code.  I tried opening an upstream issue about this but there's some resistance to enabling features by default on unknown UAs. https://github.com/livekit/client-sdk-js/issues/2026#issuecomment-5140393952

I would like to append application information to the UA instead of replacing it like the above PR does. The patch looks likes this:

```
-            webView.customUserAgent = UserAgentBuilder.makeASCIIUserAgent()
+            configuration.applicationNameForUserAgent = UserAgentBuilder.makeASCIIUserAgent()
```

This keeps the safari token that livekit is sniffing while also adding application information for logging purposes which is what I think the original PR was intended for.

Well I considered modifying livekit js sdk... but that's not great. I also use the remote URL override which is great but requires downloading 40mb on every call which is less great.

**Additional context**
Here's in the upstream repo where they sniff out ios and safari https://github.com/livekit/client-sdk-js/blob/v2.19.2/src/room/utils.ts#L235-L256 
If we satisfy these checks, it should turn on safari features for the browser that is already being used

pixlwave:
Sets Element Call's user-agent back to the default but including the bundle name.

Apparently livekit-js-sdk does feature detection based on UA sniffing.

Fixes #6005.

### DIFF
--- a/ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
+++ b/ElementX/Sources/Screens/CallScreen/View/CallScreen.swift
@@ -117,6 +117,7 @@ struct CallView: UIViewRepresentable {
             configuration.userContentController = userContentController
             configuration.allowsInlineMediaPlayback = true
             configuration.allowsPictureInPictureMediaPlayback = true
+            configuration.applicationNameForUserAgent = InfoPlistReader.main.bundleDisplayName
             
             if let script = viewModelContext.viewState.script {
                 let userScript = WKUserScript(source: script, injectionTime: .atDocumentEnd, forMainFrameOnly: false)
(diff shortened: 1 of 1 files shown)


stefanceriu (approved):
👌

drewwells:
Thanks ! 😀

PR #6012, opened 2026-08-12
Load MapLibre lazily instead of at every app launch

stefanceriu:
MapLibre ships as a dynamic framework and the app links it, which means dyld loads it before `main` runs on every launch: 7.3 MB paged in, 133 C++ static initialisers run, a hundred-odd classes registered. None of that is needed unless someone opens a location screen, which most sessions never do.

Moved the map components into a new `MapLibreShim` framework which is embedded in the bundle but deliberately left unlinked, and is the only image that links MapLibre. The app links a second `MapLibreInterface` framework that holds the handful of types that cross the boundary.

> Measured it with two otherwise identical probe apps on an iPhone 17 Pro, launched alternately. Linking MapLibre takes pre-main from 11.1 ms to 23.2 ms, and on a cold launch from 32.9 ms to 60.7 ms. Loading it on demand instead brings pre-main back to 11.1 ms, and the ~11 ms it costs to `dlopen` lands when the map first appears. Numbers are most likely significantly worse on older silicon.


ara4n:
ftr on older silicon (iPhone 12 Pro Max on iOS 26.5) this took the pre-main launch time from 480ms to 80ms, hence the original motivation :)

pixlwave (approved):
We talked about introducing a new `Components` directory for modules, variants, dev assets, secrets etc to help on the project organisation level, but for a future PR.

Otherwise sounds good to me 👍

stefanceriu:
> We talked about introducing a new `Components` directory for modules, variants, dev assets, secrets etc to help on the project organisation level

Now done in https://github.com/element-hq/element-x-ios/pull/6014

PR #6014, opened 2026-08-12
Move various top level folder to the `Components` one

stefanceriu:
In order to clean up the project's root.


project.yml
```diff
@@ -24,9 +24,7 @@ options:
     - UITests
     - AccessibilityTests
     - IntegrationTests
-    - DevelopmentAssets
-    - SDKMocks
-    - Secrets
+    - Components
```
pixlwave, line 27: No strong opinion here, but pointing out there's an argument to put `Components` above the tests.

project.yml
```diff
@@ -59,12 +57,12 @@ settings:
...
 - path: app.yml
-- path: ElementX/MapLibre/MapLibreInterface/SupportingFiles/target.yml
-- path: ElementX/MapLibre/MapLibreShim/SupportingFiles/target.yml
+- path: Components/MapLibre/MapLibreInterface/SupportingFiles/target.yml
+- path: Components/MapLibre/MapLibreShim/SupportingFiles/target.yml
 - path: ElementX/SupportingFiles/target.yml
 - path: NSE/SupportingFiles/target.yml
 - path: ShareExtension/SupportingFiles/target.yml
-- path: SDKMocks/SupportingFiles/target.yml
+- path: Components/SDKMocks/SupportingFiles/target.yml
```
pixlwave: And then put all the `Componets/` here so the order is consistent.

pixlwave (approved):
👏

project.yml
stefanceriu: I think it's correct this way, with targets having more priority being the reason why we moved things into components to being with.

project.yml
stefanceriu: Like so 027d4396963f31303d27fa5b5f754532a585cb7d? (assuming you're okay with my previous comment)

project.yml
pixlwave: Fair 👍

project.yml
pixlwave: Yep LGTM 👍

PR #6016, opened 2026-08-13
Avoid re-applying identical room list filters

stefanceriu:
The home screen's filter observation fires on every view state change, so setFilter was repeatedly pushing the same filter down into the SDK. Cache the last applied filter and skip the SDK call when it hasn't changed.


pixlwave (approved):
Makes sense to me.

pixlwave:
Maybe we should also remove duplicates from the home screen view state filter changes too for correctness?

stefanceriu:
> we should also remove duplicates from the home screen view

Already taken care of in https://github.com/element-hq/element-x-ios/blob/5cb4a2fe5ac073a73d8a11a9900bf37b95130fbc/ElementX/Sources/Screens/HomeScreen/HomeScreenViewModel.swift#L162-L166

PR #6017, opened 2026-08-13
Avoid race conditions between the home screen's crash alert and Sentry's `onLastRunStatusDetermined`

stefanceriu:
By converting the `lastCrashEventID` to a publisher that the VM can just listen to instead of relying on start to match a pre-set event


pixlwave (approved):
Nice, makes sense 👍

PR #6022, opened 2026-08-14
Fix the flaky ElementCallServiceTests by replacing TestClock

stefanceriu:
The suite intermittently hung on CI, most recently for the full 2 minute time limit on lifetimeIsCapped(). A spindump from a failed run shows the process suspended rather than spinning (0.247s of CPU over 190s) with a background priority worker thread sitting runnable for roughly 145 seconds without ever being scheduled.

That band is exactly what swift-clocks' TestClock depends on to make progress: both advance() and checkSuspension() await Task.megaYield(), which serially awaits 20 detached tasks at background priority. On an oversubscribed runner those never run, so the test waits forever instead of failing. The same mechanism, less starved, is why two of the passing tests took 1.0s and 0.77s for what is otherwise pure logic. checkSuspension() made it worse, since asking whether a timer was armed cost a megaYield per poll.

ManualClock replaces it. Sleeps register a continuation, advance(by:) resumes the ones that are due synchronously, and nextScheduledSleep() suspends until a timer is armed instead of polling for it. Nothing in the path touches background priority and no loop's iteration count depends on the scheduler.

### DIFF
--- a/UnitTests/Sources/TestUtilities/ManualClock.swift
+++ b/UnitTests/Sources/TestUtilities/ManualClock.swift
@@ -1,0 +1,151 @@
+//
+// Copyright 2026 Element Creations Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Synchronization
+
+/// A clock whose sleeps only finish when the test advances it.
+final nonisolated class ManualClock: Clock {
+    struct Instant: InstantProtocol {
+        let offset: Duration
+        
+        func advanced(by duration: Duration) -> Self {
+            Self(offset: offset + duration)
+        }
+        
+        func duration(to other: Self) -> Duration {
+            other.offset - offset
+        }
+        
+        static func < (lhs: Self, rhs: Self) -> Bool {
+            lhs.offset < rhs.offset
+        }
+    }
+    
+    private struct Sleep {
+        let deadline: Instant
+        let continuation: CheckedContinuation<Void, any Error>
+    }
+    
+    private struct State {
+        var now = Instant(offset: .zero)
+        /// A test steps through one timer at a time, so a single sleep can be pending.
+        var sleep: Sleep?
+        var sleepObserver: CheckedContinuation<Duration, Never>?
+    }
+    
+    private let state = Mutex(State())
+    
+    let minimumResolution: Duration = .zero
+    
+    var now: Instant {
+        state.withLock { $0.now }
+    }
+    
+    /// Suspends until ``advance(by:)`` moves the clock to `deadline`, or the calling task is
+    /// cancelled, in which case this throws a `CancellationError`. Deadlines that have already
+    /// passed return without suspending. `tolerance` is ignored.
+    func sleep(until deadline: Instant, tolerance: Duration? = nil) async throws {
+        try await withTaskCancellationHandler {
+            try Task.checkCancellation()
+            
+            try await withCheckedThrowingContinuation { continuation in
+                switch schedule(deadline: deadline, continuation: continuation) {
+                case .elapsed:
+                    continuation.resume()
+                case .scheduled(let duration, let observer):
+                    observer?.resume(returning: duration)
+                    
+                    // Cancellation that landed while scheduling found nothing to cancel.
+                    if Task.isCancelled {
+                        cancelSleep()
+                    }
+                }
+            }
+        } onCancel: {
+            cancelSleep()
+        }
+    }
+    
+    /// Suspends until a sleep is pending and returns how long it has left to run. A sleep that is
+    /// already pending returns straight away, so a scheduled sleep can't be missed.
+    ///
+    /// Await this before advancing a timer that runs in an unstructured task — advancing the clock
+    /// before the sleep has been scheduled would leave it sleeping past the new deadline.
+    @discardableResult
+    func waitForScheduledSleep() async -> Duration {
+        await withCheckedContinuation { continuation in
+            let duration = state.withLock { state -> Duration? in
+                guard let sleep = state.sleep else {
+                    state.sleepObserver = continuation
+                    return nil
+                }
+                
+                return state.now.duration(to: sleep.deadline)
+            }
+            
+            if let duration {
+                continuation.resume(returning: duration)
+            }
+        }
+    }
+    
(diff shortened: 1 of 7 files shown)


UnitTests/Sources/TestUtilities/ManualClock.swift
```diff
@@ -0,0 +1,163 @@
...
+//
+
+import Foundation
+import Synchronization
+
+/// A clock whose sleeps only finish when the test advances it.
+///
+/// This is used instead of swift-clocks' `TestClock` because both `advance` and `checkSuspension`
+/// await `Task.megaYield()`, which hops through background priority tasks. Busy CI runners starve
+/// that priority band for minutes at a time, hanging the test instead of running it.
```
pixlwave: I don't think this has any relevance…

UnitTests/Sources/TestUtilities/ManualClock.swift
```diff
@@ -0,0 +1,163 @@
...
+        /// Sleeps cancelled before their continuation was installed.
+        var cancelledSleeps: Set<UUID> = []
+        /// Durations of the sleeps that have been scheduled but not yet observed by the test.
+        var scheduledDurations: [Duration] = []
+        var observers: [UnsafeContinuation<Duration, Never>] = []
+    }
+    
+    private let state = Mutex(State())
+    
+    let minimumResolution: Duration = .zero
```
pixlwave, line 42: This is unused I think?

UnitTests/Sources/TestUtilities/ManualClock.swift
```diff
@@ -0,0 +1,163 @@
...
+    /// Suspends until ``advance(by:)`` moves the clock to `deadline`, or the calling task is
+    /// cancelled, in which case this throws a `CancellationError`. Deadlines that have already
+    /// passed return without suspending. `tolerance` is ignored, the clock is exact.
+    func sleep(until deadline: Instant, tolerance: Duration? = nil) async throws {
+        let id = UUID()
+        
+        try await withTaskCancellationHandler {
+            try Task.checkCancellation()
+            
+            try await withUnsafeThrowingContinuation { continuation in
```
pixlwave: I know it's just a test, but does this really need to be an unsafe continuation?

UnitTests/Sources/TestUtilities/ManualClock.swift
```diff
@@ -0,0 +1,163 @@
...
+    }
+    
+    /// Suspends until a sleep has been scheduled and returns the duration it was scheduled for.
+    /// Sleeps scheduled before this call are returned in order, so none of them are missed.
+    ///
+    /// Await this before advancing a timer that runs in an unstructured task — advancing the clock
+    /// before the sleep has been scheduled would leave it sleeping past the new deadline.
+    @discardableResult
+    func nextScheduledSleep() async -> Duration {
+        await withUnsafeContinuation { continuation in
```
pixlwave: And this?

UnitTests/Sources/TestUtilities/ManualClock.swift
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+    
+    /// Suspends until a sleep has been scheduled and returns the duration it was scheduled for.
+    /// Sleeps scheduled before this call are returned in order, so none of them are missed.
+    ///
+    /// Await this before advancing a timer that runs in an unstructured task — advancing the clock
+    /// before the sleep has been scheduled would leave it sleeping past the new deadline.
+    @discardableResult
+    func nextScheduledSleep() async -> Duration {
```
pixlwave: I can't think of an answer, but based on looking at the name of this at the call site it wasn't clear to me what it was doing (I guess partly because the `@discardableResult` means it can have 2 different uses).

UnitTests/Sources/TestUtilities/ManualClock.swift
```diff
@@ -0,0 +1,163 @@
...
+    }
+    
+    private struct Sleep {
+        let deadline: Instant
+        let continuation: UnsafeContinuation<Void, any Error>
+    }
+    
+    private struct State {
+        var now = Instant(offset: .zero)
+        var sleeps: [UUID: Sleep] = [:]
```
pixlwave: It might just be me, but this also seems a little over-engineered to me too. As in, why would you want to call

```
clock.sleep(until: .someTime)
clock.sleep(until: .someLaterTime)
clock.advanceBy(.someDuration)
```

in a test? I don't see the advantage of scheduling multiple sleeps, wouldn't want to walk step-by-step through the test situation?

pixlwave (approved):
Looks fine to me although I can't say I fully grok the ManualClock implementation, would be nice if was possible to simplify it a bit more.

stefanceriu:
I've asked claude to take care of your suggestions, it seems to have agreed 😅

UnitTests/Sources/TestUtilities/ManualClock.swift
```diff
@@ -5,14 +5,13 @@
...
 import Synchronization
 
 /// A clock whose sleeps only finish when the test advances it.
 ///
-/// This is used instead of swift-clocks' `TestClock` because both `advance` and `checkSuspension`
-/// await `Task.megaYield()`, which hops through background priority tasks. Busy CI runners starve
-/// that priority band for minutes at a time, hanging the test instead of running it.
+/// This replaces swift-clocks' `TestClock`, whose `advance` awaits `Task.megaYield()` and so only
+/// makes progress once background priority tasks are scheduled. Busy CI runners starve that priority
+/// band for minutes at a time, hanging the test instead of running it.
```
pixlwave: Lol Claude, just remove it 😆

UnitTests/Sources/TestUtilities/ManualClock.swift
```diff
@@ -0,0 +1,156 @@
...
+    private struct State {
+        var now = Instant(offset: .zero)
+        /// A test steps through one timer at a time, so a single sleep can be pending.
+        var sleep: Sleep?
+        var sleepObserver: CheckedContinuation<Duration, Never>?
+    }
+    
+    private let state = Mutex(State())
+    
+    /// Required by `Clock`. The clock is exact, it has no resolution to speak of.
```
pixlwave: I think this could've been a reply to my comment rather than documentation 🙃

UnitTests/Sources/TestUtilities/ManualClock.swift
stefanceriu: Protocol requirement.

PR #6030, opened 2026-08-14
Update the SDK.

pixlwave:
Also use the new `Client.isProfilesSlidingSyncExtensionSupported` to check for user profile subscription support so that the user's profile is still reactive if the homeserver disables the User Status feature.


stefanceriu (approved):
👌

PR #6037, opened 2026-08-17
Update the SDK fixing some user status bugs.

pixlwave:
- Room heroes weren't being updated when their status changed.
- A user's own profile would go missing after removing their avatar (or setting any field to `null`).


stefanceriu (approved):
Yey for bug fixes 🎉

PR #6038, opened 2026-08-17
Delay secondary room summary provider setup

stefanceriu:
To avoid creating room summaries for them on startup, especially as the static one uses a page size as big as the whole account.

### DIFF
--- a/ElementX/Sources/Services/Client/ClientProxy.swift
+++ b/ElementX/Sources/Services/Client/ClientProxy.swift
@@ -1571,13 +1571,11 @@ private struct ClientProxyServices {
                                                         notificationSettings: notificationSettings,
                                                         appSettings: appSettings)
         
-        // Delay alternate and static provider full setup until the main one as
-        // they have no consumers until after the app is fully running.
-        // Avoid creating room summaries for all of them on startup.
+        // Setting a provider's room list will create summaries for every room so
+        // wait until the app is fully running for the alternate and static providers.
         Task { [roomSummaryProvider, alternateRoomSummaryProvider, staticRoomSummaryProvider] in
-            // Wait for actual content (or a loaded-but-empty account): the loading state
-            // alone reports loaded straight from the cache, before the first summaries
-            // have been built and published.
+            // Wait for actual content (or a loaded-but-empty account) as the loading state
+            // doesn't take into account the app having build and published any summaries.
             for await rooms in roomSummaryProvider.roomListPublisher.values {
                 if !rooms.isEmpty {
                     break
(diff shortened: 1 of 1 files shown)


ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -1563,15 +1563,38 @@ private struct ClientProxyServices {
...
                                                         eventStringBuilder: eventStringBuilder,
                                                         name: "StaticAllRooms",
                                                         roomListPageSize: .max,
                                                         notificationSettings: notificationSettings,
                                                         appSettings: appSettings)
-        try await staticRoomSummaryProvider.setRoomList(roomListService.allRooms())
+        
+        // Delay alternate and static provider full setup until the main one as
+        // they have no consumers until after the app is fully running.
+        // Avoid creating room summaries for all of them on startup.
```
pixlwave:
```suggestion
        // Setting a provider's room list will create summaries for every room so
        // wait until the app is fully running for the alternate and static providers.
```

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -1563,15 +1563,38 @@ private struct ClientProxyServices {
...
+        Task { [roomSummaryProvider, alternateRoomSummaryProvider, staticRoomSummaryProvider] in
+            // Wait for actual content (or a loaded-but-empty account): the loading state
+            // alone reports loaded straight from the cache, before the first summaries
+            // have been built and published.
```
pixlwave:
```suggestion
            // Wait for actual content (or a loaded-but-empty account) as the loading state
            // doesn't take into account the app having build and published any summaries.
```

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -1563,15 +1563,38 @@ private struct ClientProxyServices {
...
+                if case .loaded(0) = roomSummaryProvider.statePublisher.value {
+                    break
+                }
+            }
+            
+            do {
+                try await alternateRoomSummaryProvider.setRoomList(roomListService.allRooms())
+                try await staticRoomSummaryProvider.setRoomList(roomListService.allRooms())
+            } catch {
+                MXLog.error("Failed setting up the deferred room summary providers: \(error)")
```
pixlwave: FWIW, previously these throws would have resulted in a `UserSessionStoreError.failedSettingUpClientProxy` error being handled as a `fatalError`. Maybe we should do the same as part of the app's functionality will be quietly broken otherwise?

pixlwave (approved):
Nice 😎

Just a couple of suggestions on the comments as they weren't particularly clear to me.

PR #6039, opened 2026-08-17
Fix the emoji picker's title when setting a custom status.

pixlwave:
The title now says "Custom emoji" like in the Figma.


stefanceriu (approved):
Noice, thanks!

PR #6041, opened 2026-08-18
Fail fast on the integration tests when encountering the first failed assert

stefanceriu:
And especially don't keep going when login fails as if everything is in order.


pixlwave (approved):
Oh yeah makes sense 👍

PR #6043, opened 2026-08-18
Update the toolbars in the Chats/Spaces screens.

pixlwave:
- The avatar buttons are now full-sized buttons on iOS 26 without the glass.
    - Fixes a bug with User Status where the status emoji would be cropped inside the glass button.
- The large navigation title has been collapsed into the inline title (with a slight reduction in font size).
- iOS 18 remains unchanged.
- The Search screen's title matches the font size of the Chats screen.
- The Spaces headers have reduced font sizes for when we show a title on the Spaces with the Orgs work.

I've also included a small fix to the BuildExtensions package as my previews were failing due to targeting iOS 12 apparently.


stefanceriu (approved):
Noice, looks great!

PR #6054, opened 2026-08-21
Use the `ClientFactory` pattern everywhere replacing raw `ClientBuilders`.

pixlwave:
Allows the use of a single `ClientFactoryHook` to handle both authentication and restoration.


stefanceriu (approved):
🤘

ElementX/Sources/Services/Client/ClientFactoryProtocol.swift
stefanceriu, line 1: This is just lovely! 🧑‍🍳 👌

stefanceriu (approved):
Looks great, thank you!

PR #6055, opened 2026-08-21
Attributed string builder tweaks

### LINKED ISSUE #6026: Messages using `<strike>` lose their strikethrough formatting
### Steps to reproduce

1. Send a message whose formatted body uses `<strike>`, for example from a client that emits it, or over a bridge.
2. Look at the message in the timeline.

#### What did you expect?

The text is struck through, the same as when the sender uses `<s>` or `<del>`.

#### What happened instead?

The text renders as ordinary text with no strikethrough and no indication that the sender formatted it at all. `<s>` and `<del>` both work, only `<strike>` is dropped.

`<strike>` is in the list of tags the Matrix specification permits in `org.matrix.custom.html` bodies, so discarding it loses information the sender deliberately included. The text itself is preserved, only the styling is lost.

Element X Android fixed the same gap: element-hq/element-x-android#7341.

### Your phone model

n/a

### Operating system version

n/a

### Application version

`develop` at c62aed4e

### Homeserver

n/a

### Will you send logs?

No

stefanceriu:
Fix the indentation for the first items in lists and add support for the `<strike>` tag

Fixes https://github.com/element-hq/element-x-ios/issues/6026


pixlwave (approved):
Hell yes for the indentation fix! 🙌

PR #6056, opened 2026-08-21
Fix missing back pagination attempt

stefanceriu:
Avoiding not populating past the initial events

### DIFF
--- a/ElementX/Sources/Screens/Timeline/TimelineTableViewController.swift
+++ b/ElementX/Sources/Screens/Timeline/TimelineTableViewController.swift
@@ -56,10 +56,6 @@ class TimelineTableViewController: UIViewController {
             
             applySnapshot()
             
-            if timelineItemsDictionary.isEmpty {
-                paginatePublisher.send()
-            }
-            
             sendLastVisibleItemReadReceipt()
         }
     }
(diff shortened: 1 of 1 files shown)


ElementX/Sources/Screens/Timeline/TimelineTableViewController.swift
```diff
@@ -390,6 +386,9 @@ class TimelineTableViewController: UIViewController {
         // Re-evaluate after the snapshot has been applied so the new layout is reflected.
         DispatchQueue.main.async { [weak self] in
             self?.updateReadMarkerVisibility()
+            
+            // The other paginate triggers can fire against the previous timeline's geometry.
```
stefanceriu: You won't like this comment but just so you understand 😁

pixlwave: Ha, how did you guess?! 😄

I'm sure it could just say

> Make sure we paginate with the final timeline geometry

pixlwave (approved):
Oh nice, I've been wondering what the problem was here for a while 🙌

ElementX/Sources/Screens/Timeline/TimelineTableViewController.swift
stefanceriu: Cool cool, exactly how I imagined this going 😁

PR #6060, opened 2026-08-24
Stop the integration tests from backgrounding the web authentication session

stefanceriu:
Tapping the SafariViewService app object makes XCUITest activate it as a standalone app. That backgrounds Element X, which hosts the web view, so the hosted scene goes to the background too and WebKit suspends it mid load. The sign in page is fetched successfully but never rendered, and the test times out waiting for the username field.

The tap dates from when the element was a web view inside the app, where it was a plain in app tap, so it no longer does what its comment describes.

### DIFF
--- a/IntegrationTests/Sources/Common.swift
+++ b/IntegrationTests/Sources/Common.swift
@@ -61,7 +61,6 @@ extension XCUIApplication {
         
         let webAuthenticationView = XCUIApplication(bundleIdentifier: "com.apple.SafariViewService")
         XCTAssertTrue(webAuthenticationView.waitForExistence(timeout: 10.0))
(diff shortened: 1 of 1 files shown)


pixlwave:
Sorry I hadn't realised this one was assigned to me. Let's see what this does first maybe?

https://github.com/element-hq/element-x-ios/actions/runs/32726147473

pixlwave (approved):
But yeah, if it fixes things then LGTM.

stefanceriu:
We shall see soon enough if it does anything 😅

pixlwave:
The run failed 😕 https://github.com/element-hq/element-x-ios/actions/runs/32726147473/job/97427759619#step:4:146

PR #6061, opened 2026-08-24
Preserve the blank lines typed after a list

stefanceriu:
The newline to break replacement only matched paragraph to paragraph, so a blank line after a list was dropped and the text following it glued to the last item. Only a following paragraph is matched, because without a blank line its text would have continued the last item instead of starting a paragraph of its own.

Follow up to https://github.com/element-hq/element-x-ios/pull/2224

### DIFF
--- a/ElementX/Sources/Other/Extensions/String.swift
+++ b/ElementX/Sources/Other/Extensions/String.swift
@@ -52,9 +52,10 @@ nonisolated extension String {
 }
 
 nonisolated extension String {
+    /// Drops stray new lines everywhere but paragraphs and lists when other paragraphs follow them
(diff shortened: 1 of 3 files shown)


pixlwave (approved):
🙌

pixlwave (commented):
Silly question (which is still sending on matrix.org), doesn't SwiftSoup have some kind of property that tells us which tags result in new lines. Could we use that here?

stefanceriu:
> Silly question (which is still sending on matrix.org), doesn't SwiftSoup have some kind of property that tells us which tags result in new lines. Could we use that here?

I honestly don't rememeber of any but we can come back to it if we do.

PR #6064, opened 2026-08-25
Fixed one accessibility test and an UI test snapshot

stefanceriu:


UITests/Sources/__Snapshots__/Application/userSessionScreen.testAcceptSpaceInvite-iPad-en-GB-6.png
stefanceriu, line 1: No idea why this decided to start failing just now ¯\_(ツ)_/¯

pixlwave: huh weird.

AccessibilityTests/Sources/AccessibilityTests.swift
pixlwave, line 1: 2 thoughts rather than ignoring the entire view should we either:
- Add isRunningTests to [this guard](https://github.com/element-hq/element-x-ios/blob/024b02324e907fc2562d8d432a7e7924f426e5dc/ElementX/Sources/Screens/Settings/SettingsScreen/View/SettingsScreenUserStatusRow.swift#L32)?
- Update the accessibility tests to ignore "the keyboard's empty prediction bar cells"?

Probably the latter makes more sense in the long run, we never care about those.

pixlwave (approved):
Looks fine to me but I'd prefer to fix the issue as suggested below rather than ignoring a whole view.

AccessibilityTests/Sources/AccessibilityTests.swift
stefanceriu: Okay, let's try 4d43ddb4e85f4998d4412492394a20295b30793b then -> https://github.com/element-hq/element-x-ios/actions/runs/32837748236

pixlwave: Nice! 🤩

stefanceriu: Heh, it passed, awesome!

PR #6065, opened 2026-08-25
Add tests for the UserSessionStore.

pixlwave:
Follow-on from #6054, the new client factory allows us to inject mock `Clients` so we can easily test the user session store now too!

The first commit adds the necessary SDK mock configurations and the second commit adds the tests.


stefanceriu (approved):
This is super super cool, nicely done! 👏

PR #6067, opened 2026-08-25
Update sdk to 26.08.25

Velin92:
name says it all

### DIFF
--- a/ElementX/Sources/Services/UserSession/UserSessionStore.swift
+++ b/ElementX/Sources/Services/UserSession/UserSessionStore.swift
@@ -142,7 +142,7 @@ class UserSessionStore: UserSessionStoreProtocol {
                     .passphrase(passphrase: credentials.restorationToken.passphrase))
             .withSearchIndexStore(path: credentials.restorationToken.sessionDirectories.dataPath,
                                   password: credentials.restorationToken.passphrase)
-            .username(username: credentials.userID)
+            .serverNameFromUserId(userId: credentials.userID)
             .homeserverUrl(url: homeserverURL)
         
         do {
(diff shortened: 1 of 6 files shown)


ElementX/Sources/Services/UserSession/UserSessionStore.swift
```diff
@@ -142,7 +142,7 @@ class UserSessionStore: UserSessionStoreProtocol {
                     .passphrase(passphrase: credentials.restorationToken.passphrase))
             .withSearchIndexStore(path: credentials.restorationToken.sessionDirectories.dataPath,
                                   password: credentials.restorationToken.passphrase)
-            .username(username: credentials.userID)
+            .serverNameFromUserId(userId: credentials.userID)
```
pixlwave: We don't need this, our `username` here was a no-op on build because of the `homeserverUrl`. The new one can be removed (serverName is about discovery but we've already discovered the server as we know its homeserverURL).

```suggestion
```

NSE/Sources/NSEUserSession.swift
```diff
@@ -75,7 +75,7 @@ final nonisolated class NSEUserSession: NSEUserSessionProtocol {
             .sqliteStore(config: .init(dataPath: credentials.restorationToken.sessionDirectories.dataPath,
                                        cachePath: credentials.restorationToken.sessionDirectories.cachePath)
                     .passphrase(passphrase: credentials.restorationToken.passphrase))
-            .username(username: credentials.userID)
+            .serverNameFromUserId(userId: credentials.userID)
```
pixlwave: Same

```suggestion
```

pixlwave (approved):
LGTM, just 2 things to remove based on the docs from https://github.com/matrix-org/matrix-rust-sdk/pull/6900

ElementX/Sources/Services/UserSession/UserSessionStore.swift
pixlwave: Also sorry, I did handle this in https://github.com/element-hq/element-x-ios/pull/6054 but decided not to merge until after the RC).

PR #6071, opened 2026-08-26
Allow copying blocked user id

### LINKED ISSUE #6028: No way to copy a blocked user's Matrix ID
### Steps to reproduce

1. Go to **Settings**, **Blocked users**.
2. Try to copy the Matrix ID of one of the listed users.

#### What did you expect?

Long pressing the entry copies that user's Matrix ID to the clipboard and confirms with "Copied to clipboard", the same gesture the user profile screen already offers.

#### What happened instead?

There is no way to get the ID out of the app. The Matrix ID is displayed on the row, but a long press does nothing and a tap opens the unblock confirmation instead. Copying it by hand means retyping it from the screen.

Someone who has blocked a user for abuse needs that ID to report them onwards, and the blocked users list is the one place it is still visible after the block.

Element X Android added the long press: element-hq/element-x-android#7372.

Related: element-hq/element-x-ios#2781 is about the same screen, where a long Matrix ID is truncated on the row. This one is about getting the ID out of the app at all.

### Your phone model

n/a

### Operating system version

n/a

### Application version

`develop` at c62aed4e

### Homeserver

n/a

### Will you send logs?

No

Velin92:
fixes #6028

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -14,6 +14,7 @@ PRs must follow rules. Prefer Xcode MCP tools over terminal commands.
 - **SwiftLint** (.swiftlint.yml) + **SwiftFormat** (.swiftformat) enforce style on build. Warnings show in Xcode. Build before **XcodeRefreshCodeIssuesInFile**.
 - Follow [Swift API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/) everywhere, Rust SDK wrappers too (`ID` not `Id`, `URL` not `Url`, `configuration` not `config` or `cfg`).
 - File headers live in `IDETemplateMacros.plist`.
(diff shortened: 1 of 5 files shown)


stefanceriu (approved):
👌

UnitTests/Sources/BlockedUsersScreenViewModelTests.swift
```diff
@@ -10,6 +10,7 @@ import Combine
 @testable import ElementX
 import Foundation
 import Testing
+import UIKit
```
pixlwave: Tiny nit. I think this should probably be a point in the CLAUDE.md to import SwiftUI instead of UIKit and that SwiftUI re-exports Foundation

```suggestion
import Testing
import SwiftUI
```

pixlwave (commented):
Wooo 🥳

AGENTS.md
pixlwave, line 1: Nice, thanks! 👌

PR #6072, opened 2026-08-26
Media view / Quick Look tweaks and improvements

stefanceriu:
Replacement for https://github.com/element-hq/element-x-ios/pull/6062 in which the per commit split was off.

### DIFF
--- a/ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
+++ b/ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
@@ -122,8 +122,9 @@ class TimelineMediaPreviewDataSource: NSObject, QLPreviewControllerDataSource {
             .flatMap { $0.previewableMedia(allowedGalleryItemTypes: allowedGalleryItemTypes) }
             .map { newItem in
                 // If an item already exists use that instead to preserve the file handle, download error etc.
-                if let oldItem = previewItems.first(where: { $0.id == newItem.id }) {
+                if let oldItem = previewItems.first(where: { $0.id == newItem.id || $0.isLocalEcho(of: newItem) }) {
                     oldItem.content = newItem.content
+                    oldItem.id = newItem.id // A sent local echo: transaction ID → event ID.
                     return oldItem
                 }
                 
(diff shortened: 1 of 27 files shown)


ElementX/Sources/Screens/Settings/AdvancedSettingsScreen/View/AdvancedSettingsScreen.swift
```diff
@@ -40,6 +40,10 @@ struct AdvancedSettingsScreen: View {
                     .onChange(of: context.optimizeMediaUploads) {
                         context.send(viewAction: .optimizeMediaUploadsChanged)
                     }
+                
+                ListRow(label: .plain(title: UntranslatedL10n.screenAdvancedSettingsPreloadMediaTitle,
```
pixlwave, line 44: I would think this should go in Developer Options for some time to test/bug-hunt in Nightly first right?

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
```diff
@@ -32,12 +32,20 @@ class TimelineMediaPreviewDataSource: NSObject, QLPreviewControllerDataSource {
     
     private var backwardPadding: Int
     private var forwardPadding: Int
+    /// Whether a real (post-load) pagination state has been seen. `.initial` is `endReached` on
+    /// both sides as a "don't paginate yet" sentinel, indistinguishable by value from a genuinely
```
pixlwave, line 36: Hmmm, I wonder if we should separately fix this with an `initial` case on `PaginationState` instead.

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
```diff
@@ -32,12 +32,20 @@ class TimelineMediaPreviewDataSource: NSObject, QLPreviewControllerDataSource {
...
+    /// exhausted small room, so we only collapse the phantom padding (below) once a real state has
+    /// arrived; until then the padding stays, keeping room to paginate.
+    private var hasReceivedRealPaginationState: Bool
```
pixlwave, line 39: Noting to myself that I'm surprised this doesn't need to do something to make sure swiping goes to the correct index.

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewDataSource.swift
```diff
@@ -267,8 +268,23 @@ enum TimelineMediaPreviewItem: Equatable {
...
         
         /// A stable identifier that's unique per preview item — including individual gallery
-        /// attachments that would otherwise share their parent event's ID.
-        let id: MediaPreviewItemID
+        /// attachments that would otherwise share their parent event's ID. Only changes when a
+        /// local echo is sent and its transaction ID gives way to the event ID.
+        fileprivate(set) var id: MediaPreviewItemID
+        
+        /// Whether this is the local echo that `sentItem` is the remote echo of: sending swaps the
+        /// transaction ID for an event ID (so `id` changes) but the timeline recycles the unique ID.
```
pixlwave, line 276: > but the timeline recycles the unique ID

Noting that I'm curious why we don't use the uniqueID alone as the `MediaPreviewItemID` instead… There's probably a reason but it certainly stands out as being wrong to me now.

pixlwave (commented):
Ok after an initial look (I haven't tested anything yet) I think we should split this PR into 3 (in this order):

- Bug fixes
    - [Only show a spinner when opening the media viewer is slow](https://github.com/element-hq/element-x-ios/pull/6072/changes/96c3ef83111cb1294189a9cf1f5233de7ea35419)
    - [Keep a sent local echo's page in the media viewer](https://github.com/element-hq/element-x-ios/pull/6072/changes/1188a87a562651f569c2e90446d974656ecce4a0)
    - [Keep the viewer's own chrome visible through QuickLook's churn](https://github.com/element-hq/element-x-ios/pull/6072/changes/8960da126d1d4bfcdccb7618827cec5d42e40d01)
- Proactive loading/padding fixes (the big one, should be clearly split between loading and then all the padding adjustments).
- Thumbnail support.

This will be easier to review and test the changes in isolation of each other.

PR #6076, opened 2026-08-27
Don't focus the user status text field during tests.

pixlwave:
Hopefully will fix the a11y test keyboard failure, but I can't reproduce it at all here 🤷‍♂️


stefanceriu (approved):
Works for me 👍

PR #6077, opened 2026-08-27
Add support for <details> and <summary> tags

### LINKED ISSUE #6070: Render `<summary>` and `<details>` tags
The matrix client-server spec (under `10.2.1.1 m.room.message msgtypes`) suggests that clients implement rendering of various tags, including `<summary>` and `<details>`. These tags are quite useful for bots, allowing details to be hidden without clogging an entire message. Sadly, element-x-ios doesn't render them and ignores them entirely.

stefanceriu:
Fixes #6070


ElementX/Sources/Other/Extensions/AttributedString.swift
```diff
@@ -17,23 +17,26 @@ nonisolated extension AttributedString {
     var formattedComponents: [AttributedStringBuilderComponent] {
         var components = [AttributedStringBuilderComponent]()
         
-        for run in runs[\.blockquote, \.codeBlock] {
+        for run in runs[\.blockquote, \.codeBlock, \.details] {
             let isBlockquote = run.0 != nil
             let isCodeBlock = run.1 != nil
-            var attributedString = AttributedString(self[run.2])
+            let summary = run.2
```
pixlwave: I think this should be `let details` no?

ElementX/Sources/Other/HTMLParsing/ElementXAttributeScope.swift
pixlwave, line 1: Should we group `blockquote`, `details`, `codeblock` and `inlineCode` together in this file, seems odd to have code so far apart from quotes/details.

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FormattedBodyText.swift
```diff
@@ -111,7 +121,7 @@ struct FormattedBodyText: View {
                     CodeBlockView(attributedString: component.attributedString, mode: .layout)
                         .timelineBubbleLayoutSize(.bubbleWidth(mode: .layout))
                         .hidden()
-                case .plainText:
+                case .details, .plainText:
```
pixlwave, line 124: Presumably we're happy that if the details contains wide content it doesn't expand the width of the bubble. We should probably document that here.

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FormattedBodyText.swift
```diff
@@ -50,10 +51,15 @@ struct FormattedBodyText: View {
...
+    private var containsDetails: Bool {
+        attributedString.runs[\.details].contains { $0.0 != nil }
+    }
+    
     var body: some View {
         layout
             .tint(.compound.textLinkExternal)
-            .accessibilityElement(children: .ignore)
+            .accessibilityElement(children: containsDetails ? .contain : .ignore)
             .accessibilityLabel(Text(attributedString))
```
pixlwave, line 63: This feels kind of odd to me. That you could swipe through the parts of the message only if there's a `<details>` in it. I'm wondering if we could expose an additional accessibility action alongside the one that opens the actions menu. It could i.e. expand all details (through the environment or something).

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FormattedBodyText.swift
```diff
@@ -157,6 +167,34 @@ struct FormattedBodyText: View {
...
+                    }
+                    .font(.compound.bodyLGSemibold)
+                    .foregroundStyle(.compound.textPrimary)
+                }
+                .buttonStyle(.plain)
+                
+                if isExpanded {
+                    MessageText(attributedString: attributedString)
+                        .fixedSize(horizontal: false, vertical: true)
+                }
```
pixlwave, line 193: Why a button and not a `DisclosureGroup`?

PreviewTests/Sources/__Snapshots__/PreviewTests/formattedBodyText.details-iPhone-en-GB.png
pixlwave, line 1: The chevron seems misaligned to me?

pixlwave (approved):
Very cool. Although given the conversation this morning, I wonder if we need to wait for Android support before merging this?

ElementX/Sources/Other/Extensions/AttributedString.swift
stefanceriu: The run is details but the associated value is a summay 🤪

ElementX/Sources/Other/HTMLParsing/ElementXAttributeScope.swift
stefanceriu: Yeah, makes perfect sense, no idea how the blockquote ended up on top 🤔

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FormattedBodyText.swift
stefanceriu: Not sure I follow, if it has details then it's all mixed together and read as one otherwise you just get the message? 🤔

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FormattedBodyText.swift
stefanceriu: Apparently it breaks the layout

> Tested it rather than guessing: `DisclosureGroup` breaks bubble sizing.
> A `DisclosureGroup` isn't used here as it takes all of the available width to push its chevron to the trailing edge, which would stretch the bubble beyond its natural width.

pixlwave: Even when you provide it with a label?

ElementX/Sources/Screens/Timeline/View/TimelineItemViews/FormattedBodyText.swift
pixlwave: We discussed this and I was misremembering what the second layout pass was for.

PR #6082, opened 2026-08-27
Use server name from user id function

### LINKED ISSUE #6027: Entering a full Matrix ID as the account provider is rejected
### Steps to reproduce

1. Start signing in and reach **Enter your account provider**.
2. Type a full Matrix ID, for example `@alice:example.org`.
3. Tap **Continue**.

#### What did you expect?

The homeserver is resolved from the server part of the ID, so `@alice:example.org` finds `example.org` and sign in continues.

#### What happened instead?

The field shows "We couldn't reach this homeserver. Please check that you have entered the homeserver URL correctly." and sign in cannot proceed. The whole ID is treated as a server address.

Typing a Matrix ID here is a natural mistake: it is the identifier people know and copy around, and the field asks for something they may never have thought about separately. There is nothing on screen to say the `@alice:` part has to be removed.

Element X Android accepts a full Matrix ID here, keeping any port: element-hq/element-x-android#7375.

### Your phone model

n/a

### Operating system version

n/a

### Application version

`develop` at c62aed4e

### Homeserver

n/a

### Will you send logs?

No

Velin92:
This PR does essentially 3 things:
- updates the sdk
- replaces all instances of trying to parse the server name from the user id with a dedicated sdk function
- allows to determine the server name from a user id, if a user inputs it in the server selection screen by mistake

fixes #6027

### DIFF
--- a/ElementX/Sources/Screens/Authentication/ServerSelectionScreen/ServerSelectionScreenViewModel.swift
+++ b/ElementX/Sources/Screens/Authentication/ServerSelectionScreen/ServerSelectionScreenViewModel.swift
@@ -7,6 +7,7 @@
 //
 
 import Combine
+import MatrixRustSDK
 import SwiftUI
 
 typealias ServerSelectionScreenViewModelType = StateStoreViewModelV2<ServerSelectionScreenViewState, ServerSelectionScreenViewAction>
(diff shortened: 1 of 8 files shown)


ElementX/Sources/Screens/Authentication/ServerSelectionScreen/ServerSelectionScreenViewModel.swift
```diff
@@ -101,7 +102,9 @@ class ServerSelectionScreenViewModel: ServerSelectionScreenViewModelType, Server
     
     /// Updates the login flow using the supplied homeserver address, or shows an error when this isn't possible.
     private func configureHomeserver() async {
-        let homeserverAddress = state.bindings.homeserverAddress
```
stefanceriu, line 104: Don't we need to get rid of the state binding or something? 🤔

stefanceriu (commented):
Good stuff 👍

ElementX/Sources/Screens/Authentication/ServerSelectionScreen/ServerSelectionScreenViewModel.swift
Velin92: uhm, no why would we? the line bellow checks the user input and uses it in case is not a user id (since the `serverNameFromUserId` function will fail), or will simply use it.

stefanceriu: Ah, I get it now, cool cool 👍

stefanceriu (approved):

pixlwave:
😎

PR #6089, opened 2026-08-31
Autoplay the next voice message if it is the next timeline item.

Velin92:
This PR implements autoplay for voice messages when they are close in the timeline. So for example if we have 3 voice messages one after another and we play the first one, once the first is fully played, the second one will start and so on...

It essentially works very similarly to the WA behaviour so...
- Autoplays only when the next timeline item is a voice message, if there is any other timeline item separating voice messages the autoplay won't continue.
- If the second message was already played but not finished (for example being at 0:10 duration) and we play the first message, once the first is finished and the second autoplays, it resumes from where it was left off at 0:10.
- A system sound plays between autoplays to inform the user that another voice message is now playing.

what is missing:
- I think WA also does a scroll (but this autoscroll behaviour is a bit inconsistent)
- background handling to stop autoplay (which means that as of right now, we keep autoplay also when backgrounding after a voice message has already been started), however this is the same behaviour as Signal so it's probably okay.

To review commit by commit

(EXAMPLE WITH NO SOUND)
https://github.com/user-attachments/assets/83495bc9-dee2-4ddd-8bfb-d3a804a2b8de


ElementX/Sources/Services/Audio/Player/AudioPlayerState.swift
```diff
@@ -128,7 +128,6 @@ class AudioPlayerState: ObservableObject, Identifiable {
     
     private func subscribeToAudioPlayer(audioPlayer: AudioPlayerProtocol) {
         audioPlayerSubscription = audioPlayer.actions
-            .receive(on: DispatchQueue.main)
```
Velin92, line 131: these actions are already sent on the main queue (since we now use MainActor isolation), so the receive here, just handled subscription on the next main thread update, which resulted in some weird behaviour with autoplay (like the first finished message not wrapping up to 0:0 before the second one starts)

bmarty:
I did not check recently, but I believe that WA plays a small sound between 2 voice messages

Velin92:
> I did not check recently, but I believe that WA plays a small sound between 2 voice messages

Yeah it does, should we play a system sound or some kind of custom sound? I guess this is a question for design/product @amshakal @Anderas

UnitTests/Sources/ArrayTests.swift
```diff
@@ -33,4 +34,84 @@ struct ArrayTests {
         
         #expect([0, 0, 0, 1, 0, 2, 3, 0, 0].groupBy { $0 == 0 } == [[0, 0, 0], [1], [0], [2], [3], [0, 0]])
     }
+    
+    // MARK: - Voice message autoplay
+    
+    @Test("The directly following voice message is found")
```
Velin92: we never used test descriptions, but I kinda like them

Velin92:
For now I have resolved in using a system sound called "Tink" which plays like a very quick metallic tick (like the one of a metronome).

ElementX/Sources/Other/Extensions/URL.swift
pixlwave, line 1: Could we maybe put this beneath e.g. `appGroupTemporaryDirectory` so all the app dirs stay at the top?

ElementX/Sources/Services/MediaPlayer/MediaPlayerProviderProtocol.swift
```diff
@@ -8,13 +8,28 @@
 
 import Foundation
 
+/// Short sounds played on top of whatever the app is already playing. They go through the app's
+/// own audio session, so unlike the sounds built into iOS they are heard in silent mode too.
```
pixlwave: This documentation seems more related to the play function than the sound effect file, no?

ElementX/Sources/Services/MediaPlayer/MediaPlayerProvider.swift
```diff
@@ -6,11 +6,14 @@
...
 //
 
+import AVFoundation
 import Foundation
 
 class MediaPlayerProvider: MediaPlayerProviderProtocol {
     private lazy var audioPlayer = AudioPlayer()
     private var audioPlayerStates: [String: AudioPlayerState] = [:]
+    /// Held onto for the duration of the playback, an unowned player is silent.
+    private var soundEffectPlayer: AVAudioPlayer?
```
pixlwave, line 16: Just a thought, would `AudioServicesPlaySystemSound(1057)` work the same way, as an alternative to creating a dedicated `AVAudioPlayer` for this?

ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
```diff
@@ -493,6 +511,34 @@ class TimelineInteractionHandler {
         }
     }
     
+    private func handleAudioPlayerAction(_ action: AudioPlayerAction) async {
+        switch action {
+        case .didStartPlaying:
+            playingVoiceMessageID = requestedVoiceMessageID
+        case .didFinishPlaying:
+            await autoplayVoiceMessageFollowingTheCurrentOne()
+        default:
```
pixlwave: Should `playingVoiceMessageID` be set back to `nil` when playback finishes?

Also what about `requestedVoiceMessageID`, if nothing is playing should that still contain a value?

pixlwave (approved):
Looks cool! 😎

Although surprised there isn't a setting to enable/disable this functionality – seems like it could be potentially annoying/surprising to some users who may not want/expect autoplay.

Velin92:
> Looks cool! 😎
> 
> Although surprised there isn't a setting to enable/disable this functionality – seems like it could be potentially annoying/surprising to some users who may not want/expect autoplay.

Well the android PR doesn't do that, and WA too doesn't have it, also this is something that we need to get approved by design/product so... yeah things may change

ElementX/Sources/Services/MediaPlayer/MediaPlayerProvider.swift
Velin92: I implemented this with `AudioServicesPlaySystemSound` first... and it worked fine but with 1 huge problem... audio services mute when silent mode is on, and by testing the WA implementation, the autoplay sound effect plays REGARDLESS of silent mode.

ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
Velin92: `playing` does get cleared in the autoplay function, regardless if there is a next track to autoplay or not.
For requested I think instead of having both requested and playing vars, we should have 1 single var to differentiate between the requesting state and the playing state

mxandreas:
> background handling

Can you please describe this limitation more precisely? Does playing stop at once when the app goes to background, or does it not simply start playing the next one? 

> Although surprised there isn't a setting to enable/disable this functionality

When I looked into this, I concluded that people either did not use voice messages or if they did, then wanted also auto-play. I think it is safe to try without setting and add if needed.

> For now I have resolved in using a system sound called "Tink" which plays like a very quick metallic tick

Fine by me.

> If the second message was already played but not finished (for example being at 0:10 duration) and we play the first message, once the first is finished and the second autoplays, it resumes from where it was left off at 0:10

If the second message was played already, then we skip it completely?

Velin92:
> Can you please describe this limitation more precisely? Does playing stop at once when the app goes to background, or does it not simply start playing the next one?

It will simply keep autoplaying also when backgrounding the app

> When I looked into this, I concluded that people either did not use voice messages or if they did, then wanted also auto-play. I think it is safe to try without setting and add if needed.

Yeah agree, makes sense

> If the second message was played already, then we skip it completely?

It works exactly like WA, if the message was already played and finished in the past (which makes the voice message considered stopped, and brought back to 0:00), it will autoplay from the start again when the first message is finished.
If the second message was played in the past and left unfinished, like at 0:10/0:30 for example (paused state), when the first message finishes, it will autoplay from where it was left off (resumed from 0:10)

mxandreas:
> It will simply keep autoplaying also when backgrounding the app

Ok, but this should be a feature, not a bug? When looking into this topic that I noticed that Signal kept playing while backgrounded and it seemed like this was what the users wanted.

Velin92:
> > It will simply keep autoplaying also when backgrounding the app
> 
> Ok, but this should be a feature, not a bug? When looking into this topic that I noticed that Signal kept playing while backgrounded and it seemed like this was what the users wanted.

I mean, we just don't have any background handling so yeah it would just keep playing what was playing in foreground and keep autoplay if the next item was a voice message, if we wanted to stop autoplay or do something else, we would need some bg handling for that, but I don't think it's required, I feel the feature is perfectly fine as it is.

mxandreas:
> if we wanted to stop autoplay or do something else, we would need some bg handling for that, but I don't think it's required, I feel the feature is perfectly fine as it is.

We do want the autoplay to continue, for the "why" see my comment in here: https://github.com/element-hq/element-x-android/pull/6454#issuecomment-4213417299

ElementX/Sources/Services/MediaPlayer/MediaPlayerProvider.swift
pixlwave: Right yes, makes sense 👍

ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
pixlwave: Nice, that looks tidy now 👌

mxandreas:
Created a story for this: https://github.com/element-hq/element-meta/issues/3308

PR #6096, opened 2026-09-02
Feature flag tweaks

stefanceriu:
* promote galleries to labs
* promote knocking to labs
* move the view source flag to the developer options
* merge together the threads labs flag with the room thread list developer option

image


compound-ios/Inspector/Sources/CompoundInspectorApp.swift
pixlwave, line 1: Not sure if this is a default rule being disabled or you're using an old SwiftFormat version, but I think that change was only committed on Monday.

ElementX/Sources/Screens/Settings/AdvancedSettingsScreen/View/AdvancedSettingsScreen.swift
pixlwave, line 1: Not sure I understand why this one is moving tbh, I thought we had intentionally put it outside of the developer options.

pixlwave (approved):
Yay! 🎉

compound-ios/Inspector/Sources/CompoundInspectorApp.swift
pixlwave: Ok well looking at CI I think its the latter 😅

stefanceriu: stupid thing, my upgrade got stuck in a background terminal

PR #6098, opened 2026-09-02
Redaction reason

### LINKED ISSUE #6083: Let the user give a reason when removing a message - #7538
The user cannot provide a reason when deleting a message. Also there is no confirmation dialog for this desctructive action

Add a dialog when the user want to remove a message.
Design: https://www.figma.com/design/oNN0xG59IkD9rb05uhNCot/Aaron-s-playground?node-id=748-51313

None

**Additional context**
Android PR is at https://github.com/element-hq/element-x-android/pull/7538

Velin92:
fixes #6083 

This PR adds a redact confirmation bottom sheet when trying to delete a message from the timeline, so that a reason can be provided
Simulator Screenshot - iPhone 17 - 2026-09-02 at 13 32 27

It also extends the delete file confirmation screen of the media preview, to provide a reason:
Simulator Screenshot - iPhone 17 - 2026-09-02 at 13 27 41

Displaying the reason is out of scope.


Velin92:
@americanrefugee since for when trying to redact/delete a media from the media preview we already had this figma:
https://www.figma.com/design/Ni6Ii8YKtmXCKYNE90cC67/Timeline--new-?node-id=1732-80677&t=OUgqfZLkOpjilCEv-4

I just added the reason text field in it, like this:
Simulator Screenshot - iPhone 17 - 2026-09-02 at 13 27 41

Velin92:
Simulator Screenshot - iPhone 17 - 2026-09-02 at 14 13 35


@americanrefugee we also have this alert for when trying to deletre a poll from its edit screen. Maybe we can keep it simple and just add a text field to the alert? (Like with the remove/ban alert)

ElementX/Resources/Localizations/ar.lproj/SAS.strings
pixlwave, line 1: Did you mean to commit all these SAS.strings files?

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewViewModel.swift
```diff
@@ -267,8 +267,9 @@ class TimelineMediaPreviewViewModel: TimelineMediaPreviewViewModelType {
         }
     }
     
-    private func redactItem(_ item: TimelineMediaPreviewItem.Media) {
-        timelineViewModel.context.send(viewAction: .handleTimelineItemMenuAction(itemID: item.timelineItem.id, action: .redact(isMedia: true)))
+    private func redactItem(_ item: TimelineMediaPreviewItem.Media, reason: String) {
+        // This sheet is the confirmation, so don't ask again in the timeline.
```
pixlwave: Not sure this comment is valuable?

ElementX/Sources/Screens/Timeline/View/RedactConfirmationView.swift
```diff
@@ -0,0 +1,91 @@
...
+struct RedactConfirmationView: View {
+    @Environment(\.dismiss) private var dismiss
+    
+    let confirm: (String) -> Void
+    
+    @State private var reason: String
+    @State private var sheetHeight: CGFloat = .zero
+    
+    init(reason: String = "", confirm: @escaping (String) -> Void) {
+        _reason = State(initialValue: reason)
```
pixlwave: I might be wrong but I thought we could do `self.reason = reason` nowadays? Also couldn't we just remove the init here and remove the `private` from reason too?

ElementX/Sources/Screens/Timeline/View/RedactConfirmationView.swift
```diff
@@ -0,0 +1,91 @@
...
+                .font(.compound.bodyLG)
+                .foregroundStyle(.compound.textSecondary)
+        }
+        .multilineTextAlignment(.center)
+        .padding(.top, 8)
+        .padding(.bottom, 24)
+    }
+    
+    private var buttons: some View {
+        HStack(spacing: 12) {
```
pixlwave: We might want to use a `ViewThatFits` here to switch to a `VStack` when the text gets too large/long.

ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
```diff
@@ -114,13 +115,23 @@ class TimelineInteractionHandler {
         }
     }
     
+    /// Redacting needs the event alone, so it works even when the item isn't part of this timeline,
+    /// such as one held by a media preview that was built from a different one.
```
pixlwave: I think this should be a comment on the guard and not a documentation comment about the whole function? That would match how it was just below.

ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
```diff
@@ -114,13 +115,23 @@ class TimelineInteractionHandler {
...
-        // such as one held by a media preview that was built from a different one.
         if case .redact = action {
-            guard case let .event(_, eventOrTransactionID) = itemID else { fatalError() }
-            Task { await timelineController.redact(eventOrTransactionID) }
+            // The server hasn't accepted this message yet, it is still sending or has failed,
+            // so there is nobody to give a reason to.
+            if case .event(_, .eventID) = itemID {
+                actionsSubject.send(.showRedactConfirmation(itemID: itemID))
+            } else {
+                redact(itemID, reason: nil)
```
pixlwave, line 134: I'm not sure I understand, how come we aren't showing a confirmation for an unsent (`.event(_, .transactionID)`) event?

ElementX/Sources/Screens/Timeline/TimelineModels.swift
```diff
@@ -219,6 +223,11 @@ struct ReadReceiptSummaryInfo: Identifiable {
     let id: TimelineItemIdentifier
 }
 
+/// The item that the user has asked to remove and must confirm, optionally giving a reason.
+struct RedactConfirmationInfo: Identifiable {
```
pixlwave: Should this be scoped to the timeline?

```suggestion
struct TimelineRedactConfirmationInfo: Identifiable {
```

Or nested within the sheet as something like.

`RedactConfirmationView.Item`

UnitTests/Sources/TimelineViewModelTests.swift
```diff
@@ -596,6 +596,99 @@ final class TimelineViewModelTests {
     
     // MARK: - Helpers
     
+    // MARK: - Redaction
```
pixlwave, line 597: That looks wrong? 👀

ElementX/Sources/Other/SwiftUI/Views/RedactionReasonTextField.swift
pixlwave, line 1: Compltely optional: Maybe its just me, this seems a little OTT to extract considering the only things they share are those 2 strings which can be edited together in Localazy.

ElementX/Sources/Screens/Timeline/View/RedactConfirmationView.swift
```diff
@@ -0,0 +1,91 @@
...
+    
+    init(reason: String = "", confirm: @escaping (String) -> Void) {
+        _reason = State(initialValue: reason)
+        self.confirm = confirm
+    }
+    
+    var body: some View {
+        ScrollView {
+            VStack(spacing: 0) {
+                closeButton
```
pixlwave: Doesn't this mean that with e.g. large dynamic type you could scroll the button offscreen? I would have thought this layout should use a `NavigationStack` with a `.toolbar`.

pixlwave (approved):
Nice 😎

ElementX/Sources/Screens/Timeline/View/RedactConfirmationView.swift
Velin92: damn you're right, I wanted to avoid using a nav stack given how simple is the view but that's true...

ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
Velin92: well there is technically no need because there is no reason that will ever be sent for the local only scenario.

Velin92: also this is the same thing that was done in android

Issue #4507, opened 2025-09-12
EXI: When a verification request comes in and we are verified, check whether we are missing secrets and if so, tell user we can't be used for verification.


andybalaam:
Design: https://www.figma.com/design/jvZ7TIyd7V9Vn8sIDB9vMd/Compound-iOS-Templates?node-id=4032-209431&t=YPYbJ2HYdSwixdPK-1

<img width="300" height="167" alt="Image" src="https://github.com/user-attachments/assets/d2c93e23-88d3-4357-a804-2d69d1be636b" />

PR #6099, opened 2026-09-02
Say so when this session has no secrets to verify another one with

hayaksi1:
When an incoming verification request arrives, this session offers "Start verification" whatever state its own cross-signing secrets are in. If the recovery state is `incomplete` the secrets are not on this device, so the flow can only fail — and it fails at the very end, after the user has compared the emojis.

The screen now shows the alert step from the design instead — "Verification with this device is currently unavailable" / "Please use another device or recovery key." — with only the Ignore action. It applies to the initial step of a responder flow only; a verification already under way is untouched, as is the request acknowledgement the view model already sends. `SecureBackupRecoveryState.incomplete` is the signal `SecureBackupScreen` already uses to tell the user their recovery is broken, so no new notion of "missing secrets" is introduced.

The two new strings use the same keys as the Android ones, so they line up on the Localazy import.

`SessionVerificationViewModelTests` covers both sides: a responder flow whose recovery is `incomplete` reaches the unavailable state with the design's title and subtitle and never calls `acceptVerificationRequest`, and the same flow with `enabled` recovery is unaffected. The state is added to `SessionVerification_Previews`, so the snapshot and accessibility tests cover it too.

iOS counterpart of element-hq/element-x-android#7551. Part of #4507, using the copy from the design attached to the issue.

### Pull Request Checklist

- [ ] Voiceover enabled.

### DIFF
--- a/ElementX/Resources/Localizations/en.lproj/Untranslated.strings
+++ b/ElementX/Resources/Localizations/en.lproj/Untranslated.strings
@@ -4,6 +4,8 @@
 "screen_search_no_results_message" = "There are no results for “%1$@.” Try a new search term.";
 "screen_search_tab_chats" = "Chats";
 "screen_search_tab_messages" = "Messages";
+"screen_session_verification_unavailable_subtitle" = "Please use another device or recovery key.";
+"screen_session_verification_unavailable_title" = "Verification with this device is currently unavailable";
 "soft_logout_clear_data_dialog_content" = "Clear all data currently stored on this device?\nSign in again to access your account data and messages.";
 "soft_logout_clear_data_dialog_title" = "Clear data";
 "soft_logout_clear_data_notice" = "Warning: Your personal data (including encryption keys) is still stored on this device.\n\nClear it if you’re finished using this device, or want to sign in to another account.";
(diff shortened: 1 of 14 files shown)


stefanceriu:
> isMissingSecrets: secureBackupController.recoveryState.value == .incomplete

@hayaksi1 can I ask why this hasn't been fixed/tweaked/raised SDK side instead? `RecoveryState.Incomplete` can mean a number of things which might or might not directly link to the verification process, the app is basically oblivious of the underlying crypto parts.

(especially so considering https://github.com/matrix-org/matrix-rust-sdk/blob/7a2a8996c9bd59d1315287b2507621dfc0a06c84/bindings/matrix-sdk-ffi/src/session_verification.rs#L250-L251)

hayaksi1:
You're right — `Incomplete` also covers a missing backup key, so it fires on devices that could verify fine, and someone who never set up recovery reads `Disabled`, so it misses the case this is for.

I opened this for the iOS counterpart of element-hq/element-x-android#7551. If it belongs SDK-side, I'll leave tracking that with you — the SDK already has the check in `SessionVerificationController::process_incoming_verification_request`, it's just skipped for own-device requests. Closing this.

stefanceriu:
@andybalaam I see you've reviewed the Android PR, what are your thoughts on this?

andybalaam:
> @andybalaam I see you've reviewed the Android PR, what are your thoughts on this?

Hi, I agree that we should refuse to verify if we don't have the right secrets locally. I don't have much skin in the game about where this should happen, but if it can be handled in the SDK that sounds like the right answer.

PR #6102, opened 2026-09-03
Add a developer option and selection mode for timeline messages

Velin92:
First step of the multi-select messages feature (element-meta#3255, iOS counterpart of element-x-android#6918).

- Adds a "Select" action to the timeline item menu for remote messages in the room, focused and thread timelines. Choosing it enters selection mode with that message selected.
- While selecting, every bubbled row shows a checkbox on its leading edge and tapping or long pressing a row toggles it. State events, redacted messages, live locations, unsupported events and local echoes cannot be selected and stay inert. Selected rows get the accent background.
- The navigation bar swaps to a Cancel button and an "N selected" title, the back button is hidden and the composer is removed until the selection ends (Cancel or deselecting the last item). This is the hook where the bulk action bar will land in the next PR.
- The selection is capped at 30 messages
- Unit tests for the menu action gating and the selection state transitions, plus RoomScreen and bubble styler previews for the new state.

https://github.com/user-attachments/assets/b323fd8a-8897-4b5d-91e5-5ea014c2674b

### DIFF
--- a/ElementX/Sources/Other/SwiftUI/CollapsedInPlaceModifier.swift
+++ b/ElementX/Sources/Other/SwiftUI/CollapsedInPlaceModifier.swift
@@ -15,7 +15,7 @@ private struct CollapsedInPlaceModifier: ViewModifier {
     func body(content: Content) -> some View {
         content
             .frame(height: isCollapsed ? 0 : nil, alignment: .top)
-            .clipped()
+            .opacity(isCollapsed ? 0 : 1)
             .allowsHitTesting(!isCollapsed)
             .accessibilityHidden(isCollapsed)
     }
--- a/ElementX/Sources/Other/SwiftUI/CollapsedInPlaceModifier.swift
+++ b/ElementX/Sources/Other/SwiftUI/CollapsedInPlaceModifier.swift
@@ -9,7 +9,7 @@ import SwiftUI
 
 /// Collapses the view to zero height while keeping it in the hierarchy, so that its state is
 /// preserved and no `onAppear`/`onDisappear` side effects fire, unlike an `if` statement.
-private struct CollapsedModifier: ViewModifier {
+private struct CollapsedInPlaceModifier: ViewModifier {
     let isCollapsed: Bool
     
     func body(content: Content) -> some View {
@@ -22,7 +22,7 @@ private struct CollapsedModifier: ViewModifier {
 }
 
 extension View {
-    func collapsed(_ isCollapsed: Bool) -> some View {
-        modifier(CollapsedModifier(isCollapsed: isCollapsed))
+    func collapsedInPlace(_ isCollapsed: Bool) -> some View {
+        modifier(CollapsedInPlaceModifier(isCollapsed: isCollapsed))
     }
 }
--- a/ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
+++ b/ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
@@ -226,6 +226,8 @@ class TimelineInteractionHandler {
             actionsSubject.send(.viewInRoomTimeline(eventID: eventID))
         case .downloadMedia:
             break // Handled inline in the media preview screen.
+        case .select:
+            break // Handled by the TimelineViewModel before reaching here.
         case .translate:
             guard let messageTimelineItem = timelineItem as? EventBasedMessageTimelineItemProtocol else { return }
             actionsSubject.send(.showTranslation(text: messageTimelineItem.body))
--- a/ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
+++ b/ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
@@ -224,10 +224,8 @@ class TimelineInteractionHandler {
             analyticsService.trackInteraction(name: .PinnedMessageListViewTimeline)
             guard let eventID = itemID.eventID else { return }
             actionsSubject.send(.viewInRoomTimeline(eventID: eventID))
-        case .downloadMedia:
-            break // Handled inline in the media preview screen.
-        case .selectMessages:
-            break // Handled by the TimelineViewModel before reaching here.
+        case .downloadMedia, .selectMessages:
+            break // Handled by the media preview screen and the TimelineViewModel respectively.
         case .translate:
             guard let messageTimelineItem = timelineItem as? EventBasedMessageTimelineItemProtocol else { return }
             actionsSubject.send(.showTranslation(text: messageTimelineItem.body))
--- a/ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
+++ b/ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
@@ -226,7 +226,7 @@ class TimelineInteractionHandler {
             actionsSubject.send(.viewInRoomTimeline(eventID: eventID))
         case .downloadMedia:
             break // Handled inline in the media preview screen.
-        case .select:
+        case .selectMessages:
             break // Handled by the TimelineViewModel before reaching here.
         case .translate:
             guard let messageTimelineItem = timelineItem as? EventBasedMessageTimelineItemProtocol else { return }
@@ -374,8 +374,7 @@ class TimelineInteractionHandler {
         await voiceMessageRecorder.stopRecording()
     }
     
-    /// Stops an in-progress recording so it isn't left running while the composer is out of
-    /// sight. The recording is kept and the composer moves to the preview state.
+    /// Stops the recording when one is in progress, moving the composer to the preview state.
     func stopRecordingVoiceMessageIfNeeded() async {
         guard voiceMessageRecorder.isRecording else { return }
         await voiceMessageRecorder.stopRecording()
--- a/ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
+++ b/ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
@@ -374,6 +374,13 @@ class TimelineInteractionHandler {
         await voiceMessageRecorder.stopRecording()
     }
     
+    /// Stops an in-progress recording so it isn't left running while the composer is out of
+    /// sight. The recording is kept and the composer moves to the preview state.
+    func stopRecordingVoiceMessageIfNeeded() async {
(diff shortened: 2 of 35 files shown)


ElementX/Sources/Screens/Timeline/View/ItemMenu/TimelineItemMenuAction.swift
```diff
@@ -68,6 +68,7 @@ enum TimelineItemMenuAction: Identifiable, Hashable {
     case reply(isThread: Bool)
     case replyInThread
     case forward(itemID: TimelineItemIdentifier)
+    case select
```
pixlwave: Maybe this should be called `selectMessages`/`selectItem` and then in the future #5451 could add `selectText`?

ElementX/Sources/Screens/Timeline/View/ItemMenu/TimelineItemMenuActionProvider.swift
```diff
@@ -147,9 +152,13 @@ struct TimelineItemMenuActionProvider {
...
     }
     
-    private func makeEncryptedItemActions() -> TimelineItemMenuActions? {
+    private func makeEncryptedItemActions(for item: EventBasedTimelineItemProtocol) -> TimelineItemMenuActions? {
         var actions: [TimelineItemMenuAction] = [.copyPermalink]
         
+        if isMultiSelectEnabled, item.isBulkSelectable {
+            actions.append(.select)
+        }
+
```
pixlwave: Is this actually necessary given it's presumably going to return `false` when the item isn't decrypted?

ElementX/Sources/Screens/Timeline/View/Style/LongPressWithFeedback.swift
pixlwave, line 1: 👌

ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
```diff
@@ -360,6 +362,13 @@ class TimelineInteractionHandler {
         await voiceMessageRecorder.stopRecording()
     }
     
+    /// Stops an in-progress recording so it isn't left running while the composer is out of
+    /// sight. The recording is kept and the composer moves to the preview state.
```
pixlwave: FWIW, this method has nothing to do with the composer being out of sight…

ElementX/Sources/Screens/Timeline/TimelineModels.swift
```diff
@@ -193,6 +201,33 @@ struct TimelineViewStateBindings {
...
+    
+    var count: Int {
+        selectedEventIDs.count
+    }
+    
+    var isAtCap: Bool {
+        count >= Self.maxCount
+    }
+    
+    func isSelected(_ eventID: String?) -> Bool {
```
pixlwave: From an API perspective, calling this `includes` or `contains` would be less repetitive.
```
selection.isSelected(myID)
```

vs

```
selection.includes(myID)
```

ElementX/Sources/Screens/Timeline/TimelineModels.swift
```diff
@@ -193,6 +201,33 @@ struct TimelineViewStateBindings {
     var textToBeTranslated: String?
 }
 
+/// The state of the multi-selection of messages, active as soon as an item is selected.
+struct TimelineSelectionState: Equatable {
+    static let maxCount = 30
+    
+    /// Mirrors the `messageMultiSelectEnabled` labs flag.
```
pixlwave: Given this isn't in Labs?

```suggestion
    /// Mirrors the `messageMultiSelectEnabled` feature flag.
```

ElementX/Sources/Screens/Timeline/TimelineModels.swift
```diff
@@ -310,6 +345,15 @@ enum ScrollDirection: Equatable {
 }
 
 extension TimelineViewState {
+    /// Multi-selection is only offered in the room and thread timelines.
+    var canSelectMessages: Bool {
+        guard selection.isEnabled else { return false }
+        switch timelineKind {
+        case .live, .detached, .thread: return true
+        case .pinned, .media: return false
```
pixlwave: Super minor: Whilst I like single line cases, I think they're awkward when there's a `return` in them. So maybe multi-line or switch as an expression:
```suggestion
        return switch timelineKind {
        case .live, .detached, .thread: true
        case .pinned, .media: false
```

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
```diff
@@ -1159,6 +1179,70 @@ class TimelineViewModel: TimelineViewModelType, TimelineViewModelProtocol {
...
+extension TimelineViewModel {
+    private func enterSelection(itemID: TimelineItemIdentifier) {
+        guard state.canSelectMessages, let eventID = selectableEventID(for: itemID) else { return }
+        
+        // The composer is collapsed while selecting, so don't leave the microphone open behind it.
+        Task { await timelineInteractionHandler.stopRecordingVoiceMessageIfNeeded() }
+        actionsSubject.send(.composer(action: .removeFocus))
+        
+        guard !state.selection.isAtCap || state.selection.selectedEventIDs.contains(eventID) else {
+            showSelectionCapToast()
```
pixlwave: Completely optional but maybe `Limit` would be clearer than `Cap` for all of this?

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
```diff
@@ -196,8 +198,16 @@ class TimelineViewModel: TimelineViewModelType, TimelineViewModelProtocol {
             }
         case .displayTimelineItemMenu(let itemID):
             timelineInteractionHandler.displayTimelineItemActionMenu(for: itemID)
+        case .handleTimelineItemMenuAction(let itemID, .select):
+            enterSelection(itemID: itemID)
         case .handleTimelineItemMenuAction(let itemID, let action):
             timelineInteractionHandler.handleTimelineItemMenuAction(action, itemID: itemID)
+        case .enterSelection(let itemID):
+            enterSelection(itemID: itemID)
```
pixlwave: I'm slightly confused by the name `enterSelection` here:

- Is it enabling selection with the given item, in which case `activateMessageSelection`, `activateItemSelection`, `enableSelectionMode` or something along those lines would be clearer.
- Or is it selecting the given item, in which case simply `selectItem` would be clearer (with the byproduct of enabling selection if not already enabled).

ElementX/Sources/Screens/Timeline/View/Style/TimelineItemAccessibilityModifier.swift
```diff
@@ -8,11 +8,39 @@
...
+            if let item = timelineItem as? EventBasedTimelineItemProtocol, item.isBulkSelectable {
+                content
+                    .accessibilityElement(children: .combine)
+                    .accessibilityAddTraits(isSelected ? [.isButton, .isSelected] : .isButton)
+                    .accessibilityAction(.default, action)
+            } else {
+                // Items that can't be selected are inert while selecting.
+                content
+                    .accessibilityElement(children: .combine)
+            }
```
pixlwave: Errr, these bypass things like `.accessibilityRepresentation` and `.accessibilityLabel` that we've added below for clarity to VoiceOver users? That doesn't seem right?

ElementX/Sources/Other/SwiftUI/CollapsedInPlaceModifier.swift
pixlwave, line 1: a) This is a cool trick
b) I'm not sure `collapse` fully indicates that it doesn't trigger onAppear/Disappear. Maybe `.collapsedInPlace`, `.collapsedInHierarchy` or something like `.zeroHeight` would be helpful. But even then I don't have huge confidence in the clarity of those names

ElementX/Sources/Screens/Timeline/View/Style/TimelineItemBubbledStylerView.swift
```diff
@@ -42,6 +42,19 @@ struct TimelineItemBubbledStylerView<Content: View>: View {
...
+    }
+    
+    private var isSelected: Bool {
+        context.viewState.selection.isSelected(timelineItem.id.eventID)
+    }
+    
+    private func toggleSelection() {
+        guard timelineItem.isBulkSelectable else { return }
+        context.send(viewAction: .toggleSelection(itemID: timelineItem.id))
+    }
```
pixlwave: Shouldn't functions come beneath the `body`/subviews?

ElementX/Sources/Screens/Timeline/View/Style/TimelineItemBubbledStylerView.swift
```diff
@@ -53,6 +66,46 @@ struct TimelineItemBubbledStylerView<Content: View>: View {
...
     var body: some View {
+        HStack(spacing: 0) {
+            if isSelectionActive {
+                selectionIndicator
+            }
+            
+            bubble
+        }
+        .padding(EdgeInsets(top: 1, leading: 8, bottom: 1, trailing: 8))
+        .background(isSelected ? Color.compound.bgAccentSelected : .clear)
```
pixlwave, line 72: The green multi-selection colour looks weird to me on iOS (I feel like that's possibly an Android/Material thing).

iOS usually highlights multi-selected items in grey (most lists/forms) or doesn't highlight them at all. My guess is that it uses the tap-selection colour if there is one, otherwise it doesn't apply a highlight.

pixlwave (approved):
☑️☑️☑️

ElementX/Sources/Screens/Timeline/View/ItemMenu/TimelineItemMenuActionProvider.swift
Velin92: uhm good call maybe is useless

ElementX/Sources/Screens/Timeline/View/Style/TimelineItemBubbledStylerView.swift
Velin92: I'll leave this to @americanrefugee

ElementX/Sources/Screens/Timeline/TimelineInteractionHandler.swift
Velin92: You're right, this comment belongs where this function is called

PR #6103, opened 2026-09-03
Add an `AccountProvider` type and use it within the authentication flow.

pixlwave:
This PR makes 3 changes:
- Adds a dedicated `AccountProvider` type instead of using a plain `String` for app configuration.
    - Used in both `AppSettings` (first commit) and `LoginHomeserver` (second commit).
- Reinstates the checks for allowed servers when using QR code login as long as the allowed account provider(s) are `.managed` (removed in [#5196](https://github.com/element-hq/element-x-ios/pull/5196)).
- Renames any remaining uses of `homeserverAddress` to `serverNameOrBaseURL`.

I'll be honest: I'm unsure whether the second commit is a good idea or not as it'll always end up with a `.generic` provider when `AuthenticationService.configure` is called, so it seems potentially confusing. Open to input on whether to just drop it.


ElementX/Sources/Screens/Authentication/LoginScreen/LoginHomeserver.swift
```diff
@@ -10,34 +10,27 @@ import Foundation
...
     /// - Lowercase the address.
-    /// - Ensure the address contains a scheme, otherwise make it `https`.
+    /// - Removes an https scheme (treating it as implicit).
     /// - Remove any trailing slashes.
-    static func sanitized(_ address: String) -> String {
-        var address = address.trimmingCharacters(in: .whitespacesAndNewlines).lowercased()
-        
-        if !address.contains("://") {
-            address = "https://\(address)"
-        }
```
stefanceriu, line 36: Logic here seems to have changed quite a bit, are you sure it's all still correct?

ElementX/Sources/Services/Authentication/AuthenticationService.swift
```diff
@@ -64,15 +64,15 @@ class AuthenticationService: AuthenticationServiceProtocol {
...
+        homeserverSubject = .init(LoginHomeserver(accountProvider: appSettings.defaultAccountProvider, loginMode: .unknown))
         flow = .login
     }
     
     // MARK: - Public
     
     func configure(for serverNameOrBaseURL: String, flow: AuthenticationFlow) async -> Result<Void, AuthenticationServiceError> {
         do {
-            var homeserver = LoginHomeserver(serverNameOrBaseURL: serverNameOrBaseURL, loginMode: .unknown)
+            var homeserver = LoginHomeserver(accountProvider: .generic(serverNameOrBaseURL), loginMode: .unknown)
```
stefanceriu, line 78: I think that's fine but to be honest, totally up to you!

stefanceriu (approved):
Looks good to me, just a small question. 👍

ElementX/Sources/Screens/Authentication/LoginScreen/LoginHomeserver.swift
pixlwave: Yeah, this all came from Classic:

https://github.com/element-hq/element-ios/blob/2552faab76b23e505d80f10ca09848b24069b53d/RiotSwiftUI/Modules/Authentication/Common/AuthenticationModels.swift#L82

Element X/Rust basically considers the `https://` to be implicit so we're fine. The rest of it was just a tidy-up to make it functional. I checked a bunch of cases and they all seemed good to me.

pixlwave: And if you notice above, we were removing the `https://` immediately after adding it:

`Self.sanitized(address).components(separatedBy: "://").last`

PR #6107, opened 2026-09-04
Fix the tab rail flickering on narrow iPad windows

### LINKED ISSUE #6085: App flickers constantly on ipados 27
### Steps to reproduce

Not sure what triggers it. But maybe..

1. Install ipados 27 beta
2. Open element x
3. App flickers constantly 

#### What did you expect?
The app works normally . 

The app is unusable

### Your phone model

Ipad mini 7th gen

### Operating system version

Ipad os 27 beta

### Application version

26.08.2 (240)

### Homeserver

matrix.org

### Will you send logs?

Yes

stefanceriu:
The rail was hidden when the split view collapsed, but that freed up enough width for the split to expand again, showing the rail and collapsing the split. The split's state now only hides the rail and showing it again is driven by the container's width growing.

Fixes https://github.com/element-hq/element-x-ios/issues/6085

### DIFF
--- a/ElementX/Sources/Application/Navigation/NavigationTabCoordinator.swift
+++ b/ElementX/Sources/Application/Navigation/NavigationTabCoordinator.swift
@@ -307,6 +307,48 @@ extension EnvironmentValues {
     @Entry var tabViewHorizontalSizeClass: UserInterfaceSizeClass?
 }
 
+/// Tracks whether the tab rail should be shown alongside the split view's sidebar.
+///
+/// The rail is laid out beside the split view, so hiding it hands that width back to the split, which may
+/// then have room for its sidebar again, which would show the rail, ad infinitum. The split reports the
+/// same column visibility whether it collapsed on its own or because the sidebar was toggled, and at the
+/// same width, so the two are told apart by when they arrive: the rebound lands in the next layout pass.
+private struct TabRailVisibility {
+    /// The longest a collapse can follow the rail being shown and still be considered caused by it.
+    private static let reboundDuration: TimeInterval = 0.5
+    
+    private(set) var isVisible = true
+    
+    private var containerWidth: CGFloat = 0
+    /// The container width at which the rail and the sidebar have been shown not to fit together.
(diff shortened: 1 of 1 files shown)


pixlwave:
I can't reply on matrix as you've got an unverified device, but at first glance this fails to re-show the rail after having manually hid the sidebar:

https://github.com/user-attachments/assets/9fa2b09a-60d8-4e67-a254-20e435004bd9

stefanceriu:
Great, I somehow didn't see any of that 🤦‍♂️ 

Let's try this again..

ElementX/Sources/Application/Navigation/NavigationTabCoordinator.swift
```diff
@@ -317,6 +317,11 @@ private struct NavigationTabCoordinatorView<Tag: Hashable>: View {
     @State private var isFullScreen = true
     @State private var isRailVisible = true
     @State private var railBackgroundColor: Color = .compound.bgCanvasDefault
+    @State private var containerWidth: CGFloat = 0
+    /// The container width at which the rail and the split view's sidebar have been shown not to fit together.
+    @State private var railBlockedWidth: CGFloat?
+    /// When the rail was last shown, used to spot the split view collapsing as a direct result of it.
+    @State private var railShownDate = Date.distantPast
```
pixlwave: part of me wonders if we should combine all 3 of these (possibly with `isRailVisible` too) into a single struct about this "management". We could maybe even then put the calculations on the struct to keep the navigation coordinator slimmer.

pixlwave (approved):
I haven't tried this one out but the code looks much more convincing to me this time around.

ElementX/Sources/Application/Navigation/NavigationTabCoordinator.swift
stefanceriu: <del>we could but let's see if it even works first?</del>

stefanceriu: Nice idea, yeah! 💪

PR #6112, opened 2026-09-07
Minor quicklook fixes.

pixlwave:
This first 2 commits from "Bug fixes" https://github.com/element-hq/element-x-ios/pull/6072#pullrequestreview-5111398023 plus a fix for showing/hiding the caption on iOS 26

There's more to review for that 3rd commit – not showing the caption after hiding it does potentially make sense given it currently sits on top of the progress view, but we might not need to use the dedicated tap gesture now.


stefanceriu (approved):
Good stuff, let's go! 💪

PR #6133, opened 2026-09-08
Add a couple of points about code style/ordering to the AGENTS.md

pixlwave:
Might help with the random placement of additions.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -15,6 +15,8 @@ PRs must follow rules. Prefer Xcode MCP tools over terminal commands.
 - Follow [Swift API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/) everywhere, Rust SDK wrappers too (`ID` not `Id`, `URL` not `Url`, `configuration` not `config` or `cfg`).
 - File headers live in `IDETemplateMacros.plist`.
 - **Import `SwiftUI`, not `UIKit`** — need `UIPasteboard` etc, `SwiftUI` bring it. `SwiftUI` re-export `Foundation` too, so no `import Foundation` beside it.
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -15,6 +15,8 @@ PRs must follow rules. Prefer Xcode MCP tools over terminal commands.
 - Follow [Swift API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/) everywhere, Rust SDK wrappers too (`ID` not `Id`, `URL` not `Url`, `configuration` not `config` or `cfg`).
 - File headers live in `IDETemplateMacros.plist`.
 - **Import `SwiftUI`, not `UIKit`** — need `UIPasteboard` etc, `SwiftUI` bring it. `SwiftUI` re-export `Foundation` too, so no `import Foundation` beside it.
+- **Clean Code style** (*Clean Code*, Robert C. Martin) — small functions, intention-revealing names, low noise. Newspaper order: headline first, detail down. Stepdown rule: caller above callee, calls point down. Suite: test cases first, helpers below. Swift idiom always win on clash.
```
pixlwave: I'm not sure if `small functions` will end up being counter-productive, maybe should remove/reword it.

Velin92: small functions is a bit scary, considering AI already writes some surprisingly stuipid small functions already. Maybe we should give it a more precise lines range, or avoid mentioning small at all.

stefanceriu: > *Clean Code*, Robert C. Martin

I wonder if this is going backfile in some spectacular way. I think it might better to not reference the whole thing but just literally tell it what you want.

pixlwave: Removed it.

pixlwave: We discussed and will see how it goes (I've had it in a side project for a while and seems fine). Easy enough to remove if it's bad.

PR #6135, opened 2026-09-08
Introduce a local mechanism for managing the app's badge count

stefanceriu:
Which reads the correct count from the SDK and then increments it by one on every received push notification, only to be overriden again later with the correct value from the main process.

Set unde the same feature flag as the room list notification counts.

### DIFF
--- a/ElementX/Sources/FlowCoordinators/ChatsTabFlowCoordinator.swift
+++ b/ElementX/Sources/FlowCoordinators/ChatsTabFlowCoordinator.swift
@@ -341,6 +341,7 @@ class ChatsTabFlowCoordinator: FlowCoordinatorProtocol {
                     Task {
                         let roomSummaries = self.userSession.clientProxy.staticRoomSummaryProvider.roomListPublisher.value
                         await self.flowParameters.notificationManager.removeDeliveredNotificationsForFullyReadRooms(roomSummaries)
+                        await self.flowParameters.notificationManager.updateAppBadgeCount()
                     }
                 default:
                     break
--- a/ElementX/Sources/Services/Client/ClientProxyProtocol.swift
+++ b/ElementX/Sources/Services/Client/ClientProxyProtocol.swift
@@ -137,6 +137,9 @@ protocol ClientProxyProtocol: AnyObject {
     
     var pusherNotificationClientIdentifier: String? { get }
     
+    /// The total number of unread notifications across all joined, non-muted rooms, as computed by the SDK.
+    var totalUnreadNotifications: UInt64 { get }
+    
     var mediaLoader: MediaLoaderProtocol { get }
     
     var contentScanner: ContentScannerProxyProtocol? { get }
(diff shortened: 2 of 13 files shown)


ElementX/Sources/FlowCoordinators/ChatsTabFlowCoordinator.swift
```diff
@@ -341,6 +341,7 @@ class ChatsTabFlowCoordinator: FlowCoordinatorProtocol {
                     Task {
                         let roomSummaries = self.userSession.clientProxy.staticRoomSummaryProvider.roomListPublisher.value
                         await self.flowParameters.notificationManager.removeDeliveredNotificationsForFullyReadRooms(roomSummaries)
+                        await self.flowParameters.notificationManager.updateAppBadgeCount()
```
pixlwave, line 344: Maybe just me, but these seem like they should be the same method from an API perspective?

ElementX/Sources/Services/Client/ClientProxyProtocol.swift
```diff
@@ -137,6 +137,9 @@ protocol ClientProxyProtocol: AnyObject {
     
     var pusherNotificationClientIdentifier: String? { get }
     
+    /// The total number of unread notifications across all joined, non-muted rooms, as computed by the SDK.
```
pixlwave, line 140: Presumably this isn't specifically excluding muted rooms, it's just that there aren't any notifications for them to count?

```suggestion
    /// The total number of unread notifications across all joined room, as computed by the SDK.
```

NSE/Sources/NotificationHandler.swift
```diff
@@ -79,15 +84,37 @@ nonisolated class NotificationHandler {
...
+            notificationContent.badge = NSNumber(value: incrementBadgeCount())
+            MXLog.info("\(tag) Delivering notification, new badge value: \(settings.lastKnownBadgeCount)")
+        } else {
+            MXLog.info("\(tag) Delivering notification")
+        }
         contentHandler(notificationContent)
     }
     
+    private func incrementBadgeCount() -> Int {
+        // `handleTimeExpiration` can deliver a notification this handler already delivered.
```
pixlwave, line 97: Errr that doesn't sound good?!

pixlwave (approved):
Cool. Although for this to work we also need to opt out of the silent pushes that decrement the badge when you read a room too?

ElementX/Sources/Services/Client/ClientProxyProtocol.swift
stefanceriu: Yeah, that's a bit extra

ElementX/Sources/FlowCoordinators/ChatsTabFlowCoordinator.swift
stefanceriu: Hmm.. good point, I'm not sure actually as we have nothing to replace them with from the NSE's point of view 🤔

PR #6143, opened 2026-09-09
Experimental Native Call gated behind developper flag

BillCarsonFr:
Allows to test the new [element-call-ios](https://github.com/element-hq/element-call-ios) component instead of the webview based Element-Call.

Accessible only behind a lab flag:
image

Made available as early access to get some feedbacks.
Integrates better with the os than the webview, and offers:
- Proper Audio Call via callkit (in background phone locked)
- Screen Sharing support
- Audio routing

### Pull Request Checklist

- [ ] Voiceover enabled.


ElementX.xcodeproj/project.pbxproj
```diff
@@ -10112,6 +10153,7 @@
 				OTHER_LDFLAGS = (
 					"$(inherited)",
 					"-ObjC",
+					"-ObjC",
 				);
```
pixlwave, line 10167: I don't think we need 2 of these…

ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -465,3 +467,8 @@ final nonisolated class AppSettings: @unchecked Sendable {
 }
 
 nonisolated extension AppSettings: CommonSettingsProtocol { }
+
+/// Mirrors `MatrixRtcElementCallCompat` without the app's shared code depending on the RTC framework.
+enum NativeCallElementCallCompat: String, Codable, CaseIterable {
+    case off, stickyEvents, stateEvents
+}
```
pixlwave: This should go in its own file in this `Settings` directory like the other ones please.

<del>Also what does `NativeCallElementCallCompat` mean here?</del>

Also lets follow the Swift API design guidelines and call this `NativeCallElementCallCompatibility`.
Or to be honest, `nativeCallCompatibility: NativeCallCompatibility`, it's quite clear that it uses Element Call.

ElementX/Sources/Screens/NativeCallScreen/NativeCallScreenCoordinator.swift
```diff
@@ -0,0 +1,35 @@
+//
+// Copyright 2026 Element Creations Ltd.
+// Copyright 2026 New Vector Ltd.
```
pixlwave: We only need `Element Creations Ltd` on new files

```suggestion
```

ElementX/Sources/Services/NativeCall/MatrixRtcLogBridge.swift
```diff
@@ -0,0 +1,27 @@
...
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+import ElementCallKit
+import Foundation
+
+/// Feeds the Rust RTC core's log records into `MXLog` so they end up in rageshakes next to the SDK's.
+nonisolated enum MatrixRtcLogBridge {
```
pixlwave: Again, applying the Swift API design guidelines:

```suggestion
nonisolated enum MatrixRTCLogBridge {
```

ElementX/Sources/Services/NativeCall/MatrixRtcLogBridge.swift
```diff
@@ -0,0 +1,27 @@
...
+    /// Idempotent, must run before any other use of the call package.
+    static func install() {
+        MatrixRtcLogging.install { record in
+            let message = "[\(record.target)] \(record.message)"
```
pixlwave: We don't really do logs in this style of `[Target] message` as the SDK has built in formatting. Would something like `MXLog.error(record.message, file: record.target)` work here? Presumably the record could be extended to include file/line numbers too.

ElementX/SupportingFiles/target.yml
```diff
@@ -147,6 +147,14 @@ targets:
...
         SWIFT_OBJC_INTERFACE_HEADER_NAME: GeneratedInterface-Swift.h
         PILLS_UT_TYPE_IDENTIFIER: $(BASE_BUNDLE_IDENTIFIER).pills
+        # Required by the call package's media library. libwebrtc's Objective-C categories are
+        # dead-stripped out of its static archive without this, and the app then aborts at runtime
+        # with an unrecognised selector rather than failing to build. See the package's README.
+        #
+        # Measure the binary size delta when changing this: `-ObjC` loads every object file in a
+        # static archive that contains Objective-C, and `-force_load` on the archive alone is the
+        # documented fallback if the delta is worse than expected.
+        OTHER_LDFLAGS: [$(inherited), -ObjC]
```
pixlwave: I'm guessing this is what can be removed given inherited likely already contains that `-ObjC`.

ElementX/SupportingFiles/target.yml
```diff
@@ -260,6 +268,16 @@ targets:
...
+    # The native call stack. ElementCallMatrix is the only one that touches the Matrix SDK; the
+    # other three know nothing about it.
+    - package: ElementCall
+      product: ElementCallKit
+    - package: ElementCall
+      product: ElementCall
+    - package: ElementCall
+      product: ElementCallUI
+    - package: ElementCall
+      product: ElementCallMatrix
```
pixlwave: How important is it that we have 4 packages here? Can one of them be considered the main/top-level one and re-export the rest so that there's only one `import ElementCall{XYZ}` necessary throughout the code?

Tools/Sourcery/AutoMockable.stencil
```diff
@@ -7,6 +7,11 @@
 import AnalyticsEvents
 import AVFoundation
 import CallKit
+// The call package appears in mocked protocol signatures, notably the transport factory on
+// ClientProxyProtocol.
```
pixlwave: I don't think we need this comment added to all of our mocks 😆

ElementX/Sources/Application/Navigation/NavigationTabCoordinator.swift
```diff
@@ -377,12 +384,21 @@ private struct NavigationTabCoordinatorView<Tag: Hashable>: View {
...
                 .animation(.elementDefault, value: navigationTabCoordinator.overlayModule)
             }
+            .overlay(alignment: .top) {
+                if navigationTabCoordinator.overlayModule != nil,
+                   navigationTabCoordinator.overlayPresentationMode == .minimized,
+                   let accessory = navigationTabCoordinator.minimizedOverlayAccessory {
+                    accessory
+                        .transition(.move(edge: .top).combined(with: .opacity))
+                }
+            }
```
pixlwave: I have a feeling this is going to be
a) obscured (partially?) by sheets on an iPhone which will look broken and interactivity will be either limited or disabled until the sheet is dismissed.
b) Similar on iPad except it won't be obscured by the sheet, so depending on the sheet, tapping e.g. Hang Up may dismiss the sheet and not hang-up the call.

Using a Picture In Picture controller fixes this, but if we have to have a bar, then it probably needs to live on the window rather than the tab coordinator.

It would probably be sensible to scope this one out of the integration PR and do it in a follow-up feature PR if possible?

ElementX/Sources/Services/NativeCall/ElementCallOptionsAdapter.swift
```diff
@@ -0,0 +1,51 @@
...
+///
+/// This covers only the package's Swift logging. The Rust core's output is installed separately by
+/// `MatrixRtcLogBridge`.
+struct ElementCallLoggerAdapter: ElementCallLogging {
+    func log(_ level: ElementCallLogLevel, _ message: String) {
+        switch level {
+        case .debug: MXLog.verbose("ElementCall: \(message)")
+        case .info: MXLog.info("ElementCall: \(message)")
+        case .warning: MXLog.warning("ElementCall: \(message)")
+        case .error: MXLog.error("ElementCall: \(message)")
```
pixlwave: Same thought here on the logs, maybe we can use Rust's formatter and pass file/line numbers etc?

ElementX/Sources/Services/NativeCall/ElementCallRoomContextAdapter.swift
```diff
@@ -0,0 +1,71 @@
...
+/// Serves the call package's room port from a room proxy.
+///
+/// Deliberately not in the package, even though the underlying data is all in the SDK. The proxy is
+/// an asset rather than a wrapper: it keeps a cached current value for room info and members,
+/// refreshed off the SDK's own subscription, and reads members from disk before the network. A
+/// version built on the raw SDK would re-derive all of that and lose the caching.
+///
+/// The display name is also a product decision rather than protocol, which is the other reason it
+/// belongs here.
+@MainActor
```
pixlwave: We're `@MainActor` by default in the Element X target, this shouldn't be necessary unless it is used by the NSE.

pixlwave (commented):
I've taken an initial look through at the code and put an initial set of comments in here.

One big request though: I see a mixture of `NativeCallXYZ`, `ElementCallXYZ` and `NativeCallElementCallXYZ` types/names. It would be nice if we could go with a blanker `NativeCallXYZ` in the app itself to make it clear to see/find everything specifically related to the new calling feature.

ElementX.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -171,6 +180,15 @@
         "version" : "26.9.8"
       }
     },
+    {
+      "identity" : "matrix-rust-rtc",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/BillCarsonFr/matrix-rust-rtc",
```
pixlwave, line 186: Out of interest, will this one live on `element-hq`/`matrix-org` before the PR gets merged?

ElementX/SupportingFiles/target.yml
stefanceriu: Yes please!

Tools/Sourcery/AutoMockable.stencil
stefanceriu: As it happens it's just one file as everything lives in GeneratedMocks.swift but still, we don't need 3 different packages, do we?

ElementX/Sources/Application/Navigation/NavigationTabCoordinator.swift
stefanceriu: I fully agree here, if we do need more than what the default CallKit controls provide then we should add it separately.

ElementX/Sources/Application/Settings/AppSettings.swift
```diff
@@ -455,6 +455,18 @@ final nonisolated class AppSettings: @unchecked Sendable {
...
+    @UserPreference(defaultValue: false)
+    var nativeCallEnabled: Bool
+    
+    /// Shows minimized native video calls in the system Picture in Picture window.
+    @UserPreference(defaultValue: true)
+    var nativeCallPictureInPictureEnabled: Bool
+    
+    /// How native calls publish their membership; state events is what Element Web speaks today.
+    @UserPreference(defaultValue: NativeCallElementCallCompat.stateEvents)
+    var nativeCallElementCallCompat: NativeCallElementCallCompat
```
stefanceriu: Why do we need so many feature flags though, can't these go inside the ElementCall UI itself? 🤔

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -318,15 +320,34 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
...
+                    }
                 default:
                     break
                 }
             }
             .store(in: &cancellables)
         
+        // Created with the session rather than with the first call: media keys arrive over to-device
+        // and cannot be caught up on, so a stack started when the user taps the call button has
+        // already missed keys sent while the room was ringing.
```
stefanceriu: This feels super wasteful, is there no better way?

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -522,11 +561,105 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
...
+            }
+            .store(in: &cancellables)
+        
+        Task { await stack.start() }
+    }
+    
+    private func presentElementCallScreen(configuration: ElementCallConfiguration) {
+        if elementCallStack == nil {
+            startElementCallStack()
+        }
```
stefanceriu: If we start the stack with the session then why do we need to check it here? 🤔

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -370,6 +371,16 @@ class ClientProxy: ClientProxyProtocol {
...
     
+    // MARK: Native calls
+    
+    func makeElementCallTransport() -> ElementCallSDKTransport? {
+        guard let client = client as? Client else {
+            MXLog.warning("Cannot make an Element Call transport without a real SDK client")
+            return nil
+        }
+        return ElementCallSDKTransport(client: client, logger: ElementCallLoggerAdapter())
+    }
```
stefanceriu: This doesn't belong on the client proxy.

ElementX/Sources/Services/Client/ClientProxyProtocol.swift
```diff
@@ -168,6 +169,19 @@ protocol ClientProxyProtocol: AnyObject {
...
+    /// A ready-made Matrix transport for the call package, or nil when this proxy is not backed by a
+    /// real SDK client, which is the case in mocks and previews.
+    ///
+    /// The package supplies the implementation; all this does is hand it the client. The downcast to
+    /// the concrete SDK type stays in here rather than leaking through the protocol, which is the
+    /// same shape the content scanner already uses.
+    ///
+    /// Concrete rather than existential, and taking no logger, because Sourcery cannot generate a
+    /// mock for `(any Protocol)?` and the logger has nothing to configure.
+    func makeElementCallTransport() -> ElementCallSDKTransport?
```
stefanceriu: Same here, wrong place, not to mention it hardcodes a ElementCallLoggerAdapter

ElementX/Sources/Services/ElementCall/ElementCallServiceProtocol.swift
```diff
@@ -28,4 +31,14 @@ protocol ElementCallServiceProtocol {
...
+    
+    // MARK: Native calls
+    
+    /// Whether an answered incoming call is handed to the native stack (kept alive in CallKit)
+    /// rather than to the web view (ended right after answering).
+    func setNativeCallModeEnabled(_ enabled: Bool)
+    /// Reports an outgoing native call to CallKit, or attaches to the ringing call it answers.
+    func startNativeCallSession(roomID: String, roomDisplayName: String, isVideo: Bool) async
+    func reportNativeCallConnected(roomID: String)
+    func endNativeCallSession(roomID: String)
```
stefanceriu: Why do we need parallel methods for native vs web? We should use the protocol and hide the implementations.

ElementX/Sources/Services/NativeCall/ElementCallOptionsAdapter.swift
```diff
@@ -0,0 +1,50 @@
...
+//
+
+import ElementCall
+import ElementCallKit
+
+/// Serves the call package's settings port from `AppSettings`.
+///
+/// Every member reads the setting on access rather than capturing it, so a developer flipping a
+/// toggle mid-session is seen by the next call without restarting anything.
+struct ElementCallOptionsAdapter: ElementCallOptions {
```
stefanceriu: This doesn't adapt anything, just implement a protocol. `ElementCallOptions` should be called `ElementCallOptionsProtocol` and `ElementCallOptionsAdapter` just `ElementCallOptions`

ElementX/Sources/Services/NativeCall/ElementCallRoomContextAdapter.swift
```diff
@@ -0,0 +1,71 @@
...
+final class ElementCallRoomContextAdapter: ElementCallRoomContext {
```
stefanceriu: Same exact thing here, no adapting, just implementing.

ElementX/Sources/Services/NativeCall/NativeCallRoomContext.swift
```diff
@@ -0,0 +1,71 @@
...
+///
+/// The display name is also a product decision rather than protocol, which is the other reason it
+/// belongs here.
+@MainActor
+final class ElementCallRoomContextAdapter: ElementCallRoomContext {
+    private let roomProxy: JoinedRoomProxyProtocol
+    
+    init(roomProxy: JoinedRoomProxyProtocol) {
+        self.roomProxy = roomProxy
+        Task { await roomProxy.updateMembers() }
```
stefanceriu: Why force update the members?

ElementX/Sources/Services/NativeCall/NativeCallRoomContextAdapter.swift
```diff
@@ -0,0 +1,71 @@
...
+    
+    var memberProfilesPublisher: AnyPublisher<[String: ElementCallMemberProfile], Never> {
+        roomProxy.membersPublisher.map(Self.profiles).eraseToAnyPublisher()
+    }
+    
+    /// The app's fallback chain, which is why this is not in the package: what to call a room with
+    /// no name is Element's opinion, not something the protocol settles.
+    private static func name(from info: RoomInfoProxyProtocol, roomID: String) -> String {
+        info.displayName ?? info.rawName ?? info.canonicalAlias ?? roomID
+    }
```
stefanceriu: This should probably go on the RoomInfoProxyProtocol extension if it doesn't already exist somewhere.

ElementX/Sources/Services/NativeCall/ElementCallStyleAdapter.swift
```diff
@@ -0,0 +1,154 @@
...
+// Colours, icons and avatars for the call, read from this app's Compound rather than the package's
+// default tokens.
+//
+// This is the whole reason the package takes a theme rather than depending on Compound itself.
+// `Color.compound` is a single shared instance that Element Pro re-brands at runtime from remote
+// settings, so a second copy linked inside the package would never see the override and a branded
+// deployment would get a stock-coloured call screen. Every member below is a computed property, so
+// each read goes to the live instance.
+
+struct ElementCallCompoundTheme: ElementCallTheme {
```
stefanceriu: What's a StyleAdapter? This looks like it's just a theme.

ElementX/Sources/Services/NativeCall/NativeCallSystem.swift
```diff
@@ -0,0 +1,61 @@
...
+// Copyright 2026 Element Creations Ltd.
+//
+// SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial.
+// Please see LICENSE files in the repository root for full details.
+//
+
+import Combine
+import ElementCall
+
+/// Serves the call package's system-call port from `ElementCallService`.
```
stefanceriu, line 11: I honestly don't know how to parse this line.

ElementX/Sources/Services/NativeCall/ElementCallSystemAdapter.swift
```diff
@@ -0,0 +1,61 @@
...
+
+import Combine
+import ElementCall
+
+/// Serves the call package's system-call port from `ElementCallService`.
+///
+/// This stays app-side because the provider is process-wide and shared with the web-view call path:
+/// the same `CXProvider` and VoIP push registry answer both, so it cannot be handed over.
+@MainActor
+final class ElementCallSystemAdapter: ElementCallSystemProviding {
```
stefanceriu: An system adapter for providing a system.. something? 🤔 Can we call it a "ElementCallSystemInteractionProtcol" maybe so it's lcear to everybody what it does?

ElementX/Sources/Services/NativeCall/NativeCallSystem.swift
```diff
@@ -0,0 +1,61 @@
...
+/// the same `CXProvider` and VoIP push registry answer both, so it cannot be handed over.
+@MainActor
+final class ElementCallSystemAdapter: ElementCallSystemProviding {
+    private let service: ElementCallServiceProtocol
+    
+    init(service: ElementCallServiceProtocol) {
+        self.service = service
+    }
+    
+    var events: AnyPublisher<ElementCallSystemEvent, Never> {
```
stefanceriu, line 22: Looks like this ElementCallSystemEvent mixes events with requests/actions which I don't think is a good idea.

stefanceriu (commented):
Not a bad start but it looks like it was build in isolation from the EX iOS codebase itself and it ended up not respecting many of our conventions, naming schemes etc. It might be a good time to port over chunks from our Agents.md.

ElementX/Sources/Application/Navigation/NavigationTabCoordinator.swift
BillCarsonFr: I have removed this from the scope of this PR, and I will do a follow up PR.
Without this the screen share button is pretty useless, but that would be a limitation of this first integration

ElementX.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
BillCarsonFr: I'll move it, but not sure when. Tell me if it is a blocker for you

ElementX/Sources/Services/NativeCall/ElementCallOptionsAdapter.swift
BillCarsonFr: done

ElementX/Sources/Application/Settings/AppSettings.swift
BillCarsonFr: Done

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
BillCarsonFr: The comment is wrong, but the end goal of the rust-rtc stack is to be plugged in the sdk and it should notify you of any new incoming call, so it should be started early. It is still experimental so not sure where it will be at the end.

ElementX/Sources/Services/NativeCall/NativeCallSystem.swift
BillCarsonFr: Created an issue in the element-call-ios repo

ElementX/Sources/Services/NativeCall/NativeCallRoomContext.swift
BillCarsonFr: removed

ElementX/Sources/Services/NativeCall/NativeCallRoomContextAdapter.swift
BillCarsonFr: done

toshanmugaraj:
The call status should display a "Ringing" state with a ringing audio tone. The call duration timer should only start after the callee accepts/joins the call.

BillCarsonFr:
> The call status should display a "Ringing" state with a ringing audio tone. The call duration timer should only start after the callee accepts/joins the call.

Thx, this is not yet implemented. The component is still experimental, ringing will be added on next updates
Maybe you can create an issue here https://github.com/element-hq/element-call-ios/issues

ElementX.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
pixlwave: Not a blocker no, just thinking that it would be nice from the perspective of being sure something would exist when checking out old commits, but I think it would be unlikely for there not to be.

ElementX/Sources/Application/Navigation/NavigationTabCoordinator.swift
```diff
@@ -377,6 +377,7 @@ private struct NavigationTabCoordinatorView<Tag: Hashable>: View {
                     if let coordinator = navigationTabCoordinator.overlayModule?.coordinator {
                         coordinator.toPresentable()
                             .opacity(navigationTabCoordinator.overlayPresentationMode == .minimized ? 0 : 1)
+                            .allowsHitTesting(navigationTabCoordinator.overlayPresentationMode == .fullScreen)
```
pixlwave, line 380: I think this is left over and should also be removed for now?

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -226,6 +228,8 @@ class ClientProxy: ClientProxyProtocol {
             contentScanner = nil
         }
         
+        nativeCallTransport = (client as? Client).flatMap { ElementCallSDKTransport(client: $0) }
```
pixlwave, line 231: Should we guard this on `appSettings.nativeCallEnabled`? It's fine if the app needs a restart for native calls, we can include that in the Developer Option toggle's label.

ElementX/Sources/Services/ElementCall/CXProviderProtocol.swift
pixlwave, line 1: Minor nit: Could we order these logically so
- delegate
- new call methods
- updated call method
- ended call method.

ElementX/Sources/Services/ElementCall/ElementCallService.swift
```diff
@@ -26,8 +26,11 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
         let roomID: String
         let rtcNotificationID: String?
         let isVoiceCall: Bool
+        /// A native call keeps its CallKit call alive; a web view call must not (see `CXAnswerCallAction`).
```
pixlwave, line 29: This doesn't document the property, I think this should be moved to a comment at the point where `isNative` is used.

ElementX/Sources/Services/ElementCall/ElementCallService.swift
```diff
@@ -313,6 +394,10 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
         // This gets called for no reason on simulators, where CallKit
         // isn't even supported, ignore it.
         #else
+        // Logged because the system ends a call for reasons of its own, and a native call that
+        // disappears looks like the app's doing until you can see this line.
```
pixlwave, line 398: I don't think we need to explain why we logs things.

ElementX/Sources/Services/ElementCall/ElementCallService.swift
```diff
@@ -337,6 +422,9 @@ class ElementCallService: NSObject, ElementCallServiceProtocol, PKPushRegistryDe
     
     private func tearDownCallSession(sendEndCallAction: Bool = true) {
         if sendEndCallAction, let ongoingCallID {
+            // Logged so that the end call action the provider performs next can be told apart from
+            // one the system raised by itself, which otherwise look identical.
```
pixlwave, line 426: Ditto

ElementX/Sources/Services/ElementCall/ElementCallService.swift
pixlwave, line 1: Generally `unattached` seems an odd term to use to me. Maybe something like `pending` would be a reasonable description of it?

ElementX/Sources/Services/NativeCall/MatrixRTCLogBridge.swift
pixlwave, line 1: Nice, thank you.

ElementX/Sources/Services/NativeCall/NativeCallOptions.swift
pixlwave, line 1: Might just be me, but this seems a bit OTT to use a protocol here vs just passing a struct directly. Again it is fine for the feature flag to return an app restart.

ElementX/Sources/Services/NativeCall/NativeCallStyle.swift
```diff
@@ -0,0 +1,152 @@
...
+import ElementCallAll
+import SwiftUI
+
+// Colours, icons and avatars for the call, read from this app's Compound.
+//
+// Why the package takes a theme instead of linking Compound itself: `Color.compound` is a single
+// shared instance that Element Pro re-brands at runtime, and a second copy inside the package would
+// never see the override. Every member is computed so each read goes to the live instance.
+
+struct NativeCallTheme: ElementCallTheme {
```
pixlwave, line 18: Feel free to ignore this seeing as ideal end-goal will be to import compound-ios directly into Element Call but a simplification that could be made here would be:

Split this up into e.g. `ElementCallColors` and `ElementCallFonts` and extend `CompoundColors` and `CompoundFonts` to conform to those separately. Then pass `Color.compound` and `Font.compount` directly to the call package and you don't need any updates to Element X if you want to access new tokens.

UnitTests/Sources/ElementCallServiceTests.swift
```diff
@@ -240,6 +245,161 @@ final class ElementCallServiceTests {
...
+        #expect(service.ongoingCallRoomIDPublisher.value == "!room:example.com")
+        
+        // And the watchdog must be gone, otherwise it ends the call that was just taken over.
+        var endedCall = false
+        callProvider.reportCallWithEndedAtReasonClosure = { uuid, _, _ in
+            endedCall = endedCall || uuid == answeredUUID
+        }
+        clock.advance(by: .seconds(60))
+        for _ in 0..<3 {
+            await Task.yield()
```
pixlwave, line 354: Please can we avoid the use of `Task.yield()`, it rarely ever does what you want it to (especially on CI, resulting in flakey tests).

There are numerous ways to signal such as `deferFullfilment`, `deferFailure`, `waitForConfirmations`, `confirmations` etc etc. There's very likely already a pattern in the tests that could be re-used here.

ElementX/SupportingFiles/target.yml
```diff
@@ -260,6 +260,10 @@ targets:
       implicit: true
       link: false
       embed: true
+    # The native call stack. The umbrella re-exports the package's four modules, whose split is
+    # what keeps the Matrix SDK confined to ElementCallMatrix.
+    - package: ElementCall
+      product: ElementCallAll
```
pixlwave, line 266: From a module name this is a bit weird. I would've thought either
- `ElementCall` with the existing on renamed to something like `ElementCallCore`
- `ElementCallNative`

Would make perfect sense here. We also don't need the comments, that's an implementation detail as far as the app is concerned.

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
pixlwave, line 1: I think this is the only file that stands out to me as not fitting in with our standard patterns. As much as possible, a flow coordinator's sole responsibility should be for navigation. As I see it:
- The native call stack can be started and stopped, therefore it should live in `NativeCallScreenCoordinator`.
- The call screen coordinator should forward any presentation actions upwards to the flow coordinator for it to handle.
- Any other actions should be handled internally within the call screen coordinator.
- The call stack no longer needs to exist when there isn't a running call (and we don't need to think about tearing it down when signing out).

pixlwave (commented):
Nice, thank you for the last round of changes. Other than the `UserSessionFlowCoordinator` all my comments are minor.

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
BillCarsonFr: > The call stack no longer needs to exist when there isn't a running call 

The RTC stack is expected to run while the session is active, utlimatly it will notify when a call is started

ElementX/Sources/Services/Client/ClientProxy.swift
```diff
@@ -8,6 +8,7 @@
 
 @preconcurrency import Combine
 import CryptoKit
+import ElementCallAll
```
stefanceriu, line 11: There's no reason for this not to be called just plainly `ElementCall`

ElementX/Sources/Screens/Settings/DeveloperOptionsScreen/View/DeveloperOptionsScreen.swift
```diff
@@ -133,6 +133,14 @@ struct DeveloperOptionsScreen: View {
...
                         }
                     }
+                Toggle(isOn: $context.nativeCallEnabled) {
+                    Text("Experimental Native Call")
+                }
+                
+            } header: {
+                Text("Calls")
+            } footer: {
+                Text("The URL overrides where Element Call is loaded from in the web view. It has no effect on native calls.")
```
stefanceriu, line 143: This text is wrong. Also presumably these 2 won't work correctly together, might be worth pointing that out.

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
stefanceriu: Not sure what that means, a session existing doesn't mean sync is running, there's no guarantee this will tell you about an incoming call. Up to now we relied on push notifications and `PKPushRegistry` updates, why the change?

ElementX/Sources/FlowCoordinators/UserSessionFlowCoordinator.swift
```diff
@@ -522,8 +534,175 @@ class UserSessionFlowCoordinator: FlowCoordinatorProtocol {
...
+    
+    private var nativeCallStack: ElementCallStack?
+    
+    private var nativeCallController: ElementCallController? {
+        nativeCallStack?.controller
+    }
+    
+    /// Builds the call stack for this session. The arguments are the whole integration surface: a
+    /// transport, the system call provider, settings, the look and a log sink.
+    private func startNativeCallStack() {
```
stefanceriu, line 580: Oh I see, is comment above outdated now? If we can start and stop this "call stack" then it should be moved to the ElementCallService. No reason a service should call a flowCoordinator to setup service specific components.

stefanceriu (changes requested):

BillCarsonFr:
As dicussed, closing in favor of https://github.com/element-hq/element-x-ios/pull/6159 with clean history and review comments addressed

PR #6144, opened 2026-09-09
Scope dark mode change only to the presenting sheet

### LINKED ISSUE #6093: Sending an image *sometimes* causes the app to switch to dark mode
https://github.com/user-attachments/assets/5a734a57-5a90-4fcb-b8a3-aad424a5d9e2

Reporting this on behalf of a friend. This was also reproduced by another user: https://matrix.to/#/!nqXEcZbBVWpiHLmQhVANyKwuRD3d672nXjd1iEKPuTU/$dfRbXyYgM8AXVyfm_SWkZ6tCijlJkTXv7t3_lmzFNjI?via=matrix.org&via=element.io&via=tchncs.de

Velin92:
instead of applying the dark mode to the whole window
This replaces `preferredColorScheme` (which forces the window) to a custom `presentationColorScheme` which only affects the UIKit presentation part of the modal view. It takes a `ColorScheme?` like the original modifier, and maps it to the `UIUserInterfaceStyle` inside the representable, so the views don't need to know about UIKit

fixes #6093


ElementX/Sources/Other/SwiftUI/PresentationInterfaceStyleModifier.swift
```diff
@@ -0,0 +1,72 @@
...
+
+import SwiftUI
+
+extension View {
+    /// Forces the interface style of the sheet or popover containing this view without touching the
+    /// rest of the app. Does nothing when the view isn't presented modally.
+    ///
+    /// Prefer this over `preferredColorScheme` for modals: SwiftUI applies that modifier to the whole
+    /// window and can leave it applied when the presentation is torn down abnormally (#6093).
+    func presentationInterfaceStyle(_ style: UIUserInterfaceStyle) -> some View {
```
pixlwave: I guess my only question on this is whether it should take a `ColorScheme` at the SwiftUI level and map that to the `UIUserInterfaceStyle` in the representable?

ElementX/Sources/Other/SwiftUI/PresentationInterfaceStyleModifier.swift
```diff
@@ -0,0 +1,72 @@
...
+        
+        private func applyStyle() {
+            guard window != nil else { return }
+            targetViewController?.overrideUserInterfaceStyle = style
+        }
+        
+        /// The view controller that was actually presented, so the override covers its navigation bar too.
+        private var targetViewController: UIViewController? {
+            var responder = next
+            while let current = responder {
```
pixlwave: Totally unimportant, but I think there's a cool thing you can do with `sequence(first:next:)` here instead of using the while loop.

ElementX/Sources/Screens/FilePreviewScreen/View/TimelineMediaPreviewRedactConfirmationView.swift
```diff
@@ -14,7 +14,7 @@ struct TimelineMediaPreviewRedactConfirmationView: View {
     
     let item: TimelineMediaPreviewItem.Media
     @ObservedObject var context: TimelineMediaPreviewViewModel.Context
-    var preferredColorScheme: ColorScheme? = .dark
+    var interfaceStyle: UIUserInterfaceStyle = .dark
```
pixlwave: Ok yeah, I definitely think we should do the mapping in the representable now I see this being stored/computed in various places.

ElementX/Sources/Screens/MediaUploadPreviewScreen/View/MediaUploadPreviewScreen.swift
```diff
@@ -165,19 +166,13 @@ struct MediaUploadPreviewScreen: View {
             Button { context.send(viewAction: .cancel) } label: {
                 Text(L10n.actionCancel)
             }
-            // Fix a bug with the preferredColorScheme on iOS 18 where the button doesn't
```
pixlwave, line 168: Nice 😎

pixlwave (approved):
Very cool 🙌

PR #6148, opened 2026-09-09
Fix some isolation errors when compiling with Swift 6.4.

pixlwave:
This allows the project to be built with Xcode 27.

### DIFF
--- a/compound-ios/Sources/Compound/Colors/CompoundUIColors.swift
+++ b/compound-ios/Sources/Compound/Colors/CompoundUIColors.swift
@@ -6,7 +6,7 @@
 // Please see LICENSE files in the repository root for full details.
 //
 
-@preconcurrency import CompoundDesignTokens
+import CompoundDesignTokens
 import Synchronization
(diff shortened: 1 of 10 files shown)


compound-ios/Sources/Compound/Colors/CompoundUIColors.swift
```diff
@@ -6,12 +6,13 @@
 // Please see LICENSE files in the repository root for full details.
 //
 
-import CompoundDesignTokens
+@preconcurrency import CompoundDesignTokens
```
pixlwave: Should probably update the token generation instead of the `@preconcurrency` tbf.

Velin92 (approved):

compound-ios/Sources/Compound/Colors/CompoundUIColors.swift
pixlwave: https://github.com/element-hq/compound-design-tokens/pull/293

PR #6149, opened 2026-09-10
Fix some SwiftLint warnings.

pixlwave:
Looks like there's a new `legacy_swiftui_aspect_ratio` rule that whilst seemed odd to me (it isn't deprecated), is based on the Apple's docs:

https://developer.apple.com/documentation/swiftui/view/scaledtofit()

> This method is equivalent to calling [aspectRatio(_:contentMode:)](https://developer.apple.com/documentation/swiftui/view/aspectratio(_:contentmode:)) with a nil aspectRatio and a content mode of [ContentMode.fit](https://developer.apple.com/documentation/swiftui/contentmode/fit).

https://developer.apple.com/documentation/swiftui/view/scaledtofill()

> This method is equivalent to calling [aspectRatio(_:contentMode:)](https://developer.apple.com/documentation/swiftui/view/aspectratio(_:contentmode:)) with a nil aspectRatio and a content mode of [ContentMode.fill](https://developer.apple.com/documentation/swiftui/contentmode/fill).

So I ran `swiftlint lint --fix`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,22 +18,9 @@ DerivedData/
 .swiftpm
 .netrc
 
-# fastlane
-#
-# It is recommended to not store the screenshots in the git repo.
-# Instead, use fastlane to re-generate the screenshots whenever they are needed.
-# For more information about the recommended setup visit:
-# https://docs.fastlane.tools/best-practices/source-control/#source-control
-
-fastlane/report.xml
-fastlane/Preview.html
-fastlane/screenshots/**/*.png
-fastlane/test_output
-towncrier.toml.tmp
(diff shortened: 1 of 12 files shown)


.gitignore
```diff
@@ -18,22 +18,9 @@ DerivedData/
 .swiftpm
 .netrc
 
-# fastlane
```
stefanceriu, line 21: 🤘

stefanceriu (approved):

.gitignore
pixlwave: 😄

PR #6150, opened 2026-09-13
Fix: pinned timeline for inaccessible events #5702

### LINKED ISSUE #5702: Pinning a message another user does not have access to is handled incorrectly
### Steps to reproduce

steps:

1. Have a room that contains messages and an account that does not yet have access to this room

2. Invite your second account to this room (make sure he won't have access to the past messages)

3. Accept the invitation on the second account

4. Have the first account pin a message which the second account has no access to.

5. Look at the chat on the second account

### Outcome

What did you expect?
The are no pinned messages or it is not shown
In the settings of the screen you see 0 and there are non shown if you click on it
What happened instead?
- Settings > pinned messsages 0 > click > you "1 pinned message" but you do not see any messages 

### Your phone model

_No response_

### Operating system version

_No response_

### Application version

_No response_

### Homeserver

_No response_

### Will you send logs?

Yes

pr3m-kamble:
## What happened?

The pinned events timeline used `pinnedEventIDs` to determine its title count and empty state.

This includes pinned event IDs for events that the current user may not have access to, which could result in the timeline showing a pinned message count even though no corresponding message was visible.

## What did I change?

Use the event IDs from the materialized timeline item view states to determine the visible pinned message count and empty state once the timeline has loaded.

While the timeline is loading, the existing `pinnedEventIDs` are still used to avoid briefly showing an incorrect empty state.

Fixes #5702.

### Pull Request Checklist

- [ ] Voiceover enabled.


ElementX/Sources/Screens/PinnedEventsTimelineScreen/View/PinnedEventsTimelineScreen.swift
```diff
@@ -13,12 +13,22 @@ struct PinnedEventsTimelineScreen: View {
...
+            guard !pinnedEventIDs.isEmpty else {
+                return L10n.screenPinnedTimelineScreenTitleEmpty
+            }
+            return L10n.screenPinnedTimelineScreenTitle(pinnedEventIDs.count)
+        }
+        guard !visiblePinnedEventIDs.isEmpty else {
             return L10n.screenPinnedTimelineScreenTitleEmpty
         }
-        return L10n.screenPinnedTimelineScreenTitle(pinnedEventIDs.count)
+        return L10n.screenPinnedTimelineScreenTitle(visiblePinnedEventIDs.count)
```
pixlwave, line 31: I don't think this would have made sense anyway. The timeline paginates in older messages as you scroll up. So unless the pinned events were in the last 20/30 messages at the point you open the room, then the banner would stay hidden.

pixlwave (commented):
Hey @pr3m-kamble thanks for the PR.

However, I'm going to close this as we need input from Product before we handle this. For example would it be better to keep the count correct but show a "you don't have access" message instead otherwise 2 different users could be confused why one sees a pinned banner and the other does not.

We likely need an element-meta ticket for this discussion as it will affect both iOS and Android (please see the contributing guidelines for the full explanation).

PR #6157, opened 2026-09-16
Various search tab tweaks

stefanceriu:
- show breadcrumbs of previous search tab interactions instead of the empty state (when available)
- only show joined rooms in search results
- hide the segmented control until a query is entered
- clear the query when leaving the tab
- dismiss the keyboard when interacting with the results
- dismiss the whole tab when tapping the search field cancellation button (the big X on the right)


manuroe (approved):
From I am able to understand, LGTM. Thanks for this update!

> The queries the user searched for and the rooms they opened from the results, most recent first.

I would add to it the "room breadcrumbs" as we have on Element Web and Classic to ease room switching but it is another topic ;)

PR #6161, opened 2026-09-18
Tweak the BuildSDK command line tool now that ElementCall also links the SDK

stefanceriu:
Otherwise we're left with invalid package dependencies.

> unable to override package 'MatrixRustSDK' because its identity 'matrix-rust-components-swift' doesn't match override's identity (directory name) 'matrix-rust-sdk'


manuroe (approved):
Fun as I said in the chat. Well done for the fix!

PR #6170, opened 2026-09-21
Switch all clients to the new `.highEntropyPassphrase` mechanism.

stefanceriu:
Which will automatically migrate the existing passphrase and allow us to later switch directly to the raw key everywhere.

More details on https://github.com/matrix-org/matrix-rust-sdk/pull/6878


pixlwave (approved):
Can we also change the parameters to `Data` and then drop the base64 encoding in the authentication service at the very least? It would also be nice to add some logic to the decoding in the `Session` too, so that that exposes `Data` too.

ElementX/Sources/Services/Client/ClientFactory.swift
```diff
@@ -67,9 +67,9 @@ nonisolated struct ClientFactory: ClientFactoryProtocol {
             .enableAutomaticBackPagination(enableAutomaticBackPagination: appSettings.automaticBackPaginationEnabled)
             .sqliteStore(config: .init(dataPath: credentials.restorationToken.sessionDirectories.dataPath,
                                        cachePath: credentials.restorationToken.sessionDirectories.cachePath)
-                    .passphrase(passphrase: credentials.restorationToken.passphrase))
+                    .highEntropyPassphrase(passphrase: credentials.restorationToken.passphrase, base64Variant: .padded))
             .withSearchIndexStore(path: credentials.restorationToken.sessionDirectories.dataPath,
-                                  password: credentials.restorationToken.passphrase)
+                                  password: credentials.restorationToken.passphrase.base64EncodedString())
```
stefanceriu, line 72: 🤷‍♂️

ElementX/Sources/Services/UserSession/RestorationToken.swift
```diff
@@ -53,7 +58,7 @@ nonisolated extension RestorationToken: Codable {
         try container.encode(session, forKey: .session)
         try container.encode(sessionDirectories.dataDirectory, forKey: .sessionDirectory)
         try container.encode(sessionDirectories.cacheDirectory, forKey: .cacheDirectory)
-        try container.encode(passphrase, forKey: .passphrase)
+        try container.encode(passphrase.base64EncodedString(), forKey: .passphrase)
```
stefanceriu, line 61: Do you think we need more than this (plus the counterpart above)?

ElementX/Sources/Services/Client/ClientFactory.swift
pixlwave: Baby-steps I guess. Presumably this one will get a key at some stage too?

pixlwave: Also, shouldn't we have this `.withSearchIndexStore` on `makeAuthenticationClient` too??

ElementX/Sources/Services/Client/ClientFactory.swift
```diff
@@ -95,7 +95,7 @@ nonisolated struct ClientFactory: ClientFactoryProtocol {
             .systemIsMemoryConstrained()
             .sqliteStore(config: .init(dataPath: credentials.restorationToken.sessionDirectories.dataPath,
                                        cachePath: credentials.restorationToken.sessionDirectories.cachePath)
-                    .passphrase(passphrase: credentials.restorationToken.passphrase))
+                    .highEntropyPassphrase(passphrase: credentials.restorationToken.passphrase, base64Variant: .padded))
```
pixlwave, line 100: Maybe a silly question but just to be sure, we're good if the SDK were to do the migration by receiving the highEntropyPassphrase from the NSE before the app was launched right?

ElementX/Sources/Services/UserSession/RestorationToken.swift
pixlwave: We discussed, we'll add something like `key: Data` along with any additional logic when we're ready to use the key directly.

UnitTests/Sources/KeychainControllerTests.swift
```diff
@@ -141,7 +141,7 @@ struct KeychainControllerTests {
                                                                      oidcData: nil,
                                                                      slidingSyncVersion: .proxy(url: "https://sync.example.com")),
                                                   sessionDirectory: .sessionsBaseDirectory.appending(component: UUID().uuidString),
-                                                  passphrase: "passphrase",
+                                                  passphrase: Data("passphrase".utf8).base64EncodedString(),
```
pixlwave: Ok I see where I was getting confused before. I think we should replace all of these `RestorationTokenVX` values with

```suggestion
                                                  passphrase: "cGFzc3BocmFzZQ==",
```

for simplicity.

UnitTests/Sources/RestorationTokenTests.swift
pixlwave, line 1: So same here too.

UnitTests/Sources/RestorationTokenTests.swift
```diff
@@ -106,7 +106,7 @@ struct RestorationTokenTests {
                                                               oauthData: "data-from-mas",
                                                               slidingSyncVersion: .native),
                                              sessionDirectories: .init(),
-                                             passphrase: "passphrase",
+                                             passphrase: Data("passphrase".utf8),
```
pixlwave: Or maybe instead of a literal we store both somewhere like

```
enum RestorationTokenFixtures {
    static let passphrase = Data("passphrase".utf8)
    static let base64Passphrase = Data("passphrase".utf8).base64EncodedString()
}
```

and then use that for all of the tests.

pixlwave (approved):
Nice, thanks for the switch to `Data`, looks cleaner to me 🙌

ElementX/Sources/Services/Client/ClientFactory.swift
stefanceriu: Yes, I believe this is all idempotent.

ElementX/Sources/Services/Client/ClientFactory.swift
stefanceriu: My bad, I keep forgetting that's the clien that gets used after registration.

PR #6171, opened 2026-09-21
Drop Intel device support from the build scripts and project

stefanceriu:


Tools/Sources/Commands/BuildSDK.swift
```diff
@@ -11,9 +11,7 @@ enum Profile: String, ExpressibleByArgument {
 enum Target: String, ExpressibleByArgument, CaseIterable {
     case iOS = "aarch64-apple-ios"
     case simulatorARM64 = "aarch64-apple-ios-sim"
-    case simulatorIntel = "x86_64-apple-ios"
     case macARM64 = "aarch64-apple-darwin"
```
pixlwave, line 14: Technically we can drop this one too given the amount of work that would be required for real macOS support.

pixlwave (approved):
Finally! 🥳

Tools/Sources/Commands/BuildSDK.swift
stefanceriu: I'm still somewhat hoping that Apple with do some magic at some point. I'm delusional, I know 😅

PR #6172, opened 2026-09-21
TabRail layout fixes

stefanceriu:
Previously it used to either cause flickering #6107 or disappear on window resizing and never appear back


pixlwave:
> or disappear on window resizing and never appear back

I thought we fixed that? Asking because I'm not really a fan of this idea 😕

stefanceriu:
> > or disappear on window resizing and never appear back
> 
> I thought we fixed that? Asking because I'm not really a fan of this idea 😕

Yeah well no, I was able to reproduce it on the mac by just randomly resizing the windows. Couldn't find anything else that works so I just landed on it. At least it's not broken ¯\_(ツ)_/¯

pixlwave (commented):
We had a call and decided to try moving the rail into the sidebar on iOS 27 (now that Liquid Glass has dropped the inset sidebar design) so that it can continue to be hidden with the sidebar on most devices.

ElementX/Sources/Application/Navigation/NavigationCoordinators.swift
```diff
@@ -416,9 +415,11 @@ private struct NavigationSplitCoordinatorView: View {
         .animation(.elementDefault, value: navigationSplitCoordinator.sidebarModule)
         .animation(.noAnimation, value: navigationSplitCoordinator.detailModule) // Don't crossfade the detail transition on iPad.
     }
-    
-    @ViewBuilder
-    var sidebarDivider: some View {
+}
+
+/// The vertical divider that separates the sidebar from its neighbours on iOS 27, where the sidebar is no longer inset.
+struct SidebarDivider: View {
```
pixlwave: Maybe `NavigationSidebarDivider` on this?

pixlwave (approved):
Works for me 👍

PR #6177, opened 2026-09-22
Fix HDR UI bugs on iOS 27.0.

pixlwave:
It seems that putting a pure white colour into the toolbar results in it being considered as HDR (even if as with `.compound.bgCanvasDefault` it is defined as being sRGB)…

Two fixes:

- Don't tint `ToolbarButton`s that should be clear.
- Explicitly disable HDR on `LoadableAvatarImage`'s background for when it is included inside a `.toolbar`.


stefanceriu (approved):
Fascinating stuff

PR #6180, opened 2026-09-22
Fix the split view layout on iOS 27.

pixlwave:
iOS 27 no longer uses a Liquid Glass inset sidebar and goes back to the iOS 18 full-height behaviour. However unlike iOS 18 the divider has been removed in favour of a different background colour. So this PR:

- Adds our own divider as we use the same colour in both columns.
- Additionally includes the divider on the tab rail.
- Adds additional padding to the rail that was previously provided for free by the inset sidebar.
- Fixes the background colour of the rail now that the detail's background no longer extends past the sidebar.

This isn't based on a design, but it at least makes the app look semi-reasonable for now (we probably need to think about what to do in the top safe area to avoid clashing with the date/menu).

| Before | After |
| - | - |
| ipad-before | ipad-after |


stefanceriu (approved):
So much better!

PR #6183, opened 2026-09-22
Show the call buttons on macOS when the experimental native calling feature is enabled.

pixlwave:
In theory there's no need to hide the call buttons on macOS now that we're not using a web view so let's show them and test how well it works?


ElementX/Sources/Application/Windowing/WindowManagerProtocol.swift
```diff
@@ -11,6 +11,7 @@ import SwiftUI
 enum SecondaryWindowType: Hashable, Codable {
     case room(roomID: String)
     case settings
+    case call
```
stefanceriu, line 14: Oh now that's cool!

stefanceriu (approved):

PR #6188, opened 2026-09-23
Only make toolbar buttons prominent when tinted.

pixlwave:
#6177 worked fine on iOS, but macOS seemingly handles things differently. So this PR simply removes the prominent tint/treatment all together for regular buttons.


stefanceriu (approved):
Suuure 🙈

Velin92 (approved):

PR #6190, opened 2026-09-23
Multi select forward

Velin92:
This PR adds forward with multi selection, it heavily refactors the existing multi selection POC, in fact now the Select option has completely been removed in favour of having Foward start multi selection directly when the FF is on.

Screenshot iPhone 17 23-09-2026 at 19 24 48


stefanceriu:
Please rename this PR as whatever you name it will go into the change log, under the work in progress section for which I just added the label.

ElementX/Resources/Localizations/en.lproj/Untranslated.strings
```diff
@@ -1,4 +1,6 @@
 "screen_home_tab_search" = "Search";
+"screen_room_message_forwarded" = "Message forwarded";
+"screen_room_messages_forwarded" = "Messages forwarded";
```
stefanceriu, line 3: Best move these to localazy (and parameterize them?)

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -1293,6 +1294,18 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
         }
     }
     
+    /// Ends the message selection in the timeline that started the forwarding, now that it has completed.
+    private func clearSelection(in state: State) {
+        switch state {
+        case .room:
+            roomScreenCoordinator?.clearSelection()
+        case .thread:
+            childThreadScreenCoordinators.last?.clearSelection()
```
stefanceriu, line 1303: I wonder if we should allow threads to forward into their own parent room, feels like it might be useful to resurface old talked about content 🤔

stefanceriu: Somewhat related, I remember that slack had a thread option to "also send in room" which was useful at times.

ElementX/Sources/Screens/Timeline/TimelineViewModel.swift
```diff
@@ -198,7 +198,7 @@ class TimelineViewModel: TimelineViewModelType, TimelineViewModelProtocol {
             }
         case .displayTimelineItemMenu(let itemID):
             timelineInteractionHandler.displayTimelineItemActionMenu(for: itemID)
-        case .handleTimelineItemMenuAction(let itemID, .selectMessages):
+        case .handleTimelineItemMenuAction(let itemID, .forward) where state.canSelectMessages:
```
stefanceriu, line 201: It should never reach this point if `state.canSelectMessages` is false, no?

ElementX/Sources/Screens/Timeline/TimelineModels.swift
```diff
@@ -207,7 +209,7 @@ struct TimelineViewStateBindings {
 
 /// The state of the multi-selection of messages, active as soon as an item is selected.
 struct TimelineSelectionState: Equatable {
```
stefanceriu, line 211: I'm finding this `TimelineSelection` naming scheme a bit lacking, too generic. Can we rename it throughout to `TimelineMultiMessageSelection` or something along those lines?

ElementX/Sources/Screens/Timeline/TimelineModels.swift
```diff
@@ -207,7 +209,7 @@ struct TimelineViewStateBindings {
...
-    static let limit = 30
+    static let limit = 10
     
     /// Mirrors the `messageMultiSelectEnabled` feature flag.
```
stefanceriu, line 214: Such an unnecessary comment 😅

ElementX/Sources/Screens/Timeline/TimelineModels.swift
stefanceriu, line 229: Why would it ever go over the limit?!

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -1293,6 +1294,18 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
         }
     }
     
+    /// Ends the message selection in the timeline that started the forwarding, now that it has completed.
+    private func clearSelection(in state: State) {
```
stefanceriu, line 1298: I'm sometimes wondering if we should revamp the whole action system so it supports callbacks/async finishes and we no longer have to expose these kinds of things. I'm not a fan of having all kinds of public methods for internal functionality. In this particular case it would get better if we rename "Selection" to the thing I wrote about here elsewhere.

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
```diff
@@ -1293,6 +1294,18 @@ class RoomFlowCoordinator: FlowCoordinatorProtocol {
...
+        switch state {
```
stefanceriu, line 1299: Does it really matter what state we come from, can't we just `clearSelection` throughout?

ElementX/Sources/Screens/FilePreviewScreen/TimelineMediaPreviewViewModel.swift
```diff
@@ -136,7 +136,7 @@ class TimelineMediaPreviewViewModel: TimelineMediaPreviewViewModelType {
     }
     
     private func forwardItem(itemID: TimelineItemIdentifier) async {
-        guard let forwardingItem = await timelineViewModel.makeForwardingItem(for: itemID) else { return }
+        guard let forwardingItem = await timelineViewModel.makeForwardingItem(for: [itemID]) else { return }
```
stefanceriu, line 139: Surely we sould rename this `MessageForwardingItem` not that it contains **multiple** items, right?

ElementX/Sources/Screens/ThreadTimelineScreen/View/ThreadTimelineScreen.swift
```diff
@@ -44,14 +44,20 @@ struct ThreadTimelineScreen: View {
...
-                    .collapsedInPlace(isSelectionActive)
+                VStack(spacing: 0) {
+                    composer
+                        .padding(.top, 8)
+                        .background(Color.compound.bgCanvasDefault.ignoresSafeArea())
+                        .environmentObject(timelineContext)
+                        .environment(\.timelineContext, timelineContext)
+                        // Make sure the reply header honours the hideTimelineMedia setting too.
+                        .environment(\.shouldAutomaticallyLoadImages, !timelineContext.viewState.hideTimelineMedia)
+                        .collapsedInPlace(isSelectionActive)
```
stefanceriu, line 55: What's the context around this `collapsedInPlace`, on what level would those "onAppear`/`onDisappear` side effects fire" happen?

ElementX/Sources/Services/Timeline/TimelineItems/EventBasedTimelineItemProtocol.swift
```diff
@@ -28,17 +28,14 @@ nonisolated extension EventBasedTimelineItemProtocol {
...
-    /// Whether the item can be part of a multi-selection of messages.
-    var isBulkSelectable: Bool {
         isRemoteMessage
             && !isRedacted
             && !hasFailedDecryption
             && !(self is StateRoomTimelineItem)
             && !(self is UnsupportedRoomTimelineItem)
+            && !(self is PollRoomTimelineItem)
             && !(self is LiveLocationRoomTimelineItem)
+            && !(self is StickerRoomTimelineItem) // Stickers aren't m.room.message events so there is no content to forward.
```
stefanceriu, line 38: I don't understand this comment, so what if it's not an `m.room.message`? 🤔  I most certainly expect stickers to be forwardable and as a matter of fact they do work on element web.

stefanceriu (commented):
Looks and works surprisingly well, nicely done!

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
Velin92: That's a good point! I'll mention it to product and design, we may include it in another PR.

ElementX/Sources/Screens/ThreadTimelineScreen/View/ThreadTimelineScreen.swift
Velin92: before we were just hiding the composer, but this made the onDisappear get called which triggered stuff like saving a draft, or clear the currently stored voice message

ElementX/Sources/Services/Timeline/TimelineItems/EventBasedTimelineItemProtocol.swift
Velin92: I think this is an SDK issue, the stickers are not considered msg like By the SDK so they can't be forwarded. Is not an m.room.message event issue, more of the SDK lacking an implementation that allows stickers to work properly. I should update this comment to explain it better.

ElementX/Sources/Screens/ThreadTimelineScreen/View/ThreadTimelineScreen.swift
stefanceriu: Makes sense, thanks!

ElementX/Sources/Services/Timeline/TimelineItems/EventBasedTimelineItemProtocol.swift
Velin92: Okay so the issue is that: 

> Forwarding goes through Timeline.send, which only takes m.room.message content, and stickers are m.sticker events, so there's no content to hand it. On develop, Forward was already shown on stickers but did nothing when tapped. 
> Android hides it for the same reason (element-x-android#2161).

ElementX/Sources/Screens/Timeline/TimelineModels.swift
Velin92: nah it can't go over the limit normally but the >= is a generally a safer check than == so I would prefer to keep it.

ElementX/Sources/Services/Timeline/TimelineItems/EventBasedTimelineItemProtocol.swift
stefanceriu: That's so funky but okay, fair enough, let's leave it alone for now.

ElementX/Sources/FlowCoordinators/RoomFlowCoordinator.swift
Velin92: noted, we should do this in the future

